PR #12182, opened 2018-06-04
Stellar integration: First pass, sagas and reducers

cjb:
@keybase/react-hackers 

This PR is to unblock us to start hooking up component containers -- here's enough code to load my wallets, assets and payments into the store.  Better error handling's needed, and the pagination and pending payment RPCs aren't in on the core side yet.

(The idea will be to have a separate RPC to get pending payments, so that you can always pull out pending ones without having to worry about pagination.)

Here's a screenshot of what `store.wallets` looks like with these immutable records converted to JS:

screen shot 2018-06-03 at 11 14 53 pm

The new actions are:

* `loadWallets()` - load every wallet
* `loadAssets(accountID)` - load every asset in a given wallet
* `loadPayments(accountID)` - load every payment in a given wallet
* `loadEverything()` - debugging only: load all wallets, then all assets and all payments

The `loadEverything()` action is hooked up to the (feature-flagged) Wallets nav, so just clicking on the nav icon is enough to fully populate the store.  When we're merging in component containers, they should only be calling actions to load for the account they need data for, rather than using this debug action.


cjb:
@keybase/react-hackers bump

shared/actions/wallets.js
```diff
@@ -1,15 +1,67 @@
...
+  const {accountID} = action.payload
+  return Saga.put(
+    WalletsGen.createPaymentsReceived({
+      accountID,
+      payments: res.map(elem => Constants.paymentResultToPayment(elem)),
+    })
+  )
+}
+
+function* loadEverything(action: WalletsGen.LoadEverythingPayload) {
```
chrisnojima: lets just wrap this in __DEV__ so it never gets released

shared/constants/types/wallets.js
```diff
@@ -9,6 +9,50 @@ export type _Reserve = {
 export type Reserve = I.RecordOf<_Reserve>
 
 export type _State = {
-  hello: string,
+  assetsMap: any,
```
chrisnojima: lets not leave the anys in here

shared/constants/wallets.js
```diff
@@ -1,14 +1,133 @@
...
+    sourceType,
+    statusDescription,
+    statusDetail,
+    statusSimplified,
+    target,
+    targetType,
+    time,
+    worth,
+    worthCurrency,
+  } = w.payment
```
chrisnojima: i don't think this kind of destructuring very readable, like it helps makes the lines below less wide but adds 15 lines to the file. if we did
`const p = w.payment` then below

```
id: p.id,
note: p.note,
``` 
etc that doesn't seem so bad to me.

You don't have to change this as its a style thing but i find if the file is longer lines wise its just harder to consume

shared/reducers/wallets.js
```diff
@@ -9,11 +10,23 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
...
     case WalletsGen.resetStore:
       return initialState
-    case WalletsGen.walletsRefresh:
-      return state.set('hello', String(Math.random()))
+    case WalletsGen.walletsReceived:
+      const walletMap = I.Map(action.payload.wallets.map(wallet => [wallet.accountID, wallet]))
+      return state.set('walletMap', walletMap)
+    case WalletsGen.assetsReceived:
+      const {assets} = action.payload
+      return state.update('assetsMap', assetsMap => assetsMap.set(action.payload.accountID, I.List(assets)))
```
chrisnojima: i think this can be
```
state.setIn(['assetsMap', action.payload.accountID], I.List(action.payload.assets))
```

shared/reducers/wallets.js
```diff
@@ -9,11 +10,23 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
...
+    case WalletsGen.paymentsReceived:
+      const {payments} = action.payload
+      return state.update('paymentsMap', paymentsMap =>
```
chrisnojima: same here

shared/wallets/container.js
```diff
@@ -1,26 +1,27 @@
...
   refresh: dispatchProps.refresh,
   title: 'Wallets',
 })
 
-export default connect(mapStateToProps, mapDispatchToProps, mergeProps)(
-  isMobile ? HeaderHoc(Wallets) : Wallets
-)
+export default compose(
+  connect(mapStateToProps, mapDispatchToProps, mergeProps),
+  lifecycle({
```
chrisnojima: we can leave this as its temp but fyi we're avoiding lifecycle as its very untyped

chrisnojima:
also i'd run flow coverage on these files just to make sure nothing is surprisingly untyped. might as well get good coverage while its simple

cjb:
@chrisnojima Thanks!  Made all of these changes, ready for another look.

Also renamed "wallets" to "accounts" per Patrick, the nomenclature is that you have one wallet with many accounts inside.

shared/actions/wallets.js
```diff
@@ -1,15 +1,72 @@
...
+
+const loadAccountsSuccess = (res: ?Array<Types.Account>) =>
+  Saga.put(
+    WalletsGen.createAccountsReceived({
+      accounts: (res || []).map(account => Constants.accountResultToAccount(account)),
+    })
+  )
+
+const loadAssets = (action: WalletsGen.LoadAssetsPayload) => {
+  const {accountID} = action.payload
```
chrisnojima: if we don't have these lines to extract the one param they could all just return the non {} version, aka

```
const loadAssets = (action: WalletsGen.LoadAssetsPayload) = 
  Saga.call(RPCTypes.localGetAccountAssetsLocalRpcPromise, {accountID: action.payload.accountID})
```

but just a style thing

shared/constants/types/wallets.js
```diff
@@ -8,7 +8,53 @@ export type _Reserve = {
...
+export type Account = I.RecordOf<_Account>
+
+export type Assets = I.RecordOf<_Assets>
+
+export type Payment = I.RecordOf<_Payment>
+
 export type _State = {
-  hello: string,
+  accountMap: I.Map<AccountID, _Account>,
+  assetsMap: I.Map<AccountID, I.List<_Assets>>,
```
chrisnojima: these should be Assets not _Assets right, same for the others

cjb:
@chrisnojima Thanks, done too.

cjb:
(I think I must like destructuring so much because it removes an opportunity to mess up the right-hand side..)

PR #12183, opened 2018-06-04
Remove unused files.

adamjspooner:
It doesn't look like these are used. @chrisnojima? @patrickxb?


chrisnojima:
We shouldn't just delete these files. We should update the protocol script to do this on every run

patrickxb:
Why?  They don't get generated here anymore.

chrisnojima:
Ok I wasn't sure if they were still part of the process. We don't use any
json files from this flow

patrickxb:
Should we remove all the json files then?

adamjspooner:
@chrisnojima, so, it's safe to remove these?

adamjspooner:
These files aren't recreated after `make`ing. `protocol/bin/flow.js` uses the JSON files under `protocol/json/<thing>1/`, not the JSON files in `protocol/json/`.

patrickxb:
Trash them!

adamjspooner:

PR #12186, opened 2018-06-04
Add GetWalletSettingsLocal RPC, based on GetUserSettingsLocal

zapu:
We used to have `GetUserSettingsLocal` that returned `disclaimerAccepted` boolean, but the spec talks about different set of RPCs. This PR adds `GetWalletSettingsLocal` that conforms to the spec, but we still need to make the change how `displayCurrency` works - designs want it per user (wallet), not per account.


go/stellar/stellar.go
```diff
@@ -47,7 +62,10 @@ func CreateWallet(ctx context.Context, g *libkb.GlobalContext) (created bool, er
 	default:
 		return false, err
 	}
-	return true, err
+	if err := setNewWalletDisplayCurrency(ctx, g, clearBundle); err != nil {
+		g.Log.CWarningf(ctx, "Failed to set default currency for primary account: %s", err)
+	}
```
mlsteele: Why do this on the client? If it's done on the server it can be guaranteed to succeed and we can change it whenever.

zapu:
My thought behind this is that we can be mode creative on the client on how to detect current region/locale. But we can also send additional info in headers. What do you think?

go/stellar/stellar.go
```diff
@@ -21,6 +21,21 @@ import (
 	"github.com/stellar/go/amount"
 )
 
+// setNewWalletDisplayCurrency takes initial bundle that user creates
+// and sets primary account display currency based on OS region
+// settings (that part's TODO - right now it just sets "USD").
```
mlsteele: This is a pretty far reaching TODO. I'd say just inline the function and if you want add the comment about what it really does to CreateWallet's comment. To implement that region detection deserves its own ticket.

protocol/avdl/stellar1/local.avdl
```diff
@@ -91,11 +91,13 @@ protocol local {
   // OutsideCurrencyCode examples: "USD", "EUR". Has to be one of
   // supported currencies, returned from getDisplayCurrenciesLocal.
   void changeDisplayCurrencyLocal(int sessionID, AccountID accountID, OutsideCurrencyCode currency);
+  OutsideCurrencyCode getDisplayCurrencyLocal(int sessionID, AccountID accountID);
```
mlsteele: I tried to imagine what frontend would use this RPC for. Maybe they'd want it to return `FormatCurrencyLabel` as well since they won't format it. Or maybe currency information would just come along for the ride in a bigger RPC, or at the same time as the other choices? But I'm just imagining. Ask frontend if this is useful its current form?

mlsteele:
> My thought behind this is that we can be mode creative on the client on how to detect current region/locale. But we can also send additional info in headers. What do you think?

Ah interesting. This is fine

protocol/avdl/stellar1/local.avdl
zapu: Good point. This wasn't in the doc at all because it wasn't on the same page regarding where displayCurrency actually belong (user or account). So I just wanted to add anything. I'll make it return more stuff instead of just code and ask them if it's fine.

zapu:
I'm leaning towards moving this to the server (saving one RPC upon init) as additional request parameter. So the client could determine its preferred currency and send it along with the initial bundle in the same request.

protocol/avdl/stellar1/local.avdl
mlsteele: Don't make it return more stuff just because of my theories about frontend.

zapu:
Ah nevermind, after attempting this, code is much nicer in current state rather than with additional parameter for stellar/bundle and key/multi, on both client and the server.

go/stellar/stellar.go
```diff
@@ -62,8 +47,13 @@ func CreateWallet(ctx context.Context, g *libkb.GlobalContext) (created bool, er
...
 	}
-	if err := setNewWalletDisplayCurrency(ctx, g, clearBundle); err != nil {
-		g.Log.CWarningf(ctx, "Failed to set default currency for primary account: %s", err)
+	primary, err := clearBundle.PrimaryAccount()
+	if err != nil {
+		g.Log.CErrorf("We've just posted a bundle that's missing PrimaryAccount: %s", err)
+		return err
+	}
+	if err := remote.SetAccountDefaultCurrency(ctx, g, primary.AccountID, "USD"); err != nil {
+		return err
```
mlsteele: Shouldn't this still be a warning?

zapu: crap, thanks. I got distracted.

PR #12193, opened 2018-06-04
implement new folder creation

songgao:
Depends on https://github.com/keybase/client/pull/12041

This lays ground work for renames too, which I'll do in a different PR maybe after the upload stuff.

~Not done:~

~Auto select doesn't work on mobile. @chrisnojima any hint for how to fix this?~

screen shot 2018-06-04 at 4 08 11 pm

screen shot 2018-06-04 at 4 08 14 pm


shared/actions/fs/index.js
```diff
@@ -334,6 +334,58 @@ function* _loadMimeType(path: Types.Path) {
...
 
+const commitEdit = (action: FsGen.CommitEditPayload, state: TypedState) => {
+  const {editID} = action.payload
+  const edit = state.fs.edits.get(editID)
+  if (!edit) {
+    return null
+  }
+  const {parentPath, name, type} = edit
+  switch (type) {
+    case 'new-folder':
```
chrisnojima: before i get too far into this pr, is this how you want this to be structured? a generic 'edit' with these sub types ? instead of it just being a high level type. aka why not just have this be a NewFolderPayload etc

songgao: It seems to be something that can be shared between this and rename, and maybe upload (I.e. new entry) as well. Plus it needs to be plumbed into the store. I guess we could also have separate maps for new folder creation and renames if they were separate independent types, but then we still need to check for them separately in containers.

To be clear, I am planning to have separate types for rename, but both satisfying the Edit type.

Does that sound good to you?

shared/common-adapters/input.desktop.js
```diff
@@ -134,6 +135,7 @@ class Input extends React.PureComponent<Props, State> {
   focus = () => {
     const n = this._input
     n && n.focus()
+    this.props.selectTextOnFocus && this.select()
```
chrisnojima: still marked as mobile only in the flow file

shared/fs/row/editing-container.js
```diff
@@ -0,0 +1,47 @@
...
+import Editing from './editing'
+
+type OwnProps = {
+  editID: Types.EditID,
+  routePath: I.List<string>,
+}
+
+const mapStateToProps = (state: TypedState, {editID}: OwnProps) => {
+  const _edit = state.fs.edits.get(editID, Constants.makeNewFolder()) // TODO make missing get better
+  const _username = state.config.username || undefined
```
chrisnojima: why need undefined

shared/fs/row/editing-container.js
```diff
@@ -0,0 +1,47 @@
...
+  return {
+    _username,
+    _edit,
+  }
+}
+
+const mapDispatchToProps = (dispatch: Dispatch, {editID, routePath}: OwnProps) => ({
+  onSubmit: () => dispatch(FsGen.createCommitEdit({editID})),
+  onUpdate: name => dispatch(FsGen.createNewFolderName({editID, name})),
```
chrisnojima: you need to type name else its untyped. flow coverage will show that

shared/fs/row/still.js
```diff
@@ -0,0 +1,147 @@
...
+        color: globalColors.black_40,
+      },
+      '& .fs-path-item-hover-icon:hover': {
+        color: globalColors.black_60,
+      },
+    })
+
+const RowMeta = ({badgeCount, isNew, isIgnored, needsRekey, resetParticipants}) => {
+  if (isIgnored || !(isNew || isIgnored || needsRekey || badgeCount || resetParticipants)) {
+    return <Box />
```
chrisnojima: why box vs null?

shared/fs/row/still.js
```diff
@@ -0,0 +1,147 @@
...
+        <Box style={rowStyles.itemBox}>
+          <Text
+            type={props.itemStyles.textType}
+            style={rowStyles.rowText(props.itemStyles.textColor)}
+            lineClamp={isMobile ? 1 : undefined}
+          >
+            {props.name}
+          </Text>
+          {props.type === 'folder' && (!props.resetParticipants || props.resetParticipants.length === 0) ? (
+            <Box />
```
chrisnojima: same

shared/reducers/fs.js
```diff
@@ -38,6 +39,15 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
           return original
         }
+
+        toRemove = toRemove.concat(
+          original.children
+            .filter(child => !item.children.includes(child))
+            .toArray()
+            .map(name => Types.pathConcat(path, name))
+        )
+        console.log(`Removing entries in state.fs.pathItems: ${JSON.stringify(toRemove)}`)
```
chrisnojima: 🔥

shared/reducers/fs.js
```diff
@@ -155,6 +167,44 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
         ]
       }, [])
       return state.mergeIn(['pathItems'], resetsToMerge)
+    case FsGen.newFolderRow:
+      const {parentPath} = action.payload
+      const parentPathItem = state.pathItems.get(parentPath, Constants.makeUnknownPathItem())
+      if (parentPathItem.type !== 'folder') {
+        console.warn(`bad parentPath: ${Types.pathToString(parentPath)}`)
```
chrisnojima: we might not want to log people's paths

shared/reducers/fs.js
```diff
@@ -155,6 +167,44 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
+              name: newFolderName,
+              hint: newFolderName,
+              parentPath,
+            }),
+          ],
+        ]
+      )
+    case FsGen.newFolderName:
+      return state.updateIn(
+        // $FlowFixMe
```
chrisnojima: what is flow saying here and below

shared/fs/row/still.js
songgao: I think this might be from before the refactor but I’ll try changing it to null.

shared/reducers/fs.js
songgao: Does it mean this is on fire or Song your pants will be on fire if you do this? I’m guessing later. But how do I log this? Just the array itself or construct an object that contains the array?

shared/reducers/fs.js
songgao: We already log paths in KBFS logs though. But I guess I can remove the path here.

shared/reducers/fs.js
songgao: Flow says:

```
Cannot call state.updateIn because EditID [1] is incompatible with key set [2] in type argument K2.

     reducers/fs.js
      193│         ]
      194│       )
      195│     case FsGen.newFolderName:
      196│       return state.updateIn(
      197│         ['edits', action.payload.editID],
      198│         editItem => editItem && editItem.set('name', action.payload.name)
      199│       )
      200│     case FsGen.editSuccess:
      201│     case FsGen.discardEdit:
      202│       // $FlowFixMe

     actions/fs-gen.js
 [1]  123│   editID: Types.EditID,

     node_modules/immutable/dist/immutable.js.flow
 [2] 1722│   updateIn<K: $Keys<T>, K2: $KeyOf<$ValOf<T, K>>, S: $ValOf<$ValOf<T, K>, K2>>(keyPath: [K, K2], updater: (value: $ValOf<$ValOf<T, K>, K2>) => S): this & T;
```

songgao:
🍄

songgao:
🏓

PR #12202, opened 2018-06-05
Disallow --public and --exploding-lifetime

joshblum:
fixes error if you tried to send an ephemeral message with `--public` set for `keybase chat send` and `keybase chat upload`

Help:

```
$ kb chat send --help
Running `go install github.com/keybase/client/go/keybase`... Done.
NAME:
   keybase chat send - Send a message to a conversation

USAGE:
   keybase chat send [command options] [conversation [message]]

OPTIONS:
   --topic-type "chat"          Specify topic type of the conversation. Has to be chat or dev
   --channel                    Specify the conversation channel.
   --public                     Only select public conversations. Exclusive to --private
   --private                    Only select private conversations. Exclusive to --public. If both --public and --private are present, --private takes priority.
   --set-headline               Set the headline for the conversation
   --clear-headline             Clear the headline for the conversation
   --nonblock                   Send message without success confirmation
   --exploding-lifetime "0s"    Make this message an exploding message and set the lifetime for the given duration.
                                The maximum lifetime is 168h0m0s (one week) and the minimum lifetime is 30s. Cannot be used in conjunction with --public.
 ```
error handling:
```
$ kb chat send --exploding-lifetime=30s --public test3 hi
Running `go install github.com/keybase/client/go/keybase`... Done.
▶ ERROR Cannot send ephemeral messages with --public set.
```

### DIFF
--- a/go/chat/errors.go
+++ b/go/chat/errors.go
@@ -80,6 +80,18 @@ func (e EphemeralUnboxingError) Error() string {
 
 //=============================================================================
 
+type EphemeralKeyError struct{}
+
+func NewEphemeralKeyError() EphemeralKeyError {
+	return EphemeralKeyError{}
+}
+
+func (e EphemeralKeyError) Error() string {
+	return "Cannot use ephemeral messages for a public team."
+}
+
+//=============================================================================
+
 type ConsistencyErrorCode int
 
 const (
--- a/go/chat/errors.go
+++ b/go/chat/errors.go
@@ -80,13 +80,13 @@ func (e EphemeralUnboxingError) Error() string {
 
 //=============================================================================
 
-type EphemeralKeyError struct{}
+type PublicTeamEphemeralKeyError struct{}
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: go/chat/errors.go [hunk 1]
```diff
@@ -80,6 +80,18 @@ func (e EphemeralUnboxingError) Error() string {
 
 //=============================================================================
 
+type EphemeralKeyError struct{}
```
oconnor663: Maybe name this type something public-specific?

joshblum: sgtm!

dabura667:
***Curiosity peek***

I still don't think I fully understand what --public does, so it's hard for me to grasp why this needs to throw an Error. If anyone has time to explain --public or link me to a doc explaining the difference (the help basically says it's the opposite of --private and is mutually exclusive, which doesn't help me understand much) and why this doesn't jive with exploding messages, I would appreciate it.

Thanks! Can't wait for these new features!

PR #12203, opened 2018-06-05
check if pipe is served by administrators alias

zanderz:
Check it out, @maxtaco , I think this might fix it. Will have some users test a build with this first though.

Here's the relevant debug from a session which fails without this change:
```
2018-06-05T08:39:53.256982-07:00 - [DEBU keybase pipeowner_windows.go:98] 010 Pipe account: Administrators, BUILTIN, 4, <nil>
2018-06-05T08:39:53.259974-07:00 - [DEBU keybase pipeowner_windows.go:99] 011 User account: Steve, SteveSamsung9, 1, <nil>
```


go/libkb/pipeowner_windows.go
```diff
@@ -85,10 +86,21 @@ func IsPipeowner(name string) (bool, error) {
...
 			return false, err // return original busy error
 		}
-		fileSid, err = GetFileUserSid(name)
+		pipeSid, err = GetFileUserSid(name)
 	}
 	if err != nil {
 		return false, err
 	}
-	return windows.EqualSid(fileSid, userSid), nil
+	isEqual := windows.EqualSid(pipeSid, userSid)
```
maxtaco: Can you throw some more debugging in right here too?

go/libkb/pipeowner_windows.go
```diff
@@ -68,13 +69,13 @@ func GetFileUserSid(name string) (*windows.SID, error) {
 	return userSID, nil
 }
 
-func IsPipeowner(name string) (bool, error) {
+func IsPipeowner(log logger.Logger, name string) (bool, error) {
```
maxtaco: it would be nice to put debugs on the entry and exit of this function.

zanderz: I don't think that helps us but will do

go/libkb/pipeowner_windows.go
zanderz: The caller always prints out that error someplace, do you want it debug logged here too, or...?

maxtaco:
ok, whatever logging you can, to hunt down further problems here

also, failed CI due to `go fmt` on `socket_windows`

PR #12204, opened 2018-06-05
[KBFS-2893] Reset Folder Banner

jzila:
This enables the reset panel for Files.

Everything should work except badging.


shared/constants/types/fs.js
```diff
@@ -55,6 +56,7 @@ export type _FolderPathItem = {
   children: I.Set<string>,
   favoriteChildren: I.Set<string>,
   resetParticipants: Array<string>,
+  teamID?: RPCTypes.TeamID,
```
songgao: Can this go into tlfMeta?

Also I wonder if we should have a Types.TlfPathItem so that we don't have to put TLF metadata into normal Folders.

shared/fs/header/header.js
```diff
@@ -80,6 +81,7 @@ const FolderHeader = ({
       )}
     </Box>
     <ConnectedFilesBanner path={path} />
+    <ConnectedResetBanner path={path} />
```
songgao: Have you tried having them together and see if they'd become too big? If so perhaps we can hide the FilesBanner if ResetBanner needs to show.

shared/fs/index.js
```diff
@@ -23,16 +25,20 @@ class Files extends React.PureComponent<FolderProps> {
...
+      </Box>
+    ) : this.props.progress === 'pending' ? (
+      <List items={['1', '2', '3']} renderItem={this._renderRowPlaceholder} />
+    ) : this.props.items.length === 0 ? (
+      <Box style={stylesEmptyContainer}>
+        <Text type="BodySmall">This is an empty folder.</Text>
+      </Box>
+    ) : (
+      <List items={this.props.items} renderItem={this._renderRow} />
+    )
```
songgao: Ugh there's gonna be quite some conflicts here. I'll wait until yours is merged before merging mine.

shared/constants/types/fs.js
jzila: Yeah, good call.

shared/fs/header/header.js
jzila: Also good call.

shared/fs/index.js
jzila: :(

shared/actions/fs/index.js
```diff
@@ -477,6 +479,17 @@ function* loadResets(action: FsGen.LoadResetsPayload): Saga.SagaGenerator<any, a
   yield Saga.call(resetRpc.run)
 }
 
+function letResetUserBackIn(action: FsGen.LetResetUserBackInPayload) {
```
chrisnojima: in actions we prefer arrows for all non generators since they can be more concise.

shared/common-adapters/usernames.js
```diff
@@ -25,6 +25,23 @@ function usernameText({
   inlineGrammar = false,
   showAnd = false,
 }: Props) {
+  const andStyle = collapseStyles([
```
chrisnojima: we should move the static styles into a thing below and use stylesheetcreate

shared/fs/reset-banner/container.js
```diff
@@ -0,0 +1,45 @@
...
+import Banner from '.'
+
+const mapStateToProps = (state: TypedState, {path}) => {
+  const _pathItem = state.fs.pathItems.get(path, Constants.makeUnknownPathItem())
+  const _username = state.config.username || undefined
+  return {
+    _pathItem,
+    _username,
+    path,
+    isUserReset:
```
chrisnojima: if you're going to bother to pass down _pathitem and _username you can calculate this in mergeProps instead

shared/fs/reset-banner/index.js
```diff
@@ -0,0 +1,147 @@
...
+        <Text type="BodySemiboldLink" style={{...bottomTextStyle, textDecorationLine: 'underline'}}>
+          {/* TODO: put in a link to opening a folder here. */}
+          open a folder without any of them.
+        </Text>
+      </Box>
+    </Box>
+  )
+}
+
+const bannerStyle = {
```
chrisnojima: should start using stylesheetcreate

shared/constants/types/fs.js
jzila: Oh actually, I thought of doing that but it complicates the `listFavorites` result coming back, since that just drop-in replaces `tlfMeta`. I'd have to selectively replace parts of `tlfMeta` otherwise.

jzila: `TlfPathItem` is a great idea but it'll be a headache to make work.

jzila: Yeah that is a refactor that should go in its own PR.

songgao: Sounds good. When we are done with these changes, I'll try to do that refactor (so that it doesn't cause conflicts)

jzila: I can do it, we just need to plan it separately.

shared/common-adapters/usernames.js
jzila: I can't figure out a way to get Flow to be happy with using `styleSheetCreate`:
```
Cannot call styleSheetCreate with object literal bound to map because:
 • property textDecoration is missing in _StylesCrossPlatform [1] but exists in object literal [2] in property
   andStyle.
```

chrisnojima:
Maybe your missing the two incorrectly. Collapse is just for components.
Stylehseetecreate is for defining static styles

jzila:
@chrisnojima nope, it yells at me at the style declaration level. I think it's because I'm using style properties that only exist on desktop and filtering them via `isMobile`.

jzila:
@chrisnojima ready for another review.

jzila:
I'd like to get this in before the end of sprint.

jzila:
@chrisnojima ping

shared/common-adapters/usernames.js
chrisnojima: you need to use platformStyles and not isMobile checks

shared/fs/banner/fileui-banner/container.js
```diff
@@ -0,0 +1,53 @@
...
+    onDismiss: () => dispatch(FsGen.createSetFlags({showBanner: false})),
+    onInstall: () => dispatch(FsGen.createInstallFuse()),
+    onUninstall: () => dispatch(FsGen.createUninstallKBFSConfirm({onSuccess: uninstall})),
+    _openInFileUI: (path: Types.Path) => dispatch(FsGen.createOpenInFileUI({path: Types.pathToString(path)})),
+  }
+}
+
+const mergeProps = (stateProps, dispatchProps, {path}: OwnProps) => ({
+  ...stateProps,
+  ...dispatchProps,
```
chrisnojima: i think its better to just extract the props you're using instead of spreading. you're going to get _openInFileUI in your component if you do it like this

jzila:
@chrisnojima ok got your feedback in, `styleSheetCreate` and all.

PR #12205, opened 2018-06-05
use UPAK loads in hourly check tracking

maxtaco:


maxtaco:
in some places we need a loadMe.  If you ever want to examine the sigchain or run identify, then you want a full user.  but if you just want to see device keys or tracks, you can use a UPAK. So *yes* if you have seen other call sites, let's try to migrate them.  

BTW, there's an environment flag to track down where the `LoadUsers` are coming from, when last I went through and tried to reduce their number: `--vdebug dump-site-load-user`

PR #12206, opened 2018-06-05
send ibms injected locally from an on-disk queue CORE-7921

mmaxim:
Patch does the following:

1.) Add a new "outbox" to be managed by the Gregor server client.
2.) The outbox is stored/restored in the normal `Save`/`Restore` flow for the Gregor state.
3.) Add `ConsumeMessage` to `Client` for use by `gregorHandler` to send new messages. For OOBMs, they go immediately to the server. For IBMs, they get put in the outbox and the send loop gets triggered.
4.) When calling `StateMachineState` on `Client`, it can optionally apply the messages in the outbox to yield the final version. The call the frontend makes will use this method.
5.) Throw up a `pushState` call to the frontend whenever they call one of the functions that can modify state. They only act on these messages, not the success of a call they make. 
6.) Some other random cleanup.

cc @buoyad (although you don't need to do anything here, the APIs don't change).

### DIFF
--- a/go/gregor/client/client.go
+++ b/go/gregor/client/client.go
@@ -16,8 +16,8 @@ import (
 )
 
 type LocalStorageEngine interface {
-	Store(gregor.UID, []byte, [][]byte, [][]byte) error
-	Load(gregor.UID) ([]byte, [][]byte, [][]byte, error)
+	Store(uid gregor.UID, state []byte, outbox [][]byte, localdismissals [][]byte) error
+	Load(uid gregor.UID) (state []byte, outbox [][]byte, localdismissals [][]byte, err error)
 }
 
 type TestingEvents struct {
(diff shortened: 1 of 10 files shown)


go/gregor/client/client.go
```diff
@@ -15,8 +16,18 @@ import (
 )
 
 type LocalStorageEngine interface {
-	Store(gregor.UID, []byte, [][]byte) error
-	Load(gregor.UID) ([]byte, [][]byte, error)
+	Store(gregor.UID, []byte, [][]byte, [][]byte) error
```
maxtaco: use named params here?

go/gregor/client/client.go
```diff
@@ -355,13 +416,103 @@ func (c *Client) filterLocalDismissals(ctx context.Context, state gregor.State)
...
+	if applyLocalState {
+		st = c.filterLocalDismissals(ctx, st)
+		st = c.applyOutboxMessages(ctx, st, t)
 	}
 	return st, nil
 }
+
+func (c *Client) outboxSend() {
+	var newOutbox []gregor.Message
+	msgs, err := c.Sm.Outbox(context.Background(), c.User)
```
maxtaco: use the same background context for the lot of them?

PR #12208, opened 2018-06-05
Escape terminal output

AMarcedone:
I am still working on the escaper itself, but the rest should be stable. 
@maxtaco want to take a look?


AMarcedone:
Ok, I think the escaper is better now

AMarcedone:
Still to do is whitelisting all uses of ColorString so that colors are not escaped.

maxtaco:
quick readthough is that this look really good.

AMarcedone:
Rebase is done.
Note that this PR does not attempt to escape anything which is printed to to stderr.

AMarcedone:
To note: the escaper works for UTF8 encoded strings. If the terminal uses a different encoding, issues might still be possible.
Maybe we can extend this fix to other encodings in another PR? This should be the most common case.

AMarcedone:
@maxtaco Highlighted outputs of chat search are now correctly escaped (color is shown).

go/escaper/escaper.go
```diff
@@ -22,10 +22,27 @@ func CleanBytes(p []byte) []byte {
...
+// Write writes p to the underlying io.Writer, after sanitizing it.
+// It returns n = len(p) on a successful write (regardless of how much data is written).
+// This is because the escaping function might alter the actual dimension of the data, but the caller is interested
+// in knowing how much of what they wanted to write was actually written. In case of errors it (conservatively) returns n=0
+// and the error, and no other writes are possible.
+func (w *Writer) Write(p []byte) (int, error) {
+	if w.err != nil {
+		return 0, w.err
+	}
+	if _, err := w.Writer.Write(CleanBytes(p)); err == nil {
```
maxtaco: remind me: is it guaranteed there won't be any "short writes"?

AMarcedone: Not sure about what you're asking here. 
If you are asking about why the function returns 0 on errors, my reasoning is that because of escaping the amount of bytes written to the stream and returned by w.Writer.Write(...) does not map directly to the number of bytes the caller intended to write. 
I could have written some more code to try and figure out how many bytes of the original stream were written, but this can be problematic if the escaping substitution we are doing changes the number of bytes of the symbols, and it is probably not useful anyways (I could not find any logic to retry writes, but I might be wrong on this). So I decided it was just easier to return 0 and document that. Happy to change that if you disagree.

Also, since it can be non trivial/error prone to recover from a failed write in this case (i.e. trying again), I decided to just prevent any further writing after an error. This behavior (locking further writes on error) is consistent with what bufio.Writer does.

maxtaco: My question isn’t that. You are throwing away the first return value from write. Traditionally that is because you can pass write n bytes and it might only write out m bytes, where m<n.  In that case you need to rewrite with the remainder. This is called a “short write.”  Is that a possibility here?

maxtaco: My question isn’t that. You are throwing away the first return value from write. Traditionally that is because you can pass write n bytes and it might only write out m bytes, where m<n.  In that case you need to rewrite with the remainder. This is called a “short write.”  Is that a possibility here?

go/client/terminal.go
```diff
@@ -7,19 +7,21 @@ import (
...
-	once   sync.Once // protects opening the minterm
-	engine *minterm.MinTerm
+	once      sync.Once // protects opening the minterm
+	engine    *minterm.MinTerm
+	rawWrites bool
 }
 
-func NewTerminal(g *libkb.GlobalContext) (*Terminal, error) {
-	return &Terminal{Contextified: libkb.NewContextified(g)}, nil
+func NewTerminal(g *libkb.GlobalContext, rawWrites bool) *Terminal {
```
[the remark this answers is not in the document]
AMarcedone: In general all the input systems (i.e. PromptXXXX in the TerminalUI interface) are not escaped, so nothing should fail there. I tried to create an account, and the password prompt was not in the terminal itself but some gnome window opened asking for a password.

go/escaper/escaper.go
AMarcedone: No, that is not a guarantee. In that situation Write would also have to return an error (that's mandated in the definition of the io.Writer interface), in which case this function returns the error and prevents further writes (rather than retrying). 
Are we doing retries on short writes somewhere else in the code that this can mess up with?

maxtaco: If you get an error back from write then it is fine to just error all the way out. Thanks.

go/client/files.go
```diff
@@ -197,6 +200,31 @@ func (s *StdoutSink) Write(b []byte) (n int, err error) {
...
+// This is because the escaping function might alter the actual dimension of the data, but the caller is interested
+// in knowing how much of what they wanted to write was actually written. In case of errors it (conservatively) returns n=0
+// and the error, and no other writes are possible.
+func (s *EscapedSink) Write(p []byte) (n int, err error) {
+	if s.err != nil {
+		return 0, s.err
+	}
+	if _, err := s.Sink.Write(escaper.CleanBytes(p)); err == nil {
+		return len(p), nil
+	} else {
```
maxtaco: remove `} else {`

go/client/files.go
```diff
@@ -268,13 +296,21 @@ func (s *FileSink) HitError(e error) error {
...
+	msg     string // input
+	infile  string
+	outfile string
 }
 
 func initSink(g *libkb.GlobalContext, fn string) Sink {
 	if len(fn) == 0 || fn == "-" {
-		return &StdoutSink{}
+		if g.Env.GetDisplayRawUntrustedOutput() || !isatty.IsTerminal(os.Stdout.Fd()) {
+			return &StdoutSink{}
```
maxtaco: Set `Contextified : libkb.NewContextified(g)` here, upon construction.

go/client/files.go
```diff
@@ -268,13 +296,21 @@ func (s *FileSink) HitError(e error) error {
...
+		} else {
+			return &EscapedSink{Sink: &StdoutSink{}}
```
maxtaco: As above, set `Contextified` on construction

go/client/files.go
```diff
@@ -293,19 +329,27 @@ func initSource(msg, infile string) (Source, error) {
 }
 
 func (u *UnixFilter) FilterInit(g *libkb.GlobalContext, msg, infile, outfile string) (err error) {
-	u.source, err = initSource(msg, infile)
-	if err == nil {
-		u.sink = initSink(g, outfile)
-	}
-	return err
+	u.Contextified = libkb.NewContextified(g)
```
maxtaco: remove this since we're setting it above.

go/client/terminal.go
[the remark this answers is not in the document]
maxtaco: See the first comment about style for `NewTerminal`

go/escaper/escaper.go
```diff
@@ -0,0 +1,48 @@
+package escaper
```
maxtaco: Can you rename this package to something like `terminal_escaper`? There are lot of things we could potentially be escaping.

go/escaper/escaper.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"io"
+	"strings"
+)
+
+func Clean(s string) string {
```
maxtaco: A comment for `Clean` would be nice.

go/escaper/escaper.go
```diff
@@ -0,0 +1,48 @@
...
+// This is because the escaping function might alter the actual dimension of the data, but the caller is interested
+// in knowing how much of what they wanted to write was actually written. In case of errors it (conservatively) returns n=0
+// and the error, and no other writes are possible.
+func (w *Writer) Write(p []byte) (int, error) {
+	if w.err != nil {
+		return 0, w.err
+	}
+	if _, err := w.Writer.Write(CleanBytes(p)); err == nil {
+		return len(p), nil
+	} else {
```
maxtaco: remove `} else {` here

go/escaper/escaper_test.go
```diff
@@ -0,0 +1,114 @@
...
+			t.Fail()
+		}
+		if n != len([]byte(a)) {
+			t.Logf("Write for %#v returned wrong length: %#v \n", a, n)
+			t.Fail()
+		}
+	}
+}
+
+func TestWriterLocksOnErr(t *testing.T) {
```
maxtaco: Do you mean `TestWriterErrorsOnError`?  `Locks` has a very particular meaning that we don't want to overload.

go/libcmdline/cmdline.go
```diff
@@ -447,6 +450,10 @@ func (p *CommandLine) PopulateApp(addHelp bool, extraFlags []cli.Flag) {
 			Name:  "debug, d",
 			Usage: "Enable debugging mode.",
 		},
+		cli.BoolFlag{
+			Name:  "display-raw-untrusted-output",
+			Usage: "Display outputs from users (messages, chats, ...) in the terminal without escaping terminal codes. WARNING: maliciously crafted unescaped outputs can overwrite anything you see on the terminal.",
```
maxtaco: I'd write: `"Display output from users..."`

`outputs` -> `output`

go/escaper/escaper_test.go
AMarcedone: What I want to imply is that after the first error, all subsequent writes will error (even if the underlying writer would allow them). Maybe `Freezes` is better than `Locks`? I can also use `TestWriterErrorsOnError` if you think it is clean enough.

maxtaco: `TestOutputStopsOnError`?

go/escaper/escaper.go
AMarcedone: I am going with `terminalescaper`, as underscore is not recommended in go package names.

AMarcedone: Maybe `terminal-escaper`?

maxtaco: sure, that's good with me

AMarcedone: I cannot use the `-` in package names, so I am going with `terminalescaper`

maxtaco: ok, that's fine. thanks

AMarcedone:
All the requested changes are done.
Thanks for the review!

maxtaco:
Great job on this first PR. It was a challenging one. Well done!

PR #12218, opened 2018-06-05
fix a few text cutoffs

songgao:
Fixed: 

screen shot 2018-06-05 at 3 13 17 pm
screen shot 2018-06-05 at 3 13 01 pm
screen shot 2018-06-05 at 3 12 43 pm
screen shot 2018-06-05 at 3 13 58 pm
screen shot 2018-06-05 at 3 14 56 pm

Not fixed:

screen shot 2018-06-05 at 3 14 26 pm

i.e. in the header on desktop, "long word" (no spaces in between) titles don't wrap. This is supposed to be controlled by `overflow-wrap: break-word;`, but doesn't work for the header. The same thing wraps in the action popup. It seems to be related to the fact that the title box in the header is a flex box and doesn't have fixed width. On mobile it seems it always breaks the word so it's never a problem.


shared/fs/popups/row-action-popup.js
```diff
@@ -32,16 +32,32 @@ const Popup = (props: Props) => {
...
-          {props.name}
-        </Text>
+        <Box style={stylesNameTextBox}>
+          <Text
+            type="BodySmallSemibold"
+            style={
+              // $FlowFixMe
+              {
+                color: props.itemStyles.textColor,
+                overflowWrap: 'break-word',
```
songgao: This is only supported by `text.desktop.js`. Will I have to break this file into .desktop and .native just for this, or is there a better way to do it?

shared/fs/header/header.desktop.js
```diff
@@ -106,27 +109,36 @@ const styleFolderHeaderEnd = {
...
   paddingRight: 0,
+  flexShrink: 0,
 }
 
+const lastFolderBreadcrumbStyle = {
+  ...folderBreadcrumbStyle,
+  flexShrink: 1,
+}
+
+const stylesLastNameText = {}
```
songgao: This is where I tried `overflowWrap: 'break-word'` but it didn't work. I also tried adding a few `width: '100%'`s and it didn't help either.

shared/fs/popups/row-action-popup.js
jzila: You could do:
```javascript
{
  color: ...,
  textAlign: ...,
  ...(isMobile ? {} : {overflowWrap: 'break-word'}),
}
```

akalin-keybase: I think using `platformStyles` is the preferred way to handle these

shared/fs/index.stories.js
```diff
@@ -113,6 +115,61 @@ const load = () => {
...
     ))
+    .add('Footer Cards', () => (
+      <Box>
+        <Box style={{height: 8}} />
+        <Download
+          filename="fjweio"
+          completePortion={0.42}
+          progressText="4 s"
+          isDone={false}
+          open={action('dismiss')}
```
akalin-keybase: `action('open')` etc, may want to have a `commonProps = { open: action(...) }` variable and have `{...commonProps}` here and below

shared/fs/header/header.native.js
```diff
@@ -13,28 +13,30 @@ type Props = {
 
 const Header = ({title, path, onBack}: Props) => (
   <Box style={stylesFolderHeaderContainer}>
+    <BackButton title={null} onClick={onBack} />
```
akalin-keybase: just omit `title` here

shared/fs/header/header.desktop.js
akalin-keybase: not sure how to fix this...maybe someone else with better CSS-fu can help? @buoyad perhaps?

shared/fs/popups/row-action-popup.js
songgao: Not sure how I didn't think of trying those. I'm kind of embarrassed for asking the question 😂

shared/fs/header/header.desktop.js
chrisnojima: this should belong in a storybook
i'm not 100% what your trying to do here but can't you just do wordBreak: 'break-word'

buoyad: I had no luck with `overflowWrap`, didn't try `wordBreak` though

songgao: I think overflowWrap and wordBreak are aliases to each other. Former is in the standard.

songgao: @chrisnojima if you haven’t, please take a look at PR description. It worked in the pop up but not header.

chrisnojima: can you put this into a storybook then i'll look

songgao: Alright; I added both header and row action popup to the storybook. Header is the one having problems. Thanks!!

songgao:
@chrisnojima Please take another look. I made a few changes on top of yours, mainly to make it consistent with Cécile's thoughts. Thanks!

screen shot 2018-06-12 at 11 45 57 am

songgao:
Thanks Nojima!

PR #12220, opened 2018-06-06
once upon a time there was an upload row

songgao:
Also add the missing download badge to Still row.

screen shot 2018-06-06 at 4 02 31 pm


shared/fs/row/uploading.js
```diff
@@ -0,0 +1,46 @@
...
+)
+
+const xOffset = -28
+const yOffset = 20
+const stylesIconBadge = {
+  marginLeft: xOffset,
+  marginRight: -xOffset,
+  marginTop: yOffset,
+  width: 0,
+  zIndex: 100,
```
chrisnojima: is this really necessary

songgao: Seems like so. Somehow it's affected by opacity of the thing under it:

screen shot 2018-06-08 at 11 45 48 am

Is there an alternative here?

chrisnojima: just put it later in the dom

songgao: I placed this as the last in the containing Box (and changed to row-reverse because it needs the fix-sized PathItemIcon to compute offset), and it didn't work.

songgao: (still affected by opacity of underlying component)

chrisnojima: lemme look

songgao: Thansk!

chrisnojima: ok so this is because a new stacking context is created since the icon has an opacity applied to it. so that forces you to do something like this which is kind of annoying. so we can keep it for now. if you happend to get a semi transparent image from cecile it actually would have worked

songgao: Hmm that's beyond my knowledge to understand. A semi transparent image meaning there's at least one pixel in the PNG that has an effective alpha value? It's still nice to not dependent on PNG having that though I guess.

chrisnojima: No I mean you have an opacity value applied through css which causes this behavior. If the transparency want handled this way it'd stack as expected

songgao: Ah I see. Makes sense.

PR #12225, opened 2018-06-06
cleanup the cache flush story

maxtaco:
- previously, we observed in CI, that it's possible to shut a cache down twice on logout
- this cleaner story should prevent that, and also cleans up the code a little bit.
- just watch out for certain tests that want reset caches while preserving cache properties

### DIFF
--- a/go/service/gregor.go
+++ b/go/service/gregor.go
@@ -376,6 +376,11 @@ func (g *gregorHandler) resetGregorClient(ctx context.Context) (err error) {
 }
 
 func (g *gregorHandler) getGregorCli() (*grclient.Client, error) {
+
+	if g == nil {
+		return nil, errors.New("gregorHandler client unset")
+	}
+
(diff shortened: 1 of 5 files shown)


maxtaco:
cc: @AMarcedone

maxtaco:
cc @mmaxim  on the new fixes to gregor on backgroundConvLoader

go/service/gregor.go
```diff
@@ -1783,11 +1797,15 @@ func newGregorRPCHandler(xp rpc.Transporter, g *libkb.GlobalContext, gh *gregorH
 func (g *gregorHandler) getState(ctx context.Context) (res gregor1.State, err error) {
 	var s gregor.State
 
-	if g == nil || g.gregorCli == nil {
+	g.gregorCliMu.Lock()
+	gcli := g.gregorCli
```
mmaxim: Maybe just change this to use the `getGregorCli` function?

PR #12233, opened 2018-06-06
new chat msg type REACTION

joshblum:


dabura667:
killing it, at always.

Reactions on the way *trumpets play*

PR #12241, opened 2018-06-06
add UpdateCategory RPC for dismissing an entire category and replacing with a new item CORE-8055

mmaxim:
Patch does the following:

1.) Adds `UpdateCategory` service RPC that dismisses the category and creates the new item.
2.) Get rid of small `EndTime_` on message range dismiss, since the server now handles them better. 

cc @buoyad

### DIFF
--- a/go/service/gregor.go
+++ b/go/service/gregor.go
@@ -1762,6 +1762,7 @@ func (g *gregorHandler) UpdateCategory(ctx context.Context, cat string, body []b
 	if err != nil {
 		return nil, err
 	}
+	msgID := msg.Ibm_.StateUpdate_.Md_.MsgID_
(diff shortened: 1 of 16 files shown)


mmaxim:
Requires https://github.com/keybase/keybase/pull/2556

go/service/gregor.go
```diff
@@ -1691,12 +1691,6 @@ func (g *gregorHandler) DismissCategory(ctx context.Context, category gregor1.Ca
 		Ranges_: []gregor1.MsgRange{
 			gregor1.MsgRange{
 				Category_: category,
-				// A small non-zero offset that effectively means "now",
```
mmaxim: @maxtaco of note this section, I think it is safe to remove this now with the server changes.

maxtaco: 👍

PR #12246, opened 2018-06-06
Use CopyText on Git repo URLs.

adamjspooner:


shared/git/row/index.js
```diff
@@ -136,27 +145,9 @@ class Row extends React.Component<Props, State> {
                 }}
```
buoyad: might need a `width` somewhere around here
image

adamjspooner: Good catch! Thanks!

cecileboucheron:
I think they could be wider on desktop - like could we have them be their maxWidth (should be set to 460px) and shrink if the window is smaller? It's nice to see the whole string if there is space for it. 

image

adamjspooner:
@keybase/react-hackers, this is ready for another review. There are two problems related to the `CopyText` component on mobile:
1. The `Toast` component is covering the entire view on iOS making it impossible to tap anything.
1. The text is wrapping on Android (even with `numberOfLines={1}`) making it clipped on top and bottom.

adamjspooner:
I can't review those changes, but they look good to me. Thanks, @buoyad!

buoyad:
[My changes](https://github.com/keybase/client/pull/12246/files/09bc145898756c0107781468aedf7b619ef399ef..e21ce4f5fae4232fdae22bd4beca220ab390b5ca), if someone would like to look

shared/common-adapters/toast.native.js
```diff
@@ -8,28 +8,40 @@ import type {Props} from './toast'
...
+        }).start()
+      )
     }
     if (!this.props.visible && prevProps.visible) {
       NativeAnimated.timing(this.state.opacity, {
         duration: 100,
         easing: NativeEasing.linear,
         toValue: 0,
       }).start()
+      this.timeoutID = setTimeout(() => this.setState({visible: false}), 100)
```
adamjspooner: Should the callback also `clearTimeout`?

shared/common-adapters/toast.native.js
```diff
@@ -8,31 +8,43 @@ import type {Props} from './toast'
...
-        duration: 50,
+        duration: 100,
         easing: NativeEasing.linear,
         toValue: 0,
       }).start()
+      this.timeoutID = setTimeout(() => this.setState({visible: false}), 100)
     }
   }
 
+  componentWillUnmount() {
```
chrisnojima: we should use timerhoc and not do this outselves

buoyad: done 👍

shared/common-adapters/toast.native.js
buoyad: I don't think it's necessary, by the time the callback happens the timer is already done

adamjspooner: Yeah, total nitpick.

adamjspooner:
@keybase/react-hackers, bump.

Issue #11296, opened 2018-04-08
Debian Buster: run_keybase does not succeed


miwie:
Already created an account and added two (mobile, Android) devices. 
Installed the Debian package _keybase_amd64.deb_ (1.0.47-20180403160041.242bcf96eb) on a Debian Buster system. No errors so far. 
After installing I tried to run _run_keybase_ as suggested on the webpage.   
All I get is this error message:

```
error: Usage: Need either 0,1 or 2 arguments for setting values in config
error: Need either 0,1 or 2 arguments for setting values in config
```

Any hints what's going wrong?

miwie:
Yes, `$XDG_RUNTIME_DIR` is set to `/run/user/1000`

oconnor663:
I'm still not entirely sure where this error is happening. The obvious spot would be `keybase config set mountdir "$mountdir"`, except that I really don't think that variable can be empty.

@miwie could you make a copy of `/usr/bin/run_keybase`, and add this line near the top:

    set -v -x

The please rerun this copy of `run_keybase`. It'll print out a ton of stuff before (presumably) running into this error again. Could you please send us all the output?

miwie:
Looks indeed like an error with `keybase config set mountdir "$mountdir"` but I don't know how to handle this.

[run_keybase-err.txt](https://github.com/keybase/client/files/1890548/run_keybase-err.txt)

oconnor663:
Thanks for the logs. This is still super confusing! It really looks like that command should be valid. I wonder if you have a `keybase` binary that's shadowing the installed one, or something else surprising like that. Could you please show us the output of these commands?

- `dpkg -s keybase`
- `dpkg -L keybase`
- `which keybase`
- `keybase version`

miwie:
Oops, the output of `which keybase` explains the error!
There has been an (very) old (unfortunately forgotten) installation in `/usr/local/bin` of `keybase`.
After deleting this everything seems to work.
Sorry for the trouble ...

oconnor663:
![image](https://user-images.githubusercontent.com/860932/38570440-ddae59d8-3cbb-11e8-90c7-39d41cf8ebe2.png)

xarses:
Also had problems trying to update from an very old version, really obscure error message. removing the old version fixed the problem.

tom-dalton-fanduel:
I've also just encountered this.

oconnor663:
The PR above adds a warning for this case in `run_keybase`.

[...]

PR #12250, opened 2018-06-07
add a warning for unexpected executable paths

oconnor663:
This should help with situations like
https://github.com/keybase/client/issues/11296.

r? @maxtaco

### DIFF
--- a/packaging/linux/run_keybase
+++ b/packaging/linux/run_keybase
@@ -225,7 +225,26 @@ start_background() {
   write_startup_token "background"
 }
 
+# Sometimes people get into weird configurations, where the `keybase` binary in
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Probably a more elegant approach would be to use something like https://invisible-island.net/conflict/conflict.html (packaged in the Arch repositories).

`conflict -p keybase` ==> one line per detected command on path, indicates something is weird.

`conflict` is quite fancy, but if you just need to check one command to see where it is anywhere on the $PATH, then bash has this built in as `type -ap keybase` The current run_keybase script is a horrible portmanteau of POSIX sh and GNU bash, declaring the bash shebang and using some bash features but then falling back on POSIX compatibility for some things... so I've got no idea if bashisms are something you want to pursue.

PR #12112, opened 2018-05-30
WIP: Less thread jumping

chrisnojima:
- [x] Lock to bottom/jump to bottom on send
- [x] Only throw away single cache row height instead of all
- [x] Update react-virtualized
- [x] Only use componentDidUpdate
- [x] Debounce scroll handler for lockToBottom. Without this the new items will cause a callback before it comes, during, and after which causes the state to thrash on/off/on etc

PR #12251, opened 2018-06-07
subset of thread fixes

chrisnojima:
this is a subset of the old thread jumping pr
https://github.com/keybase/client/pull/12112/files
should be better than master, still not great

@keybase/react-hackers @mmaxim


chrisnojima:
k it shouldn't affect that i hope, merging we can unmerge if its worse

PR #12252, opened 2018-06-07
Dump json files on load and save on android

patrickxb:
More attempts to try to track down strange android bugs...

### DIFF
--- a/go/libkb/json.go
+++ b/go/libkb/json.go
@@ -80,6 +80,17 @@ func (f *JSONFile) Load(warnOnNotFound bool) error {
 		return err
 	}
 	f.jw = jsonw.NewWrapper(obj)
+
+	if runtime.GOOS == "android" {
+		// marshal it into json without indents to make it one line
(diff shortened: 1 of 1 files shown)


go/libkb/json.go
```diff
@@ -80,6 +80,17 @@ func (f *JSONFile) Load(warnOnNotFound bool) error {
 		return err
 	}
 	f.jw = jsonw.NewWrapper(obj)
+
+	if runtime.GOOS == "android" {
```
maxtaco: should we just do this on all platforms?

patrickxb: I could go either way...I'm not aware of any issues on other platforms.

Only downside is cluttering up the logs, and maybe some privacy concerns about logging the information.  The device and username stuff is public, but maybe the fact that user X is also user Y?  or that user X uses proxy Z?

maxtaco: good point.... ok android only...

PR #12255, opened 2018-06-07
hold onto tsec'ed passphrase as input for PGP synced key decryption

maxtaco:
- there are two ways we could have handled this
  - the first is when decrypting a synced PGP key, to ask for a salt and generate a full passphrase stream,
    since we can read the salt out of the P3SKB blob and therefore don't need to ask the server for it.
  - the second is just to handle the case of a partial passphrase stream cache (with a tsec but with a passphraseStream)
- we opted for the second option here, since it's slightly less disruptive, but we might revisit this idea later on

### DIFF
--- a/go/engine/passphrase_change.go
+++ b/go/engine/passphrase_change.go
@@ -299,6 +299,11 @@ func (c *PassphraseChange) runStandardUpdate(m libkb.MetaContext) (err error) {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


maxtaco:
actually @patrickxb hold off, something is up...

maxtaco:
ok @patrickxb  all good

PR #12264, opened 2018-06-07
Add wallet rpc preamble with RequireWallet

mlsteele:
Make the error `logged-in user has no wallet accounts` harder to get (if stellar is enabled for you). With this PR, when an RPC that has RequireWallet is called, it checks if the user has a wallet and if not tries to make one and if that fails errors out. Quickly in the common success case (<1ms locally).

Previously running `wallet list` wouldn't even try to get you a wallet. So you could race with the background maker for example by waiting more than 15 seconds to login on a logged-out device. Or by having stellar enabled for you while your service was running. I'm not sure how often to expect that sort of scenario, but with this PR it no longer matters.

### DIFF
--- a/go/stellar/stellar.go
+++ b/go/stellar/stellar.go
@@ -23,13 +23,16 @@ import (
 
 // CreateWallet creates and posts an initial stellar bundle for a user.
 // Only succeeds if they do not already have one.
-// Safe to call even if the user has a bundle already.
+// Safe (but wasteful) to call even if the user has a bundle already.
 func CreateWallet(ctx context.Context, g *libkb.GlobalContext) (created bool, err error) {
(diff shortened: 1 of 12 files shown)


go/stellar/stellar.go
```diff
@@ -77,7 +104,7 @@ func CreateWalletSoft(ctx context.Context, g *libkb.GlobalContext) {
 		err = fmt.Errorf("yielding to guard")
 		return
 	}
-	_, err = CreateWalletGated(ctx, g)
+	_, _, err = CreateWalletGated(ctx, g)
```
patrickxb: This `err` is ignored...check it?  log it?

mlsteele: It's logged by the trace

patrickxb: ok

PR #12268, opened 2018-06-07
fix protocol registration race and cleanup init code

songgao:
Sister PR of https://github.com/keybase/kbfs/pull/1604 .


songgao:
Yup I did test both Files and Git.

PR #12280, opened 2018-06-08
remeasure if wrapper changes. move orange line data out of meta

chrisnojima:
@keybase/react-hackers this should fix some measuring / thrash around orange line handling

@buoyad


shared/chat/conversation/messages/wrapper/container.js
```diff
@@ -15,8 +14,8 @@ const mapStateToProps = (state: TypedState, {message, previous, innerClass, isEd
   const isYou = state.config.username === message.author
   const isFollowing = state.config.following.has(message.author)
   const isBroken = state.users.infoMap.getIn([message.author, 'broken'], false)
-  const meta = Constants.getMeta(state, message.conversationIDKey)
-  const orangeLineAbove = !!previous && meta.orangeLineOrdinal === previous.ordinal
+  const orangeLineOrdinal = state.chat2.orangeLineMap.get(message.conversationIDKey)
+  const orangeLineAbove = !!previous && orangeLineOrdinal === previous.ordinal
```
buoyad: The orange line isn't showing when it's the first time I've loaded the thread

shared/chat/conversation/messages/wrapper/index.js.flow
```diff
@@ -18,6 +18,7 @@ export type Props = {
   isFollowing: boolean,
   isRevoked: boolean,
   isYou: boolean,
+  measure: null | (() => void),
```
buoyad: `?(() => void)`

chrisnojima:
This keeps the current behavior. We don't mutate the orange line based on
incoming messages etc. the null or function thing came from flow being
wacky about undefined

PR #12281, opened 2018-06-08
Add flag to send from a non-primary account

patrickxb:


protocol/avdl/stellar1/local.avdl
```diff
@@ -137,7 +137,7 @@ protocol local {
 
   SendResultCLILocal sendCLILocal(string recipient, string amount, Asset asset, string note,
                                   string displayAmount, string displayCurrency, boolean forceRelay,
-                                  string publicNote);
+                                  string publicNote, AccountID fromAccountID);
```
mlsteele: This argument order is getting weird. Put from first and comment that it's optional

go/client/cmd_wallet_send.go
```diff
@@ -67,6 +72,7 @@ func (c *CmdWalletSend) ParseArgv(ctx *cli.Context) error {
 	}
 	c.Note = ctx.String("message")
 	c.ForceRelay = ctx.Bool("relay")
+	c.FromAccountID = stellar1.AccountID(ctx.String("from-account-id"))
```
mlsteele: Error out before asking for confirmation

```
fromAccountID := ctx.String("from-account-id")
if fromAccountID != "" {
  c.FromAccountID, err = libkb.ParseStellarAccountID(fromAccountID)
  if err != nil {
    return err
  }
}
```

go/client/cmd_wallet_send.go
```diff
@@ -29,6 +30,10 @@ func newCmdWalletSend(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Co
 			Name:  "m, message",
 			Usage: "Include a message with the payment.",
 		},
+		cli.StringFlag{
+			Name:  "f, from-account-id",
```
mlsteele: `-f` is often force, which we might want since this currently asks for confirmation.

How about just `from`?

go/stellar/stellar.go
```diff
@@ -187,6 +187,21 @@ func OwnAccount(ctx context.Context, g *libkb.GlobalContext, accountID stellar1.
...
 
+func LookupSender(ctx context.Context, g *libkb.GlobalContext, accountID stellar1.AccountID) (stellar1.BundleEntry, error) {
+	bundle, _, err := remote.Fetch(ctx, g)
+	if err != nil {
+		return stellar1.BundleEntry{}, err
+	}
+
+	for _, account := range bundle.Accounts {
+		if account.AccountID.Eq(accountID) {
+			return account, nil
```
mlsteele: Add the Signers length checks from LookupSenderPrimary

go/stellar/stellar.go
```diff
@@ -294,14 +309,21 @@ type DisplayBalance struct {
 // User with wallet ready : Standard payment
 // User without a wallet  : Relay payment
 // Unresolved assertion   : Relay payment
-func SendPayment(m libkb.MetaContext, remoter remote.Remoter, to stellarcommon.RecipientInput, amount string, note string, displayBalance DisplayBalance, forceRelay, quickReturn bool, publicNote string) (res stellar1.SendResultCLILocal, err error) {
+func SendPayment(m libkb.MetaContext, remoter remote.Remoter, to stellarcommon.RecipientInput, amount string, note string, displayBalance DisplayBalance, forceRelay, quickReturn bool, publicNote string, fromAccountID stellar1.AccountID) (res stellar1.SendResultCLILocal, err error) {
```
mlsteele: Move `fromAccountID` to before `to`. Could just be `from` since it's typed. Also comment that it's optional.

protocol/avdl/stellar1/local.avdl
patrickxb: the order doesn't matter...it turns into a struct.

PR #12283, opened 2018-06-08
Add delete history to exploding messages

buoyad:
r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/message-popup/exploding/container.js
+++ b/shared/chat/conversation/messages/message-popup/exploding/container.js
@@ -4,6 +4,7 @@ import * as Constants from '../../../../../constants/chat2'
 import * as TeamConstants from '../../../../../constants/teams'
 import * as Types from '../../../../../constants/types/chat2'
 import * as Chat2Gen from '../../../../../actions/chat2-gen'
+import {navigateAppend} from '../../../../../actions/route-tree'
 import {connect, type TypedState} from '../../../../../util/container'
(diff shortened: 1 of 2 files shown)


shared/chat/conversation/messages/message-popup/exploding/container.js
```diff
@@ -19,12 +20,12 @@ type OwnProps = {
 const mapStateToProps = (state: TypedState, ownProps: OwnProps) => {
   const yourMessage = ownProps.message.author === state.config.username
   const meta = Constants.getMeta(state, ownProps.message.conversationIDKey)
-  const canExplodeNow =
-    meta.teamType === 'adhoc' ||
-    yourMessage ||
-    TeamConstants.getCanPerform(state, meta.teamname).manageMembers
+  const canDeleteHistory =
+    meta.teamType === 'adhoc' || TeamConstants.getCanPerform(state, meta.teamname).deleteChatHistory
```
chrisnojima: Can perform can't be used in adhoc? We could make it right

buoyad: Everyone can do everything in adhoc conversations right?

chrisnojima:
Yes but can user perform could just return those flags instead of everyone
always checking

buoyad:
That might be useful. Existing code about this is pretty team specific and in the teams store etc so it might be a big change. Merging this for now

PR #12291, opened 2018-06-10
fix syncing gregor state

mmaxim:
Fix syncing Gregor state, this could cause us to think we have a blank Gregor state locally in the following circumstance:

1.) Initiate restart of Gregor.
2.) Desktop clients all re-connect and re-sync state.
3.) There is another bug which causes us to miss the hash check for some reason (we can deal with this separately), causing the client to sync down the entire Gregor state fresh. Since Gregor is restarting, it is very slow to respond to this request.
4.) We had some bogus timeout set on this call of 5 seconds, and so it could easily miss this timeout in the restart scenario.
5.) `freshSync` clears the local state before hearing back from the server, so if the server call returns an error then we are left with a blank local state. 
6.) `OnConnect` did not fail out if the Gregor state sync routine returns an error, and so we are left fully connected to Gregor with a blank local state. This would only be fixable on a Gregor restart or local client restart.

Patch fixes this by doing the following:

1.) Failing out `OnConnect` if Gregor state sync fails.
2.) Increasing timeout on the call after a local state hash failure to the default of 30 seconds.
3.) Doesn't really affect anything after the first two, but we also don't clear the local state until hearing back from the server with no error. 

cc @buoyad @chrisnojima I think this should address the problems with getting a blank Gregor state back from `GetState`.

### DIFF
--- a/go/gregor/client/client.go
+++ b/go/gregor/client/client.go
@@ -232,7 +232,7 @@ func (c *Client) freshSync(ctx context.Context, cli gregor1.IncomingInterface, s
 
 	if state == nil {
 		state = new(gregor.State)
-		*state, err = c.State(cli)
+		*state, err = c.State(ctx, cli)
 		if err != nil {
 			return msgs, err
 		}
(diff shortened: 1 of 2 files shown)


chrisnojima:
Cool. Maybe if we can't connect we should tell the GUI we're offline

PR #12299, opened 2018-06-11
fix mobile

chrisnojima:
1. Red box on mobile due to lineheight
1. Send animation totally offscreen

@keybase/react-hackers 
cc: @cecileboucheron this was broken in your styling changes.
you can't pass a lineheight string to react native. use platformStyles to do this automatically


cecileboucheron:
Oh ok, oops.

PR #12303, opened 2018-06-11
Some belated logging that went in after the hotfix shipped

zanderz:


go/libkb/pipeowner_windows.go
```diff
@@ -93,16 +95,16 @@ func IsPipeowner(log logger.Logger, name string) (bool, error) {
...
 		log.Debug("Pipe account: %s, %s, %v, %v", pipeAccount, pipeDomain, pipeAccType, pipeErr)
 		log.Debug("User account: %s, %s, %v, %v", userAccount, userDomain, userAccType, userErr)
 		// If the pipe is served by an admin, let local security policies control access
 		if pipeAccount == "Administrators" && pipeDomain == "BUILTIN" && pipeAccType == syscall.SidTypeAlias {
-			return true, nil
+			isOwner = true
 		}
 	}
-	return isEqual, nil
+	return
```
maxtaco: But still do `return isOwner, nil`, since that's less error-prone as the code changes in the future.

PR #12304, opened 2018-06-11
stepping stone to per platform flow

chrisnojima:
@keybase/react-hackers this is a step towards per platform flow typing.

Here's how it'll work and how we can work on this in the short term while keeping the 'old' way working

Old:
We have:
```
foo.desktop.js
foo.native.js
foo.js.flow
```

flow sees `foo.desktop.js` and `foo.native.js`. When things import `foo` it reads `foo.js.flow` as the interface. This means if `foo.js.flow` and `foo.desktop.js` or `foo.native.js` differ flow won't catch this.

New:
We have:
```
foo.desktop.js
foo.native.js
```

we'll have multiple flow configs that'll turn on / off the various file types. So on a flow-desktop run flow will only see `foo.desktop.js` and on a flow-ios run flow will only see `foo.native.js`. Files that import foo will only see the real .desktop or .native file. This would make flow more closely align with how the bundlers see the world.

This means we could have Props that are different for desktop and native and that would resolve correctly and be ok. We obviously have a lot of components which are split but share props. Those can just import the shared props from a `foo.types.js.flow` file.

As a short term bridge (so we can keep the old style until everything is new style compatible) we can basically pull what was in `foo.js.flow` and put it into a `foo.types.js.flow`. Then `foo.native.js` `foo.desktop.js` and `foo.js.flow` will all pull that in so old/new style will still all work and point to the same props style. When we get the multi flow thing going we can then delete `foo.js.flow`

I wanted to put a small example of how this would work on a pr before going further.

thoughts?


chrisnojima:
also we could ditch the `foo.types.js.flow` and just copy the props into the desktop and native files. Might be cleaner that way. Also longer term we should collapse a bunch of these split files as they don't need to exist in a lot of cases

chrisnojima:
@keybase/react-hackers bump

chrisnojima:
@keybase/react-hackers bump

songgao:
Sounds good to me!

songgao:
Sounds good to me!

PR #12306, opened 2018-06-11
Support avatar uploads.

adamjspooner:


adamjspooner:

Two things need improvement:

1. Zooming on desktop isn't great. It zooms from the top left instead of from the center of the area you're viewing.
1. Zooming on Android is temporarily deactivated. I can't figure out how to get the cropping coordinates.

This only adds the ability to edit your user avatar. The ability to edit team avatars will come in a later PR.

shared/profile/edit-avatar/index.native.js
```diff
@@ -1,33 +1,127 @@
...
+
+class EditAvatar extends React.Component<Props> {
+  _h: number = 0
+  _w: number = 0
+  _x: number = 0
+  _y: number = 0
+
+  _onSave = () => {
+    const filename = isIOS ? this.props.image.uri.replace('file://', '') : this.props.image.path
+    if (this._h && this._w && this._x && this._y) {
```
chrisnojima: isn't x/y =0 valid?

shared/profile/edit-avatar/index.native.js
```diff
@@ -1,33 +1,127 @@
...
+      this.props.onSave(filename, this._getCropCoordinates())
+      return
+    }
+    this.props.onSave(filename)
+  }
+
+  _getCropCoordinates = () => {
+    const x = this._x
+    const y = this._y
+    const rH = this.props.image.height / this._h
```
chrisnojima: need to sanity check this never gets called with zero _h or _w

shared/profile/edit-avatar/index.js.flow
```diff
@@ -1,10 +1,11 @@
 // @flow
 import {Component} from 'react'
+import * as RPCTypes from '../../constants/types/rpc-gen'
 
 export type Props = {
-  keybaseUsername: string,
-  hasAvatar: boolean,
-  onAck: () => void,
+  image: Object,
```
chrisnojima: why not type this

shared/profile/edit-avatar/index.native.js
```diff
@@ -1,33 +1,127 @@
...
+              waitingKey={null}
+            />
+          </ButtonBar>
+        </Box>
+      </StandardScreen>
+    )
+  }
 }
 
+const container = platformStyles({
```
chrisnojima: container should get in styleSheetCreate

shared/profile/edit-avatar/index.native.js
```diff
@@ -1,33 +1,127 @@
...
+const styles = styleSheetCreate({
+  button: {
+    marginTop: globalMargins.tiny,
+    width: '100%',
+  },
+  zoomContainer: {
+    alignSelf: 'center',
+    backgroundColor: globalColors.lightGrey2,
+    borderRadius: AVATAR_SIZE,
+    flexShrink: 1,
```
chrisnojima: why shrink 1?

shared/profile/index.native.js
```diff
@@ -104,18 +105,31 @@ class Profile extends Component<Props, State> {
...
+      this.props.onEditAvatar(response)
+    })
+  }
+
   _makeUserBio(loading: boolean) {
     return (
       <UserBio
         type="Profile"
-        editFns={this.props.bioEditFns}
+        // editFns={this.props.bioEditFns}
```
chrisnojima: ???

shared/profile/profile.stories.js
```diff
@@ -164,6 +164,7 @@ const props = {
   currentFriendshipsTab: 'Followers',
   currentlyFollowing: false,
   error: null,
+  // filename: null,
```
chrisnojima: ???

shared/common-adapters/zoomable-box.android.js
```diff
@@ -212,7 +212,6 @@ class ZoomableBox extends React.Component<Props, State> {
         onLayout={this._onLayout}
         style={{
           ...this.props.style,
-          position: 'absolute',
```
chrisnojima: does this break attachment full screen view?

shared/profile/edit-avatar/index.js.flow
adamjspooner: It's a prop that's only used on native, and I wasn't sure how to do that since this is shared with desktop. How do I do that? Is there a better way to do this?

adamjspooner:
Sorry! I apparently left a lot of testing code around. I'll clean this up before asking for another review.

shared/profile/edit-avatar/index.native.js
adamjspooner: It's because `StandardScreen` is expecting an `Object` and not `StylesCrossPlatform`. Should I fix that as part of this PR?

I get this when I try to use it in `styleSheetCreate`:

```
Cannot create `StandardScreen` element because read-only array type [1] is incompatible with object type [2] in property `style`.
```

chrisnojima: short term i'd just change standard screen to take any instead of object. any loses typing but object is just incorrect

shared/profile/edit-avatar/index.js.flow
chrisnojima: you can do ?{x: number, y: number, /etc) whatever it is, and its ok for it to be null

adamjspooner: When I do that, Flow complains:

```
Cannot get `this.props.image.uri` because property `uri` is missing in null or undefined [1].
```

Do I need to check for `this.props.image`'s existence every time I want to access its properties?

chrisnojima: yes cause thats what you're saying, it can be null sometimes

shared/profile/edit-avatar/index.native.js
adamjspooner: I tried `any` and `?Object | ?StylesCrossPlatform`, and both end up with this error:

screen shot 2018-06-29 at 3 15 01 pm

...which I don't understand. It's where the wrapped component's props are being expanded: https://github.com/keybase/client/blob/master/shared/common-adapters/header-hoc.native.js#L42

chrisnojima: thats not a flow error, thats an actual runtime error
in RN you can't do 
```
{
  foo: 'bar',
  ...(test ? {baz: whut} : null)
}
```
it doesn't allow you to spread null
also if your spreading the output of stylesheetcreate that won't work as its numbers and not an object. thats what collapseStyles is for

adamjspooner: Thanks for the explanation!

adamjspooner:
@keybase/react-hackers, this has been cleaned up and is ready for another review.

chrisnojima:
i'm looking

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,382 @@
...
+    }
+    this._filePickerSetValue('')
+  }
+
+  _onDragLeave = () => {
+    this.setState({dropping: false})
+  }
+
+  _onDrop = (e: SyntheticDragEvent<any>) => {
+    e.persist()
```
chrisnojima: shouldn't be necessary

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,382 @@
...
+        } catch (e) {
+          // TODO: Show a red error banner on failure: https://zpl.io/2jlkMLm
+        }
+      }
+      this._paintImage(paths.pop())
+    }
+  }
+
+  _validDrag = (e: SyntheticDragEvent<any>) => {
+    e.persist()
```
chrisnojima: shouldn't be necessary

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,382 @@
...
+    return Array.prototype.map.call(e.dataTransfer.types, t => t).includes('Files')
+  }
+
+  _onDragOver = (e: SyntheticDragEvent<any>) => {
+    e.persist()
```
chrisnojima: shouldn't be necessary

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,382 @@
...
+      e.dataTransfer.dropEffect = 'none'
+    }
+  }
+
+  _paintImage = (path: string) => {
+    this.setState({imageSource: path})
+  }
+
+  _onImageLoad = (e: SyntheticEvent<HTMLImageElement>) => {
+    e.persist()
```
chrisnojima: shouldn't be necessary

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,382 @@
...
+
+    this.setState({
+      scale,
+      scaledImageHeight,
+      scaledImageWidth,
+    })
+  }
+
+  _onMouseDown = (e: SyntheticMouseEvent<any>) => {
+    e.persist()
```
chrisnojima: shouldn't be necessary

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,382 @@
...
+      dragStopY:
+        this._image && this._image.style.top ? parseInt(this._image.style.top, 10) : this.state.dragStopY,
+      dragging: false,
+      offsetLeft: this._image ? this._image.offsetLeft : this.state.offsetLeft,
+      offsetTop: this._image ? this._image.offsetTop : this.state.offsetTop,
+    })
+  }
+
+  _onMouseMove = (e: SyntheticMouseEvent<any>) => {
+    e.persist()
```
chrisnojima: shouldn't be necessary

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,382 @@
...
+    const crop = {
+      x0: Math.round(x * rW),
+      x1: Math.round((x + AVATAR_SIZE - AVATAR_BORDER_WIDTH * 2) * rW),
+      y0: Math.round(y * rH),
+      y1: Math.round((y + AVATAR_SIZE - AVATAR_BORDER_WIDTH * 2) * rH),
+    }
+    this.props.onSave(this.state.imageSource, crop)
+  }
+
+  render = () => {
```
chrisnojima: use 
```
render () {
```

shared/profile/edit-avatar/index.native.js
```diff
@@ -1,33 +1,121 @@
...
+  _w: number = 0
+  _x: number = 0
+  _y: number = 0
+
+  _onSave = () => {
+    if (!this.props.image) {
+      throw new Error('Missing image when saving avatar')
+    }
+    const filename = isIOS ? this.props.image.uri.replace('file://', '') : this.props.image.path
+    // Cropping is temporarily deactivated on Andoird.
```
chrisnojima: Android

shared/profile/edit-avatar/index.native.js
```diff
@@ -1,33 +1,121 @@
...
+              // Temporarily deactive zooming on Android.
+              maxZoom={isIOS ? 10 : 1}
+              onZoom={isIOS ? this._onZoom : null}
+              showsHorizontalScrollIndicator={false}
+              showsVerticalScrollIndicator={false}
+              style={styles.zoomContainer}
+            >
+              <NativeImage
+                resizeMode="contain"
+                source={{uri: `data:image/jpeg;base64,${this.props.image ? this.props.image.data : ''}`}}
```
chrisnojima: is this necessary? can't you just pass the path in?

shared/profile/index.native.js
```diff
@@ -104,7 +105,23 @@ class Profile extends Component<Props, State> {
...
+        console.error(response.error)
+        throw new Error(response.error)
+      }
+      this.props.onEditAvatar(response)
+    })
+  }
+
   _makeUserBio(loading: boolean) {
+    if (this.props.isYou && this.props.bioEditFns) {
+      this.props.bioEditFns.onEditAvatarClick = this._onClickAvatar
```
chrisnojima: we should never write to props

chrisnojima:
first test on ios simulator w/ built in images: 
screen shot 2018-07-03 at 11 37 28 am

adamjspooner:
@chrisnojima: https://github.com/keybase/keybase/pull/2642

chrisnojima:
The aspect ratio isn't maintained, so if i upload a small image it squeezes it. also putting in small images doesn't let you pan around at all

chrisnojima:
app_store_badge_320x94
dog

chrisnojima:
( i was looking for files in safari since i can't use the camera roll due to the files being too big)

shared/profile/index.native.js
adamjspooner: I'm not sure what to do here. I need to use `react-native-image-picker`'s `showImagePicker`. So, I can't override it in the container (which seems like the right thing to do). The desktop fails to render with:

```
Failed to compile.

./node_modules/react-native-image-picker/index.js
Module parse failed: Unexpected token (22:2)
You may need an appropriate loader to handle this file type.
| 
| module.exports = {
|   ...ImagePickerManager,
|   showImagePicker: function showImagePicker(options, callback) {
|     if (typeof options === 'function') {
```

chrisnojima:
Just write to your own var and pass that down. Don't write into the parents
value

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,388 @@
...
+  }
+
+  render() {
+    return (
+      <MaybePopup
+        onClose={this.props.onClose}
+        styleCover={collapseStyles([
+          styles.cover,
+          {
+            cursor: !this.state.hasPreview ? 'default' : this.state.dragging ? 'grabbing' : 'grab',
```
adamjspooner: @chrisnojima, this doesn't work, but I don't understand why. Can you explain why it doesn't?

shared/common-adapters/index.js
```diff
@@ -11,14 +11,22 @@ const MaybePopup = isMobile
...
       styleContainer?: any,
     }) => (
       <PopupDialog
         onClose={props.onClose}
-        styleCover={props.cover ? {..._styleCover, ...props.styleCover} : {}}
+        onMouseUp={props.onMouseUp}
+        onMouseDown={props.onMouseDown}
+        onMouseMove={props.onMouseMove}
+        styleCover={
+          props.cover ? {..._styleCover, ...props.styleCover} : props.styleCover ? props.styleCover : {}
```
chrisnojima: this should just use collapse styles, its smart
```
collapseStyles([props.cover && _styleCover, props.styleCover]}
```

adamjspooner: Oh, nice! Will do.

adamjspooner:
@chrisnojima, this is ready for another look. It has user and team avatar uploads. Zooming and panning on Android is all that's lacking.

adamjspooner:
@chrisnojima, this now has Android support.

chrisnojima:
looking

chrisnojima:
looking. you need to merge master in to fix circle

shared/common-adapters/avatar.js
```diff
@@ -163,7 +168,8 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps): PropsWithout
 
   const style = collapseStyles([
     ownProps.style,
-    onClick && platformStyles({isElectron: desktopStyles.clickable}),
+    (onClick || (flags.avatarUploadsEnabled && ownProps.onEditAvatarClick)) &&
```
chrisnojima: this logic can be in mapDispatchToProps, just make onEditAvatarClick undefined there

shared/common-adapters/avatar.js
```diff
@@ -225,19 +233,24 @@ class AvatarConnector extends React.PureComponent<Props> {
...
         children={this.props.children}
+        editable={flags.avatarUploadsEnabled && this.props.editable}
         followIconSize={this.props.followIconSize}
         followIconStyle={this.props.followIconStyle}
         followIconType={this.props.followIconType}
         isTeam={this.props.isTeam}
         loadingColor={this.props.loadingColor}
         onClick={this.props.onClick}
+        onEditAvatarClick={
+          flags.avatarUploadsEnabled && this.props.onEditAvatarClick
```
chrisnojima: then you can just use this.props.onEditAvatarclick here

shared/common-adapters/avatar.render.desktop.js
```diff
@@ -106,7 +106,7 @@ class AvatarRender extends React.PureComponent<Props, State> {
 
     return (
       <div
-        onClick={this.props.onClick}
+        onClick={this.props.onEditAvatarClick ? this.props.onEditAvatarClick : this.props.onClick}
```
chrisnojima: you can just contorl what onClick means in the connector. this can just be this.props.onclick

shared/profile/edit-avatar-placeholder/container.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import EditAvatar from '.'
+import {connect, type TypedState} from '../../util/container'
+import {navigateUp} from '../../actions/route-tree'
+
+const mapStateToProps = (state: TypedState) => {
+  const username = state.config.username
+  if (!username) {
```
chrisnojima: maybe better to not throw. i know we do this in places but i think its more dangerous, if no username i'd just return mostly blank stuff and have it not render

chrisnojima: well, actually this is old so we can keep it, just for future reference

adamjspooner: Should we get an issue in JIRA to clean this up?

adamjspooner:

adamjspooner:

adamjspooner:
@keybase/react-hackers, bump.

shared/common-adapters/avatar.render.desktop.js
```diff
@@ -148,6 +148,17 @@ class AvatarRender extends React.PureComponent<Props, State> {
...
+        {this.props.editable && (
+          <Icon
+            type="iconfont-edit"
+            onClick={this.props.onEditAvatarClick}
+            style={{
+              bottom: this.props.isTeam ? -2 : 0,
+              position: 'absolute',
+              right: this.props.isTeam ? -18 : 0,
+            }}
+          />
```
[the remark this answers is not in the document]
adamjspooner: I removed that `onClick` because it's on a layer behind the avatar, and the avatar's `onClick` will take care of it. So, it's as if they're wrapped in a `ClickableBox` but not really.

shared/common-adapters/image.js.flow
```diff
@@ -3,7 +3,8 @@
 export type Props = {
   src: string,
   style?: any,
-  onLoad?: () => void,
+  onDragStart?: (e: SyntheticEvent<>) => void,
+  onLoad?: (e: SyntheticEvent<>) => void,
```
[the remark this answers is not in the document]
adamjspooner: 🙌🏻

shared/common-adapters/name-with-icon.js
```diff
@@ -46,6 +48,8 @@ const NameWithIconVertical = (props: Props) => {
     >
       {isAvatar && (
         <Avatar
+          editable={props.editableIcon}
+          onEditAvatarClick={props.editableIcon ? props.onEditIcon : undefined}
```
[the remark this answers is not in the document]
chrisnojima: its better to use undefined, then the component can do something else. if you pass a function it can't know if its doing something so it won't fallback. aka sometimes it'll wrap itself differently or have different behavior on RN

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -244,14 +244,15 @@ class OrientedImage extends React.Component<Props, State> {
...
-        <canvas ref={el => (this._canvasRef = el)} style={styleCanvas} />
         {this.state.srcTransformed && (
           <ImageComponent
             src={this.state.srcTransformed}
             style={this.props.style}
+            onDragStart={this.props.onDragStart}
             onLoad={this.props.onLoad}
           />
         )}
+        <canvas ref={el => (this._canvasRef = el)} style={styleCanvas} />
```
chrisnojima: what is this change for?

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,494 @@
...
+              <Box style={styles.spinnerContainer}>
+                <ProgressIndicator style={iconCastPlatformStyles(styles.spinner)} />
+              </Box>
+            )}
+            <OrientedImage
+              ref={this._imageSetRef}
+              src={this.state.imageSource}
+              style={{
+                height: this.state.scaledImageHeight,
+                left: `${this.state.offsetLeft}px`,
```
chrisnojima: shouldn't need the px

shared/profile/edit-avatar/index.desktop.js
```diff
@@ -1,33 +1,494 @@
...
+                opacity: this.state.loading ? '0' : '1',
+                position: 'absolute',
+                top: `${this.state.offsetTop}px`,
```
chrisnojima: same here

shared/profile/index.js.flow
```diff
@@ -1,11 +1,11 @@
 // @flow
 import {Component} from 'react'
 import * as Types from '../constants/types/profile'
+import type {Response} from 'react-native-image-picker'
```
chrisnojima: we shouldn't have RN only things in a shared file like this

shared/teams/team/header/container.js
```diff
@@ -36,6 +37,10 @@ const mapDispatchToProps = (dispatch: Dispatch, {teamname}: OwnProps) => ({
   },
   onChat: () => dispatch(Chat2Gen.createPreviewConversation({teamname, reason: 'teamHeader'})),
   onEditDescription: () => dispatch(navigateAppend([{props: {teamname}, selected: 'editTeamDescription'}])),
+  onEditIcon: (image?: Response) =>
+    dispatch(
+      navigateAppend([{props: {image, sendChatNotification: true, teamname}, selected: 'editTeamAvatar'}])
```
chrisnojima: do we really have to use routeprops for this

shared/common-adapters/oriented-image.desktop.js
adamjspooner: I need a reference to the image (not the canvas).

```
_imageSetRef = (ref: ?OrientedImage) => {
  this._image = ref
}
```

That was giving me a canvas before this change.

shared/common-adapters/avatar.render.desktop.js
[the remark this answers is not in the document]
thebearjew: Ah awesome!

shared/common-adapters/name-with-icon.js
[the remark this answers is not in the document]
thebearjew: Did not know this, thank you!

shared/profile/index.js.flow
adamjspooner: What's the right thing to do? Use `any` instead?

shared/teams/team/header/container.js
adamjspooner: How else can I do it?

shared/profile/index.js.flow
chrisnojima: thats fine for now, or you could make the actual type yourself (whatever part you use), aka
type Response ={
data: string
} or whatever

shared/teams/team/header/container.js
chrisnojima: it could go into the store. i think this is fine for now as this PR is too big and too old but as a future reference i think we should avoid routeProps unless absolutely necessary (aka when a single screen is used multiple places and needs context passed in). instead probably simpler to just use the store

adamjspooner: Sounds good. Thanks for the info!

adamjspooner:

chrisnojima:
should turn off the feature flag

adamjspooner:
Do you mean enable it for everyone?

chrisnojima:
might as well. this is ready to go live right

adamjspooner:
What if we dog food it while I'm out and I enable it for everyone when I'm back ... you know, in case something is horribly wrong.

chrisnojima:
ok well you need it to be true on mobile else it won't be in testflight

chrisnojima:
and make a ticket to turn the feature flag on or leave the ticket in review or something

PR #12311, opened 2018-06-11
Remove VersionErrors from storage when our client is able to understand them

joshblum:
When we bump a `MessageBoxed`/`Header`/`Body` versions we store a `PermanentUnboxingError` on old clients since they cannot yet understand these messages. When these clients update, we want them to try to refetch the message from the server since they can now try to unboxed.

### DIFF
--- a/go/chat/storage/storage.go
+++ b/go/chat/storage/storage.go
@@ -897,7 +897,14 @@ func (s *Storage) FetchMessages(ctx context.Context, convID chat1.ConversationID
 			}
 		}
 		sres = rc.Result()
-		res = append(res, &sres[0])
+		msg := &sres[0]
+
+		// If we have a versioning error but our client now understands the new
+		// version, don't return the error message
+		if msg.IsError() && msg.Error().ParseableVersion() {
+			msg = nil
+		}
+		res = append(res, msg)
 	}
 
 	return res, nil
--- a/go/protocol/chat1/extras.go
+++ b/go/protocol/chat1/extras.go
@@ -339,12 +339,10 @@ func MessageUnboxedDebugLines(ms []MessageUnboxed) string {
 	return strings.Join(MessageUnboxedDebugStrings(ms), "\n")
 }
 
-type VersionKind string
-
 const (
 	VersionErrorMessageBoxed VersionKind = "messageboxed"
-	VersionErrorHeader                   = "header"
-	VersionErrorBody                     = "body"
+	VersionErrorHeader       VersionKind = "header"
+	VersionErrorBody         VersionKind = "body"
 )
 
 // NOTE: these values correspond to the maximum accepted values in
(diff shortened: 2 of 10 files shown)


go/protocol/chat1/extras.go
```diff
@@ -334,6 +335,52 @@ func MessageUnboxedDebugLines(ms []MessageUnboxed) string {
...
+
+// NOTE: these values correspond to the maximum accepted values in
+// chat/boxer.go. If these values are changed, they must also be accepted
+// there.
+var MaxMessageBoxedVersion MessageBoxedVersion = MessageBoxedVersion_V3
+var MaxHeaderVersion HeaderPlaintextVersion = HeaderPlaintextVersion_V1
+var MaxBodyVersion BodyPlaintextVersion = BodyPlaintextVersion_V1
+
+// Check if this error has a version that is now able to be understood by our client.
+func (m MessageUnboxedError) ParseableVersion() bool {
```
joshblum: we can also store version/kind info on this error type so we can drop the parsing later, should we?

go/protocol/chat1/extras.go
```diff
@@ -334,6 +335,52 @@ func MessageUnboxedDebugLines(ms []MessageUnboxed) string {
...
+)
+
+// NOTE: these values correspond to the maximum accepted values in
+// chat/boxer.go. If these values are changed, they must also be accepted
+// there.
+var MaxMessageBoxedVersion MessageBoxedVersion = MessageBoxedVersion_V3
+var MaxHeaderVersion HeaderPlaintextVersion = HeaderPlaintextVersion_V1
+var MaxBodyVersion BodyPlaintextVersion = BodyPlaintextVersion_V1
+
+// Check if this error has a version that is now able to be understood by our client.
```
mmaxim: Should use the function name first here for `godoc`

go/protocol/chat1/extras.go
mmaxim: Yes I think we should do that.

go/chat/storage/storage.go
```diff
@@ -897,7 +897,14 @@ func (s *Storage) FetchMessages(ctx context.Context, convID chat1.ConversationID
 			}
 		}
 		sres = rc.Result()
-		res = append(res, &sres[0])
+		msg := &sres[0]
+
+		// If we have a versioning error but our client now understands the new
+		// version, don't return the error message
+		if msg.IsError() && msg.Error().ParseableVersion() {
```
mmaxim: Think we should just stored this on the error like you said.

joshblum: we have to leave the parsing in for a few releases otherwise old clients wont' be able to refetch messages which was what the hope was for introducing V4 with reputability for ephemeral emssages in the next release

go/protocol/chat1/extras.go
joshblum: done

PR #12312, opened 2018-06-11
Force wallclock time for ephemeral calculations

joshblum:
@buoyad noticed that an ephemeral message had reverted to the original lifetime on one of his devices after a few days had passed. After a db nuke, the remaining lifetime was calculated correctly. I _think_ that the issue was caused by his device having been asleep and then monotonic clock being off when he received the message, which looked like the server was trying to extend the lifetime so he defaulted to the original (signed) lifetime. 

we try to `ForceWallClock` time wherever we do ephemeral calculations (even if it's redundant in some places since we are deserializing/serializing a `time.Time`, it will hopefully prevent future bugs from happening)

### DIFF
--- a/go/protocol/chat1/extras.go
+++ b/go/protocol/chat1/extras.go
@@ -420,6 +420,7 @@ func (m MessageUnboxedValid) Etime() gregor1.Time {
 }
 
 func (m MessageUnboxedValid) RemainingEphemeralLifetime(now time.Time) time.Duration {
+	now = keybase1.ForceWallClock(now)
 	remainingLifetime := m.Etime().Time().Sub(now).Round(time.Second)
 	return remainingLifetime
 }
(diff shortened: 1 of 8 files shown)


go/protocol/chat1/extras.go
```diff
@@ -420,6 +420,7 @@ func (m MessageUnboxedValid) Etime() gregor1.Time {
 }
 
 func (m MessageUnboxedValid) RemainingEphemeralLifetime(now time.Time) time.Duration {
+	now = keybase1.ForceWallClock(now)
 	remainingLifetime := m.Etime().Time().Sub(now).Round(time.Second)
```
joshblum: I think the original bug happened here, if the `now` that was passed in was well in the past compared to the wall clock time.

maxtaco: :+1:

joshblum:
i'm not convinced this is the issue actually after further debugging through the logs

PR #12314, opened 2018-06-12
WIP: styling

chrisnojima:
@songgao this is an example of a way to style this.
you can skip all the splitting if you are ok with each folder gettings its own line if its long (seems fine to me). if you do do the splitting (like on comma , but you could be smarter here) then the individual items could wrap nicely. depends on how much work you want to put into this.


songgao:
Thanks Nojima for the magical fix! I’ll play with it tomorrow.

PR #12315, opened 2018-06-12
Explicitely clear API server uidmap cache to avoid cache race flake

zapu:
The race happens during the following (or similar) path:
1) user A does something that changes their eldest_seqno: e.g. resets or reprovisions after reset,
2) user B queries `user/names` endpoint *immediately* after.

Even though API request in (1) finishes (be it `key/multi` for reprovisioning, or the reset API), the serverside uidmap cache is still in flux because pubsub fire-and-forget is used to propagate the change across API server fleet - even if it's just one server on CI. So if a request to `user/names` is made quickly enough, server will still return the old value, also poisoning the clientside cache.

I went through several attempts at fixing this "properly" but considering that: 
1) it's a relatively obscure bug, and
2) it only affects CI, because any real operations on client don't trust uidmap.

I decided that ultimately forcing server to clear its cache during tests is the way to go. Serverside cache clearing is a side effect of calling `user/names` with `no-cache` parameter. Not only will the server always pull the latest value from the DB, but will also update its own cache while on it. So any subsequent `user/names` call from any context will be able to get proper data.

This `clearServerUIDMapCache` should probably be added in test user functions like `loginAfterReset`, but I'm still investigating which tests are affected. In some, I went nuclear and disabled uidmap cache entirely, where a more subtle fix like this one would be more adequate.

### DIFF
--- a/go/service/team_handler.go
+++ b/go/service/team_handler.go
@@ -180,6 +180,9 @@ func (r *teamHandler) changeTeam(ctx context.Context, cli gregor1.IncomingInterf
 			continue
 		}
 
+		r.G().Log.CDebugf(ctx, "Checking reset badges: got total %d badges for team %q",
+			len(badges), row.Name)
+
 		// load this team and see if any of the badge users are active members
 		details, err := teams.Details(ctx, r.G(), row.Name, false)
 		if err != nil {
--- a/go/systests/team_reset_test.go
+++ b/go/systests/team_reset_test.go
@@ -680,10 +680,7 @@ func TestTeamAfterDeleteUser(t *testing.T) {
 	bob.readChats(team, 1)
 }
 
-// TestTeamResetBadges checks that badges show up for admins
-// when a member of the team resets, and that they are dismissed
-// when the reset user is added.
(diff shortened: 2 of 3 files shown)


zapu:
cc @patrickxb about the bug writeup

maxtaco:
@zapu thanks for tracking this down. I'm still missing the larger context of why we need to consult the uidmapper in this case. It wasn't really intended for this situation, but more so for large information queries. Can't we go to the source of truth here in these important cases, since we're only dealing with a small handful of users?

maxtaco:
As described over chat, I'm not a fan of this solution. I'd much rather that we fix the login in clearing the badge on the client side here: https://github.com/keybase/client/blame/master/go/service/team_handler.go#L173-L189

I don't think using `teams.Details` is correct here. I think we should ask the merkle tree for the UPAK and the team list, and clear the badge if the two match.   

To move things along, it's ok to merge this, but please file a ticket to fix this in the client, and then to undo the changes to the test that disable the uidmapper cache.  In general if we're doing that, it's just masking a deeper design bug, as in this case.

PR #12316, opened 2018-06-12
New waypoint+measure threadview

chrisnojima:
This is a complete rewrite of how the thread view works. My hope is the comments in the code is enough to describe whats going on. if not please point out whats confusing so we can update the comments


shared/chat/conversation/list-area/normal/index.stories.js
```diff
@@ -0,0 +1,214 @@
+// @flow
+import React from 'react'
```
chrisnojima: renders some mock messages and can be turned into a more interactive mode. uses a prng so its static

shared/chat/conversation/messages/index.js
```diff
@@ -15,7 +15,7 @@ import SetDescription from './set-description/container'
 import SetChannelname from './set-channelname/container'
 import Placeholder from './placeholder/container'
 import Wrapper from './wrapper/container'
-import {connect, compose, lifecycle, type TypedState} from '../../../util/container'
+import {setDisplayName, connect, compose, lifecycle, type TypedState} from '../../../util/container'
```
chrisnojima: make these components storybook compatible

chrisnojima:

shared/chat/conversation/list-area/normal/index.desktop.js
```diff
@@ -1,156 +1,162 @@
...
 
-  _onResize = ({width}) => {
-    if (this._cellCache.columnWidth({index: 0}) !== width) {
-      this._cellCache.clearAll()
+  _onResize = debounce(({bounds}) => {
+    if (this.state.isLockedToBottom) {
+      this._scrollToBottom()
     }
+
+    this.setState(p => (p.width === bounds.width ? null : {width: bounds.width}))
```
buoyad: I scrolled back 4 times or so and then made the screen a little wider, and the thread rerendered once but then was unresponsive for about a minute

buoyad: It's a little hard to repro, it might be that I'm catching this function multiple times so they end up rerendering on top of eachother. Maybe a longer debounce?

chrisnojima: yeah resizing is going to suck, unclear if there's a good fix actually

chrisnojima: lemme think on it, maybe i'll just remeasure the visible ones and just have more jank while you scroll

zanderz:
seems basically fine on Windows

shared/chat/conversation/list-area/normal/index.desktop.js
```diff
@@ -165,43 +160,251 @@ class Thread extends React.Component<Props, State> {
...
+// All the waypoints keep a key to re-render if we get a measure callback
+type TopBottomWaypointProps = {
+  conversationIDKey: Types.ConversationIDKey,
+  onPositionChange: () => void,
+}
+type TopBottomWaypointState = {
+  keyCount: number,
+}
+
+class TopWaypoint extends React.PureComponent<TopBottomWaypointProps, TopBottomWaypointState> {
```
buoyad: can this and `BottomWaypoint` be consolidated and take a prop to tell whether it's top / bottom?

mmaxim:
I just want to mention again it might be useful to think about removing the ordinal list altogether. In a world where the JS just hits the service everytime it wants to put messages on the screen, it doesn't seem necessary for the ordering information to be present in Redux at all. It seems like not having to maintain that list might be a perf win, as well as simplifying the code.

shared/chat/conversation/input-area/normal/container.js
```diff
@@ -102,7 +102,6 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps): Props => ({
     } else {
       dispatchProps._onPostMessage(stateProps.conversationIDKey, text)
     }
-    ownProps.onScrollDown()
```
chrisnojima: scrlldown is based on if you send something last and not a side effect of sending

chrisnojima:
We likely are going to merge this in for the release candidate Monday. It's changed enough that it needs rereview

chrisnojima:
@react-hackers

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -61,14 +61,12 @@ class ImageAttachment extends React.PureComponent<Props, State> {
...
-          <Box style={progressContainerStyle}>
-            <Text type={'BodySmall'} style={progressLabelStyle}>
-              {this.props.progressLabel}
-            </Text>
-            {this.props.hasProgress && <ProgressBar ratio={this.props.progress} />}
-          </Box>
-        )}
+        <Box style={progressContainerStyle}>
+          <Text type={'BodySmall'} style={progressLabelStyle}>
+            {this.props.progressLabel || '\u00A0'}
```
buoyad: ?

chrisnojima: added a comment to explain

shared/chat/conversation/list-area/normal/index.desktop.js
```diff
@@ -24,21 +24,31 @@ if (module.hot) {
...
 
 class Thread extends React.PureComponent<Props, State> {
-  state = {isLockedToBottom: true, isScrolling: false}
+  state = {isLockedToBottom: true}
   _listRef = React.createRef()
+  // so we can turn pointer events on / off
+  _pointerWrapperRef = React.createRef()
+  // Not a state so we don't rerender, just mutate the dom
+  _isScrolling = false
+  _programaticScrollToBottomRefCount = 0
```
buoyad: :sheep: `programmatic`

buoyad: Also could you add a comment about what this is for?

shared/chat/conversation/list-area/normal/index.desktop.js
```diff
@@ -160,74 +213,78 @@ class Thread extends React.PureComponent<Props, State> {
...
-      />
-    )
+    items.push(<BottomItem key="bottomItem" conversationIDKey={this.props.conversationIDKey} />)
 
-    return waypoints
+    return items
+  }
+
+  _onResize = ({scroll}) => {
+    if (this._scrollHeight) {
```
buoyad: Should this do something different if `isLockedToBottom`? You can get scrolled up on making the window wider.

PR #12317, opened 2018-06-12
Animate exploding messages on desktop

buoyad:
- Changes `ExplodingHeightRetainer` to throw away stored height on remount
  - Pipes `measure` through and uses `m.ordinal` instead of `m.id` to key the message

r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
+++ b/shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
@@ -21,7 +21,7 @@ const copyChildren = children =>
 
 const animationDuration = 1.5
 
-const messageHeights = {}
+const retainedHeights = {}
 
 type State = {
   children: ?React.Node,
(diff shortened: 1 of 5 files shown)


buoyad:
cc @cecileboucheron & @malgorithms if you'd like to change the colors, they're defined [here](https://github.com/keybase/client/pull/12317/files#diff-ecf15b17035331881283719eb6fdf71eR155). I chose arbitrarily from `globalColors`

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
```diff
@@ -4,40 +4,73 @@ import * as ReactDOM from 'react-dom'
...
 import {urlsToImgSet} from '../../../../../common-adapters/icon.desktop'
 import {Box, ConnectedUsernames, Text} from '../../../../../common-adapters'
-import {collapseStyles, globalColors, styleSheetCreate} from '../../../../../styles'
+import {
+  collapseStyles,
+  glamorous,
+  globalColors,
+  platformStyles,
+  styleSheetCreate,
+} from '../../../../../styles'
```
chrisnojima: so whats the native story here?

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
```diff
@@ -4,40 +4,73 @@ import * as ReactDOM from 'react-dom'
...
+  React.Children.map(children, child => (child ? React.cloneElement(child) : child))
+
+const animationDuration = 1.5
+
+const retainedHeights = {}
 
 type State = {
-  height: ?number,
+  animating: boolean,
+  children: ?React.Node,
```
chrisnojima: why do we need to clone the children?

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
```diff
@@ -4,40 +4,73 @@ import * as ReactDOM from 'react-dom'
...
     }
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: Props) {
     if (this.props.retainHeight) {
+      if (!prevProps.retainHeight && !this.timeoutID) {
+        // destroy local copy of children when animation finishes
+        this.setState({animating: true}, () => {
+          this.timeoutID = setTimeout(
```
chrisnojima: likely need to clear the timeout if this gets his if one exists still

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
buoyad: Nothing on native yet, I'll look into it today

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
buoyad: As soon as the message explodes the text is deleted from the store; cloning the children is so the text doesn't disappear while the animation is running

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
buoyad: done :+1: thanks

PR #12319, opened 2018-06-12
Improve logging and fix early deletion bug

joshblum:
Improves logging for ephemeral messages and catches a bug in key deletion where we were comparing the wrong units and deleting keys early. Device keys that were supposed to have an extended lifetime because of a gap in generations were deleted early

### DIFF
--- a/go/ephemeral/common.go
+++ b/go/ephemeral/common.go
@@ -122,30 +122,34 @@ type keyExpiryMap map[keybase1.EkGeneration]keybase1.Time
 // it is still used to encrypt a different key. This only applies to deviceEKs
 // or userEKs since they can have a dependency above them.  A teamEK expires
 // after `KeyLifetimeSecs` without exception, so it doesn't call this.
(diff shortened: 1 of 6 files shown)


go/ephemeral/common.go
```diff
@@ -122,30 +122,34 @@ type keyExpiryMap map[keybase1.EkGeneration]keybase1.Time
...
+		keyCtime := keyMap[generation]
 		if i < len(keys)-1 {
-			nextCTime = keyMap[keys[i+1]]
+			nextKeyCtime = keyMap[keys[i+1]]
 		} else {
-			nextCTime = nowCTime
+			nextKeyCtime = nowCtime
 		}
-		expiryOffset = nextCTime - currentCTime
-		if expiryOffset > KeyLifetimeSecs { // Offset can be max KeyLifetimeSecs
```
joshblum: wrong units

go/ephemeral/common.go
```diff
@@ -122,30 +122,34 @@ type keyExpiryMap map[keybase1.EkGeneration]keybase1.Time
...
-			expiryOffset = KeyLifetimeSecs
```
joshblum: wrong units

PR #12321, opened 2018-06-12
Add rtime as early as possible when unboxing

joshblum:
Addresses a bug where a client downloads a message from the server, sleeps and then later adds its received (`rtime`) extending the lifetime of an ephemeral message. We add the `rtime` as early as we can in the unboxing process


protocol/avdl/chat1/common.avdl
```diff
@@ -298,6 +298,10 @@
     // server's view of now(), used to calculate ephemeral lifetimes
     @mpackkey("n") @jsonkey("n")
     gregor1.Time now;
+    // clients mark the received time as soon as they pull the message from the
+    // server
+    @mpackkey("rt") @jsonkey("rt")
+    gregor1.Time rtime;
```
mmaxim: Ack actually could you make this a nullable thing?

mmaxim: I just don't want the server wasting any bandwidth on it

joshblum: yeah

PR #12325, opened 2018-06-12
add logging of pipeowner names on fail to GUI

zanderz:
at least one log came in with a pipeowner failure, but it wasn't being logged


shared/engine/index.platform.desktop.js
```diff
@@ -62,6 +62,17 @@ function checkRPCOwnership(): Promise<void> {
         resolve()
         return
       }
+      // log the names of the owners
+      const args2 = ['pipeowner', socketPath, '--names']
+      execFile(binPath, args2, {windowsHide: true}, (error, stdout, stderr) => {
```
songgao: It looks like this is the same call as above, except for the `--names` arg. If the command were printing `json` (e.g. `{"result": true}` or `{"result": false, names: [...]}`, we could have just had one call, and inspect the `json` which would have `names` populated if `!result`. Not a huge deal though.

zanderz: Good idea, I'll do that.

PR #12326, opened 2018-06-12
Remove more getDerivedStateFromProps

chrisnojima:
This removes some getDerivedStateFromProps that we didn't actually need. There's a couple of leftovers but maybe the authors should take a look


shared/app/nav.native.js
```diff
@@ -185,25 +184,17 @@ class MainNavStack extends Component<any, any> {
     this.setState({verticalOffset: frameData.height - (isIPhoneX ? 45 : 20)})
   }
 
-  static getDerivedStateFromProps(nextProps, prevState) {
-    const {routeSelected, routeStack} = nextProps
-    if (tabIsCached.hasOwnProperty(routeSelected)) {
```
chrisnojima: my understanding is we keep a subset of key/values around always. Instead we can just do this in the render function as it never triggers anything so it doesn't need to be in state

shared/chat/conversation/normal/index.desktop.js
```diff
@@ -41,11 +39,7 @@ const Offline = () => (
 
 class Conversation extends React.PureComponent<Props, State> {
   _mounted = false
-  state = {
-    conversationIDKey: null,
-    infoPanelOpen: false,
```
chrisnojima: infoPanelOpen isn't even used anymore so this is all removed

shared/common-adapters/avatar.render.native.js
```diff
@@ -99,11 +94,10 @@ class AvatarRender extends React.PureComponent<Props, State> {
     }
   }
 
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (nextProps.url !== prevState.url) {
-      return {loaded: false, url: nextProps.url}
+  componentDidUpdate(prevProps: Props) {
+    if (this.props.url !== prevProps.url) {
```
chrisnojima: we can just set the state after we update

shared/common-adapters/dropdown.js
```diff
@@ -21,24 +21,9 @@ type Props = {
...
-    }
-  }
-
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (nextProps.selected !== prevState.selected) {
-      return {selected: nextProps.selected}
-    }
-    return null
-  }
+  state = {expanded: false}
```
chrisnojima: no need to keep selected in local state at all

shared/common-adapters/floating-box.desktop.js
```diff
@@ -8,32 +8,24 @@ import type {Props} from './floating-box'
...
-      : {targetRect: null}
-  }
-
+class FloatingBox extends React.Component<Props> {
   _onHidden = () => {
     this.props.onHidden && this.props.onHidden()
   }
 
   render() {
+    let targetRect = null
```
chrisnojima: targetRect can be derived when we render

shared/common-adapters/input.native.js
```diff
@@ -14,8 +14,6 @@ import {checkTextInfo} from './input.shared'
 type State = {
   focused: boolean,
   height: ?number,
-  // Only changed for controlled components.
-  value?: string,
```
chrisnojima: apply the changes i made to input.desktop earlier today

shared/teams/team/list.native.js
```diff
@@ -10,10 +10,9 @@ type State = {
 class List extends React.Component<Props, State> {
   _renderSectionHeader = ({section}) => (section.key === 'body' ? this.props.renderRow(section.header) : null)
   _renderRow = data => this.props.renderRow(data.item)
-  state: State = {sections: []}
 
-  static getDerivedStateFromProps(nextProps: Props) {
-    const rows = nextProps.rows || []
+  render() {
+    const rows = this.props.rows || []
```
chrisnojima: just derive this when we render

chrisnojima:
read this for info: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

shared/actions/teams.js
```diff
@@ -594,12 +591,12 @@ const _getTeamPublicity = function*(action: TeamsGen.GetTeamPublicityPayload): S
   )
 
   let tarsDisabled = false
-  // Find out whether team access requests are enabled. Throws if you aren't admin.
-  if (yourOperations.changeTarsDisabled) {
```
chrisnojima: this logic is incorrect. i get maybe your operations could be not loaded yet or something? anyways, lets just load this and catch the error if it happens. else you can never load it (in my testing)

shared/common-adapters/error-boundary.js
```diff
@@ -106,17 +106,15 @@ type Props = {
...
 
 class ErrorBoundary extends React.PureComponent<Props, State> {
-  state = {info: null, children: null}
+  state = {info: null}
 
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (nextProps.children !== prevState.children) {
-      return {info: null, children: nextProps.children}
+  componentDidUpdate(prevProps: Props) {
+    if (this.props.children !== prevProps.children) {
```
chrisnojima: just clear after we render

shared/common-adapters/save-indicator.js
```diff
@@ -121,25 +121,6 @@ class SaveIndicator extends React.Component<Props, State> {
     this.state = {saving: false, lastSave: new Date(0), saveState: 'steady', lastJustSaved: new Date(0)}
   }
 
-  static getDerivedStateFromProps = (nextProps: Props, prevState: State) => {
-    if (nextProps.saving === prevState.saving) {
-      return null
-    }
-
-    // Just set saving and lastSave here -- run the state machine from
```
chrisnojima: just do this all in componentDidUpdate

shared/teams/team/settings-tab/index.js
```diff
@@ -180,35 +180,37 @@ const IgnoreAccessRequests = (props: SettingProps) =>
...
       publicitySettingsChanged: false,
       retentionPolicyChanged: false,
       retentionPolicyDecreased: false,
     }
   }
 
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    // We just got new settings for this team, reset any user selections
-    // to reflect the actual settings.
-    return {
```
chrisnojima: this is explicitly called out as being incorrect behavior in the react blog post, we can't just always return a new state w/o looking at the input at all

shared/common-adapters/floating-box.desktop.js
```diff
@@ -9,26 +9,41 @@ const StyleOnlyBox = (props: any) => <Box children={props.children} />
 const RelativeFloatingBox = ModalPositionRelative(StyleOnlyBox)
 
 type State = {
-  targetRect: ?ClientRect,
+  targetRect: ?React.Component<any, any>,
```
buoyad: I get a flow error in render:
```
Cannot create RelativeFloatingBox element because React.Component [1] is incompatible with ClientRect [2] in property targetRect.
```

chrisnojima: fixed

PR #12327, opened 2018-06-12
 fix android build: dex jumbo

chrisnojima:


chrisnojima:
@keybase/react-hackers the android build was failing. i think cause our binary was having too many methods (like we had on ios). 
```Execution failed for task ':app:transformClassesWithDexForRelease'.
> com.android.build.api.transform.TransformException: com.android.ide.common.process.ProcessException: java.util.concurrent.ExecutionException: com.android.dex.DexIndexOverflowException: Cannot merge new index 65545 into a non-jumbo instruction!```

this pr adds the jumbo flag which lets us get over this limit.

PR #12329, opened 2018-06-12
update outbox prev pointers to correct values after sprinkling into a thread CORE-8067

mmaxim:
The point of the patch is to update the `Prev` value (used by the JS to order pending messages) to the correct value in case the original prev is now no longer correct. This can happen under the followig circumstance:

1.) Send 8 messages in a row that stay pending.
2.) Watch 4 get sent.
3.) Click in and out of the thread.

Before this patch, the 4 remaining messages will appear above the messages sent. After this patch, they will show up at the bottom.

### DIFF
--- a/go/chat/storage/outbox.go
+++ b/go/chat/storage/outbox.go
@@ -399,20 +399,10 @@ func (o *Outbox) RemoveMessage(ctx context.Context, obid chat1.OutboxID) error {
 }
 
 func (o *Outbox) getMsgOrdinal(msg chat1.MessageUnboxed) chat1.MessageID {
-	typ, err := msg.State()
-	if err != nil {
-		return 0
+	if msg.IsValid() && msg.Valid().ClientHeader.OutboxInfo != nil {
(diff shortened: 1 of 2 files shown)


mmaxim:
Also if you approve, you should just merge this in case I am out of commission. It should get in before next release.

joshblum:
going to cleanup a few comments and then merge

go/chat/storage/outbox.go
```diff
@@ -408,6 +408,9 @@ func (o *Outbox) getMsgOrdinal(msg chat1.MessageUnboxed) chat1.MessageID {
 	case chat1.MessageUnboxedState_OUTBOX:
 		return msg.Outbox().Msg.ClientHeader.OutboxInfo.Prev
 	default:
+		if msg.IsValid() && msg.Valid().ClientHeader.OutboxInfo != nil {
```
joshblum: this can be simplified since the `OUTBOX` check was added to `GetMessageID`

PR #12330, opened 2018-06-12
Add setting to override notification sound with system default

buoyad:
In section `Security` but will probably move to `Phone - push notifications`


mmaxim:
Merging this since the code is already merged on the server side, which means I can't re-vendor `chat1` without breaking the server compile.

mmaxim:
Also @buoyad, in the future, it is usually a good idea to merge the client PRs first for this reason, otherwise the `vendor` directory on the server is out ahead and can end up getting reverted if someone revendors from master.

PR #12333, opened 2018-06-13
New component: WaitingButton

cjb:
@keybase/react-hackers 

`<WaitingButton />` is a new common-adapter that will either track a given `waitingKey` in the generic waiting store, or if none is passed, just spin when clicked until it unmounts.

This PR adds the component and also:

* switches `git/` from using its own loading bool in the store to using the generic waiting store and WaitingButton
* same for `app/push`
* hooks up the `wallets/loadEverything` saga to the waiting store and button
* makes `attachment-get-titles` use the non-waiting-key version
* adds a waiting button to `teams/edit-team-description`
* switches `teams/manage-channels` over to waitingbutton
* switches `devices` over to waitingbutton

(It's unsatisfying that I had to make a new mock PropProvider for tests since the WaitingButton is connected.. will keep thinking about that.)


shared/common-adapters/waiting-button.js
```diff
@@ -0,0 +1,47 @@
+// @flow
+import * as React from 'react'
+import Button, {type Props as ButtonProps} from './button'
+import {compose, connect, setDisplayName, withStateHandlers, type TypedState} from '../util/container'
+
+export type WaitingButtonProps = {
```
chrisnojima: lets add some comments that describe what this component is

chrisnojima:
re: the propsprovider thing, we could just have a common union of these, like defaultProvider that does waiting and usernames and avatars and etc

PR #12337, opened 2018-06-13
Enable ephemeral messages throughout

joshblum:
remove feature flags from ephemeral messages. Let's hold off on merging until we're cutting the release

cc @buoyad let's do the same for the gui?


buoyad:
I added the desktop flag too, r? @keybase/react-hackers

chrisnojima:
looks fine but we're waiting till monday

PR #12338, opened 2018-06-13
team.avdl: findNextMerkleRootAfterTeamRemoval

maxtaco:
- simple plug into existing framework
- a simple test that piggybacks on TestMemberRemove


maxtaco:
cc: @strib

PR #12342, opened 2018-06-13
fix team member remove flake

maxtaco:


zapu:
oh sorry for not thumbing this one, I looked at it before the call and thought it was a good play! thanks

PR #12351, opened 2018-06-13
metacontext-ify merkle client

maxtaco:
- this is one small step to meta-contextified api.go, but i wanted to split it up into small chunks


maxtaco:
need to rebase to rerun tests, etc.

PR #12352, opened 2018-06-13
Clear outbox periodically

joshblum:
Bundles outbox cleaning into the existing `EphemeralPurge` method which is called periodically (rn every 5m by `BackgroundEphemeralPurger`).We could have different time cutoffs for ephemeral vs regular messages but wanted to see your opinion first @mmaxim

### DIFF
--- a/go/chat/ephemeral_purger.go
+++ b/go/chat/ephemeral_purger.go
@@ -256,7 +256,7 @@ func (b *BackgroundEphemeralPurger) loop() {
 			b.queueLock.Unlock()
 		case <-b.outboxTicker.C:
 			// Check the outbox for stuck ephemeral messages that need purging
-			storage.NewOutbox(b.G(), b.uid).EphemeralPurge(context.Background())
+			storage.NewOutbox(b.G(), b.uid).OutboxPurge(context.Background())
 		case ch := <-b.stopCh: // caller will reset looping=false
 			b.Debug(bgctx, "loop: shutting down for %s", b.uid)
 			close(ch)
--- a/go/chat/storage/outbox.go
+++ b/go/chat/storage/outbox.go
@@ -26,7 +26,7 @@ type Outbox struct {
 }
 
 const outboxVersion = 4
-const ephemeralPurgeCutoff = time.Minute * 10
+const errorPurgeCutoff = time.Minute * 10
 
 type diskOutbox struct {
 	Version int                  `codec:"V"`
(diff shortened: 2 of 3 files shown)


go/chat/storage/outbox.go
```diff
@@ -197,7 +197,7 @@ func (o *Outbox) PullAllConversations(ctx context.Context, includeErrors bool, r
 		}
 	}
 	if remove {
-		// Write out box
+		// Write outbox
```
mmaxim: I think this was originally using "Write out" as the verb and "box" as noun

mmaxim:
10 minutes is way too short for what I had in mind here. I was thinking like 1 week.

joshblum:
ok cool we can split the logic for ephemeral and non-ephemeral

go/chat/ephemeral_purger.go
```diff
@@ -256,7 +256,7 @@ func (b *BackgroundEphemeralPurger) loop() {
 			b.queueLock.Unlock()
 		case <-b.outboxTicker.C:
 			// Check the outbox for stuck ephemeral messages that need purging
-			storage.NewOutbox(b.G(), b.uid).EphemeralPurge(context.Background())
+			storage.NewOutbox(b.G(), b.uid).OutboxPurge(context.Background())
```
mmaxim: I'm not too excited about this being a general thing invoked from the exploding purger. Is there any other way that isn't too much work so that we can decouple this error purger from this loop?

joshblum: we can move it back into the service i don't think it would be that much extra

go/chat/storage/outbox.go
```diff
@@ -522,19 +531,26 @@ func (o *Outbox) EphemeralPurge(ctx context.Context) (err error) {
...
+	if len(ephemeralPurged) > 0 {
 		act := chat1.NewChatActivityWithFailedMessage(chat1.FailedMessageInfo{
-			OutboxRecords:    purged,
+			OutboxRecords:    ephemeralPurged,
 			IsEphemeralPurge: true,
 		})
 		o.G().NotifyRouter.HandleNewChatActivity(context.Background(),
 			keybase1.UID(o.GetUID().String()), &act)
 	}
+	if len(purged) > 0 {
```
mmaxim: I don't think you need this, the main goal here is to just get these things out of the box, not get them off the screen, that can just happen whenever.

joshblum: makes sense, we just need it for ephemeral stuff

go/service/main.go
```diff
@@ -566,6 +567,31 @@ func (d *Service) writeServiceInfo() error {
...
+	m := libkb.NewMetaContextBackground(d.G()).WithLogTag("OBOXPRGE")
+	d.G().PushShutdownHook(func() error {
+		m.CDebugf("stopping chatOutboxPurgeCheck loop")
+		ticker.Stop()
+		return nil
+	})
+	go func() {
+		for {
+			<-ticker.C
+			m.CDebugf("| checking outbox purge")
```
mmaxim: Can we delete this?

joshblum: done

PR #12354, opened 2018-06-13
Don't reuse any message components

buoyad:
Was causing the exploding animation to re-explode messages sometimes. r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/list-area/normal/index.desktop.js
+++ b/shared/chat/conversation/list-area/normal/index.desktop.js
@@ -5,6 +5,7 @@ import * as React from 'react'
 import Message from '../../messages'
 import SpecialTopMessage from '../../messages/special-top-message'
 import SpecialBottomMessage from '../../messages/special-bottom-message'
+import {ordinalToNumber} from '../../../../constants/types/chat2'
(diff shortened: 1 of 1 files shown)


buoyad:
cc @malgorithms

shared/chat/conversation/list-area/normal/index.desktop.js
```diff
@@ -26,12 +27,13 @@ class Thread extends React.Component<Props, State> {
...
-        return 'specialBottom'
+        return `${this.props.conversationIDKey}:specialBottom`
       } else if (index === 0) {
-        return 'specialTop'
+        return `${this.props.conversationIDKey}:specialTop`
       } else {
         const ordinalIndex = index - 1
-        return this.props.messageOrdinals.get(ordinalIndex)
+        const ordinal = this.props.messageOrdinals.get(ordinalIndex)
+        return `${this.props.conversationIDKey}:${ordinal ? ordinalToNumber(ordinal) : 'null'}`
```
chrisnojima: better use index instead of null

chrisnojima:
this could solve some measuring issues

PR #12360, opened 2018-06-14
RPC: teams.findNextMerkleRootAfterTeamRemovalBySigningKey

maxtaco:
- a thin wrapper around libkb.FindNextMerkleRootAfterTeamRemoval, for use from KBFS,
who knows the user's UID and signing key from the MD, and wants to work backward to when,
if ever, this user was kicked out of the team, or downgraded to writer.


maxtaco:
cc: @strib

PR #12362, opened 2018-06-14
fix globals race in Shutdown v Logout

maxtaco:
- make sure we hold the cacheMu when doing both shutdown and logout
- small refactor to dedup code

### DIFF
--- a/go/libkb/globals.go
+++ b/go/libkb/globals.go
@@ -396,20 +396,33 @@ func (g *GlobalContext) ConfigureAPI() error {
 	return nil
 }
 
-func (g *GlobalContext) configureMemCachesLocked(isFlush bool) {
-	// shutdown any existing ones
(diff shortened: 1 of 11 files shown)


maxtaco:
ok @patrickxb now it should be ready for review.

go/libkb/globals.go
```diff
@@ -396,29 +399,62 @@ func (g *GlobalContext) ConfigureAPI() error {
 	return nil
 }
 
-func (g *GlobalContext) configureMemCachesLocked(isFlush bool) {
-	// shutdown any existing ones
-	if g.TrackCache != nil {
-		g.TrackCache.Shutdown()
+// shutdownCachesLocked shutsown and non-nil caches that have running goroutines
```
patrickxb: shutsdown any

PR #12363, opened 2018-06-14
Pass ToAssertion field that we may get from API server

zapu:
It's only returned for relay payments and only to the sender of the payment.


go/client/stellar_common.go
```diff
@@ -41,11 +41,15 @@ func printPayment(g *libkb.GlobalContext, p stellar1.PaymentCLILocal, verbose bo
 	}
 	var to string
 	switch {
+	case p.ToUsername != nil && p.ToAssertion != nil:
```
mlsteele: Maybe you want to add `&& *p.ToUsername != *p.ToAssertion`? Because isn't it the case that in the usual path the sender posts the ToAssertion as the username they're sending to?

zapu: hmm in the case where recipient has no wallet? good idea, I missed this

mlsteele: I meant regardless of whether the recipient has a wallet, I thought that assertion=username was posted. Why would whether the recipient had a wallet affect it?

zapu: my understanding of when ToAssertion is set (or returned) is a bit messy then. I thought it's only for relay payments hence if the recipient has a wallet, it's never a relay, so this field will never be used there

mlsteele: Oh yeah you're absolutely right. Yes, for that case then :D

PR #12364, opened 2018-06-14
remove AlreadyDeleted error from outbox

joshblum:
wrap the `AlreadyDeleted` error in a new `OutboxError` type so we can clear it from the user's outbox. if a user tried to delete a message that was wiped out from a background process (ephemeral purge, retention or maybe they were offline and another device deleted it) this error message would hang around in the outbox in definitely. https://github.com/keybase/client/pull/12352 would clear this out after some time but i don't think the error should ever be user facing 

cc @mmaxim

### DIFF
--- a/go/chat/sender.go
+++ b/go/chat/sender.go
@@ -834,15 +834,21 @@ func (s *Deliverer) doNotRetryFailure(ctx context.Context, obr chat1.OutboxRecor
 }
 
 func (s *Deliverer) failMessage(ctx context.Context, obr chat1.OutboxRecord,
-	oserr chat1.OutboxStateError) error {
+	oserr chat1.OutboxStateError) (err error) {
 	var marked []chat1.OutboxRecord
-	var err error
 	switch oserr.Typ {
 	case chat1.OutboxErrorType_TOOMANYATTEMPTS:
 		s.Debug(ctx, "failMessage: too many attempts failure, marking whole outbox failed")
 		if marked, err = s.outbox.MarkAllAsError(ctx, oserr); err != nil {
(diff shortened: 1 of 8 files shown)


mmaxim:
You could give this the same treatment as we do the double send error

mmaxim:
It might be better from a consistency standpoint

joshblum:
i originally had it in the switch with `OutboxErrorType_DUPLICATE` to not send the `NewChatActivityWithFailedMessage` notification but the ui would leave a sending icon around so i removed it:

screen shot 2018-06-14 at 11 33 07 am

mmaxim:
That's their bug, they aren't doing something right with the sending animation and pending messages. I filed another bug a couple days ago about a similar problem.

@chrisnojima  ^

go/chat/sender.go
```diff
@@ -845,6 +845,10 @@ func (s *Deliverer) failMessage(ctx context.Context, obr chat1.OutboxRecord,
 				s.outbox.GetUID(), err.Error())
 			return err
 		}
+	case chat1.OutboxErrorType_ALREADY_DELETED:
```
joshblum: what's the reason we don't do this for `OutboxErrorType_DUPLICATE`?

joshblum:
gotcha i'll change it back

go/chat/sender.go
mmaxim: I guess I just didn't care. This is fine too, but maybe also bring duplicate in too? mayb can combine those two switches.

joshblum: sgtm

joshblum:
@mmaxim merging but lmk if you want any subsequent changes

PR #12369, opened 2018-06-14
Stellar: CreateWalletAccountLocal

zapu:


go/stellar/stellarsvc/frontend_test.go
```diff
@@ -241,14 +241,25 @@ func TestLinkNewWalletAccountLocal(t *testing.T) {
 	require.NoError(t, err)
```
mlsteele: btw you can get of this CreateWallet thing in tests now because
1. creating the initial wallet is synchronous now
2. most wallet RPCs will create one if they need one
and it's nice to test that stuff

go/stellar/stellarsvc/frontend_test.go
```diff
@@ -241,14 +241,25 @@ func TestLinkNewWalletAccountLocal(t *testing.T) {
...
 
 	// link a new account
-	a1, s1 := randomStellarKeypair()
-	arg := stellar1.LinkNewWalletAccountLocalArg{
-		SecretKey: s1,
-		Name:      "my other account",
+	var linkID stellar1.AccountID
```
mlsteele: linkID seems like it would be the sigchain link

go/stellar/stellarsvc/frontend_test.go
zapu: good idea!

PR #12370, opened 2018-06-14
move PairwiseMacs from MessageBoxed to ClientHeader

oconnor663:
MessageBoxed gets pieced together by the server, and it was a mistake to
add the new field there. ClientHeader gets messagepack-deserialized from
the protocol, so putting the field there instead will have the behavior
we were originall hoping for.

r? @joshblum 

cc @mmaxim

### DIFF
--- a/protocol/avdl/chat1/common.avdl
+++ b/protocol/avdl/chat1/common.avdl
@@ -379,6 +379,12 @@
 
     @mpackkey("em") @jsonkey("em")
     union { null, MsgEphemeralMetadata } ephemeralMetadata;
+
+    // [V1, V2]: Missing
+    // [V3, V4]: Optional map of pairwise MACs, used for exploding messages on
+    //           small teams. In V4, if MACs are present, the verifyKey is a dummy.
+    @mpackkey("pm") @jsonkey("pm")
+    map<keybase1.KID, bytes> pairwiseMacs;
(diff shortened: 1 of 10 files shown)


oconnor663:
After this lands, I'll do a vendoring PR on the server, and then I'll do a third PR that switches on pairwise MACs and doesn't break tests :)

protocol/avdl/chat1/common.avdl
```diff
@@ -379,6 +379,11 @@
 
     @mpackkey("em") @jsonkey("em")
     union { null, MsgEphemeralMetadata } ephemeralMetadata;
+
+    // [V1, V2]: Missing
+    // [V3, V4]: Optional map of pairwise MACs, used for exploding messages on
+    //           small teams. In V4, if MACs are present, the verifyKey is a dummy.
+    map<keybase1.KID, bytes> pairwiseMacs;
```
joshblum: can you add `@mpackkey("pm) @jsonkey("pm")` here to keep the key sizes small ?

joshblum: missed this in the initial review sorry

oconnor663:
I'm going to ignore the "General SCM" errors and land away. @mmaxim gave a :+1: over chat.

PR #12372, opened 2018-06-14
use a dummy signing key when pairwise MAC'ing, bumping version to V4

oconnor663:
This depends on protocol changes in https://github.com/keybase/client/pull/12370, including getting those changes vendored on the server (not yet submitted). Tests will fail until that's done.

r? @joshblum

### DIFF
--- a/go/chat/boxer.go
+++ b/go/chat/boxer.go
@@ -742,19 +742,27 @@ func (b *Boxer) unboxV2orV3orV4(ctx context.Context, boxed chat1.MessageBoxed,
 	// even though signEncryptOpen will check a signature in the end, we no
 	// longer care what signing key it's using. That means we're compatible
 	// with two different scenarios, to support gradual rollout:
-	// 1) Pairwise MACs are included, but also the signing key is still the
-	//    same as in regular messages. Clients with MAC support can check MACs,
-	//    but clients without MAC support ignore the MACs and keep working. We
-	//    don't have repudiability yet, but we can gradually roll out support.
-	// 2) Pairwise MACs are included, and the signing key is an all-zeros
-	//    dummy. Clients with MAC support won't notice this change. At this
-	//    point clients without MAC support will break (so we should pair this
-	//    with a version bump), and we will finally have repudiability.
+	// V3) Pairwise MACs are included, but also the signing key is still the
+	//     same as in regular messages. Clients with MAC support can check MACs,
+	//     but clients without MAC support ignore the MACs and keep working. We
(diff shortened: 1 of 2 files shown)


go/chat/boxer.go
```diff
@@ -1318,12 +1345,19 @@ func (b *Boxer) BoxMessage(ctx context.Context, msg chat1.MessagePlaintext,
...
-			// but we do have backwards compatibility.
+			// As noted above, bump the version to V4 when we're MAC'ing.
+			if version == chat1.MessageBoxedVersion_V3 {
+				version = chat1.MessageBoxedVersion_V4
+			}
+			// Replace the signing key with a dummy. Using the real signing key
+			// would sabotage the repudiability that pairwise MACs are
+			// providing. We could avoid signing entirely, but this approach
+			// keeps the difference between the two modes very small.
+			signingKeyPair = dummySigningKey()
```
oconnor663: The real change is right here (and the version bump immediately above).

go/chat/boxer_test.go
```diff
@@ -370,6 +374,58 @@ func TestChatMessageUnboxWithPairwiseMacsMissing(t *testing.T) {
...
+	if boxed.ClientHeader.OutboxID == msg.ClientHeader.OutboxID {
+		t.Fatalf("defective test: %+v   ==   %+v", boxed.ClientHeader.OutboxID, msg.ClientHeader.OutboxID)
+	}
+
+	// need to give it a server header...
+	boxed.ServerHeader = &chat1.MessageServerHeader{
+		Ctime: gregor1.ToTime(time.Now()),
+	}
+
+	unboxed, err := boxer.unbox(context.TODO(), *boxed, chat1.ConversationMembersType_KBFS, key, nil)
```
joshblum: i thought we disabled this for `KBFS` type?

oconnor663: Ah this is some questionable copypasta. The reason this has no effect is that that param is used in `UnboxMessage`, but these tests are calling the lower level `unbox` function, and manually passing in the values like `key` that that higher level function would normally load for us.

oconnor663:
Ugh, this got closed because the upstream branch got deleted.

go/chat/boxer_test.go
joshblum: ah ok, let's change it for consistency at least

oconnor663:
Will use https://github.com/keybase/client/pull/12374 for the CI run.

go/chat/boxer_test.go
oconnor663: Went ahead and changed it.

PR #12375, opened 2018-06-14
check GetExtendedStatus for revokes and resets

maxtaco:
- we're now going to the API for each user section on the config file to check that the listed
users are still unrevoked/unreset. we'll filter that list down before sending it back up
to the UI
- this is now slower, since there's a merkle API path call for each such check. but we're not really
expecting users to have more than a small handful of sockpuppets
- tested by hand

### DIFF
--- a/go/libkb/all_provisioned_usernames.go
+++ b/go/libkb/all_provisioned_usernames.go
@@ -1,0 +1,59 @@
+package libkb
+
+func getUsernameIfProvisioned(m MetaContext, uc UserConfig) (ret NormalizedUsername, err error) {
+	m.CDebugf("getUsernameIfProvisioned(%+v)", uc)
+	did := uc.GetDeviceID()
+	if did.IsNil() {
+		m.CDebugf("- no valid username since nil deviceID")
+		return ret, nil
+	}
+	err = checkDeviceValidForUID(m.Ctx(), m.G().GetUPAKLoader(), uc.GetUID(), did)
+	switch err.(type) {
+	case nil:
+		m.CDebugf("- checks out")
+		return uc.GetUsername(), nil
+	case DeviceNotFoundError:
+		m.CDebugf("- user was likely reset (%s)", err)
+		return ret, err
+	case KeyRevokedError:
+		m.CDebugf("- device was revoked (s)", err)
+		return ret, err
+	default:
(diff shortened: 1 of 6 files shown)


maxtaco:
cc: @chrisnojima 

This wasn't causing the bug that I showed you, but there are other cases I thought (like if the device is offline at the time of the reset) where we could indeed cause that bug.

go/libkb/all_provisioned_usernames.go
```diff
@@ -0,0 +1,59 @@
...
+		current, err = getUsernameIfProvisioned(m, *currentUC)
+		if err != nil {
+			return current, nil, err
+		}
+	}
+
+	for _, u := range allUCs {
+		tmp, err := getUsernameIfProvisioned(m, u)
+		if err != nil {
+			return current, nil, err
```
patrickxb: don't you just want a continue here?

maxtaco: `getUsernameIfProvisioned` should only error when the error should propagate all the way up.

go/libkb/all_provisioned_usernames.go
```diff
@@ -0,0 +1,59 @@
...
+		return ret, nil
+	}
+	err = checkDeviceValidForUID(m.Ctx(), m.G().GetUPAKLoader(), uc.GetUID(), did)
+	switch err.(type) {
+	case nil:
+		m.CDebugf("- checks out")
+		return uc.GetUsername(), nil
+	case DeviceNotFoundError:
+		m.CDebugf("- user was likely reset (%s)", err)
+		return ret, err
```
patrickxb: isn't this propagating all the way up?  and you just want to skip over this one in the loop?

go/libkb/all_provisioned_usernames.go
```diff
@@ -0,0 +1,59 @@
...
+	case KeyRevokedError:
+		m.CDebugf("- device was revoked (s)", err)
+		return ret, err
```
patrickxb: and this?

go/libkb/all_provisioned_usernames.go
maxtaco: yeah, my bad. refresh, thank you.

PR #12379, opened 2018-06-14
Fix native styling for edit msg input

buoyad:
Also make sure bomb icon never shows while editing.
Before: 
editing_before
After:
editing_after
r? @keybase/react-hackers


adamjspooner:
Did something change with the time badge? There's some extra space on the bottom.

Here's what I see on the desktop:
screen shot 2018-06-14 at 17 17 30

buoyad:
@adamjspooner no, this PR shouldn't affect it

chrisnojima:
also that left part should have flexShrink: 0 so it doesn't collapse

PR #12380, opened 2018-06-14
uint overflow fix

joshblum:
Was seeing a panic when we tried to `clearUpthrough` an overflowed uint so we abort early to fix. I tried grepping around to see if we had any similar bugs and there may have been one in the `storage_blockengine` I think there may still be a bug though, i'm not sure why we call `ClearBefore` (https://github.com/keybase/client/compare/joshblum/convload-panic-CORE-8082?expand=1#diff-cf8add58c26f1cbee6e6e1e5f4a20d6eR1091) on a `DELETE` message -- don't we only want to clear out the single message from the `DELETE` instead of everything `upto`?

### DIFF
--- a/go/chat/convsource.go
+++ b/go/chat/convsource.go
@@ -1066,7 +1066,7 @@ func (s *HybridConversationSource) mergeMaybeNotify(ctx context.Context,
 }
 
 // ClearFromDelete clears the current cache if there is a delete that we don't know about
-// and returns true to the caller if it schedule a background loader job
+// and returns true to the caller if it schedules a background loader job
 func (s *HybridConversationSource) ClearFromDelete(ctx context.Context, uid gregor1.UID,
 	convID chat1.ConversationID, deleteID chat1.MessageID) bool {
 	defer s.Trace(ctx, func() error { return nil }, "ClearFromDelete")()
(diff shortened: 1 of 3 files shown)


mmaxim:
How was the param overflown? What was the root cause?

joshblum:
if our max message is a `DELETE` we call `ClearFromDelete` in the background: https://github.com/keybase/client/blob/master/go/chat/sync.go#L407 but don't find a delete so we call expunge:
https://github.com/keybase/client/blob/master/go/chat/convsource.go#L1076 which then finds a min message ID of `0` which `ClearBefore` passes  as `upto - 1= -1` to `clearUpthrough` 
 
I'm a little confused with the logic for calling the expunge stuff it. Why do we call expunge on a single `DELETE` and not a `DELETE_HISTORY`? Things seems to have changed here https://github.com/keybase/client/commit/670aa984651bcbe6a18d88aff8e4af49abb4e6a7#diff-cf8add58c26f1cbee6e6e1e5f4a20d6eL1036 when we used to call `storage.Expunge` which aborted on `Upto=0` https://github.com/keybase/client/blob/master/go/chat/storage/storage.go#L635

PR #12382, opened 2018-06-14
Fix native attachment preview not rendering

buoyad:
Also fixes max width for attachments and attachment upload error handling. cc @mmaxim r? @keybase/react-hackers

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -1599,9 +1599,10 @@ function* attachmentUpload(action: Chat2Gen.AttachmentUploadPayload) {
     if (ordinal) {
       yield Saga.put(Chat2Gen.createAttachmentUploaded({conversationIDKey, ordinal}))
     }
-  } catch (e) {}
-  // TODO better error
-  logger.warn('Upload Attachment Failed')
+  } catch (e) {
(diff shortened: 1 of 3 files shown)


shared/actions/chat2/index.js
```diff
@@ -1599,9 +1599,10 @@ function* attachmentUpload(action: Chat2Gen.AttachmentUploadPayload) {
     if (ordinal) {
       yield Saga.put(Chat2Gen.createAttachmentUploaded({conversationIDKey, ordinal}))
     }
-  } catch (e) {}
-  // TODO better error
-  logger.warn('Upload Attachment Failed')
+  } catch (e) {
+    // TODO better error
+    logger.warn(`Upload Attachment Failed: ${e}`)
```
jzila: debugging left in on purpose?

buoyad: An earlier PR moved the `logger.warn` out of the `catch`, so it was getting logged on every run.

mmaxim:
Does this make it render not right on the right side of the screen?

buoyad:
@mmaxim Yup

buoyad:
I think the larger avatars broke that originally

PR #12393, opened 2018-06-15
fs: implement upload

songgao:
This PR has upload implemented for both desktop and mobile, and prepare for overlaying KBFS journal status. It doesn't track journal status yet. That is, GUI for now doesn't know whether a file is in journal only, or has been done uploading. It also doesn't know whether the journal is still uploading anything or is idle. I will have a separate PR for it.

### DIFF
--- a/shared/actions/fs/common.native.js
+++ b/shared/actions/fs/common.native.js
@@ -2,6 +2,8 @@
 import * as Saga from '../../util/saga'
 import * as FsGen from '../fs-gen'
 import {putActionIfOnPath, navigateAppend} from '../route-tree'
+import {showImagePicker} from 'react-native-image-picker'
+import {isIOS} from '../../constants/platform'
 
 export const share = (action: FsGen.SharePayload) =>
   Saga.put(
@@ -31,3 +33,18 @@ export function* save(action: FsGen.SavePayload): Saga.SagaGenerator<any, any> {
     )
   )
 }
+
+export const pickAndUpload = ({payload: {type}}: FsGen.PickAndUploadPayload) =>
+  new Promise((resolve, reject) =>
+    showImagePicker(
+      {mediaType: 'photo'}, // TODO: support other types
+      response =>
+        !response.didCancel &&
+        (response.error
+          ? reject(response.error)
+          : resolve(isIOS ? response.uri.replace('file://', '') : response.path))
+    )
+  )
+
+export const pickAndUploadSuccess = (localPath: string, action: FsGen.PickAndUploadPayload) =>
+  localPath && Saga.put(FsGen.createUpload({localPath, parentPath: action.payload.parentPath}))
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -119,7 +119,7 @@ function* folderList(action: FsGen.FolderListLoadPayload): Saga.SagaGenerator<an
   yield Saga.put(FsGen.createFolderListLoaded({pathItems, path: rootPath}))
 }
 
-function* monitorTransferProgress(key: string, opID: RPCTypes.OpID) {
+function* monitorDownloadProgress(key: string, opID: RPCTypes.OpID) {
   // This loop doesn't finish on its own, but it's in a Saga.race with
   // `SimpleFSWait`, so it's "canceled" when the other finishes.
   while (true) {
@@ -129,7 +129,7 @@ function* monitorTransferProgress(key: string, opID: RPCTypes.OpID) {
       continue
     }
     yield Saga.put(
-      FsGen.createTransferProgress({
+      FsGen.createDownloadProgress({
         key,
         endEstimate: progress.endEstimate,
         completePortion: progress.bytesWritten / progress.bytesTotal,
@@ -175,8 +175,7 @@ function* download(action: FsGen.DownloadPayload): Saga.SagaGenerator<any, any> 
   const key = Constants.makeDownloadKey(path, localPath)
 
   yield Saga.put(
-    FsGen.createTransferStarted({
-      type: 'download',
+    FsGen.createDownloadStarted({
       key,
       path,
       localPath,
@@ -200,13 +199,13 @@ function* download(action: FsGen.DownloadPayload): Saga.SagaGenerator<any, any> 
 
   try {
     yield Saga.race({
-      monitor: Saga.call(monitorTransferProgress, key, opID),
+      monitor: Saga.call(monitorDownloadProgress, key, opID),
       wait: Saga.call(RPCTypes.SimpleFSSimpleFSWaitRpcPromise, {opID}),
     })
 
     // No error, so the download has finished successfully. Set the
     // completePortion to 1.
-    yield Saga.put(FsGen.createTransferProgress({key, completePortion: 1}))
+    yield Saga.put(FsGen.createDownloadProgress({key, completePortion: 1}))
 
     const mimeType = yield Saga.call(_loadMimeType, path)
 
@@ -215,11 +214,11 @@ function* download(action: FsGen.DownloadPayload): Saga.SagaGenerator<any, any> 
     intentEffect && (yield intentEffect)
   } catch (error) {
     console.log(`Download for intent[${intent}] error: ${error}`)
-    yield Saga.put(FsGen.createTransferFinished({key, error}))
+    yield Saga.put(FsGen.createDownloadFinished({key, error}))
     return
   }
 
-  yield Saga.put(FsGen.createTransferFinished({key}))
+  yield Saga.put(FsGen.createDownloadFinished({key}))
 }
 
 function* upload(action: FsGen.UploadPayload) {
@@ -227,28 +226,8 @@ function* upload(action: FsGen.UploadPayload) {
   const opID = Constants.makeUUID()
   const name = Types.getLocalPathName(localPath)
   const path = Types.pathConcat(parentPath, name)
-  const key = Constants.makeUploadKey(localPath, path)
 
-  // Call stat to figure out path type.
-  const dirent = yield Saga.call(RPCTypes.SimpleFSSimpleFSStatRpcPromise, {
-    path: {
-      PathType: RPCTypes.simpleFSPathType.local,
-      local: localPath,
-    },
-  })
-  const entryType = Types.direntToPathType(dirent)
-
-  yield Saga.put(
-    FsGen.createTransferStarted({
-      type: 'upload',
-      entryType,
-      key,
-      path,
-      localPath,
-      intent: 'none',
-      opID,
-    })
-  )
+  yield Saga.put(FsGen.createUploadStarted({path}))
 
   // TODO: confirm overwrites?
   // TODO: what about directory merges?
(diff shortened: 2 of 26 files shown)


songgao:
Please hold off for review. I'm likely gonna make more changes to the store.

songgao:
@keybase/react-hackers OK I think this is finally ready again!

shared/fs/header/add-new-container.js
```diff
@@ -4,14 +4,49 @@ import * as FsGen from '../../actions/fs-gen'
...
+  isDarwin
+    ? [
+        {
+          onClick: () => _upload(path, 'both'),
+          icon: 'iconfont-upload',
+          title: 'Upload a file or folder',
+        },
+      ]
+    : [
+        // Linux/Windows don't support accepting file and dir at the same time.
```
jzila: What does this mean?

shared/reducers/fs.js
```diff
@@ -31,31 +32,40 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
+
+        if (originalFolder.progress === 'loaded' && newItem.progress === 'pending') {
+          // We don't want to override a loaded folder into pending, because
+          // otherwise next time user goes into that folder we'd show
+          // placeholders. We also don't want to simply use the original
+          // PathItem, since it's possible some metadata has updated. So use
+          // the new item, but reuse children, progress, and tlfMeta fields.
+          if (originalFolder.type === 'folder' && newItem.type === 'folder') {
+            // make flow happy
+            newItem = newItem.set('children', originalFolder.children).set('progress', 'loaded')
```
jzila: What about original's other properties?

shared/actions/fs/common.native.js
```diff
@@ -4,22 +4,39 @@ import * as Types from '../../constants/types/fs'
...
   Saga.sequentially([
     Saga.put(FsGen.createDownload({intent: 'camera-roll', path})),
-    getTransferPopupAction(path, routePath),
+    getDownloadPopupAction(path, routePath),
   ])
+
+export const pickAndUpload = ({payload: {type}}: FsGen.PickAndUploadPayload) =>
+  new Promise((resolve, reject) =>
+    showImagePicker(
+      {mediaType: isIOS ? 'mixed' : 'photo'}, // 'mixed' is not supported on Android. TODO: find something better.
```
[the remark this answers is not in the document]
songgao: Yes, like chat, Android uploads only supports image.

I brought this up and found a couple libraries earlier when I started working on this, and @chrisnojima said we can discuss what library to pull in. So (I think in a sprint meeting) we decided to go with image only for now, and extend it later.

Better than the vanilla RN one, there are also:

https://github.com/Elyx0/react-native-document-picker

https://github.com/luisfuertes/react-native-file-picker

The iOS one even supports picking from other apps!

shared/actions/fs/index.js
```diff
@@ -232,17 +275,31 @@ function* pollSyncStatusUntilDone(): Saga.SagaGenerator<any, any> {
   }
   polling = true
   try {
-    let status: RPCTypes.FSSyncStatus = yield Saga.call(RPCTypes.SimpleFSSimpleFSSyncStatusRpcPromise)
-    if (status.totalSyncingBytes <= 0) {
-      return
-    }
+    while (1) {
```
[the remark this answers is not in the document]
songgao: If a RPC times out, the saga would just fail. Are you suggesting we put it in a try/catch block? I guess we should at least reset the `polling` var.

shared/fs/container.js
```diff
@@ -39,69 +36,107 @@ const mapStateToProps = (state: TypedState, {path}) => {
...
+): Array<SortableRowItem> =>
+  stills.map(still => {
+    const {name, type, path} = still
+    if (type === 'folder') {
+      // Don't show an upload row for folders.
+      return still
+    }
+    if (!uploads.writingToJournal.has(path) && !uploads.syncingPaths.has(path)) {
+      // The entry is absent from uploads. So just show a still row.
+      return still
```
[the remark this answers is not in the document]
songgao: Hmm I actually prefer to have them separate for better readability.

shared/fs/header/add-new-container.js
songgao: You can have a popup modal dialog to choose either a file, or a folder, on macOS. On Linux/Windows, you have to decide whether you want a file or a folder, and that popup is just for that.

shared/reducers/fs.js
songgao: Good catch. I think this is missing tlfMeta, etc.

shared/fs/header/add-new-container.js
jzila: Hmmm, that's pretty terrible UX. I'd almost rather just do files then. Do people often want to do whole-folder actions on mobile?

shared/actions/fs/common.native.js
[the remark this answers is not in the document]
jzila: Yeah I'd prefer to avoid having separate libraries for Android vs iOS.

shared/fs/header/add-new-container.js
songgao: I feel like it can be useful, and it's in the [design](https://app.zeplin.io/project/56b25ffc64fc91773a73aaf8/screen/5a7212277692923af00a8836). Would it be less confusing if we just split them into separate items on all platform?

jzila: That's the desktop design, and it's definitely useful there. I'm questioning whether there's any point to it on mobile.

songgao: Oh I missed the "on mobile" part. This is just for the desktop. The mobile ones are from `mobileUploadItems` below.

jzila: Oh right. Ugh, I guess we don't have a choice then :(

shared/actions/fs/common.native.js
```diff
@@ -4,22 +4,39 @@ import * as Types from '../../constants/types/fs'
...
 export const saveMedia = ({payload: {path, routePath}}: FsGen.SaveMediaPayload) =>
   Saga.sequentially([
     Saga.put(FsGen.createDownload({intent: 'camera-roll', path})),
-    getTransferPopupAction(path, routePath),
+    getDownloadPopupAction(path, routePath),
   ])
+
+export const pickAndUpload = ({payload: {type}}: FsGen.PickAndUploadPayload) =>
+  new Promise((resolve, reject) =>
+    showImagePicker(
```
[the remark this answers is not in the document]
songgao: Yeah it looks like so: https://github.com/react-community/react-native-image-picker/blob/develop/android/src/main/java/com/imagepicker/ImagePickerModule.java

shared/actions/fs/index.js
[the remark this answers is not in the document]
songgao: Oh actually we already do those; it's in a try/finally block and we reset `polling` to `false` in `finally`.

shared/actions/fs/common.native.js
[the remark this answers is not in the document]
jzila: Maybe it's just Android then: https://github.com/keybase/client/blob/master/shared/actions/fs/platform-specific.android.js#L14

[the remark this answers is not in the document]
songgao: That's the download path, where we use a different thing. The URL above I linked from `react-native-image-picker` is for Android.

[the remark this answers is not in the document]
jzila: Understood, thanks.

shared/fs/container.js
```diff
@@ -39,66 +36,107 @@ const mapStateToProps = (state: TypedState, {path}) => {
...
+      rowType: 'still',
+      path: Types.pathConcat(parentPath, item.name),
+      name: item.name,
+
+      type: item.type,
+      tlfMeta: item.tlfMeta,
+      lastModifiedTimestamp: item.lastModifiedTimestamp,
+    }))
+
+// TODO: when we have renames, reconsile editing rows in here too.
```
jzila: *reconcile

songgao:
bump!

shared/actions/fs/index.js
```diff
@@ -236,17 +279,31 @@ function* pollSyncStatusUntilDone(): Saga.SagaGenerator<any, any> {
...
+        RPCTypes.SimpleFSSimpleFSSyncStatusRpcPromise
+      )
+      yield Saga.put(
+        FsGen.createJournalUpdate({
+          syncingPaths: (syncingPaths || []).map(Types.stringToPath),
+          totalSyncingBytes,
+          endEstimate,
+        })
+      )
+      console.log({
```
chrisnojima: 🔥

shared/fs/container.js
```diff
@@ -39,66 +36,107 @@ const mapStateToProps = (state: TypedState, {path}) => {
...
-    isUserReset: stateProps.isUserReset,
-    resetParticipants: stateProps.resetParticipants,
-    path: stateProps.path,
-    progress: stateProps.progress,
-    routePath,
+    .map(([editID, edit]) => ({
+      rowType: 'editing',
+      editID,
+      name: edit.name,
+
```
chrisnojima: 🔥

shared/fs/container.js
```diff
@@ -1,35 +1,32 @@
...
-    itemDetail &&
-    itemDetail.type === 'folder' &&
-    !!itemDetail.tlfMeta &&
-    itemDetail.tlfMeta.resetParticipants.length > 0
+    itemDetail.type === 'folder' && !!itemDetail.tlfMeta && itemDetail.tlfMeta.resetParticipants.length > 0
       ? itemDetail.tlfMeta.resetParticipants.map(i => i.username)
       : []
   const isUserReset = resetParticipants.includes(_username)
+  const _downloads = state.fs.downloads
+  console.log({msg: 'songgao', path, itemDetail, resetParticipants})
```
chrisnojima: 🔥

shared/fs/container.js
```diff
@@ -39,66 +36,107 @@ const mapStateToProps = (state: TypedState, {path}) => {
...
+  names: Array<string>
+): Array<SortableStillRowItem> =>
+  names
+    .map(name => pathItems.get(Types.pathConcat(parentPath, name), Constants.makeUnknownPathItem({name})))
+    .filter(item => !(item.tlfMeta && item.tlfMeta.isIgnored))
+    .map(item => ({
+      rowType: 'still',
+      path: Types.pathConcat(parentPath, item.name),
+      name: item.name,
+
```
chrisnojima: 🔥

shared/fs/container.js
```diff
@@ -39,66 +36,107 @@ const mapStateToProps = (state: TypedState, {path}) => {
...
+      type: 'folder',
+    }))
+
+const getStillRows = (
+  pathItems: I.Map<Types.Path, Types.PathItem>,
+  parentPath: Types.Path,
+  names: Array<string>
+): Array<SortableStillRowItem> =>
+  names
+    .map(name => pathItems.get(Types.pathConcat(parentPath, name), Constants.makeUnknownPathItem({name})))
```
chrisnojima: maybe simpler to just do reduce and do the filter / transform in one block

shared/fs/files-loading-hoc.js
```diff
@@ -0,0 +1,49 @@
...
+  return {
+    syncingPaths,
+    loadFolderList,
+    loadFavorites,
+    path,
+    routePath,
+  }
+}
+
+const FilesLoadingHoc = compose(
```
chrisnojima: just fyi we prefer to just make a simple wrapper class vs using lifecycle. lifecycle loses a lot of flow typing

shared/fs/footer/upload.desktop.js
```diff
@@ -0,0 +1,46 @@
...
+`
+
+const Upload = ({files, timeLeft}: UploadProps) => (
+  <Box style={stylesBox} className="uploadBoxAnimation">
+    <style>{realCSS}</style>
+    <Text
+      key="files"
+      type="BodySemibold"
+      style={stylesText}
+    >{`Encrypting and uploading ${files} files...`}</Text>
```
chrisnojima: can just use jsx
```
>Encrypting and uploading {files} files...<
```

shared/fs/footer/upload.native.js
```diff
@@ -0,0 +1,78 @@
...
+const stylesText = {
+  color: globalColors.white,
+}
+
+const stylesBox = {
+  ...globalStyles.flexBoxColumn,
+  alignItems: 'center',
+  justifyContent: 'center',
+  height: 48,
+  marginTop: -48,
```
chrisnojima: does this work in android

songgao: Looks like it doesn't :'(

I'm gonna address other feedback and merge this in first, and fix Android in a follow-up PR.

PR #12395, opened 2018-06-15
allow admins to readmit owners after reset

maxtaco:
- a test to check that it works
- needs https://github.com/keybase/keybase/pull/2591 to pass CI


maxtaco:
Agreed. It is just two people.

PR #12398, opened 2018-06-15
Fix starting a new convo from a search with enter/tab

cjb:
@keybase/react-hackers 

exitSearch fires when you hit enter/tab on a searchv3 conversation result, and we want to go to that conversation you found.


cjb:
@chrisnojima Makes sense.  How's this?

PR #12402, opened 2018-06-15
feature flag fs writes

songgao:


shared/util/feature-flags.native.js
```diff
@@ -5,7 +5,8 @@ import type {FeatureFlags} from './feature-flags'
 const ff: FeatureFlags = {
   admin: __DEV__,
   explodingMessagesEnabled: true,
-  fsEnabled: __DEV__,
+  fsEnabled: true,
```
songgao: I'm realizing this wasn't turned on when we made M1 release. But luckily we don't check for this for native (at least now), so it didn't matter.

PR #12404, opened 2018-06-16
add breadcrumb to action popups

songgao:


shared/fs/common/static-breadcrumb.js
```diff
@@ -0,0 +1,87 @@
...
+          type={getIcon(pathElements[1])}
+          color={globalColors.blue}
+          style={stylesIconFolderType}
+          key="icon"
+        />
+      ),
+      <Text key="text" type="BodySmallSemibold">
+        {pathElements[1]}
+      </Text>,
+      ...pathElements.slice(2, includeLast ? undefined : pathElements.length - 1).map((elem, idx) => [
```
jzila: What happens if `pathElements` has length 2 and `includeLast` is false?

shared/fs/common/static-breadcrumb.js
```diff
@@ -0,0 +1,87 @@
...
+          color={globalColors.black_20}
+          fontSize={12}
+        />,
+
+        // TODO: make this into a component.
+        // We are splitting on ',' here, so it won't work for
+        // long names that don't have comma. If this becomes a
+        // problem, we might have to do smarter splitting that
+        // involve other characters, or just break the long name
+        // apart into 3-character groups.
```
jzila: This seems to only handle simple comma-separated TLF names. Not conflicting names, and not subdirectories. And it'll split subdirectories on commas. Is this really the desired behavior?

songgao: As explained in the comment, yes this only deals with comma for now, which is the most common case. We can add other stuff later (or make it fancier) if they become a problem. Do conflicting names not have any space inside? If they have spaces then it's fine. I'm not sure what you mean by subdirectories. It just needs to split long names into chunks so it can wrap properly.

This approach was because we failed to make CSS work for long name wrapping, and Nojima proposed the splitting [here](https://github.com/keybase/client/pull/12314)

shared/fs/common/static-breadcrumb.js
songgao: It'd be an empty slice, so should be fine:

```
%  node
> [1,2].slice(2, 1)
[]
```

songgao:
@jzila mind taking another look at this? THanks!

jzila:
@songgao: want to resolve conflicts first?

PR #12406, opened 2018-06-17
Store raw error for EphemeralUnboxingError

joshblum:
cc @chrisnojima if you want to build off this for verbose debugging


joshblum:
was debugging something with chris and he didn't have full logs (was running the service in debug mode when it happened) so i thought it would be useful to store when it happens

mmaxim:
The service always runs in debug mode

joshblum:
I’m not sure his setup that was doing something to the logs

PR #12407, opened 2018-06-17
fs tweaks

songgao:
1. Show ... menu for folders too, so we can see metadata;
2. Collapse 'Share to other apps' into main popup menu;
3. When time remaining for transfers is not known, just don't show any
thing, ratehr than show '? s';
4. When we don't know the modified timestamp (e.g. for
public/private/team folders in root, or for a TLF in a TLF list), just
don't show a time, rather than showing the epoch;
5. Make clickable area larger for row action ... button and close button in bare view;
6. Fix download badge position on mobile.

### DIFF
--- a/shared/actions/fs/common.native.js
+++ b/shared/actions/fs/common.native.js
@@ -1,33 +1,25 @@
 // @flow
+import * as I from 'immutable'
+import * as Types from '../../constants/types/fs'
 import * as Saga from '../../util/saga'
 import * as FsGen from '../fs-gen'
 import {putActionIfOnPath, navigateAppend} from '../route-tree'
 
-export const share = (action: FsGen.SharePayload) =>
+const getTransferPopupAction = (path: Types.Path, routePath?: I.List<string>) =>
   Saga.put(
-    putActionIfOnPath(
-      action.payload.routePath,
-      navigateAppend([
-        {
-          props: {path: action.payload.path, isShare: true},
-          selected: 'pathItemAction',
-        },
-      ])
-    )
+    routePath
+      ? putActionIfOnPath(routePath, navigateAppend([{props: {path}, selected: 'transferPopup'}]))
+      : navigateAppend([{props: {path}, selected: 'transferPopup'}])
   )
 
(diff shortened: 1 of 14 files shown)


songgao:
It’s probably fine, but I’m ok with not including it in the release too. @jzila WDYT?

chrisnojima:
ok lets defer

shared/fs/popups/row-action-popup-container.js
```diff
@@ -123,11 +123,17 @@ const getRootMenuItems = (stateProps, dispatchProps) => {
       onClick: () => saveImage(path),
     })
   isMobile &&
+    pathItem.type === 'file' &&
+    // TODO: re-enable this (or refactor) when we have in-app share. This
+    // should probably be split up into separate components anyway.
```
jzila: I thought we decided to include in-app share items in the root menu:
* Share to other app
* Share to chat conversation

songgao: Oh; I wasn't aware of that (or I didn't remember). Will clarify in the TODO comment.

jzila:
As @chrisnojima says, let's wait. We're already going through the release checklist.

songgao:
Please take a look at last two commits ("remove second level ... " and "fix download badge ..."). Thanks!

shared/actions/fs/common.native.js
```diff
@@ -1,33 +1,25 @@
...
-}
+export const shareNative = ({payload: {path, routePath}}: FsGen.ShareNativePayload) =>
+  Saga.sequentially([
+    Saga.put(FsGen.createDownload({intent: 'share', path})),
+    getTransferPopupAction(path, routePath),
+  ])
+
+export const saveMedia = ({payload: {path, routePath}}: FsGen.SaveMediaPayload) =>
+  Saga.sequentially([
+    Saga.put(FsGen.createDownload({path, intent: 'camera-roll'})),
```
jzila: please keep `intent` and `path` in the same order for the various `FsGen.createDownload` calls.

PR #12416, opened 2018-06-18
Some exploding style cleanup

buoyad:


shared/common-adapters/input.js.flow
```diff
@@ -87,5 +87,7 @@ declare export default class Input extends Component<Props> {
   selection: () => Selection;
   // transformText must be called only on uncontrolled Input
   // components. The transformation may be done asynchronously.
-  transformText: (fn: (TextInfo) => TextInfo) => void;
+  // @param reflectChange: desktop only. If true, `onChangeText`
```
buoyad: opt-in so it only affects this spot

buoyad:
@keybase/react-hackers the animation colors aren't finalized, but o/w this is ready!

shared/chat/conversation/messages/wrapper/exploding-meta.js
```diff
@@ -207,11 +209,44 @@ const styles = styleSheetCreate({
...
     },
   }),
-  countdownContainer: {
-    borderRadius: 2,
-    paddingLeft: 4,
-    paddingRight: 4,
-  },
+  countdown: platformStyles({
+    common: {color: globalColors.white, fontSize: 10, fontWeight: 'bold'},
+    isElectron: {lineHeight: '1.4em'},
```
chrisnojima: youcan just put lineHeight in common and it'll do teh right thign in platform styles

PR #12422, opened 2018-06-18
fs: Fix zoomable images on native

jzila:


jzila:
Jenkins passed...merging without CircleCI.

PR #12425, opened 2018-06-18
fix a bug where we weren't respecting symmetric key context

oconnor663:
This was causing different applications to derive the same symmetric
encryption key. Currently only the stellar code is affected, which isn't
yet publicly launched. We believe we can replace the bad boxes in the
employee accounts that have issued them, as part of switching this code
over to a proper context string.

r? @mlsteele


oconnor663:
I checked for all the callers by changing the name of the function and building all tests. Only stellar code broke.

oconnor663:
Added task ID of https://keybase.atlassian.net/browse/CORE-8135 in the TODO comments.

PR #12429, opened 2018-06-18
keybase1: simpler edit notification types

strib:
These will be used for the new edit history protocol.

(Circle will likely fail until I merge the corresponding KBFS PR.)

### DIFF
--- a/protocol/avdl/keybase1/kbfs_common.avdl
+++ b/protocol/avdl/keybase1/kbfs_common.avdl
@@ -51,11 +51,28 @@ protocol kbfsCommon {
     UID writerUid;
     Time localTime;
     FolderType folderType;
-}
+  }
 
   record FSEditListRequest {
     Folder folder;
     int requestID;
+  }
+
+  record FSFolderWriterEdit {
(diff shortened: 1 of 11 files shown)


protocol/avdl/keybase1/kbfs_common.avdl
```diff
@@ -51,13 +51,30 @@ protocol kbfsCommon {
...
+
+  record FSFolderWriterEditHistory {
+    string writerName;
+    array<FSFolderWriterEdit> edits;
+  }
+
+  record FSFolderEditHistory {
+    Folder folder;
+    Time serverTime;
+    array<FSFolderWriterEditHistory> history;
```
songgao: Is there a reason why this isn't a map? It probably won't make much a difference and I'm just curious.

strib: > Is there a reason why this isn't a map? It probably won't make much a difference and I'm just curious.

Sorry, I forgot to answer this.  It's not a map because it needs to be ordered, with the most recent edits first.

PR #12439, opened 2018-06-19
Add PaymentStatusMsg handler to stellargregor

zapu:
It does not do anything with the message yet. This PR does, however, introduce a new record type: `stellar1.PaymentStatusMsg` that will have to be revendored to server PR.

### DIFF
--- a/go/stellar/stellargregor/stellargregor.go
+++ b/go/stellar/stellargregor/stellargregor.go
@@ -62,6 +62,7 @@ func (h *Handler) autoClaim(mctx libkb.MetaContext, cli gregor1.IncomingInterfac
 }
 
 func (h *Handler) paymentStatus(mctx libkb.MetaContext, cli gregor1.IncomingInterface, category string, item gregor.Item) error {
+	defer h.G().GregorDismisser.DismissItem(mctx.Ctx(), cli, item.Metadata().MsgID())
 	mctx.CDebugf("%v: %v received", h.Name(), category)
 	var msg stellar1.PaymentStatusMsg
 	if err := json.Unmarshal(item.Body().Bytes(), &msg); err != nil {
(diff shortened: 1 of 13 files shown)


go/stellar/stellargregor/stellargregor.go
```diff
@@ -56,3 +60,17 @@ func (h *Handler) autoClaim(mctx libkb.MetaContext, cli gregor1.IncomingInterfac
 	mctx.G().GetStellar().KickAutoClaimRunner(mctx, item.Metadata().MsgID())
 	return nil
 }
+
+func (h *Handler) paymentStatus(mctx libkb.MetaContext, cli gregor1.IncomingInterface, category string, item gregor.Item) error {
```
mlsteele: at least initially while the format may still be in flux, it might be good to dismiss these even if they fail to unmarshal

zapu: done, thanks!

protocol/avdl/stellar1/remote.avdl
```diff
@@ -166,4 +166,16 @@ protocol remote {
...
+  record PaymentStatusMsg {
+    KeybaseTransactionID keybaseID;
+    TransactionID stellarID;
+    boolean pending;
+    boolean relay; // Is it a relay payment?
+    // After relay payment was submitted, was there already a
+    // claimant user?
+    boolean autoClaimed;
+    union { null, string } errMsg;
+  }
```
mlsteele: Move to stellar1/gregor.avdl to just to stay organized.

mlsteele: This is a sprawling combination of fields for a notification that we don't know yet how it will be used. Seems simpler to either just send the IDs so that the client can call up to refresh, or send down a full Summary type. For now I'd recommend just sending kbTxID and txID (rename them). We can add more later if needed. So the message just means 'something has changed'.

protocol/avdl/stellar1/gregor.avdl
```diff
@@ -0,0 +1,11 @@
...
+protocol gregor {
+  import idl "github.com/keybase/client/go/protocol/keybase1" as keybase1;
+  import idl "common.avdl";
+
+  // stellar.payment_status gregor message body
+  record PaymentStatusMsg {
+    KeybaseTransactionID kbTxID;
+    TransactionID txID;
+  }
+}
```
zapu: @mlsteele is this one fine?

PR #12440, opened 2018-06-19
dont count all placeholder results as hit from PullLocalOnly

mmaxim:
We don't get much value from returning 10 placeholder messages from `PullLocalOnly` to the UI, so just count those as cache misses. 

cc @chrisnojima

### DIFF
--- a/go/chat/convsource.go
+++ b/go/chat/convsource.go
@@ -783,9 +783,27 @@ func (p *pullLocalResultCollector) String() string {
 	return fmt.Sprintf("[ %s: t: %d ]", p.Name(), p.num)
 }
 
+func (p *pullLocalResultCollector) realHits() (total int) {
+	for _, m := range p.Result() {
+		st, err := m.State()
+		if err != nil {
+			// count these
+			total++
+			continue
+		}
+		switch st {
+		case chat1.MessageUnboxedState_PLACEHOLDER:
+			// don't count!
+		default:
+			total++
+		}
+	}
+	return total
+}
+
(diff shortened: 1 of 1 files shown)


go/chat/convsource.go
```diff
@@ -783,9 +783,27 @@ func (p *pullLocalResultCollector) String() string {
 	return fmt.Sprintf("[ %s: t: %d ]", p.Name(), p.num)
 }
 
+func (p *pullLocalResultCollector) realHits() (total int) {
+	for _, m := range p.Result() {
+		st, err := m.State()
+		if err != nil {
```
joshblum: could be `if err != nil || st != chat1. MessageUnboxedState_PLACEHOLDER`

go/chat/convsource.go
```diff
@@ -783,9 +783,27 @@ func (p *pullLocalResultCollector) String() string {
 	return fmt.Sprintf("[ %s: t: %d ]", p.Name(), p.num)
 }
 
+func (p *pullLocalResultCollector) realHits() (total int) {
```
joshblum: maybe add an `minimum int` and break if we hit that total so we don't loop across all results?

joshblum: or make return a `bool` for if we hit anything real

go/chat/convsource.go
mmaxim: You know how much I like `switch` :)

go/chat/convsource.go
mmaxim: Hmm, yeah this makes sense, will do!

PR #12445, opened 2018-06-19
fix crash backing out of subteams

chrisnojima:
@keybase/react-hackers @cjb


shared/common-adapters/list.desktop.js
```diff
@@ -8,6 +8,11 @@ import type {Props} from './list'
 class List extends PureComponent<Props<any>, void> {
   _list: ?ReactList
   _itemRender = index => {
+    // ReactList has an issue where it caches the list length into its own state so can ask
+    // for indicies outside of the items...
```
cjb: indices

PR #12447, opened 2018-06-19
Whitelist 'keybase fs ls' colored output

AMarcedone:


AMarcedone:
@patrickxb I added two more quick changes:
- `keybase chat read` now always shows the usescaped output (as cat would do). I discussed this with max.
- the names output by `keybase fs` ls are now directly escaped (before I just whitelisted the whole output, but it turns out filenames might contain escape codes).

go/terminalescaper/escaper.go
```diff
@@ -2,24 +2,114 @@ package terminalescaper
...
+		return -2
 	}, s)
 }
 
+// replace returns a copy of the string s with all its characters modified
+// according to the mapping function.
+// If mapping returns -1, the character is substituted with the two character string `^[`.
+// If mapping returns any other negative value, the character is dropped from the string with no replacement.
+// This function is similar to strings.Map, except for how -1 is handled.
+func replace(mapping func(rune) rune, s string) string {
```
patrickxb: why can't you use strings.Map?

AMarcedone: strings.Map was the previous solution. The point is that strings.Map only allows to map one rune to one rune. I want to substitute `0x1b` (one rune) with `^[` (two runes), so I had to slightly modify strings.Map for this purpose. 

Another option would have been using strings.Map to escape everything else but `0x1b`, and then take care of `0x1b` -> `^[` with strings.Replace, but this would require two passes over the data instead of a single one.

patrickxb: Makes me nervous to not just use strings.Map (which is battle tested by thousands of users) as opposed to a custom function.

AMarcedone: So do you think I should go with the second option and do two passes over the data?

For what is worth, I copied over all the tests which are given for strings.Map to check that this implementation passes them as well. The changes (w.r.t. strings.Map) are quite small. I don't have a strong preference here...

maxtaco: Can `strings.Replace` just work with "0x1b" -> "^[", in one pass?

maxtaco: OK, got it!

maxtaco: OK, i think it's fine to give this a shot.  I worry that if we make two passes over the data, we're slowing down the terminal in a way that we can't easily detect, but might be perceivable.  This is basically `strings.Map`, and I realize we're forking it, but it probably changes very slowly at this point.

patrickxb: In regards to two passes, aren't you doing two passes now?  line 39 and line 78.

patrickxb: nvm, i see...

patrickxb: I'm ok with merging, just worried that we will miss upstream bugfixes by forking the function out.

Can you add a comment that this is copied from strings.Map function and a comment on where it is different?  So we can maintain it easier in the future.

AMarcedone: Comments are added. I will merge after CI passes unless I hear any objections.

Thanks a lot for the review!

PR #12453, opened 2018-06-20
in case itemDetails is null

songgao:


songgao:
Yeah; hopefully it’ll happen again and we can figure out why.

PR #12454, opened 2018-06-20
Change reset member badge handler to use User loader instead of UIDMap

zapu:
1) Pass `Uid` from `TeamMemberOutFromReset` message to `TeamMemberOutReset` (needed small JS change).
2) Add `findAndDismissResetBadges` function that will load the team, iterate through reset member badges, and get latest UV for each badge's Uid that's in the team using `Team.UserVersionByUID`:
    a) If not found, that means reset member has been removed - dismiss the badge.
    b) If found, load user and check if team has the current version - if so, dismiss the badge.
3) Wrap call to this function in goroutine, so it's done in background similar to team key rotation. Use the same lock as team rotation so only one of either can be running at time.
4) Remove `clearServerUIDMapCache` call from test, UIDMapper should not be used anymore during any of this so we are immune to cache races.
5) Bonus change: factor out unused `forceRepoll` from some of team.Details loading code, but leave the API as is for now. `forceRepoll` was obsoleted when UIDMapper was added, but the API was never changes. So there's code that consciously passes either true or false for `forceRepoll` but sadly this doesn't matter.


go/service/team_handler.go
```diff
@@ -151,6 +155,50 @@ func (r *teamHandler) abandonTeam(ctx context.Context, cli gregor1.IncomingInter
...
+	if err != nil {
+		return err
+	}
+
+	for _, badge := range badges {
+		var dismiss bool
+		teamUV, notFoundErr := team.UserVersionByUID(ctx, badge.Uid)
+		if notFoundErr == nil {
+			loadUserArg := libkb.NewLoadUserByUIDArg(ctx, r.G(), badge.Uid).WithPublicKeyOptional()
+			u, err := libkb.LoadUser(loadUserArg)
```
maxtaco: Use the UPAK loader here.

maxtaco: You might need the forceRepoll flag here.

zapu: done, thanks!

go/teams/service_helper.go
```diff
@@ -29,29 +29,36 @@ func LoadTeamPlusApplicationKeys(ctx context.Context, g *libkb.GlobalContext, id
...
-	ret.Readers, err = userVersionsToDetails(ctx, g, m.Readers, forceRepoll)
+	ret.Readers, err = userVersionsToDetails(ctx, g, m.Readers)
 	if err != nil {
 		return ret, err
 	}
 	return ret, nil
 }
 
 func Details(ctx context.Context, g *libkb.GlobalContext, name string, forceRepoll bool) (res keybase1.TeamDetails, err error) {
+	// TODO: forceRepoll is unused since 87b4ec1d699b633a957156108276df154f2fbc2a
```
maxtaco: Can you finish out this todo by removing the `forceRepoll` flag and then promoting this comment to a header comment for the `Details` function?

zapu: Sure, will do! I've already done that but I reverted to keep this PR focused on one thing.

PR #12456, opened 2018-06-20
always use a memory SecretStore, sometimes a disk

maxtaco:
- we've seen certain cases of the macOS keychain just not working; it's a bummer, but the least we can do is
  to only pester the user for a passphrase once every service restart. Previously, we were going to the secret
  store every time we needed the password.
- this PR introduces a SecretStoreMem as a write-through cache to a disk-backed SecretStore, which is optional

### DIFF
--- a/go/libkb/secret_store.go
+++ b/go/libkb/secret_store.go
@@ -130,67 +130,112 @@ func ClearStoredSecret(g *GlobalContext, username NormalizedUsername) error {
 	return ss.ClearSecret(username)
 }
 
-// SecretStoreLocked protects a SecretStoreAll with a mutex.
+// SecretStoreLocked protects a SecretStoreAll with a mutex. It wraps two different
+// SecretStoreAlls: one in memory and one in disk. In all cases, we always have a memory
+// backing. If the OS and options provide one, we can additionally have a disk-backed
+// secret store. It's a write-through cache, so on RetrieveSecret, the memory store
+// will be checked first, and then the disk store.
 type SecretStoreLocked struct {
-	SecretStoreAll
+	Contextified
 	sync.Mutex
+	mem  SecretStoreAll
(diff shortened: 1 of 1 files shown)


maxtaco:
Thanks for the review. Do you know why we haven't made this change previously? Seems like also a win for android since secret store access is so slow there

patrickxb:
there is something there already, i believe that caches retrieves, exactly for that purpose.  let me find it.

patrickxb:
it's for the temporary usage ones:

// NewSecretStore returns a SecretStore interface that is only used for                   
// a short period of time (i.e. one function block).  Multiple calls to RetrieveSecret()  
// will only call the underlying store.RetrieveSecret once.                               
func NewSecretStore(g *GlobalContext, username NormalizedUsername) SecretStore {          
        store := g.SecretStore()                                                          
        if store != nil {                                                                 
                return &SecretStoreImp{                                                   
                        username: username,                                               
                        store:    store,                                                  
                }                                                                         
        }                                                                                 
        return nil                                                                        
}

maxtaco:
Ah, so my PR does nothing?

patrickxb:
I'd have to look through the code, but I think it still does something.

maxtaco:
I will look too.  This codepath feels overly complex by two abstraction
layers.

patrickxb:
Yes, it's a mess.  And the names don't help.

maxtaco:
The problem with the cache explained above is that we're constantly making new `SecretStoreImp`s, say across two different calls to `BootstrapActiveDevice`. So the secret will be refetched from the underlying store without this change.

maxtaco:
I'll put in some comments in master.

maxtaco:
See 35cfdfd5497cf2c55aa61bb3561843cf7c947780

PR #12460, opened 2018-06-20
Exploding messages QA 2

buoyad:
- [x] Focus chat input on selecting explode time 
  - Previously, trying to type automatically focused the input. This makes the cursor show up.
- [x] Text colors on exploding popup
- [x] Usernames colored correctly & clickable in message popup
- [x] Add `clickToProfile` prop to `Avatar` to let it handle setting up click handlers

r? @keybase/react-hackers


buoyad:
@chrisnojima it updated all of the message popup storyshots 👍

buoyad:
Nothing else changed, just those 5

PR #12464, opened 2018-06-20
Don't show chat warning for users you don't follow

buoyad:
r? @keybase/react-hackers


shared/chat/conversation/bottom-banner/container.js
```diff
@@ -51,9 +53,10 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
     const broken = stateProps._meta.participants.filter(p =>
       stateProps._users.infoMap.getIn([p, 'broken'], false)
```
chrisnojima: why not just add the check here

buoyad:
@chrisnojima thanks, updated to do it all in one pass. PTAL

PR #12469, opened 2018-06-20
Wallet key derivation fix phase 1

mlsteele:
Wallet encryption was using the wrong context string to derive its key. This PR adds support for wallet crypt v2 which uses the correct context string.

This is phase 1. Add support for crypt v2 and drop support for crypt v1. There's a new unlisted command `wallet fixup` which is the only thing that still supports decrypting v1. Fixup decrypts v1 and encrypts and uploads v2. For existing accounts with crypt v1 bundles, This PR breaks all stellar commands until fixup is run.

Before phase 2 I will get everyone with a wallet to run `keybase wallet fixup`.

All throughout I'm referring to the encryption version which is moving to v2. Which is different from the data version which is staying at v1.

```
$ kbu wallet history
▶ ERROR stellar secret bundle encryption version 1 has been retired. Try `keybase wallet fixup`

$ kbu wallet fixup
Wallet bundle was busted.. but now it is fixed
✔ Success

$ kbu wallet fixup
Wallet bundle was already fixed
✔ Success
```

Phase 2, not in this PR, will be to remove all support for v1 and remove the `wallet fixup` command. After phase 2 anyone with a crypt v1 bundle as their latest would be marooned forever, but there will be no such users.

This will not pass CI until https://github.com/keybase/keybase/pull/2607 happens.


go/stellar/bundle/boxer.go
```diff
@@ -171,17 +176,34 @@ func Unbox(decodeRes DecodeResult, visibleBundleB64 string,
...
+	default:
+		return res, fmt.Errorf("unsupported stellar secret bundle encryption version: %v", encBundle.V)
+	}
+
 	// Derive key
-	// TODO: We're passing the wrong context in here. We need to migrate to libkb.DeriveReasonPUKStellarBundle. (CORE-8135)
-	symmetricKey, err := puk.DeriveSymmetricKey(libkb.DeriveReasonPUKPrev)
+	reason := libkb.DeriveReasonPUKStellarBundle
+	if encBundle.V == 1 {
+		reason = libkb.DeriveReasonPUKPrev
```
oconnor663: Maybe another comment here just to say what's up?

mlsteele: This code will be gone quicker than you can say https://github.com/keybase/client/pull/12496

PR #12475, opened 2018-06-21
More font-size tweaks

cecileboucheron:
@keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/set-explode-popup/index.desktop.js
+++ b/shared/chat/conversation/messages/set-explode-popup/index.desktop.js
@@ -20,13 +20,13 @@ const Header = (props: HeaderProps) => (
           style={platformStyles({isElectron: {display: 'block', lineHeight: '28px', maxHeight: 28}})}
         />
         <Text type="BodySmallSemibold" backgroundMode="Announcements" style={{textAlign: 'center'}}>
-          Set a timeout on your messages and watch them E&nbsp;X&nbsp;P&nbsp;L&nbsp;O&nbsp;D&nbsp;E&nbsp;!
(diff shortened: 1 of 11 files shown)


shared/chat/conversation/messages/set-explode-popup/index.desktop.js
```diff
@@ -19,17 +19,14 @@ const Header = (props: HeaderProps) => (
           fontSize={48}
           style={platformStyles({isElectron: {display: 'block', lineHeight: '28px', maxHeight: 28}})}
         />
-        <Text type="BodySemibold" backgroundMode="Announcements" style={{textAlign: 'center'}}>
+        <Text type="BodySmallSemibold" backgroundMode="Announcements" style={{textAlign: 'center'}}>
           Set a timeout on your messages and watch them
-        </Text>
-        <Text type="BodySemibold" backgroundMode="Announcements">
-          E X P L O D E
+          E&nbsp;&nbsp;&nbsp;X&nbsp;&nbsp;&nbsp;P&nbsp;&nbsp;&nbsp;L&nbsp;&nbsp;&nbsp;O&nbsp;&nbsp;&nbsp;D&nbsp;&nbsp;&nbsp;E.
```
adamjspooner: Is there a reason we're not using CSS to style this?

cecileboucheron: Sure I can do CSS if that looks messy.

chrisnojima: might as well leave it, it'll likely not work well on android , this is simpler

PR #12478, opened 2018-06-21
inject in order so we don't go crazy

mmaxim:
If `Constants.newExplodingGregorKey` makes it into the state first, then it sets off a chain reaction later on in `handleIsExplodingNew`. It triggers this code:

```
else if (!seenExploding) {
    // newExploding && !seenExploding. this should never happen
    logger.warn('Got newExploding but not seenExploding! Setting seenExploding...')
    actions.push(
      Saga.call(RPCTypes.gregorInjectItemRpcPromise, {
        cat: ChatConstants.seenExplodingGregorKey,
        body: 'true',
        dtime: {time: 0, offset: 0},
      })
    )
  }
```

Can we just delete that code block too?

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -2087,8 +2087,8 @@ function* handleSeeingExplodingMessages(action: Chat2Gen.HandleSeeingExplodingMe
   }
   // neither are set, inject both
   yield Saga.sequentially([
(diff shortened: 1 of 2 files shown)


buoyad:
We should change it so the UI uses one flag and compares the `ctime`

PR #12479, opened 2018-06-21
Skip TestTeamResetBadges* tests

zapu:
so there are two bugs:
- one related to new stellar notifications stellar.account_change that somehow clear badge state so the new reset user badge is never observed by the test
- second one related to my refactor, somehow it gets old upak (??) and does not dismiss the badge when it should


maxtaco:
Can merge without ci

PR #12482, opened 2018-06-21
Rotate Desktop Images based on EXIF Orientation Flag

thebearjew:
⚠️ WIP ⚠️ 

### Problem

Chrome does not, and historically has not, ever respected EXIF orientation flags. There has been ongoing tickets in the Chromium issue tracker to add support of EXIF orientation ([#1](https://bugs.chromium.org/p/chromium/issues/detail?id=56845) [#2](https://bugs.chromium.org/p/chromium/issues/detail?id=158753)).

This manifests when viewing images in Chrome that have an existing orientation but the EXIF Orientation flag dictates a different one. See the following resources for EXIF Orientation information:
- https://magnushoff.com/jpeg-orientation.html
- https://beradrian.wordpress.com/2008/11/14/rotate-exif-images/

### Behavior

Images rendered directly in chrome (not transcoded by us) will not respect the EXIF orientation. This can be tested:
1. Choose an image with an EXIF orientation flag.
2. Upload to a Keybase chat
3. The upload image preview will not respect the orientation - this is expected and can't be changed on our end at the moment.
3. After the image successfully uploads, the attachment thumbnail in chat **should correctly respect the orientation**
4. Clicking on the image to view in full screen **will not correctly respect the orientation** and render the image without the flag.

Example images to use for testing in Keybase application.
- [Orientation 1](https://user-images.githubusercontent.com/5200812/41726449-2e76fb60-7540-11e8-8fcc-92dfdc991ec0.jpeg)
- [Orientation 2](https://user-images.githubusercontent.com/5200812/41726468-38b009d2-7540-11e8-9cb0-259f7cf92464.jpeg)
- [Orientation 3](https://user-images.githubusercontent.com/5200812/41726489-4572a8f0-7540-11e8-9c62-50b3292b8248.jpeg)
- [Orientation 4](https://user-images.githubusercontent.com/5200812/41726503-50a96808-7540-11e8-9229-4f6e8f0be904.jpeg)
- [Orientation 5](https://user-images.githubusercontent.com/5200812/41726558-6c2da6ac-7540-11e8-8741-d8ed582d70ef.jpeg)
- [Orientation 6](https://user-images.githubusercontent.com/5200812/41726580-76ba44b8-7540-11e8-8c1b-5d20ded634eb.jpeg)
- [Orientation 7](https://user-images.githubusercontent.com/5200812/41726593-7e7956d0-7540-11e8-881c-70a4fa5762f0.jpeg)
- [Orientation 8](https://user-images.githubusercontent.com/5200812/41726619-89bb4a1c-7540-11e8-86f7-afdf32ac20a1.jpeg)

### Solution 

- Added [exif-js](https://github.com/exif-js/exif-js) as a prod dependency
- Created a new common adapter for handled exif orientations on desktop `OrientedImage`
- Currently, the only location that can be reoriented successfully is `chat/conversation/attachment-full-screen`.

### Notes

- The electron Content Security Policy was set to only accept `connect-src` from `localhost:4000` on dev and `self` in production. The CSP has been updated to allow exif.js to make HTTP requests to the local Keybase content server started by the client on `127.0.0.1:*`.
- EXIF Orientation is respected on both native platforms.


shared/common-adapters/oriented-image.desktop.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+  componentDidMount() {
+    const component = this
+    EXIF.getData({src: this.props.src}, function() {
+      const img = this
+      const orientation: number = EXIF.getTag(img, 'Orientation')
+      if (orientation && inRange(orientation, 1, 9)) {
+        const transform: StylesDesktop = makeStyleTransform(orientation)
+        component.setState({styleTransform: transform})
```
thebearjew: Not entirely happy with calling `setState()` inside of `componentDidMount` -> I've always understood this to be an anti-patter and possibly leading to two renders.

Seeing as we only want to perform this once, the other thought I had was to move the EXIF library call to a parent container and hold off the loading state until it's completed.

Looking for feedback on this.

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -0,0 +1,71 @@
...
+import {Image} from '../common-adapters'
+import {type Props} from './oriented-image.types'
+import {collapseStyles, type StylesDesktop} from '../styles'
+
+type State = {
+  styleTransform: StylesDesktop | null,
+}
+
+const makeStyleTransform = (orientation: number): StylesDesktop =>
+  exifOrientaionMap[orientation] || {transform: 'rotate(0deg)'}
```
chrisnojima: we might want to log if we get a different orientation and fallback to 0 so we can debug user logs

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -0,0 +1,71 @@
...
+
+// Orientations:
+// 1: rotate 0 deg left
+// 2: flip horizontally
+// 3: rotate 180 deg left
+// 4: flip vertivally
```
chrisnojima: vertically

shared/common-adapters/oriented-image.desktop.js
chrisnojima: lets chat about this

thebearjew: ^ chatted

shared/common-adapters/oriented-image.desktop.js
thebearjew: What log level would be appropriate? warn?

thebearjew:
@keybase/react-hackers ready

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -0,0 +1,125 @@
...
+  return transform
+}
+
+class OrientedImage extends React.Component<Props, State> {
+  static defaultProps = {
+    onLoad: noop,
+  }
+
+  state = {
+    cacheStyleTransforms: {},
```
chrisnojima: this can be module level so its not lost when this component gets dismounted

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -0,0 +1,125 @@
...
+        return this.setState({
+          styleTransform: this.state.cacheStyleTransforms[src],
+        })
+      }
+
+      const component = this
+      // EXIF will make an HTTP request locally to 127.0.0.1:* to fetch the
+      // image that the keybase service is serving
+      // img = this refers to the image ArrayBuffer fetched from the local server.
+      EXIF.getData({src}, function() {
```
chrisnojima: its annoying that getData() works this way but I guess we can't do much unless we fork it. Instead lets just limit the scope of this context so maybe

```
_handleData = (img) {
...
}

_setTranformForExifOrientation
...
const handleData = this._handleData
EXIT.getData({src}, function() {  handleData(this) })
```

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -0,0 +1,125 @@
...
+
+  state = {
+    cacheStyleTransforms: {},
+    styleTransform: {transform: ''},
+  }
+
+  _hasComponentMounted = false
+
+  _setTranformForExifOrientation(src) {
+    if (this._hasComponentMounted) {
```
chrisnojima: likely better to just check and bail early instead of having this whole thing nested

shared/common-adapters/oriented-image.desktop.js
chrisnojima: this._hasComponentMounted has to be checked here. The danger is any async call has to ensure the component is mounted, so the component.setState is whats unsafe here

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -0,0 +1,125 @@
...
+      EXIF.getData({src}, function() {
+        const img = this
+        const orientation: number = EXIF.getTag(img, 'Orientation')
+        // EXIF.getTag can return undefined for Orientation if the field is not found
+        if (isNumber(orientation)) {
+          const styleTransform: Transform = makeStyleTransform(orientation)
+          component.setState(prevState => {
+            const newTransform =
+              prevState.styleTransform.transform !== styleTransform.transform
+                ? styleTransform
```
chrisnojima: the semantics of setState with the function parameter is such that if you don't want to update you shouldn't put the value in, an equal value will cause downstream stuff to trigger

so something like

component.setState(p => p.styleTransform.transform === styleTransform.transform ? undefined : {styleTransform: newTransform})

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -0,0 +1,125 @@
...
+}
+
+class OrientedImage extends React.Component<Props, State> {
+  static defaultProps = {
+    onLoad: noop,
+  }
+
+  state = {
+    cacheStyleTransforms: {},
+    styleTransform: {transform: ''},
```
chrisnojima: should should just be a string and not the object with just the key. we usually want a flat value for state and not an hierarchical object

chrisnojima: especially if this isn't immutable you can run into rendering problems. we can discuss in person

shared/common-adapters/oriented-image.desktop.js
thebearjew: Thanks for clarifying this, seems like I misunderstood earlier!

shared/common-adapters/oriented-image.desktop.js
thebearjew: That's really good to know, thank you!

shared/common-adapters/oriented-image.desktop.js
thebearjew: Major slip up on the purpose of a cache on my part >.<

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -51,43 +46,39 @@ class OrientedImage extends React.Component<Props, State> {
   }
 
   state = {
-    cacheStyleTransforms: {},
-    styleTransform: {transform: ''},
+    styleTransform: '',
   }
 
+  _cacheStyleTransforms = {}
```
chrisnojima: this actually should go up and be a module level cache, it can't exist in the insance

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -51,43 +46,39 @@ class OrientedImage extends React.Component<Props, State> {
...
-              prevState.styleTransform.transform !== styleTransform.transform
-                ? styleTransform
-                : prevState.styleTransform
-            return {
-              cacheStyleTransforms: {[src]: newTransform},
-              styleTransform: newTransform,
-            }
-          })
-        }
+    if (this._cacheStyleTransforms[src]) {
```
chrisnojima: if you have this you can skip all this and just initialize your state with it in your constructor

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  _hasComponentMounted = false
+
+  _handleData = img => {
+    const orientation: number = EXIF.getTag(img, 'Orientation')
+    // EXIF.getTag can return undefined for Orientation if the field is not found
+    if (isNumber(orientation)) {
+      const newTransform: string = makeStyleTransform(orientation)
+      this.setState(p => (p.styleTransform === newTransform ? undefined : {styleTransform: newTransform}))
```
chrisnojima: i think you're not ever using your cache. you're never writing to it which likely needs to happen here. i'd add some logging so we can validate that this is working as expected. i'd log cache hits, cache misses, etc

thebearjew:
Rebased on master

chrisnojima:
- [ ] doesn't seem to correct itself on upload

chrisnojima:
screen shot 2018-06-22 at 12 03 19 pm

thebearjew:
>screen shot 2018-06-22 at 12 03 19 pm ...

@chrisnojima Those are meant to be oriented. All images images are originally upright (like the last one) and are then set to different orientations based on EXIF.

It would not be working if you saw three normal images with these as examples.

thebearjew:
>>screen shot 2018-06-22 at 12 03 19 pm ... ...

@thebearjew missing EXIF orientation can be seen in the upload preview (`attachment-get-titles`) because they all appear to be @ Orientation 1

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -53,30 +56,33 @@ class OrientedImage extends React.Component<Props, State> {
...
+      return this.setState({styleTransform: _cacheStyleTransforms[src]})
     }
 
     // EXIF will make an HTTP request locally to 127.0.0.1:* to fetch the
     // image that the keybase service is serving
     // img = this refers to the image ArrayBuffer fetched from the local server.
     const handleData = this._handleData
+    const _hasComponentMounted = this._hasComponentMounted
     EXIF.getData({src}, function() {
+      if (!_hasComponentMounted) return
```
thebearjew: Update with correct `this`

shared/common-adapters/oriented-image.native.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import * as React from 'react'
+import {type Props} from './oriented-image.types.js.flow'
```
chrisnojima: we never explicitly point to the js.flow files, just use oriented-image.types

PR #12483, opened 2018-06-21
Engine stats

chrisnojima:
- [ ] ~add to log send~ (turns out this isn't easy to plumb right now)
- [x] whitelist for kb

### DIFF
--- a/shared/actions/tracker.js
+++ b/shared/actions/tracker.js
@@ -618,11 +618,11 @@ function _setupTrackerHandlers() {
         }
       }
 
-      const session: Session = engine().createSession(
-        _serverCallMap(dispatch, getState, onStart, onFinish),
-        null,
-        cancelHandler
-      )
+      const session: Session = engine().createSession({
+        cancelHandler,
+        incomingCallMap: _serverCallMap(dispatch, getState, onStart, onFinish),
+        startMethod: 'keybase.1.identifyUi.delegateIdentifyUI',
+      })
 
       response && response.result(session.getId())
     }
--- a/shared/actions/tracker.js
+++ b/shared/actions/tracker.js
@@ -621,7 +621,6 @@ function _setupTrackerHandlers() {
       const session: Session = engine().createSession({
         cancelHandler,
         incomingCallMap: _serverCallMap(dispatch, getState, onStart, onFinish),
-        startMethod: 'keybase.1.identifyUi.delegateIdentifyUI',
       })
 
       response && response.result(session.getId())
--- a/shared/actions/unlock-folders.js
+++ b/shared/actions/unlock-folders.js
@@ -72,7 +72,6 @@ function _registerRekeyListener() {
         },
         'keybase.1.rekeyUI.rekeySendEvent': () => {}, // ignored debug call from daemon
       },
-      startMethod: 'keybase.1.rekeyUI.delegateRekeyUI',
     })
     response && response.result(session.id)
   })
--- a/shared/actions/unlock-folders.js
+++ b/shared/actions/unlock-folders.js
@@ -58,8 +58,9 @@ function _registerRekeyListener() {
   // else we get this also as part of delegateRekeyUI
   engine().setIncomingActionCreators('keybase.1.rekeyUI.delegateRekeyUI', (_, response, dispatch) => {
     // Dangling, never gets closed
-    const session = engine().createSession(
-      {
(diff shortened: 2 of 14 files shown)


shared/engine/transport-shared.js
```diff
@@ -7,19 +7,13 @@ import {requestIdleCallback} from '../util/idle-callback'
...
 import type {rpcLogType} from './index.platform'
 
-const {
-  transport: {RobustTransport},
-  client: {Client: RpcClient},
-} = rpc
-const KEYBASE_RPC_DELAY_RESULT: number = process.env.KEYBASE_RPC_DELAY_RESULT
-  ? parseInt(process.env.KEYBASE_RPC_DELAY_RESULT)
-  : 0
-const KEYBASE_RPC_DELAY: number = process.env.KEYBASE_RPC_DELAY ? parseInt(process.env.KEYBASE_RPC_DELAY) : 0
```
chrisnojima: i'm sure no one uses this so lets just remove it

shared/actions/tracker.js
```diff
@@ -618,11 +618,11 @@ function _setupTrackerHandlers() {
         }
       }
 
-      const session: Session = engine().createSession(
-        _serverCallMap(dispatch, getState, onStart, onFinish),
-        null,
-        cancelHandler
-      )
+      const session: Session = engine().createSession({
```
chrisnojima: lets use named parameters so its less confusing

shared/actions/unlock-folders.js
```diff
@@ -71,10 +72,8 @@ function _registerRekeyListener() {
         },
         'keybase.1.rekeyUI.rekeySendEvent': () => {}, // ignored debug call from daemon
       },
-      null,
-      null,
-      true
-    )
+      startMethod: 'keybase.1.rekeyUI.delegateRekeyUI',
```
chrisnojima: we now add a hint about the method even though its not required, helps logging

shared/engine/transport-shared.js
```diff
@@ -2,113 +2,90 @@
 // Classes used to handle RPCs. Ability to inject delays into calls to/from server
 import rpc from 'framed-msgpack-rpc'
 import {localLog} from '../util/forward-logs'
```
chrisnojima: generally cleaning up this file cause its pretty old

shared/engine/transport-shared.js
```diff
@@ -2,113 +2,90 @@
...
-import {printRPC} from '../local-debug'
+import {printRPC, printRPCStats} from '../local-debug'
 import {requestIdleCallback} from '../util/idle-callback'
-
-import type {rpcLogType} from './index.platform'
-
-const {
-  transport: {RobustTransport},
-  client: {Client: RpcClient},
-} = rpc
```
chrisnojima: I find this kind of deeply nested destructuring harder to read

shared/engine/transport-shared.js
```diff
@@ -2,113 +2,90 @@
...
+  methodInCallback: boolean,
+}): F {
+  const {handler, extra, method, type, enforceOnlyOnce, methodInCallback} = options
   let once = false
-  // $FlowIssue
-  return (...args) => {
-    if (once) {
-      rpcLog('engineInternal', 'ignoring multiple result calls', {args})
+  // $ForceType
+  const wrapped: F = (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5): void => {
```
chrisnojima: having one wrapping function is a lot simpler to understand than all these nested calls that make wrappers

shared/engine/transport-shared.js
```diff
@@ -2,113 +2,90 @@
...
+    connectCallback: () => void,
+    disconnectCallback: () => void,
+    incomingRPCCallback: (a: any) => void,
+    writeCallback: any
+  ) {
+    super(opts)
+
+    this.hooks = {
       connected: () => {
-        // $FlowIssue complains that this might be null
```
chrisnojima: flow rightly complains about using this before super is called. turns out we can just set hooks after

chrisnojima:
@keybase/react-hackers this is ready. This adds a little black bar for us in the app (even in prod mode). you can click it to hide it. other users can turn this on if they turn on the flag

chrisnojima:
@keybase/react-hackers bump

buoyad:
Oops meant to comment

PR #12484, opened 2018-06-21
Use purples in asset input

buoyad:
The changed storyshots are the 4 wallet ones, and styles/colors with the added `purple2_40`. Also adds a cross platform prop `placeholderColor` to `PlainInput`.  r? @keybase/react-hackers


shared/common-adapters/plain-input.desktop.js
```diff
@@ -170,7 +170,7 @@ class PlainInput extends React.PureComponent<InternalProps> {
 
   render = () => {
     const inputProps = this._getInputProps()
-    const css = `::-webkit-input-placeholder { color: rgba(0,0,0,.4); }
+    const css = `::-webkit-input-placeholder { color: ${this.props.placeholderColor || 'rgba(0,0,0,.4)'}; }
```
chrisnojima: kinda curious why we have to use css here instead of doing it inline? you can do WebkitInputPlaceholder i believe

shared/common-adapters/plain-input.native.js
```diff
@@ -128,6 +128,7 @@ class PlainInput extends Component<InternalProps, State> {
       onFocus: this._onFocus,
       onSubmitEditing: this.props.onEnterKeyDown,
       placeholder: this.props.placeholder,
+      placeholderTextColor: this.props.placeholderColor || 'rgba(0,0,0,.4)',
```
chrisnojima: we have black_40 already

shared/common-adapters/plain-input.desktop.js
chrisnojima: black_40

buoyad: It was a remnant of `Input.desktop`, I can't find anything about how to style pseudo-elements inline

chrisnojima: k we should still change to black_40

PR #12486, opened 2018-06-21
Fix 'You haven't chatted with yet' after canceling a pending conv

cjb:
@keybase/react-hackers 

Here's a fix for canceling a pending conversation not backing out of the "Start chatting" view on desktop.

But I think the fix suggests that `navigateToInbox()` itself should be doing this work -- it could perform the `isMobile` check and call `_maybeAutoselectNewestConversation()` itself instead of making callers do it.  I didn't do that here because I haven't looked at every caller to try to work out whether that would ever do the wrong thing.  Let's talk about it.


cjb:
@keybase/react-hackers Ready for review, added a parameter that makes `navigateToInbox()` take care of finding a new convo if you want it to.

PR #12494, opened 2018-06-21
optimization pass for keybasefriends loading

maxtaco:
- we were spending a fair amount of time in DeepCopy()s of the loader state, so try to skirt it as much as possible
- ~~attempts to get more aggressive resulted in test failures that I didn't track down~~
- the minimal full load kbfriends is about ~5.5s on my machine at this point, if the UPAK cache is warm.
- if CheapCopy() is replaced with DeepCopy(), then the load takes about ~9.5s


maxtaco:
ping @mlsteele

go/client/cmd_team.go
```diff
@@ -30,6 +30,7 @@ func NewCmdTeam(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command
 			newCmdTeamDelete(cl, g),
 			newCmdTeamAPI(cl, g),
 			newCmdTeamSettings(cl, g),
+			newCmdTeamProfileLoad(cl, g),
```
mlsteele: This command should be hidden from normal users. Is it?

maxtaco: the way you do that is to skip adding a `Usage` line.  Not the greatest, but that's what `cmd_riit.go` does, for instance.

PR #12498, opened 2018-06-21
Show close button for UnlockFolders

thebearjew:
This was a simple prop name mismatch that caused the close button to remain hidden.


thebearjew:
Rebased on master

PR #12502, opened 2018-06-22
UnboxError for erasablekvstore

joshblum:
It's possible that we can an error when trying to read deviceEKs from erasable storage. Rather than failing we want to 1. delete these keys  and 2. allow future key generation to proceed. Adds logging to try to determine if the noiseFile used for decryption is corrupted 
cc @oconnor663

### DIFF
--- a/go/ephemeral/lib.go
+++ b/go/ephemeral/lib.go
@@ -182,7 +182,7 @@ func (e *EKLib) newDeviceEKNeeded(ctx context.Context, merkleRoot libkb.MerkleRo
 	if err != nil {
 		switch err.(type) {
 		case *erasablekv.UnboxError:
-			e.G().Log.Debug(err.Error())
+			e.G().Log.Debug("newDeviceEKNeeded: DeviceEKStorage.MaxGeneration failed %v", err)
 			return true, nil
 		default:
 			return false, err
@@ -196,7 +196,7 @@ func (e *EKLib) newDeviceEKNeeded(ctx context.Context, merkleRoot libkb.MerkleRo
 	if err != nil {
 		switch err.(type) {
 		case *erasablekv.UnboxError:
-			e.G().Log.Debug(err.Error())
+			e.G().Log.Debug("newDeviceEKNeeded: DeviceEKStorage.Get failed %v", err)
 			return true, nil
 		default:
 			return false, err
--- a/go/ephemeral/lib.go
+++ b/go/ephemeral/lib.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	lru "github.com/hashicorp/golang-lru"
+	"github.com/keybase/client/go/erasablekv"
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/keybase1"
 )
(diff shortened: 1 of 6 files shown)


go/ephemeral/lib.go
```diff
@@ -179,15 +180,27 @@ func (e *EKLib) newDeviceEKNeeded(ctx context.Context, merkleRoot libkb.MerkleRo
 	s := e.G().GetDeviceEKStorage()
 	maxGeneration, err := s.MaxGeneration(ctx)
 	if err != nil {
-		return needed, err
+		switch err.(type) {
+		case *erasablekv.UnboxError:
+			e.G().Log.Debug(err.Error())
```
mmaxim: Can we annotate these error messages more?

go/erasablekv/erasable_kv_store.go
```diff
@@ -86,22 +101,28 @@ func (s *FileErasableKVStore) unbox(ctx context.Context, data []byte, noiseBytes
 	// Decode encrypted box
 	var boxed boxedData
 	if err := libkb.MPackDecode(data, &boxed); err != nil {
-		return err
+		return NewUnboxError(err.Error())
```
mmaxim: Do you want to just keep the underlying error around, instead of just the string?

go/erasablekv/erasable_kv_store.go
```diff
@@ -15,10 +17,23 @@ import (
...
+
+func (e UnboxError) Error() string {
+	return fmt.Sprintf("Unable to unbox item: %v", e.msg)
+}
+
 type boxedData struct {
 	V int
 	N [libkb.NaclDHNonceSize]byte
 	E []byte
+	H []byte // sha256 has of noise used for encryption
```
mmaxim: This only gets used when there is an error right? So there is no downside for it being blank on current installs?

joshblum: yeah it's only for the error message and if someone is unable to unbox in the wild we can't add it in since we can't decrypt. I don't think it's worth trying to add to current keys since these code changes will allow them to participate in ek gen and after a week new clients will report if this is the issue.

go/erasablekv/erasable_kv_store.go
joshblum: yes better

go/ephemeral/lib.go
joshblum: i'll add to here: https://github.com/keybase/client/pull/12502/files#diff-97b6f490cc3dd8c39619127caf5a83a6R29

mmaxim: I think it's worth adding some info to the specific call site too.

joshblum: done!

go/ephemeral/lib.go
```diff
@@ -179,15 +180,27 @@ func (e *EKLib) newDeviceEKNeeded(ctx context.Context, merkleRoot libkb.MerkleRo
 	s := e.G().GetDeviceEKStorage()
 	maxGeneration, err := s.MaxGeneration(ctx)
 	if err != nil {
-		return needed, err
+		switch err.(type) {
+		case *erasablekv.UnboxError:
```
mmaxim: Why is this a `*` here?

joshblum: `https://github.com/keybase/client/pull/12502/files#diff-97b6f490cc3dd8c39619127caf5a83a6R24` returns a pointer. should we change?

mmaxim: Yeah I'd change it to just be a normal type.

joshblum: done!

PR #12504, opened 2018-06-22
remove dependency on LookupOrCreateImplicitTeam everywhere except NewConversation CORE-8094

mmaxim:
Patch does the following:

1.) Changes `ImplicitTeamsNameInfoSource#Lookup` to use `LookupImplicitTeam` instead of `LookupOrCreateImplicitTeam`. This stops us from generating bogus teams outside of a `NewConversation` flow.
2.) Change `FindConversation` to process a missed team lookup and not error out under that circumstance. Instead, it just treats the result like finding no conversations. 
3.) Change CLI to also not use `LookupOrCreateImplicitTeam`.


mmaxim:
cc @maxtaco

PR #12505, opened 2018-06-22
jenkins: deal with random Windows lint issue

jzila:


jzila:
Yeah it's just a weird random error that we see once every ~100 windows builds.

PR #12506, opened 2018-06-22
add clean to build command

chrisnojima:
so we can do a build clean
@keybase/react-hackers


chrisnojima:
@keybase/react-hackers bump

PR #12509, opened 2018-06-22
Add forceReload argument to SecretSyncer to force key update

zapu:
It is useful to update keys even if we have latest versions sometimes, especially when trying to get accurate LastUsedTime. This commit fixes device list screen/commmand problems where users would not see accurate "last used" times on their devices.

I tried to find the minimal patch to make it possible to force update in SecretSyncer. I did not want to change Syncer interface API because that would require changes all over the codebase. So this PR adds a new field to SecretSyncer struct.

### DIFF
--- a/go/libkb/sync_secret.go
+++ b/go/libkb/sync_secret.go
@@ -63,9 +63,6 @@ type SecretSyncer struct {
 	Contextified
 	dirty bool
 	keys  *ServerPrivateKeys
-	// forceNextSync makes it so next time syncFromServer is performed,
-	// version will be disregarded and latest keys will always be fetched.
-	forceNextSync bool
 }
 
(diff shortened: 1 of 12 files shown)


go/libkb/sync_secret.go
```diff
@@ -63,6 +63,9 @@ type SecretSyncer struct {
 	Contextified
 	dirty bool
 	keys  *ServerPrivateKeys
+	// forceNextSync makes it so next time syncFromServer is performed,
+	// version will be disregarded and latest keys will always be fetched.
+	forceNextSync bool
```
maxtaco: not a huge fan of this, is there any reason we can't just change `RunSyncer` and push the force mode all the way down?

PR #12518, opened 2018-06-23
Fix Jenkins

jzila:


jzila:
Merging to fix CI, but cc @akalin-keybase and @zanderz for visibility.

PR #12530, opened 2018-06-25
Check if user was on buggy release before checking for newExploding item

buoyad:
Currently, we only check for `newExplodingGregorKey` to make decisions about whether to show the `NEW` tag or not. This is to be backwards compatible with the last release, which had a bug with the tag, and is intended to transition to the new gregor key without changing whether the `NEW` shows up. 

The problem is that the current code assumes users were on the buggy release. Since the absence of the gregor item means exploding isn't new, the `NEW` tag only shows for users who were on the bad version and happened to get into a good state, with the `NEW` tag visible. Anyone who wasn't on the bad version (most importantly new users) would never get the `NEW` tag. The fix is to check for the other item that was injected in the bad release. r? @keybase/react-hackers


buoyad:
Closing, we opted for server fix

PR #12531, opened 2018-06-25
WIP: parallelize flow coverage in circle

buoyad:


buoyad:
Hm, if circle fails before coverage is run, I'm not sure it runs coverage if no changes to `shared` are included in the next commit.

buoyad:
This works but flow only takes ~2m or so on cached runs, so it's not as big an improvement as I thought. It also makes runs take up an extra container, which would probably offset any benefit from it. Closing this, the config might be useful if we have more parallelize-able stuff in the future.

PR #12546, opened 2018-06-25
Extend ephemeral staleness lifetime to 1 month

joshblum:
Extends the staleness window for ephemeral keys to be 1 month and separates the logic so we can easily alter this value in the future. existing code is refactored to use the standard `time` package throughout so we can use `time.Duration` for our values and simplify some of the unit munging between time formats we used to do

we also drop support for existing metadata in all ek statements to match the server side changes (depends on https://github.com/keybase/keybase/pull/2618)


go/systests/ephemeral_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	context "golang.org/x/net/context"
 )
 
-func TestNewTeamEKNotif(t *testing.T) {
+func TestEphemeralNewTeamEKNotif(t *testing.T) {
```
joshblum: these changes were just so we could run `go test --run TestEphemeral` to verify server side changes more easily

PR #12547, opened 2018-06-25
Load updated avatars.

adamjspooner:
@keybase/react-hackers, this is ready for review. There isn't a separate call for user avatar updates, but @zapu can add one if we think it's necessary.


shared/actions/teams.js
```diff
@@ -1028,6 +1029,18 @@ function _setupTeamHandlers() {
       return getLoadCalls()
     }
   )
+  engine().setIncomingActionCreators(
+    'keybase.1.NotifyTeam.avatarUpdated',
+    (args: RPCTypes.NotifyTeamAvatarUpdatedRpcParam, _, __, getState) => {
```
chrisnojima: prefer
```
({name}: RPCTypes.NotifyTeamAvatarUpdatedRpcParam,...
```

shared/actions/teams.js
```diff
@@ -1028,6 +1029,18 @@ function _setupTeamHandlers() {
...
+      const actions = []
+      if (getState().teams.teamnames.includes(args.name)) {
```
chrisnojima: maybe more succinct

```
return [
  getState().teams.teamnames.includes(name) 
   ? ConfigGen.createLoadTeamAvatars({teamnames: [name]})
   : ConfigGen.createLoadAvatars({usernames: [name]})
]
```

PR #12551, opened 2018-06-25
Signup part 5

chrisnojima:
This pr

- [x] addresses some visual feedback from @cecileboucheron 
- [x] fixes some things from parts 1-4 from @akalin-keybase and @buoyad 
- [x] introduces a new util/saga helper to deal withe rpcs that return puts as promises (all the validation logic)
- [x] Checks all errors at every step of the way to be very careful about the state of the signup flow


cecileboucheron:
* Full width buttons on desktop should be 40px high: https://zpl.io/VkKl3Jg
* Inputs are now sticking to the edges on mobile, let's have some 40px left/right paddings
* Avatar still too close to the blue label on mobile
* Mobile avatar seems to be an odd size, can it be the 128px one (seems like the desktop one is fine)
* Not worth spending too much time but inputs and button's edges should be aligned ideally.

chrisnojima:
- [ ] device error not plumbed through

chrisnojima:
@buoyad @cecileboucheron updated

shared/login/signup/common.js
```diff
@@ -0,0 +1,70 @@
...
+      direction="vertical"
+      fullWidth={true}
+      fullHeight={true}
+      centerChildren={true}
+      style={styles.wrapper}
+      gap={isMobile ? 'xtiny' : 'small'}
+    >
+      {props.children}
+    </Box2>
+    <HeaderHocHeader onBack={props.onBack} headerStyle={styles.header} />
```
buoyad: this works but the bottom border still ends to the left:
image

chrisnojima: k, i  think we should just remove the border anyways, we're not using a title

shared/login/signup/device-name/index.js
```diff
@@ -25,7 +24,7 @@ class Devicename extends React.Component<Props, State> {
 
     return (
       <Wrapper onBack={this.props.onBack}>
-        <Text type="Header">A public name for this device:</Text>
+        <Text type="Header">Set a public name...</Text>
         <Icon type="icon-computer-64" />
         <Input
```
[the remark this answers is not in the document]
buoyad: I don't know how it made it past the call to `deviceCheckDeviceNameFormatRpcPromise` we make

cecileboucheron:
Looks good to me - except that error code 201 I sent you.

chrisnojima:
the 201 was unrelated to this pr

chrisnojima:

chrisnojima:
cool lemme fix that, annoying that signup and provison share this little bit of code

chrisnojima:
@akalin fixed

shared/local-debug.native.js
```diff
@@ -49,7 +49,7 @@ let config = {
 // Developer settings
 if (__DEV__) {
   config.enableActionLogging = true
-  config.enableStoreLogging = false
+  config.enableStoreLogging = true
```
akalin-keybase: intended?

chrisnojima:
I still want another thumb but in prep I'm going to merge down so ci passes. If you review you can use these prs or look at the merge down

PR #12553, opened 2018-06-26
teamEK background creation

joshblum:
If we have access to the current teamEK but it is expired (daily keygen needed) we can now do this generation in the background and temporarily use the old teamEK. Since we lazily create teamEKs for large teams, it can be a burn to get stuck with the daily generation and have you message hang (seemingly randomly) while this is happening. 

note this will conflict with https://github.com/keybase/client/pull/12546 because we reference `KeyGenLifetimeSecs` in tests

### DIFF
--- a/go/ephemeral/common.go
+++ b/go/ephemeral/common.go
@@ -65,6 +65,18 @@ func ctimeIsStale(ctime keybase1.Time, currentMerkleRoot libkb.MerkleRoot) bool 
 	return currentMerkleRoot.Ctime()-ctime.UnixSeconds() >= KeyLifetimeSecs
 }
 
+// If a teamEK is almost expired we allow it to be created in the background so
+// content generation is not blocked by key generation. We *cannot* create a
+// teamEK in the background if the key is expired however since the current
+// teamEK's lifetime (and supporting device/user EKs) is less than the maximum
+// lifetime of ephemeral content. This can result in content loss once the keys
+// are deleted.
+func backgroundKeygenPossible(ctime time.Time, currentMerkleRoot libkb.MerkleRoot) bool {
+	diff := keybase1.TimeFromSeconds(currentMerkleRoot.Ctime()).Time().Sub(ctime)
+	keygenInterval := time.Second * time.Duration(KeyGenLifetimeSecs)
+	return diff >= (keygenInterval-time.Hour) && diff < keygenInterval
+}
+
 func keygenNeeded(ctime keybase1.Time, currentMerkleRoot libkb.MerkleRoot) bool {
 	return currentMerkleRoot.Ctime()-ctime.UnixSeconds() >= KeyGenLifetimeSecs
 }
--- a/go/ephemeral/lib.go
+++ b/go/ephemeral/lib.go
@@ -11,21 +11,23 @@ import (
 	"github.com/keybase/client/go/erasablekv"
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/clockwork"
 )
 
 const SkipKeygenNilMerkleRoot = "Skipping key generation, unable to fetch merkle root"
 
-const cacheEntryLifetimeSecs = 60 * 5 // 5 minutes
+const cacheEntryLifetime = time.Minute * 5
 const lruSize = 200
 
 type EKLib struct {
 	libkb.Contextified
 	teamEKGenCache *lru.Cache
 	sync.Mutex
-}
 
-func (e *EKLib) metaContext(ctx context.Context) libkb.MetaContext {
-	return libkb.NewMetaContext(ctx, e.G())
+	// During testing we may want to stall background creation to assert cache
+	// state.
+	clock                    clockwork.Clock
+	backgroundCreationTestCh chan bool
 }
 
 func NewEKLib(g *libkb.GlobalContext) *EKLib {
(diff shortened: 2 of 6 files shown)


go/ephemeral/lib.go
```diff
@@ -11,21 +11,23 @@ import (
 	"github.com/keybase/client/go/erasablekv"
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/clockwork"
 )
 
 const SkipKeygenNilMerkleRoot = "Skipping key generation, unable to fetch merkle root"
 
-const cacheEntryLifetimeSecs = 60 * 5 // 5 minutes
+const cacheEntryLifetime = time.Minute * 5
```
oconnor663: Just confirming the intention: This is going from seconds (our previous units) to nanoseconds (the Go std lib units)?

go/ephemeral/lib.go
```diff
@@ -288,46 +303,48 @@ func (e *EKLib) newTeamEKNeeded(ctx context.Context, teamID keybase1.TeamID, mer
...
+			return true, false, latestGeneration, nil
 		default:
-			return false, latestGeneration, err
+			return false, false, latestGeneration, err
 		}
 	}
 	// Ok we can access the ek, check lifetime.
 	e.G().Log.CDebugf(ctx, "nextTeamEKNeeded at: %v", nextKeygenTime(ek.Metadata.Ctime))
-	return keygenNeeded(ek.Metadata.Ctime, merkleRoot), latestGeneration, nil
+	return keygenNeeded(ek.Metadata.Ctime, merkleRoot), true, latestGeneration, nil
```
oconnor663: Should this be changed to only true if the key is < 24 hrs old?

go/ephemeral/lib.go
```diff
@@ -384,23 +401,55 @@ func (e *EKLib) getOrCreateLatestTeamEKInner(ctx context.Context, teamID keybase
...
+				if e.backgroundCreationTestCh != nil {
+					select {
+					case <-e.backgroundCreationTestCh:
+					}
+				}
+
+				publishedMetadata, err := publishNewTeamEK(ctx, e.G(), teamID, merkleRoot)
+				// Grab the lock once we finish publishing so we do don't block
+				e.Lock()
+				defer e.Unlock()
```
oconnor663: I think there's a race here. Thread 1 publishes a new teamEK and gets to this line. Thread 2 then publishes _yet another_ teamEK (at the following generation) and gets to this line. If thread 2 happens to grab this lock first, then thread 1 will grab it after and overwrite thread 2's key.

It's a really unlikely race, but maybe we want a map of per-team locks?

go/ephemeral/lib.go
joshblum: yes since we're using `time.Time` calculations

go/ephemeral/lib.go
joshblum: i don't think it's possible since only one thread can be in here at a time since the `CreationInProgress` flag is set, blocking thread2 from reaching the code to allow background generation

joshblum:
note: we can only issue teamEK asynchronously if the current teamEK is < 1 day old since our deletion policy for teamEKs is maxEphemeralContentLifetime + 1 from the creation date. If we allowed background creation after this day window, the deletion lifetime would have to be extended to account for the gap which we cannot determine currently for the deviceek/usereks without reboxing of teameks

go/gregor/client/client.go
```diff
@@ -36,7 +36,7 @@ type Client struct {
 	Sm      gregor.StateMachine
 	Storage LocalStorageEngine
 	Log     logger.Logger
-	Clock   clockwork.Clock
```
joshblum: cc @mmaxim for any of these changes

go/ephemeral/lib.go
```diff
@@ -357,7 +377,7 @@ func (e *EKLib) getOrCreateLatestTeamEKInner(ctx context.Context, teamID keybase
 	cacheKey := e.cacheKey(teamID)
 	val, ok := e.teamEKGenCache.Get(cacheKey)
 	if ok {
-		if cacheEntry, valid := e.isEntryValid(val); valid {
+		if cacheEntry, expired := e.isEntryExpired(val); !expired || cacheEntry.CreationInProgress {
```
oconnor663: Could this cause trouble if background key creation fails? I'm worried we could be in a situation where the `CreationInProgress` flag is set, and it never gets unset, and so the `expired` bool never gets respected.

go/ephemeral/common.go
```diff
@@ -65,6 +65,18 @@ func ctimeIsStale(ctime keybase1.Time, currentMerkleRoot libkb.MerkleRoot) bool
...
+// If a teamEK is almost expired we allow it to be created in the background so
+// content generation is not blocked by key generation. We *cannot* create a
+// teamEK in the background if the key is expired however since the current
+// teamEK's lifetime (and supporting device/user EKs) is less than the maximum
+// lifetime of ephemeral content. This can result in content loss once the keys
+// are deleted.
+func backgroundKeygenPossible(ctime time.Time, currentMerkleRoot libkb.MerkleRoot) bool {
+	diff := keybase1.TimeFromSeconds(currentMerkleRoot.Ctime()).Time().Sub(ctime)
+	keygenInterval := time.Second * time.Duration(KeyGenLifetimeSecs)
+	return diff >= (keygenInterval-time.Hour) && diff < keygenInterval
```
oconnor663: This would be easier for me to follow with some more concrete variable names. Maybe `keyAge` or something instead of `diff`? And maybe give the return value an intermediate name like `isOneHourFromExpiration`?

go/ephemeral/lib.go
joshblum: we clear the cache on errors: https://github.com/keybase/client/pull/12553/files#diff-6dc458a542227a8fc53f68de4223c8baR430

go/ephemeral/common.go
joshblum: i'll change this in the subsequent PR with deletion fixes

PR #12557, opened 2018-06-26
Resize desktop add to team dropdown

nathunsmitty:
Go from this:
screen shot 2018-06-26 at 1 46 09 pm

To this:
screen shot 2018-06-26 at 1 46 24 pm

The fix was to remove a hardcoded width of 270 from the dropdown component's styling. I'm unsure if this will affect anything else, though, so please let me know if you know where the 270 came from.


nathunsmitty:
Ok! I think we're using the `Dropdown` component in 5 places across the app:
1. When selecting a team to assign a team git repo to
2. When a user is selecting one of multiple accounts when logging in
3. When a user is adjusting another user's permissions on a team by adding them via the teams pane by username
4. When a user is adjusting another user's permissions on a team by adding them via the teams pane by email
5. When a user is adjusting another user's permissions on a team by adding them via the people pane

3 was the page that got the bug initially reported, but I think it's fair to say 4 and 5 are also instances of the same bug. Currently, they look like this (in order of 3, 4, 5):

screen shot 2018-06-27 at 10 29 08 am
screen shot 2018-06-27 at 10 29 12 am
screen shot 2018-06-27 at 10 28 02 am
 
This fix will cause them to look like:
screen shot 2018-06-27 at 10 42 05 am
screen shot 2018-06-27 at 10 42 12 am
screen shot 2018-06-27 at 10 28 10 am

Note that there are some margin and wording differences between 3 and 4; I'd like to file a separate ticket to deal with those inconsistencies.

Technically, this fix turned out to be a little more complicated than I originally thought. The line I changed in my first commit was the crux of the issue, but merely adding a `style` prop to the affected `Dropdown`s caused an error due to the width of 270px being set in the `ButtonBox` component as opposed to the parent `Box` component that was receiving the `style` prop. The solution was to move the default 270 up, then override it if a different width is given. The `ButtonBox` is then set to the same width as `Box`.

nathunsmitty:
@keybase/react-hackers

shared/profile/add-to-team/index.js
```diff
@@ -143,6 +143,7 @@ const AddToTeam = (props: Props) => {
               const selectedRole: TeamRoleType = (node && node.key) || null
               props.onRoleChange(selectedRole)
             }}
+            style={{width: 'auto'}}
```
chrisnojima: i'm not sure this works on mobile, can you check

nathunsmitty: It does not. Fixing rn

nathunsmitty:
In mobile, we use the `Dropdown` component a little less than on desktop; of the 5 uses I mentioned above, only 2, 3, and 5 exist in mobile. Here are how they look currently (in order):

img_2326
img_2327
img_2328

Of these three, I think only 3 is excessively long on mobile? Here's how it'll look with the mobile fixes:
simulator screen shot - iphone 7 - 2018-06-27 at 15 39 30

chrisnojima:
@cecileboucheron you ok with this short term. i know we have a different design ultimately but this is likely better than what we have now

cecileboucheron:
LGTM

PR #12558, opened 2018-06-26
Add big teams divider on less than 6 small teams

nathunsmitty:
When a user has less than 6 small team conversations and are also in a big team, they see an inbox similar to the following:

screen shot 2018-06-26 at 3 37 44 pm

The problem with this behavior is that the small teams and big teams aren't very well separated. The solution? To render the divider whenever a user has any big teams in their inbox. Previously, we rendered a button to expand/collapse the small teams chats. Since this only makes sense to do with 6+ chats, we now conditionally render the button within the `Divider` component. In order to facilitate this, we need to pass `Divider` component an extra prop, `showSmallTeamsExpandDivider`.

The result:

screen shot 2018-06-26 at 3 44 23 pm

For reference, this is how 6+ conversations looks (note the lack of the `Big teams` header):
screen shot 2018-06-26 at 3 46 02 pm

screen shot 2018-06-26 at 3 46 08 pm


cecileboucheron:
Looks good to me! 👍

chrisnojima:
Please don't merge until approval from react hackers

nathunsmitty:
@keybase/react-hackers

PR #12559, opened 2018-06-26
Fix inbox measuring issues

chrisnojima:
We had an issue w/ inbox on mobile sometimes not showing items on the bottom. This is because we help the inbox measure itself so it doesn't have to load the entire list and we can get correct scrollbars. Because of this we need to know in advance the sizes of things. Before we implicitly did this but when there was a pass at styling the rows the sizes got changed but the measuring did not.

In order to stop this from happening again we now have a RowSizes object which is used to explicitly set the heights of rows and is used by the measurer.

The divider recently changed to actually have 2 sizes so we now plumb this through into the row type

cc: @cecileboucheron 
cc: @nathunsmitty since i changed some stuff  you were just in

I also moved to Box 2 to remove all these inner margins


chrisnojima:
@keybase/react-hackers

shared/chat/inbox/row/big-team-channel/index.js
```diff
@@ -84,57 +74,75 @@ const mutedStyle = {
...
-  paddingRight: globalMargins.tiny,
-}
-
-const selectedChannelBackgroundStyle = {
-  ...channelBackgroundStyle,
-  backgroundColor: globalColors.blue,
-  borderTopLeftRadius: 3,
-  borderBottomLeftRadius: 3,
-}
+const styles = styleSheetCreate({
```
nathunsmitty: What's with `styleSheetCreate`? Is this a preferred practice to the method of creating separate objects for each component's styles?

shared/chat/inbox/row/divider/index.js
```diff
@@ -1,85 +1,105 @@
...
 }
 
 class Divider extends React.PureComponent<Props> {
   render() {
     return (
-      <Box style={this.props.style ? {..._toggleContainer, ...this.props.style} : _toggleContainer}>
-        {this.props.showSmallTeamsExpandDivider && (
-          <ClickableBox onClick={this.props.toggle} className="toggleButtonClass" style={_toggleButtonStyle}>
-            <Text type="BodySmallSemibold" style={_buttonTextStyle}>
+      <Box2
```
nathunsmitty: What's the difference between `Box` and `Box2`? `Box2` seems to have some extra styling in it, but what was the motivation for the creation of `Box2`?

shared/chat/inbox/row/small-team/index.js
```diff
@@ -91,33 +104,42 @@ class SmallTeam extends React.PureComponent<Props> {
...
+  conversationRowSelected: {
+    ...globalStyles.flexBoxColumn,
+    backgroundColor: globalColors.blue,
+    flexGrow: 1,
+    height: '100%',
+    justifyContent: 'center',
+    paddingLeft: 8,
+    paddingRight: 8,
+  },
+  rowContainer: platformStyles({
```
nathunsmitty: Is `platformStyles` what you were telling me about during lunch? Could you possibly explain a little more how it's different from and how it interacts with the platform constants? Is it just to satisfy typing with Flow?

adamjspooner: Yup, it's the one that Nojima was talking about at lunch the other day. Take a look at it's definition to see how it works: https://github.com/keybase/client/blob/master/shared/styles/shared.js#L48

shared/chat/inbox/row/divider/index.js
adamjspooner: Take a look at the stories for `Box` and `Box2` to understand a bit more about them: https://github.com/keybase/client/blob/master/shared/common-adapters/box.stories.js

Basically `Box` is just a `div` (on desktop) and a `View` (on mobile). `Box2` has a few props for injecting normal gaps, etc. Have a look at its props to know what's available: https://github.com/keybase/client/blob/master/shared/common-adapters/box.stories.js

shared/chat/inbox/row/big-team-channel/index.js
adamjspooner: Nojima can explain this better, but basically it's a helper for creating cross-platform styles because creating a stylesheet differs between React and React Native. Take a look at the differences between them to understand a little more:

- React: https://github.com/keybase/client/blob/master/shared/styles/index.desktop.js#L113
- React Native:https://github.com/keybase/client/blob/master/shared/styles/index.native.js#L52

shared/chat/inbox/index.native.js
```diff
@@ -143,36 +140,43 @@ class Inbox extends React.PureComponent<Props, State> {
...
     }
-
-    return {index, length: this._itemTypeToHeight[data[index].type], offset}
+    const length = data[index].type === 'big' ? RowSizes.bigRowHeight : RowSizes.bigHeaderHeight
+    return {index, length, offset}
   }
 
   render() {
+    this._dividerShowButton = false
+    this._dividerIndex = this.props.rows.findIndex(r => {
```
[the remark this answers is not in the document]
chrisnojima: not really. if we're rendering its likely cause the rows changed so we need to search again. i'm sure most inboxes are < 1000 items so its not that expensive. the getitemlayout gets called as you scroll around so we want it cached like this

PR #12562, opened 2018-06-26
go/vendor: replace code.google.com/p/rsc/qr with rsc.io/qr

rsc:
This will ease your eventual transition to Go modules.
It looks like this is the only package you depend on
that doesn't exist anymore. (Sorry about that.)


maxtaco:
Thanks @rsc !

PR #12564, opened 2018-06-27
skip empty username in login accounts list

zanderz:
Empty username was coming from service. Maybe didn't need to change both sides but this fixes it.


go/libkb/secret_store.go
```diff
@@ -90,12 +90,14 @@ func NewSecretStore(g *GlobalContext, username NormalizedUsername) SecretStore {
...
-	currentUsername, otherUsernames, err := c.GetAllUserNames()
+	currentUsername, allUsernames, err := c.GetAllUserNames()
 	if err != nil {
 		return nil, err
 	}
 
-	allUsernames := append(otherUsernames, currentUsername)
+	if currentUsername != "" {
+		allUsernames = append(allUsernames, currentUsername)
+	}
```
songgao: Do we wanna filter on other ones from `allUsernames` too or does it never contain empty strings?

go/libkb/config.go
```diff
@@ -410,7 +410,7 @@ func (f JSONConfigFile) GetAllUsernames() (current NormalizedUsername, others []
 			return
 		}
 		nu := NewNormalizedUsername(name)
-		if !nu.Eq(current) {
+		if !nu.Eq(current) && !nu.IsNil() {
```
maxtaco: Instead of doing this, it's better to use `GetAllProvisionedUsernames` instead.  That's going to be more correct for various scenarios, such as when the device is question was offline at the time of the reset.

maxtaco: And BTW, the bug with the code that's currently there is that the `currentUsername` is coming back as empty, since the current user doesn't exist after the reset.

zanderz: sorry, Max, I don't quite follow. `GetAllProvisionedUsernames` instead of what, exectly? `SecretStoreContext.GetAllUserNames` ? And how to resolve the right context to call `GetAllProvisionedUsernames` from inside the secret store?

maxtaco: See the comment below. Instead of `currentUsername, otherUsernames, err := c.GetAllUserNames()`, call `GetAllProvisionedUsernames`.  Change `GetConfiguredAccounts` to take a `MetaContext` and pass that through.

go/engine/login_state_test.go
```diff
@@ -175,6 +175,7 @@ func TestLoginWithPromptPassphrase(t *testing.T) {
 }
 
 func userHasStoredSecretViaConfiguredAccounts(tc *libkb.TestContext, username string) bool {
+	m := libkb.NewMetaContext(tc, tc.G)
```
maxtaco: Change this to `NewMetaContextForTest(tc)` and it will compile

go/libkb/config.go
maxtaco: Can you remove this `nu.IsNil` check here?  I don't think it's needed.

go/libkb/secret_store.go
```diff
@@ -115,7 +118,7 @@ func GetConfiguredAccounts(c SecretStoreContext, s SecretStoreAll) ([]keybase1.C
 	for _, username := range storedSecretUsernames {
 		nu := NewNormalizedUsername(username)
 		account, ok := accounts[nu]
-		if ok {
+		if ok && !nu.IsNil() {
```
maxtaco: Let's get rid of this check too. I don't see how we could have an empty username here.

PR #12565, opened 2018-06-27
migrate to FloatingMenu and clean up

songgao:


shared/fs/common/path-item-action-container.js
```diff
@@ -74,9 +53,10 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
     : {}),
 })
 
-const getRootMenuItems = (stateProps, dispatchProps) => {
-  const {path, pathItem, fileUIEnabled, _username} = stateProps
+const getRootMenuItems = (stateProps, dispatchProps, path: Types.Path) => {
+  const {_pathItems, fileUIEnabled, _username} = stateProps
   const {_showInFileUI, _saveMedia, _shareNative, _download, _ignoreFolder} = dispatchProps
+  const pathItem = _pathItems.get(path) || Constants.makeUnknownPathItem()
```
chrisnojima: you can use .get(path, Constants.makeUnknownPathItem()) or even bake that into a helper in constants

shared/fs/common/path-item-action-container.js
```diff
@@ -118,14 +98,30 @@ const getRootMenuItems = (stateProps, dispatchProps) => {
...
 }
 
-const mergeProps = (stateProps, dispatchProps) => {
-  const {path, pathItem, _username, childrenFolders, childrenFiles} = stateProps
-  const {onHidden, _loadMimeType} = dispatchProps
+const mergeProps = (stateProps, dispatchProps, {path, actionIconClassName, actionIconFontSize}) => {
+  const {_pathItems, _username} = stateProps
+  const {_loadMimeType} = dispatchProps
+  const pathItem = _pathItems.get(path) || Constants.makeUnknownPathItem()
+  const [childrenFolders, childrenFiles] =
```
chrisnojima: might be clearer to just use an object instead of an array-as-tuple, aka
const {folders, files} = pathItem.children.reduce((...) => {}, {folder: 0, files: 0})

shared/fs/common/path-item-action.js
```diff
@@ -0,0 +1,129 @@
...
+      containerStyle={stylesFloatingContainer}
+      attachTo={props.attachmentRef}
+      visible={props.showingMenu}
+      onHidden={props.toggleShowingMenu}
+      position="bottom right"
+      closeOnSelect={true}
+      header={{
+        title: 'unused',
+        view: (
+          <Box style={stylesHeader}>
```
chrisnojima: you might want to break this out into its own component, can be in this file but outside of all this nesting

shared/fs/common/path-item-action.js
```diff
@@ -0,0 +1,129 @@
...
+            )}
+            <PathItemInfo
+              lastModifiedTimestamp={props.lastModifiedTimestamp}
+              lastWriter={props.lastWriter}
+              wrap={true}
+            />
+          </Box>
+        ),
+      }}
+      items={props.menuItems.map(a => a /* make flow happy */)}
```
chrisnojima: cause some can be null? you can do
`props.menuItems.filter(Boolean)`

shared/fs/header/breadcrumb-container.desktop.js
```diff
@@ -0,0 +1,52 @@
...
+import {fsTab} from '../../constants/tabs'
+import {navigateTo} from '../../actions/route-tree'
+import Breadcrumb from './breadcrumb.desktop'
+
+type OwnProps = {
+  path: Types.Path,
+}
+
+const mapStateToProps = (state: TypedState) => ({
+  _username: state.config.username || undefined,
```
chrisnojima: why do you need undefined?

shared/fs/header/breadcrumb-container.desktop.js
```diff
@@ -0,0 +1,52 @@
...
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  _navigateTo: (path: Types.Path) => dispatch(navigateTo([fsTab, {props: {path}, selected: 'folder'}])),
+})
+
+const mergeProps = ({_username}, {_navigateTo}, {path}: OwnProps) => {
+  const elems = Types.getPathElements(path)
+
+  let acc = Types.stringToPath('/')
+  const items = elems.map((elem, i) => {
```
chrisnojima: why not reduce?

songgao:
> do we really need to have so much split mobile/desktop files?

Breadcrumb is only used on desktop, so it's not really a split. Or did you mean something else?

shared/fs/common/path-item-action-container.js
songgao: I've been using `.get(path, Constants.makeUnknownPathItem())` but forgot to fix this one when copy/pasting.

I'm curious why a helper in constants would be better in this case? I assume it's gonna look like `getPathItem(pathItems: Map<...>, path: Path)` which seems similar to this form.

shared/fs/common/path-item-action.js
songgao: No I think it's because `menuItems` here is a different array, despite it's actually more restrictive:

```
$ /Users/songgao/gopath/src/github.com/keybase/client/shared/node_modules/.bin/flow
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ fs/common/path-item-action.js:82:14

Cannot create FloatingMenu element because:
 • inexact string literal Divider [1] is incompatible with exact MenuItem [2] in array element of property items.
 • inexact null [3] is incompatible with exact MenuItem [2] in array element of property items.

        fs/common/path-item-action.js
    [2] 29│   menuItems: Array<MenuItem>,
          :
        68│         className={props.actionIconClassName}
        69│       />
        70│     </ClickableBox>
        71│     <FloatingMenu
        72│       containerStyle={stylesFloatingContainer}
        73│       attachTo={props.attachmentRef}
        74│       visible={props.showingMenu}
        75│       onHidden={props.toggleShowingMenu}
        76│       position="bottom right"
        77│       closeOnSelect={true}
        78│       header={{
        79│         title: 'unused',
        80│         view: <PathItemActionHeader {...props} />,
        81│       }}
        82│       items={props.menuItems}
        83│     />
        84│   </Box>
        85│ )
        86│

        common-adapters/floating-menu.js
 [1][3] 12│   items: Array<MenuItem | 'Divider' | null>,
```

songgao: I'll just change the `menuItems` type here.

shared/fs/common/path-item-action-container.js
```diff
@@ -101,21 +101,21 @@ const getRootMenuItems = (stateProps, dispatchProps, path: Types.Path) => {
...
+  const pathItem = _pathItems.get(path, Constants.unknownPathItem)
+  const {childrenFolders, childrenFiles} =
     !pathItem || pathItem.type !== 'folder'
-      ? [0, 0]
+      ? {childrenFolders: 0, childrenFiles: 0}
       : pathItem.children.reduce(
-          ([currentFolders, currentFiles], p) => {
+          ({childrenFolders, childrenFiles}, p) => {
             const pi = _pathItems.get(Types.pathConcat(path, p))
             if (!pi) {
```
chrisnojima: could just be
```
const isFolder = pi && pi.type === 'folder'
return {
  childrenFolders: childrenFolders + isFolder ? 1 : 0, 
  childrenFiles + isFolder ? 0 : 1
}
```

PR #12566, opened 2018-06-27
Stellar: Show assets in the GUI

cjb:
@keybase/react-hackers 

This PR:

1. Hooks up showing assets in the GUI. They're loaded on-demand when you click on an account in the account list.  There are a few missing fields on the CORE side, nothing very important, I'm filing a CORE bug for those.
2. Fixes a black bar on master due to an RPC payload change -- `res.map` => `res.payments.map`
3. Fixes the account list on master. The account list was depending on the dev-only `loadEverything` RPC to perform its data load, but this code is live on admin builds (with `__DEV__` false) now, so this call wasn't doing anything.  Now we  `loadAccounts` when the main component loads, and `loadAssets` and `loadPayments` when you select an account.  So, `loadEverything` can be retired now.

There should soon be one scrollable List with sections for assets and payments, rather than a list of assets and a list of payments.  I haven't done that here, but we're using `<List />` for everything so combining them later should be straightforward.

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -3,9 +3,7 @@ import * as Constants from '../constants/wallets'
 import * as Types from '../constants/types/wallets'
 import * as RPCTypes from '../constants/types/rpc-stellar-gen'
 import * as Saga from '../util/saga'
-import * as WaitingGen from './waiting-gen'
 import * as WalletsGen from './wallets-gen'
-import type {TypedState} from '../util/container'
 
 const loadAccounts = (action: WalletsGen.LoadAccountsPayload) =>
   Saga.call(RPCTypes.localGetWalletAccountsLocalRpcPromise)
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -38,10 +38,16 @@ const loadPaymentsSuccess = (res: any, action: WalletsGen.LoadPaymentsPayload) =
   return Saga.put(
     WalletsGen.createPaymentsReceived({
       accountID,
-      payments: res.map(elem => Constants.paymentResultToPayment(elem)),
+      payments: res.payments.map(elem => Constants.paymentResultToPayment(elem)),
     })
   )
 }
+
+const loadAccount = (action: WalletsGen.SelectAccountPayload) =>
+  Saga.all([
+    Saga.put(WalletsGen.createLoadAssets({accountID: action.payload.accountID})),
+    Saga.put(WalletsGen.createLoadPayments({accountID: action.payload.accountID})),
+  ])
 
 function* loadEverything(action: WalletsGen.LoadEverythingPayload) {
   if (__DEV__) {
@@ -60,6 +66,7 @@ function* loadEverything(action: WalletsGen.LoadEverythingPayload) {
 
 function* walletsSaga(): Saga.SagaGenerator<any, any> {
   yield Saga.safeTakeEveryPure(WalletsGen.loadAccounts, loadAccounts, loadAccountsSuccess)
+  yield Saga.safeTakeEveryPure(WalletsGen.selectAccount, loadAccount)
   yield Saga.safeTakeEveryPure(WalletsGen.loadAssets, loadAssets, loadAssetsSuccess)
   yield Saga.safeTakeEveryPure(WalletsGen.loadPayments, loadPayments, loadPaymentsSuccess)
   // Debugging saga -- remove before launching.
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -43,36 +43,12 @@ const loadPaymentsSuccess = (res: any, action: WalletsGen.LoadPaymentsPayload) =
   )
 }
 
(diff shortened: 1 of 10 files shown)


shared/actions/wallets.js
```diff
@@ -60,6 +66,7 @@ function* loadEverything(action: WalletsGen.LoadEverythingPayload) {
 
 function* walletsSaga(): Saga.SagaGenerator<any, any> {
   yield Saga.safeTakeEveryPure(WalletsGen.loadAccounts, loadAccounts, loadAccountsSuccess)
+  yield Saga.safeTakeEveryPure(WalletsGen.selectAccount, loadAccount)
```
chrisnojima: i think we want to invert this relationship.

We want to have an array of actions which lead to a single side effect. We don't want to have an action which spawns a bunch of actions

aka

we should prefer
```
yield Saga.safeTakeEveryPure(WalletsGen.selectAccount, loadAssets)
yield Saga.safeTakeEveryPure(WalletsGen.selectAccount, loadPayments)
```

then if some other thing causes a loadAssets, like a call from the server saying assets changed we'd do
```
yield Saga.safeTakeEveryPure([WalletsGen.selectAccount, WalletsGen.recievedAssetsUpdatedNotification], loadAssets)
```

I think this leads to less logical flows and less branching. There's one place where loadAssets actually happens and the relationships are more clear.

Let me know if the above makes sense. @keybase/react-hackers this ^ is something we should all grok

shared/wallets/asset/container.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+import {Assets} from '.'
+import {connect, type TypedState, type Dispatch} from '../../util/container'
+import {getAssets, getSelectedAccount} from '../../constants/wallets'
+
+const mapStateToProps = (state: TypedState) => {
+  const selectedAccount = getSelectedAccount(state)
+  const assets = getAssets(state, selectedAccount)
```
chrisnojima: we could make these helpers just default to the selected account if you don't pass one in. likely 95% of the time this is what we'll want

shared/wallets/asset/index.js
```diff
@@ -154,3 +152,63 @@ const styles = styleSheetCreate({
 })
 
 export default Asset
+
+type AssetWrappedState = {
+  expanded: boolean,
+}
+
+export class AssetWrapped extends React.Component<Props, AssetWrappedState> {
```
chrisnojima: seems like we should just merge this into Asset and not have this composition

shared/wallets/asset/index.js
```diff
@@ -154,3 +152,63 @@ const styles = styleSheetCreate({
...
+type AssetsProps = {
+  assets: Array<Props>,
+}
+
+export class Assets extends React.Component<AssetsProps> {
+  _renderRow = (i: number, row: Row): React.Node => {
+    switch (row.type) {
+      case 'header':
+        return (
+          <Box2
```
chrisnojima: should just pull this into a component AssetsHeader or something

shared/wallets/asset/index.js
```diff
@@ -154,3 +152,63 @@ const styles = styleSheetCreate({
...
+            direction="vertical"
+            fullWidth={true}
+            style={{backgroundColor: globalColors.blue5, padding: globalMargins.xtiny}}
```
chrisnojima: lets always use styleSheetCreate in new code

shared/wallets/asset/index.js
```diff
@@ -154,3 +152,63 @@ const styles = styleSheetCreate({
...
+        /*::
+        declare var ifFlowErrorsHereItsCauseYouDidntHandleAllTypesAbove: (a: empty) => any
+        ifFlowErrorsHereItsCauseYouDidntHandleAllTypesAbove(row.type);
+        */
+        throw new Error(`Impossible case encountered: ${row.type}`)
+    }
+  }
+
+  render = () => {
+    const rows = this.props.assets.map(asset => ({
```
chrisnojima: the container should just do this

shared/wallets/asset/index.js
cjb: @chrisnojima Ah, I did this because the storybook and flowtype expect an "expanded" prop to be passed in to Asset that controls whether it's expanded.  If it's just local state inside Asset, should it still accept the prop?  If not, how would we get storybook to show both expanded and non-expanded separately?  Maybe there should be a prop, but it only controls the initial value, and after that it just uses state..?

chrisnojima: maybe best if its always a prop and expanded is controlled by the parent. so the list has a map of the expanded ids and it passes it down and there's a prop it passes down to toggle

cjb: @chrisnojima Hm, I've made all of the suggested changes except this one about where the `expanded` prop comes from.  Do you think it's important to change?  One reason it's awkward is that assets don't actually have IDs coming through from the service to store in our map (perhaps they should?), it's purely an array.

chrisnojima: k we can leave it

PR #12572, opened 2018-06-27
Make sure chat name is always centered on native

buoyad:
- Reworks styling on mobile chat header to two margins (back button + info button) of the same width and center area with chat title. Touchables before:
image
after:
image

- Makes `Badge` play nice with `styleSheetCreate`
- Add `ellipsizeMode` as a mobile-only prop as a passthrough to [this](https://facebook.github.io/react-native/docs/text#ellipsizemode)

r? @keybase/react-hackers


shared/chat/conversation/header-area/normal/index.native.js
```diff
@@ -1,97 +1,79 @@
...
-    <Icon
-      type="iconfont-shh"
-      style={iconCastPlatformStyles(styles.left)}
-      color={shhIconColor}
-      fontSize={shhIconFontSize}
-    />
-  </Box>
+  <Icon
+    type="iconfont-shh"
+    style={{marginLeft: globalMargins.xtiny}}
```
chrisnojima: lets put all styles into the styles object below

shared/chat/conversation/header-area/normal/index.native.js
```diff
@@ -2,96 +2,87 @@
...
+  <Box2 direction="horizontal" style={styles.container}>
+    <ClickableBox onClick={props.onBack} style={styles.leftMargin}>
+      <Icon type="iconfont-arrow-left" fontSize={24} color={globalColors.blue} />
+      {!!props.badgeNumber && <Badge badgeNumber={props.badgeNumber} badgeStyle={styles.badge} />}
+    </ClickableBox>
+    <Box2
+      direction="vertical"
+      style={collapseStyles([
+        styles.contentContainer,
+        !!props.badgeNumber && {paddingLeft: globalMargins.tiny, paddingRight: globalMargins.tiny},
```
chrisnojima: this should move to styles also

PR #12573, opened 2018-06-27
LTPAK: avoid loading keys if team is public and we are not a member

zapu:
*(LoadTeamPlusApplicationKeys)*

This is the first push but maybe we are seeking something different for public teams.

Let's wait for @strib 

cc @maxtaco


strib:
@zapu what do you need from me?

maxtaco:
@strib  -- are public KBFS teams encrypted, and if so with what key?

strib:
>  are public KBFS teams encrypted, and if so with what key?

Do you mean public team TLFs?  They're encrypted with the zero key.  We don't need the team application keys for public teams.

maxtaco:
OK, great, that answers the question.

PR #12588, opened 2018-06-28
Fix ek deletion times

joshblum:
Fixes some early deletion edge case scenarios for ephemeral keys. Depends on https://github.com/keybase/client/pull/12546 and https://github.com/keybase/client/pull/12553, @oconnor663 let's hold off on review until those are merged so i can clean this diff up

### DIFF
--- a/go/ephemeral/device_ek_storage.go
+++ b/go/ephemeral/device_ek_storage.go
@@ -350,7 +350,7 @@ func (s *DeviceEKStorage) DeleteExpired(ctx context.Context, merkleRoot libkb.Me
 		keyMap[generation] = ctime
 	}
 
-	expired = getExpiredGenerations(context.Background(), s.G(), keyMap, now)
+	expired = s.getExpiredGenerations(context.Background(), keyMap, now)
 	epick := libkb.FirstErrorPicker{}
 	for _, generation := range expired {
 		epick.Push(s.delete(ctx, generation))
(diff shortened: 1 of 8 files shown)


go/ephemeral/device_ek_storage.go
```diff
@@ -360,6 +360,68 @@ func (s *DeviceEKStorage) DeleteExpired(ctx context.Context, merkleRoot libkb.Me
...
+	for k := range keyMap {
+		keys = append(keys, k)
+	}
+	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
+
+	// Sort the generations we have so we can walk through them in order.
+	for i, generation := range keys {
+		keyCtime := keyMap[generation].Time()
+
+		expiryOffset1 := libkb.MaxEphemeralKeyStaleness
```
oconnor663: It looks like the reason we split this into offsets 1 & 2 is that we want to independently check the `MaxEphemeralKeyStaleness` bound on each half. If I have that right, please include a comment about it.

go/ephemeral/team_ek_box_storage.go
```diff
@@ -261,8 +261,10 @@ func (s *TeamEKBoxStorage) DeleteExpired(ctx context.Context, teamID keybase1.Te
 		return nil, nil
 	}
 
+	merkleCtime := keybase1.TimeFromSeconds(merkleRoot.Ctime()).Time()
 	for gen, teamEKBox := range teamEKBoxes {
-		if ctimeIsStale(teamEKBox.Metadata.Ctime, merkleRoot) {
+		keyAge := merkleCtime.Sub(teamEKBox.Metadata.Ctime.Time())
+		if keyAge >= libkb.MinEphemeralKeyLifetime {
```
oconnor663: Please add a comment here about why teamEKs never exceed the `MinEphemeralKeyLifetime`.

PR #12589, opened 2018-06-28
Fix hex.Decode bugs

mlsteele:
hex.Decode is a deathtrap

### DIFF
--- a/go/libkb/util.go
+++ b/go/libkb/util.go
@@ -865,3 +865,24 @@ func NoiseXOR(secret [32]byte, noise NoiseBytes) ([]byte, error) {
 func ForceWallClock(t time.Time) time.Time {
 	return t.Round(0)
 }
+
+// Decode decodes src into dst.
+// Errors unless all of:
+// - src is valid hex
+// - src decodes into exactly len(dst) bytes
+func DecodeHexFixed(dst, src []byte) error {
+	// hex.Decode is wrapped because it does not error on short reads and panics on long reads.
+	if len(dst) != hex.DecodedLen(len(src)) {
+		return NewHexWrongLengthError(fmt.Sprintf(
+			"error decoding fixed-length hex: expected %v bytes but got %v", len(dst), hex.DecodedLen(len(src))))
+	}
+	n, err := hex.Decode(dst, src)
+	if err != nil {
+		return err
+	}
+	if n != len(dst) {
+		return NewHexWrongLengthError(fmt.Sprintf(
+			"error decoding fixed-length hex: expected %v bytes but got %v", len(dst), n))
+	}
+	return nil
+}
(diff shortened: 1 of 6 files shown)


mlsteele:
CI was failing because golang changed to.. error out slower
https://golang.org/src/encoding/hex/hex.go?s=2080:2200#L57

mlsteele:
Probable flake
``` 	server_test.go:889: 
			Error Trace:	server_test.go:889
			            				server_test.go:226
			            				server_test.go:780
			Error:      	Not equal: 
			            	expected: "2"
			            	actual  : "1"
			Test:       	TestChatSrvGetInboxNonblockLocalMetadata
	test_logger.go:56: TEST FAILED: TestChatSrvGetInboxNonblockLocalMetadata
```

go/libkb/util.go
```diff
@@ -865,3 +865,27 @@ func NoiseXOR(secret [32]byte, noise NoiseBytes) ([]byte, error) {
...
+	}
+	if len(dst) != hex.DecodedLen(len(src)) {
+		return NewHexWrongLengthError(fmt.Sprintf(
+			"error decoding fixed-length hex: expected %v bytes but got %v", len(dst), hex.DecodedLen(len(src))))
+	}
+	n, err := hex.Decode(dst, src)
+	if err != nil {
+		return err
+	}
+	if n != len(dst) {
```
oconnor663: This should be impossible if `err` is `nil`, right? I'm not against checking it though.

go/libkb/util_test.go
```diff
@@ -151,3 +152,36 @@ func TestForceWallClock(t *testing.T) {
...
+		{"abcd", []byte{171, 205, 0}, "error decoding fixed-length hex: expected 3 bytes but got 2"},
+		{"abcd", []byte{171}, "error decoding fixed-length hex: expected 1 bytes but got 2"},
+	}
+	for i, unit := range units {
+		t.Logf("units[%v]", i)
+		buf := make([]byte, len(unit.dst))
+		err := DecodeHexFixed(buf, []byte(unit.src))
+		if unit.err != "" {
+			require.Error(t, err)
+			require.Equal(t, unit.err, err.Error())
```
oconnor663: Since we're already using a concrete error type in this case, it might be nicer to put the expected/got numbers on it as members and check them, rather than duplicating the error string here?

go/libkb/util.go
mlsteele: https://play.golang.org/p/XyCXAzazfBd

go/libkb/util_test.go
mlsteele: That sounds kind of elegant but this is easy and I don't think anyone will get bit

go/libkb/util.go
oconnor663: My bad!

PR #12590, opened 2018-06-28
Show Timestamps for All Messages

thebearjew:
## Problem

1. **Missing Timestamps on System Messages**
Previously, timestamps were only showing above `MessageTypeText` and `MessageTypeAttachment` in a conversation view. System messages did not receive timestamps and could lead to a scenario where user text message that followed a system message would also not get a timestamp:

2. **Consecutive Attachments By the Same User Were Not Grouped**
When a user sends multiple text messages in a short period of time, we group them together and show a single message "block" with a single avatar. However, this was not done for attachment messages:

consecutive-attachment-messages
 
## Solution

1. Wrap all message type with a `WrapperTimestamp` higher order component that prefixes the timestamp and orange line marker when returning to a thread.
2. Move original wrapper for `MessageTypeText` and `MessageTypeAttachment` to a new `WrapperAuthor` that handle consecutive message types from the same user and avatar positioning.
3. Update `chat/conversation/list-area/normal/index.stories.js` to pseudo-randomly generate timestamps for testing.

**Sequential attachments are grouped together from the same user**
screen shot 2018-06-28 at 9 39 34 am

### DIFF
--- a/shared/chat/conversation/list-area/normal/index.stories.js
+++ b/shared/chat/conversation/list-area/normal/index.stories.js
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react'
 import I from 'immutable'
+import moment from 'moment'
 import {Box2, Text} from '../../../../common-adapters'
 import * as Types from '../../../../constants/types/chat2'
 import {storiesOf, action} from '../../../../stories/storybook'
@@ -8,6 +9,7 @@ import * as PropProviders from '../../../../stories/prop-providers'
 import Thread from '.'
 import * as Message from '../../../../constants/chat2/message'
 import HiddenString from '../../../../util/hidden-string'
+import {formatTimeForMessages} from '../../../../util/timestamp'
 
 // set this to true to play with messages coming in on a timer
 const injectMessages = false && !__STORYSHOT__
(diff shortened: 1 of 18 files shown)


thebearjew:
@keybase/react-hackers should be good to review → Special focus on noticing any performance difference on thread scrolling (desktop)

shared/chat/conversation/messages/wrapper/container-wrapper-author.js
```diff
@@ -53,29 +52,18 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
...
-    (previous.type === 'systemJoined' || previous.type === 'systemLeft')
-  )
+  const oldEnough = Constants.enoughTimeBetweenMessages(message, previous)
 
   const timestamp =
-    stateProps.orangeLineAbove || !previous || oldEnough || previousIsJoinedLeft
-      ? formatTimeForMessages(message.timestamp)
-      : null
-  const includeHeader = !previous || !continuingTextBlock || !!timestamp
+    stateProps.lastPositionExists || !previous || oldEnough ? formatTimeForMessages(message.timestamp) : null
```
chrisnojima: why actually build the timestamp w/ formatTimeForMessages? we just need the boolean right

shared/chat/conversation/list-area/normal/index.stories.js
```diff
@@ -52,6 +55,53 @@ class Rnd {
...
+
+      // Reset threashold and count
+      messagesThreshold = r.next() % 10 + 1
+      generatedCount = 1
+
+      return currentTimestamp
+    }
+
+    if (generatedCount <= messagesThreshold) {
+      generatedCount += 1
```
buoyad: :sheep: `generatedCount++`

shared/chat/conversation/list-area/normal/index.stories.js
```diff
@@ -52,6 +55,53 @@ class Rnd {
...
+      end.add(dayRange, 'days')
+
+      const diff = end.diff(start)
+      // Multiply the epoch time different by some floating point between [0, 1]
+      const newDiff = diff * (r.next() / 2147483647)
+      const newTimestamp = moment(start.valueOf() + newDiff)
+      currentTimestamp = newTimestamp.valueOf()
+
+      // Reset threashold and count
+      messagesThreshold = r.next() % 10 + 1
```
buoyad: You could parameterize the `10` here and `7` on L78 to make it easily configurable. it might also make this fcn more self explanatory.

shared/chat/conversation/list-area/normal/index.stories.js
```diff
@@ -73,6 +123,7 @@ const ordinalToMessage = o => {
 
   const message = Message.makeMessageText({
     text: new HiddenString(String(o) + extra),
+    timestamp: generateTimestamp(),
```
buoyad: cool!

shared/chat/conversation/messages/wrapper/wrapper-timestamp.js
```diff
@@ -0,0 +1,36 @@
+// @flow
```
buoyad: This folder is getting kinda crowded. Do you think this + `wrapper-timestamp-container` could move into a subfolder?

shared/chat/conversation/messages/wrapper/container-wrapper-author.js
buoyad: `SendIndicator` takes the timestamp from `props.message.timestamp`, which is the `number`. But it shouldn't be reaching into the `message` to grab that, so you should pass in `message.timestamp` directly as `timestamp` and use that for `SendIndicator` instead

thebearjew:
@keybase/react-hackers This is good for another review & test

PR #12591, opened 2018-06-28
Integrate showing transactions & layout of wallet page

buoyad:
The wallet page will be a native `SectionList` eventually. I added a desktop port to common adapters and used it to set up the layout of the wallet page. Also makes some changes to the transaction component to support transactions that don't have a display currency. r? @keybase/react-hackers


buoyad:
Affected storyshots are for the transactions, assets, and wallet rows

buoyad:
Hm, I can't repro, when I concatenate the returned asset array to itself in the saga I end up with two identical rows in the GUI for it. How did you lengthen the array?

shared/wallets/wallet/index.js
```diff
@@ -0,0 +1,60 @@
...
+    if (section.title === 'Your assets') {
+      children.push(
+        <Asset accountID={props.accountID} index={item.item} key={`${props.accountID}:${item.item}`} />
+      )
+      if (index !== section.data.length - 1) {
+        // don't put divider after last thing in section
+        children.push(<Divider key={`${props.accountID}:${item.item}:divider`} />)
+      }
+    }
+    // TODO
```
cjb: (what's the TODO here?)

buoyad: to render the transaction rows, I'll remove it in the second PR

PR #12598, opened 2018-06-28
Payment fixes for frontend

patrickxb:
1. sourceType, targetType changed from string to ParticipantType enum
2. include amountValue in PaymentLocal and PaymentDetailsLocal that is like amountDescription but without the + or -.
3. fix bug where target, targetType were empty when sending from keybase -> stellar.

cc @buoyad


patrickxb:
CI failed because of flow...

protocol/avdl/stellar1/local.avdl
```diff
@@ -40,21 +40,29 @@ protocol local {
...
+  }
+
   record PaymentLocal {
     TransactionID id;               // For relay payments, the funding tx.
     TimeMs time;
     PaymentStatus statusSimplified; // we are working on making this correct for relay payments...
     string statusDescription;       // "pending", "completed", "error"
     string statusDetail;
     string amountDescription;       // "+ 1,323.1234567 XLM"
+    string amountValue;             // "1,323.1234567 XLM" (no + or -)
```
mlsteele: This is too much. Can we just have `amountDescription` without the sign and gui can display + or - or neither based on `delta`? @buoyad

protocol/avdl/stellar1/local.avdl
```diff
@@ -40,21 +40,29 @@ protocol local {
...
     string worthCurrency;           // "USD", "CAD", etc. (the symbols can be ambiguous)
 
-    string source;             // keybase username or stellar public key
-    string sourceType;         // "keybase" or "stellar"
-    string target;             // keybase username, stellar public key, or sbs assertion
-    string targetType;         // "keybase", "stellar", or "sbs"
+    string source;              // keybase username or stellar public key
+    ParticipantType sourceType; // "keybase" or "stellar"
+    string target;              // keybase username, stellar public key, or sbs assertion
+    ParticipantType targetType; // "keybase", "stellar", or "sbs"
```
mlsteele: ```// keybase or stellar```
without quotes

protocol/avdl/stellar1/local.avdl
mlsteele: I guess that's not really my objection. It's that we use the type `string` for actual machine readable amounts and the word 'value' sounds very canonical to me. If yall are into the doubled representation then how about `amountDescription` and `amountDescriptionSigned` or something?

buoyad: `amountDescription` sans sign is fine, we can render the + or -.

patrickxb:
Ok, @buoyad amountDescription no longer has +/-.  Please use delta.

PR #12605, opened 2018-06-28
fs: fix scrolling by being honest about fixedHeight

songgao:
also cleanup the divider stuff, and placeholder icon fix.


shared/fs/index.js
```diff
@@ -19,30 +19,34 @@ type FolderProps = {
   routePath: I.List<string>,
 }
 
+export const wrapRow = (rowContent, key: string) => (
```
chrisnojima: don't make helper functions that return a component, just make it a component
```
const WrapRow = ({children, id}) => 
```

shared/fs/index.js
```diff
@@ -19,30 +19,34 @@ type FolderProps = {
...
+    <Divider key="divider" style={stylesDivider} />
+  </Box>
+)
+
 class Files extends React.PureComponent<FolderProps> {
   _rowRenderer = (index: number, item: Types.RowItem) => {
     switch (item.rowType) {
       case 'placeholder':
-        return <Placeholder key={`placeholder:${item.name}`} />
+        return wrapRow(<Placeholder type={item.type} />, `placeholder:${item.name}`)
```
chrisnojima: then
```
<WrapRow id=``><Placeholder ...
```

shared/fs/index.js
songgao: Sure; Does this have something to do with re-rendering?

chrisnojima: no really its just if its a function that takes in a component and returns a component, then it itself is a component. we don't need 2 forms of this

songgao: Got it thanks!

songgao:
@chrisnojima This is ready for another look; thanks!

PR #12615, opened 2018-06-29
Fix ReAddMemberAfterReset so it finds invite superseding membership

zapu:
This PR fixes a bug in ReAddMemberAfterReset which makes the following possible:
1) Have an implicit team "alice,bob".
2) Bob resets (is now `bob%0`).
3) Alice re-adds bob (creates an invite link `bob%0%`. at this point his previous crypto membership is still not-removed).
4) Alice can call ReAddMemberAfterReset again and post pointless link like:
```
CanceledInvites: [bob%0]
NewInvites: { Owner: [bob%0] }
```

Changed `ReAddMemberAfterReset` to look for invites first, and for memberships second. It was the other way round before, and it was always finding the membership first, which resulted in condition `existingUV.EldestSeqno == uv.EldestSeqno` not being hit.

### DIFF
--- a/go/teams/implicit_test.go
+++ b/go/teams/implicit_test.go
@@ -402,6 +402,32 @@ func TestImplicitInvalidLinks(t *testing.T) {
 	}
 }
 
+func TestImpTeamAddInviteWithoutCanceling(t *testing.T) {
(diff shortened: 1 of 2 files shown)


zapu:
Tagging @mlsteele because we've recently went through this change_membership+invite flow. Thanks in advance!

PR #12618, opened 2018-06-29
Don't delete all message types on getting a delete history message

buoyad:
Adds a property to `messagesWereDeleted`, `deletableMessageTypes`, to support deleting a subset of message types. Calls the `getStaticConfig` rpc @mlsteele added and references `deletableByDeleteHistory` while iterating through messages to delete. r? @keybase/react-hackers


shared/actions/chat2/index.js
```diff
@@ -2106,6 +2107,39 @@ function* handleSeeingExplodingMessages(action: Chat2Gen.HandleSeeingExplodingMe
...
+  return Saga.call(RPCChatTypes.localGetStaticConfigRpcPromise)
+}
+
+const loadStaticConfigSuccess = (res: ?RPCChatTypes.StaticConfig) => {
+  if (!res) {
+    // we already had it
+    return
+  }
+  if (!res.deletableByDeleteHistory) {
+    logger.error('chat.loadStaticConfig: got no deletableByDeleteHistory in static config')
```
buoyad: this would have the affect of deleting all message types on getting a delete history until bootstrap happens again. but it should never happen.

shared/constants/chat2/message.js
```diff
@@ -25,6 +25,45 @@ export const getMessageID = (m: RPCChatTypes.UIMessage) => {
   }
 }
 
+// Note flow doesn't type check these values correctly
+// https://github.com/facebook/flow/issues/252
+// so double check these!
+export const serviceMessageTypeToMessageTypes: {
```
chrisnojima: maybe better to make a function which does the switch default check so we can actually update this if that changes

buoyad: nice, that's a lot better. the lack of flow checking on this makes me uneasy

buoyad: hmm. Flow doesn't seem to figure this out. if I switch it to `case <number literal>:` it works, but if I use `case RPCChatTypes.common....<msgType>:` flow complains that numbers 0-13 are incompatible with `empty`. It still buys typing on the return value though, which is nice.

buoyad: https://flow.org/try/#0MYewdgzgLgBAtgQwA4wLwwN4CgYygUwA8oAuGARgBocYEooFgALOfMUmAJmoF8ssoATyT4YAWXwQICAOb4AKsNHpyMAD5d+oSLCEj5IAIIAnYwkFoYACg4SpshUoCUaAHyYsASAgB3AJZQzNZQLti44TDACBCiiEgAdATEJDQRuMb4UACuxmAwAAyp4VEx8MjxdAzMrOwpaeEZ2bkURbgAJvgAZghZADYcreEA9ABUJHX1MB3AvQgZMABuczB+nQBivSA+AKKmIMYQABL4GQCSUBAAwj0xAJogWQAifm3shwhgbb34hr29dtI5IoRBBDAAjEALfBkGxkfBwJBCFyodwfQSDXCrDZbXbGfZHE74c5XG74e5PF5vD5fH5-AEOYGScGQ-A2JwAbgxMBGQyKfD4WG00BgIAA1oyDCYzBZ0LDxJJAY4RMj3NhvP5AkxgqFBiVROQJvVGjk8oV6nquIa0sbmuRBh1un0BpMYKNxlzprN5ktjCt1psdnsDsczhdrlk7g9nq8oO9Pt9fv8FQylKCIVCYRx4YjBCraGB0S6sQHcfiQ0Sw6TydGqfHaUn7EDU8yoWzOS6eXysDwgA

shared/constants/chat2/message.js
```diff
@@ -25,6 +25,67 @@ export const getMessageID = (m: RPCChatTypes.UIMessage) => {
   }
 }
 
+// Map service message types to our message types.
+export const serviceMessageTypeToMessageTypes = (t: RPCChatTypes.MessageType): Array<Types.MessageType> => {
+  // const types = RPCChatTypes.commonMessageType
```
chrisnojima: 🔥

buoyad:
I'll push the comment :fire: and merge once circle passes so it doesn't have to restart

PR #12620, opened 2018-06-29
Whitelist team sigchain link

mlsteele:
This should make it possible for users in this team to load it again. Please review the link in question to make sure you also think it's good.


mlsteele:
tweaked could you look again?

PR #11516, opened 2018-04-24
give some notification that the remote user's attachment upload is st…

zanderz:
…ill pending

I know this stuff is changing, too, but is this the right approach?

PR #12621, opened 2018-06-29
show spinner until chat image arrives

zanderz:
This is the case where the sender uploads fine but the receiver can't reach our servers to get it.
Also requested by @mmaxim  in https://github.com/keybase/client/pull/11516

I don't know how to block specific urls to test on mobile.

image


shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -55,12 +56,16 @@ class ImageAttachment extends React.PureComponent<Props, State> {
...
             {
               height: this.props.height,
               width: this.props.width,
             },
           ])}
         >
+          {!this.state.loaded && (
+             <ProgressIndicator style={{margin: 'auto'}} />
+          )}
           {!!this.props.path && (
```
songgao: If looks like if we do `!this.state.loaded ? ....`, this block can just be part of what's after `:`. Then we don't need `display: none` and can get rid of `opacity` below.

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -55,12 +56,16 @@ class ImageAttachment extends React.PureComponent<Props, State> {
         <Box
           style={collapseStyles([
             styles.loading,
+            !this.state.loaded ? styles.spinner : {},
```
songgao: Hmm I wonder if styles.spinner can just go into `styles.loading`?

chrisnojima: you can just do 
```
!this.state.loaded && styles.spinner
```

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -55,12 +56,16 @@ class ImageAttachment extends React.PureComponent<Props, State> {
...
             {
               height: this.props.height,
               width: this.props.width,
             },
           ])}
         >
+          {!this.state.loaded && (
+             <ProgressIndicator style={{margin: 'auto'}} />
```
chrisnojima: i doubt this works in mobile

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -72,6 +77,7 @@ class ImageAttachment extends React.PureComponent<Props, State> {
                   opacity: this.state.loaded ? 1 : 0,
                   width: this.props.width,
                 },
+                this.state.loaded ? null : {display: 'none'},
```
chrisnojima: also this

zanderz:
thanks guys, I jumped the gun, it doesn't work on mobile, sorry.

zanderz:
spinner still not vertically centered on mobile, but better than nothing.

PR #12626, opened 2018-06-30
Pull loop for prev pointers

joshblum:
In a chat with only exploded messages (and all of the ash lines have been hidden) it's possible for the prev pointer addition to fail. This adds a retry loop to fetch further pages before allowing the message to be sent without prev points (similar to https://github.com/keybase/client/compare/joshblum/prev-fix?expand=1#diff-a0fb8fa7e91c4b2c1548dc6363b637f9R124)

cc @oconnor663

### DIFF
--- a/go/chat/sender.go
+++ b/go/chat/sender.go
@@ -81,57 +81,69 @@ func (s *BlockingSender) addSenderToMessage(msg chat1.MessagePlaintext) (chat1.M
 }
 
 func (s *BlockingSender) addPrevPointersAndCheckConvID(ctx context.Context, msg chat1.MessagePlaintext,
-	conv chat1.Conversation) (chat1.MessagePlaintext, error) {
+	conv chat1.Conversation) (resMsg chat1.MessagePlaintext, err error) {
 
 	// Make sure the caller hasn't already assembled this list. For now, this
 	// should never happen, and we'll return an error just in case we make a
 	// mistake in the future. But if there's some use case in the future where
 	// a caller wants to specify custom prevs, we can relax this.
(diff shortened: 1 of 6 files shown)


joshblum:
added

mmaxim:
Thinking about this some more, would it be easier to just put `Pull` in a mode where it doesn't drop the ashed messages?

joshblum:
we could, but i don't think it solves the problem in general where there is nothing to prev against and we might get hit with this again if we add some other type of filter during post process.

go/chat/sender.go
```diff
@@ -81,57 +81,71 @@ func (s *BlockingSender) addSenderToMessage(msg chat1.MessagePlaintext) (chat1.M
...
+			// the check on numPrev below.
+			hasPrev = len(newPrevsForRegular) > 0 || len(newPrevsForExploding) > 0
+		}
+
+		if hasPrev {
+			break
+		} else if thread.Pagination.Last {
+			return resMsg, fmt.Errorf("Could not find previous messages for prev pointers (of %v)", len(thread.Messages))
+		} else {
+			s.Debug(ctx, "Could not find previous messages for prev pointers (of %v), attempt: %v, retrying", len(thread.Messages), i)
```
oconnor663: It looks like the intention is that after hitting this case 5 times, the loop just exits, and allows the prev list to be empty. I'm fine with that behavior overall, but it's a little hard to tell that that's what it's doing just from reading it. Should we have it print a separate debug message after it gives up or something like that?

joshblum: added a comment and tweaked the log line

PR #12633, opened 2018-07-01
add a wrapper to make any RPC handler cancellable 

mmaxim:
Patch does the following:

1.) Add a new wrapper function `CancellingProtocol` which will convert the handlers in a given protocol to have cancellable contexts passed down.
2.) Add `RPCCanceller` and attach it to `G`.
3.) In `Server.OnLogout` and `AppState.Update` we can cancel all live RPC contexts to kill off the calls.


mmaxim:
bump

go/libkb/rpccanceller.go
```diff
@@ -0,0 +1,51 @@
...
+		liveCtxs: make(map[string]liveContext),
+	}
+}
+
+func (r *RPCCanceller) RegisterContext(ctx context.Context) (context.Context, string) {
+	r.Lock()
+	defer r.Unlock()
+	var lc liveContext
+	lc.ctx, lc.cancelFn = context.WithCancel(ctx)
+	id, _ := RandHexString("", 8)
```
maxtaco: can you make a special type for this, rather than string?

maxtaco: Also, can you use `RandStringB64(3)` as we do in `WithLogTag`?

maxtaco:
Sorry to be unclear, urg, not really worth it, but what I meant was, in `libkb/net_context.go`, make a `func newRandomContextID() string { return RandStringB64(3); }`, and call it from `libkb/net_context.go` and here.

mmaxim:
Ah sorry, I wasn't really sure, but I see now.

PR #12637, opened 2018-07-02
Stellar receive modal: Integration

cjb:
@keybase/react-hackers 

@buoyad, this PR has your tx-2 branch as a base, for reviewing, but it's totally fine for you to merge in first and I can just point the PR at master instead.

An unsatisfying thing about this PR was having to pass navigateAppend through from the main (routed) wallets container to the wallet/header/ component, so that it can be used for the Receive popup.  Let me know if you have other ideas!  I don't expect that making the header itself a route is the right thing.

### DIFF
--- a/shared/wallets/routes.js
+++ b/shared/wallets/routes.js
@@ -1,9 +1,14 @@
 // @flow
 import {makeRouteDefNode, makeLeafTags} from '../route-tree'
 import Container from './container'
+import ReceiveModal from './receive-modal/container'
 
 const routeTree = makeRouteDefNode({
-  children: {},
+  children: {
+    receive: {
+      component: ReceiveModal,
+    },
+  },
   component: Container,
   defaultSelected: '',
   tags: makeLeafTags({}),
--- a/shared/constants/wallets.js
+++ b/shared/constants/wallets.js
@@ -15,9 +15,9 @@ const makeReserve: I.RecordFactory<Types._Reserve> = I.Record({
 })
 
 const makeState: I.RecordFactory<Types._State> = I.Record({
+  accountMap: I.Map(),
   assetsMap: I.Map(),
   paymentsMap: I.Map(),
-  accountMap: I.Map(),
   selectedAccount: Types.noAccountID,
 })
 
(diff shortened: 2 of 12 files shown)


cjb:
(Oh, I see on the link-existing branch @buoyad just imports navigateAppend directly from Route.  I usually like to use the one from ownProps, but maybe that's a better idea here too?)

shared/wallets/routes.js
```diff
@@ -1,9 +1,14 @@
 // @flow
 import {makeRouteDefNode, makeLeafTags} from '../route-tree'
 import Container from './container'
+import ReceiveModal from './receive-modal/container'
 
 const routeTree = makeRouteDefNode({
-  children: {},
+  children: {
+    receive: {
+      component: ReceiveModal,
```
buoyad: The background disappears when this is up, `layerOnTop: !isMobile`

chrisnojima:
this is why we have navigateToInbox / etc for chat

cjb:
@chrisnojima Oh, I'm not sure that helps here.  Our complaint is that we have to thread `navigateAppend` through a few child components, because the component that wants the `navigateAppend` to happen is far from the route-connected component.

If we move to using `putActionIfOnPath`, I think we still have to thread through a few child components, it's just that we thread `routePath` down instead of `navigateAppend`, since that's an input to `putActionIfOnPath`.  So I think it wouldn't reduce any lines of code, just change which prop is being copied around.  Does that make sense?

.. unless we hardcode this `routePath` in the saga instead of supplying it in the action?  But I don't think you were proposing that, and it creates two sources of truth for route paths, which doesn't sound like a good idea.

cjb:
@buoyad Thanks, made the layerOnTop change.

chrisnojima:
I spoke to danny about the routing thing yesterday. I think it's simplest if the saga itself knows how to push and what it can push on top of. Then the logic is, say you want a whitelist of screens it can be on and it can't be pushed on top of itself. The saga can just look at the route state for the accountsTab and pull the last item and decide to push or not. We dont need to plumb the current path all the way through from the components perspective

cjb:
Will go ahead and merge, since the review change asked for and made was a one-liner.

PR #12642, opened 2018-07-02
Add `largerBadgeMinWidthFix` option to fix badge sizing

nathunsmitty:
My previous badge sizing fix actually doesn't work on all badge sizes 😢

This is just a quick fix that adds an option (`largerBadgeMinWidthFix`) for the larger badge size on mobile that I was working with (it seems like all other badges are correctly sized, but please let me know if this is not the case!) I know there's probably a better and more elegant solution to the badge sizing problem, but this is just a quick fix. I'd love to come back later and make this better and welcome all suggestions you may have on how to do that!

Here's what the settings screen looks like with my change:
simulator screen shot - iphone 8 plus - 2018-07-02 at 11 48 31

### DIFF
--- a/shared/common-adapters/badge.js.flow
+++ b/shared/common-adapters/badge.js.flow
@@ -5,7 +5,7 @@ export type Props = {
   badgeNumber: ?number,
   badgeStyle?: any,
   badgeNumberStyle?: Object,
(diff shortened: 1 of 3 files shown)


nathunsmitty:
@keybase/react-hackers let me know what you think!

PR #12644, opened 2018-07-02
Unbox notifications in the background

joshblum:
Boxed messages are attached to both regular and silent notifications now. We always attempt to unbox the silent notification in the background, if the user has enabled plaintext notifications we also display the unboxed message. If we've attached the boxed message to the regular notification we also attempt unboxing when the user selects the notification. the aim is to fill the user's cache as aggressively as possible.


shared/actions/push.js
```diff
@@ -95,16 +98,19 @@ function* pushNotificationSaga(notification: PushGen.NotificationPayload): Saga.
       break
     case 'chat.newmessageSilent_2':
       try {
-        logger.info('Push notification: silent notification received')
-        const membersType: RPCChatTypes.ConversationMembersType =
-          // $ForceType
-          typeof payload.t === 'string' ? parseInt(payload.t) : payload.t
+        logger.info('Push notification: silent notification received, displayPlaintext: ', payload.d)
```
chrisnojima: don't log this

shared/actions/push.js
```diff
@@ -125,6 +131,16 @@ function* pushNotificationSaga(notification: PushGen.NotificationPayload): Saga.
         // ignore it
         break
       }
+      // If a boxed message is attached to the notification, unbox.
+      if (payload.m) {
+        logger.info('Push notification: unboxing notification message')
+        yield Saga.call(RPCChatTypes.localUnboxMobilePushNotificationRpcPromise, {
```
chrisnojima: likely shouldn't do this if no convID or payload

shared/react-native/ios/Keybase/AppDelegate.m
```diff
@@ -221,23 +221,29 @@ - (void)application:(UIApplication *)application didReceiveRemoteNotification:(N
 - (void)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)notification fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler {
   NSString* type = notification[@"type"];
   NSString* body = notification[@"m"];
+  NSLog(@"Remote notification handle for %@ started...", type);
```
chrisnojima: don't log this

shared/actions/push.js
joshblum: done

shared/actions/push.js
joshblum: done

shared/react-native/ios/Keybase/AppDelegate.m
joshblum: done

shared/actions/push.js
```diff
@@ -132,6 +141,16 @@ function* pushNotificationSaga(notification: PushGen.NotificationPayload): Saga.
           logger.error('Push chat notification payload missing conversation ID')
           break
         }
+        // If a boxed message is attached to the notification, unbox.
+        if (payload.m) {
+          logger.info('Push notification: unboxing notification message')
+          yield Saga.call(RPCChatTypes.localUnboxMobilePushNotificationRpcPromise, {
+            convID: convID || '',
```
chrisnojima: these || '' aren't necessary anymore right?

joshblum: i left them in since i was getting some flow errors:

```
Error ------------------------------------------------------------------------------------------- actions/push.js:147:17

Cannot call `Saga.call` because string [1] is incompatible with undefined [2] in property `payload` of the first
argument of type argument `Fn`.

   actions/push.js:147:17
                        v-------------------------------------------------------------------
   147|           yield Saga.call(RPCChatTypes.localUnboxMobilePushNotificationRpcPromise, {
   148|             convID: convID,
   149|             membersType,
   150|             payload: payload.m,
   151|             shouldAck: false,
   152|           })
                  -^

References:
   constants/types/rpc-chat-gen.js:557:76
   557| export type LocalUnboxMobilePushNotificationRpcParam = $ReadOnly<{payload: String, convID: String, membersType: ConversationMembersType, pushIDs?: ?Array<String>, shouldAck: Boolean}>
                                                                                   ^^^^^^ [1]
   constants/types/push.js:10:7
    10|   m?: string,
              ^^^^^^ [2]



Found 3 errors
```

shared/react-native/ios/Keybase/AppDelegate.m
```diff
@@ -223,21 +223,26 @@ - (void)application:(UIApplication *)application didReceiveRemoteNotification:(N
...
-        NSString* pushID = [notification[@"p"] objectAtIndex:0];
-        int badgeCount = [notification[@"b"] intValue];
-        int unixTime = [notification[@"x"] intValue];
-        NSString* soundName = notification[@"s"];
-        PushNotifier* pusher = [[PushNotifier alloc] init];
-        KeybaseHandleBackgroundNotification(convID, membersType, messageID, pushID, badgeCount, unixTime, body, soundName, pusher, &err);
+        NSString* msg = nil;
+        msg = KeybaseUnboxNotification(convID, membersType, body, &err);
+        // If d is set, display the plaintext notification
+        bool displayPlaintext = notification[@"d"];
```
mmaxim: This key also can't be right.

joshblum: changing to `n`

shared/actions/push.js
chrisnojima: this is because flow doesn't know if there's logic inbetween which could have mutated the state of the values you already checked. aka if you did 
```
const {convID, m} = payload
then passed m it'd work
```

chrisnojima:
did you test this with large payloads? there are limits to how large a push payload can be so i wonder if we'll get more messages dropped with this change

joshblum:
we have a threshold below the max limit for sns above which we won't attach the msgBoxed

chrisnojima:
k but did we actually test this out

joshblum:
yes.

joshblum:
thanks!

shared/react-native/ios/Keybase/AppDelegate.m
```diff
@@ -223,21 +223,26 @@ - (void)application:(UIApplication *)application didReceiveRemoteNotification:(N
...
-        int messageID = [notification[@"d"] intValue];
-        NSString* pushID = [notification[@"p"] objectAtIndex:0];
-        int badgeCount = [notification[@"b"] intValue];
-        int unixTime = [notification[@"x"] intValue];
-        NSString* soundName = notification[@"s"];
-        PushNotifier* pusher = [[PushNotifier alloc] init];
-        KeybaseHandleBackgroundNotification(convID, membersType, messageID, pushID, badgeCount, unixTime, body, soundName, pusher, &err);
+        NSString* msg = nil;
+        msg = KeybaseUnboxNotification(convID, membersType, body, &err);
+        // If d is set, display the plaintext notification
```
mmaxim: Fix comment

shared/react-native/ios/Keybase/AppDelegate.m
```diff
@@ -223,21 +223,26 @@ - (void)application:(UIApplication *)application didReceiveRemoteNotification:(N
...
+        bool displayPlaintext = notification[@"n"];
```
mmaxim: Think this needs to be ` [notification[@"m"] boolValue]`

shared/react-native/ios/Keybase/AppDelegate.m
```diff
@@ -223,21 +223,26 @@ - (void)application:(UIApplication *)application didReceiveRemoteNotification:(N
...
+        if (displayPlaintext) {
```
mmaxim: Can we just do this display check in Go just to limit the amount of code in native/JS? I realize it will just be passing in a bool to check for false, but I prefer to have all the logic for this stuff in Go if possible

joshblum: done

PR #12649, opened 2018-07-02
add anyRoleAllowed to nextMerkleRoot searches in team role permission downgrades

xgess:
Currently, when a user gets kicked out of a team, and another client wants to know the first merkle root after that happened, we're using a bit of a broad stroke. We're finding the last update where they don't have permissions to write (or to read). 

The idea here is to allow a client to request whether this is someone being dropped from their status as a reader of the team (i.e. completely kicked out) or as a writer (e.g. Writer -> Reader) but not necessarily removed. More nuance. 

* add the anyRoleAllowed boolean field to FindNextMerkleRootAfterTeamRemovalBySigningKeyArg and related other places
* update the logic on logpoint searching to find the first value of the lower permission AFTER the last value of the higher permission.

### DIFF
--- a/go/teams/chain.go
+++ b/go/teams/chain.go
@@ -131,28 +131,6 @@ func (t TeamSigChainState) GetUserLogPoint(user keybase1.UserVersion) *keybase1.
 	return &tmp
 }
 
-// GetLastChangeLogPointWithPredicates gets the first user logpoint in the series at which
-// the second predicate is true AFTER the last user logpoint at which the first predicate is true.
-func (t TeamSigChainState) GetLastChangeLogPointWithPredicates(user keybase1.UserVersion,
-	f1 func(keybase1.UserLogPoint) bool, f2 func(keybase1.UserLogPoint) bool) *keybase1.UserLogPoint {
-
-	points := t.inner.UserLog[user]
-	if len(points) == 0 {
-		return nil
-	}
-	var closestCandidate int
-	for i := len(points) - 1; i >= 0; i-- {
-		if f2(points[i]) {
-			closestCandidate = i
-		}
-		if f1(points[i]) && closestCandidate != 0 {
-			tmp := points[closestCandidate].DeepCopy()
-			return &tmp
-		}
-	}
-	return nil
-}
-
 func (t TeamSigChainState) GetAdminUserLogPoint(user keybase1.UserVersion) *keybase1.UserLogPoint {
 	ret := t.GetUserLogPoint(user)
 	if ret == nil {
--- a/go/teams/chain.go
+++ b/go/teams/chain.go
@@ -145,7 +145,7 @@ func (t TeamSigChainState) GetLastChangeLogPointWithPredicates(user keybase1.Use
 		if f2(points[i]) {
 			closestCandidate = i
 		}
-		if f1(points[i]) {
+		if f1(points[i]) && closestCandidate != 0 {
 			tmp := points[closestCandidate].DeepCopy()
 			return &tmp
 		}
--- a/go/teams/chain.go
+++ b/go/teams/chain.go
@@ -131,16 +131,22 @@ func (t TeamSigChainState) GetUserLogPoint(user keybase1.UserVersion) *keybase1.
 	return &tmp
 }
 
-// GetLastUserLogPointWithPredicate gets the last user logpoint in the series for which the given
-// predicate is true.
-func (t TeamSigChainState) GetLastUserLogPointWithPredicate(user keybase1.UserVersion, f func(keybase1.UserLogPoint) bool) *keybase1.UserLogPoint {
+// GetLastChangeLogPointWithPredicates gets the first user logpoint in the series at which
+// the second predicate is true AFTER the last user logpoint at which the first predicate is true.
+func (t TeamSigChainState) GetLastChangeLogPointWithPredicates(user keybase1.UserVersion,
+	f1 func(keybase1.UserLogPoint) bool, f2 func(keybase1.UserLogPoint) bool) *keybase1.UserLogPoint {
+
 	points := t.inner.UserLog[user]
 	if len(points) == 0 {
 		return nil
 	}
+	var closestCandidate int
 	for i := len(points) - 1; i >= 0; i-- {
-		if f(points[i]) {
-			tmp := points[i].DeepCopy()
+		if f2(points[i]) {
+			closestCandidate = i
+		}
+		if f1(points[i]) {
+			tmp := points[closestCandidate].DeepCopy()
 			return &tmp
 		}
 	}
--- a/go/teams/member_test.go
+++ b/go/teams/member_test.go
@@ -170,7 +170,7 @@ func TestMemberAddInvalidRole(t *testing.T) {
 // a thin wrapper around libkb.FindNextMerkleRootAfterTeamRemoval. Test that the plumbing works
 // properly. Pass in the values from the libkb inner function, to confirm that this function
 // returns the same.
-func testFindNextMerkleRootAfterRemoval(t *testing.T, tc libkb.TestContext, user *kbtest.FakeUser, id keybase1.TeamID, wasReader bool, seqno keybase1.Seqno) {
+func testFindNextMerkleRootAfterRemoval(t *testing.T, tc libkb.TestContext, user *kbtest.FakeUser, id keybase1.TeamID, anyRoleAllowed bool, seqno keybase1.Seqno) {
 	m := libkb.NewMetaContextForTest(tc)
 	upak, _, err := tc.G.GetUPAKLoader().LoadV2(libkb.NewLoadUserArgWithMetaContext(m).WithUID(user.GetUID()))
 	require.NoError(t, err)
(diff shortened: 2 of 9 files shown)


go/teams/chain.go
```diff
@@ -131,16 +131,22 @@ func (t TeamSigChainState) GetUserLogPoint(user keybase1.UserVersion) *keybase1.
...
 		return nil
 	}
+	var closestCandidate int
 	for i := len(points) - 1; i >= 0; i-- {
-		if f(points[i]) {
-			tmp := points[i].DeepCopy()
+		if f2(points[i]) {
+			closestCandidate = i
+		}
+		if f1(points[i]) {
```
mmaxim: This function is a little weird, what happens if I hit `f1` without `f2` ever firing?

xgess: huh. yeah i guess it should give the same `nil` response. i'll update that.

xgess: just kidding. i think it should keep going until it finds the switch it's looking for.

strib:
I didn't review the actual code, but let me see if I understand the interface:
* The caller should set `isWriter` to `true` if they want to find the merkle root for just after the last point in time the specified user had write permissions in the team.  If the user was never a writer, the RPC will return an error. (Returning an error is important, as the previous version of the code didn't do that as far as I know.)
* If `isWriter` is `false`, then the caller wants to find the merkle root for just after the last point in time the specified user was a member of the team in any capacity.  If the user was never a member, the RPC will return an error.

Is that right?

If that's right, the name `isWriter` is a bit confusing to me, since "is" implies present tense.  Also, it would be nice if the default value for the field matched the current behavior without that value set, so that I don't have to update KBFS immediately with the new param.  So can we change the param, and the corresponding logic, to something like `anyRoleAllowed bool` or `wasReader bool`?

strib:
(Also, thanks for taking this on @xgess!)

xgess:
@strib your read on the behavior is correct, and it does return a `libkb.NotFoundError` error if the permissions were never there (or if the downgrade isn't found). i didn't change that actually, so it should behave that way currently. or it will continue to behave however it did before if the permission change wasn't found. 

past-tense and same-default-behavior requests make sense. i guess that would make it `wasntWriter`. can't say i love that (not at all opposite that its opposite is `wasntReader` as opposed to `wasWriter`). and the opposite of `wasReader` being `wasWriter` is also less obvious to me as than vice versa. i'll think on it for another couple minutes and probably just go with `wasReader`. thanks!

mmaxim:
Thanks @xgess! Let's wait for confirmation from @strib before merging.

strib:
I am nitpicking here, and I don't mind if you merge as is, but `wasReader` is still ambiguous because writers are readers too.  That's why I had suggested `anyRoleAllowed` as being a bit more explicit.

Cool re: error, I may have misunderstood what max told me about its behavior.

go/teams/chain.go
```diff
@@ -131,16 +131,22 @@ func (t TeamSigChainState) GetUserLogPoint(user keybase1.UserVersion) *keybase1.
 	return &tmp
 }
 
-// GetLastUserLogPointWithPredicate gets the last user logpoint in the series for which the given
-// predicate is true.
-func (t TeamSigChainState) GetLastUserLogPointWithPredicate(user keybase1.UserVersion, f func(keybase1.UserLogPoint) bool) *keybase1.UserLogPoint {
+// GetLastChangeLogPointWithPredicates gets the first user logpoint in the series at which
+// the second predicate is true AFTER the last user logpoint at which the first predicate is true.
+func (t TeamSigChainState) GetLastChangeLogPointWithPredicates(user keybase1.UserVersion,
```
maxtaco: Do you think it's possible do to this with one predicate that stores some state?  Otherwise, maybe we should get rid of this function and just implement what we want directly.  It doesn't seem like we're ever going to have a second use case for this particular function.

go/teams/service_helper.go
```diff
@@ -1582,9 +1582,25 @@ func FindNextMerkleRootAfterRemoval(mctx libkb.MetaContext, arg keybase1.FindNex
...
 
 	uv := vers.ToUserVersion()
-	logPoint := team.chain().GetLastUserLogPointWithPredicate(uv, func(p keybase1.UserLogPoint) bool {
-		return p.Role == keybase1.TeamRole_NONE || p.Role == keybase1.TeamRole_READER
-	})
+
+	var logPoint *keybase1.UserLogPoint
+	if arg.AnyRoleAllowed {
+		reader := func(p keybase1.UserLogPoint) bool {
+			return p.Role == keybase1.TeamRole_READER || p.Role == keybase1.TeamRole_WRITER || p.Role == keybase1.TeamRole_ADMIN || p.Role == keybase1.TeamRole_OWNER
```
maxtaco: `p.Role.IsReaderOrAbove()`

go/teams/service_helper.go
```diff
@@ -1582,9 +1582,25 @@ func FindNextMerkleRootAfterRemoval(mctx libkb.MetaContext, arg keybase1.FindNex
...
+		}
+		notReader := func(p keybase1.UserLogPoint) bool {
```
maxtaco: `!p.Role.IsReaderOrAbove()`

go/teams/service_helper.go
```diff
@@ -1582,9 +1582,25 @@ func FindNextMerkleRootAfterRemoval(mctx libkb.MetaContext, arg keybase1.FindNex
...
+			return p.Role == keybase1.TeamRole_NONE
+		}
+		logPoint = team.chain().GetLastChangeLogPointWithPredicates(uv, reader, notReader)
+	} else {
+		writer := func(p keybase1.UserLogPoint) bool {
```
maxtaco: `p.Role.IsWriterOrAbove()`

go/teams/service_helper.go
```diff
@@ -1582,9 +1582,25 @@ func FindNextMerkleRootAfterRemoval(mctx libkb.MetaContext, arg keybase1.FindNex
...
+			return p.Role == keybase1.TeamRole_WRITER || p.Role == keybase1.TeamRole_ADMIN || p.Role == keybase1.TeamRole_OWNER
+		}
+		notWriter := func(p keybase1.UserLogPoint) bool {
```
maxtaco: `!p.Role.IsWriterOrAbove()`

go/teams/chain.go
maxtaco: Seems like from above, that `f1(x) = !f2(x)` in all use cases, so we can simplify this...

xgess: ahh yes yes. i wasn't actually sure if this would be strictly true based on underlying data and non-obvious uses. but yes, it definitely seems to be the case if this usecase is executed in isolate. so... lovely! i will simplify this.

go/teams/service_helper.go
```diff
@@ -1582,10 +1582,23 @@ func FindNextMerkleRootAfterRemoval(mctx libkb.MetaContext, arg keybase1.FindNex
...
+		}
+		return !p.Role.IsWriterOrAbove()
+	}
+	var earliestDemotion int
+	var logPoint keybase1.UserLogPoint
+	for i := len(logPoints) - 1; i >= 0; i-- {
+		if demotionPredicate(logPoints[i]) {
+			earliestDemotion = i
+		} else if earliestDemotion != 0 {
+			logPoint = logPoints[earliestDemotion].DeepCopy()
```
maxtaco: hmm, wait a second... What if AnyRoleAllowed is false and the sequence is:

Link 1: Writer
Link 2: Reader
Link 3: Writer
Link 4: Reader

Won't this return Link 2?  It should return link 4.

xgess: oh blerg. you're right. i'm not exiting out of the loop once the `logPoint` is set. :/

xgess: updated the test and added the `break`

go/teams/service_helper.go
```diff
@@ -1582,10 +1582,24 @@ func FindNextMerkleRootAfterRemoval(mctx libkb.MetaContext, arg keybase1.FindNex
...
+			break
+		}
+	}
+	if &logPoint == nil {
```
maxtaco: oh i didn't see it before, this will never return false.

go/teams/member_test.go
```diff
@@ -262,26 +264,41 @@ func TestMemberRemove(t *testing.T) {
...
+	assertRole(tc, name, other.Username, keybase1.TeamRole_WRITER)
 
+	if err := SetRoleReader(context.TODO(), tc.G, name, other.Username); err != nil {
+		t.Fatal(err)
+	}
 	assertRole(tc, name, owner.Username, keybase1.TeamRole_OWNER)
 	assertRole(tc, name, other.Username, keybase1.TeamRole_READER)
+
+	teamID, seqno := pollForNextMerkleRootAfterRemovalViaLibkb(t, tc, other, name)
+	testFindNextMerkleRootAfterRemoval(t, tc, other, teamID, anyRoleAllowed, seqno)
```
maxtaco: Is there any way to test that you're getting the last demotion and not the first demotion?

go/teams/service_helper.go
xgess: huh. that's a little tricky, but i guess it makes sense. looks like i can't compare logPoint with an empty keybase1.UserLogPoint (because `struct containing keybase1.SignatureMetadata cannot be compared`) so im going to keep it as a pointer the whole time. is there another way to do this?

maxtaco: In somecases, we can put an `IsNil()` around these XDR structs, to distinguish between an unset variable and one that's been populated. In this case, we likely don't have that, so a pointer is the better option.

go/teams/member_test.go
xgess: yeah - is this not doing that? `pollForNextMerkleRootAfterRemovalViaLibkb` essentially just gets the current sequence number.

go/teams/service_helper.go
xgess: code fix and test are imminent

go/teams/member_test.go
maxtaco: ok great, so this test failed before the `break` was inserted into the loop?

xgess: yes.

go/teams/service_helper.go
xgess: cool. thanks.

go/teams/member_test.go
maxtaco: 👍

PR #12656, opened 2018-07-02
Slow merkle down for a test

mlsteele:
`TestTeamSignedByRevokedDevice` now acts as a regression test for CORE-8233. It tells merkled to slow that team down, which if the server does not have https://github.com/keybase/keybase/pull/2639 causes the error `ERROR: proof error for proof 'team link before user key revocation'` on my machine. CI may have different timing so idk if it's actually a regression test on CI.


mlsteele:
Yep it needs https://github.com/keybase/keybase/pull/2639

PR #12657, opened 2018-07-02
[KBFS-3055] Folder Header Chat Buttons

jzila:


shared/fs/header/container.js
```diff
@@ -0,0 +1,45 @@
...
+    ...Util.tlfToParticipantsOrTeamname(Types.pathToString(path)),
+  })),
+})
+
+const mergeProps = ({_kbfsEnabled}, {_onBack, onChat, _openInFileUI, _openFinderPopup}, {path}) => {
+  const elems = Types.getPathElements(path)
+  return {
+    path,
+    title: elems.length > 1 ? elems[elems.length - 1] : 'Keybase Files',
+    openInFileUI: isMobile
```
chrisnojima: you could just do this check in mapDispatchToProps so its not so nested here

shared/fs/header/header.native.js
```diff
@@ -23,6 +24,20 @@ const Header = ({title, path, onBack}: Props) => (
       <Box style={stylesAddNewBox}>
         <AddNew path={path} style={stylesAddNew} />
       </Box>
+      {onChat && (
+        <Box style={stylesAddNewBox}>
+          <Icon
+            type="iconfont-chat"
+            style={{
+              marginLeft: globalMargins.tiny,
```
chrisnojima: you don't have to make this change now but we do want to move to using styleSheetCreate everywhere. I'd advocate doing that kind of cleanup when revisiting components. again you don't need to do this in this pr but in general likely a good practice

shared/constants/types/fs.js
```diff
@@ -237,6 +237,10 @@ export const getVisibilityFromElems = (elems: Array<string>) => {
       return null
   }
 }
+export const getFolderNameFromElems = (elems: Array<string>): ?string => {
+  if (elems.length < 3 || !elems[2]) return null
+  return elems[2]
+}
```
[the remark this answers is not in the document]
jzila: Nope, I'll remove. Good catch!

PR #12658, opened 2018-07-03
fix Android upload banner and make animation right

songgao:


shared/fs/footer/upload-container.js
```diff
@@ -1,16 +1,42 @@
...
 import Upload, {type UploadProps} from './upload'
 import {formatDurationFromNowTo} from '../../util/timestamp'
 
 const mapStateToProps = (state: TypedState) => ({
   _uploads: state.fs.uploads,
 })
 
-const mergeProps = ({_uploads}, dispatchProps, ownProps) =>
+// NOTE flip this to show a button to debug the upload banner animations.
+const enableDebugUploadBanner = true
```
chrisnojima: 🔥

shared/fs/footer/upload-container.js
```diff
@@ -1,16 +1,42 @@
...
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  debugToggleShow: !enableDebugUploadBanner
+    ? undefined
+    : (() => {
```
chrisnojima: lets not put this in here inline. just make a closure up top. and likely wrap this in a if (__DEV__)

shared/fs/footer/upload-container.js
```diff
@@ -1,16 +1,42 @@
...
+          tmp = !tmp
+        }
+      })(),
+})
+
+const mergeProps = ({_uploads}, {debugToggleShow}) =>
   ({
     files: _uploads.syncingPaths.merge(_uploads.writingToJournal).size,
-    timeLeft: formatDurationFromNowTo(_uploads.endEstimate),
+    timeLeft: '32 s' || formatDurationFromNowTo(_uploads.endEstimate),
```
chrisnojima: i assume this is debug leftover? 🔥

shared/fs/footer/upload.native.js
```diff
@@ -1,66 +1,145 @@
...
+      return
+    }
+    this.animations[animation].stop()
+    this.animations[animation] = null
+  }
+
+  enter() {
+    this.setState({mounted: true})
+    this.stopAnimation('out')
+    this.startAnimationIn()
```
chrisnojima: don't you need to stop in and loop also before starting them again?

shared/fs/footer/upload.native.js
```diff
@@ -1,66 +1,145 @@
...
+    this.startAnimationLoop()
+  }
+
+  exit() {
+    this.stopAnimation('in')
+    this.startAnimationOut(() => {
+      this.stopAnimation('loop')
+      this.setState({mounted: false})
```
chrisnojima: i don't believe this is safe

songgao: Could you explain why this is not safe? It's in the callback that's called when the animation has finished cleanly.

chrisnojima: any asynchronous callback can't safely call setState as the component could be unmounted. you either need to bookkeep the mounting or cancel any calls like this while unmounting

songgao: I see. What's the consequences of calling it in async and race happens?

In this case, I need this to determine whether to render the banner or not. If the whole component is unmounted, it doesn't really matter. Also, the component is always mounted as long as user is in the Files tab.

By bookkeeping, do you mean I should have a class variable that's set to `true` in `componentWillUnmount`, and only call this `setState` if that variable is `false`?

chrisnojima: react will give you warnings telling you its unsafe to do this. this is just a general rule. so if you do
```
setTimeout(() => this.setState({foo: 'bar'}), 10000)
```
that's unsafe.

yes, by bookkeeping i mean either

```
this.timeoutID = setTimeout(() => this.setState({foo: 'bar'}), 10000)

componentWillUnmount() {
clearTimeout(this.timeoutID)
}

// or

_mounted = true
componentWillUnmount () {
this._mounted = false
}

componentDidMount () {
this._mounted = true
}

setTimeout(() => this._mounted && this.setState({foo: 'bar'}), 10000)
```

PR #12664, opened 2018-07-03
make sure to always send a response from sync to notify clients

mmaxim:
We run a variety of filters on the conversations the server sends down to sync, most notably we do not send up notifications for non-selected big team channels. If those were the only thing to have changed, then we would just send nothing and the frontend blue loading bar could get stuck. Patch does the following:

1.) If we have 0 conversations to sync, then just send up a `CURRENT` response, even if the server said it was incremental.
2.) If we have at least 1 conversation, then send a notification on each `TopicType` channel. This will make it so that it is possible to receive an `INCREMENTAL` response with nothing in it. cc @chrisnojima  hopefully the JS is cool with this.
3.) Only filter big team channels out of the sync that are of type `CHAT`, and let other topic types through.

### DIFF
--- a/go/chat/sender_test.go
+++ b/go/chat/sender_test.go
@@ -60,11 +60,15 @@ func (n *chatListener) ChatThreadsStale(uid keybase1.UID, updates []chat1.Conver
 		panic("timeout on the threads stale channel")
 	}
 }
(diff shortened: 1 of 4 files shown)


chrisnojima:
should be fine, js protects against this case already
https://github.com/keybase/client/blob/9a8f30bf3cf3b67e2513383ade173aa7802c7bf4/shared/actions/chat2/index.js#L416

PR #12665, opened 2018-07-03
Optionally add extended status in log send

joshblum:
Right now the mobile status is pretty limited compared to desktop reports. The extended status is especially helpful for debugging deviceEK state since we log the filenames stored on disk.

- moves `GetExtendedStatus` to `libkb` and merges the result with the json string the frontend gives us
- adds library https://github.com/fatih/structs to simplify the json merging process (unless there is an easier way of doing this?)

an example log with these changes is here: https://keybase.io/_/admin/logdump_viewer?360376538#9d2deedf4facfd1e9391271c

### DIFF
--- a/go/libkb/log_send.go
+++ b/go/libkb/log_send.go
@@ -7,6 +7,7 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -21,6 +22,7 @@ import (
 
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/client/go/protocol/keybase1"
+	"golang.org/x/net/context"
 )
 
 // Logs is the struct to specify the path of log files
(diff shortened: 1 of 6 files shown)


go/libkb/status.go
```diff
@@ -35,3 +37,94 @@ func GetCurrentStatus(ctx context.Context, g *GlobalContext) (res CurrentStatus,
...
+	for i, u := range all {
+		p[i] = u.String()
+	}
+	res.ProvisionedUsernames = p
+	res.PlatformInfo = getPlatformInfo()
+	res.DefaultDeviceID = g.Env.GetDeviceID()
+	res.RememberPassphrase = g.Env.RememberPassphrase()
+	// DeviceEK status, can be nil if user is logged out
+	deviceEKStorage := g.GetDeviceEKStorage()
+	if deviceEKStorage != nil {
```
joshblum: function copied verbatim from `service/config` except we now check if `devicekEKStorage` is present since it is cleared from `g` on logout

go/libkb/log_send.go
```diff
@@ -500,3 +507,26 @@ func (l *LogSendContext) LogSend(statusJSON, feedback string, sendLogs bool, num
...
+
+// mergeExtendedStatus adds the extended status to the given status json blob.
+// If any errors occur the original status is returned unmodified.
+func (l *LogSendContext) mergeExtendedStatus(status string) string {
+	var statusObj map[string]interface{}
+	if err := json.Unmarshal([]byte(status), &statusObj); err != nil {
+		return status
+	}
+
+	extStatus, err := GetExtendedStatus(context.TODO(), l.G())
```
mmaxim: As discussed in chat, I think we can do this simpler without the need of this library.

joshblum: done

joshblum:
new log: https://keybase.io/_/admin/logdump_viewer?771470706#e56485ae2e17d66ec9eae91c

PR #12667, opened 2018-07-03
Integrate link existing account dialog

buoyad:
r? @keybase/react-hackers


mlsteele:
@patrickxb avdl, frontend{_test}.go

protocol/avdl/stellar1/local.avdl
```diff
@@ -117,6 +117,12 @@ protocol local {
   }
   array<CurrencyLocal> getDisplayCurrenciesLocal(int sessionID);
 
+  void validateAccountIDLocal(int sessionID, AccountID accountID);
+  void validateSecretKeyLocal(int sessionID, SecretKey secretKey);
+  // Validates a name for a new account. Checks if any accounts have the same name.
+  // If there is an error getting account names, returns true.
+  void validateAccountNameLocal(int sessionID, string name);
```
patrickxb: shouldn't these return `boolean`?

mlsteele: They return an RPC error if the input failed validation. There is no "something went unexpectedly wrong" value.

patrickxb: "If there is an error getting account names, returns true."

mlsteele: but that comment was wrong

shared/actions/wallets.js
```diff
@@ -41,12 +44,64 @@ const loadPaymentsSuccess = (res: RPCTypes.PaymentsPageLocal, action: WalletsGen
...
+}
+
 function* walletsSaga(): Saga.SagaGenerator<any, any> {
   yield Saga.safeTakeEveryPure(WalletsGen.loadAccounts, loadAccounts, loadAccountsSuccess)
   yield Saga.safeTakeEveryPure(WalletsGen.loadAssets, loadAssets, loadAssetsSuccess)
   yield Saga.safeTakeEveryPure(WalletsGen.loadPayments, loadPayments, loadPaymentsSuccess)
   yield Saga.safeTakeEveryPure(WalletsGen.selectAccount, loadAssets, loadAssetsSuccess)
   yield Saga.safeTakeEveryPure(WalletsGen.selectAccount, loadPayments, loadPaymentsSuccess)
+  yield Saga.safeTakeEveryPurePromise(
+    action => action.type === WalletsGen.linkExistingAccount && !action.error,
```
chrisnojima: i'm not necessarily against this form but it might be useful to just have the handler check for error, so as the acceptable types grows that check can be there instead of these patterns getting longer

shared/actions/wallets.js
```diff
@@ -41,12 +44,64 @@ const loadPaymentsSuccess = (res: RPCTypes.PaymentsPageLocal, action: WalletsGen
...
+const linkExistingAccount = (state: TypedState, action: WalletsGen.LinkExistingAccountPayload) => {
+  const {name, secretKey} = action.payload
+  return RPCTypes.localLinkNewWalletAccountLocalRpcPromise(
+    {
+      name,
+      secretKey: secretKey.stringValue(),
+    },
+    Constants.linkExistingWaitingKey
+  )
+    .then(accountID => [
```
chrisnojima: this might not work after my change to util/saga thats in pr. This thing is supposed to return a single thing to yield, so returning an array should give you a warning, instead we should be using sequentially

shared/actions/wallets.js
```diff
@@ -41,12 +44,64 @@ const loadPaymentsSuccess = (res: RPCTypes.PaymentsPageLocal, action: WalletsGen
...
+      WalletsGen.createLoadAccounts(),
+      WalletsGen.createSelectAccount({accountID: Types.stringToAccountID(accountID), show: true}),
```
chrisnojima: also this is the pattern we want to avoid. When this thing is done it should emit some kind of action, and as a side effect these things should happen, or maybe one happens, and as a side effect the other things happen etc.
createLinkExistingAccountError shouldn't be necessary as the reducer for the action that is emitted by this should handle it in the reducer

shared/actions/wallets.js
```diff
@@ -41,12 +44,64 @@ const loadPaymentsSuccess = (res: RPCTypes.PaymentsPageLocal, action: WalletsGen
...
+      WalletsGen.createLinkExistingAccountError({error: '', name, secretKey}),
+    ])
+    .catch(err => WalletsGen.createLinkExistingAccountError({error: err.desc, name, secretKey}))
+}
+
+const validateAccountName = (state: TypedState, action: WalletsGen.ValidateAccountNamePayload) => {
+  const {name} = action.payload
+  return RPCTypes.localValidateAccountNameLocalRpcPromise({name})
+    .then(() => WalletsGen.createValidateAccountNameError({error: '', name}))
```
chrisnojima: i think this should be a different action. this handles the validateAccountName action, and happens to not be called on error but really i think there's 2 semantic actions
1. validateAccountName (aka i want it validated)
2. validatedAccountName (aka i'm done validating)

shared/actions/json/wallets.json
```diff
@@ -1,37 +1,70 @@
...
+      "show?": "boolean"
+    },
+    "linkExistingAccount": {
+      "_description": "Link an existing Stellar account with this Keybase user.",
+      "name": "string",
+      "secretKey": "HiddenString",
+      "canError": {
+        "name": "string",
+        "secretKey": "HiddenString",
+        "error": "string"
```
chrisnojima: can the error be sensitive

shared/reducers/wallets.js
```diff
@@ -19,6 +19,29 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
...
+            .set('accountNameValidationState', action.payload.error ? 'error' : 'valid')
+        : state.set('accountNameValidationState', 'waiting')
+    case WalletsGen.validateSecretKey:
+      return action.error
+        ? state
+            .set('secretKeyError', action.payload.error)
+            .set('secretKeyValidationState', action.payload.error ? 'error' : 'valid')
+        : state.set('secretKeyValidationState', 'waiting')
+    case WalletsGen.clearErrors:
+      return state
```
chrisnojima: likely better to just use merge

shared/util/saga.js
```diff
@@ -127,7 +127,7 @@ function safeTakeEveryPurePromise<A, RA>(
     const state: TypedState = yield select()
     const toPut = yield call(f, state, action)
     if (toPut) {
-      yield put(toPut)
+      yield Array.isArray(toPut) ? sequentially(toPut.map(p => put(p))) : put(toPut)
```
chrisnojima: i've changed how this works, lets discuss

shared/wallets/link-existing/container.js
```diff
@@ -0,0 +1,40 @@
...
+      })
+    )
+  },
+  onCheckName: (name: string) => {
+    dispatch(WalletsGen.createValidateAccountName({name}))
+  },
+  onClearErrors: () => dispatch(WalletsGen.createClearErrors()),
+  onDone: (sk: string, name: string) => {
+    dispatch(WalletsGen.createClearErrors())
+    dispatch(WalletsGen.createLinkExistingAccount({name, secretKey: new HiddenString(sk)}))
```
chrisnojima: really this should clear the errors, not this component

shared/util/saga.js
chrisnojima: actually i changed simple, not promise really ,but still it doing a single put was on purpose

shared/actions/wallets.js
buoyad: I started out doing that with `safeTakeEveryPure` and `actionCreatorsWithResult`, etc. The issue there was that `actionCreatorsWithResult/Error` always ran and made a `createLinkExistingAccountError` (either to set or clear), which resulted in an infinite loop if I forgot to check for the error in those handlers too. With this form I can change it to check in the one handler :+1:

shared/actions/json/wallets.json
buoyad: No, also @mlsteele says the service will log them so I'll log them here too.

shared/reducers/wallets.js
buoyad: :+1:

buoyad:
Ok @chrisnojima I added the post-actions as separate ones and moved more towards the side-effect model. At first I was checking for `action.error` in a bunch of the `safeTakeEveryPure` success handlers, so I switched them to use `safeTakeEveryPurePromise` so we can bail out in one place (cc @cjb). 

I couldn't completely get rid of the `clearErrors` action because we need to call it explicitly in some places, specifically when unmounting the dialog and when switching from `enter secret key` -> `enter name`.  PTAL

shared/reducers/wallets.js
```diff
@@ -19,6 +19,47 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
...
+      })
+    case WalletsGen.clearErrors:
+      return state.merge({
+        accountNameError: '',
+        accountNameValidationState: 'none',
+        linkExistingAccountError: '',
+        secretKeyError: '',
+        secretKeyValidationState: 'none',
+      })
+    case WalletsGen.linkExistingAccount:
```
chrisnojima: could make this a fallthrough w/ clearErrors

shared/reducers/wallets.js
```diff
@@ -19,6 +19,47 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
       return state.setIn(['paymentsMap', action.payload.accountID], I.List(action.payload.payments))
     case WalletsGen.selectAccount:
       return state.set('selectedAccount', action.payload.accountID)
+    case WalletsGen.validateAccountName:
```
chrisnojima: to be super safe we should likely validate the params are still good, lets chat

shared/reducers/wallets.js
buoyad: It ended up being a saga-only action

shared/reducers/wallets.js
buoyad: 👍

PR #12668, opened 2018-07-03
raise max size of chat messages in dev channels to 1MB

mmaxim:
cc @strib


joshblum:
can you add a test case with the max plaintext value so we know the boxed max value is good

joshblum:
otherwise lgtm

PR #12673, opened 2018-07-03
New Badge component

nathunsmitty:
Adds a new badge component, `Badge2`, that provides a little nicer API and promises better styling than the original `Badge`. The eventual plan is for `Badge2` to supersede `Badge`, but @chrisnojima suggested that we start small and replace it one page at a time.

Usage of `Badge2` looks as follows:
```js
<Badge2 radius={11} topBottomPadding={0} badgeStyle={badgeStyle} badgeNumber={badgeNumber} />
```

The props it supports are:
```js
export type Badge2Props = {
  badgeNumber: number, // same as Badge
  radius: number, // The radius of the badge. (Default: 12) 
  // This determines the height, minimum width, and border radius of the badge.
  // If the badge's number is small enough, the badge will be a circle with that radius.
  // If it is not, the badge with be a rounded rectangle with that border radius and a height of 2*radius.
  fontSize: number, // The size of the font of the number (Default: 11) 
  leftRightPadding: number, // The left and right padding of the badge. (Default: 6)  
  // Padding is included in the radius of the badge.
  topBottomPadding: number, // The top and bottom padding of the badge. (Default: 4) 
  badgeStyle?: any, // A style object to override any of the badge's styles.
  badgeNumberStyle?: Object, // A style object to override any of the badge's number's styles.
}
```

The defaults were chosen to closely match the default styles of `Badge`.

A screenshot of it in action!
simulator screen shot - iphone 8 plus - 2018-07-03 at 16 05 48

A couple limitations as of right now:
- Badge2 is only implemented for mobile.
- Badge2 is not supported on the storybook currently.
- Badge2 is currently only in use on the mobile nav page.

Would especially love feedback on either the styling or interface of `Badge2` so lemme know what you think @keybase/react-hackers!


shared/common-adapters/badge.native.js
```diff
@@ -48,4 +48,49 @@ const styles = styleSheetCreate({
...
+  static defaultProps = {fontSize: 11, radius: 12, leftRightPadding: 6, topBottomPadding: 4}
+
+  render() {
+    return (
+      <Box
+        style={collapseStyles([
+          badge2Styles.badge,
+          {
+            borderRadius: this.props.radius,
+            height: this.props.radius * 2,
```
chrisnojima: i think a prop of height is clearer than a radius. a designer will know what height means but radius is more confusing

chrisnojima: i don't think we need leftRightPadding or topBottomPadding to be props. can't we drive this with just the fontSize and height?

shared/common-adapters/badge.native.js
```diff
@@ -48,4 +48,49 @@ const styles = styleSheetCreate({
...
+          },
+          this.props.badgeStyle,
+        ])}
+      >
+        <Text
+          type="BodyTinySemibold"
+          style={collapseStyles([
+            badge2Styles.text,
+            this.props.badgeNumberStyle,
+            {fontSize: this.props.fontSize, lineHeight: this.props.fontSize + 5},
```
chrisnojima: what is this 5 about

shared/common-adapters/badge.js.flow
```diff
@@ -1,11 +1,31 @@
...
 
-declare export default class Badge extends Component<Props> {}
+export type Badge2Props = {
+  badgeNumber: number,
+  radius: number,
+  fontSize: number,
+  leftRightPadding: number,
+  topBottomPadding: number,
+  badgeStyle?: any,
+  badgeNumberStyle?: Object,
```
chrisnojima: these should likely be StylesCrossPlatform

shared/common-adapters/badge.native.js
chrisnojima: i hope 99% of the usage of this will be something like
`<Badge2 fontSize={14} />`
where radius is calculated automatically and can be overridden if people want more control

shared/common-adapters/badge.native.js
nathunsmitty: In all of our text styles, [lineHeight is exactly + 5 of font size](https://github.com/keybase/client/blob/nathan/DESKTOP-7129-additional-badge-fixes/shared/common-adapters/text.meta.native.js#L25). I needed a way to scale line height with font size and that was the first thing I came up with – although now that I'm looking at it again, I don't like how hacky it is. Perhaps a better solution would be to extract the `fontSizeToSizeStyle` function into a utility file and call it here?

shared/common-adapters/badge.native.js
nathunsmitty: If a user is just defining `fontSize`, does that mean the radius would be calculated as `radius = lineHeight + topPadding + bottomPadding`?

shared/common-adapters/badge.native.js
chrisnojima: does just aligning it center just fix this and you can ignore lineheight?

shared/common-adapters/badge.native.js
chrisnojima: i mean, i should be able to just set fontsize and get an appropriate padding w/o overriding it. whatever that is, we'd have to play with it. maybe it makes sense for it to be fixed, maybe it scales up based on the fontsize, i'd have to see it

shared/common-adapters/badge.native.js
nathunsmitty: It doesn't by default but setting `lineHeight` equal to the `height` [does work](https://stackoverflow.com/questions/8865458/how-do-i-vertically-center-text-with-css)!

A problem with this approach, however, is that we can't seem add any top or bottom padding – this seems to throw off the centering. I don't think that's supper important considering the top and bottom paddings are sort of implicitly determined by height, though.

nathunsmitty:
A few changes based on @chrisnojima's review!

The `Badge2` api now looks like:
```js
export type Badge2Props = {
  badgeNumber: number,
  fontSize: number,
  height?: number,
  leftRightPadding: number,
  badgeStyle?: StylesCrossPlatform,
  badgeNumberStyle?: StylesCrossPlatform,
}
```
`badgeNumber` is the only prop that is required; all other props either have defaults or are optional. Height is calculated from font size if not specified. Based on looking at current `Badge` usage, I was thinking it may also be good to add margin as a first-class property since it seems to be used a lot to position badges. This currently isn't implemented though, so let me know if I should do it.

Here's a little sample usage of `Badge2` with this new api and a corresponding screenshot.

```js
<Badge2 fontSize={10} badgeNumber={number} />
<Badge2 badgeNumber={number} />
<Badge2 fontSize={12} height={30} badgeNumber={number} />
<Badge2 fontSize={14} leftRightPadding={2} badgeNumber={number} />
<Badge2 fontSize={16} badgeNumber={number} badgeStyle={{backgroundColor: 'green'}} />
```

simulator screen shot - iphone 8 plus - 2018-07-05 at 11 40 04

cecileboucheron:
I feel like the one on the latest builds right now is smaller than what we used to have. I liked it when it was a bit bigger.

nathunsmitty:
@cecileboucheron is there a specific height you'd like the badges to be at?

shared/common-adapters/badge.js
```diff
@@ -0,0 +1,119 @@
...
+    paddingLeft: 4,
+    paddingRight: 4,
+  },
+  text: platformStyles({
+    common: {
+      color: globalColors.white,
+      flex: 0,
+    },
+    isElectron: {
+      lineHeight: '8px',
```
chrisnojima: prefer lineHeight: 8, which platformStyles adds 'px' to for you automatically

chrisnojima:
lets chat with @cecileboucheron and update this, then get this in

nathunsmitty:
@cecileboucheron @chrisnojima are we good to go on this or are design things I should change?
simulator screen shot - iphone 8 plus - 2018-07-11 at 15 46 42

adamjspooner:
Is the text next to the icons now aligned to the bottom of the icon? They're centered on master.

nathunsmitty:
@adamjspooner Yes! Just fixed 🙂

adamjspooner:
Thanks!

nathunsmitty:
@keybase/react-hackers would love some review for this!

shared/settings/nav/settings-item.js
```diff
@@ -0,0 +1,90 @@
...
+  platformStyles,
+  globalStyles,
+  globalColors,
+  globalMargins,
+  styleSheetCreate,
+  hairlineWidth,
+  collapseStyles,
+} from '../../styles'
+
+export default function SettingsItem({
```
chrisnojima: lets not spread like this

shared/settings/nav/settings-item.js
```diff
@@ -0,0 +1,90 @@
...
+  badgeNumber,
+  icon,
+  largerBadgeMinWidthFix,
```
chrisnojima: 🔥

shared/styles/css.js.flow
```diff
@@ -265,7 +265,7 @@ export type _StylesDesktop = {|
   left?: Dimension,
   letterSpacing?: string,
   lineBreak?: string,
-  lineHeight?: string,
+  lineHeight?: Dimension,
```
chrisnojima: we actually want this to be a string likely. people too often just pass a number to it when thats the wrong thing

shared/common-adapters/badge.js
chrisnojima: this is still the string

nathunsmitty: was waiting to merge #12927 in, since that changed the lineHeight type. rebasing now

nathunsmitty:
@keybase/react-hackers bump? :)

PR #12675, opened 2018-07-03
Git hint

mlsteele:
I usually `git remote add` when a project becomes worth saving. Maybe other people do the same.

### DIFF
--- a/go/client/cmd_git_create.go
+++ b/go/client/cmd_git_create.go
@@ -81,7 +81,11 @@ func (c *CmdGitCreate) Run() error {
 	}
 
 	dui := c.G().UI.GetDumbOutputUI()
-	dui.Printf("Repo created! You can clone it with:\n  git clone %s\n", urlString)
+	dui.Printf(`Repo created! You can clone it with:
+  git clone %s
+Or add it as a remote to an existing repo with:
+  git remote add origin %s
+`, urlString, urlString)
 	return nil
 }
 


go/client/cmd_git_create.go
```diff
@@ -81,7 +81,11 @@ func (c *CmdGitCreate) Run() error {
 	}
 
 	dui := c.G().UI.GetDumbOutputUI()
-	dui.Printf("Repo created! You can clone it with:\n  git clone %s\n", urlString)
+	dui.Printf(`Repo created! You can clone it with:
+  git clone %s
+Or add it as a remote to an existing repo with:
+  git remote add origin %s
```
strib: Not sure we should be recommending `origin` to people for existing repos, since that could easily fail if there's already an `origin` and it would probably be too verbose to explain all the subtleties here.  But I really don't know for sure, and I think @malgorithms is the best choice of reviewer for user-facing wording issues.

mlsteele: That's fair. I usually use 'keybase' anyway. But.. for use cases where you want to keep your stuff encrypted I'm not sure what other origin you could have while simultaneously not understanding what the command does.

mlsteele: I'll leave it to either of you to decide the origin name. Anything to copy and paste would reduce friction for me.

malgorithms: I don't have an opinion here, as long as the regular clone command stays, for people creating new ones. For adding it to an existing repo with other remotes, I'm ok with using either `origin` or not...seems like pros and cons either way. You could do:

```

Or add it as a remote to an existing repo with:

    git remote add <pick_a_remote_name> keybase://private/chris/asteroids.git

```

That may slow down your copy/pasting since you'll have to edit, but at least it's educational.

mlsteele: Cool. Alright I'm ready to go with `origin` or `keybase`. I think `origin` is good for those who don't know what an origin is, and those who do know also know what to change. @strib

PR #12682, opened 2018-07-04
[KBFS-3052] protocol: Make new SimpleFS API for recursive ls to a certain depth

jzila:


songgao:
I think you need to add it to https://github.com/keybase/client/blob/master/protocol/bin/enabled-calls.json

jzila:
K I think this is ready. I might have to do a few more fixes after I merge the changes to KBFS, to correctly parse the paths from ListRecursive.

protocol/avdl/keybase1/simple_fs.avdl
```diff
@@ -175,6 +185,11 @@ protocol SimpleFS {
 */
   void simpleFSListRecursive(OpID opID, Path path, ListFilter filter, boolean refreshSubscription);
 
+  /**
+   Begin recursive list of items in directory at path up to a given depth
+   */
+  void simpleFSListRecursiveToDepth(OpID opID, Path path, ListFilter filter, int depth);
```
songgao: Did you mean to use `ListToDepthArgs` here rather than flattening them out?

jzila: Nope I want them flat because it's called with either a `ListArgs` or `ListToDepthArgs`.

songgao: Do we want `refreshSubscription` here too?

songgao:
Oh actually, I think you also need to change `go/service/simplefs.go` and `go/client` (a test mock, I think) to add the new RPC to satisfy the interface.

PR #12686, opened 2018-07-04
move more actions into platform specific. split out avatar stuff

chrisnojima:
@keybase/react-hackers


chrisnojima:
@keybase/react-hackers bump

PR #12689, opened 2018-07-04
WIP: 

chrisnojima:
WIP:

- [x] Moves route storage to platform-specific.native
- [x] Moves push configure to platform-specific
- [x] Removes usernameSelector and loggedInSelector. Really we should just use flow and the state to get simple values out. Or we should use helpers in the constants/config to do this. Not a generic 'selectors' file. We definitely haven't been filling that out and a per-substore-constants-helpers thing makes more sense i think
- [x] Updated safeTakeEveryPureSimple to not yield (again) the results. What it was doing was doing a call, then getting the results, then yielding that. Instead making this more like safeTakeEveryPure (which was the idea) and doing the actual call then yielding the results. This lets you return Saga.sequentially() w/o it complaining
- [ ] re-test signup to ensure the util/saga change doesn't break that code
- [ ] change takeEveryPureSimple to use takeEvery and not takeEveryPure


chrisnojima:
https://github.com/keybase/client/pull/12711 replaced by

PR #12694, opened 2018-07-05
Stellar: Export Secret Key screen

cjb:
@keybase/react-hackers 

First up: @patrickxb or @mlsteele, are you okay with moving `exportSecretKeyLocal` to `enabled-calls.json` now?  It's used by the frontend for this PR.

Redux security considerations: Frontenders, I decided to use a `HiddenString` for the Stellar secret key value in the redux store, and to clear that `HiddenString` out from the store as soon as the export dialog is closed.  Sound right?


mlsteele:
@cjb yep sounds good

cjb:
@zanderz Thanks -- is that a thumb? :) Pinentry's switched over to the non-passphrase version now.

shared/actions/wallets.js
```diff
@@ -121,6 +133,7 @@ function* walletsSaga(): Saga.SagaGenerator<any, any> {
     [WalletsGen.selectAccount, WalletsGen.linkedExistingAccount],
     navigateToAccount
   )
+  yield Saga.safeTakeEveryPure(WalletsGen.exportSecretKey, exportSecretKey, exportSecretKeySuccess)
```
chrisnojima: likely better to use safeTakeEveryPurePromise

shared/reducers/wallets.js
```diff
@@ -18,6 +18,10 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
       return state.setIn(['assetsMap', action.payload.accountID], I.List(action.payload.assets))
     case WalletsGen.paymentsReceived:
       return state.setIn(['paymentsMap', action.payload.accountID], I.List(action.payload.payments))
+    case WalletsGen.secretKeyReceived:
+      return state.setIn(['secretKeyMap', action.payload.accountID], action.payload.secretKey)
+    case WalletsGen.secretKeySeen:
+      return state.set('secretKeyMap', I.Map())
```
chrisnojima: why have this be a map and set it based on account but clear the entire map? likely this can just be one value right

chrisnojima: if you switch accounts it can just clear it

shared/constants/wallets.js
```diff
@@ -163,6 +164,9 @@ const getFederatedAddress = (state: TypedState, accountID?: Types.AccountID) =>
   return username && account.isDefault ? `${username}*keybase.io` : ''
 }
 
+const getSecretKey = (state: TypedState, accountID: Types.AccountID) =>
```
chrisnojima: likely better to return a string always so you don't have to check and call stringValue() later

shared/wallets/export-secret-key/container.js
```diff
@@ -0,0 +1,36 @@
...
+export type OwnProps = {
+  accountID: Types.AccountID,
+}
+
+const mapStateToProps = (state: TypedState, {routeProps}) => {
+  const accountID = routeProps.get('accountID')
+  const secretKey = Constants.getSecretKey(state, accountID)
+  return {
+    accountID,
+    secretKey: secretKey && secretKey.stringValue(),
```
chrisnojima: then here its just `secretKey,`

shared/wallets/export-secret-key/index.js
```diff
@@ -0,0 +1,93 @@
...
+    marginBottom: globalMargins.medium,
+  },
+  secretKeyContainer: {
+    marginBottom: globalMargins.medium,
+    maxWidth: 272,
+    width: '100%',
+  },
+})
+
+const containerStyle = platformStyles({
```
chrisnojima: this should go into styles

cjb:
@chrisnojima Thanks! Made all of these changes.

PR #12699, opened 2018-07-05
protocol: add FSPathUpdated notification and SimpleFS subscription

strib:
Issue: KBFS-3120


go/libkb/notify_router.go
```diff
@@ -377,6 +379,31 @@ func (n *NotifyRouter) HandleFSActivity(activity keybase1.FSNotification) {
...
+	n.cm.ApplyAll(func(id ConnectionID, xp rpc.Transporter) bool {
+		// If the connection wants the `Kbfs` notification type
+		if n.getNotificationChannels(id).Kbfs {
+			// In the background do...
+			go func() {
+				// A send of a `FSPathUpdated` RPC with the notification
+				(keybase1.NotifyFSClient{
+					Cli: rpc.NewClient(xp, NewContextifiedErrorUnwrapper(n.G()), nil),
+				}).FSPathUpdated(context.Background(), path)
+			}()
```
jzila: Any reason not to do:
```go
go (keybase1.NotifyFSClient{Cli: rpc.NewClient(...)}).FSPathUpdated(context.Background(), path)
```

strib: I don't see one, but I was just being consistent with the other functions in this file (not written by me).

strib:
@jzila @songgao could one of you please point me in the right direction for the saga failure in this PR?

jzila:
@strib you need to add a `refreshSubscription` parameter here: https://github.com/keybase/client/blob/f37ccf692d539832d5cc79771bdcf23829222f6a/shared/actions/fs/index.js#L97

jzila:
(as discussed in Keybase chat, I pushed a fix to the CI error)

protocol/avdl/keybase1/simple_fs.avdl
```diff
@@ -156,16 +156,24 @@ protocol SimpleFS {
...
+   Begin list of items in directory at path.
+   Retrieve results with readList().
+   Can be a single file to get flags/status.
+   If `refreshSubscription` is true and the path is a KBFS path, simpleFS
+     will begin sending `FSPathUpdated` notifications for the for the
+     corresponding TLF, until another call refreshes the subscription on a
+     different TLF.
    */
-  void simpleFSList(OpID opID, Path path, ListFilter filter);
+  void simpleFSList(OpID opID, Path path, ListFilter filter, boolean refreshSubscription);
```
songgao: Can we make `refreshSubscription` a string or an integer instead? This is so we can track different "subscribers" rather than only having one.

strib: What should simplefs do with that string?  All notifications go through the notify-router to all listeners, so I guess you'd want a list of all matching subscriptions in each notification?  We could do that, but it would be a lot of work and a lot more complicated, and if it's not going to be used right away I'd rather wait until we need it.  It's the kind of thing that's easy to fix later since it's all within a single client (e.g. by adding a new field).

songgao: > so I guess you'd want a list of all matching subscriptions in each notification? 

Yeah that's fine. We'd only have one connection from GUI to core about kbfs any way. All notifications would go through that path. This is basically to make sure multiple paths can be subscribed to at the same time, and for each distinct string, we only need to keep track the latest path.

strib: > multiple paths can be subscribed to at the same time

Huh?  This is the exact opposite of what you said you needed originally.  I thought the GUI is only going to ever want to subscribe to one TLF at a time?

songgao:
> Also, would it be hard to support notification on SimpleFSStat too?

As discussed on kbchat, this isn't needed. So feel free to ignore it!

protocol/avdl/keybase1/simple_fs.avdl
songgao: Hmm is it possible that I meant for each distinct refresh tag we only need to subscribe to the latest path? If I said otherwise, perhaps it was before we had the same mechanism in GUI. Anyways, would it make a lot of difference for KBFS? I was under the impression we’d just need to have a (refreshTag->path) map instead of a string of path.

songgao: Actually thinking about it more, I think for now only having one path might be enough. The reason why we had two o GUI was we were losing mimetyoes from http server too. I don’t think we need two for folder list. Sorry for the confusing comment!

strib: > Anyways, would it make a lot of difference for KBFS?

It adds multiple maps, because we need to track the tag->`FolderBranch` relationship too.  And we'd have to refcount in order to properly unsubscribe from `folderBranchOps` events when no one is listening on a particular path anymore.

> The reason why we had two o GUI was we were losing mimetyoes from http server too.

I'm having trouble parsing this, but I thought the intent was to only refresh the screen that was currently showing to the user, and as far as I understand the UX there can only be one folder showing in the screen at one time.

songgao: > I'm having trouble parsing this, but I thought the intent was to only refresh the screen that was currently showing to the user

Sorry for not making this clear. The loadMimeType thing is separate, and is used to decide what component we use to display the file content, and for Android share. I needed to track both in GUI code and refresh both when there are FS updates. But the mime type stuff is irrelevant to folder listing. I thought of the tracking map and mistakenly thought we need multiple tracking for listing too, but we don’t. And you are right we only need the latest because that’s where the user is on the screen.

PR #12705, opened 2018-07-05
Docs for android 28

joshblum:
Documents some issues and fixes i ran into when using android emulator on latest OS

### DIFF
--- a/shared/README.md
+++ b/shared/README.md
@@ -371,14 +371,14 @@ server you should manually change the dev server URL and then kill and restart
 the app:
 
 ```sh
-# First enable loopback
+# Enable loopback
 adb reverse tcp:8081 tcp:8081
-# Opens react native debug menu
-# https://facebook.github.io/react-native/docs/debugging.html#accessing-the-in-app-developer-menu
-adb shell input keyevent 82
-# Now tap "Dev settings" and set "Debug server host & port for device to
-# "127.0.0.1:8081"
 ```
+
+Then open the [react native debug
+menu](https://facebook.github.io/react-native/docs/debugging.html#accessing-the-in-app-developer-menu),
+tap "Dev settings" and set "Debug server host & port for device" to
+`127.0.0.1:8081`
 
 #### Native inspector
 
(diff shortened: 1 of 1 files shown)


shared/README.md
```diff
@@ -364,6 +364,22 @@ See [this
...
+Metro Bundler from running properly. We have manually allowed `127.0.0.1` to
+have HTTP traffic, so if you see an error about connecting to the bundler
+server you should manually change the dev server URL and then kill and restart
+the app:
+
+```sh
+# First enable loopback
+adb reverse tcp:8081 tcp:8081
+# Opens react native debug menu
+# https://facebook.github.io/react-native/docs/debugging.html#accessing-the-in-app-developer-menu
```
chrisnojima: you can just hit up volume

joshblum: ah good to know

akalin-keybase: Since the linked page doesn't actually mention hitting up volume, may be good to mention this here too

shared/README.md
```diff
@@ -364,6 +364,22 @@ See [this
...
+
+```sh
+# Enable loopback
+adb reverse tcp:8081 tcp:8081
+```
+
+Then open the [react native debug
+menu](https://facebook.github.io/react-native/docs/debugging.html#accessing-the-in-app-developer-menu),
+tap "Dev settings" and set "Debug server host & port for device" to
+`127.0.0.1:8081`
```
akalin-keybase: is the default not already `127.0.0.1:8081`?

chrisnojima:
It should be actually

shared/README.md
akalin-keybase: nojima said:
> It should be actually

so maybe the language should be
`...to 127.0.0.1:8081 (the default) if it isn't already`

joshblum: it wasn't the default for me but maybe something has been changed since i set things up

shared/README.md
joshblum: i'd rather leave as is and let their docs explain in case things change in the future

Issue #12701, opened 2018-07-05
Loading folders takes forever and then times out


scolby33:
Keybase GUI Version: 2.1.2-20180625190314+836e4aeaca
Client:  2.1.2-20180625190314+836e4aeaca
Service: 2.1.2-20180625190314+836e4aeaca
OS: macOS 10.13.5

See https://github.com/vimwiki/vimwiki/issues/525#issuecomment-402841434

```sh
$ time ls /keybase/private/scolby33/vimwiki
ls: /keybase/private/scolby33/vimwiki: Input/output error
      129.95 real         0.00 user         0.00 sys
```

Opening the folder in Finder shows no content.

What's up with this?

scolby33:
The log id is 27a911fd252e28c4c939521c.

strib:
@maxtaco @patrickxb these hangs appear to be from identify calls that are hanging, i.e.:

```
2018-07-05T13:27:03.204296-07:00 ▶ [DEBU kbfs(FBO 1b14d745) folder_branch_ops.go:1013] 1cbf Running identifies on /keybase/private/scolby33 [tags:FID=POBABK3J8KUa203ZwXkBHQ]
2018-07-05T13:27:22.278498-07:00 ▶ [DEBU kbfs(FBO 1b14d745) folder_branch_ops.go:1017] 1cc0 Identify finished with error: context deadline exceeded [tags:FID=POBABK3J8KUa203ZwXkBHQ]
```

The corresponding service logs seem to just stop after doing this:

```
2018-07-05T13:27:03.205496-07:00 ▶ [DEBU keybase engine.go:140] 3e08 IdentifyUI wanted for engine "Identify2WithUID" [tags:ENG=DXjcwAfzYgfa,ID2=YHXY4MxWlUxn,FID=POBABK3J8KUa203ZwXkBHQ,IDL=PfZn9aS1QAyY]
```

Not sure what this means, please advise.

maxtaco:
Hahah, looks like we're chasing this one all the way up. I think it's hanging indefinitely on electron.

maxtaco:
@keybase/react-hackers is it possible we never got a reply from `DelegateIdentifyUI`?  That's my best read of this log.

maxtaco:
@scolby33 for now, can you try quitting out of the app, and the restarting?

scolby33:
(a) I don't know why I didn't try that
(b) That seems to have fixed the problem, albeit probably temporarily.

[...]

PR #12710, opened 2018-07-06
dont wait for a response from pushes into electron

mmaxim:
cc @chrisnojima 

This is to fix a problem where Electron just stops responding to service RPC calls into it. Because `pushState` and `pushOutOfBandMessages` are on the main thread of processing Gregor messages, this can back up the whole system. 

```
2018-07-05T20:25:50.380639-04:00 ▶ [DEBU keybase gregor.go:931] 57fea ++Chat: - PushHandler: broadcastMessageOnce -> ok (5h49m45.56188765s)
```


maxtaco:
Are we worried about a bigger problem?  This one came in today: https://github.com/keybase/client/issues/12701

Basically, electron didn't reply to the `DelegateUI` RPC, and the FS operation hung.

PR #12711, opened 2018-07-06
WIP: more saga cleanup

chrisnojima:


chrisnojima:
- [ ] after release

chrisnojima:
@keybase/react-hackers bump

chrisnojima:
@keybase/react-hackers bump

chrisnojima:
@keybase/react-hackers bump

zanderz:
So, this route storage was a no-op on desktop anyway?
And our selectors were so simple there was little benefit?
LGTM

chrisnojima:
I moved the per platform stuff into the per platform actions and not have
that delegated by config. Selectors are ok. I don't see the value if it's
just grabbing a value but if they exist it should be in the constants of
the sub store and not this generic selectors file

chrisnojima:
@react-hackers one more thumb please

PR #12712, opened 2018-07-06
run all chat notify through a single thread to guarantee ordering 

mmaxim:
If all these calls just go to `NotifyRouter` right away, they can all line up on the `ConnectionManager` lock in `ApplyAll`. If that happens, then all the threads waiting on the lock can get out of order. Patch does the following:

1.) Introduce `ActivityNotifier` a new interface in `globals`.
2.) Implement with `NotifyRouterActivityRouter` which just serializes calls into `NotifyRouter`. 
3.) Don't run any notifications from `OutboxPurge`, do it in a package outside.


mmaxim:
cc @maxtaco this is to prevent any out of order notifications as a result of my other change to unlock the Gregor broadcast thread from having to wait on the frontend.

PR #12717, opened 2018-07-06
Partial pairwiseMAC reader support

joshblum:
cc @maxtaco  @mmaxim just as a sanity check that we don't need any changes to the upcoming release for pairwise mac support. Since reader support for `chat1.MessageBoxedVersion_V4` has had some time in the wild, once upak cache busting is supported we should be able to turn on pairwiseMAC sending and old clients will be able to decrypt these messages.

This pr just cleans up some comments about pairwiseMAC support.

### DIFF
--- a/go/chat/boxer.go
+++ b/go/chat/boxer.go
@@ -416,7 +416,7 @@ func (b *Boxer) unbox(ctx context.Context, boxed chat1.MessageBoxed,
 	case chat1.MessageBoxedVersion_V2, chat1.MessageBoxedVersion_V3, chat1.MessageBoxedVersion_V4:
 		res, err := b.unboxV2orV3orV4(ctx, boxed, membersType, encryptionKey, ephemeralSeed)
 		if err != nil {
-			b.Debug(ctx, "error unboxing message version: %v, %v", boxed.Version, err)
+			b.Debug(ctx, "error unboxing message version: %v, %s", boxed.Version, err)
 		}
 		return res, err
 	// NOTE: When adding new versions here, you must also update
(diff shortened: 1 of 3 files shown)


go/chat/boxer.go
```diff
@@ -416,7 +416,7 @@ func (b *Boxer) unbox(ctx context.Context, boxed chat1.MessageBoxed,
 	case chat1.MessageBoxedVersion_V2, chat1.MessageBoxedVersion_V3, chat1.MessageBoxedVersion_V4:
 		res, err := b.unboxV2orV3orV4(ctx, boxed, membersType, encryptionKey, ephemeralSeed)
 		if err != nil {
-			b.Debug(ctx, "error unboxing message version: %v", boxed.Version)
+			b.Debug(ctx, "error unboxing message version: %v, %v", boxed.Version, err)
```
mmaxim: %s

go/chat/boxer.go
```diff
@@ -741,16 +741,7 @@ func (b *Boxer) unboxV2orV3orV4(ctx context.Context, boxed chat1.MessageBoxed,
 	// intended for our device, and error out if it's missing or invalid. If it
 	// is valid, then we *don't* validate the message signing key. That is,
 	// even though signEncryptOpen will check a signature in the end, we no
-	// longer care what signing key it's using. That means we're compatible
```
mmaxim: What is wrong with this comment?

joshblum: we never ended up optionally including pairwise macs but pushed reader support for v4 when we initially released exploding messages. Once we fix the caching issues we'll release with writing pairwise macs on v4 since existing clients with exploding support have mac reader support

PR #12721, opened 2018-07-06
turn on power save blocker for electron

mmaxim:
This doesn't pass Flow, maybe because the flow package we use for Electron is out of date.  If this gets approved, maybe someone can help me update these types. 

Docs for this feature:
https://electronjs.org/docs/api/power-save-blocker

cc @maxtaco


chrisnojima:
we own the electron types we use so just update it here: https://github.com/keybase/client/blob/master/shared/libs/electron_vx.x.x.js.flow

chrisnojima:
does this actually work? what are the side effects?

mmaxim:
I'm running it now, but it is very hard to test it. I basically need to get the OS to attempt to suspend the app, which seems to only happen after the machine has been locked for some amount of time. This seems like it hits the exact problem we are seeing though, so if it doesn't work I would be surprised. 

I guess a downside would be the app looks like it is consuming more energy than it used to, but I think that is a fine price to pay considering the behavior that we see now.

PR #12718, opened 2018-07-06
Better Initial Chat Experience

thebearjew:
## New Inbox Components

Adds a new button `StartNewChat` to the top of the inbox in place of `ChatFilterRows` when the user has not chats.

Additionally this adds a button at the bottom of the inbox, `BuildTeam` when the user is not part of a any teams (and thus has no chats), which takes the user to the teams page.

screen shot 2018-07-06 at 5 13 54 pm

screen shot 2018-07-09 at 9 50 37 am

## Storybook
Most of the work in this PR involves rewriting the stories for `Chat/Inbox`. Previously only the inbox rows were being rendered in a story. Now the stories will render an entire inbox including the header components (`StartNewConversation` or `ChatFilteredRow`), the row components, and `BuildTeam` component if it applies.

PR #12722, opened 2018-07-06
Use custom icons in for exploding messages in inbox

nathunsmitty:
## Description
Currently, we use the emojis 💥 and 💣 to signify exploded and soon-to-explode messages in small teams in a user's inbox. We use custom icons everywhere else, so this update replaces these emojis with those custom icons.

## Solution
It's actually [defined in the API](https://github.com/keybase/client/blob/82a7496704e5471d314f262b3165f0260608f0c8/protocol/avdl/chat1/chat_ui.avdl#L20) to return these emojis, so I wrote a `switch` statement to convert these emojis (or the lack of one) into the appropriate `Icon` component. There was also some difficulty centering the `Icon` components, so I placed them in a `Box2`.

I also added a couple more conversations with the new icons to the [simple inbox story in Storybook](http://localhost:6006/?selectedKind=Chat%2FInbox&selectedStory=Simple&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) and updated our snapshot accordingly.


nathunsmitty:
@keybase/react-hackers

shared/chat/inbox/row/small-team/bottom-line.js
```diff
@@ -93,14 +93,27 @@ class BottomLine extends PureComponent<Props> {
...
+      // `snippetDecoration` will either be an explosion emoji, bomb emoji, or empty string.
+      // We want to use these emojis to render the correct custom icon.
+      switch (this.props.snippetDecoration) {
+        case '\u{1F4A5}': // Explosion (Collision) emoji (💥)
+          snippetDecoration = <Icon type="iconfont-boom" fontSize={16} />
+          break
+        case '\u{1F4A3}': // Bomb emoji (💣)
+          snippetDecoration = <Icon type="iconfont-bomb" fontSize={12} />
+          break
+      }
```
chrisnojima: we need a story for what happens if this changes
At a bare minimum if this.props.snippetDecoration doesn't match those 2 cases it should default: to passing the value straight through
aka
```
default:
  snippetDecoration = this.props.snippetDecoration
```

nathunsmitty: added! also added some conditional rendering of the containing box, just in case we get a falsey value like empty string.

shared/chat/inbox/row/index.stories.js
```diff
@@ -90,6 +90,22 @@ const mocks = [
...
+    snippetDecoration: '💣',
+  },
+  {
+    ...simpleCommon,
+    conversationIDKey: '7',
+    participants: ['nathunsmitty'],
+    snippet: 'lol',
+    timestamp: '1:37 pm',
+    snippetDecoration: '💥',
+  },
```
thebearjew: Fair warning, I just removed this file in favor of a story that tests the entire inbox component rather than just the row components.

My PR: #12718 

I can update the necessary stories after this merges.

nathunsmitty: Thanks for the heads up! I can also update this branch if your changes get merged in first.

nathunsmitty:
@keybase/react-hackers could probably use a bump if anyone has a few minutes. Lemme know merging strategy with #12718, too.

nathunsmitty:
Also updated the look a little bit to better reflect @cecileboucheron's designs. Here's what my inbox looks like now:
screen shot 2018-07-10 at 12 23 20 pm

adamjspooner:
@cecileboucheron, should those `boom!`s be gray instead of black?

cecileboucheron:
Yeah agreed, let's do black_40 for the boom icon. Also the bomb could be a bit smaller I think - maybe 12px font-size on desktop and 16 on mobile, since the snippet is using body small.

nathunsmitty:
Updated!
screen shot 2018-07-10 at 5 32 11 pm

cecileboucheron:
LGTM!

PR #12727, opened 2018-07-06
Clear chat from infopanel, not individual message

zanderz:
also some UI changes.
Note that @cecileboucheron said the delete confirmation should be a popup, where it isn't presently, but I would like to merge this in and do that later because iconfonts.


shared/chat/conversation/info-panel/container.js
```diff
@@ -21,15 +21,18 @@ type OwnProps = {
...
   let admin = false
   let canEditChannel = false
   let canSetRetention = false
+  let canDeleteHistory = false
   if (meta.teamname) {
     const yourOperations = getCanPerform(state, meta.teamname)
     admin = yourOperations.manageMembers
     canEditChannel = yourOperations.editChannelDescription
     canSetRetention = yourOperations.setRetentionPolicy
+    canDeleteHistory = lastMessage && (meta.teamType === 'adhoc' || yourOperations.deleteChatHistory)
```
chrisnojima: why isn't yourOperations just have this as true for adhoc, why do we need these extra checks?

shared/chat/conversation/messages/message-popup/exploding/container.js
```diff
@@ -25,7 +24,7 @@ const mapStateToProps = (state: TypedState, ownProps: OwnProps) => {
   const yourMessage = ownProps.message.author === state.config.username
   const meta = Constants.getMeta(state, ownProps.message.conversationIDKey)
   const _canDeleteHistory =
-    meta.teamType === 'adhoc' || TeamConstants.getCanPerform(state, meta.teamname).deleteChatHistory
+  meta.teamType === 'adhoc' || TeamConstants.getCanPerform(state, meta.teamname).deleteChatHistory
```
chrisnojima: same here, i understand this isn't part of this pr but this is really error prone i think

shared/actions/chat2/index.js
```diff
@@ -1742,7 +1733,12 @@ const markThreadAsRead = (
 const deleteMessageHistory = (action: Chat2Gen.MessageDeletePayload, state: TypedState) => {
```
chrisnojima: we should update the rpc so we don't have to pass a message id anymore and bookkeep this. The UX is just that we want the entire thing cleared so we should be able to just call some rpc that does that and takes just the conversationIDKey

shared/chat/conversation/info-panel/container.js
zanderz: I don't know, I just copied the existing logic.

shared/actions/chat2/index.js
zanderz: I considered that, but won't this operation then be broken for older clients? e.g., admin has this version and sends clear message with no message ID, and another team member has an older client. If this case isn't a big deal then yeah I agree.

chrisnojima: i'm talking about the rpc between the app and the daemon, which is just local so it doesn't affect other clients. the daemon can derive what the last message is

shared/chat/conversation/messages/message-popup/exploding/container.js
zanderz: CanPerform is per team, not per convo. Should we have a separate chat1.CanUserPerform() or something?

shared/chat/conversation/info-panel/container.js
zanderz: `teamType` seems to be a conversation property, not a team property. Maybe we need a `chat1.canUserPerform()`?

chrisnojima:
But your rights apply to all the convos in the team right?

zanderz:
@keybase/react-hackers This should be ready now

shared/chat/delete-history-warning/index.js
```diff
@@ -1,44 +1,53 @@
...
-    </Text>
-    <Box style={{...globalStyles.flexBoxRow, marginTop: globalMargins.xlarge}}>
-      <Button type="Secondary" style={{marginLeft: globalMargins.tiny}} onClick={onBack} label="Cancel" />
-      <Button
-        type="Danger"
-        style={{marginLeft: globalMargins.tiny}}
-        onClick={onDeleteHistory}
-        label="Yes, delete these messages"
-      />
+const Wrapper = ({children, onBack}) =>
```
chrisnojima: we have MaybePopup  for this

chrisnojima:
Can we fix the layout on mobile on the warning page: just make the buttons vertically aligned in the buttonbar
screen shot 2018-07-11 at 2 08 59 pm

cecileboucheron:
Also let's do a Cancel in the top-left corner (vs back button), and use the vertical full-width version of the buttons at the bottom:
image
image
Also let's drop the header title, center-align both texts and have them use the same left/right margins (40px from edges) so they look aligned.
Full specs for mobile confirm window: https://zpl.io/agG9WZO

zanderz:
ok, now Mobile looks like this:
image

shared/chat/delete-history-warning/index.js
```diff
@@ -7,11 +7,16 @@ type Props = {
...
   name: string,
   onBack: () => void,
+  onCancel: () => void,
   onDeleteHistory: () => void,
 }
 
-const DeleteHistoryWarning = ({errorText, name, onBack, onDeleteHistory}: Props) => (
-  <MaybePopup onClose={onBack}>
+const cancelButton = (onClick: () => void) => {
+  return <Button type="Secondary" style={styleButton} onClick={onClick} label="Cancel" fullWidth={isMobile} />
```
chrisnojima: This should just be a component. Not a function

shared/chat/delete-history-warning/index.js
```diff
@@ -1,55 +1,89 @@
...
+    >
+      <Icon type={isMobile ? 'icon-message-deletion-64' : 'icon-message-deletion-48'} />
+      <Text style={{padding: globalMargins.small}} type="Header">
+        Delete conversation history?
+      </Text>
+      <Text style={{padding: globalMargins.small}} type="Body">
+        You are about to delete all the messages in this conversation. For everyone.
+      </Text>
+      <Box style={styleButtonBox}>
+        {!isMobile && cancelButton(onCancel || onBack)}
```
chrisnojima: You can just style this with flex direction instead of putting in the top or bottom. Column-reverse etc

shared/chat/delete-history-warning/container.js
```diff
@@ -3,37 +3,34 @@ import * as Types from '../../constants/types/chat2'
...
   return {
     teamname,
-    timestamp,
   }
 }
 
 const mapDispatchToProps = (dispatch: Dispatch, {navigateUp, routeProps}: OwnProps) => ({
-  onBack: () => dispatch(navigateUp()),
+  onCancel: isMobile ? () => dispatch(navigateUp()) : null,
+  onBack: isMobile ? null : () => dispatch(navigateUp()),
```
chrisnojima: Let's just pass oncancel

zanderz:
ok made more changes

zanderz:
should be ok, pending storyshots

shared/chat/delete-history-warning/container.js
```diff
@@ -3,37 +3,33 @@ import * as Types from '../../constants/types/chat2'
...
   onDeleteHistory: () => {
-    const message = routeProps.get('message')
+    const conversationIDKey = routeProps.get('conversationIDKey')
     dispatch(navigateUp())
     dispatch(
       Chat2Gen.createMessageDeleteHistory({
-        conversationIDKey: message.conversationIDKey,
-        ordinal: message.ordinal,
+        conversationIDKey: conversationIDKey,
```
chrisnojima: can just be `conversationIDKey,`

PR #12736, opened 2018-07-07
this shouldnt be possible but see if it fixes that crash

chrisnojima:
@keybase/react-hackers this is an extra check . we got some crash reports (@cjb ) where something.offline was referring to undefined and this is really the only place where that's possible. 
cc: @mmaxim would the service ever return null here? seems to violate the avdl...

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -883,7 +883,9 @@ const loadMoreMessages = (
         },
         waitingKey: loadingKey,
       })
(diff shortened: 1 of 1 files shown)


mmaxim:
Ok, it might be worth be worth thinking about if it is possible to get `null` here outside of a `null` response from the service, just in case. I looked at that log and don't see any kind of unusual response.

chrisnojima:
Yeah I'll have to see. Maybe the event channel gets hosed by backgrounding
or something. Will have to investigate

PR #12737, opened 2018-07-07
add simpleFSSuppressNotifications to reduce FS notifications

songgao:
This mechanism is to ask KBFS not to send FSActivity and FSSyncActivity when we are already polling.

Also:

1) act on both FSActivity and FSSyncActivity;
2) increase the polling interval from 2s to 4s. When we have UI tweaks for the countdown timer, this won't be as noticeable as now.

cc @mmaxim


protocol/avdl/keybase1/simple_fs.avdl
```diff
@@ -307,4 +307,17 @@ protocol SimpleFS {
    recorded by the server) of their most recent edit.
    */
   FSFolderEditHistory simpleFSFolderEditHistory(Path path);
+
+  // suppressNotifications suppresses KBFS notifications, particularly FSEvent
+  // and FSSyncEvent.
+  //
+  // Caller should call this again after nextSuppressIn to continue
+  // suppressing.  If no such call is made in 2*nextSuppressIn, KBFS will
```
strib: What are the units of `nextSuppressIn`?  Is it supposed to be a time duration (in which case you should have a `Sec` or `MSec` suffix)?

protocol/avdl/keybase1/simple_fs.avdl
```diff
@@ -307,4 +307,17 @@ protocol SimpleFS {
...
+  // resume notifications.
+
+  // When resuming notifications, if a notification would have happened since
+  // last suppressNotifications call but was suppressed, this notification will
+  // go through. If multiple notifications would have happened, only one of
+  // them is sent at resuming.
```
strib: Why does it make sense to send one event?  Seems kind of arbitrary, since all the events are independent.

protocol/avdl/keybase1/simple_fs.avdl
songgao: It used to be the keybase1.DurationSec type. I changed it into int because I didn’t like it to be a floating number, but forgot to add the sec suffix to name. Will fix!

protocol/avdl/keybase1/simple_fs.avdl
songgao: At this moment we treat all notifications equally without inspecting its content. We just use it as a signal to start the polling, so it wouldn’t break anything if we miss a couple of them. Sending all the missed one one the other hand is a bit wasteful. Perhaps we should change the notification semantics or add a new (that doesn’t make it distinct from each other) one to avoid confusion?

strib: I don't want to change semantics in this PR.  I guess just explain all this in the comment so it makes it clear that the listener shouldn't depend on any differential nature of the data between different notifications.

PR #12742, opened 2018-07-08
Catch all exceptions on background job

joshblum:
Wrap background job in a try/catch. imo we should only do this since the root cause of crashers is undetermined, and we should dig in further after release


chrisnojima:
merging and kicking a build

PR #12750, opened 2018-07-09
Select conversation first, then unbox

joshblum:
select conv and then unbox


chrisnojima:
What motivates this change

joshblum:
@mmaxim was thinking this might improve performance when tapping in

chrisnojima:
Bite were code frozen so unless this really solves something let's wait

mmaxim:
We should get this in, otherwise coming from a push notification is kind of bustedl

PR #12752, opened 2018-07-09
Avatar cache fixes

zapu:
This PR introduces the following changes:
1) Fixes a bug in `populateCacheWorker` where HTTP response was not properly disposed of. On some platforms it might have caused a resource leak.
2) Added a new method to `avatars.Source` interface: `OnCacheCleared`. It's called after DB nuke and gives a chance to avatar source to clean up after cache database has been cleared. `FullCachingSource` removes all avatar files from cache directory. Other sources don't have to do anything.
3) Moved avatar cache storage path to `avatars/` subdirectory (instead of main keybase cache dir). `FullCachingSource` will ensure this dir exists when it's initialized. Old cached avatars will continue to live on old path but new ones will be put in new path. Avatars from old directory will be slowly phased out with cache eviction.

### DIFF
--- a/go/avatars/fullcaching.go
+++ b/go/avatars/fullcaching.go
@@ -182,7 +182,9 @@ func (c *FullCachingSource) commitAvatarToDisk(ctx context.Context, data io.Read
 		// started saving them to `avatars/` subdir. If user has just
 		// updated to client with new path, it's fine to have them
 		// start clean.
-		c.unlinkAllAvatars(ctx, c.G().GetCacheDir())
+		if len(c.tempDir) == 0 {
+			c.unlinkAllAvatars(ctx, c.G().GetCacheDir())
+		}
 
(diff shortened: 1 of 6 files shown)


mmaxim:
Can this also just delete all the avatars from the old directory? It isn't so bad if we just start from scratch.

zapu:
Because it stores paths in leveldb, it would just resave avatars to old directory if we only deleted files.

So we would have to either go through leveldb and clean old entries up, or make it ignore old paths
in e.g. `commitAvatarToDisk` https://github.com/keybase/client/pull/12752/files#diff-2f8f36b16ddaa8bf55e3faca35d46568L182

mmaxim:
Can we just change that to an `Open` call? I'm not sure why I used `OpenFile`

zapu:
and then make it go ahead the tempfile route if it can't open? sounds good!

PR #12757, opened 2018-07-09
Attempt ephemeral keygen when going into background active state

joshblum:
Although the mobile apps wake up during a `BackgroundSync`  periodically, the amount of active runtime often does not exceed the time for our routine checks to fire. Instead we monitor the app state to manually fire the check if necessary

### DIFF
--- a/go/ephemeral/lib.go
+++ b/go/ephemeral/lib.go
@@ -80,6 +80,11 @@ func (e *EKLib) backgroundKeygen() {
 			runIfNeeded()
 		case state = <-e.G().AppState.NextUpdate(&state):
 			if state == keybase1.AppState_BACKGROUNDACTIVE {
+				// Before running  we pause briefly so we don't stampede for
+				// resources with other background tasks. libkb.BgTicker
+				// handles this internally, so we only need to throttle on
+				// AppState change.
+				time.Sleep(time.Second)
 				runIfNeeded()
 			}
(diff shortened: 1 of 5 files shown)


go/ephemeral/lib.go
```diff
@@ -36,11 +36,44 @@ func NewEKLib(g *libkb.GlobalContext) *EKLib {
...
+	ekLib := &EKLib{
 		Contextified:   libkb.NewContextified(g),
 		teamEKGenCache: nlru,
 		clock:          clockwork.NewRealClock(),
 	}
+	go ekLib.backgroundKeygen()
+	return ekLib
+}
+
+func (e *EKLib) backgroundKeygen() {
```
mmaxim: how does this loop end?

joshblum: fixed with `Shutdown` method

go/ephemeral/lib.go
```diff
@@ -36,10 +37,55 @@ func NewEKLib(g *libkb.GlobalContext) *EKLib {
...
+	}
+	go ekLib.backgroundKeygen()
+	return ekLib
+}
+
+func (e *EKLib) Shutdown() {
+	e.Lock()
+	defer e.Unlock()
+	if e.stopCh != nil {
+		e.stopCh <- struct{}{}
```
mmaxim: Why not `close` it?

PR #12759, opened 2018-07-09
Do not verify team.leave links

mlsteele:
Don't verify team.leave links. Needs https://github.com/keybase/keybase/pull/2661. The thing that's not being verified is whether the signing KID has anything to do with the signing UV. So effectively the server could make leave anyone leave by signing a link with any KID.

If we plan to do this
- [ ] decide whether it's safe
- [ ] write more comments explaining `fullVerify`
- [ ] not sure whether omitting the link map part could stall order checking
- [ ] put an extra note in `chain.go` that it should never support key rotation on leave links

### DIFF
--- a/go/teams/loader2.go
+++ b/go/teams/loader2.go
@@ -129,6 +129,21 @@ func (l *TeamLoader) loadUserAndKeyFromLinkInner(ctx context.Context,
 	return signerUV, key, linkMap, nil
 }
 
+// Get the UV from a link but using server-trust and without verifying anything.
+func (l *TeamLoader) loadUserAndKeyFromLinkInnerNoVerify(ctx context.Context,
+	link *chainLinkUnpacked) (signerUV keybase1.UserVersion, err error) {
+	defer l.G().CTraceTimed(ctx, fmt.Sprintf("TeamLoader#loadUserAndKeyFromLinkInnerNoVerify(%d)", int(link.inner.Seqno)), func() error { return err })()
+	keySection := link.inner.Body.Key
+	if keySection == nil {
+		return signerUV, libkb.NoUIDError{}
+	}
+	// Use the UID from the link body and EldestSeqno from the server-trust API response.
+	if link.source.EldestSeqno == 0 {
+		return signerUV, fmt.Errorf("missing server hint for team sigchain link signer")
+	}
+	return NewUserVersion(keySection.UID, link.source.EldestSeqno), nil
+}
+
 func (l *TeamLoader) verifySignatureAndExtractKID(ctx context.Context, outer libkb.OuterLinkV2WithMetadata) (keybase1.KID, error) {
 	return outer.Verify(l.G().Log)
 }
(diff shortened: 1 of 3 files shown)


mlsteele:
Talked with @maxtaco . Scan back to find the first link obviously (by type) signed by an admin. Only FullVerify team.leave links _after_ that link. So inductively admins checked the stuff. Put link in teamMap even when not doing FullVerify, since at least an admin checked the SigID.

go/teams/loader2.go
```diff
@@ -129,6 +130,21 @@ func (l *TeamLoader) loadUserAndKeyFromLinkInner(ctx context.Context,
...
+func (l *TeamLoader) loadUserAndKeyFromLinkInnerNoVerify(ctx context.Context,
+	link *chainLinkUnpacked) (signerUV keybase1.UserVersion, err error) {
+	defer l.G().CTraceTimed(ctx, fmt.Sprintf("TeamLoader#loadUserAndKeyFromLinkInnerNoVerify(%d)", int(link.inner.Seqno)), func() error { return err })()
+	keySection := link.inner.Body.Key
+	if keySection == nil {
+		return signerUV, libkb.NoUIDError{}
+	}
+	// Use the UID from the link body and EldestSeqno from the server-trust API response.
+	if link.source.EldestSeqno == 0 {
+		return signerUV, fmt.Errorf("missing server hint for team sigchain link signer")
```
maxtaco: shouldn't we revert to full verify in this case?

go/teams/loader2.go
```diff
@@ -195,26 +211,44 @@ func (l *TeamLoader) verifyLink(ctx context.Context,
...
-	if err != nil {
-		return nil, err
-	}
+	// FullVerify all links except for `team.leave` links for which there is
+	// a admin link later in the chain.
+	// Such a link has effectively been verified for us by the admin who signed on top.
+	// This trick can be used on `team.leave` links because they do not add admins.
+	fullVerify := (link.LinkType() != libkb.SigchainV2TypeTeamLeave) ||
+		(link.Seqno() >= fullVerifyCutoff) ||
+		(link.source.EldestSeqno == 0)
```
maxtaco: ah i guess it's handled here...

go/teams/loader.go
```diff
@@ -524,6 +524,19 @@ func (l *TeamLoader) load2InnerLockedRetry(ctx context.Context, arg load2ArgT) (
 		ret = &tmp
 	}
 
+	// A link which was signed by an admin. Sloppily the latest such link.
+	// Sloppy because this calculation misses out on e.g. a rotate_key signed by an admin.
+	// This value is used for skipping fullVerify on team.leave links, see `verifyLink`.
+	var fullVerifyCutoff keybase1.Seqno
+	for i := len(links) - 1; i >= 0; i-- {
+		l.G().Log.CDebugf(ctx, "xxx %v :: %v", i, len(links))
```
maxtaco: is this intended to stay in?

mlsteele: arghxxx no

go/teams/loader2.go
maxtaco: Comment we should never hit this case?

go/teams/loader2.go
```diff
@@ -195,26 +211,44 @@ func (l *TeamLoader) verifyLink(ctx context.Context,
 		return nil, err
 	}
 
-	signerUV, key, linkMap, err := l.loadUserAndKeyFromLinkInner(ctx, *link.inner)
-	if err != nil {
-		return nil, err
-	}
+	// FullVerify all links except for `team.leave` links for which there is
+	// a admin link later in the chain.
```
maxtaco: `an admin link`

PR #12760, opened 2018-07-10
[KBFS-3131] Fix username text in Files list

jzila:
@buoyad I changed the default click handler on the usernames control to never open the tracker popup (instead, it now always opens the profile). If there are unforeseen consequences to that LMK, but as far as I understand that's what we want anyway. Tracker popups are for when an external activity causes an identify that needs your attention.

cc @cecileboucheron

### DIFF
--- a/shared/common-adapters/usernames.js
+++ b/shared/common-adapters/usernames.js
@@ -175,8 +175,7 @@ const mapStateToProps = (state: TypedState) => {
 }
 
 const mapDispatchToProps = (dispatch: Dispatch) => ({
-  _onUsernameClicked: (username: string) =>
-     dispatch(createShowUserProfile({username}))
+  _onUsernameClicked: (username: string) => dispatch(createShowUserProfile({username})),
 })
 
 const mergeProps = (stateProps, dispatchProps, ownProps) => {


shared/fs/common/path-item-info.js
```diff
@@ -57,20 +57,19 @@ const PathItemInfo = (props: Props) => (
     {props.lastWriter ? (
       <Text type="BodySmall" style={writerTextStyle} lineClamp={isMobile ? 1 : undefined}>
         &nbsp;by&nbsp;
-        <Text type="BodySmall" style={writerStyle}>
-          {props.lastWriter}
-        </Text>
+        <ConnectedUsernames
```
buoyad: can you add `underline={true}`?

buoyad: can you add `underline={true}`?

jzila: The design doesn't have an underline.

cecileboucheron: I don't know if that's what Danny is referring to but it should be underlined on hover on desktop.

jzila: Understood, thanks @cecileboucheron!

PR #12769, opened 2018-07-10
add a HOC for smoothing count down timer to 1s resolution and make sure the upload banner shows up for at least 2s

songgao:


shared/fs/footer/upload-countdown-hoc.js
```diff
@@ -0,0 +1,160 @@
...
+      */
+          break
+      }
+    }
+
+    _startTicker = () => {
+      if (this._tickerID) {
+        return
+      }
+      this._tickerID = this.props.setInterval(this._tick, tickInterval)
```
chrisnojima: likely want to call this._stopTicker here

shared/fs/footer/upload-countdown-hoc.js
```diff
@@ -0,0 +1,160 @@
...
+  class extends React.PureComponent<Props, State> {
+    state = {
+      displayDuration: 0,
+      mode: 'hidden',
+      glueTTL: 0,
+    }
+
+    _tickerID: ?IntervalID = null
+
+    _tick = () => {
```
chrisnojima: you need to protect against unmount here

shared/fs/footer/upload-countdown-hoc.js
```diff
@@ -0,0 +1,160 @@
...
+      const {mode, glueTTL, displayDuration} = this.state
```
chrisnojima: this is sort of dangerous, setState is async so you reading it and then calling setState won't necessarily be correct all the time, likely safer to use
```
this.setState(prevState => {}) 
```
form

shared/fs/footer/upload-countdown-hoc.js
songgao: I thought the TimerHOC already did that. Is it not enough?

shared/fs/footer/upload-countdown-hoc.js
chrisnojima: oh i didn't realize it was composed with that, tnx

shared/fs/footer/upload-countdown-hoc.js
chrisnojima: oh i didn't realize it was composed with that, tnx

shared/fs/footer/upload-countdown-hoc.js
songgao: I meant this call to be idempotent, i.e. start the ticker if it hasn't been started. The `if` condition above already covers the case where we already have a ticker ID. I'll add a comment to clarify.

songgao:
@chrisnojima ready for another look. Thanks!

songgao:
@chrisnojima ping 🏓

PR #12773, opened 2018-07-10
Improve revoke device page

nathunsmitty:
A handful of minor things:

- Adds a `<ProgressIndicator />` component while the client waits to receive a list of all endangered TLFs.
- Adds a storybook component for the revoke device page.
- Some general refactoring with props/styles/types.
- Makes endangered TLF list text selectable.

@akalin-keybase's original ticket also mentioned potentially making the list of endangered TLFs links. I think this may be best implemented with a `<Files />` component embedded within the page, e.g.,
screen shot 2018-07-10 at 6 11 13 pm
I didn't implement this feature in this PR because it looked like it would involve some refactor of the files component if we did it that way and I wanted to get some feedback first.

I also want to note that I couldn't get any endangered TLFs to occur when I tried to reproduce a situation in which they would occur. This could be due to an error on my part, though.

Currently, we don't show any type of special warning or list of files when a user revokes their current and only device. I'm probably going to file a ticket for that.


shared/devices/device-page/index.js
```diff
@@ -76,9 +77,15 @@ const Render = (props: Props) => {
     metaOne = <Meta title="revoked" style={styles.meta} backgroundColor={globalColors.red} />
   }
 
+  const icon: IconType = {
+    backup: 'icon-paper-key-64',
+    desktop: 'icon-computer-64',
+    mobile: 'icon-phone-64',
```
songgao: Are there no 48 versions of these like the revoke ones?

shared/devices/device-revoke/index.js
```diff
@@ -71,33 +101,34 @@ const DeviceRevoke = (props: Props) => (
...
+    alignItems: 'flex-start',
+    alignSelf: 'center',
+    borderWidth: 1,
+    borderStyle: 'solid',
+    borderColor: globalColors.black_05,
+    borderRadius: 4,
+    height: 162,
+    marginBottom: globalMargins.small,
+    marginTop: globalMargins.small,
+    // overflowY: 'scroll',
```
songgao: Did you mean to remove this line?

nathunsmitty: Nope! Good catch!

shared/devices/device-page/index.js
chrisnojima: in zeplin its 64 on both

shared/devices/device-revoke/index.js
```diff
@@ -1,31 +1,49 @@
...
+      <Icon type={icon} />
+      <Text type="BodySemibold" style={styles.name}>
+        {name}
+      </Text>
+    </Box>
+  )
+}
 
-const Body = ({endangeredTLFs, name, currentDevice}) => (
+const Body = ({
```
chrisnojima: we're moving towards not destructuring these props as parameters, it just makes the file take a lot of vertical space for very little
instead
```
const Body  = (props: Props) => (
... this.props.waiting? ...
)
```


or if you use a prop a bunch (usually unlikely) you can just destructure that one prop
```
const Body = (props: Props) => {
  const {waiting} = props
...
}
```

shared/devices/device-revoke/index.js
```diff
@@ -35,34 +53,46 @@ const Body = ({endangeredTLFs, name, currentDevice}) => (
...
     )}
   </Box>
 )
 
 const DeviceRevoke = (props: Props) => (
   <Confirm
     body={
-      <Body endangeredTLFs={props.endangeredTLFs} name={props.name} currentDevice={props.currentDevice} />
+      <Body
+        // endangeredTLFs={props.endangeredTLFs}
```
chrisnojima: this has to go back

shared/devices/device-revoke/index.js
```diff
@@ -35,34 +53,46 @@ const Body = ({endangeredTLFs, name, currentDevice}) => (
...
+        endangeredTLFs={['nathunsmitty']}
```
chrisnojima: 🔥 🔥 🔥

shared/devices/device-revoke/index.js
```diff
@@ -1,31 +1,49 @@
...
-  </Box>
-)
+const Header = ({name, type}: {name: string, type: DeviceType}) => {
+  const icon: IconType = {
+    backup: isMobile ? 'icon-paper-key-revoke-64' : 'icon-paper-key-revoke-48',
+    desktop: isMobile ? 'icon-computer-revoke-64' : 'icon-computer-revoke-48',
+    mobile: isMobile ? 'icon-phone-revoke-64' : 'icon-phone-revoke-48',
+  }[type]
+  return (
+    <Box style={{...globalStyles.flexBoxColumn, alignItems: 'center'}}>
```
chrisnojima: should move to styles

shared/devices/device-revoke/index.js
nathunsmitty: Oh actually it looks like `overflowY` is not a available react native value and `overflow` has [some limitations](https://facebook.github.io/react-native/docs/layout-props.html#overflow). Anyone have any suggestions on how to get the type of scrolling behavior we get on desktop on mobile also then?

chrisnojima: just use a <ScrollView> from common adapters, you can't style overflow in RN like desktop

shared/devices/device-revoke/index.js
nathunsmitty: Generally, are there any cases where styles should be defined inline like this or are we moving to the practice of putting all styles in the `styles` object?

chrisnojima: ideally they'd all be in there. plus you already have a styleSheetCreate so might as well put it in

nathunsmitty:
@keybase/react-hackers bumping this

nathunsmitty:
@keybase/react-hackers go for another look over!

shared/common-adapters/list.desktop.js
```diff
@@ -36,21 +36,9 @@ class List extends PureComponent<Props<any>, void> {
...
-        <div style={globalStyles.fillAbsolute}>
-          <div
-            style={{
-              height: '100%',
-              overflowY: 'auto',
-              width: '100%',
-            }}
-          >
+      <div style={{flexGrow: 1, position: 'relative', ...this.props.style}}>
+        <div style={collapseStyles([globalStyles.fillAbsolute, {padding: 'inherit'}])}>
```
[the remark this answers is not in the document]
chrisnojima: lets discuss

shared/common-adapters/list.native.js
```diff
@@ -30,13 +30,7 @@ class List extends PureComponent<Props<any>, void> {
...
   render() {
     return (
-      <View
-        style={{
-          flexGrow: 1,
-          position: 'relative',
-          ...this.props.style,
-        }}
-      >
+      <View style={collapseStyles([styles.outerView, this.props.style])}>
```
nathunsmitty: I was getting errors with spreading `this.props.style` on mobile, so it seemed like a good opportunity to do some minor refactoring using `collapseStyles` and `styleSheetCreate`.

shared/devices/device-revoke/index.js
```diff
@@ -71,33 +100,56 @@ const DeviceRevoke = (props: Props) => (
...
+      borderWidth: 1,
+      marginBottom: globalMargins.small,
+      marginTop: globalMargins.small,
+    },
+    isElectron: {
+      height: 162,
+      width: 440,
+    },
+    isMobile: {
+      height: 176,
```
nathunsmitty: The endangeredTLFs list is currently kinda buggy and I think height was set sort of implicitly before, so I decided to just set a list height of 176, which looks pretty good on all phone sizes.

shared/common-adapters/list.native.js
chrisnojima: 👍 we definitely want all the common-adapters to be able to take ids

shared/devices/device-revoke/index.js
chrisnojima: can you show me this

chrisnojima:
also ci isn't passing

shared/devices/device-revoke/index.js
```diff
@@ -1,103 +1,195 @@
...
-              </Text>
-              <Text type="BodySemibold">{tlf}</Text>
-            </Box>
-          ))}
-        </Box>
-      </Box>
-    )}
-  </Box>
+const ActionButtons = (props: {onCancel: () => void, onSubmit: () => void, waiting: boolean}) => (
+  <Box2 direction={isMobile ? 'vertical' : 'horizontalReverse'} style={styles.actionButtonsContainer}>
```
nathunsmitty: This is the reason why I added `horizontalReverse` and `verticalReverse` – gives us the delete button on top in mobile and to the right on desktop!

nathunsmitty:
@chrisnojima I know you approved, but may want to take a look at the row-reverse Box2 stuff

shared/common-adapters/box.desktop.js
```diff
@@ -27,9 +27,27 @@ const injectGaps = (Component, _children, gap, gapStart, gapEnd) => {
...
   render() {
-    const horizontal = this.props.direction === 'horizontal'
+    let directionStyle
+    let horizontal = this.props.direction === 'horizontal' || this.props.direction === 'horizontalReverse'
+    switch (this.props.direction) {
+      case 'horizontal':
+        directionStyle = globalStyles.flexBoxRow
+        break
+      case 'horizontalReverse':
+        directionStyle = {display: 'flex', flexDirection: 'row-reverse'}
```
chrisnojima: this has to pull from styles, not make a new object every time or you'll re render over and over

shared/common-adapters/box.desktop.js
```diff
@@ -62,13 +80,7 @@ const styles = {
...
   fullHeight: {height: '100%'},
   fullWidth: {width: '100%'},
-  hbox: {
-    ...globalStyles.flexBoxRow,
-    alignItems: 'stretch',
-    justifyContent: 'flex-start',
-  },
-  vbox: {
-    ...globalStyles.flexBoxColumn,
+  commonDirectionStyles: {
```
chrisnojima: this isn't supposed to be common, its different vertical / horizontal

shared/common-adapters/box.desktop.js
```diff
@@ -27,9 +27,27 @@ const injectGaps = (Component, _children, gap, gapStart, gapEnd) => {
...
+    let directionStyle
+    let horizontal = this.props.direction === 'horizontal' || this.props.direction === 'horizontalReverse'
+    switch (this.props.direction) {
+      case 'horizontal':
+        directionStyle = globalStyles.flexBoxRow
+        break
+      case 'horizontalReverse':
+        directionStyle = {display: 'flex', flexDirection: 'row-reverse'}
+        break
+      case 'verticalReverse':
```
chrisnojima: it should just map from the 4 values to 4 things in the stylesheet create

chrisnojima:
need to add these cases to box.stories.js

nathunsmitty:
@keybase/react-hackers this could use some 👀

shared/devices/device-revoke/index.js
```diff
@@ -1,103 +1,195 @@
...
+    <Button
+      fullWidth={!!isMobile}
```
chrisnojima: can jsut be isMobile, its already a boolean

shared/devices/device-revoke/index.js
```diff
@@ -1,103 +1,195 @@
...
+      type="Danger"
+      onClick={props.waiting ? null : props.onSubmit}
+      label="Yes, delete it"
+      disabled={!!props.waiting}
```
chrisnojima: same here

shared/devices/device-revoke/index.js
```diff
@@ -1,103 +1,195 @@
...
+      style={isMobile ? {marginBottom: globalMargins.tiny} : {marginLeft: globalMargins.tiny}}
```
chrisnojima: cant you just use gapEnd in the Box2 parent

shared/devices/device-revoke/index.js
```diff
@@ -1,103 +1,195 @@
...
-    onSubmit={props.waiting ? null : props.onSubmit}
-    disabled={!!props.waiting}
-    submitLabel="Yes, delete it"
-    theme="public"
-  />
+  <Box2 direction="vertical" fullHeight={true} fullWidth={true}>
+    <Box style={styles.deviceRevokeContainer}>
+      <Header name={props.name} type={props.type} />
+      <BodyText name={props.name} currentDevice={props.currentDevice} />
+      <EndangeredTLFList endangeredTLFs={props.endangeredTLFs} waiting={props.waiting} style={{flex: 1}} />
```
chrisnojima: use styles below

nathunsmitty:
@chrisnojima made your fixes!

PR #12784, opened 2018-07-11
Deleted messages don't count towards page limit

mlsteele:
Requires https://github.com/keybase/keybase/pull/2665

With one local repro this seems to fix the problem. Not sure if there are other pathways though.
- `db nuke -f` and start (solved by [server patch](https://github.com/keybase/keybase/pull/2665))
- warm cache (solved by this client patch)

### DIFF
--- a/go/chat/convsource.go
+++ b/go/chat/convsource.go
@@ -449,6 +449,8 @@ type HybridConversationSource struct {
 	numExpungeReload int
 	storage          *storage.Storage
 	lockTab          *conversationLockTab
+
+	blackoutPullForTesting bool
 }
 
 var _ types.ConversationSource = (*HybridConversationSource)(nil)
@@ -737,6 +739,9 @@ func (s *HybridConversationSource) Pull(ctx context.Context, convID chat1.Conver
 			if err = s.postProcessThread(ctx, uid, conv, &thread, query, nil, true, true); err != nil {
 				return thread, err
 			}
+			if s.blackoutPullForTesting {
+				thread.Messages = nil
+			}
 			return thread, nil
 		}
 		s.Debug(ctx, "Pull: cache miss: err: %s", err.Error())
@@ -782,6 +787,9 @@ func (s *HybridConversationSource) Pull(ctx context.Context, convID chat1.Conver
 	// Run post process stuff
 	if err = s.postProcessThread(ctx, uid, unboxConv, &thread, query, nil, true, true); err != nil {
 		return thread, err
+	}
+	if s.blackoutPullForTesting {
+		thread.Messages = nil
 	}
 	return thread, nil
 }
--- a/go/chat/convsource.go
+++ b/go/chat/convsource.go
@@ -25,6 +25,8 @@ type baseConversationSource struct {
 
 	boxer *Boxer
 	ri    func() chat1.RemoteInterface
+
+	blackoutPullForTesting bool
 }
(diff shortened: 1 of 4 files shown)


mlsteele:
Looks like the experiment went ok so this and https://github.com/keybase/keybase/pull/2665 are ready for review

go/chat/convsource.go
```diff
@@ -449,6 +449,8 @@ type HybridConversationSource struct {
 	numExpungeReload int
 	storage          *storage.Storage
 	lockTab          *conversationLockTab
+
+	blackoutPullForTesting bool
```
mmaxim: Is there any other way?

go/chat/sender_test.go
```diff
@@ -1042,14 +1047,8 @@ func TestPrevPointerAddition(t *testing.T) {
...
 			world.Fc.Advance(time.Second*time.Duration(*ephemeralLifetime) + chat1.ShowExplosionLifetime)
-			// Mock out server call for new messages
-			ri.GetThreadRemoteFunc = func(m *kbtest.ChatRemoteMock, ctx context.Context, arg chat1.GetThreadRemoteArg) (chat1.GetThreadRemoteRes, error) {
-				return chat1.GetThreadRemoteRes{
-					Thread: chat1.ThreadViewBoxed{
-						Pagination: &chat1.Pagination{},
-					},
-				}, nil
-			}
+			// Mock out pulling messages to return no messages
```
mmaxim: Actually why did you do this? I don't like all that random code in convsource

mlsteele: This test depends on `ConvSource` returning zero messages from a `Pull` (line 1068). Previously mocking `GetThreadRemoteFunc` seemed to be enough. Note that the pagination number is set to 2 for this test (line 197). 

This patch changes the definition of pagination. Meaning that even with the count set to 2 _any fetch_ will fill in the whole conv in cache in this test since a bunch of messages are deleted, and so do not count towards the paging limit. In addition to whatever caching mechanisms there are background processes that fill in cache. Those background processes would have to be shut off for this test to work. The background processes include the ephemeral doohicky, and idk whether BackgroundLoader is running or not. But it's easier just to mock out `Pull`. It didn't seem too far fetched since the existing mocking out of `GetThreadRemoteFunc` was not real behavior anyway.

cc @joshblum

mmaxim: Ok at least stick it in `postProcessThread` then, so that stuff is only in one place. I just dislike having too much `if <I am testing>` code hanging around.

mlsteele: Can do. What do you mean in one place? Where's the `if <I am testing>` in [`postProcessThread`](https://github.com/keybase/client/blob/master/go/chat/convsource.go#L73)?

mlsteele: Oh I see that it's the common callee

mmaxim:
This is going to mess with how placeholders work, since they no longer count now in the simple collector. Why don't `IsValid` false messages count here?

mlsteele:
What are placeholders? If they're not `IsValidDeleted` then they'd still be counted. `IsValid->false` messages _do_ still count in the collectors. 

```
// IsValidDeleted returns whether a message is valid and has been deleted.
```

mmaxim:
Ah I see, I got lost in the `!`'s across the functions

PR #12787, opened 2018-07-12
[KBFS-2986] Fix broken Windows viewing in Explorer

jzila:


jzila:
Tested on Windows and it now works. cc @zanderz

jzila:
@keybase/react-hackers ping

PR #12793, opened 2018-07-12
serve pending attachment preview from the http server, and store encrypted CORE-7689

mmaxim:
Main goal of this patch is to remove the use of temp files in the top level cache directory for the pending image previews for chat attachments. Instead, the server produces a URL from `MakePreview` and the UI can source the image data from that URL. Specific:

1.) Expand the `encrypteddb` package to allow for easily saving encrypted files using the chat storage encryption key.
2.) Introduce `storage.PendingPreviews` that uses `encrypteddb.File` to store all the pending previews.
3.) Change `AttachmentHTTPSrv` to have a new endpoint to return pending attachment previews. All the previews are indexed solely by outbox ID. 
4.) Upon receiving a message in `HybridConversationSource` with an outbox ID in the pending file store, we remove it since we can get it from S3 now.

cc @chrisnojima


mmaxim:
Also note this patch is setting up the ability for us to be able to render pending attachments from the chat outbox.

PR #12796, opened 2018-07-12
When canceling out of a pending convo, find a new one to replace it with

cjb:
@keybase/react-hackers 

Fixes:

1) go to new conversation, choose someone you have no existing convo with

2) click the X next to the pending convo in the chat inbox

3) now you see "You haven't chatted with (blank) before" instead of a replacement convo.

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -1376,7 +1376,8 @@ const _maybeAutoselectNewestConversation = (
     action.payload.conversationIDKey === selected
   ) {
     // Intentional fall-through -- force select a new one
-  } else if (selected !== Constants.noConversationIDKey) {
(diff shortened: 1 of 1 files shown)


shared/actions/chat2/index.js
```diff
@@ -1376,7 +1376,8 @@ const _maybeAutoselectNewestConversation = (
     action.payload.conversationIDKey === selected
   ) {
     // Intentional fall-through -- force select a new one
-  } else if (selected !== Constants.noConversationIDKey) {
+  } else if (selected !== Constants.noConversationIDKey && selected !== Constants.pendingConversationIDKey) {
```
chrisnojima: maybe use this: https://github.com/keybase/client/blob/master/shared/constants/types/chat2/common.js#L23

cjb:
@chrisnojima Thanks, works.

PR #12797, opened 2018-07-12
Support EXIF Orientation for Attachment Previews

thebearjew:
### Problem

In #12482, support for reading and rotating images according to EXIF Orientation metadata was added. However, during the upload preview stage, the preview image was being loaded from the user's local filesystem, preventing us from reading the image data via JavaScript.

This also became an issue for Avatar uploads via the desktop client #12306 whereby previews of avatars (during zoom/cropping stage) would not be oriented correctly.

### Solution

When `OrientedImage` has `preview={true}` Use `fs#readFile` to load the image file into an`ArrayBuffer` then pass the buffer to `EXIF#readFromBinaryFile` which can operate on the bytes directly which is highly performant.

### Notes

1. Performing `fs.readFile` followed by `EXIF#readFromBinaryFile` is very fast. For a 20mb image it took 8.3ms.
2. If the image is not `jpeg` then EXIF will return false and `NO_TRANSFORM` will be applied.

### DIFF
--- a/shared/chat/conversation/attachment-get-titles/index.js
+++ b/shared/chat/conversation/attachment-get-titles/index.js
@@ -110,8 +110,7 @@ class GetTitles extends React.Component<Props, State> {
             <Box style={{...globalStyles.flexBoxCenter, height: 150, width: 150}}>
               {info.type === 'image' ? (
                 <OrientedImage
-                  preview={true}
-                  onLoad={() => {}}
+                  localFile={true}
                   src={path}
                   style={isMobile ? {height: 150, width: 150} : {maxHeight: '100%', maxWidth: '100%'}}
                 />
--- a/shared/chat/conversation/attachment-get-titles/index.js
+++ b/shared/chat/conversation/attachment-get-titles/index.js
@@ -4,7 +4,6 @@ import {
   Box,
   Button,
   Icon,
-  // Image,
   OrientedImage,
   Input,
   PopupDialog,
--- a/shared/chat/conversation/attachment-get-titles/index.js
+++ b/shared/chat/conversation/attachment-get-titles/index.js
@@ -4,7 +4,8 @@ import {
   Box,
   Button,
   Icon,
-  Image,
+  // Image,
+  OrientedImage,
   Input,
   PopupDialog,
   Text,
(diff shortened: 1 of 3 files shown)


shared/chat/conversation/attachment-get-titles/index.js
```diff
@@ -109,7 +109,9 @@ class GetTitles extends React.Component<Props, State> {
           <Box style={isMobile ? stylesMobile : stylesDesktop}>
             <Box style={{...globalStyles.flexBoxCenter, height: 150, width: 150}}>
               {info.type === 'image' ? (
-                <Image
+                <OrientedImage
+                  preview={true}
+                  onLoad={() => {}}
```
chrisnojima: can we make onLoad be optional so we don't need to do this

shared/common-adapters/oriented-image.types.js.flow
```diff
@@ -5,4 +5,5 @@ export type Props = {
   src: string,
   style?: StylesDesktop,
   onLoad: () => void,
+  preview?: boolean,
```
chrisnojima: maybe this is more accurate to say localFile or something

shared/chat/conversation/attachment-get-titles/index.js
thebearjew: Got some strange Flow errors last time, I'll give it another try.

shared/common-adapters/oriented-image.types.js.flow
thebearjew: Updated!

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -86,20 +141,14 @@ class OrientedImage extends React.Component<Props, State> {
...
-
-    try {
-      EXIF.getData({src}, function() {
-        if (!_hasComponentMounted()) return
-        handleData(this)
-      })
-    } catch (_) {
-      // Mark src as null in the cache to avoid making subsequent calls.
-      _cacheStyleTransforms[src] = NO_TRANSFORM
+    if (this.props.preview) {
```
buoyad: Since `_readExifLocal` is always for local files and `_fetchExifUploaded` is always for 'remote' files, can this component figure out which to do based on `src` rather than a prop?

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -74,6 +88,42 @@ class OrientedImage extends React.Component<Props, State> {
...
+      })
+      if (!ret) reject(new Error('EXIF failed to fetch image data'))
+    })
+  }
+
+  // Read the file contents directly into a buffer and pass it to EXIF which
+  // can extract the EXIF data
+  _readExifLocal = src => {
+    return new Promise((resolve, reject) => {
+      fs.readFile(src, (err, data) => {
```
buoyad: you can use `readFileSync` to avoid this extra level of nesting.

PR #12801, opened 2018-07-12
blast fs store on logout

songgao:


shared/actions/fs/index.js
```diff
@@ -536,7 +537,10 @@ const letResetUserBackIn = ({payload: {id, username}}: FsGen.LetResetUserBackInP
 
 const letResetUserBackInResult = () => undefined // Saga.put(FsGen.createLoadResets())
 
+const resetStore = () => Saga.put({payload: undefined, type: FsGen.resetStore})
```
jzila: Do you need the explicit payload?

songgao: Seems not; I just copied this from existing code without trying to remove it. I'll push a commit to get rid of the `payload` filed.

PR #12802, opened 2018-07-13
limit download cards to 3 and add button to open download folder

songgao:
screen shot 2018-07-12 at 5 45 04 pm


shared/fs/footer/download.js
```diff
@@ -52,6 +52,7 @@ const stylesDownload = memoize((errored: boolean) => ({
   justifyContent: 'flex-start',
   marginLeft: globalMargins.xtiny,
   width: 140,
+  minWidth: 140, // needed to avoid shrinking
```
chrisnojima: What shrinking? Can you use flexShrink 0?

shared/fs/footer/downloads.js
```diff
@@ -0,0 +1,68 @@
...
+import {globalStyles, globalMargins, globalColors, styleSheetCreate, isMobile} from '../../styles'
+import {Box2, Box, Icon, ClickableBox} from '../../common-adapters'
+import Download from './download-container'
+
+export type DownloadsProps = {
+  downloadKeys: Array<string>,
+  openDownloadFolder?: () => void,
+}
+
+const maxNumCards = isMobile ? 1 : 3
```
chrisnojima: Just limit the cards in the connector

shared/fs/footer/download.js
songgao: When building in storybook, if I resize the window too small, the download cards shrinks. But I'm realizing now the real Electron window doesn't go that small, so I'll remove this line.

shared/fs/footer/downloads.js
songgao: The component also decides whether to show the "..." button. If we do that container, we'd need an additional prop for whether to show the "..." button, so component seemed like a better place to me. Would you mind explaining why it's better to put it in container here?

chrisnojima: i think its better if the components are dumb and the containers are smart. especially if the component is split into native/desktop
if it was me (you don't have to change this, but)
I'd make
```
cards: Array<whatever>
thereAreMoreCard: boolean 
```
or something, that encodes all you need

songgao: OK; will do. Thanks!

PR #12807, opened 2018-07-13
client side of device clone detection

xgess:
This is part of an effort to identify and notify about possible device clones. Periodically (currently every 6 hours) the client will send two tokens (`prior`, and `stage`) to the server, and based on its response, the client will decide whether or not to notify about a possible clone. Here's what the token interaction looks like from the client's perspective:
1. Read [`prior`, `stage`] off persistent disk
2. If `stage` is `""` generate a new token for `stage` and write [`prior`, `stage`] to disk.
3. Post [`prior`, `stage`] to the server. 
4. In the happy path, the server will update the token it's saving for this device from `prior` to `stage`. If there's a clone, we'll see one `prior` become multiple `stage`s. 
5. The client then updates it's persistent disk from [`prior`, `stage`] to [`stage`, `""`] and sleeps until the next run.

This PR includes all of the token interaction/persistence and an RPC notification to the frontend.


mmaxim:
I am going to xfer this review to @mlsteele since I think he has been following along closely.

go/libkb/device_clone_token.go
```diff
@@ -0,0 +1,45 @@
...
+
+package libkb
+
+import "fmt"
+
+const (
+	DefaultCloneTokenValue string = "00000000000000000000000000000000"
+)
+
+type DeviceCloneToken struct {
```
mlsteele: `DeviceCloneToken` strikes me as a strange name for something that is not a single token. Maybe something like `DeviceCloneTokens` or `DeviceCloneState` or `CloneState`.

go/libkb/device_clone_token.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (f JSONConfigFile) GetDeviceCloneToken(un NormalizedUsername) DeviceCloneToken {
+	pPath, sPath, cPath := deviceCloneTokenPaths(un)
+	p, _ := f.GetStringAtPath(pPath)
+	s, _ := f.GetStringAtPath(sPath)
+	c, _ := f.GetIntAtPath(cPath)
+	return DeviceCloneToken{Prior: p, Stage: s, Clones: c}
+}
+
+func (f *JSONConfigFile) SetDeviceCloneToken(un NormalizedUsername, d DeviceCloneToken) error {
```
mlsteele: Let's set these atomically. Is it crash safe if saved in this order?

go/libkb/notify_router.go
```diff
@@ -523,6 +525,35 @@ func (n *NotifyRouter) HandleFavoritesChanged(uid keybase1.UID) {
 	n.G().Log.Debug("- Sent favorites changed notification")
 }
 
+// HandleDeviceCloneNotification is called whenever
+// ALEX: add better comment here
```
mlsteele: yup

go/service/main.go
```diff
@@ -639,6 +639,23 @@ func (d *Service) hourlyChecks() {
...
 
+func (d *Service) deviceCloneSelfCheck() error {
+	eng := engine.NewDeviceCloneTokenEngine(d.G())
+	m := libkb.NewMetaContextBackground(d.G())
+
+	beforeClones := eng.ClonesCount(m)
+	if err := engine.RunEngine2(m, eng); err != nil {
+		return err
+	}
+	newClones := eng.ClonesCount(m) - beforeClones
```
mlsteele: Using an Engine this way is unusual. In general an Engine is a glorified function so its lifecycle is construct, Run, and optionally pull more results off it with methods. Can you get this information directly through `G.Env.GetConfig().GetDeviceCloneToken(un).Clones`?

maxtaco: Agreed that it seems possible to put all code for this in one place!

go/engine/device_clone_token.go
```diff
@@ -0,0 +1,111 @@
...
+			"stage":     libkb.S{Val: s},
+		},
+		NetContext:     m.Ctx(),
+		AppStatusCodes: []int{libkb.SCOk},
+	}
+	res, err := m.G().API.Post(arg)
+	if err != nil {
+		return err
+	}
+	persistedToken, err := res.Body.AtKey("token").GetString()
```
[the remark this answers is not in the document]
xgess: allows slight changing of the protocol (in some ways) directly from the server without needing to touch the client. i thought it was more important that the device save the thing that server told it (for confidence in the next request) than that the device assume it knows exactly what the server is doing with that data. it really shouldn't matter. just seemed a little more flexible this way.

go/engine/device_clone_token.go
```diff
@@ -0,0 +1,111 @@
...
+	if count < 1 {
+		count = 1
+	}
+	return count
+}
+
+// Run starts the engine.
+func (e *DeviceCloneTokenEngine) Run(m libkb.MetaContext) error {
+	d := e.getToken(m)
+	p, s := d.Prior, d.Stage
```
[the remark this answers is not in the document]
xgess: i'm definitely partial to longer, more descriptive variable names, so that's one comment I'm always happy to get. i'm still trying to feel out `go` styles. any insights there are welcome.

go/engine/device_clone_token.go
```diff
@@ -0,0 +1,111 @@
...
+	if p == "" {
+		//first run
+		p = libkb.DefaultCloneTokenValue
+	}
+	if s == "" {
+		buf := make([]byte, 16)
+		rand.Read(buf)
+		s = hex.EncodeToString(buf)
```
maxtaco: we have helper functions for this.  see libkb/utils.go

go/engine/device_clone_token.go
```diff
@@ -0,0 +1,111 @@
...
+func (e *DeviceCloneTokenEngine) Run(m libkb.MetaContext) error {
+	d := e.getToken(m)
+	p, s := d.Prior, d.Stage
+	if p == "" {
+		//first run
+		p = libkb.DefaultCloneTokenValue
+	}
+	if s == "" {
+		buf := make([]byte, 16)
+		rand.Read(buf)
```
maxtaco: should handle the error code here.

go/engine/device_clone_token.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	res, err := m.G().API.Post(arg)
+	if err != nil {
+		return err
+	}
+	persistedToken, err := res.Body.AtKey("token").GetString()
+	if err != nil {
+		return err
+	}
+	clones, err := res.Body.AtKey("clones").GetInt()
```
maxtaco: can you use the `PostDecode` method of the `API` interface?  In general, we've been trying to do that going forward, rather than using `jsonw`...

go/libkb/device_clone_token.go
```diff
@@ -0,0 +1,45 @@
...
+	DefaultCloneTokenValue string = "00000000000000000000000000000000"
+)
+
+type DeviceCloneToken struct {
+	Prior  string
+	Stage  string
+	Clones int
+}
+
+func deviceCloneTokenPaths(un NormalizedUsername) (p, s, c string) {
```
maxtaco: Is it possible to store these things in LevelDB?  It'd prefer that over a whole new set of configuration files.  It's my understanding that the protocol should work even if these tokens are blasted away (since it's usually OK to blast away LevelDB).

go/libkb/interfaces.go
```diff
@@ -176,6 +176,7 @@ type ConfigReader interface {
 	GetUpdateLastChecked() keybase1.Time
 	GetUpdateURL() string
 	GetUpdateDisabled() (bool, bool)
+	GetDeviceCloneToken(NormalizedUsername) DeviceCloneToken
```
maxtaco: storing in levelDB would also obviate the need for these changes.

maxtaco:
I talked it over with @xgess  and @mlsteele, and maybe the best thing to do is to introduce a new clone detection config file for this.  It just makes me very nervous to be constantly rewriting `config.json`

go/engine/device_clone_token.go
xgess: 👍

go/engine/device_clone_token.go
```diff
@@ -0,0 +1,111 @@
...
+	if p == "" {
+		//first run
+		p = libkb.DefaultCloneTokenValue
+	}
+	if s == "" {
+		buf := make([]byte, 16)
+		rand.Read(buf)
+		s = hex.EncodeToString(buf)
+		tmp := libkb.DeviceCloneToken{Prior: p, Stage: s, Clones: d.Clones}
+		err := e.persistToken(m, tmp)
```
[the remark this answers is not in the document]
xgess: 👍

go/engine/device_clone_token.go
xgess: 👍 yeah - this is definitely nicer

go/libkb/device_clone_token.go
xgess: i'm 95% sure that this could fail at any point and we'd be ok. the worst case scenario for crashing in the middle and the next run completing is simply that the notification would be delayed until then. i'll take a look at doing them atomically though. i'd rather future me (and anyone else) not have to figure that out again.

xgess:
i tried pushing shared json file functionality down into `json.go` but it exposed something weird we're doing with the mutex. as per discussion with @maxtaco, we can circle back and clean that up later because it's going to touch unrelated things. there's a commit and its reversion in the history of this PR if you're curious. or if I want to look at it in the future.

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+	Status AppStatus `json:"status"`
+	Token  string    `json:"token"`
+	Clones int       `json:"clones"`
+}
+
+func (d *cloneDetectionResponse) GetAppStatus() *AppStatus {
+	return &d.Status
+}
+
+func UpdateDeviceCloneState(m *MetaContext) (before int, after int, err error) {
```
maxtaco: Always pass `MetaContext` by value, and not by reference.

go/libkb/globals.go
```diff
@@ -353,6 +353,16 @@ func (g *GlobalContext) ConfigureConfig() error {
 func (g *GlobalContext) ConfigReload() error {
 	err := g.ConfigureConfig()
 	g.ConfigureUpdaterConfig()
+	g.ConfigureDeviceCloneState()
+	return err
+}
+
+func (g *GlobalContext) ConfigureDeviceCloneState() error {
+	c := NewDeviceCloneStateJSONFile(g)
+	err := c.Load(false)
```
maxtaco: why open this here?  why not just open it whenever you need to access it?

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+		return 0, 0, err
+	}
+
+	err = SetDeviceCloneState(m, DeviceCloneState{Prior: res.Token, Stage: "", Clones: res.Clones})
+	after = res.Clones
+	return
+}
+
+func configPaths(m *MetaContext) (p, s, c string) {
+	deviceID := m.G().ActiveDevice.deviceID
```
maxtaco: use the `DeviceID` accessor.

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+	// Zero out the TX so that we don't abort it in the defer()
+	tx = nil
+	return nil
+}
+
+func NewDeviceCloneStateJSONFile(g *GlobalContext) *DeviceCloneStateJSONFile {
+	return &DeviceCloneStateJSONFile{NewJSONFile(g, g.Env.GetDeviceCloneStateFilename(), "device clone state")}
+}
+
+func deviceCloneStateReader(m *MetaContext) (DeviceCloneStateJSONFile, error) {
```
maxtaco: call this `newDeviceCloneStateReader`

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+	f := NewDeviceCloneStateJSONFile(m.G())
+	err := f.Load(false)
+	return *f, err
+}
+
+func deviceCloneStateWriter(m *MetaContext) (*DeviceCloneStateJSONFile, error) {
```
maxtaco: call this `newDeviceCloneStateWriter`

go/libkb/device_clone_state.go
maxtaco: Here and everywhere else. The point of `MetaContext` is to wrap `context.Context`, which you should always pass by value, and represents the call tree for the current "context", whether an RPC call from electron, or a background job, or whatever.  It also wraps `*GlobalContext`, which is shared across all threads of computation and represents global state.  So that you want to pass by reference.  `MetaContext`, if passed by value, satisfies both.

xgess: ahh. interesting.

go/libkb/globals.go
xgess: It was just so I could put it in the ConfigReload method which gets called in a bunch of places. It seemed like the kind of thing that might be (or might one day) be called to naively and hopefully clean things up that might be in a bad state. But yes I agree that it doesn't at all seem necessary. still want me to remove it from here?

go/libkb/device_clone_state.go
xgess: 👍

go/libkb/device_clone_state.go
xgess: 👍

go/libkb/device_clone_state.go
xgess: 👍

go/libkb/globals.go
xgess: i'll remove it. that doesn't really seem like a compelling enough reason to have it here.

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+	after = res.Clones
+	return
+}
+
+func configPaths(m MetaContext) (p, s, c string) {
+	deviceID := m.G().ActiveDevice.DeviceID()
+	p = fmt.Sprintf("%s.prior", deviceID)
+	s = fmt.Sprintf("%s.stage", deviceID)
+	c = fmt.Sprintf("%s.clones", deviceID)
+	return
```
maxtaco: `return p,s,c`;  we try to never do "naked" returns anymore.

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	var res cloneDetectionResponse
+	err = m.G().API.PostDecode(arg, &res)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	err = SetDeviceCloneState(m, DeviceCloneState{Prior: res.Token, Stage: "", Clones: res.Clones})
+	after = res.Clones
+	return
```
maxtaco: `return 0, 0, nil`

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+	err := f.Load(false)
+	return f, err
+}
+
+func (f DeviceCloneStateJSONFile) GetStringAtPath(p string) (ret string, isSet bool) {
+	i, isSet := f.getValueAtPath(p, getString)
+	if isSet {
+		ret = i.(string)
+	}
+	return
```
maxtaco: `return ret, isSet`

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (f DeviceCloneStateJSONFile) GetIntAtPath(p string) (ret int, isSet bool) {
+	i, isSet := f.getValueAtPath(p, getInt)
+	if isSet {
+		ret = i.(int)
+	}
+	return
```
maxtaco: `return ret, isSet`

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (f DeviceCloneStateJSONFile) getValueAtPath(p string, getter valueGetter) (ret interface{}, isSet bool) {
+	var err error
+	ret, err = getter(f.jw.AtPath(p))
+	if err == nil {
+		isSet = true
+	}
+	return
```
maxtaco: `return ret, isSet`

go/service/main.go
```diff
@@ -650,6 +650,22 @@ func (d *Service) hourlyChecks() {
...
 
+func (d *Service) deviceCloneSelfCheck() error {
+	m := libkb.NewMetaContextBackground(d.G())
+	before, after, err := libkb.UpdateDeviceCloneState(m)
+	if err != nil {
+		return err
+	}
+	newClones := after - before
+
+	d.G().Log.Debug("deviceCloneSelfCheck: is there a new clone? %v", newClones > 0)
```
maxtaco: `m.CDebugf`

go/service/main.go
```diff
@@ -650,6 +650,22 @@ func (d *Service) hourlyChecks() {
...
+	if newClones > 0 {
+		d.G().Log.Debug("deviceCloneSelfCheck: notifying user %v -> %v restarts", before, after)
```
maxtaco: `m.CDebugf`

go/service/main.go
```diff
@@ -650,6 +650,22 @@ func (d *Service) hourlyChecks() {
 	}()
 }
 
+func (d *Service) deviceCloneSelfCheck() error {
+	m := libkb.NewMetaContextBackground(d.G())
```
maxtaco: `m = m.WithLogTag("CLONE")`

go/libkb/device_clone_state.go
xgess: cool

maxtaco: Yeah, I think the danger of "naked returns" comes with refactoring something higher up in the function, and then forgetting there's a naked return hanging out below.

xgess: yeah - that makes perfect sense.

go/service/main.go
xgess: oooo cool!

go/service/main.go
xgess: 👍

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+type DeviceCloneStateJSONFile struct {
+	*JSONFile
+}
+type cloneDetectionResponse struct {
+	Status AppStatus `json:"status"`
+	Token  string    `json:"token"`
+	Clones int       `json:"clones"`
+}
+
+func (d *cloneDetectionResponse) GetAppStatus() *AppStatus {
```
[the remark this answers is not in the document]
xgess: makes sense there's an abstraction for this 👍

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return 0, 0, err
+	}
+
+	err = SetDeviceCloneState(m, DeviceCloneState{Prior: res.Token, Stage: "", Clones: res.Clones})
+	after = res.Clones
+	return before, after, err
+}
+
+func configPaths(m MetaContext) (string, string, string) {
```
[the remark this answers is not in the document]
xgess: ahh right right. yeah, i'll change it.

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Zero out the TX so that we don't abort it in the defer()
+	tx = nil
+	return nil
+}
+
+func NewDeviceCloneStateJSONFile(g *GlobalContext) *DeviceCloneStateJSONFile {
```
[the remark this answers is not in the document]
xgess: this is all basically copy-pasta from updater_config and config. and i'd like to cut a bunch of the complexity around dealing with those (and any future) json files significantly. aside from following the logic, there's also a little bug with a mutex in there. 
but yes. i agree. will change here first.

go/libkb/device_clone_state.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	if err == nil {
+		err = writer.SetIntAtPath(cPath, d.Clones)
+	}
+	if err == nil {
+		err = tx.Commit()
+	}
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
xgess: yyyeeeeaaahhhh. it's long and repetitive, but i guess it's the most obvious and go-y way to do it.

PR #12808, opened 2018-07-13
Add storybooks for components with badges

nathunsmitty:
As I've been working with badges, I've realized it's hard to know how a change will affect badges throughout the apps. A big part of this is due to the fact that most components that use badges aren't in our storybook. #12673 would add a Badge component storybook entry, but wouldn't give us the context we really need to see how badges look in the different places they're used.

This PR aims to add entries for every component that uses a badge (if there wasn't already one that existed). A search through the codebase led me to the following places:

#### Shared
- Inbox dividers: [Big teams](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/chat/inbox/row/big-teams-divider/index.js#L48) and ["show more/less"](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/chat/inbox/row/divider/index.js#L42)
- [_Files_](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/fs/row/still.js#L64)
- [_Team List_](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/teams/main/team-list.js#L72)
- [**Team Tab**](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/teams/team/tabs/index.js)

#### Desktop
- [**Desktop Tab Bar**](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/common-adapters/tab-bar.desktop.js)
- [**Menubar**](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/menubar/index.desktop.js)

#### Mobile
- [**Mobile Tab Bar**](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/common-adapters/tab-bar.native.js)
- Back buttons, both in a [**chat conversations**](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/chat/conversation/header-area/normal/index.native.js#L32) and on [**screens**](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/common-adapters/back-button.native.js#L26)
- [Settings screen](https://github.com/keybase/client/blob/05fbab3f17e7c0f816d882bc8c6951603ea86f43/shared/settings/nav/index.native.js#L50)

The _italicized_ components already had stories, the **bolded** components got stories in the PR, and the non-bolded, non-italicized components do not have stories implemented yet. 

The files row storybook already existed, but I added an extra row to demo the badge. The team list storybook also already existed and tested the badge, so I didn't touch that.

The settings screen storybook is implemented in #12673.

Currently, the inbox related components are not included in this PR. After talking with @thebearjew, we decided that it'd be best to wait until his revamp of the inbox storybooks (#12718) was merged in before I tackled those. I plan on removing the `[WIP]` once those changes happen, but the rest of the code should be ready for review.

I changed a couple of things in some components in order to fix some errors and/or weird behavior I was getting in the storybook. I know that it isn't really super preferred to do this, so I've tried to note where and why in a self-review. If you know a better way to do something, please lmk.


shared/menubar/index.desktop.js
```diff
@@ -43,9 +43,9 @@ class MenubarRender extends Component<Props, State> {
 
   constructor(props: Props) {
     super(props)
-    SafeElectron.getRemote()
-      .getCurrentWindow()
-      .on('show', this._onShow)
+    if (SafeElectron.getCurrentWindowFromRemote) {
```
nathunsmitty: I was getting errors about SafeElectron not having a `getRemote` or `getCurrentWindow` property when I ran this in storybook. I think the reason why is in [`__mocks__/electron.js`](https://github.com/keybase/client/blob/637af207887e94ee23e916bf76b485f047225e74/shared/__mocks__/electron.js#L8), where `remote` is initialized to an almost empty object.

I also noticed that there was a single function (probably written after this code was last touched?) called [`getCurrentWindowFromRemote`](https://github.com/keybase/client/blob/637af207887e94ee23e916bf76b485f047225e74/shared/util/safe-electron.desktop.js#L83) that basically wrapped around these two functions. The cleanest solution seemed like to just check for this function before using it, but I could see this potentially not throwing an exception somewhere when we would want it to?

shared/chat/conversation/header-area/normal/index.types.js.flow
```diff
@@ -4,7 +4,7 @@ export type Props = {|
   canOpenInfoPanel: boolean,
   channelName: ?string,
   muted: boolean,
-  onCancelPending: () => void,
+  onCancelPending: null | (() => void),
```
nathunsmitty: I'm not exactly sure what onCancelPending is being used for but in all of the conversations I had, it was equal to null. It displays an extra close icon on desktop only if it is defined.

shared/stories/prop-providers.js
```diff
@@ -21,11 +21,10 @@ const defaultFollowers = ['max', 'akalin']
...
   Usernames: (ownProps: _Usernames.ConnectedProps): _Usernames.Props => {
     const {usernames} = ownProps
-    const users = (usernames || []).map(username => ({
-      username,
-      following: following.includes(username),
-      you: username === you,
-    }))
+    const users = (usernames || [])
+      .map(username => ({username, following: following.includes(username), you: username === you}))
+      .filter(u => !ownProps.skipSelf || !u.you)
```
nathunsmitty: Added this filter line to work with the chat conversation header, since mergeProps in `common-adapters/usernames.js` had it.

nathunsmitty: ConnectedUsernames is the only component that has a skipSelf prop so I don't think so. It also didn't affect any of the existing storyshots.

shared/chat/inbox/index.stories.js
```diff
@@ -471,12 +519,34 @@ const provider = createPropProvider(
...
   }
 )
 
+class Wrapper extends React.Component<any, any> {
+  state = {
+    props: propsInboxExpanded,
+  }
+
+  componentDidMount() {
+    setTimeout(() => {
```
nathunsmitty: Had an issue with displaying the the `BigTeamsDivider` – the internal logic depends on state, which is set on didComponentUpdate. The official storybook solution is to [pass down the desired state via a prop](https://storybook.js.org/basics/faq/#can-i-modify-react-component-state-in-stories), which isn't super great so @thebearjew came up with the solution of wrapping the Inbox in a timeout to ensure that didComponentUpdate fires.

nathunsmitty:
Modified the inbox story to include tests for the Big teams divider and show more/less!

nathunsmitty:
probably ready for another review @keybase/react-hackers

shared/package.json
```diff
@@ -91,7 +91,7 @@
       "react-list": "<rootDir>/__mocks__/react-list.js",
       "desktop/app/resolve-root": "<rootDir>/__mocks__/resolve-root.js",
       "react-redux": "<rootDir>/__mocks__/react-redux.js",
-      "electron": "<rootDir>/__mocks__/electron.js"
```
nathunsmitty: The regex was picking up and mocking safeelectron, so this fixes that.

shared/chat/inbox/index.stories.js
chrisnojima: we shouldn't have any kind of timing logic in the stories else storyshots might get messed up. better to pass a prop or wrap this in a `if (!__STORYSHOTS__)` check

shared/__mocks__/react-redux.js
```diff
@@ -23,7 +23,7 @@ const selectorDelegatorFactory = (dispatch, options) => {
     } catch (err) {
       throw new Error(
         `In calling propSelector for ${options.wrappedComponentName}:
-          Your propProvider is probably missing a key for this connected component. See shared/storybook/README.md for more details.
+          Your propProvider is probably missing a key for this connected component. See shared/stories/README.md for more details.
```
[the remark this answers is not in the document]
buoyad: oh I probably just typo'd

shared/chat/inbox/index.stories.js
nathunsmitty: great point, hadn't thought of that!

shared/chat/conversation/header-area/normal/index.stories.js
```diff
@@ -0,0 +1,40 @@
...
+
+const load = () => {
+  storiesOf('Chat/Header', module)
+    .addDecorator(provider)
+    .add('Username Header', () => <UsernameHeader {...defaultProps} />)
+    .add('Username Header with info panel open', () => (
+      <UsernameHeader {...defaultProps} infoPanelOpen={true} />
+    ))
+    .add('Username Header muted', () => <UsernameHeader {...defaultProps} muted={true} />)
+    .add('Username Header with onCancelPending', () => (
```
buoyad: Can you also pass in the other props that depend on `isPending`? From the connector:
```
  canOpenInfoPanel: !stateProps._isPending,
  onCancelPending: stateProps._isPending ? dispatchProps._onCancel : null,
  onOpenFolder: stateProps._isPending ? null : dispatchProps._onOpenFolder,
```

shared/chat/conversation/header-area/normal/index.types.js.flow
buoyad: It's used when you're in a pending conversation with a fixed set of users. You can get there by clicking a chat button on a profile or team member list.

shared/common-adapters/back-button.stories.js
```diff
@@ -0,0 +1,24 @@
...
+  badgeNumber: 0,
+  onClick: action('onClick'),
+  onPress: action('onPress'),
+}
+
+const load = () => {
+  storiesOf('Common', module).add('Back Button', () => (
+    <Box2 direction="vertical">
+      <BackButton {...defaultProps} />
+      <BackButton {...defaultProps} badgeNumber={5} />
```
buoyad: Should I see a badge?
image

shared/teams/team/tabs/index.stories.js
```diff
@@ -0,0 +1,54 @@
...
+  storiesOf('Teams/Tabs', module)
+    .add('Standard', () => <TeamTabs {...commonProps} />)
+    .add('Loading', () => <TeamTabs {...commonProps} loading={true} />)
+    .add('Admin', () => <TeamTabs {...commonProps} admin={true} />)
+    .add('User can Manage Subteams', () => (
+      <TeamTabs {...commonProps} yourOperations={{...yourOperations, manageSubteams: true}} />
+    ))
+    .add('Has Subteams', () => <TeamTabs {...commonProps} numSubteams={3} />)
+    .add('Has reset users', () => <TeamTabs {...commonProps} resetUserCount={7} />)
+    .add('Has team requests', () => <TeamTabs {...commonProps} admin={true} numRequests={4} />)
```
buoyad: This and `Has reset users` look a little funky:
image
image

shared/common-adapters/back-button.stories.js
nathunsmitty: The back button only shows badges on mobile so nope

nathunsmitty: unless that is mobile

shared/teams/team/tabs/index.stories.js
nathunsmitty: Oh, this is how it looks on master currently

nathunsmitty: Maybe #12798 is the cause? (cc @adamjspooner) I wrote this story a while ago, and I remember it looking fine then

shared/common-adapters/back-button.stories.js
buoyad: oh that's not mobile. cool :+1:

shared/chat/conversation/header-area/normal/index.stories.js
```diff
@@ -0,0 +1,44 @@
...
+  onCancelPending: null,
+  onOpenFolder: action('onOpenFolder'),
+  onShowProfile: action('onShowProfile'),
+  onToggleInfoPanel: action('onToggleInfoPanel'),
+  participants: ['joshblum', 'ayoubd'],
+  smallTeam: true,
+  teamName: 'keybase',
+}
+
+const isPendingProps = {
```
nathunsmitty: @buoyad

shared/teams/team/tabs/index.stories.js
nathunsmitty: Fixed by https://github.com/keybase/client/commit/f57fcc4d5ed8ff4dede1977c6800ced7c1230a22

shared/constants/platform.desktop.js
```diff
@@ -13,9 +13,10 @@ const isAndroidNewerThanM: boolean = false
 const isAndroidNewerThanN: boolean = false
 
 const isElectron = true
-const isDarwin = process.platform === 'darwin'
-const isWindows = process.platform === 'win32'
-const isLinux = process.platform === 'linux'
+// For storyshots, we only want to test macOS
+const isDarwin = process.platform === 'darwin' || __STORYSHOT__
```
nathunsmitty: @chrisnojima is this the best way to set the constants for storyshots?

PR #12809, opened 2018-07-13
Team batch add members

mlsteele:
Add a new RPC `teamAddMembers` and call that from the gui instead of looping over `teamAddMember`. The motivation behind this change is to save on team sigchain links. Since each call to `teamAddMember` is one chain link.

This implementation of `teamAddMembers` takes the happy-path addees and puts them in a transaction. The more complicated additions get looped over same as before. But we should be able to fix this on the core side without further changes to the RPC interface (@zapu).

There is one user-facing downside to this. If some additions succeed and others fail, the gui no longer removes the successful addees from search. It would be possible to spruce up the return value to fix this. However I think it is more pragmatic to leave it until this method gets fixed up to be a full transaction, at which point it will not be possible for it to partially succeed.

There are a distinct lack of tests. I will write them but want to sanity check this approach with @zapu first.


shared/actions/teams.js
```diff
@@ -91,50 +91,28 @@ const _addPeopleToTeam = function*(action: TeamsGen.AddPeopleToTeamPayload) {
   const state: TypedState = yield Saga.select()
   const ids = SearchConstants.getUserInputItemIds(state, {searchKey: 'addToTeamSearch'})
   logger.info(`Adding ${ids.length} people to ${teamname}`)
-  const collectedErrors = []
-  const errorUsers = []
   for (const id of ids) {
```
buoyad:
```
logger.info(`Adding ${ids.join(',')}`)
```

shared/actions/teams.js
```diff
@@ -91,50 +91,28 @@ const _addPeopleToTeam = function*(action: TeamsGen.AddPeopleToTeamPayload) {
...
-    }
+    logger.info(`Adding ${id}`)
   }
-  if (collectedErrors.length === 0) {
-    logger.info(`Successfully added ${ids.length} users to ${teamname}`)
+  try {
+    yield Saga.call(RPCTypes.teamsTeamAddMembersRpcPromise, {
+      name: teamname,
+      assertions: ids,
+      role: role ? RPCTypes.teamsTeamRole[role] : RPCTypes.teamsTeamRole.none,
```
buoyad: `role,`

buoyad: `RPCTypes.teamsTeamRole[role] || RPCTypes.teamsTeamRole.none`

mlsteele:
@zapu this is ready for another look!

@buoyad fyi this is a transaction now which means that when an error occurs not removing anyone from the search is the right thing. Here's an example
image

mlsteele:
Also for the record the stuff in the original comment is now mostly untrue.

go/teams/transactions.go
```diff
@@ -334,7 +375,39 @@ func (tx *AddMemberTx) AddMemberByUsername(ctx context.Context, username string,
...
 	}
 
-	return tx.addMemberByUPKV2(ctx, upak.Current, role)
+	_, err = tx.addMemberByUPKV2(ctx, upak.Current, role)
+	return err
+}
+
+// AddMemberByAssertion adds an assertion to the team.
+// Does not attempt to resolve the assertion.
+// Returns (uv, username) which can both be zero-valued if the assertion is not a keybase user.
```
zapu: looks like it returns `username, uv, invite`

go/teams/service_helper.go
```diff
@@ -260,6 +263,84 @@ func AddMember(ctx context.Context, g *libkb.GlobalContext, teamname, username s
...
+		tx := CreateAddMemberTx(team)
+		type sweepEntry struct {
+			Assertion string
+			UV        keybase1.UserVersion
+		}
+		var sweep []sweepEntry
+		for i, assertion := range assertions {
+			username, uv, invite, err := tx.AddMemberByAssertion(ctx, assertion, role)
+			if err != nil {
+				if strings.Contains(err.Error(), "doesn't have a Keybase account yet") {
```
zapu: since this error also comes from the service (and not api or anything), wouldn't you rather have this as separate error type?

go/systests/teams_test.go
```diff
@@ -1337,3 +1337,29 @@ func TestTeamCanUserPerform(t *testing.T) {
...
+	assertions := []string{
+		bob.username,
+		john.username,
+		rob.username,
+		bob.username + "@rooter",
+		rob.username + "@rooter",
+	}
+	role := keybase1.TeamRole_WRITER
+	res, err := teams.AddMembers(context.Background(), alice.tc.G, teamName, assertions, role)
+	_ = res
```
zapu: it would be nice to do the proper thing here and inspect team afterwards. Originally I was running that test and inspecting team chain manually, but now that this function is done and working, we can finish the test.

go/teams/transactions.go
```diff
@@ -334,7 +375,39 @@ func (tx *AddMemberTx) AddMemberByUsername(ctx context.Context, username string,
...
+			WithName(assertion).
+			WithPublicKeyOptional().
+			WithForcePoll(true))
+		if err != nil {
+			return "", uv, false, err
+		}
+		invite, err := tx.addMemberByUPKV2(ctx, upak.Current, role)
+		return libkb.NewNormalizedUsername(upak.Current.Username), upak.Current.ToUserVersion(), invite, err
+	}
+	typ, name, err := team.parseSocial(assertion)
```
zapu: What do you think about making parseSocial not method of `Team` object? a bit nitpicky and probably outside of the scope of this PR. reasons are that it's not even truly a "team" operation, and also its behavior does not depend on given team. We can also add a 0.5p ticket

zapu:
ci fails with:
```
[go] Running shell script
+ make -s lint
teams/transactions.go:297:9: if block ends with a return statement, so drop this else and outdent its block
Makefile:11: recipe for target 'lint' failed
make: *** [lint] Error 1[go] Running s
```

go/service/teams.go
```diff
@@ -236,6 +236,46 @@ func (h *TeamsHandler) TeamAddMember(ctx context.Context, arg keybase1.TeamAddMe
...
+		func() error { return err })()
+	if len(arg.Assertions) == 0 {
+		return fmt.Errorf("attempted to add 0 users to a team")
+	}
+	if err := h.assertLoggedIn(ctx); err != nil {
+		return err
+	}
+	res, err := teams.AddMembers(ctx, h.G().ExternalG(), arg.Name, arg.Assertions, arg.Role)
+	// Check the error after spawning a goroutine to send chat notifications.
+	savedErr := err
```
zapu: I'm not sure what's going on here, would you mind clearing it up for me?

go/service/teams.go
```diff
@@ -236,6 +236,46 @@ func (h *TeamsHandler) TeamAddMember(ctx context.Context, arg keybase1.TeamAddMe
...
+		go func() {
+			h.G().Log.CDebugf(ctx, "sending team welcome messages")
+			ctx := libkb.WithLogTag(context.Background(), "BG")
+			for _, res := range res {
+				if !res.Invite && !res.Username.IsNil() {
+					err := teams.SendTeamChatWelcomeMessage(ctx, h.G().ExternalG(), arg.Name, res.Username.String())
+					if err != nil {
+						h.G().Log.CDebugf(ctx, "send team welcome message () err: %v", res.Username, err)
+					} else {
+						h.G().Log.CDebugf(ctx, "send team welcome message () success", res.Username, err)
```
zapu: arguments and format string mismatch in both Debugf calls here

go/service/teams.go
mlsteele: Thanks for pointing this out. This is from when it could partially succeed. Doesn't make sense anymore.

PR #12811, opened 2018-07-13
has pairwise macs

joshblum:
adds `has_pairwise_macs` field to json api and also exposes it to the UI. 

@malgorithms @cecileboucheron should we add something to the message info popup to indicate when this is used?

### DIFF
--- a/go/chat/boxer.go
+++ b/go/chat/boxer.go
@@ -905,6 +905,7 @@ func (b *Boxer) unversionHeaderMBV2(ctx context.Context, serverHeader *chat1.Mes
 			KbfsCryptKeysUsed: hp.KbfsCryptKeysUsed,
 			EphemeralMetadata: hp.EphemeralMetadata,
 			Rtime:             rtime,
+			HasPairwiseMacs:   hp.HasPairwiseMacs,
 		}, hp.BodyHash, nil
 	// NOTE: When adding new versions here, you must also update
 	// chat1/extras.go so MessageUnboxedError.ParseableVersion understands the
@@ -1592,6 +1593,7 @@ func (b *Boxer) boxV2orV3orV4(ctx context.Context, messagePlaintext chat1.Messag
 	}
 
 	// create the v1 header, adding hash
+	usePairwiseMacs := len(pairwiseMACRecipients) > 0
 	headerVersioned := chat1.NewHeaderPlaintextWithV1(chat1.HeaderPlaintextV1{
 		Conv:              messagePlaintext.ClientHeader.Conv,
 		TlfName:           messagePlaintext.ClientHeader.TlfName,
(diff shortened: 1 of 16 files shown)


malgorithms:
> @malgorithms @cecileboucheron should we add something to the message info popup to indicate when this is used?

Nah, I don't think it's necessary.

protocol/avdl/chat1/local.avdl
```diff
@@ -292,6 +292,11 @@ protocol local {
     // This is the only difference between V2 and V3.
     @mpackkey("em") @jsonkey("em")
     union { null, MsgEphemeralMetadata } ephemeralMetadata;
+
+    // Whether or not the message was signed using pairwise MACs. Only valid
+    // for V4 where pairwise MACs are used for exploding messages.
+    @mpackkey("pm") @jsonkey("pm")
+    union { null, boolean } hasPairwiseMacs;
```
mmaxim: Why does this need to be here? We can't just get this information from somewhere else?

protocol/avdl/chat1/common.avdl
```diff
@@ -420,6 +420,9 @@
     // ephemeral lifetime calculations
     @mpackkey("rt") @jsonkey("rt")
     gregor1.Time rtime;
+
+    @mpackkey("pm") @jsonkey("pm")
```
mmaxim: Why not boot this guy too? Juts make `HasPairwiseMacs` do what we do to set the bool in boxer.

joshblum: i think it's required here since we need to copy from the `MessageClientHeader` to `MessageClientHeaderVerified` in some form. I'm not sure what bool you're referring to but calling `ShouldPairwiseMAC` https://github.com/keybase/client/blob/e8e11b3639573f66cc45b1730fb5f3f17a73fc69/go/chat/boxer.go#L1336 would be expensive

go/chat/boxer.go
```diff
@@ -846,6 +846,8 @@ func (b *Boxer) unboxV2orV3orV4(ctx context.Context, boxed chat1.MessageBoxed,
 	atMentions, atMentionUsernames, chanMention, channelNameMentions :=
 		b.getAtMentionInfo(ctx, clientHeader.Conv.Tlfid, membersType, body)
 
+	clientHeader.HasPairwiseMacs = len(boxed.ClientHeader.PairwiseMacs) > 0
```
mmaxim: I'm talking about this, just return this len thing from the function you just created in extras. Can that not happen, what am I missing?

joshblum: we can't since `boxed.ClientHeader` is `MessageClientHeader` which has a `PairwiseMacs` field. We don't copy this over to `MessageClientHeaderVerified` on the unboxed message (besides this bool). I thought it would be better to just have this floating around instead all of the pairwise macs since we don't use them after we verify

PR #12812, opened 2018-07-13
add tooltip for Downloads and fix center

songgao:
* [x] implement a `WithTooltip` component that shows a tooltip for desktop
* [x] use it for the Downloads folder buttons
* [x] fix "center" positions in `relative-popup-hoc.desktop.js`

[w=1](https://github.com/keybase/client/pull/12812/files?w=1) is your friend!


shared/common-adapters/relative-popup-hoc.desktop.js
```diff
@@ -203,8 +207,10 @@ function ModalPositionRelative<PP>(
       const style = collapseStyles([
         computePopupStyle(this.props.position, targetRect, popupNode.getBoundingClientRect()),
         this.props.style,
+        need2ndCall ? {opacity: 0} : {},
       ])
       this.setState({style})
+      need2ndCall && setTimeout(() => this._computeStyle(this.props.targetRect, true))
```
songgao: This is kind of hacky, but is the only way I could make it work.

chrisnojima: lets discuss, you can likely use componentDidUpdate or getSnapshotBeforeUpdate

shared/common-adapters/clickable-box.desktop.js
```diff
@@ -62,20 +76,31 @@ class ClickableBox extends React.Component<Props & {children: any}, {mouseDown:
...
-        onMouseEnter={needMouseEnterLeaveHandlers(this.props) ? this._onMouseEnter : undefined}
-        onMouseLeave={needMouseEnterLeaveHandlers(this.props) ? this._onMouseLeave : undefined}
-        onMouseUp={this._onMouseUp}
-        onClick={onClick}
-        style={collapseStyles([_containerStyle, onClick ? desktopStyles.clickable : null, style])}
-      >
-        {underlay}
-        {children}
-      </div>
+      <React.Fragment>
```
chrisnojima: i think we should just make a new thing instead of changing this component. else you'll have to test every part of the app again

songgao: Sounds good; I'll try to make a `<WithTooltip />`

shared/common-adapters/relative-popup-hoc.desktop.js
songgao: This is already called in `componentDidUpdate`. I think it has to do with browser figuring out the actual width. I'll see if I can make `getSnapshotBeforeUpdate` work.

songgao: It works!! Thanks

songgao:
@chrisnojima ready for another look; thanks!

shared/common-adapters/tooltip.desktop.js
```diff
@@ -0,0 +1,61 @@
+// @flow
```
buoyad: We recently added an adapter `toast.desktop.js` that looks pretty similar to this. Could that be used here?

songgao: Ahhh I had no idea! I searched for "tooltip" and found nothing alike. Will switch to this. Thanks!

songgao:
🏓Ready for another look!

zanderz:
I tried it, LGTM

songgao:
Thanks folks!

PR #12824, opened 2018-07-16
Add global tab shortcuts.

adamjspooner:
Take two. @keybase/react-hackers, this is ready for review.


shared/app/nav.desktop.js
```diff
@@ -5,26 +5,61 @@ import * as RouteTree from '../route-tree/render-route'
...
+  '2': chatTab,
+  '3': fsTab,
+  '4': teamsTab,
+  '5': devicesTab,
+  '6': gitTab,
+  '7': settingsTab,
+  ...(flags.walletsEnabled ? {'8': walletsTab} : {}),
+}
+
+const hotkeys = Object.keys(hotkeyTabMap).map(key => `${isDarwin ? 'command' : 'ctrl'}+${key}`)
```
nathunsmitty: Not super familiar with Electron and how we use it yet, but I think there's a [`CmdOrCtrl`](https://electronjs.org/docs/api/accelerator) value that basically does what you're trying to do here.

adamjspooner: Sadly, that's not compatible with Mousetrap (the library we're using behind the scenes).

nathunsmitty: Ah well 😞 looks good to me then! Excited to have this feature!

shared/app/nav.desktop.js
```diff
@@ -53,7 +88,14 @@ const mapStateToProps = (state: TypedState) => ({
   _username: state.config.username,
 })
 
-const mapDispatchToProps = (dispatch: Dispatch) => ({})
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  _onHotkey: (cmd: string) => {
+    const tab = hotkeyTabMap[cmd.replace(/(command|ctrl)\+/, '')]
+    if (tab) {
+      dispatch(navigateTo([tab]))
```
buoyad: using `switchTo` would allow the tab stack to persist rather than being blown away each time.

shared/app/nav.desktop.js
```diff
@@ -5,26 +5,61 @@ import * as RouteTree from '../route-tree/render-route'
 import GlobalError from './global-errors/container'
 import Offline from '../offline/container'
 import TabBar from './tab-bar/container'
+import {isDarwin} from '../constants/platform'
 import {Box, ErrorBoundary} from '../common-adapters'
-import {chatTab, loginTab, type Tab} from '../constants/tabs'
+import {
```
buoyad: can you do `import * as Tabs` here?

shared/app/nav.desktop.js
adamjspooner: TIL 🙌

PR #12825, opened 2018-07-16
Add TeamMemberStatus to TeamMember

joshblum:
protocol changes for member status. Removes the `UserEldestSeqno` field in favor of the `status` field since `UserEldestSeqno != EldestSeqno` could mean either a `RESET` or `DELETED` user, we should just be using the `status` field


mmaxim:
I think the idea was to not have to do this, and to just treat deleted users as if they were active.

go/protocol/keybase1/extras.go
```diff
@@ -1872,10 +1872,6 @@ func (t TeamMembers) AllUserVersions() []UserVersion {
 	return all
 }
 
-func (t TeamMember) IsReset() bool {
```
joshblum: we now use `TeamMember.Status` throughout

joshblum:
approved in person

PR #12826, opened 2018-07-16
dump power save block

mmaxim:
Seems like this drops an idle lock on macOS, which doesn't seem great. I guess we will need a different solution to the problem I see where Electron goes idle. 

cc @chrisnojima


tamsky:
seems like the kinda bug that you'd say:  "we should cut a release just to fix this."
wasting power does nobody any good.
just sayin'

tamsky:
thanks for the latest release!

PR #12829, opened 2018-07-16
OrientedImage: Use Canvas API for Better Results

thebearjew:
### Problem

Correctly orienting images was done in #12797 using css `transform` styles. However images with orientations that involve a `90º` or `-90º` rotation can cause the final image to be rendered out of bounds of the parent.
This is because transformed DOM elements do not participate in flow and will not cause the parent to expand/contrast after transformation. This occurs when shrinking the window size to the minimum height. The rotated image will not scale down because css `width: 100%` and `height: auto` are applied to the un-transformed DOM element.

oriented-images-exceeding-bounds

### Solution

Use the canvas api to perform the following operations
1. Render a `<canvas />` element and get a backing 2D context from `getContext('2d')`
2. Draw the image into the context
3. Apply the appropriate transformation to the image using context `rotate()` `scale()` and `transform()`. Note: This approach was adapted from [croppie.js](https://github.com/Foliotek/Croppie/blob/master/croppie.js#L296-L352).
4. Encode the entire canvas as a dataURL using `canvas.toDataUrl()`
5. Use the canvas dataURL as the final source to the rendered `<Image />`

Final result works on pre-upload preview images (served from the user's filesystem) as well as the fullscreen attachment view for images.

screen shot 2018-07-16 at 2 46 37 pm
screen shot 2018-07-16 at 2 47 00 pm

### Notes

- The final result from this process is a regular `` component whose **content** has been rotated/flipped according to the EXIF orientation flag.
- There is some latency in the process from src → img → canvas → dataURL → img. However it is usually less that 700ms for jpeg images under 5Mb and approximately 1.5 seconds for images larger than 10Mb
- When the image is large, the largest source of latency is loading the image data from the service.
- Zooming works as expected.

### DIFF
--- a/shared/common-adapters/oriented-image.desktop.js
+++ b/shared/common-adapters/oriented-image.desktop.js
@@ -4,19 +4,25 @@ import fs from 'fs'
 import EXIF from 'exif-js'
 import {noop, isNumber} from 'lodash-es'
 import logger from '../logger'
-import {Image} from '../common-adapters'
+import {Image as ImageComponent} from '../common-adapters'
 import type {Props} from './oriented-image.types'
-import {collapseStyles} from '../styles'
 
 type State = {
-  styleTransform: string,
+  srcTransformed: string,
 }
+type TransformFn = (
+  canvas: HTMLCanvasElement,
+  ctx: CanvasRenderingContext2D,
+  width: number,
+  height: number
+) => void
 
 const uploadedSrc = /https?:\/\/127.0.0.1:.*$/i
 const NO_TRANSFORM = 'notransform'
+const _cacheNoTransforms: {[src: string]: string} = {}
 
-const _cacheStyleTransforms: {[src: string]: string} = {}
-// Orientations:
+// Define transformatin functions to operate on canvas elements
+//
 // 1: rotate 0 deg left
 // 2: flip horizontally
 // 3: rotate 180 deg left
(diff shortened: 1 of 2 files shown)


jzila:
Does this work well with retina-like scaling?

shared/common-adapters/oriented-image.js.flow
```diff
@@ -2,4 +2,6 @@
 import * as React from 'react'
 import type {Props} from './oriented-image.types'
 
-export default class OrientedImage extends React.Component<Props> {}
+export default class OrientedImage extends React.Component<Props> {
+  _context: ?CanvasRenderingContext2D
```
buoyad: is this needed?

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -102,79 +164,100 @@ class OrientedImage extends React.Component<Props, State> {
...
 
   componentWillUnmount() {
     this._hasComponentMounted = false
+    this._canvasRef = null
+    this._context = null
   }
 
   componentDidUpdate(prevProps: Props, prevState: State) {
     // New src requires changing EXIF transform
     if (prevProps && prevProps.src !== this.props.src) {
```
buoyad: can `prevProps` be falsy?

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -102,79 +164,100 @@ class OrientedImage extends React.Component<Props, State> {
...
+    }
+
+    this._canvasImageTransform(orientation)
   }
 
-  // Don't perform transforms if the image cannot be loaded or there is no EXIF data
-  _handleImgeLoadFailure = src => {
-    _cacheStyleTransforms[src] = NO_TRANSFORM
+  // Mark this image path as no transform and set the ImageComponent src to the original source
+  _handleOrientationFailure = () => {
```
buoyad: maybe log the error here if one comes through?

shared/common-adapters/oriented-image.desktop.js
```diff
@@ -57,40 +81,78 @@ const makeStyleTransform = (orientation: ?number): string => {
...
+    canvas.height = height
+
+    ctx.save()
+
+    const transformFn = transformMap[orientation.toString()]
+    if (!transformFn) {
+      logger.warn(`Invalid orientation value for desktop image attachment: orientation=${orientation}`)
+      return ''
     }
+    // Appy transformation to canvas
```
buoyad: `Apply` typo

shared/common-adapters/oriented-image.desktop.js
thebearjew: 🤦‍♂️ It can't, guaranteed to be called after first render. Thanks :)

shared/common-adapters/oriented-image.js.flow
thebearjew: I was having Flow issues when calling methods on the context object. In theory flow should have inferred the type because it knows the return value of `canvas.getContext()` however it didn't seem to catch on.

shared/common-adapters/oriented-image.desktop.js
thebearjew: There are a few things that can cause a "failure"
1. `fs.readSync` was not able to read the file
2. `EXIF.getData` returned false because the image is not a jpeg, or there is no EXIF data set on the image. Both of these occur before ever accessing EXIF, so logging here would log for every image type without EXIF
3. A more critical error is getting all the way to reading the EXIF Orientation flags and it is not a value from 1 to 8.

PR #12830, opened 2018-07-16
Notify frontend of reaction deletions

joshblum:
Adds `chat1.ReactionDeleteNotif` so we can notify the frontend of deleted reactions

cc @buoyad

### DIFF
--- a/go/chat/convsource.go
+++ b/go/chat/convsource.go
@@ -1045,14 +1045,14 @@ func (s *HybridConversationSource) GetMessagesWithRemotes(ctx context.Context,
 	return res, nil
 }
 
-func (s *HybridConversationSource) expungeNotify(ctx context.Context, uid gregor1.UID,
+func (s *HybridConversationSource) notifyExpunge(ctx context.Context, uid gregor1.UID,
 	convID chat1.ConversationID, mergeRes storage.MergeResult) {
 	if mergeRes.Expunged != nil {
 		var inboxItem *chat1.InboxUIItem
 		topicType := chat1.TopicType_NONE
-		conv, err := GetVerifiedConv(ctx, s.G(), uid, convID, true)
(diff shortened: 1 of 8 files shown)


protocol/avdl/chat1/notify.avdl
```diff
@@ -82,6 +83,17 @@ protocol NotifyChat {
     union { null, InboxUIItem } conv;
   }
 
+  record ReactionDelete {
+    string reactionKey;
```
joshblum: @buoyad any other info/changes that would be good to put here?

buoyad: nope, we can look it up easily given this :+1:

go/chat/convsource.go
```diff
@@ -1066,13 +1066,31 @@ func (s *HybridConversationSource) expungeNotify(ctx context.Context, uid gregor
 	}
 }
 
+// notifyReactionDeletes notifies the GUI after reactions are deleted
+func (s *HybridConversationSource) notifyReactionDeletes(ctx context.Context, uid gregor1.UID, convID chat1.ConversationID, reactionDeletes []chat1.ReactionDelete) {
+	s.Debug(ctx, "notifyReactionDeletes: %d deletions", len(reactionDeletes))
+	if len(reactionDeletes) > 0 {
+		topicType := chat1.TopicType_NONE
```
mmaxim: We don't know the `topicType` here?

mmaxim:
Pending my comment about the topic type

go/chat/convsource.go
joshblum: we could hardcode to `CHAT` but i thought this was just a future safe thing since we do the same for ephemeral notify https://github.com/keybase/client/pull/12830/files#diff-cf8add58c26f1cbee6e6e1e5f4a20d6eR1092

mmaxim: Oh wait this is correct, sorry!

PR #12831, opened 2018-07-16
client: add `keybase fs quota` command, plus new simplefs quota method

strib:
With flags to show git usage instead of regular usage, and also to add archived usage if desired.

Looks like this:

```sh
$ keybase fs quota
Usage:       44.86 GB
Quota:       250.00 GB

$ keybase fs quota -bytes
Usage:       48171491323 bytes
Quota:       268435456000 bytes

$ keybase fs quota -git -archived
Usage:       1.95 GB
Archived:    60.95 KB
Quota:       100.00 GB
```

(Circle will fail until the corresponding KBFS PR is in.)

Issue: KBFS-3188


go/client/cmd_simplefs_quota.go
```diff
@@ -0,0 +1,114 @@
...
+			cl.SetNoStandalone()
+		},
+		Flags: []cli.Flag{
+			cli.BoolFlag{
+				Name:  "git",
+				Usage: "show git usage",
+			},
+			cli.BoolFlag{
+				Name:  "bytes",
+				Usage: "show all data in bytes",
```
jzila: Where is this used? We have `UsageBytes` and `LimitBytes`, but neither of them are guarded by a switch.

go/client/cmd_simplefs_quota.go
```diff
@@ -0,0 +1,114 @@
...
+	return cli.Command{
+		Name:  "quota",
+		Usage: "show quota usage for logged-in user",
+		Action: func(c *cli.Context) {
+			cl.ChooseCommand(
+				&CmdSimpleFSQuota{Contextified: libkb.NewContextified(g)},
+				"quota", c)
+			cl.SetNoStandalone()
+		},
+		Flags: []cli.Flag{
```
jzila: Might be nice to get a json output switch too.

go/client/cmd_simplefs_quota.go
strib: It's used here to avoid humanizing the output: https://github.com/keybase/client/pull/12831/files#diff-ee8ff213e578cc065dedd53e8b3b08e9R72

Perhaps this help text is bad.

PR #12833, opened 2018-07-17
sound on notification desktop setting

zanderz:
Saved to the local config file

### DIFF
--- a/shared/settings/notifications/index.js.flow
+++ b/shared/settings/notifications/index.js.flow
@@ -25,6 +25,8 @@ export type Props = {
   onToggleUnsubscribeAll: (group: string) => void,
   mobileHasPermissions: boolean,
   waitingForResponse: boolean,
+  sound?: boolean,
+  onToggleSound?: (sound: boolean) => void,
 }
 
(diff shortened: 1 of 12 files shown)


shared/settings/notifications/index.js.flow
```diff
@@ -25,6 +25,8 @@ export type Props = {
   onToggleUnsubscribeAll: (group: string) => void,
   mobileHasPermissions: boolean,
   waitingForResponse: boolean,
+  sound?: boolean,
+  onToggleSound?: (sound: boolean) => void,
```
songgao: It looks like these can be `sound: boolean` and `onToggleSound: (sound: boolean) => void`. They are never `undefined` right?

zanderz: They can be undefined on mobile. Flow complained about this otherwise.

songgao: Ah I see it's under the `!isMobile` gate. LGTM then!

zanderz:
I'll take that as a thumb, thanks, Song!

PR #12834, opened 2018-07-17
Tweaked inbox icons

cecileboucheron:
@keybase/react-hackers 
I'm having a flow error for which I need someone's help to decode.


shared/chat/inbox/row/small-team/bottom-line.js
```diff
@@ -100,34 +112,52 @@ class BottomLine extends PureComponent<Props> {
...
-              {this.props.snippet}
-            </Markdown>
-          )}
+        <Box2 direction="horizontal" gap="xtiny" style={styles.bottomLineBox}>
           {!!snippetDecoration && (
             <Box2 direction="vertical" centerChildren={true}>
               {snippetDecoration}
             </Box2>
           )}
+          {!exploded && (
```
adamjspooner: You could fix the Flow error by adding `!exploded && this.props.snippet &&(` or by changing the props at the top from:

```
snippet: ?string,
```

to:

```
snippet?: string,
```

chrisnojima: Do not ever do a this.props.string test. Always do !! Or you'll crash android

PR #12839, opened 2018-07-17
Stellar: CancelRequest RPC and mocks

zapu:
Clientside of stellar requests. Needs to be merged first and then protocol should be vendored in to server PR.


zapu:
@patrickxb can you look at this one too? it has to go first

PR #12842, opened 2018-07-17
Show reactions in the GUI and respond to service notifications

buoyad:
All that's left is the desktop / mobile picker and people can start sending these. I've been dispatching a `toggleMessageReaction` in my console to test this. r? @keybase/react-hackers 

- Adds actions to toggle reactions with the service, and to add / remove them from the store
- Handles `commonMessageType.reaction` and `notifyChatChatActivityType.reactionDelete`
- Adds `ReactionsRow` to display reactions below each message
- Hooks up connector for `ReactionsRow` and `ReactButton` and changes all of the view to use `Ordinal` instead of `MessageID`
- Adds a `reactions` field to all message types except `deleted` and `placeholder`


buoyad:
(also the commits from the first PR got lumped into this as well, so only the ones since around `1a46320 ` are new. the diff is correct though.)

shared/chat/conversation/messages/react-button/container.js
```diff
@@ -1,22 +1,45 @@
...
-  messageID: MessageTypes.MessageID,
+  ordinal: Types.Ordinal,
 }
 
 const mapStateToProps = (state: TypedState, ownProps: OwnProps) => {
-  // TODO
-  return {}
+  const me = state.config.username || ''
+  const message = Constants.getMessage(state, ownProps.conversationIDKey, ownProps.ordinal)
+  if (!message || message.type === 'placeholder' || message.type === 'deleted') {
```
chrisnojima: you should make a `const noEmoji = { active: false...` in the module and just return that in these cases

buoyad: good idea, done :+1:

PR #12843, opened 2018-07-17
Stellar: integrate transaction details page

cjb:
@keybase/react-hackers 

The transaction details page shows fields (`txID, publicMemo`) that we don't have from the normal payments RPC, so we now call a new RPC to get the extra data and merge it in.

We do have an ID (`.id` vs `.txID`), but it's an opaque type that we're not supposed to be looking inside.  `txID`'s guaranteed to be a string that we can use to find the transaction on an explorer.


shared/common-adapters/back-button.desktop.js
```diff
@@ -33,6 +33,7 @@ export const styles = {
     ...globalStyles.flexBoxRow,
     ...desktopStyles.clickable,
     alignItems: 'center',
+    zIndex: 1,
```
chrisnojima: is this necessary

shared/wallets/transaction-details/container.js
```diff
@@ -0,0 +1,50 @@
...
+import * as WalletsGen from '../../actions/wallets-gen'
+import TransactionDetails from '.'
+
+const mapStateToProps = (state: TypedState, ownProps) => ({
+  _transaction: Constants.getPayment(
+    state,
+    ownProps.routeProps.get('accountID'),
+    ownProps.routeProps.get('paymentID')
+  ),
+  _you: state.config.username,
```
chrisnojima: do the || '' here instead

shared/wallets/transaction-details/container.js
```diff
@@ -0,0 +1,50 @@
...
+  const yourRole = Constants.paymentToYourRole(tx, stateProps._you || '')
+  const counterpartyType = Constants.paymentToCounterpartyType(tx)
+  return {
+    amountUser: tx.worth,
+    amountXLM: tx.amountDescription,
+    counterparty: yourRole === 'sender' ? tx.target : tx.source,
+    counterpartyType,
+    delta: tx.delta,
+    memo: tx.note,
+    onBack: () => dispatchProps.navigateUp(),
```
chrisnojima: dont need the extra () , just use dispatchProps.navigateUp

shared/wallets/transaction-details/index.js
```diff
@@ -75,82 +76,91 @@ const Counterparty = (props: CounterpartyProps) => {
...
+        <Box2 direction="vertical" gap="xxtiny" fullWidth={true}>
+          <Text type="BodySmallSemibold">Status:</Text>
+          <Box2 direction="horizontal" fullHeight={true} fullWidth={true} style={{alignItems: 'center'}}>
+            <Icon
+              color={this.props.timestamp ? globalColors.green2 : globalColors.black}
+              fontSize={16}
+              type={this.props.timestamp ? 'iconfont-success' : 'icon-transaction-pending-16'}
+            />
+            <Text
+              style={{
```
chrisnojima: should move styles into stylesheetCreate below

cjb:
@chrisnojima (reverted to the zIndex change now)

cjb:
Got a verbal thumb from Noj in standup, merging in.

PR #12844, opened 2018-07-17
fix updateAllSupersededBy

joshblum:
@buoyad ran into a bug where he had set the target msg id for a reaction to `0` this caused a bug in `updateAllSupersededBy` throwing ` Storage: MergeHelper -> ERROR: internal chat storage error: getBlock: invalid block id: 0`. This PR just skips these messages since they aren't valid but also cleans up the storage code to make future debugging easier

### DIFF
--- a/go/chat/storage/storage.go
+++ b/go/chat/storage/storage.go
@@ -311,6 +311,9 @@ func (s *Storage) MaybeNuke(ctx context.Context, force bool, err Error, convID c
 		if err := s.idtracker.clear(convID, uid); err != nil {
 			s.Debug(ctx, "failed to clear max message storage: %s", err)
 		}
+		if err := s.ephemeralTracker.clear(uid); err != nil {
+			s.Debug(ctx, "failed to clear ephemeral tracker storage: %s", err)
+		}
 	}
 	return err
 }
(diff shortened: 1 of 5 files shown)


go/chat/storage/storage.go
```diff
@@ -311,6 +311,9 @@ func (s *Storage) MaybeNuke(ctx context.Context, force bool, err Error, convID c
 		if err := s.idtracker.clear(convID, uid); err != nil {
 			s.Debug(ctx, "failed to clear max message storage: %s", err)
 		}
+		if err := s.ephemeralTracker.clear(uid); err != nil {
```
joshblum: noticed this was missing

joshblum:
I think there is a deeper bug that is causing the 0 since danny is still seeing this after confirming they are now never sending 0s. I think this change is good since it completely hoses the chat if this gets in there, we're trying to setup a consistent repro to figure out whats going wrong

joshblum:
i think that when @buoyad made the protocol change to drop the `mpackkey` it zeroed out the values on clients running with the old protocol definitions causing this error. So no further fixes, but i think this change is for the better overall

PR #12847, opened 2018-07-17
Add GetTeamID(string teamName) RPC

zapu:


protocol/avdl/keybase1/teams.avdl
```diff
@@ -877,4 +877,5 @@ protocol teams {
    */
   ProfileTeamLoadRes profileTeamLoad(LoadTeamArg arg);
 
+  TeamID getTeamIDByName(string teamName);
```
zapu: I'd be happy to change this to `getTeamIDByNameUntrusted`.

The original plan was to have `trusted` (and maybe `public`) arguments to this function, but there's no way to set default values and it would make calling it unnecessarily complicated from JS side.

zapu:
cc @buoyad

zapu:
one caveat - it will be a non-error for non-existing teams, because getting teamid from teamname is just hashing.

mlsteele:
What is this for? I don't like unverified rpcs

zapu:
Changed it to use resolver, and it's not just hashing for root teams now.

protocol/avdl/keybase1/teams.avdl
mlsteele: I thought frontend usually wrapped everything. Who did you talk to about too many arguments?

zapu: cc @buoyad

mlsteele: Could be named `getTeamID`

mlsteele: Comment that it has no caching and hits the server every time (if that's right).

zapu: From a quick look at resolver it seems to be caching both in memory and on disk.

zapu: I'll look again and watch logs to see when it makes requests though

PR #12854, opened 2018-07-18
switch avatars over to MetaContext

maxtaco:
We're going to be requiring `API#GetDecode` to take a `MetaContext` first argument.  In trying to make that change, I hit resistance here. So pull out the context -> MetaContext change into this PR.

### DIFF
--- a/go/avatars/urlcaching.go
+++ b/go/avatars/urlcaching.go
@@ -1,7 +1,6 @@
 package avatars
 
 import (
-	"context"
 	"fmt"
 	"time"
 
@@ -11,8 +10,6 @@ import (
 )
 
 type URLCachingSource struct {
-	libkb.Contextified
-
 	diskLRU        *lru.DiskLRU
 	staleThreshold time.Duration
 	simpleSource   *SimpleSource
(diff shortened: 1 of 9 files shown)


go/avatars/urlcaching.go
```diff
@@ -53,24 +53,24 @@ func (c *URLCachingSource) isStale(item lru.DiskLRUEntry) bool {
 }
 
 func (c *URLCachingSource) monitorAppState() {
-	c.debug(context.Background(), "monitorAppState: starting up")
+	m := libkb.NewMetaContextBackground(c.G())
+	c.debug(m, "monitorAppState: starting up")
 	state := keybase1.AppState_FOREGROUND
 	for {
 		state = <-c.G().AppState.NextUpdate(&state)
-		ctx := context.Background()
```
maxtaco: @mmaxim  is there a good reason to make a new context everytime through this loop?  I couldn't figure it out

mmaxim: No

mmaxim:
Why are these still `libkb.Contextified`? I thought a big reason for this stuff is to get rid of that.

maxtaco:
Happy to change that too, It's a slightly bigger PR (though not much bigger)

mmaxim:
Seems like we should, I feel like that is the prime benefit right?

maxtaco:
OK done. To my mind, the prime benefit is requiring that a `context.Context` (via `libkb.MetaContext`) is passed through to the `API` functions.  Right now, it's kind of there, sometimes.  But that ends up requiring MetaContexts up and down the callstack, which is how I got here.

I'd put de-Contextifying as a secondary benefit, but it doesn't really matter.  Anyways, done!

maxtaco:
Thanks @mmaxim! We'll be psyched about all this when we implement "two simultaneously logged in users"

PR #12855, opened 2018-07-18
remove UserEldestSeqno from TeamMember

joshblum:
Once #12825 is merged/deployed on the server we can remove this field


mmaxim:
Same comment here as on other PR

joshblum:
approved over chat

PR #12856, opened 2018-07-18
swallow deleted user error on chat identify

joshblum:
depends on server side changes in https://github.com/keybase/keybase/pull/2682

Is this the right place to catch the error? Should we do it earlier?

cc @maxtaco @zapu if you guys have opinions


zapu:
this looks good but I'd also look at the identify stuff to make sure it goes through the sigchain (if there's one) before ultimately bailing out with DeletedError. This would be best of both world IMO

joshblum:
confirmed manually that the signchain is loaded and then the user deleted error is returned

zapu:
Doesn't it exit super early here: https://github.com/keybase/client/blob/master/go/libkb/sig_chain.go#L256-L262 ? or does it take different code path to load sigs?

joshblum:
that status code only gets set if we haven't been in a chat with the user before so in this case we'll check the chain

joshblum:
yup comments sound good, added

PR #12858, opened 2018-07-18
Handle stellar.request_status gregors

zapu:
placeholder for actual support, always dismiss the message


zapu:
I forgot about this in previous PR

PR #12868, opened 2018-07-18
Stellar: choose asset page

buoyad:
Uses `SectionList` to implement the list of asset choices and adds it to the storybook. 

cc @cjb: as part of this, I removed the container & platform specific code for the send form `Header`. I don't know what things the `Header` would subscribe to from the store, it seems simple enough that all the props could come from the parent. The styling also seems simple enough to be in one file. Let me know if this is a bad idea and I can revert.

r? @keybase/react-hackers


buoyad:
Affected storyshot is the send form bc of the changes to `Header`

PR #12873, opened 2018-07-19
Default to chat filter row while loading

thebearjew:
Prevents "Start a new chat" from being rendered while conversations are loading

@keybase/react-hackers

### DIFF
--- a/shared/chat/inbox/row/chat-inbox-header/container.js
+++ b/shared/chat/inbox/row/chat-inbox-header/container.js
@@ -13,6 +13,7 @@ type OwnProps = {
 
 const mapStateToProps = (state: TypedState, ownProps: OwnProps) => ({
   filter: state.chat2.inboxFilter,
+  isLoading: !state.chat2.loadingMap.isEmpty(),
 })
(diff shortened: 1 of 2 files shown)


thebearjew:
@keybase/react-hackers 
1. Read `isLoading` correctly
2. Don't render `StartNewChat` when `neverLoaded` is true.

shared/chat/inbox/row/chat-inbox-header/container.js
```diff
@@ -13,13 +14,17 @@ type OwnProps = {
 
 const mapStateToProps = (state: TypedState, ownProps: OwnProps) => ({
   filter: state.chat2.inboxFilter,
+  isLoading: Constants.anyChatWaitingKeys(state),
```
chrisnojima: why is the isLoading prop needed?

thebearjew: I suppose it's not!

thebearjew:
@keybase/react-hackers Bump :)

PR #12880, opened 2018-07-19
Revert "get rid of transport keys on MessageReaction struct" 

buoyad:
r? @keybase/react-hackers cc @joshblum


chrisnojima:
might as well use this instead: https://github.com/keybase/client/pull/12881

buoyad:
@chrisnojima sounds good

PR #12881, opened 2018-07-19
support mpack protocol avdl

chrisnojima:
@keybase/react-hackers 
@buoyad @joshblum


mmaxim:
@joshblum this sounds like the best solution to our problem. Let's put the mpackkey back and the JS can use that field.

PR #12882, opened 2018-07-19
Electron depends on libasound2

mlsteele:
Electron depends on libasound2.
As referenced here: https://electronjs.org/docs/development/build-instructions-linux
Here's an example of another project adding it: https://goo.gl/AYChLr

I don't know how to test this.

### DIFF
--- a/packaging/linux/Dockerfile
+++ b/packaging/linux/Dockerfile
@@ -13,7 +13,7 @@ MAINTAINER Keybase <admin@keybase.io>
 #   - gnupg1 to avoid a password issue in key import
 RUN apt-get update
 RUN apt-get install -y fakeroot reprepro rpm createrepo git wget \
-  build-essential curl python python-pip libc6-dev-i386 gnupg1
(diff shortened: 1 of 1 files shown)


zapu:
oh so that's just docker?

mlsteele:
What do you mean? I assumed that docker file is part of our linux packaging system.

zapu:
I think it's just for people who want to run KB in docker.

dependencies in .deb are here: https://github.com/keybase/client/blob/master/packaging/linux/deb/package_binaries.sh#L35 (although this looks wrong to only be in one prerelease if block, but I don't know)

for rpm:
https://github.com/keybase/client/blob/master/packaging/linux/rpm/package_binaries.sh#L117

for arch:
https://github.com/keybase/client/blob/master/packaging/linux/arch/PKGBUILD.bin.in#L17


cc @oconnor663 for quick look? IIRC that alsa thing was happening before

mlsteele:
Oh I see. I'll remove the Dockerfile change then, since I doubt anyone is running electron out of docker.

mlsteele:
Actually I have no idea how to test this

PR #12883, opened 2018-07-19
[DO NOT MERGE] test AddMember implemented using AddMembers

mlsteele:
Piggy back on the existing AddMember tests to validate AddMembers


zapu:
same lint failure :)

PR #12886, opened 2018-07-19
WIP: New Provision Part 3: fixes and tests

chrisnojima:


shared/reducers/provision.js
```diff
@@ -53,12 +53,12 @@ export default function(state: Types.State = initialState, action: ProvisionGen.
         error: initialState.error,
       })
     case ProvisionGen.submitDeviceName:
-      if (state.existingDevices.indexOf(state.deviceName) !== -1) {
+      if (state.existingDevices.indexOf(action.payload.name) !== -1) {
```
chrisnojima: a bug!

shared/__mocks__/hidden-string.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+if (!__STORYBOOK__) {
```
chrisnojima: mock hidden string so we can use toEqual() effectively

shared/constants/types/provision.js
```diff
@@ -15,7 +15,10 @@ export type _State = {
   codePageOtherDeviceName: string,
   codePageOtherDeviceType: 'mobile' | 'desktop',
   codePageOtherDeviceId: string,
-  codePageTextCode: HiddenString,
```
chrisnojima: i split this into 2 distinct values. before it could get mixed between the incoming code and outgoing code

PR #12887, opened 2018-07-19
Don't call getTeams inside components, and always reload teamList when it changes

cjb:
@keybase/react-hackers 

From the ticket:

1. We'll make a bootstrap call to the teamlist (already does this)

2. We'll listen to the notifications and recall teamlist (always, no matter what tab you're on), the goal being to always have this be correct

3. We'll remove any existing calls to teamlist. We should already have this data


cjb:
@akalin-keybase Thanks! Removed that one too. At first read I thought it might be necessary for badging, but looks like it isn't.

mmaxim:
Wait I just saw this, I'm not sure this can go in yet. When we talk this was supposed to be dependent on the service not sending up as many notifications that would cause the JS to reload. 

cc @buoyad @chrisnojima I'm not sure this should be in here yet.

PR #12891, opened 2018-07-19
Chat: Render Repo Create Message

thebearjew:
### Problem

`MessageSystemGitPush` receives an integer flag from the service (`pushType`), however the UI did not read this flag and render accordingly. After #12590, we render timestamps above all message types for a better UX. However, when a repo creation event comes in, we render the timestamp wrapper but not an actual system message component.

screen shot 2018-07-19 at 5 12 43 pm

### Solution 

We pass the `pushType` along with the message creation for `MessageSystemGitPush` and read it in `chat/conversation/message/system-git-push/`.

screen shot 2018-07-19 at 5 13 04 pm

### DIFF
--- a/shared/chat/conversation/messages/system-git-push/index.js
+++ b/shared/chat/conversation/messages/system-git-push/index.js
@@ -11,6 +11,11 @@ type Props = {
   onClickUserAvatar: (username: string) => void,
   onViewGitRepo: (repoID: string, teamname: string) => void,
 }
+
+const PUSH_DEFAULT = 0
+const PUSH_CREATE = 1
+const PUSH_RENAME = 3
+
 const connectedUsernamesProps = {
   clickable: true,
   colorFollowing: true,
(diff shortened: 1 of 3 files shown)


thebearjew:
This raises a question regarding the timestamp UI.

Should all timestamps remain above messages or should system messages maintain their own inner timestamp (duplicated w/ the one that is above).

shared/chat/conversation/messages/system-git-push/index.js
```diff
@@ -19,87 +24,151 @@ const connectedUsernamesProps = {
...
+              pusher={pusher}
+              timestamp={timestamp}
+              repo={repo}
+              repoID={repoID}
+              team={team}
+              onViewGitRepo={this.props.onViewGitRepo}
+            />
+          </GitPushCommon>
+        )
+      case PUSH_RENAME:
```
thebearjew: Waiting to implement this as there are no designs.

PR #12893, opened 2018-07-20
engine.ResolveAndCheck calls into resolver and then checks proofs

maxtaco:
- uses the identify2_with_uid machinery
- use it in teams where we were previously server-trusting

### DIFF
--- a/go/engine/resolve_identify2.go
+++ b/go/engine/resolve_identify2.go
@@ -15,6 +15,7 @@ type ResolveThenIdentify2 struct {
 	i2eng                 *Identify2WithUID
 	testArgs              *Identify2WithUIDTestArgs
 	responsibleGregorItem gregor.Item
+	queriedName           libkb.NormalizedUsername
 
 	// When tracking is being performed, the identify engine is used with a tracking ui.
 	// These options are sent to the ui based on command line options.
@@ -87,6 +88,8 @@ func (e *ResolveThenIdentify2) resolveUID(m libkb.MetaContext) (err error) {
 		// the resolve assertion was a keybase username or UID, so remove it
 		// from identify2 arg to allow cache hits on UID.
 		e.arg.UserAssertion = ""
+		// But still check on that way out that the username matches the UID
+		e.queriedName = rres.GetNormalizedQueriedUsername()
 	}
 
 	// An optimization --- plumb through the resolve body for when we load the
@@ -113,7 +116,18 @@ func (e *ResolveThenIdentify2) Run(m libkb.MetaContext) (err error) {
 
 	// For testing
 	e.i2eng.testArgs = e.testArgs
-	return RunEngine2(m, e.i2eng)
+	err = RunEngine2(m, e.i2eng)
+	if err != nil {
+		return err
+	}
+
+	// Check the server for cheating on a Name->UID resolution. After we do a userload (by UID),
+	// we should have a merkle-verified idea of what the corresponding name is, so we check it
+	// as a post-assertion here.
+	if !e.queriedName.IsNil() && !libkb.NewNormalizedUsername(e.Result().Upk.GetName()).Eq(e.queriedName) {
+		return libkb.NewUIDMismatchError("bad user returned for " + e.queriedName.String())
+	}
+	return nil
 }
 
 func (e *ResolveThenIdentify2) Result() *keybase1.Identify2Res {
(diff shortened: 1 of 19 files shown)


go/engine/resolve_identify2.go
```diff
@@ -141,3 +141,29 @@ func (e *ResolveThenIdentify2) GetProofSet() *libkb.ProofSet {
...
+		NoErrorOnTrackFailure: true,
+		IdentifyBehavior:      keybase1.TLFIdentifyBehavior_RESOLVE_AND_CHECK,
+	}
+	eng := NewResolveThenIdentify2(m.G(), &arg)
+	err = RunEngine2(m, eng)
+	res := eng.Result()
+	if err != nil {
+		return uid, un, err
+	}
+	return res.Upk.GetUID(), libkb.NewNormalizedUsername(res.Upk.GetName()), nil
```
mlsteele: Could we get `res.Upk.ToUserVersion()`?

protocol/avdl/keybase1/identify.avdl
```diff
@@ -44,7 +44,8 @@ protocol identify {
     boolean noSkipSelf=true,
     boolean canSuppressUI=false,
     TLFIdentifyBehavior identifyBehavior=0,
-    boolean forceDisplay=false
+    boolean forceDisplay=false,
+    boolean actLoggedOut=false
```
mlsteele: woah I didn't know about this syntax. What does it do?

go/engine/resolve_identify2.go
mlsteele: A full UPAK would be great if it's handy, not sure if it is.

protocol/avdl/keybase1/identify.avdl
maxtaco: it's AVDL syntax that allows default values

maxtaco:
@mlsteele I might need another look, a lot of stuff moved around, sorry about that.

maxtaco:
Oh wow, CI passed.  I was thinking I should add another test for an evil keybase resolver.

maxtaco:
Ok, added tests for an evil resolver too

go/teams/service_helper.go
```diff
@@ -866,39 +863,35 @@ func loadUserVersionAndPUKedByUsername(ctx context.Context, g *libkb.GlobalConte
...
 	}
-	if un != "" && !libkb.NormalizedUsername(un).Eq(libkb.NormalizedUsername(upak.Current.Username)) {
-		return keybase1.UserVersion{}, libkb.BadUsernameError{N: un}
-	}
+	return loadUserVersionByUPK2(ctx, g, upak.Current)
+}
 
-	uv := NewUserVersion(upak.Current.Uid, upak.Current.EldestSeqno)
-	if upak.Current.Status == keybase1.StatusCode_SCDeleted {
+func loadUserVersionByUPK2(ctx context.Context, g *libkb.GlobalContext, upak keybase1.UserPlusKeysV2) (keybase1.UserVersion, error) {
```
mlsteele: How about naming this `filterForMemberAddition`? And it doesn't need `g` as a parameter.

go/libkb/interfaces.go
```diff
@@ -786,3 +786,14 @@ type ChatHelper interface {
 		convID chat1.ConversationID, membersType chat1.ConversationMembersType, payload string) (string, error)
 	AckMobileNotificationSuccess(ctx context.Context, pushIDs []string)
 }
+
+type Resolver interface {
```
mlsteele:
```
All results from Resolver are unverified.
```
Add a little note because this may be as far as people get looking for docs.

go/libkb/resolve.go
```diff
@@ -49,6 +50,11 @@ func (res *ResolveResult) GetUID() keybase1.UID {
 	return res.uid
 }
 
+// SetUID should only be used by testing
+func (res *ResolveResult) SetUID(u keybase1.UID) {
```
mlsteele: SetUIDForTesting

go/engine/resolve_identify2.go
```diff
@@ -141,3 +156,49 @@ func (e *ResolveThenIdentify2) GetProofSet() *libkb.ProofSet {
...
+
+// ResolveAndCheck takes as input a name (joe), social assertion (joe@twitter)
+// or compound assertion (joe+joe@twitter+3883883773222@pgp) and invokes
+// the whole resolve/identify machinery. That is, it performs a server-trusted
+// resolution of the name to a UID, then does an ID on the UID to make sure all
+// assertions are met. Pass into it a MetaContext without any UIs set, since it
+// is meant to run without any UI interaction. Also note that tracker statements
+// are *not* taken into account. The identify is run as if the user is logged out.
+// The output, on success, is a populated UserPlusKeysV2.
+func ResolveAndCheck(m libkb.MetaContext, s string) (ret keybase1.UserPlusKeysV2, err error) {
```
mlsteele: This is a mix of things that are part of the function's contract and things that aren't. How about splitting it up like this?

```
// ResolveAndCheck takes as input a name (joe), social assertion (joe@twitter)
// or compound assertion (joe+joe@twitter+3883883773222@pgp) and resolves
// it to a user, verifying the result.
// Pass into it a MetaContext without any UIs set, since it
// is meant to run without any UI interaction. Also note that tracker statements
// are *not* taken into account.
// The output, on success, is a populated UserPlusKeysV2.
func ResolveAndCheck(...
    // Invokes the whole resolve/identify machinery. That is, it performs a server-trusted
    // resolution of the name to a UID, then does an ID on the UID to make sure all
    // assertions are met. The identify is run as if the user is logged out.
```

Also what caching is part of the contract? It will never return old results?

maxtaco: yeah, just upak caching, and it forces fresh, i believe...

PR #12894, opened 2018-07-20
Stellar: Pass reserves from stellard to GUI

zapu:
This info comes from stellard. Right now we only do default reserve (-1XLM) and reserves for each additional trustline, but this can be improved serverside.

cc @buoyad I'll need some extra help with react side

### DIFF
--- a/protocol/avdl/stellar1/common.avdl
+++ b/protocol/avdl/stellar1/common.avdl
@@ -24,7 +24,7 @@ protocol common {
 
   record AccountReserve {
     string amount; // "-1"
-    string description; // e.g. "HUGZ trust line" or "account"
+    string description; // e.g. "HUGZ/Unknown trustline" or "account"
   }
 
   enum TransactionStatus {
--- a/protocol/avdl/stellar1/common.avdl
+++ b/protocol/avdl/stellar1/common.avdl
@@ -22,6 +22,11 @@ protocol common {
     string limit;
   }
 
+  record AccountReserve {
+    string amount; // "-1"
+    string description; // e.g. "HUGZ trust line" or "account"
+  }
+
   enum TransactionStatus {
     NONE_0,
     PENDING_1,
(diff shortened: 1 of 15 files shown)


protocol/avdl/stellar1/common.avdl
```diff
@@ -22,6 +22,11 @@ protocol common {
     string limit;
   }
 
+  record AccountReserve {
+    string amount; // "-1"
+    string description; // e.g. "HUGZ trust line" or "account"
```
mlsteele: HUGZ -> HUGZ/Unknown

mlsteele: I mean.. that's what's in the designs, but whatever you think is fine.

shared/constants/types/wallets.js
```diff
@@ -17,7 +17,7 @@ export opaque type AccountID: string = string
 export const stringToAccountID = __DEV__
   ? (s: string): AccountID => {
       if (!s) {
-        throw new Error('Invalid empty converationidkey. Did you mean Constants.noConversationIDKey?')
+        throw new Error('Invalid empty AccountID. Did you mean Constants.noAccountID?')
```
mlsteele: -> I hope you didn't mean Constants.noAccountID cause there's no such thing

protocol/avdl/stellar1/common.avdl
zapu: what makes a trustline "known"?

shared/constants/types/wallets.js
zapu: hm I just meant this thing: https://github.com/keybase/client/blob/b91a6d5fd20758c5a3428cc8d2a012eab4b13e7a/shared/constants/types/wallets.js#L29

I'm not sure how these symbols are exported

mlsteele: Oh there it is. My bad!

protocol/avdl/stellar1/common.avdl
mlsteele: Showing naked asset codes can be misleading. Here it's probably not a problem. Design seems to have elected for showing the federated name rather than the issuer account ID. But we don't resolve any federated asset names yet.

PR #12899, opened 2018-07-20
Revert "Don't call getTeams inside components, and always reload teamList when it changes"

cjb:
Reverts keybase/client#12887 as @mmaxim asked for.


cjb:
CC @keybase/react-hackers

PR #12900, opened 2018-07-20
[KBFS-3052] Empty Tag

jzila:


shared/actions/fs/index.js
```diff
@@ -34,17 +34,20 @@ function* listFavoritesSaga(): Saga.SagaGenerator<any, any> {
 }
 
 const direntToMetadata = (d: RPCTypes.Dirent) => ({
-  name: d.name,
+  name: d.name.split('/').pop(),
```
zanderz: are we sure there aren't backslashes here on Windows?

jzila: FWIW I tested it on my Windows VM and it worked.

PR #12901, opened 2018-07-20
Expand jest mocks

chrisnojima:
This fixes some jest issues people were running into:

1. Mock feature flags. Now feature flags are fixed and if you want to render something thats flagged off you have to flag it on in the mock
1. hoc-timers is now mocked to do nothing. We shouldn't have timing related stuff in our storyshots
1. local-debug is mocked off
1. platforms is mocked as electron/darwin always, no per platform stuff w/ linux rendering different in ci
1. fixed the chat/inbox hmr breakage


chrisnojima:
@keybase/react-hackers

chrisnojima:
Unclear what's happening in windows. Honestly it's maybe not worth worrying
about as I think it's not really a development platform for this project

PR #12903, opened 2018-07-20
If ReAddMemberAfterReset is no-op, return nil error

mlsteele:
Some explanation in https://keybase.atlassian.net/browse/CORE-8272

This change makes it so that clicking let-them-in button repeatedly does not cause a global error.


mmaxim:
Yeah this error was there when I got here to integrate with chat, and I always suspected it was being overaggressive.

mlsteele:
Yep. Still not sure why it took 40 seconds to hide the button. Oh well

PR #12906, opened 2018-07-20
Work in progress team ek box bug

zapu:
cc @joshblum

### DIFF
--- a/go/systests/ephemeral_test.go
+++ b/go/systests/ephemeral_test.go
@@ -351,7 +351,13 @@ func readdToTeamWithEKs(t *testing.T, leave bool) {
 	tt := newTeamTester(t)
 	defer tt.cleanup()
 
-	user1 := tt.addUser("one")
+	// Make standalone user that will not run gregor. This is
+	// important in the *leave* case, where we want to observe
+	// effects of team key and EK not being rotated.
+	user1 := makeUserStandalone(t, "user1", standaloneUserArgs{
+		disableGregor:            true,
+		suppressTeamChatAnnounce: true,
+	})
 	user2 := tt.addUser("wtr")
 
 	teamID, teamName := user1.createTeam2()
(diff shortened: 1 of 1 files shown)


go/systests/ephemeral_test.go
```diff
@@ -357,11 +357,12 @@ func readdToTeamWithEKs(t *testing.T, leave bool) {
 	teamID, teamName := user1.createTeam2()
 	user1.addTeamMember(teamName.String(), user2.username, keybase1.TeamRole_WRITER)
 
-	ephemeral.ServiceInit(user1.tc.G)
-	ekLib := user1.tc.G.GetEKLib()
-
-	_, err := ekLib.GetOrCreateLatestTeamEK(context.Background(), teamID)
-	require.NoError(t, err)
+	for _, u := range tt.users {
```
joshblum: since user1 is the only one doing operations on the team only they need the `ServiceInit` call

go/systests/ephemeral_test.go
```diff
@@ -370,6 +371,14 @@ func readdToTeamWithEKs(t *testing.T, leave bool) {
 	}
 
 	user1.addTeamMember(teamName.String(), user2.username, keybase1.TeamRole_WRITER)
+
+	maxGen, err := user1.tc.G.GetTeamEKBoxStorage().MaxGeneration(context.Background(), teamID)
```
joshblum: `MaxGeneration` is only a local call and doesn't fetch what is on the server, for these tests it's better to check that the generation increased as you expected and each user can successfully call `Get` which will first try the cache and then try the server

go/systests/ephemeral_test.go
```diff
@@ -347,3 +346,52 @@ func TestEphemeralNewUserEKAndTeamEKAfterRevokes(t *testing.T) {
...
+	var expectedGeneration keybase1.EkGeneration
+	if leave {
+		user2.leave(teamName.String())
+		expectedGeneration = teamEKGeneration
+	} else {
+		user1.removeTeamMember(teamName.String(), user2.username)
+		expectedGeneration = teamEKGeneration + 1
+	}
+
+	// After leaving user2 won't have access to the current teamEK
```
zapu: my worry here is that this code can race user1's rotate key  CLKR handler after leave

go/systests/ephemeral_test.go
zapu: ah makes sense, thank you!

PR #12916, opened 2018-07-23
Order operations in AddMemberTx per UID

zapu:
#### Enforce strict membership ordering by saving seqno per UID.

In AddMemberTx, remember in which payload given UID was affected last and do not queue operations affecting that UID in payloads prior.

~Adds `reflection` code in `transactions.go` so the `findPayload` func does not have to be repeated 3 times for 3 types of payload.~ Reflection code removed in favor of reflection-less "tagged union" pattern.

#### Example of what this is trying to solve:

Assume a transaction that tries to do the following, in this order:
1) Add alice as crypto member,
2) sweep old bob@keybase invite (pukless member),
3) add bob as crypto member.

Currently, this would post two signatures:
1) `change_membership { add: [ alice, bob ] }`
2) `invite { cancel: [ bob@keybase ] }`

With this PR, it's going to post:
1) `change_membership { add: [ alice ] }`
2) `invite { cancel: [ bob@keybase ] }`
3) `change_membership { add: [ bob ] }`

Ideally, we should be posting (assuming add of alice does not depend on anything else), but this would be a bigger rewrite.
1) `invite { cancel: [ bob@keybase ] }`
2) `change_membership { add: [ alice, bob ] }`

### DIFF
--- a/go/teams/transactions.go
+++ b/go/teams/transactions.go
@@ -28,6 +28,7 @@ type AddMemberTx struct {
 	payloads []interface{}
 
 	completedInvites map[keybase1.TeamInviteID]bool
+	uidSeqno         map[keybase1.UID]int
 
 	// Override whether the team key is rotated.
 	SkipKeyRotation *bool
@@ -36,23 +37,11 @@ type AddMemberTx struct {
 type txTeamInvitesKeybase struct{ Val SCTeamInvites }
 type txTeamInvitesSocial struct{ Val SCTeamInvites }
 
-// getInviteSection returns SCTeamInvites field from either
-// *txTeamInvitesKeybase or *txTeamInvitesSocial without reflection.
-func getInviteSection(payload interface{}) (*SCTeamInvites, error) {
-	switch p := payload.(type) {
-	case *txTeamInvitesKeybase:
-		return &p.Val, nil
-	case *txTeamInvitesSocial:
-		return &p.Val, nil
-	default:
-		return nil, fmt.Errorf("getInviteSection: invalid type %T", p)
-	}
-}
-
 func CreateAddMemberTx(t *Team) *AddMemberTx {
 	return &AddMemberTx{
 		team:             t,
 		completedInvites: make(map[keybase1.TeamInviteID]bool),
+		uidSeqno:         make(map[keybase1.UID]int),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


mlsteele:
Thanks for the example. In the example it used to post 2 links but now posts 3. Why is this better?

zapu:
In this example:
1) `change_membership { add: [ alice, bob ] }`
2) `invite { cancel: [ bob@keybase ] }`

If there's old version of bob as pukless member before 1 is posted, then after 1 is posted there will be two bobs in the team. Which we allow right now, but we try to avoid to ultimately block this on the server at later time.

go/teams/transactions.go
```diff
@@ -20,24 +21,27 @@ import (
...
 
 	completedInvites map[keybase1.TeamInviteID]bool
+	uidSeqno         map[keybase1.UID]int
 
 	// Override whether the team key is rotated.
 	SkipKeyRotation *bool
 }
 
+type txTeamInvitesKeybase struct{ Val SCTeamInvites }
+type txTeamInvitesSocial struct{ Val SCTeamInvites }
```
mlsteele: Lots of freaky reflection going on now. I know go is lacking in this department. What do you think about one of these alternatives?

```
// Exactly one of these must be filled.
type txEntry struct {
  Cryptomembers *keybase1.TeamChangeReq
  InvitesKeybase *SCTeamInvites
  InvitesSocial *SCTeamInvites
}
```

```
type txEntryType string
const txEntryType_Cryptomembers txEntryType = "cryptomembers" // -> keybase1.TeamChangeReq
const txEntryType_InvitesKeybase txEntryType = "invites_keybase" // -> SCTeamInvites
const txEntryType_InvitesSocial txEntryType = "invites_social" // -> SCTeamInvites

type txEntry struct {
  Kind string
  Val interface{}
}
```

zapu: That could work, thanks for thinking about it! I'm thinking about going with option 2). The helper functions that do the real type casting are still needed, and with this kind of tagging we can avoid reflection and still keep payload finding function not duplicated.

To be honest I was hesitating before using `"reflect"` but I noticed `lru/lru.go` uses it, so I decided to go for it, and only used basic features that lru uses.

zapu:
@mlsteele can you take another look? thanks

go/teams/transactions.go
```diff
@@ -21,28 +21,41 @@ import (
...
 	team     *Team
-	payloads []interface{} // *SCTeamInvites or *keybase1.TeamChangeReq
-
-	// We need two separate payloads for social invites and keybase
-	// invites because they cannot be mixed in one link.
-	isForSocialInvites map[int]bool
+	payloads []txPayload
 
 	completedInvites map[keybase1.TeamInviteID]bool
+	uidSeqno         map[keybase1.UID]int
```
mlsteele: Add comment for what this is. If I understand it's the index of the first payload in which the uid is mentioned. Also not sure 'seqno' is right. Maybe `uidFirstPayload`

go/teams/transactions.go
```diff
@@ -53,73 +66,75 @@ func (tx *AddMemberTx) IsEmpty() bool {
...
+		tx.uidSeqno[forUID] = len(tx.payloads)
+	}
+	ret := txPayload{
+		Tag: tag,
+	}
+	switch tag {
+	case txPayloadTagCryptomembers:
+		ret.Val = &keybase1.TeamChangeReq{}
+	case txPayloadTagInviteKeybase, txPayloadTagInviteSocial:
+		ret.Val = &SCTeamInvites{}
```
mlsteele: default panic (or add error path)

zapu: added below

PR #12920, opened 2018-07-23
More accurate orange line display

nathunsmitty:
Plumbs `readMsgID` from the client and uses it to set a more accurate orange line!

Note that the experience isn't super great if you're in a conversation with lots of new messages, and you'll have to scroll up to see the orange line.


shared/chat/conversation/messages/wrapper/wrapper-author/container.js
```diff
@@ -12,8 +12,8 @@ const mapStateToProps = (state: TypedState, {message, previous, innerClass, isEd
   const isYou = state.config.username === message.author
   const isFollowing = state.config.following.has(message.author)
   const isBroken = state.users.infoMap.getIn([message.author, 'broken'], false)
-  const lastPositionOrdinal = state.chat2.orangeLineMap.get(message.conversationIDKey)
```
nathunsmitty: I'm not super familiar with the difference between ordinals and message ids, but it seems like if we're only setting the orange line on the load of an inbox then we don't need to use our own ordinals and can just use message ids provided. Definitely could be wrong though, please lmk

shared/constants/chat2/meta.js
```diff
@@ -260,6 +262,7 @@ export const makeConversationMeta: I.RecordFactory<_ConversationMeta> = I.Record
   notificationsMobile: 'never',
   offline: false,
   participants: I.List(),
+  readMsgID: -1,
```
nathunsmitty: seemed like a good default value

PR #12922, opened 2018-07-23
WIP: New Provision Part 4: Updates to some views

chrisnojima:
- [x] New unified select other device
- [ ] New unified public name
- [ ] New unified enter paper key
- [ ] Update font here and on code page


chrisnojima:
@cecileboucheron set public name is also updated

chrisnojima:
@cecileboucheron paperkey page is up now

cecileboucheron:
* Let's use the `small` list item - 40px high on desktop, 48 on mobile, and using 32px icons (specs in styleguide)
* Troubleshooting box should be within the list, so not visible at first if you have a long list of devices
* "Troubleshooting" label should be BodySmallSemibold
* "If you have" text below should be BodySemibold
* That may be a separate PR but it would be nice if the Back button from the step after was taking back to this list 
* Input text in Type your paper key screen should be use the Header styles: 16px semibold, just like for the secret key one
* Name of the paper key above should be Header (we could use the Name with icon thing here)
* Add some spacing between it and the input on desktop.

chrisnojima:
the back button cant' work that way due to issues w/ the daemon, so thats not changing anytime soon

chrisnojima:
@cecileboucheron updated ptal

cecileboucheron:
* Left/right edges of troubleshooting box should be left-aligned with list 
* `Reset your account` link should be semibold too and underlined on hover
* Paper key input should be using Source code pro font (the rest of the styles stay)
* Still missing some gap on desktop on this page between the paper key name and the input

shared/provision/select-other-device/index.js
```diff
@@ -0,0 +1,133 @@
...
+        <Box2 direction="vertical" fullWidth={true} style={styles.contentBox} gap={'medium'}>
+          <Text type={isMobile ? 'BodyBig' : 'Header'} style={styles.header}>
+            For security reasons, you need to authorize with an existing device. Which of your existing
+            devices would you like to use?
+          </Text>
+          <List
+            style={styles.list}
+            items={items}
+            renderItem={this._renderItem}
+            keyProperty="name"
```
cjb: I see a yellow box key warning on Android on this screen -- perhaps this should be `keyProperty="key"`?

chrisnojima: fixed: https://github.com/keybase/client/pull/12922/commits/cf09460808bdeb99e116e2ef08b083c400c78a12

chrisnojima:
@cecileboucheron fixed your stuff

chrisnojima:
@cjb can't repro the back button thing. works for me provisioning and being provisioned

chrisnojima:
Doing the font now

PR #12925, opened 2018-07-23
Add hover styles for inbox rows.

adamjspooner:
@keybase/react-hackers, this is ready for review. I don't like the nested ternary. I'm open to suggestions for making it better. The comment above it explains what's going on.


shared/chat/inbox/row/small-team/index.js
```diff
@@ -39,23 +39,51 @@ type Props = {
...
+  ? ClickableBox
   : glamorous(Box)({
       '& .small-team-gear': {display: 'none'},
       ':hover .small-team-gear': {display: 'unset'},
       ':hover .small-team-timestamp': {display: 'none'},
     })
 
-class SmallTeam extends React.PureComponent<Props> {
+class SmallTeam extends React.PureComponent<Props, State> {
+  constructor(props: Props) {
```
thebearjew: Perhaps you can get away with `state = { isHovered: false }` for some shorthand state initialization.

shared/chat/inbox/row/small-team/top-line.js
```diff
@@ -90,7 +90,7 @@ class _SimpleTopLine extends React.Component<Props> {
           style={platformStyles({
             common: {
               ...boldOverride,
-              backgroundColor: this.props.backgroundColor,
+              // backgroundColor: this.props.backgroundColor,
```
thebearjew: Should this be removed?

shared/chat/inbox/row/small-team/index.js
```diff
@@ -39,23 +39,51 @@ type Props = {
...
+
+  _backgroundColor = () =>
+    // props.backgroundColor should always override hover styles, otherwise, there's a
+    // moment when the conversation is loading that the selected inbox row is styled
+    // with hover styles instead of props.backgroundColor.
+    this.props.isSelected
+      ? this.props.backgroundColor
+      : this.state.isHovered
+        ? globalColors.blue4
+        : this.props.backgroundColor
```
thebearjew: Maybe return early if it's selected and use a ternary for `isHovered?

```js
if (this.props.isSelected) return this.props.backgroundColor

return this.state.isHovered ? globalColors.blue4 : this.props.backgroundColor
```

adamjspooner: I've seen this logic elsewhere in the app (the nested ternary operators). I assume it's to reduce `return`s and brackets and etc. So, I'm not sure if this is a better or just different.

I'm wondering if how I'm approaching this is wrong. I mean, this works, but is there a better way to achieve what I want without even needing this?

Thanks for the feedback, though!

thebearjew: yeah no problem! I mean everything is just binary in the end so...

 👾 do ✨ what  🌈 you 🐰 feel 🦄  like 🍠  man! 🥇

shared/chat/inbox/row/big-team-channel/index.js
```diff
@@ -23,12 +25,23 @@ type Props = {
...
 class BigTeamChannel extends PureComponent<Props> {
   render() {
     return (
       <ClickableBox onClick={this.props.onSelectConversation} style={styles.container}>
         <Box style={styles.rowContainer}>
-          <Box style={this.props.isSelected ? styles.selectedChannelBackground : styles.channelBackground}>
+          <HoverBox
+            style={collapseStyles([
+              styles.channelBackground,
+              ...(this.props.isSelected ? [styles.selectedChannelBackground] : []),
```
chrisnojima: you don't need to spread like this, just do
```
collapseStyles([
              styles.channelBackground,
              this.props.isSelected && styles.selectedChannelBackground,
]
```

shared/chat/inbox/row/big-team-channel/index.js
```diff
@@ -100,15 +115,7 @@ const styles = styleSheetCreate({
     isElectron: desktopStyles.clickable,
   }),
   selectedChannelBackground: {
-    ...globalStyles.flexBoxRow,
-    ...(isMobile ? globalStyles.fillAbsolute : {width: '100%'}),
```
chrisnojima: i don't think this is safe to get rid of. you'll no longer have correct wrapping

shared/chat/inbox/row/small-team/index.js
```diff
@@ -39,23 +39,48 @@ type Props = {
...
   : glamorous(Box)({
       '& .small-team-gear': {display: 'none'},
       ':hover .small-team-gear': {display: 'unset'},
       ':hover .small-team-timestamp': {display: 'none'},
     })
 
-class SmallTeam extends React.PureComponent<Props> {
+class SmallTeam extends React.PureComponent<Props, State> {
+  state = {
+    isHovered: false,
```
chrisnojima: why can't we just use css for this?

shared/chat/inbox/row/small-team/index.js
```diff
@@ -108,16 +133,6 @@ const styles = styleSheetCreate({
   container: {flexShrink: 0, height: RowSizes.smallRowHeight},
   conversationRow: {
     ...globalStyles.flexBoxColumn,
-    backgroundColor: isMobile ? globalColors.fastBlank : globalColors.blue5,
```
chrisnojima: removing my fastblank will likely really increase layering on android

shared/chat/inbox/row/small-team/index.js
adamjspooner: Is there a way to do it with CSS and it override `props.backgroundColor`? That's the wall I ran into.

shared/chat/inbox/row/big-team-channel/index.js
adamjspooner: We need wrapping on the channel in the inbox?

shared/chat/inbox/row/small-team/index.js
chrisnojima: instead of inlining the color you could just have it be in the css and use a classname, so the hover / not-hover would be defined using a classname so like the component would just have 
<SmallTeamBox className={this.props.selected ? 'selected-row' : 'row' />. this is likely a larger change

shared/chat/inbox/row/big-team-channel/index.js
chrisnojima: make a long channel name and make sure

shared/chat/inbox/row/small-team/index.js
adamjspooner: The other part of the problem is that we need to plumb the color through to `Avatars` where it uses the background color for the border on multi-avatars. So, I figured I'd do it in one place and use it on both.

shared/chat/inbox/row/small-team/index.js
adamjspooner: I've added `fastBlank` back in with `collapseStyles`.

shared/chat/inbox/row/small-team/index.js
adamjspooner: @chrisnojima, do you have any suggestions (WRT my last comment)?

PR #12926, opened 2018-07-23
Chat reactions: desktop integration

buoyad:
Full integration on desktop. Also adds a feature flag for it. I ran into issues with circular dependencies in storybook, which is why `ReactButtonWithTooltip` is factored out into a separate file. r? @keybase/react-hackers


shared/chat/conversation/list-area/normal/index.desktop.js
```diff
@@ -1,4 +1,5 @@
 // @flow
+/* eslint-env browser */
```
chrisnojima: why is this necessary?

buoyad: eslint was marking `MouseEvent` as undefined

chrisnojima: you're using the wrong type: https://github.com/facebook/flow/blob/v0.76.0/lib/react-dom.js#L193

buoyad: oh it's actually complaining about `HTMLInputElement` and `HTMLTextAreaElement` as well

PR #12927, opened 2018-07-23
Polish chat inbox locked out row

nathunsmitty:
Adjusts the line height to make the chat locked out row a little nicer.

screen shot 2018-07-23 at 3 56 06 pm

Also some general style refactoring + adding some inbox rows to storybook.


shared/styles/css.js.flow
```diff
@@ -265,7 +265,7 @@ export type _StylesDesktop = {|
   left?: Dimension,
   letterSpacing?: string,
   lineBreak?: string,
-  lineHeight?: string,
+  lineHeight?: Dimension,
```
nathunsmitty: Guessing we don't want to type line heights as strings?

chrisnojima: i mentioned this in the other pr i think we don't

shared/react-native/ios/Keybase/Storybook.m
```diff
@@ -24,7 +24,7 @@ + (BOOL)requiresMainQueueSetup
 - (NSDictionary *)constantsToExport
 {
   // Set this to true to enable storybook mode
-  return @{@"isStorybook": @false};
+  return @{@"isStorybook": @true};
```
chrisnojima: 🔥

chrisnojima:
lets just rename the bottom part to say 'You are currently locked out' or something

cecileboucheron:
Even shorter: `You are locked out.`

nathunsmitty:
Also caught a bug where the "Rekey Needed" alert was centering itself in mobile and taking up the full width on desktop. It should behave just like the "Reset" alert now.
screen shot 2018-07-23 at 4 56 10 pm

nathunsmitty:
And here's how the one-line locked out text looks! I think this is ready for another look
screen shot 2018-07-23 at 4 56 02 pm

nathunsmitty:
@keybase/react-hackers 👀?

shared/chat/inbox/row/small-team/bottom-line.js
```diff
@@ -159,53 +114,74 @@ class BottomLine extends PureComponent<Props> {
...
+    paddingLeft: globalMargins.xtiny,
+    paddingRight: globalMargins.xtiny,
+  },
+  rekeyNeededText: platformStyles({
+    common: {
+      color: globalColors.white,
+    },
+    isElectron: {
+      fontSize: 11,
+      lineHeight: '13px',
```
chrisnojima: this should just be  13, platformstyles does the cast for you

shared/chat/inbox/row/small-team/bottom-line.js
```diff
@@ -159,53 +114,74 @@ class BottomLine extends PureComponent<Props> {
...
+    },
+    isMobile: {
+      fontSize: 12,
+      lineHeight: 14,
+    },
+  }),
+  youAreResetText: platformStyles({
+    isElectron: {
+      fontSize: 12,
+      lineHeight: '13px',
```
chrisnojima: same here

shared/chat/inbox/row/small-team/bottom-line.js
```diff
@@ -230,8 +206,18 @@ const styles = styleSheetCreate({
...
-  bottomLineBox: {
+  contentBox: {
     width: '100%',
   },
+  alertMeta: platformStyles({
+    common: {
+      alignSelf: 'center',
+      marginRight: 6,
+    },
+    isElectron: {},
```
chrisnojima: 🔥

shared/chat/inbox/row/small-team/bottom-line.js
nathunsmitty: that gives a flow error because lineHeight has to be a string

chrisnojima: that type is wrong then, it should be Dimension

PR #12929, opened 2018-07-23
Fix mobile files

jzila:
@thebearjew thanks for finding this :)


jzila:
Merging without Circle since this fixes a break.

PR #12934, opened 2018-07-23
upload attachments in the background, and display a local notification if we fail CORE-5825

mmaxim:
Patch does the following:

1.) Add `ActiveDeliveries` to `Deliverer` to indicate to outside services if we are actively trying to send any messages.
2.) Add `NextFailure` to `Deliver` to notify outside services that a message send has failed. The caller is responsible for calling a cleanup function from this call. 
3.) Upgrades `AppDidEnterBackground` to determine if there are any pending messages. If so, we tell native that we would like to run in the background to try and send them.
4.) Add `AppBeginBackgroundTask` which will wait for 10 minute for pending messages to get sent. If the app is kill in the meantime, then it aborts.
5.) Upgrade `AppWillExit` to throw up a local push notification if we are about to be suspended and we are still tryin to deliver messages. 
6.) Add a bunch of extra debugging to the attachments `Store`. 
7.) Improve chat log tags
8.) Fix a bug where `ReadB64` could be called before we set up the loopback connection and crash the app.

### DIFF
--- a/go/bind/keybase.go
+++ b/go/bind/keybase.go
@@ -394,7 +394,7 @@ func unboxNotification(ctx context.Context, strConvID, body string, intMembersTy
 
 func pushPendingMessageFailure(convID chat1.ConversationID, pusher PushNotifier) {
 	pusher.LocalNotification("failedpending",
-		"Failed to deliver pending messages. Open the Keybase app to retry.",
+		"Heads up! One or more pending messages failed to send. Tap here to retry them.",
 		-1, "default", convID.String(), "chat.failedpending")
 }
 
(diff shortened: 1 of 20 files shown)


go/bind/keybase.go
```diff
@@ -391,21 +398,131 @@ func unboxNotification(ctx context.Context, strConvID, body string, intMembersTy
...
+			abort = true
+			return errors.New("app state change")
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	})
+	g.Go(func() error {
+		for {
+			ch, cancel := kbChatCtx.MessageDeliverer.NextFailure()
+			defer cancel()
```
patrickxb: i think it would make sense to remove this defer and call `cancel` before the return on line 483.  otherwise the `defer` stack will pile up with these funcs while the for loop runs...

mmaxim: Sick point, will do

PR #12942, opened 2018-07-24
Dedup reactions in supersedes

joshblum:
changes the `ReactionMap` struct to hold a map of `{username: reactionMessageID}` instead of an array so we can easily dedup multiple reactions from the same user.

cc @buoyad

### DIFF
--- a/go/chat/convsource_test.go
+++ b/go/chat/convsource_test.go
@@ -429,12 +429,9 @@ func TestReactions(t *testing.T) {
 	t.Logf("test +1 reaction")
 	reactionMsgID := sendReaction(":+1:", msgID)
 	expectedReactionMap := chat1.ReactionMap{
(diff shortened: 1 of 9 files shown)


mmaxim:
I made a comment on the ticket, but I'll just repeat here. I think we should probably solve this problem on the send side, it shouldn't be possible to get the same reaction our per message per user.

joshblum:
commented on ticket

mmaxim:
Wait now I’m confused. I think users should be able to react more than once, just not with the same emoji. This looks like it prevents me from adding multiple different reactions.

joshblum:
You can still react more than once since you'll appear in different entries in the `reactionMap`, you just can't appear more than once under the same `reactionText` 
`{ reactionText : { username : messageID } }`

PR #12945, opened 2018-07-24
Don't count currently selected conversation for back button badge

buoyad:
Adds a reduce on `state.chat2.badgeMap`. You can notice the difference in behavior if you receive some messages while in the inbox - currently on navigating into the chat the badge will flash in next to the back button. r? @keybase/react-hackers


mmaxim:
YES

PR #12947, opened 2018-07-24
Make PassphraseError Msg say it's bad passphrase

zapu:
Also fix linter errors


go/libkb/passphrase_login.go
```diff
@@ -148,8 +148,7 @@ func pplPost(m MetaContext, eOu string, lp PDPKALoginPackage) (*loginAPIResult,
 		return nil, err
 	}
 	if res.Status.Code == SCBadLoginPassword {
-		err = PassphraseError{"server rejected login attempt"}
-		return nil, err
+		return nil, PassphraseError{"Invalid password. Server rejected login attempt."}
```
maxtaco: can you call it `passphrase` since we say that everywhere else?

zapu: oops, sure!

PR #12950, opened 2018-07-24
Add self check in orange line

nathunsmitty:
Adds a check in the orange line display logic to ensure that a message you send after the last read message doesn't trigger the orange line. (thanks to @buoyad for finding this!)

Also adds types to ownProps because it had a type of `empty` before.

### DIFF
--- a/shared/chat/conversation/messages/wrapper/wrapper-timestamp/container.js
+++ b/shared/chat/conversation/messages/wrapper/wrapper-timestamp/container.js
@@ -7,25 +7,25 @@ import {formatTimeForMessages} from '../../../../../util/timestamp'
 
 const mapStateToProps = (
   state: TypedState,
-  props: {
+  ownProps: {
     message: Types.Message,
     previous: ?Types.Message,
   }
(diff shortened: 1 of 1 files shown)


shared/chat/conversation/messages/wrapper/wrapper-timestamp/container.js
```diff
@@ -1,19 +1,31 @@
...
 import {setDisplayName, compose, connect, type TypedState} from '../../../../../util/container'
 import {formatTimeForMessages} from '../../../../../util/timestamp'
 
-const mapStateToProps = (state: TypedState, {message, previous}) => {
-  const lastReadMessageID = state.chat2.lastReadMessageMap.get(message.conversationIDKey)
-  // Show the orange line on the first message after the last unread message, if there is one
-  const orangeLineAbove = !!previous && lastReadMessageID === previous.id
+const mapStateToProps = (
+  state: TypedState,
+  props: {
```
chrisnojima: lets pull this out as type OwnProps

PR #12953, opened 2018-07-24
WIP: Inbox hover styles

adamjspooner:
@chrisnojima, this is how I _wanted_ to do the hover styles. The problem I ran into was the border color on multiavatars on hover because it's getting it from `props.backgroundColor`. I'm using red here to exaggerate the problem:
screen shot 2018-07-24 at 1 14 06 pm

I'd rather not use state. It seems pretty flakey (as @buoyad pointed out). Do you have any suggestions?

(Also, it looks like the `div` making the border is a pixel or two too wide. I just noticed after taking the screenshot.)

### DIFF
--- a/shared/chat/inbox/row/small-team/index.js
+++ b/shared/chat/inbox/row/small-team/index.js
@@ -39,43 +39,25 @@ type Props = {
   youNeedToRekey: boolean,
 }
 
-type State = {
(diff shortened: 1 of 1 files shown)


chrisnojima:
lets just keep it w/ state and we can make it more conducive to css later.

PR #12955, opened 2018-07-24
Add messageType to query

joshblum:
`TestChatSrvSetConvMinWriterRole` was flaking when pulling the thread, this adds a `MessageType` to the thread query to not flakey

cc @mmaxim will merge after ci


joshblum:
Doesn’t affect circle, merging

PR #12961, opened 2018-07-24
Send frontend paymentNotification when gregor msg handled

patrickxb:


mmaxim:
This is an IBM because we can't miss it?

patrickxb:
OOBM should be fine.  I'll change it.

patrickxb:
@keybase/react-hackers can you fix this?  CI complaining.  thank you!

chrisnojima:
looking

chrisnojima:
@patrickxb that should fix it. i assume we want these notifications...

PR #12962, opened 2018-07-24
Git looks for pre-reset devices

mlsteele:


joshblum:
can you add to `put_get_test.go` so we'll know if anything else breaks?

mlsteele:
Is it possible to reset an account in `git` tests?

joshblum:
i think so they just use `kbtest` users

mlsteele:
added regression test

PR #12963, opened 2018-07-24
move minWriterRole to convSettings

joshblum:
Creates a `ConvSettings` object to hold `MinWriterRoleInfo` and channel wide explosion policies in the future. Coming soon there will be to a `ConvSettings` and `TeamConvSettings` which can set generic server enforced rules about the conversation/team set by team admins. `ConvSettings` will have priority over `TeamConvSettings` if both are set.

depends on https://github.com/keybase/keybase/pull/2702


joshblum:
(circle fail expected until protocol changes are in kbfs)

PR #12970, opened 2018-07-25
New Provision: switch qr back to using image, update qr code to generate blue

chrisnojima:
The original implementation use an image for the qr code. During development of new provision i switched to using divs so we had a lot more control of styling but android really chokes on this plus initializing the camera at the same time so i went back and used image and changed the lib to output blue also

@keybase/react-hackers @cjb 
cc: @thebearjew since this touches image which orientedimage uses. Looks fine to me on desktop and ios

This also kills the huge update to storyshots


chrisnojima:
got an ok to merge this down into the root pr from @buoyad

PR #12977, opened 2018-07-25
Team: Add People fixes

cjb:
@keybase/react-hackers CC @cecileboucheron

* Add button now shows the number of people that will be added
* Clicking search input "x" only clears the search selections
* Show "Already a member" next to members, make their avatar transparent and don't color their rows on hover
* Remove blue + button
* Clicking "Add" brings up the role picker
* Allow customizing the title, add button label on the role picker
* Allow hiding the "Back" label on the back button in header hoc header


shared/common-adapters/back-button.desktop.js
```diff
@@ -18,11 +18,12 @@ class BackButton extends Component<Props> {
     return (
       <div style={collapseStyles([styles.container, this.props.style])} onClick={e => this.onClick(e)}>
         <Icon type="iconfont-arrow-left" style={styles.icon} color={this.props.iconColor} />
-        {this.props.title !== null && (
-          <Text type="BodyPrimaryLink" style={this.props.textStyle} onClick={e => this.onClick(e)}>
-            {this.props.title || 'Back'}
-          </Text>
-        )}
+        {this.props.title !== null &&
```
chrisnojima: why not just pass title = null?

shared/teams/add-people/container.js
```diff
@@ -63,11 +74,13 @@ const mapDispatchToProps = (dispatch: Dispatch, {navigateUp, routePath, routePro
...
         {
           props: {
+            addButtonLabel: 'Add',
             allowOwner,
+            headerTitle: 'Add them as:',
             onComplete,
             selectedRole: role,
-            sendNotificationChecked: sendNotification,
-            showNotificationCheckbox: true,
+            sendNotificationChecked: true,
```
chrisnojima: why is this true always now?

shared/teams/role-picker/index.js
```diff
@@ -240,6 +208,49 @@ export const RoleConfirm = ({
...
+  },
+  headerBox: {
+    marginBottom: globalMargins.small,
+    marginTop: globalMargins.small,
+  },
+  headerTitle: {
+    color: globalColors.black_40,
+  },
+  promptBox: {
+    margin: globalMargins.tiny,
```
chrisnojima: using box 2 likely removes the need for some of these margins

cecileboucheron:
Looking so much better already! A few comments:
* The blue check on the right side should show up for people already in list up top, not for people already in team. 
* These people (already in list up top) should show a disabled avatar and no hover state (right now they look like you can add them again)
* Window should be 620 wide by 520 maximum with 40px margins around it. If the viewport gets smaller than the total of all that, window should shrink and scroll. Right now it's going edge to edge (screenshot).
* No back button (right now there's even 2) - on the mockup there was one because you were coming from the `Name your team` step.
image

shared/common-adapters/back-button.desktop.js
cjb: Ah, because the page has a title -- I'm using `<HeaderHocHeader title={..}`>, and that component passes the same title it receives down to `<BackButton>` here, so passing a null title isn't an option given that I want the header title.

shared/teams/add-people/container.js
cjb: Because the design doesn't have the send notification checkbox in anymore -- @cecileboucheron, if we aren't showing the option anymore, would we like the notification to be always on or off?

cjb:
@cecileboucheron Thanks!  I'll work on these and let you know when it's ready for another look.

cjb:
@cecileboucheron @chrisnojima Ready for another look!

shared/search/result-row/container.js
```diff
@@ -12,12 +14,17 @@ export type OwnProps = {|
...
   onShowTracker?: () => void,
+  searchKey: string,
 |}
 
 const emptySearch = makeSearchResult()
 
 const mapStateToProps = (state: TypedState, ownProps: OwnProps) => {
+  const {searchKey} = ownProps
+  const selectedIds = getUserInputItemIds(state, {searchKey})
+  const parsedIds = selectedIds.map(id => parseUserId(id))
```
chrisnojima: all this work should likely happen in merge props

cjb:
@chrisnojima Good point, thanks, done.

cecileboucheron:
* The search should re-initialize after you picked someone, meaning:
** Results should disappear, and recommendations should appear (right now results aren't cleared up)
** Service filters should still be there (right now they disappear)
** Focus should be brought up to top input, after the last pill added, ideally
* The word "Back" seems to be on the wrong Z-index:
image
* Modal window should have rounded corners at the top (8px)
* Let's add an 8px spacing (tiny) to the bottom of the filters, they're super stuck to the content below right now:
image
image

cecileboucheron:
More feedback on mobile:
* Add button should be located in the upper right corner. Design: https://zpl.io/VKEEwqq (let's drop the bottom bar completely)
* Let's drop the blue `+` button next to user pills and make sure we bring the focus in all cases after picking someone
* Same filters disappearing/search results not going away issue as on desktop

cjb:
@cecileboucheron These should be all done now!  The back zIndex bug is still there, but that's being fixed in #13066.

@chrisnojima New large changes:

* HeaderHocHeader now supports an action label in the top right on mobile, we didn't have that before.

* If you're doing a profile or team search, we want to show recommendations again after every selection, but if you're building a conversation then we don't (because we want to show any matching convo with the participants you selected).  The old way to parameterize this probably would have been adding a special Hoc; I just added a saga that tests whether we're building a convo instead.

* Service filter logic was busted, testing a prop that was never set in the codebase. New logic after talking to Cecile: profile and team searches just always show the service filter.  Chat new convo building only shows the service filter if you're in the middle of typing a username.

cecileboucheron:
Soooo much better!
My last comments:
* Let's add a right margin to the blue check on mobile - right now it's right on the edge
image
* Add button on mobile should be disabled (30% opaque) and not clickable when list is empty (right now it lets you go through)
* "•" char before "Already in team" should only be visible if there's a full name before
image

cjb:
@cecileboucheron k, have done these too now.

cecileboucheron:
LGTM!

PR #12994, opened 2018-07-26
Don't use non-qualified code points for emoji maps

buoyad:
The non qualified code points don't include modifiers to render characters as actual emojis so they fall back to text (e.g. ❤️ vs ❤, which look the same but github renders the second smaller. it's a [heavy black heart](http://unicodeheart.com/)). This falls back _again_ on iOS and MacOS to the emoji, which results in an emoji like the various kissing ones rendering as `👨‍❤️‍💋‍👩`. It's fixed by removing the non qualified code points from our map. r? @keybase/react-hackers


buoyad:
@akalin-keybase nothing changes for desktop so I don't know if tests will ensure much. I'll add some more emoji ones anyway.

shared/markdown/generate-emoji-parser.js
```diff
@@ -35,9 +42,6 @@ function genEmojiData() {
       )
     }
     addEmojiLiteral(emoji.unified, emoji.short_name)
-    if (emoji.non_qualified) {
```
akalin-keybase: Ah, looking at this again, I don't think this is correct. I'm guessing you're doing this because in
```
const emojiIndexByName = ${JSON.stringify(invert(emojiIndexByChar))}
```
the emoji name will point to the non-qualified char, which will cause problems on mobile.

But we also use `emojiIndexByChar` (and `emojiRegex`/`emojiExp`) to render `NativeEmoji` in `parser.pegjs`:
```
   while ((match = emojiExp.exec(emojiText)) !== null) {
     results.push(emojiText.substring(idx, match.index))
     results.push({type: 'native-emoji', children: [emojiIndexByChar[match[0]]]})
     idx = match.index + match[0].length
   }
```
so removing non-qualified versions means that we won't convert them to native-emoji objects, which seems wrong.

I think what we want to do is have separate `emojiIndexByCharMain` and `emojiIndexByCharNonQualified` lists, and have `emojiIndexByName` be the inverse of `emojiIndexByCharMain` (possibly barfing if there are duplicates), and then merge `emojiIndexByCharMain` and `emojiIndexByCharNonQualified` into `emojiIndexByChar`.

And I guess we should have tests for qualified and non-qualified native emojis turning into `native-emoji` objects with respective qualified versions (or do we always want to convert non-qualified to qualified? not sure.)

buoyad: About your last question, it looks like unicode leaves it up to [implementation](http://unicode.org/reports/tr51/#Emoji_Implementation_Notes) whether to render non-qualified versions as emoji. I think for display we should always use the qualified version, since that's the one iOS accepts for the more problematic ones. I'll change the maps so we'll also handle non-qualified on the input as well. ty for looking again!

buoyad:
OK @akalin-keybase, I added the non qualified versions to `emojiIndexByChar` and added a test for the emojis that contain both variations. PTAL!

shared/markdown/__test__/parser.test.js
```diff
@@ -23,6 +24,11 @@ function check(md, options) {
   }
 }
 
+function codePointsToChar(series) {
```
akalin-keybase: seems to belong inside the new test IMO

shared/markdown/__test__/parser.test.js
```diff
@@ -124,6 +130,31 @@ describe('Markdown parser', () => {
...
   it('parses native zwj emoji correctly', () => {
     check('👩‍❤️‍💋‍👩 👩‍👩‍👧‍👧!')
+    check('👩‍❤️‍💋‍👨 👨‍👧 👨‍👦 👨‍👩‍👧‍👦 👨‍👨‍👧‍👧 👩‍👩‍👧‍👦 👩‍❤️‍👩')
+  })
+  it('parses qualified and non-qualified emoji identically', () => {
+    // Check that the trees are identical
+    // and contain the same emoji
+    const verify = (asts, emojiString) => {
+      asts.forEach(ast => {
+        expect(ast.type).toBe('markup')
```
akalin-keybase: i think you can replace this entire block with something like

```
expect(ast).toEqual({
  type: 'markup',
  children: [{
    type: 'text-block',
    children: [{
      type: 'native-emoji',
      children: [emojiString],
    }]
  }]
})
```
and maybe inline it below

shared/markdown/__test__/parser.test.js
```diff
@@ -124,6 +130,31 @@ describe('Markdown parser', () => {
...
+        expect(ast.children[0].children[0].type).toBe('native-emoji')
+        expect(ast.children[0].children[0].children.length).toBe(1)
+        expect(ast.children[0].children[0].children[0]).toBe(emojiString)
+      })
+    }
+
+    emojiData.forEach(emoji => {
+      if (emoji.unified && emoji.non_qualified) {
+        // check that both unified and non_qualified parse to same thing
+        const tests = [emoji.unified, emoji.non_qualified].map(codePointsToChar)
```
akalin-keybase: seems clearer to me to do a single map with a test function, e.g.
```
const emojiName = `:${emoji.short_name}:`
const expectedAst = { type: 'markup', /* as above ... */ emojiName }
[emoji.unified, emoji.non_qualified].map(codePoints => {
  const emoji = codePointsToChar(codePoints)
  const ast = parser.parse(emoji)
  expect(ast).toEqual(expectedAst)
})
```

akalin-keybase: may as well check that the name maps back to the qualified version, e.g.

```
expect(emojiIndexByName[emojiName].toBe(emoji.unified))
```

akalin-keybase:
Lgtm!

PR #12997, opened 2018-07-26
Update “Join a team” styles.

adamjspooner:


shared/teams/join-team/index.native.js
```diff
@@ -1,52 +1,77 @@
...
     <Box style={globalStyles.flexBoxColumn}>
-      <Box
-        style={{
-          ...globalStyles.flexBoxColumn,
-          alignItems: 'center',
-          justifyContent: 'center',
-          marginTop: globalMargins.small,
-        }}
-      >
+      <Box style={styles.container}>
```
chrisnojima: likely better to use box2 with gapStart/End = true and fullWidth

adamjspooner:

shared/teams/join-team/index.native.js
```diff
@@ -1,52 +1,80 @@
...
+  info: {
+    maxWidth: 280,
+    textAlign: 'center',
+  },
+  input: {
+    borderRadius: 4,
+    borderStyle: 'solid',
+    borderWidth: 1,
+    marginTop: globalMargins.small,
+    paddingBottom: globalMargins.small,
```
chrisnojima: if they're all the same you can just use padding

adamjspooner: I thought it should, but the top and bottom paddings are ignored.

chrisnojima: it shoudn't maybe something else is happening. what does react-devtools say. maybe you need flexShrink: 0

adamjspooner: This is why: https://github.com/keybase/client/blob/master/shared/common-adapters/input.native.js#L255-L256

chrisnojima: you should use PlainInput if you're going to change the styling so much

adamjspooner: Cool beans. Will do.

adamjspooner:

PR #13000, opened 2018-07-26
do a two stage process for UI nonblock attachment send CORE-8383

mmaxim:
It is possible to background the mobile app before the attachment send has landed in the outbox, but the preview is on the screen. Patch does the following:

1.) Break up `PostFileAttachmentLocalNonblock` into two RPCs, `PostFileAttachmentMessageLocalNonblock` and `PostFileAttachmentUploadLocalNonblock`. The two latter RPCs first post the message into the outbox (it won't send since the upload hasn't even been started), and the second one begins the upload.
2.) JS changed to first put the message into the outbox, then put the preview on the screen, and finally to begin the upload process. This should make it so that when backgrounding, we always attempt to keep uploading.
3.) Pass in caller preview again from JS.


patrickxb:
LGTM

PR #13011, opened 2018-07-26
lookup linked TLFID for upgrade convs in untrusted lookup

mmaxim:


mmaxim:
Going to merge this to fix CI, let me know if you don't like something.

PR #13014, opened 2018-07-26
Reacjis: move tooltip to reactions row and refine positioning

buoyad:
- Move `ReactionTooltip` to `ReactionsRow` to get rid of timeout before hiding it
- Make tooltip either `top center` or `bottom center` with an 8px margin around it
- Make `NewReactionButton` only appear when hovering over `ReactionsRow`

r? @keybase/react-hackers cc @cecileboucheron


buoyad:
@keybase/react-hackers bump

shared/chat/conversation/messages/reactions-row/index.js
```diff
@@ -1,36 +1,81 @@
...
+            key={emoji}
+          >
+            <ReactButton
+              ref={r => (this._attachmentRefs[emoji] = r)}
+              conversationIDKey={this.props.conversationIDKey}
+              emoji={emoji}
+              ordinal={this.props.ordinal}
+              style={styles.button}
+            />
+            {/* $FlowIssue */}
```
chrisnojima: whats the flow issue

shared/chat/conversation/messages/reactions-row/index.js
```diff
@@ -1,36 +1,81 @@
...
+            <ReactionTooltip
+              attachmentRef={this._attachmentRefs[emoji]}
+              conversationIDKey={this.props.conversationIDKey}
+              emoji={emoji}
+              onHidden={() => {}}
```
chrisnojima: we should just allow this to be undefined

buoyad: We need this prop for mobile, sorry should've `TODO`d here.

buoyad:
@chrisnojima thanks for looking! That flow issue was the one I mentioned at the end of the day yesterday w/ `ST is not a react component`. It was getting confused by some optional props, which was fixed by typing `OwnProps` in `ReactionTooltip`s `mergeProps`. PTAL

PR #13015, opened 2018-07-26
ActiveDevice now holds a UserVersion, not a UID

maxtaco:
- plumb the difference up and down


maxtaco:
(still some more WIP to actually use this new feature, but let's get CI running now)

maxtaco:
@mlsteele ok, ready for review!

PR #13016, opened 2018-07-26
Cancel outbox messages of dup reactions in Prepare

joshblum:
-moves `getSupersederEphemeralMetadata` from the server call to `BlockingSender.Prepare` (removes duplication, nonblock send will be marginally faster)
-moves `processReactionMessage` as well, and now we scan the outbox and cancel any duplicate pending messages.

### DIFF
--- a/go/chat/errors.go
+++ b/go/chat/errors.go
@@ -441,6 +441,19 @@ func (e SenderTestImmediateFailError) IsImmediateFail() (chat1.OutboxErrorType, 
 
 //=============================================================================
 
+type DuplicateSendAbortedError struct {
+}
+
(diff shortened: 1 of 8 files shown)


go/chat/errors.go
```diff
@@ -441,6 +441,19 @@ func (e SenderTestImmediateFailError) IsImmediateFail() (chat1.OutboxErrorType,
 
 //=============================================================================
 
+type DuplicateSendAbortedError struct {
```
mmaxim: Let's make this more reaction specific

joshblum: removed

mmaxim:
Sick, hope it works in the wild. Thanks!

PR #13017, opened 2018-07-26
Add reaction ctime to reactionMap

joshblum:
Add `Ctime` field to the reactionMap so we can sort by time in the popup etc

cc @buoyad for frontend changes, want to take over this PR for the new protocol changes?


buoyad:
@keybase/react-hackers I added a few lines to make the typing pass and not break the UI. I'll make another PR with the sorting changes

PR #13022, opened 2018-07-27
Reacjis: sort by timestamp in row and tooltip

buoyad:
Both by order added. r? @keybase/react-hackers cc @cecileboucheron

### DIFF
--- a/shared/chat/conversation/messages/reactions-row/container.js
+++ b/shared/chat/conversation/messages/reactions-row/container.js
@@ -1,10 +1,11 @@
 // @flow
-import {compose, connect, createSelector, setDisplayName, type TypedState} from '../../../../util/container'
+import {compose, connect, setDisplayName, type TypedState} from '../../../../util/container'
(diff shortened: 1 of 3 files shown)


shared/chat/conversation/messages/reactions-row/container.js
```diff
@@ -4,6 +4,19 @@ import * as Constants from '../../../../constants/chat2'
 import * as Types from '../../../../constants/types/chat2'
 import ReactionsRow from '.'
 
+// Get array of emoji names in the order of their earliest reaction
+const getOrderedReactions = (reactions: ?Types.Reactions) => {
+  if (!reactions) {
+    return []
+  }
+  const mins = reactions
```
chrisnojima: 🐑  you can just do (reactions || []) and remove that above if

buoyad: `reactions` is a `Map` so I'd need to add an import to do that

PR #13024, opened 2018-07-27
remove retry in attachment store CORE-8360

mmaxim:
Status quo is we have three things retrying attachment uploads: `Deliverer`, `attachments.S3Store`, and the S3 library we use. This patch drop the retry mechanic in `S3Store`, since I have observed it getting stuck retrying a fatal error returned by the S3 library, which only fixed itself on trying the upload from scratch.

### DIFF
--- a/go/chat/attachments/s3.go
+++ b/go/chat/attachments/s3.go
@@ -8,18 +8,15 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"time"
 
 	"github.com/keybase/client/go/chat/s3"
 	"github.com/keybase/client/go/chat/types"
-	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/chat1"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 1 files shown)


go/chat/attachments/s3.go
```diff
@@ -351,27 +307,6 @@ func (a *S3Store) uploadPart(ctx context.Context, task *UploadTask, b job, previ
 	return nil
 }
 
-// putRetry sends a block to S3, retrying retryAttempts times w/ backoff.
```
patrickxb: I feel like this one was useful during multipart uploads to just get one part uploaded during a transient failure.  Have you tested multipart upload retries without this?

mmaxim: I think if there is a transient failure, the goamz library will pick it up anyway. But even if it doesn't, we will be retrying this upload again from the `Deliverer` retry, which will kick in right away. It's just this way it counts as a "failure" over in the `Deliverer` world.

patrickxb: ok

PR #13030, opened 2018-07-27
better error for timeout

chrisnojima:
@keybase/react-hackers @songgao

### DIFF
--- a/shared/util/errors.js
+++ b/shared/util/errors.js
@@ -118,6 +118,8 @@ export const niceError = (e: RPCError) => {
       return 'Looks like the internet is gone...'
     case RPCTypes.constantsStatusCode.scbadsignupusernamedeleted:
       return 'Looks like this user was deleted, or something'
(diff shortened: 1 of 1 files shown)


shared/util/errors.js
```diff
@@ -118,6 +118,8 @@ export const niceError = (e: RPCError) => {
       return 'Looks like the internet is gone...'
     case RPCTypes.constantsStatusCode.scbadsignupusernamedeleted:
       return 'Looks like this user was deleted, or something'
+    case RPCTypes.constantsStatusCode.scstreameof:
+      return 'Looks like we took too long. Try again, but a little bit quicker maybe'
```
songgao: Does EOF always mean time out though?

Also, 1) in this case user would need to generate another QR code; 2) for me there wasn't a way to go back, and I had to kill the app.

chrisnojima: on error provision has to be canceled and restarted

PR #13033, opened 2018-07-27
Reacjis: mobile integration

buoyad:
- [x] Hook up emoji picker to route tree
  - [x] via `NewReactionButton`
  - [x] via message menu
  - [x] via reaction tooltip
- [x] Open `ReactionTooltip` via long press on `ReactionsRow` (after #13014)
- [x] Misc styling

### DIFF
--- a/shared/chat/conversation/messages/message-popup/attachment/index.js
+++ b/shared/chat/conversation/messages/message-popup/attachment/index.js
@@ -12,7 +12,7 @@ type Props = {
   deviceName: string,
   deviceType: DeviceType,
   deviceRevokedAt: ?number,
-  onAddReaction: () => void,
+  onAddReaction: null | (() => void),
   onDelete: null | (() => void),
   onDownload: null | (() => void),
   onHidden: () => void,
@@ -51,7 +51,7 @@ const AttachmentPopupMenu = (props: Props) => {
       ? [{disabled: props.pending, onClick: props.onShareAttachment, title: 'Share'}]
       : []),
     ...(props.onDownload ? [{disabled: props.pending, onClick: props.onDownload, title: 'Download'}] : []),
-    ...(isMobile ? [{onClick: props.onAddReaction, title: 'Add a reaction'}] : []),
+    ...(props.onAddReaction ? [{onClick: props.onAddReaction, title: 'Add a reaction'}] : []),
   ]
 
   const header = {
--- a/shared/chat/conversation/messages/message-popup/attachment/index.js
+++ b/shared/chat/conversation/messages/message-popup/attachment/index.js
@@ -12,6 +12,7 @@ type Props = {
   deviceName: string,
   deviceType: DeviceType,
   deviceRevokedAt: ?number,
+  onAddReaction: () => void,
   onDelete: null | (() => void),
   onDownload: null | (() => void),
   onHidden: () => void,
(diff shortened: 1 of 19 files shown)


buoyad:
@keybase/react-hackers I still have some styling to do but this is ready for a look

buoyad:
@chrisnojima I pushed a bunch of changes to `EmojiPicker` to make it faster. PTAL

buoyad:
@chrisnojima I changed `EmojiPicker` to make it faster (and fixed that bug I told you about). PTAL

shared/chat/conversation/messages/message-popup/attachment/index.js
```diff
@@ -50,6 +51,7 @@ const AttachmentPopupMenu = (props: Props) => {
       ? [{disabled: props.pending, onClick: props.onShareAttachment, title: 'Share'}]
       : []),
     ...(props.onDownload ? [{disabled: props.pending, onClick: props.onDownload, title: 'Download'}] : []),
+    ...(isMobile ? [{onClick: props.onAddReaction, title: 'Add a reaction'}] : []),
```
chrisnojima: really should be controlled by the connector. then have the check be props.onAddReaction (like onDownload above)

shared/chat/conversation/messages/message-popup/exploding/container.js
```diff
@@ -97,6 +108,13 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
       title: 'Explode now',
     })
   }
+  if (isMobile) {
```
chrisnojima: same here, just check dispatchProps._onAddReaection and have the isMobile check there. I think i prefer these menuitem things to be made in the component and not in the connector but we can leave this for now

shared/chat/conversation/messages/message-popup/text/index.js
```diff
@@ -51,6 +52,7 @@ const TextPopupMenu = (props: Props) => {
           },
         ]
       : []),
+    ...(isMobile ? [{onClick: props.onAddReaction, title: 'Add a reaction'}] : []),
```
chrisnojima: same here

shared/chat/conversation/messages/react-button/emoji-picker/index.js
```diff
@@ -27,78 +48,116 @@ type State = {
...
+      // Nothing to do if we don't have a width
+      return
+    }
+    if (this.props.width === cachedWidth) {
+      this.setState(s => (s.sections === cachedSections ? null : {sections: cachedSections}))
+      return
     }
+    // width is different from cached. make new sections & cache for next time
+    let sections = []
+    const emojisPerLine = Math.floor(this.props.width / emojiWidthWithPadding)
```
chrisnojima: lets talk about this irl

shared/chat/conversation/messages/message-popup/exploding/container.js
buoyad: the attachment popup does the check similarly in `mergeProps`, so it seems kinda arbitrary to move it up a level here. I'm changing the other ones over though.

buoyad: I also see why making the menuitems in the component would be nice for storybook and such. on the other hand, I went with the container here because using undefined props to switch the view is pretty difficult to read IMO. 🤔

buoyad:
Thanks @chrisnojima, addressed your feedback!

PR #13034, opened 2018-07-27
[FloatingMenu Refactor, pt. 2]: Internal refactor of FloatingMenu

nathunsmitty:
Part 2 of the FloatingMenu refactor. This one's a little bigger than part 1 and is basically FloatingMenu taking the good PopupMenu code and putting that on top of Overlay.


shared/chat/conversation/attachment-fullscreen/index.desktop.js
```diff
@@ -9,7 +9,10 @@ import {
   ProgressBar,
   ProgressIndicator,
 } from '../../../common-adapters'
-import {FloatingMenuParentHOC, type FloatingMenuParentProps} from '../../../common-adapters/floating-menu'
+import {
+  FloatingMenuParentHOC,
+  type FloatingMenuParentProps,
+} from '../../../common-adapters/floating-menu/parent-hoc'
```
nathunsmitty: With some many files in common-adapters, it seemed like a good idea to start putting components into folders. I also split up the floating menu component and the parent HOC into two separate files for easier readability.

shared/chat/conversation/messages/message-popup/attachment/index.js
```diff
@@ -74,15 +74,10 @@ const AttachmentPopupMenu = (props: Props) => {
       onHidden={props.onHidden}
       closeOnSelect={true}
       position={props.position}
-      style={{...stylePopup, ...props.style}}
+      containerStyle={props.style}
```
nathunsmitty: I made the props of FloatingMenu an exact type, and caught some errors where a `style` prop was being passed in to `FloatingMenu` instead of a `containerStyle` prop. Since these props weren't doing anything and the display of all the popups seemed correct, the best course of action seemed to be to delete all the styles and change the name if any props are passed down.

shared/common-adapters/floating-menu/index.stories.js
```diff
@@ -0,0 +1,55 @@
...
+  {...commonItemProps, title: 'Four'},
+]
+
+const commonProps = {
+  onHidden: action('onHidden'),
+  visible: true,
+  items: commonItems,
+}
+
+const load = () => {
```
nathunsmitty: Added a story for FloatingMenu! It's based mostly off of the PopupMenu story.

shared/chat/conversation/attachment-fullscreen/index.desktop.js
nathunsmitty: Probably, considering we export other HOCs from `common-adapters/index.js`. Not sure why `FloatingMenuParentHOC` never got that same treatment.

PR #13035, opened 2018-07-27
handle 10 bad kex2 phrase attempts

maxtaco:


patrickxb:
what's the change here besides moving to a function?

maxtaco:
that we call `cancel()` after we fail 10x. before, the app just hung indefinitely

patrickxb:
ok, thanks.  my human diff didn't see that.

maxtaco:
sure thing.  yeah, i refactored so it was easier to forget to cancel on error.

PR #13037, opened 2018-07-27
Add ephemeral retention policy type

joshblum:
Protocol changes for adding EphemeralRetentionPolicy this is just for the server PR i'll do another pass in the client to test and add cli support

### DIFF
--- a/protocol/avdl/chat1/common.avdl
+++ b/protocol/avdl/chat1/common.avdl
@@ -491,13 +491,15 @@
     NONE_0,
     RETAIN_1, // Keep messages forever
     EXPIRE_2, // Delete after a while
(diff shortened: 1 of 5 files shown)


protocol/avdl/chat1/common.avdl
```diff
@@ -509,6 +511,11 @@
 
   record RpInherit {}
 
+  record RpEphemeral {
+    // Messages must be exploding at have at most this lifetime.
```
mlsteele: at->and

joshblum: will change in https://github.com/keybase/keybase/pull/2716

PR #13038, opened 2018-07-27
filter chat search by sender username

joshblum:


go/chat/searcher.go
```diff
@@ -29,7 +29,7 @@ func NewSearcher(g *globals.Context) *Searcher {
 }
 
 func (s *Searcher) SearchRegexp(ctx context.Context, uiCh chan chat1.ChatSearchHit, conversationID chat1.ConversationID, re *regexp.Regexp,
-	maxHits, maxMessages, beforeContext, afterContext int) (hits []chat1.ChatSearchHit, err error) {
+	sentBy string, maxHits, maxMessages, beforeContext, afterContext int) (hits []chat1.ChatSearchHit, err error) {
```
mmaxim: Probably going to want a "query" type at some point here as we keep adding things.

joshblum: yup agreed

PR #13043, opened 2018-07-28
Stellar: use new pending payments RPC

cjb:
@keybase/react-hackers (CC @mlsteele @patrickxb) 

Gets pending payments from their own RPC and combines them with the rest of the payments before submitting all together to the reducer as one array.  What do y'all think about this way of combining data from two RPCs into one action?

```js
  RPCTypes.localGetPaymentsLocalRpcPromise({accountID: action.payload.accountID}).then(payments =>
    RPCTypes.localGetPendingPaymentsLocalRpcPromise({accountID: action.payload.accountID}).then(pending =>
      WalletsGen.createPaymentsReceived({
        accountID: action.payload.accountID,
        payments: (payments.payments || [])
          .concat(pending || [])
          .map(elem => Constants.paymentResultToPayment(elem))
          .filter(Boolean),
      })
    )
```

We could just take the loadPayments action into two separate sagas, each of which loads half of the data and then sends that half to a reducer which merges it.  I didn't like that because we end up going through store mutation and re-rendering twice instead of once, but it does seem the more Redux-y way to do it..?


chrisnojima:
I think we want to do what you implemented but it can be done in parallel instead of sequentially. The downside of the 2 totally sep paths that you suggest is dealing with errors and reconciling if we have multiple in flight etc. just harder to reason about. Also causes ux issues with a list showing with no pending or it being half out of date

PR #13050, opened 2018-07-30
[KBFS-3240] Properly fix Windows uploads

jzila:


shared/constants/types/fs.js
```diff
@@ -280,6 +280,7 @@ export const getLocalPathName = (localPath: LocalPath): string => {
   return ''
 }
 export const getLocalPathDir = (p: LocalPath): string => p.slice(0, p.lastIndexOf(localSep))
+export const getNormalizedLocalPath = (p: LocalPath): LocalPath => p.replace(new RegExp(localSep === '\\' ? '\\\\' : localSep, 'g'), '/')
```
songgao: Seems `... => localSep === '\\' ? p.replace(/\\\\/g, '/') : p` would be better?

Also there's a `pathToURL` in `actions/fs/platfrm-specific.desktop.js`, with similar substitutions, and it talks about drive letter being prefixed by a slash. I thought I'd bring it up in case it applies here too.

jzila:
@songgao good to merge?

PR #13059, opened 2018-07-30
Finesse proof posting buttons.

adamjspooner:

It makes the first step (posting to whatever service) the primary button instead of a link, and then it shows the “OK posted! Check for it!” after you've pressed the button for the first action.

### DIFF
--- a/shared/profile/post-proof/container.js
+++ b/shared/profile/post-proof/container.js
@@ -43,7 +43,7 @@ export default compose(
   connect(mapStateToProps, mapDispatchToProps),
   lifecycle({
     componentDidMount() {
-      // Active the proof check after they've completed the first step for these services.
+      // Activate the proof check after they've completed the first step for these services.
       if (['facebook', 'twitter', 'reddit', 'github', 'hackernews'].includes(this.props.platform)) {
         this.props.onAllowProofCheck(false)
       }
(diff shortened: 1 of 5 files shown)


shared/profile/post-proof/container.js
```diff
@@ -43,8 +43,8 @@ export default compose(
   connect(mapStateToProps, mapDispatchToProps),
   lifecycle({
     componentDidMount() {
-      // Facebook proof checking gets enabled after they click continue.
-      if (this.props.platform === 'facebook') {
+      // Active the proof check after they've completed the first step for these services.
+      if (['facebook', 'twitter', 'reddit', 'github', 'hackernews'].includes(this.props.platform)) {
```
adamjspooner: This could get out of hand. I'm open to suggestions.

adamjspooner:
@cecileboucheron, how does this look?
screen shot 2018-07-31 at 09 40 52

adamjspooner:
@keybase/react-hackers and @cecileboucheron, bump.

cecileboucheron:
I'd just kill this verbose and do:
`Post your proof to Facebook, and make sure the post is _public_.`
WDYT?

adamjspooner:
I think a brief explanation on why might alleviate any concerns. What about:

```
Post your proof to Facebook. Make sure it's _public_, or we won't be able to read it.
```

adamjspooner:
Are we dropping the GIF?

cecileboucheron:
Sure. No I would keep the GIF but some time we should make a better one...

PR #13067, opened 2018-07-31
support video attachments better CORE-7902

mmaxim:
The status quo for video attachments is pretty poor. On desktop they can be sent, but they look just like another other file. If those sent files are attempted to be viewed on mobile, nothing can be done. They are literally un-viewable on Android, and are basically un-viewable on iOS (you can sort of view them by sharing them to yourself, but I find it doesn't always work). On mobile, video can just not be sent on either platform. 

This patch adds a new approach which brings video support for both platforms in a much better way. On desktop, when sending a video a proper preview still is now generated, and inserted into the thread. Clicking this previews downloads the video and displays "Show in Finder". On mobile, iOS can now upload videos and generate preview stills. Both mobile platforms allow tapping a video preview to download it. A subsequent tap will display the video in fullscreen mode on iOS. On both platforms the user can save the video to their camera roll. On iOS, share continues to work. I think the experience is much improved. cc @malgorithms  for the user experience. I am happy to jump on a call to discuss it. 

Code-wise patch does the following:
**CORE**
1.) Add a new function `previewVideo`, which is implemented on iOS and macOS. On other platforms it produces a black rectangle. It works by grabbing a still from the video, JPEG encoding it, and then running it through the image preview code to get the proper size.
2.) Set the preview MIME type properly.
3.) Workaround the Go HTTP package MIME detector not detecting QuickTime video properly.
4.) When serving video from `AttachmentHTTPSrv`, use `http.ServeContent` instead of writing directly to the response writer. Serving video involves handling `Range` headers on the request, and this doesn't work if you just write the whole thing out. We use a temp file to make this happen, and delete it after the content is served.
5.) Add `IsAssetLocal` to `AttachmentFetcher` in order to communicate to the frontend that we have the video stored locally, and it can be viewed in the fullscreen viewer. Add a field to the UI type to fill in from this information. 

**Frontend**
1.) Change the `ImageAttachment` row to fire off a new action, `attachmentPreviewSelect` instead of just navigating to fullscreen view. 
2.) When handling this new action, decide the best course of action for handling the selection. If we selected an image, we just open the normal fullscreen viewer. If we selected a video, then we check to see if it has been downloaded. If so, we pop open the video fullscreen view on iOS. 
3.) Add a new component `AttachmentVideoFullscreen` which is used to just load a `WebView` component with the URL of the video. This is only used on native (and only on iOS for now).
4.) Control the `Save` and `Share` menu options for video based on whether or not the video is downloaded. 
5.) Change iOS `Save` to use the temp file method for video, since it doesn't work to send it a URI that is not `file://` for video. 

There is a lot going on here, so let me know if anyone wants to jump on a call to discuss.


chrisnojima:
i think you might be able to get this to work with a uri in the webview. We might want to just use a 3rd party library to get it to work on ios and android w/ uri and remove all this temp file code (https://github.com/react-native-community/react-native-video). might be worth playing with in a branch.
anyways, exciting!

mmaxim:
Which temp file code in particular? The stuff in Go? The complication here is the stuff we bring down from S3 is encrypted, so we need to run it through all the decryption stuff before sending it up to the `WebView`.

mmaxim:
That is, we do give the `WebView` a URI, it's just a URI on the service local HTTP server.

chrisnojima:
i'm fine with half steps to get it better, but ideally we could just hit the http server and it'd stream / decypt on the fly and not have to download the whole thing before doing anything

mmaxim:
It would be a great lift for us to do that I think (I'd have to refactor a lot of code). If you are ok with it, I think trying this first works. The UI feels pretty good too, instead of a choppy buffering viewing experience tapping directly in, you get a nice progress bar telling you we are grabbing the video, and then tapping it it plays smoothly.

mmaxim:
I experimented with a degraded mode where the user just couldn't "seek" in the video (drag the time control forward), but couldn't get it to work. I settled on the method I described here because I didn't have to change too much to get it to work, and it feel pretty good too in the app.

mmaxim:
Telegram has a similar experience (where you download then view), but they have kind of a nice UI where you see like a download icon instead of the play button on the video preview. We could also do something like that instead of using the green icon in the lower right.

go/chat/attachments/preprocess.go
```diff
@@ -201,7 +201,12 @@ func PreprocessAsset(ctx context.Context, log utils.DebugLabeler, filename strin
 	}
 	// MIME type detection failed us, try using an extension map
 	if p.ContentType == "application/octet-stream" {
-		if typ, ok := mimeTypes[strings.ToLower(filename)]; ok {
+		var ext string
```
patrickxb: https://golang.org/pkg/path/#Ext should do this...

mmaxim: Ah good call of course, will change!

go/chat/attachments/preprocess.go
```diff
@@ -189,17 +196,30 @@ func PreprocessAsset(ctx context.Context, log utils.DebugLabeler, filename strin
 	p = Preprocess{
 		ContentType: http.DetectContentType(head),
 	}
-	log.Debug(ctx, "preprocessAsset: detected attachment content type %s", p.ContentType)
 	if _, err := src.Seek(0, 0); err != nil {
 		return p, err
 	}
+	// MIME type detection failed us, try using an extension map
+	if p.ContentType == "application/octet-stream" {
+		var ext string
```
patrickxb: i think my comment got lost, but `path.Ext` does this.

https://golang.org/pkg/path/#Ext

shared/chat/conversation/attachment-video-fullscreen/index.desktop.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+import React from 'react'
+import {WebView} from '../../../common-adapters'
+import {globalColors} from '../../../styles'
+import type {Props} from './index.types'
+
+// Unused for now, we don't view videos in the app on desktop
```
chrisnojima: ? is this true? why return webview?

shared/chat/conversation/attachment-video-fullscreen/index.native.js
```diff
@@ -0,0 +1,70 @@
+// @flow
+import React from 'react'
+import {WebView, Box, ClickableBox, Text} from '../../../common-adapters'
+import {globalColors, platformStyles, globalMargins, globalStyles} from '../../../styles'
+import {FloatingMenuParentHOC, type FloatingMenuParentProps} from '../../../common-adapters/floating-menu'
+import type {Props} from './index.types'
+
+class _VideoFullscreen extends React.Component<Props & FloatingMenuParentProps, {loaded: boolean}> {
+  state = {loaded: false}
+  _setLoaded = () => this.setState({loaded: true})
```
chrisnojima: why bother with this state?

shared/chat/conversation/attachment-video-fullscreen/index.desktop.js
mmaxim: Desktop it just downloads it, doesn't do the view in here. Is there something smaller I can return from here? I just wanted to satisfy flow.

shared/chat/conversation/attachment-video-fullscreen/index.native.js
mmaxim: I just left it in there by mistake, I'll remove it.

mmaxim:
ok @chrisnojima this is ready for another look, thanks!

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -31,8 +31,10 @@ type Props = {
   progress: number,
   progressLabel: string,
   showPlayButton: boolean,
+  showFilmButton: boolean,
```
chrisnojima: should likely be something
`showButton: null | 'play' | 'film'`
since its really a state and we'd never show both at the same time

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -82,6 +84,17 @@ class ImageAttachment extends React.PureComponent<Props, State> {
           {this.props.showPlayButton && (
             <Icon type="icon-play-64" style={iconCastPlatformStyles(styles.playButton)} />
           )}
+          {this.props.showFilmButton && (
+            <Icon type="icon-film-64" style={iconCastPlatformStyles(styles.playButton)} />
+          )}
+          {this.props.videoDuration &&
```
chrisnojima: this is dangerous in react native as android will crash if this is a string and it gets passed down into the render. Instead do
```
{!!this.props.videoDuration &&
```

eventually we'll get flow to catch this but until then we just have to be careful

shared/constants/types/chat2/message.js
```diff
@@ -139,14 +139,15 @@ export type _MessageAttachment = {
...
+  previewTransferState: 'downloading' | null, // only for preview
   reactions: Reactions,
   submitState: null | 'deleting' | 'pending' | 'failed',
   timestamp: number,
   title: string,
   transferProgress: number, // 0-1 // only for the file
   transferState: 'uploading' | 'downloading' | 'remoteUploading' | null,
-  previewTransferState: 'downloading' | null, // only for preview
   type: 'attachment',
+  videoDuration: ?string,
```
chrisnojima: lets just type this as string and have it be '' if there's no duration, keeps the typing and checking simpler downstream

PR #13068, opened 2018-07-31
Stellar: lookup federation addresses

zapu:
TODO:
- [x] remove stellartoml/mocks.go in stellar-org fork and revendor


go/client/cmd_wallet_lookup.go
```diff
@@ -0,0 +1,72 @@
...
+
+	dui := c.G().UI.GetDumbOutputUI()
+	res, err := cli.LookupCLILocal(context.Background(), c.Name)
+	if err != nil {
+		return err
+	}
+
+	dui.Printf("%q: %s\n", c.Name, res.AccountID)
+	if res.Username != nil {
+		dui.Printf("Belonging to https://keybase.io/%s\n", *res.Username)
```
patrickxb: I'd leave off the URL here.  If they are in CLI they probably want to stay in CLI and can do `keybase id <username>` if they feel like it...

You could use some colors if you want :)

PR #13074, opened 2018-07-31
Payment protocol changes for transfers

mlsteele:


shared/constants/wallets.js
```diff
@@ -110,13 +110,13 @@ const paymentResultToPayment = (w: RPCTypes.PaymentOrErrorLocal) => {
     id: p.id,
     note: p.note,
     noteErr: p.noteErr,
-    source: p.source,
-    sourceType: partyTypeToString[p.sourceType],
+    source: p.fromUsername || p.fromAccountID,
```
cjb: Should this line be:
```
     source: p.fromUsername || p.fromAccountID || p.fromAccountName,
```
?

mlsteele: fromAccountID is always set. But it shouldn't be much like that, this was just enough to be a good starting point with valid flow.

mlsteele:
As is this PR passes flow but makes the app a bit worse. If a user (tst1) sends from a secondary account to the primary account this branch shows "tst1...tst1 sent you 1 XLM" with the scuba icon. On master it shows "You transferred 1 XLM from tst1 to this account" with the wallet icon. Neither is right, but master is better. The right thing I think is "You transferred 1 XLM from "squirrel fund" to GC...RAT" if the first has a name. So I think it would make sense to get the gui half of this done before merging.

The motivation for this PR is (@buoyad) to move the code that detects when a payment is a transfer between the user's own account from gui to service. The gui can now use [OWNACCOUNT](https://github.com/keybase/client/pull/13074/files#diff-d7ca6ef6559e7c1f0d0349bc72a32f8bR77) to decide that it's a own-transfer. So I imagine finishing this PR swap out that code.

A secondary effect is that more info is available in the rpc if desired. It used to be that if the recipient had a username and an account ID the RPC did not include the account ID because it was all packed into two fields [target, targetType]. For example on the details page it would be nice to show the recipient account ID as well as their username when applicable. But that can happen another time. Up to you whether it's worth future proofing by not throwing out fields.

mlsteele:
Here are my thoughts on the logic in case it helps.

If `fromType == OWNACCOUNT` then the icon should be wallet.
If the counterparty is of type `KEYBASE` then the icon should be their profile picture.
Otherwise the icon should be scuba.

For the text of sender and recipient, there's a mapping. I know it's really more complicated than this since there's different sentence structures.
```
switch toType
KEYBASE -> toUsername
STELLAR -> toAccountID
SBS -> toAssertion
OWNACCOUNT -> toAccountName
```

cjb:
@mlsteele Pushed the JS changes, let me know what you think.  I recorded fromAccountID and toAccountID in the store so that we can decide whether to use them on the transaction details page, as you suggested.

More conflicts showed up immediately!  Mainly it's PR #13789's changes to frontend_test.go (`recipPayments`), which'll need to be modified to test against the new fields instead of source/target.  Please could I ask you to do that merge again?

mlsteele:
Looks nice! Conflicts fixed

go/stellar/loader.go
```diff
@@ -255,18 +256,17 @@ func (p *Loader) uiPaymentInfo(m libkb.MetaContext, summary *stellar1.PaymentLoc
...
-		// check if sending to self
-		if summary.TargetType == stellar1.ParticipantType_KEYBASE {
-			if libkb.NewNormalizedUsername(summary.Target).Eq(username) {
-				info.Delta = stellar1.BalanceDelta_NONE
-			}
-		}
+	info.Delta = stellar1.BalanceDelta_NONE
+
+	// Calculate the payment delta
+	if summary.FromType == stellar1.ParticipantType_OWNACCOUNT {
```
patrickxb: do you also want to check if `summary.ToType == OWNACCOUNT`?

mlsteele: "Either neither or both are OWNACCOUNT". But sure might as well. Faith in comment invariants and all that

PR #13077, opened 2018-07-31
fix ConnectionManager#WaitForClientType

maxtaco:
- don't keep bumping the deadline off into the future, we'll never get there!


maxtaco:
cc @chrisnojima

chrisnojima:
works tnx

PR #13085, opened 2018-08-01
Purge resolver cache on subteam rename

joshblum:
When renaming a subteam we purge the resolver cache for the subteam id. Previously, when we would try to resolve a name, we would get an error because the cached name does not match the name in the team chain.

### DIFF
--- a/go/systests/teams_test.go
+++ b/go/systests/teams_test.go
@@ -581,6 +581,23 @@ func (u *userPlusDevice) waitForTeamChangedAndRotated(teamID keybase1.TeamID, to
 	u.tc.T.Fatalf("timed out waiting for team rotate %s", teamID)
 }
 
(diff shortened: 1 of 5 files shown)


go/systests/teams_test.go
```diff
@@ -581,6 +581,23 @@ func (u *userPlusDevice) waitForTeamChangedAndRotated(teamID keybase1.TeamID, to
 	u.tc.T.Fatalf("timed out waiting for team rotate %s", teamID)
 }
 
+func (u *userPlusDevice) waitForTeamChangeRenamed(teamID keybase1.TeamID) {
+	// process 10 team rotations or 10s worth of time
```
zapu: I found that this method is not greatest especially when there are pending changes that will be immediately selected. Here's what I propose we should start doing for these loops: https://github.com/keybase/client/blob/master/go/systests/teams_test.go#L512-L514

joshblum: good with me!

PR #13088, opened 2018-08-01
Rename "wallet" strings to "account"

thebearjew:


thebearjew:
Just want to confirm that the ticket was to change display text rather than variable names (it's been a while) @chrisnojima

PR #13091, opened 2018-08-01
Add a width to the FloatingMenu on the Profile page

nathunsmitty:
This resolves a visual error that @cecileboucheron noticed. She also noted a crash, though, which I haven't been able to reproduce yet.


shared/profile/index.desktop.js
```diff
@@ -625,11 +632,18 @@ const styleShowcasedTeamAvatar = {
...
-  ...globalStyles.flexBoxRow,
+  ...Styles.globalStyles.flexBoxRow,
   alignItems: 'center',
   justifyContent: 'center',
   alignSelf: 'center',
-  paddingLeft: globalMargins.tiny,
+  paddingLeft: Styles.globalMargins.tiny,
 }
 
+const styles = Styles.styleSheetCreate({
```
nathunsmitty: I didn't want to get too into an entire style refactor of the Profile screen so it seemed like the best action was to just create a new `styles` object for the `FloatingMenu` style.

PR #13095, opened 2018-08-01
Team load optimization: don't copy team link maps when adding to proofSet

mlsteele:
Team link maps were being copied all over the place but in the end it would [just use](https://github.com/keybase/client/blob/master/go/teams/proofs.go#L244) the one from the [final chain](https://github.com/keybase/client/blob/master/go/teams/loader2.go#L594).

This was maybe 5% of cpu profile. For a load with warm upak but cold team.

### DIFF
--- a/go/protocol/keybase1/upk.go
+++ b/go/protocol/keybase1/upk.go
@@ -5,7 +5,6 @@ package keybase1
 
 import (
 	"errors"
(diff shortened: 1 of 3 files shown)


mlsteele:
circle is about simplefs which this doesn't touch

PR #13099, opened 2018-08-01
Store chainlink revocations

mlsteele:
This saves 5% of cpu time spent during a cold load of keybasefriends.

Is this safe? I'm not sure what the memory rules are around this code.

### DIFF
--- a/go/libkb/chain_link.go
+++ b/go/libkb/chain_link.go
@@ -4,6 +4,7 @@
 package libkb
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


mlsteele:
GetRevocations is not on the map anymore. So it can't shave another 9?

mlsteele:
Bah it should've only halved. This doesn't make sense.

mlsteele:
The expected result of this change would be for GetRevocations to halve in its cpu contribution. The svg output of pprof made it seem like it went from 10% to negligible. Hence my confusion. But `pprof -top` tells the truth that it went from 10% to 5%. I wonder why the svg did that.

mlsteele:
Anyway now it's 4%

maxtaco:
So that means it helped?

mlsteele:
Yep!

maxtaco:
cool!

go/libkb/chain_link.go
```diff
@@ -198,16 +199,17 @@ var badWhitespaceChainLinks = map[keybase1.SigID]string{
...
+	id               LinkID
+	hashVerified     bool
+	sigVerified      bool
+	payloadVerified  bool
+	chainVerified    bool
+	storedLocally    bool
+	revoked          bool
+	unsigned         bool
+	dirty            bool
+	revocationsCache *[]keybase1.SigID
```
maxtaco: could you just get away with `revocationsCache []keybase1.SigID` and then set it to be `[]keybase1.SigID{}` if caching an empty value?  might be slightly simpler.  I don't feel strongly.

PR #13101, opened 2018-08-01
decouple TLF metadata out of PathItems 1: store and saga

songgao:


shared/actions/fs/index.js
```diff
@@ -15,22 +15,17 @@ import {isMobile} from '../../constants/platform'
...
-    const username = state.config.username || ''
-    const loggedIn = state.config.loggedIn
-    const folders = Constants.folderToFavoriteItems(results && results.body, username, loggedIn)
-
-    yield Saga.put(FsGen.createFavoritesLoaded({folders}))
-  } catch (e) {
-    logger.warn('Error listing favorites:', e)
-  }
-}
+const loadFavorites = (state: TypedState) =>
```
chrisnojima: you likely want your error path also, else it'll black bar

shared/constants/types/fs.js
```diff
@@ -28,41 +28,43 @@ export type ResetMember = {
...
   resetParticipants: Array<ResetMember>,
   teamId: RPCTypes.TeamID,
-|}
-
-export type _FavoriteItem = {
-  badgeCount: number,
-  name: string,
-  tlfMeta?: FavoriteMetadata,
-  favoriteChildren?: I.Set<string>,
+  waitingForParticipantUnlock?: Array<ParticipantUnlock>,
```
chrisnojima: we should only use immutable our data, prefer I.List

shared/reducers/fs.js
```diff
@@ -172,13 +150,15 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
     case FsGen.favoriteIgnoreError:
-      return state.mergeIn(['pathItems', Types.pathToString(action.payload.path), 'tlfMeta'], {
-        isIgnored: false,
+      const elems = Types.getPathElements(action.payload.path)
+      const visibility = Types.getVisibilityFromElems(elems)
+      if (!visibility) {
+        return state
+      }
+      return state.mergeIn(['tlfs', visibility, elems[2]], {
+        isIgnored: action.type === FsGen.favoriteIgnore,
```
jzila: I don't think this is right--it could be that one `favoriteIgnore` succeeded and then another one failed, in which case the failing one would reset the value of `isIgnored`.

songgao: Hmm I think this is equivalent to the version before. I don't think it really matters much though. `isIgnore` is just for update the UI state right? On the next loadFavorites call, this won't be part of it if ignore succeeded.

jzila: Yeah true.

shared/constants/fs.js
```diff
@@ -394,90 +406,74 @@ const _fillMetadataInFavoritesResult = (
...
+        resetParticipants: reset_members || [],
+        teamId: team_id || '',
+        waitingForParticipantUnlock,
+        youCanUnlock,
       })
-    )
+      return {
+        private:
+          folderType === RPCTypes.favoriteFolderType.private
+            ? {...tlfs.private, [tlf.name]: tlf}
```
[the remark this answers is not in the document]
songgao: OK will do. FWIW I was trying to avoid that since it felt weird to mutate the reduction object in `reduce()` and return the same object.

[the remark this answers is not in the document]
jzila: That's a pretty standard pattern: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#Grouping_objects_by_a_property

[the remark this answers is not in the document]
songgao: Hmm OK! Thanks for the link.

shared/actions/fs/index.js
chrisnojima: You don't  need the return type. Better to just let flow infer

songgao: I’m OK with either way here, but I find having a return type on the function signature helpful for reading code sometimes especially on long or complicated functions. Is there a reason not to have them other than not having to figure the exact type and type then put?

chrisnojima: I think overly typing functions that are used within a module (where flows inference is strongest) usually ends up making things overly verbose. One of flows strengths is its inference capabilities. Sometimes the inferred type can be very complex and would actually be hard to type correctly. Obviously not here. If you feel like it aids in readability that's fine but adding the extra types can often just make the code more noisy

jzila: @chrisnojima I'm pretty perplexed as to when we're supposed to type things explicitly. Earlier the guidance was that Flow sometimes defaults to `any`, which defeats a lot of the type checking we want to rely on. So I errred on the side of specifying type whenever possible. But it seems that now the guidance is that we should rely on inference when we can.

Is there a check we should be running to test this?

chrisnojima: When it's externally facing you have to type it. When it's just in the module, flow can totally infer the input and output so yo don't need typing. You can test this out. Note things like star work different inside vs out. (aka not at all outside a module). When in doubt flow coverage and introspect the type yourself with type-at-pos

songgao:

PR #13102, opened 2018-08-01
decouple TLF metadata out of PathItems 2: component changes - rows

songgao:


shared/fs/row/still.js
```diff
@@ -1,152 +1,66 @@
 // @flow
 import * as React from 'react'
 import * as Types from '../../constants/types/fs'
-import {globalColors, globalMargins, isMobile, glamorous} from '../../styles'
-import rowStyles from './styles'
-import {Badge, Box, Box2, ClickableBox, Icon, Meta, Text} from '../../common-adapters'
-import PathItemIcon from '../common/path-item-icon'
+import {globalColors, globalMargins, isMobile} from '../../styles'
```
chrisnojima: Moving towards using import * as Styles and import * as Kb

shared/fs/row/styles.js
```diff
@@ -1,6 +1,20 @@
 // @flow
-import {globalStyles, globalMargins, platformStyles} from '../../styles'
-import {memoize} from 'lodash-es'
+import {
```
chrisnojima: same here

shared/fs/row/styles.js
```diff
@@ -67,13 +73,105 @@ const rightBox = {
   alignItems: 'center',
 }
 
-export default {
-  rowBox,
-  itemBox,
-  pathItemIcon,
-  pathItemIcon_30,
+const pathItemActionIcon = {
```
chrisnojima: should use Styles.styleSheetCreate

shared/fs/row/tlf-container.js
```diff
@@ -0,0 +1,58 @@
...
+import * as StateMappers from '../utils/state-mappers'
+
+type OwnProps = $Diff<Types.TlfRowItem, {rowType: 'tlf'}> & {
+  routePath: I.List<string>,
+}
+
+const mapStateToProps = (state: TypedState, {tlfType, name}: OwnProps) => {
+  const _tlf = Constants.getTlfFromTlfs(state.fs.tlfs, tlfType, name)
+  const _kbfsEnabled = StateMappers.mapStateToKBFSEnabled(state)
+  const _username = state.config.username || undefined
```
chrisnojima: why || undefined

shared/fs/row/tlf-type-container.js
```diff
@@ -0,0 +1,50 @@
...
+import {compose, connect, setDisplayName, type TypedState, type Dispatch} from '../../util/container'
+import TlfType from './tlf-type'
+import * as StateMappers from '../utils/state-mappers'
+
+type OwnProps = $Diff<Types.TlfTypeRowItem, {rowType: 'tlf-type'}> & {
+  routePath: I.List<string>,
+}
+
+const mapStateToProps = (state: TypedState, {name}: OwnProps) => {
+  const _tlfList = Constants.getTlfListFromType(state.fs.tlfs, name)
```
chrisnojima: might as well just do _tlfList: Constants.getTlf... so its all inline and remove the braces/return

shared/fs/row/tlf-type.js
```diff
@@ -0,0 +1,48 @@
...
+  badgeCount: number,
+}
+
+const RowMeta = ({badgeCount}) => {
+  if (!badgeCount) {
+    return null
+  }
+
+  return (
+    <Box style={{width: 0, display: 'flex'}}>
```
chrisnojima: what is this box for

shared/fs/row/tlf.js
```diff
@@ -0,0 +1,63 @@
...
+  resetParticipants: Array<string>,
+}
+
+const RowMeta = ({isNew, isIgnored, needsRekey}) => {
+  if (isIgnored || !(isNew || needsRekey)) {
+    return null
+  }
+
+  return (
+    <Box style={{width: 0, display: 'flex'}}>
```
chrisnojima: what is this box for

chrisnojima: also this likely doens't work on android

shared/fs/row/tlf.js
```diff
@@ -0,0 +1,63 @@
...
+    path={props.path}
+    onOpen={props.onOpen}
+    openInFileUI={props.openInFileUI}
+  >
+    <RowMeta isIgnored={props.isIgnored} isNew={props.isNew} needsRekey={props.needsRekey} />
+    <Box style={rowStyles.itemBox}>
+      <Box2 direction="horizontal" fullWidth={true}>
+        <Text
+          type={props.itemStyles.textType}
+          style={{...rowStyles.rowText, color: props.itemStyles.textColor}}
```
chrisnojima: use collapseStyles

shared/fs/row/tlf-type.js
```diff
@@ -0,0 +1,48 @@
...
+    path={props.path}
+    onOpen={props.onOpen}
+    openInFileUI={props.openInFileUI}
+  >
+    <RowMeta badgeCount={props.badgeCount} />
+    <Box style={rowStyles.itemBox}>
+      <Box2 direction="horizontal" fullWidth={true}>
+        <Text
+          type={props.itemStyles.textType}
+          style={{...rowStyles.rowText, color: props.itemStyles.textColor}}
```
chrisnojima: collapseStyles

shared/fs/row/still.js
```diff
@@ -1,152 +1,66 @@
...
+    path={props.path}
+    onOpen={props.onOpen}
+    openInFileUI={props.openInFileUI}
+  >
+    <RowMeta isDownloading={props.isDownloading} />
+    <Box style={rowStyles.itemBox}>
+      <Box2 direction="horizontal" fullWidth={true}>
+        <Text
+          type={props.itemStyles.textType}
+          style={{...rowStyles.rowText, color: props.itemStyles.textColor}}
```
chrisnojima: collapseStyles

shared/fs/row/tlf.js
```diff
@@ -0,0 +1,63 @@
...
+          lineClamp={isMobile ? 1 : undefined}
+        >
+          {props.name}
+        </Text>
+      </Box2>
+      {!props.resetParticipants || props.resetParticipants.length === 0 ? null : (
```
chrisnojima: shouldnt need this first check, it always exists

shared/fs/row/uploading.js
```diff
@@ -39,7 +39,7 @@ const Uploading = (props: UploadingProps) => (
     <Box key="main" style={rowStyles.itemBox}>
       <Text
         type={props.itemStyles.textType}
-        style={rowStyles.rowText_30(props.itemStyles.textColor)}
+        style={{...rowStyles.rowText_30, color: props.itemStyles.textColor}}
```
chrisnojima: collapseStyles

shared/fs/row/index.js
```diff
@@ -1,7 +1,11 @@
 // @flow
 import Placeholder from './placeholder'
+import TlfType from './tlf-type-container'
+import Tlf from './tlf-container'
```
jzila: Instead of doing all these imports then a single export, you can do:
```javascript
export {default as Tlf} from './tlf-container'
```

shared/fs/row/still-container.js
```diff
@@ -1,62 +1,50 @@
...
+  const _pathItem: Types.PathItem = state.fs.pathItems.get(path, Constants.unknownPathItem)
   const _username = state.config.username || undefined
   const _downloads = state.fs.downloads
   return {
-    _username,
     _downloads,
-    path,
-    kbfsEnabled: StateMappers.mapStateToKBFSEnabled(state),
-    pathItem,
+    _kbfsEnabled: StateMappers.mapStateToKBFSEnabled(state),
```
jzila: This whole `state-mappers` file doesn't seem useful to me. It just has one function and it's not named in a way that is intuitive. This util function probably belongs in `constants/fs.js`.

shared/fs/row/still.js
```diff
@@ -1,152 +1,66 @@
 // @flow
 import * as React from 'react'
 import * as Types from '../../constants/types/fs'
-import {globalColors, globalMargins, isMobile, glamorous} from '../../styles'
-import rowStyles from './styles'
-import {Badge, Box, Box2, ClickableBox, Icon, Meta, Text} from '../../common-adapters'
-import PathItemIcon from '../common/path-item-icon'
+import {globalColors, globalMargins, isMobile} from '../../styles'
+import {rowStyles, StillCommon, type StillCommonProps} from './styles'
```
jzila: I don't think `StillCommon` and `StillCommonProps` should be in a file/directory called `styles`.

shared/fs/row/index.js
jzila: I also believe having separate row types for `TlfType` and Tlf` is overkill - it forces you to do further indirection like `StillCommon` and `StillCommonProps`, making it painful to find the one place you finally need to implement a change. `TlfType` and `Tlf` should just exist in the same place as `Still`, and all the encapsulation should occur in that file. Callers should just include a `Still` row as necessary.

shared/fs/row/still.js
```diff
@@ -1,152 +1,66 @@
...
+  <StillCommon
+    itemStyles={props.itemStyles}
+    name={props.name}
+    path={props.path}
+    onOpen={props.onOpen}
+    openInFileUI={props.openInFileUI}
+  >
+    <RowMeta isDownloading={props.isDownloading} />
+    <Box style={rowStyles.itemBox}>
+      <Box2 direction="horizontal" fullWidth={true}>
```
jzila: Why are we using both `Box` and `Box2`? Can we just use one?

shared/fs/row/tlf-container.js
```diff
@@ -0,0 +1,58 @@
...
+    _username,
+    _kbfsEnabled,
+    _tlf,
+  }
+}
+
+const mapDispatchToProps = (dispatch: Dispatch, {routePath}: OwnProps) => {
+  return {
+    _onOpen: (path: Types.Path) => dispatch(FsGen.createOpenPathItem({path, routePath})),
+    _openInFileUI: (path: Types.Path) => dispatch(FsGen.createOpenInFileUI({path: Types.pathToString(path)})),
```
jzila: Lots of duplicated logic here because of the aforementioned over-separation.

shared/fs/row/still.js
songgao: I think we are migrating to Box2, but there are still a lot of Box left over. They still work so probably not worthwhile to do them together in this ticket?

shared/fs/row/still.js
songgao: There's a TODO in `./styles` to rename it in `common.js`, and is addressed in the 6th of this batch of PRs (mechanical changes).

shared/fs/row/still-container.js
songgao: I just copied them from the old code. I'm OK with getting rid of this or organize them in a different way, but didn't want to address too much initially. I can try to clean these up in a follow-up commit to this branch.

jzila:
You've duplicated a lot of the container logic in tlf, tlf-type, and still though. You can put them all in the same place and reduce that duplication.

jzila:
To be clear, I'm not saying they need to be the same object. I'd just like an abstraction layer hiding them away. Unless the intent is to no longer do `SimpleFSList` for the root and `TlfType` levels, then doing a `SimpleFSList` at the `Tlf` level, and finally `SimpleFSListRecursiveToDepth` at subfolder levels...

songgao:
> I'd just like an abstraction layer hiding them away.

Sure! I'll try to abstract the common part away from the container. Thanks for bring it up.

> Unless the intent is to no longer do SimpleFSList for the root level, then doing a SimpleFSList at the TlfType level, and finally SimpleFSListRecursiveToDepth at subfolder levels...

I've been trying to make it rely on SimpleFSListRecursiveToDepty for only TLF and everything inside, and just use favorites to construct TLF lists. I didn't really like layering results from both (favorites and simplefslist) calls together, and it seemed like the TLF lists only nee stuff from favorites, thus the change.

jzila:
> I've been trying to make it rely on SimpleFSListRecursiveToDepty for only TLF and everything inside, and just use favorites to construct TLF lists

This seems like 100% the right approach to me.

shared/fs/row/styles.js
songgao: These are actually fed into `styleSheetCreate` below when `rowStyles` is exported.

shared/fs/row/tlf.js
songgao: It was this structure in the previous version already. I'm reluctant to change it in this ticket, but I'm happy to try changing it later next time we need to touch it.

songgao:

PR #13104, opened 2018-08-01
decouple TLF metadata out of PathItems 4: component changes - other

songgao:


shared/fs/banner/reset-banner/container.js
```diff
@@ -12,18 +12,10 @@ import {createGetProfile} from '../../../actions/tracker-gen'
...
-    _username,
-    path,
-    resetParticipants:
-      _pathItem.type === 'folder' && _pathItem.tlfMeta ? _pathItem.tlfMeta.resetParticipants : [],
-    _teamId: _pathItem.type === 'folder' && _pathItem.tlfMeta ? _pathItem.tlfMeta.teamId : '',
-  }
-}
+const mapStateToProps = (state: TypedState, {path}) => ({
+  _tlf: Constants.getTlfFromPath(state.fs.tlfs, path),
+  _username: state.config.username || undefined,
```
chrisnojima: why ||  undefined

shared/fs/common/path-item-action-container.js
```diff
@@ -14,28 +14,25 @@ import PathItemAction from './path-item-action'
...
-  return {
-    _pathItems,
-    _username,
-    fileUIEnabled: state.favorite.fuseStatus ? state.favorite.fuseStatus.kextStarted : false,
-  }
-}
+const mapStateToProps = (state: TypedState) => ({
+  _pathItems: state.fs.pathItems,
+  _tlfs: state.fs.tlfs,
+  _username: state.config.username || undefined,
```
chrisnojima: same here

shared/fs/common/path-item-action-container.js
```diff
@@ -53,62 +50,127 @@ const mapDispatchToProps = (dispatch: Dispatch, {path}: OwnProps) => ({
...
-  Constants.showIgnoreFolder(path, pathItem, _username) &&
-    menuItems.push({
-      title: 'Ignore this folder',
-      onClick: () => _ignoreFolder(),
-      subTitle: 'The folder will no longer appear in your folders list.',
-      danger: true,
-    })
-
-  menuItems.push({
+type MenuItemAppender = (
```
chrisnojima: I think its likely better for this construct to change. I was speaking to danny about this yesterday. I think it likely makes more sense for the mapStateToDispatch to just create the dispatchProps and the component itself to make these menu items. 

My argument is the role of the connector is to bridge the simple props the component needs to draw itself and to give it callbacks it can call to do stuff. The data structure MenuItem is an implementation detail of the component and doesn't really have anything to do with redux.

I argue better to just be like 
```
const mergeProps...
return {
  showInFileUI: !isMobile && stateProps._fileUIEnabled && dispatchProps._showInFileUI
...
}
```

then the component itself does in its render

```
const menuItems = [
 ...(this.props.showInFileUI ? {onClick: this.props.showInFileUI, label: 'Show in Findere'} : {}),
]
```

For example if we change how menu works or if the view changes to be a button bar instead of a menu the connector would have to change, which I argue isn't ideal

shared/fs/common/path-item-action-container.js
```diff
@@ -53,62 +50,127 @@ const mapDispatchToProps = (dispatch: Dispatch, {path}: OwnProps) => ({
...
+) =>
+  appenders.reduce(
+    (menuItems: Array<MenuItem>, appender: MenuItemAppender) =>
+      appender(menuItems, stateProps, dispatchProps, path),
+    ([]: Array<MenuItem>)
+  )
+
+const getRootMenuItemsByPathLevel = (pathLevel: number, stateProps, dispatchProps, path: Types.Path) => {
+  switch (pathLevel) {
+    case 0:
```
chrisnojima: maybe you want a constant like
```
const getPathType(path): 'keybase' | 'tlfRoot' | 'tlf' | 'tlfSub'
```

then your code will be less 
```
const needFolderList = type === 'folder' && pathElements.length >= 3,
```
and instead
```
const needFolderList = pathType === 'tlfSub'
```

shared/fs/common/path-item-action-container.js
songgao: Sounds good; will make the change. Thanks for explaining the details!

shared/fs/common/path-item-action-container.js
songgao: Hmm I tried this but I don't think there's gonna be a difference here other than making it more clear. We still need to check if the thing is a folder.

chrisnojima: sure it can be 
'keybase' | 'tlfRoot' | 'tlf' | 'tlfSub' | 'file'

songgao: Hmm I'm not sure that'd work well: 1) To know if it's a file or a folder, we'd need to consult the store `state.fs.pathItems`, and look at a `PathItem`'s `type` field. So `getPathType` would need to accept a `PathItem` arg; 2) If we are trying to make it reflect value from the `type` field, we'd need to include not only `'file'`, but also `'symlink'`. Then we can't simply use `pathType !== 'file'` determine if a thing is a folder.

songgao:

PR #13109, opened 2018-08-01
decouple TLF metadata out of PathItems 6: mechanical changes

songgao:


songgao:

PR #13110, opened 2018-08-01
reacjis in outbox

joshblum:
this is still WIP on the GUI integration from @buoyad  but @mmaxim can you take a look at the core side?


buoyad:
@keybase/react-hackers this is ready for a look! it changes the reaction update API

Previously on clicking the button we send `postReactionNonblock` and listen for incoming messages & deletes to update the state of the buttons. We stored the reaction msg ID to look up deletes later.

Now on clicking the button we send `postReactionNonblock` and add the reaction to the store immediately. We ignore incoming messages and deletes and listen for a new notification, `ReactionUpdate` to give us the new reaction map for the message. On getting the notification we blow away our version of the reaction map and replace it with the service's. We don't store the reaction msg ID because we don't manually delete individual reactions anymore.

mmaxim:
Merge?

PR #13112, opened 2018-08-02
Don't show a timestamp due to orange line above a message if we haven't loaded that message yet

cjb:
@keybase/react-hackers 

We were actually already testing whether a timestamp is non-zero when deciding whether to display it, in `Constants.enoughTimeBetweenMessages()`.  This bug appeared when we changed the logic to check for that function returning true *or* the orange line being above a message.  The bug appears when you're trying to load a new message but haven't got it yet, and so your orange line is above that message, and then we try to render a timestamp above that new message but haven't loaded its timestamp yet.

So this seems like the fix.  A bit difficult to test.

### DIFF
--- a/shared/chat/conversation/messages/wrapper/wrapper-timestamp/container.js
+++ b/shared/chat/conversation/messages/wrapper/wrapper-timestamp/container.js
@@ -13,7 +13,7 @@ type OwnProps = {
 const mapStateToProps = (state: TypedState, ownProps: OwnProps) => {
   const lastReadMessageID = state.chat2.lastReadMessageMap.get(ownProps.message.conversationIDKey)
   // Show the orange line on the first message after the last read message
-  // Messages sent sent by you don't count
+  // Messages sent by you don't count
   const orangeLineAbove =
     !!ownProps.previous &&
     lastReadMessageID === ownProps.previous.id &&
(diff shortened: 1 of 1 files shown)


cjb:
@chrisnojima This is about a placeholder message, not a pending outgoing message -- we initialize to 0 for placeholder messages:
```js
const makeMessageMinimum = {
  author: '',
  conversationIDKey: noConversationIDKey,
  id: Types.numberToMessageID(0),
  ordinal: Types.numberToOrdinal(0),
  timestamp: 0,
}

const makeMessageCommon = {
  ...makeMessageMinimum,
  deviceName: '',
  deviceRevokedAt: null,
  deviceType: 'mobile',
  errorReason: null,
  hasBeenEdited: false,
  outboxID: Types.stringToOutboxID(''),
}

export const makeMessagePlaceholder: I.RecordFactory<MessageTypes._MessagePlaceholder> = I.Record({
  ...makeMessageCommon,
  type: 'placeholder',
})
```

chrisnojima:
we should be able to get this timestamp from the pending i think, can we @mmaxim

cjb:
Hm, it looks like we usually get a `ctime` with messages but not in this case:
```
export type MessageUnboxedPlaceholder = $ReadOnly<{messageID: MessageID, hidden: Boolean}>
```

cjb:
Changed the test after convincing myself that we could see a placeholder message in the `!previous` case as well as the `orangeLineAbove` case: when you're loading a convo and the first message is a placeholder.

mmaxim:
We don't know anything about placeholder messages other than they exist, so unfortunately no time is available.

cjb:
Okay, thanks! @chrisnojima I think this PR's the right thing, then.

PR #13122, opened 2018-08-02
Fix orange line logic on entering a conversation with no orange line

nathunsmitty:
## Problem
If a user entered a conversation with no orange line then received a chat, then they would see an orange line. (thx @buoyad for noticing!)

The reason for this bug is in how I initially made the check to display the orange line. The problem line was this line in `shared/chat/conversation/messages/wrapper/wrapper-timestamp/container.js`:
```js
const orangeLineAbove = !!ownProps.previous && lastReadMessageID === ownProps.previous.id
```
Essentially, the orange line was displayed if there was ever a message after a user's last read message. This meant that the orange line could be displayed after a conversation was selected, when new messages came in. This was most notable when a user would send a message in a chat with no orange line, which was fixed by #12950. However, this problem of a user receiving a chat in a conversation with no orange line displayed was still there.

## Solution
This PR should get rid of any issues with new messages affecting the display of the orange line by caching the message id (if any) an orange line should be shown above at versus the current solution of caching the last read message id and determining whether to display the orange line. As a result, the orange line logic has moved from `shared/chat/conversation/messages/wrapper/wrapper-timestamp/container.js` to `shared/reducers/chat2.js`. I also found that I needed to plumb a new property, `maxMsgID`, in.


shared/chat/conversation/messages/wrapper/wrapper-author/container.js
```diff
@@ -12,8 +12,8 @@ const mapStateToProps = (state: TypedState, {message, previous, innerClass, isEd
   const isYou = state.config.username === message.author
   const isFollowing = state.config.following.has(message.author)
   const isBroken = state.users.infoMap.getIn([message.author, 'broken'], false)
-  const lastReadMessageID = state.chat2.lastReadMessageMap.get(message.conversationIDKey)
-  const lastPositionExists = !!previous && lastReadMessageID === previous.id
+  const {readMsgID, maxMsgID} = Constants.getMeta(state, message.conversationIDKey)
```
nathunsmitty: Some please correct me if I'm wrong, but it seems like `Constants.getMeta` is the preferred way to get values from the chat2 metaMap?

shared/chat/conversation/messages/wrapper/wrapper-author/container.js
```diff
@@ -12,8 +12,8 @@ const mapStateToProps = (state: TypedState, {message, previous, innerClass, isEd
...
+  const lastPositionExists = maxMsgID > readMsgID
```
nathunsmitty: `maxMsgID > readMsgID` is the same logic as `!!previous && lastReadMessageID === previous.id` – are there messages after the last read message?

I like `maxMsgID > readMsgID` because it doesn't rely on `orangeLineMap`. I think it's better that that map be used only for displaying the orange line.

shared/constants/chat2/index.js
```diff
@@ -202,6 +202,7 @@ export {
 
 export {
   allMessageTypes,
+  enoughTimeBetweenMessages,
```
nathunsmitty: I'm not really sure why `enoughTimeBetweenMessages` wasn't being exported from chat2 constants but was being exported from the messages file. Since there were files that needed both `getMeta` and `enoughTimeBetweenMessages`, it made sense to add `enoughTimeBetweenMessages`!

shared/constants/chat2/index.js
```diff
@@ -31,7 +31,7 @@ export const makeState: I.RecordFactory<Types._State> = I.Record({
     [pendingConversationIDKey, makeConversationMeta({conversationIDKey: noConversationIDKey})],
   ]),
   moreToLoadMap: I.Map(),
-  lastReadMessageMap: I.Map(),
+  orangeLineMap: I.Map(),
```
nathunsmitty: Renamed `lastReadMessageMap` back to `orangeLineMap` to make it clearer that this map should really only be used to display an orange line.

chrisnojima:
merging for the release

PR #13123, opened 2018-08-02
Add CLI 'wallet rename'

mlsteele:
```
$ keybase wallet rename GBUFL2MYSRVOCZ6NZMAQJZZZHV6MXB35CP2CE4L6ITBZKAKZJQRQOBT3 'funny money'
✓ Account renamed
```

### DIFF
--- a/go/client/cmd_wallet_rename.go
+++ b/go/client/cmd_wallet_rename.go
@@ -1,0 +1,72 @@
+// Copyright 2018 Keybase, Inc. All rights reserved. Use of
+// this source code is governed by the included BSD license.
+
+package client
+
+import (
+	"errors"
+
+	"github.com/keybase/cli"
+	"github.com/keybase/client/go/libcmdline"
+	"github.com/keybase/client/go/libkb"
+	"github.com/keybase/client/go/protocol/stellar1"
+	"golang.org/x/net/context"
+)
+
+type cmdWalletRename struct {
+	libkb.Contextified
+	AccountID stellar1.AccountID
+	Name      string
+}
+
+func newCmdWalletRename(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	cmd := &cmdWalletRename{
+		Contextified: libkb.NewContextified(g),
+	}
+	return cli.Command{
+		Name:         "rename",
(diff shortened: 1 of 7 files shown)


go/client/cmd_wallet_rename.go
```diff
@@ -0,0 +1,72 @@
...
+	c.Name = ctx.Args()[1]
+	return nil
+}
+
+func (c *cmdWalletRename) Run() (err error) {
+	cli, err := GetWalletClient(c.G())
+	if err != nil {
+		return err
+	}
+	err = cli.ChangeWalletAccountNameLocal(context.TODO(), stellar1.ChangeWalletAccountNameLocalArg{
```
zapu: IIRC we wanted to go with `context.Background()` everywhere, unless some place is actually "TODO" because context is not plumbed yet.

go/stellar/stellar.go
```diff
@@ -990,7 +1002,8 @@ func ChangeAccountName(m libkb.MetaContext, accountID stellar1.AccountID, newNam
 			// Change Name in place to modify Account struct.
 			nextBundle.Accounts[i].Name = newName
 			found = true
-			break
+		} else if acc.Name == newName {
+			return fmt.Errorf("that account name is already taken")
```
zapu: maybe "you already have an account with that name" ? so we don't imply that names are unique across the entire network

go/stellar/stellarsvc/frontend_test.go
```diff
@@ -224,8 +224,8 @@ func TestChangeWalletName(t *testing.T) {
 	require.Error(t, err)
 	chk("savings", "")
 
-	chk("an account used for saving up for the elephant's new pajamas", "")
```
zapu: 🐘😢

go/stellar/stellar.go
zapu: oh that's weird. we already return that error from `ValidateAccountNameLocal`. maybe nevermind then

mlsteele: I like yours better, will change both

go/client/cmd_wallet_rename.go
mlsteele: This one is actually TODO isn't it?

PR #13125, opened 2018-08-03
Header -> BodySemibold in stellar account input

zanderz:
The ticket was unclear on which text exactly needed to be changed. It sounded like the input text, but it was not set to "Header" - see e.g. https://github.com/keybase/client/blob/master/shared/common-adapters/input.desktop.js#L244

### DIFF
--- a/shared/wallets/receive-modal/index.js
+++ b/shared/wallets/receive-modal/index.js
@@ -1,15 +1,6 @@
 // @flow
 import * as React from 'react'
-import {
-  Box2,
-  Button,
-  CopyText,
-  Icon,
-  InfoNote,
-  MaybePopup,
-  Text,
-  iconCastPlatformStyles,
-} from '../../common-adapters'
+import * as Kb from '../../common-adapters'
 import {globalMargins, isMobile, platformStyles, styleSheetCreate} from '../../styles'
 
 type Props = {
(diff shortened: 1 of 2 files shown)


shared/wallets/receive-modal/index.js
```diff
@@ -1,15 +1,6 @@
...
-  Box2,
-  Button,
-  CopyText,
-  Icon,
-  InfoNote,
-  MaybePopup,
-  Text,
-  iconCastPlatformStyles,
-} from '../../common-adapters'
+import * as Kb from '../../common-adapters'
```
jzila: Why this change? I haven't seen this pattern anywhere else...

adamjspooner: It's the new hotness.

jzila: But it makes the markup distinctly less readable...

chrisnojima: I think this ultimately boils down to a stylistic choice. I don't find namespaces components really that unreadable. We have started using this style recently. I find that it's just easier development wise instead of having to keep managing the imports and add all this vertical space to pull in the same components over and over

zanderz: I like it. I assumed we did it the old way because there was some advantage to not importing everything like that.

zanderz:
This seems to be outdated and someone else changed it already.

PR #13133, opened 2018-08-03
Add NeedApplicationsAtGenerations refresher

joshblum:
Fixes a bug where the users cache is in a bad state. This could happen if the user was first the implicit admin of a subteam, cached the team secrets and later was added as a member but the cache was without the reader key masks. To fix we introduce a new refresher `NeedApplicationsAtGenerations` which applications should use in favor of `NeedKeyGeneration` to ensure the RKMs are also present. 

cc @zapu


protocol/avdl/keybase1/teams.avdl
```diff
@@ -516,6 +516,9 @@ protocol teams {
   record TeamRefreshers {
     // Load at least up to the keygen. Returns an error if the keygen is not loaded.
     PerTeamKeyGeneration needKeyGeneration;
+    // Load the reader key mask for the given applications and the given generations. Returns an error
+    // any of the RKMs are not loaded.
```
mlsteele: any -> if any

go/teams/loader.go
```diff
@@ -517,6 +519,10 @@ func (l *TeamLoader) load2InnerLockedRetry(ctx context.Context, arg load2ArgT) (
 			l.G().Log.CDebugf(ctx, "TeamLoader fetching: NeedKeyGeneration: %v", err)
 			fetchLinksAndOrSecrets = true
 		}
+		if err := l.satisfiesNeedApplicationsAtGenerations(ctx, arg.needApplicationsAtGenerations, ret); err != nil {
+			l.G().Log.CDebugf(ctx, "TeamLoader fetching: NeedApplicationsAtGenerations: %v", err)
+			fetchLinksAndOrSecrets = true
+		}
```
mlsteele: I would've thought that in order to fix the problem this `satisfiesNeedApplicationsAtGenerations` call would need to be outside of it `! hasSyncedSecrets` block. Since I thought `hasSyncedSecrets` would be true at line 925 in the test. Do you get why it works this way? Great test by the way.

protocol/avdl/keybase1/teams.avdl
joshblum: extentions

go/teams/loader.go
joshblum: ah i think that these changes get `repoll=true` so we fetch the merkle root and see the new link with us as a member and then refetch. I'll move this outside of the conditional block though i think it makes more sense

go/teams/loader.go
```diff
@@ -527,6 +529,9 @@ func (l *TeamLoader) load2InnerLockedRetry(ctx context.Context, arg load2ArgT) (
 			l.G().Log.CDebugf(ctx, "TeamLoader fetching: primary load")
 			fetchLinksAndOrSecrets = true
 		}
+	} else if err := l.satisfiesNeedKeyGeneration(ctx, arg.needKeyGeneration, ret); err != nil { // hasSyncedSecrets does not account for RKMs so we verify it separately.
+		l.G().Log.CDebugf(ctx, "TeamLoader fetching: NeedKeyGeneration: %v", err)
+		fetchLinksAndOrSecrets = true
```
mlsteele: I think this should be `if` not `} else if`. The others are in a else-if cascade because the first two set `fetchLinksAndOrSecrets`. But the `!hasSyncedSecrets` block doesn't necessarily do that. So there could be a case where `!hasSyncedSecrets` and `!satisfiesNeedKeyGeneration` but `fetchLinksAndOrSecrets` ends up unset but when it should be set.

go/teams/loader_test.go
```diff
@@ -874,3 +874,68 @@ func TestLoaderCORE_7201(t *testing.T) {
...
+	_, err = Load(context.TODO(), tcs[1].G, keybase1.LoadTeamArg{
+		ID:          subBID,
+		ForceRepoll: true,
+	})
+	require.NoError(t, err)
+
+	t.Logf("U0 adds U1 to A.B")
+	_, err = AddMember(context.TODO(), tcs[0].G, subBName.String(), fus[1].Username, keybase1.TeamRole_WRITER)
+	require.NoError(t, err, "add member")
+	t.Logf("setup complete")
```
mlsteele: If there were two versions of this test, one where the order of 
`U0 adds U1 to A.B`
`U1 loads A.B without refreshing`
are swapped, would that catch some of the bugs we're discussing in loader.go?

go/teams/loader.go
joshblum: ah yes you're right here

mlsteele: Could change the others from `} else if` to `if !fetchLinksAndOrSecrets &&` if you think that's easier to think about.

go/teams/loader_test.go
joshblum: hmm i don't think so since when we still force `repoll=true` and get the new chain link

joshblum: i'm not sure we can force the state where we have all of the links but no rkms.. but still think it's best to have  `fetchLinksAndOrSecrets` set correctly

PR #13135, opened 2018-08-03
IncomingMessage local and remote

joshblum:
Separates incoming messages by ChatActivity source to clean up testing code and help prevent races and CI flakes.

This also re-enables the `TestChatSrvSetConvMinWriterRole`, @patrickxb lmk if this still fails locally for you after the patch


patrickxb:
I checked out your branch and the test still fails locally.

patrickxb:
passed!

PR #13139, opened 2018-08-03
[KBFS-3231] New widget dropdown items

jzila:


shared/app/tab-bar/index.desktop.js
```diff
@@ -36,7 +36,7 @@ const _labels = {
 const _tabs = [
   Tabs.peopleTab,
   Tabs.chatTab,
-  ...(flags.fsEnabled ? [Tabs.fsTab] : [Tabs.folderTab]),
+  Tabs.fsTab,
```
songgao: Might as well remove the `fsEnabled` feature flag?

shared/menubar/index.desktop.js
```diff
@@ -153,7 +188,18 @@ class MenubarRender extends Component<Props, State> {
...
 
-    const badgeTypes: Array<Tab> = [peopleTab, chatTab, fsTab, teamsTab]
+    const badgeTypesInHeader: Array<Tab> = [peopleTab, chatTab, fsTab, teamsTab]
+    const badgesInMenu: Set<string> = new Set([
+      ...(flags.walletsEnabled ? [walletsTab] : []),
+      gitTab,
+      devicesTab,
+      settingsTab,
+    ])
+    const badgeCountInMenu = Object.entries(this.props.badgeInfo).reduce(
```
songgao: I'm curious why a `reduce` is used here rather than counting on `badgesInMenu`. Do we have situations where things exist in `badgesInMenu` but not `this.props.badgeInfo`?

shared/menubar/index.desktop.js
```diff
@@ -74,14 +85,14 @@ class MenubarRender extends Component<Props, State> {
...
-            type="iconfont-hamburger"
+            type="iconfont-nav-more"
             onClick={() => this.setState(prevState => ({showingMenu: !prevState.showingMenu}))}
             ref={this.attachmentRef}
           />
           <FloatingMenu
             visible={this.state.showingMenu}
             attachTo={this.attachmentRef.current}
-            items={this._menuItems()}
+            items={this._menuItems(this.props.badgeInfo || {})}
```
songgao: `badgeInfo` is `{[string]: number}`, i.e. can't be `null` or `undefined`. So this shouldn't be needed right? Same for below.

shared/menubar/index.desktop.js
jzila: Good point. I'll change it.

shared/menubar/index.desktop.js
jzila: I know but Flow yelled at me. ¯\_(ツ)_/¯

PR #13143, opened 2018-08-03
Add minWriterRole dropdown to info panel

buoyad:
r? @keybase/react-hackers cc @keybase/design for copy

Admin view:
image

Non-admin view:
image


shared/chat/conversation/info-panel/min-writer-role/container.js
```diff
@@ -0,0 +1,40 @@
...
+
+const emptyProps = {
+  canSetMinWriterRole: false,
+  minWriterRole: 'reader',
+}
+
+const mapStateToProps = (state: Container.TypedState, ownProps: OwnProps) => {
+  const meta = Constants.getMeta(state, ownProps.conversationIDKey)
+  if (!meta.teamname) {
+    return emptyProps
```
adamjspooner: What is `emptyProps`?

shared/chat/conversation/info-panel/min-writer-role/index.js
```diff
@@ -0,0 +1,161 @@
...
+      }
+      this._setSelected(this.props.minWriterRole)
+    }
+  }
+
+  render() {
+    const items = TeamConstants.teamRoleTypes.map(role => ({
+      onClick: () => this._selectRole(role),
+      title: upperFirst(role),
+    }))
```
adamjspooner: Does this need to happen inside `render()`?

shared/chat/conversation/info-panel/min-writer-role/container.js
buoyad: I don't think this will ever happen since `meta.teamname` is used in the info panel to decide whether this will render. But if this does happen it'll render the non-admin view from the original comment.

shared/chat/conversation/info-panel/min-writer-role/index.js
buoyad: Good point - I'll move it to `componentDidUpdate`

buoyad: actually it's a little more complicated that way and I don't think this will render that often. I'll TODO and skip for now.

shared/chat/conversation/info-panel/min-writer-role/container.js
adamjspooner: Oh, I just didn't see where it was defined was all.

PR #13146, opened 2018-08-05
use floating menu exclusively for selecting attachment type

mmaxim:
@keybase/react-hackers 

Since taking video on iOS is broken in `react-native-image-picker` (it doesn't get the right set of perms), the easiest path forward is to just not allow it for now. This patch dumps invoking the native control from `showImagePicker` and just uses `FloatingMenu` for everything, taking advantage of `launchCamera` and `launchImageLibrary`. The Android experience is now improved, and I think the whole thing feels faster (the native popup on Android was pretty sluggish, this new deal just feels snappier). Here is what it looks like when you tap the camera icon on both platforms:

iOS:
image

Android:
image

### DIFF
--- a/shared/chat/conversation/input-area/filepicker-popup/index.native.js
+++ b/shared/chat/conversation/input-area/filepicker-popup/index.native.js
@@ -3,6 +3,7 @@ import * as React from 'react'
 import {FloatingMenu} from '../../../../common-adapters'
 import {Box2, Text} from '../../../../common-adapters/mobile.native'
 import type {Props} from './index.types'
+import {isIOS} from '../../../../constants/platform'
 
 const Prompt = () => (
   <Box2 direction="horizontal" fullWidth={true} gap="xtiny" style={promptContainerStyle}>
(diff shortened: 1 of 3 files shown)


mmaxim:
FYI: this same crasher is present in Files if you "Take Photo" and try to do video, not sure if you all know about that cc @songgao @jzila

shared/chat/conversation/input-area/filepicker-popup/index.types.js.flow
```diff
@@ -5,5 +5,5 @@ export type Props = {
   attachTo: ?React.Component<any, any>,
   visible: boolean,
   onHidden: () => void,
-  onSelect: string => void,
+  onSelect: (string, string) => void,
```
chrisnojima: we should name this param now that there's 2

shared/chat/conversation/input-area/filepicker-popup/index.native.js
```diff
@@ -16,16 +17,60 @@ const promptContainerStyle = {
...
-
-  onVideo = () => {
-    this.props.onSelect('video')
-  }
-
   render() {
-    const items = [...[{onClick: this.onImage, title: 'Photo'}], ...[{onClick: this.onVideo, title: 'Video'}]]
+    const items = isIOS
+      ? [
+          ...[
```
chrisnojima: i know this is based on the old code but this... isn't necessary at all, can be
```
items = isIOS ? [
 {
              onClick: () => {
                this.props.onSelect('photo', 'camera')
              },
              title: 'Take Photo',
            },
{
              onClick: () => {
                this.props.onSelect('mixed', 'library')
              },
              title: 'Photo or Video from Library',
            },
]
```

or even better the whole block can be

```
const items = [
    {
        onClick: () => { this.props.onSelect('photo', 'camera') },
        title: 'Take Photo',
    },
    ...(isIOS 
    ? [{
        onClick: () => { this.props.onSelect('mixed', 'library') },
        title: 'Photo or Video from Library',
      }] 
    : [{
        onClick: () => { this.props.onSelect('video', 'camera') },
        title: 'Take Video',
    },
    {
        onClick: () => { this.props.onSelect('photo', 'library') },
        title: 'Photo from Library',
    },
    {
        onClick: () => { this.props.onSelect('video', 'library') },
        title: 'Video from Library',
    },
    ])
]


```

mmaxim: Dropped the `...`, but kept it separate, I think it is nice to be able to read it clearly for each platform.

shared/chat/conversation/input-area/filepicker-popup/index.types.js.flow
```diff
@@ -5,5 +5,5 @@ export type Props = {
   attachTo: ?React.Component<any, any>,
   visible: boolean,
   onHidden: () => void,
-  onSelect: string => void,
+  onSelect: (mediaType: string, location: string) => void,
```
chrisnojima: actually should be 
mediaType: 'photo' | 'video' | 'mixed', location: 'library' | 'camera'

PR #13148, opened 2018-08-05
don't compute a plaintext hash anymore for uploads, use outboxID for stash CORE-8432

mmaxim:
Patch does the following:

1.) Drop computing the plaintext hash for uploads. It can be pretty slow for large enough files (big video files).
2.) Replace `plaintextHash` on `UploadTask` with a `taskHash` of the outboxID and whether or not the upload is a preview or full asset. We then SHA256 a string of those two values to get the hash of the upload.

### DIFF
--- a/go/chat/attachments/s3.go
+++ b/go/chat/attachments/s3.go
@@ -119,6 +119,8 @@ func (a *S3Store) putMultiPipeline(ctx context.Context, r io.Reader, size int64,
 		list, err := multi.ListParts(ctx)
 		if err != nil {
 			a.Debug(ctx, "putMultiPipeline: ignoring multi.ListParts error: %s", err)
+			// dump previous since we can't check it anymore
+			previous = nil
 		} else {
 			for _, p := range list {
 				previousParts[p.N] = p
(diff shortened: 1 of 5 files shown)


patrickxb:
I believe it was a plaintext hash over the whole file so that if a multipart upload didn't finish then when the user did another attempt to upload the exact same file, it would be found in the stash.  

Does the new logic handle that?

mmaxim:
Yes, I just changed the key to be `<outboxID, preview>`. The only slight disadvantage here is that if the file changes in between attempts, then we will fail on the `ListParts` phase more often with this method. I think it is better though overall, since that hash over the whole file could be pretty slow on phones. I tried uploading a 2.5GB video and it would stall for 5-10 seconds on the hash.

mmaxim:
I think before since there was no real way to retry an attachment, the file hash was the only way to identify the user was trying to upload the same thing. However, now that a failed attachment stays in the thread view and exposes a "Retry" button, using outbox ID I think is the better play here.

patrickxb:
Sounds good, just wanted to make sure the original intention wasn't lost.

PR #13158, opened 2018-08-06
Defer PTK verification until access

mlsteele:
Defer the check that a per-team key box matches the sigchain. It used to check when downloaded from the server. Now it stores it in `PerTeamKeySeedsUnverified` and verifies it every time it is accessed. This change makes loading an initial team faster at the expense of future operations involving its keys.

Saves 5 android seconds. And about 150ms on macos. The difference stemming from key generation operations being slow probably due to there being assembly implementations only for amd64.


mlsteele:
It got me
```
Test:       	TestUnits
Messages:   	unexpected team load success in ptk_mismatch_secret.json
```

PR #13159, opened 2018-08-06
let re-admit button show up when it should

buoyad:
little typo. r? @keybase/react-hackers


zapu:
oh!

mmaxim:
We should get this in the release

PR #13164, opened 2018-08-06
Add common-adapters/banner storybook

nathunsmitty:
Wanted to check out the Banner component and didn't see a story for it, so I just wrote one up. Pretty bare bones, but I figure it's a good starting point if anyone else ends up needing to use Banner.


nathunsmitty:
@keybase/react-hackers this should be pretty quick and easy!

PR #13179, opened 2018-08-07
Better team member sort

buoyad:
I noticed a problem with the member sort while testing reset & deleted users. The previous logic was getting getting contrived; I replaced it with some weights and I think it's much simpler. r? @keybase/react-hackers


chrisnojima:
- [ ] after release

buoyad:
@keybase/react-hackers bump!

PR #13183, opened 2018-08-07
dont center on adnroid due to keyboard avoiding crappiness

chrisnojima:
@keybase/react-hackers this moves the paperkey input up on android. keyboard avoiding on android basically doens't work
https://github.com/facebook/react-native/issues/16826

### DIFF
--- a/shared/provision/paper-key/index.js
+++ b/shared/provision/paper-key/index.js
@@ -8,6 +8,7 @@ import {
   globalStyles,
   styleSheetCreate,
   isMobile,
+  isIOS,
   platformStyles,
 } from '../../styles'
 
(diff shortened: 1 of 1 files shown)


shared/provision/paper-key/index.js
```diff
@@ -27,7 +28,7 @@ const PaperKey = (props: Props) => (
     <Box2
       direction="vertical"
       style={styles.contents}
-      centerChildren={true}
+      centerChildren={isIOS /* android keyboardAvoiding doesnt work well */}
```
buoyad: Actually, I think this should be `!isAndroid`

PR #13185, opened 2018-08-07
changelog

chrisnojima:
@keybase/react-hackers


adamjspooner:
Do we want to mention keyboard shortcuts?

mmaxim:
You could add a line about attachment upload working better maybe. Is this just for desktop? Maybe something like "Improve attachment upload experience".

PR #13168, opened 2018-08-06
Stellar send participants list

zanderz:
Note the Stellar icon is TBD still.

image

image

image

PR #13187, opened 2018-08-07
Storybook: Stellar confirm payment screen

nathunsmitty:
[Check it out!](http://localhost:6006/?selectedKind=Wallets%2FConfirmForm&selectedStory=Normal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

Note that I moved `Banner` up a directory and created a component (`WalletEntry`) that uses a lot of `WalletRow` code. As components in the `SendForm` get fleshed out, `ConfirmForm` would probably be worth a look into what code is unique to it vs what can be genericized.


shared/wallets/banner/container.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import Banner from '.'
```
nathunsmitty: I'm not sure `Banner` will really need a container, in `ConfirmScreen` I used it without and passed in the props when needed. Because `ConfirmScreen` uses the banner props to determine whether it should be rendered and because it's only one level of prop passing, I think this is fine. I left the container as to not break other files, though.

nathunsmitty:
Mobile styling is done! See it [here](http://localhost:7007/?selectedKind=Wallets%2FConfirmForm&selectedStory=Normal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel).

nathunsmitty:
@keybase/react-hackers this is ready for a look!

nathunsmitty:
@keybase/react-hackers bump

shared/wallets/confirm-form/index.js
```diff
@@ -0,0 +1,111 @@
...
+      <Header
+        amount={props.amount}
+        assetType={props.assetType}
+        assetConversion={props.assetConversion}
+        onBack={props.onBack}
+      />
+      <Kb.ScrollView style={styles.scrollView}>
+        {!!props.bannerBackground &&
+          !!props.bannerText && <Banner background={props.bannerBackground} text={props.bannerText} />}
+        <Participants receivingUsername="nathunsmitty" receivingFullName="Nathan Smith" />
```
cjb: PR looks good!  Except the `<Participants>` are hardcoded inside the component here.  Maybe that was intentional 'cause @zanderz has an open PR on the participants component?  Although his PR doesn't handle the to-another-wallet case.  So maybe we should agree on some props here?

Like:
```
toKeybaseUser: string
toStellarAddress: string
toAnotherWallet: string
```
and have the participants component check those in turn..?

nathunsmitty: Yeah I wasn't entirely sure how I should handle the `Participants` component since I know @zanderz was working on it in #13168 but also the confirm screen's `Participants` were different enough to not totally match the send screen's. I think I would like to see the two components eventually be merged into one or at least share their common code, but that can probably happen when one (or both) of our PRs get merged in.

Definitely will add the extra props 👍

PR #13198, opened 2018-08-07
select keys specifically for relevant generations CORE-8525

mmaxim:


go/chat/teams.go
```diff
@@ -227,7 +238,7 @@ func (t *TeamsNameInfoSource) Lookup(ctx context.Context, name string, public bo
 	if err != nil {
 		return res, err
 	}
-	return t.makeNameInfo(ctx, team, public)
+	return t.makeNameInfo(ctx, team, nil, public)
```
maxtaco: do we ever call `#Lookup` on the way into chat?

mmaxim:
Fails `TestTeamReset` for some reason...

go/chat/boxer.go
```diff
@@ -275,26 +275,13 @@ func (b *Boxer) UnboxMessage(ctx context.Context, boxed chat1.MessageBoxed, conv
 				boxed.ClientHeader.TlfPublic))), nil
 	}
 	keyMembersType := b.getEffectiveMembersType(ctx, boxed, conv.GetMembersType())
-	nameInfo, err := CtxKeyFinder(ctx, b.G()).FindForDecryption(ctx,
+	encryptionKey, err := CtxKeyFinder(ctx, b.G()).FindForDecryption(ctx,
 		tlfName, boxed.ClientHeader.Conv.Tlfid, conv.GetMembersType(),
 		conv.IsPublic(), boxed.KeyGeneration, keyMembersType == chat1.ConversationMembersType_KBFS)
 	if err != nil {
 		// Check to see if this is a permanent error from the server
 		return chat1.MessageUnboxed{}, b.detectKBFSPermanentServerError(err, tlfName)
```
mlsteele: I suspect any `DecryptionKeyNotFoundError` should be wrapped in `NewPermanentUnboxingError` before returning from `UnboxMessage`. Could add it to `detectKBFSPermanentServerError` (which is due for a rename to kitchen sink).

go/chat/teams.go
```diff
@@ -414,7 +432,7 @@ func (t *ImplicitTeamsNameInfoSource) makeNameInfo(ctx context.Context, team *te
 		return res, errors.New("blank TLF ID given")
 	}
 	res.CanonicalName = impTeamName.String()
-	if res.CryptKeys, err = getTeamKeys(ctx, team, public); err != nil {
+	if res.CryptKeys, err = getTeamCryptKeys(ctx, team, public); err != nil {
```
mlsteele: Is this still called all the time? Does NameInfo need keys on it? If refactoring chat is a pain we can cache the verification of keys on teams so getting all keys is fastish again.

go/chat/boxer.go
mmaxim: Yeah true, I'll look at that.

go/chat/teams.go
mmaxim: I'm not quite done, I am going to refactor a little more and make this call very rare.

PR #13202, opened 2018-08-07
Fix android rawtext crasher in ReactionTooltip

buoyad:
r? @keybase/react-hackers


chrisnojima:
this is the motivation for the per platform flow

PR #13210, opened 2018-08-08
Add lockdown mode RPCs and a CLI command

zapu:


go/service/account.go
```diff
@@ -104,3 +104,65 @@ func (h *AccountHandler) ResetAccount(ctx context.Context, arg keybase1.ResetAcc
...
+
+	mctx := libkb.NewMetaContext(ctx, h.G())
+	ss, err := mctx.ActiveDevice().SyncSecrets(mctx)
+	if err != nil {
+		return ret, err
+	}
+	devs, err := ss.ActiveDevices(libkb.AllDeviceTypes)
+	if err != nil {
+		return ret, err
+	}
```
zapu: is it better to call "list devices" engine here?

go/client/cmd_account_lockdown.go
```diff
@@ -0,0 +1,190 @@
...
+		Name:  "lockdown",
+		Usage: "Manage lockdown mode",
+		Flags: flags,
+		Action: func(c *cli.Context) {
+			cl.ChooseCommand(cmd, "lockdown", c)
+		},
+		Description: `Lockdown mode can be enabled for any Keybase account with one of
+   provisioned devices. Enabling lockdown mode does not add a link to
+   user's sigchain. Lockdown mode is an extension of Keybase API
+   server access control.
```
mlsteele: I think it's better to start with "When lockdown mode is enabled, some op..." since that's the main virtue of lockdown mode and provides some context. Move this first paragraph of details to the end.

In the first sentence "with one of provisioned devices" -> something else, or delete the sentence since you'd have to be on a device to be able to run this command anyway.

go/service/account.go
```diff
@@ -104,3 +104,65 @@ func (h *AccountHandler) ResetAccount(ctx context.Context, arg keybase1.ResetAcc
...
+
+type GetLockdownResponse struct {
+	Enabled bool                       `json:"enabled"`
+	Status  libkb.AppStatus            `json:"status"`
+	History []keybase1.LockdownHistory `json:"history"`
+}
+
+func (r *GetLockdownResponse) GetAppStatus() *libkb.AppStatus {
+	return &r.Status
+}
```
mlsteele: libkb.AppStatusEmbed is there to help

go/client/cmd_account_lockdown.go
zapu: thanks, I reworded that thing according to your suggestions. let me know if it can still be improved

go/service/account.go
zapu: wow we should change everything to use that

go/service/account.go
mlsteele: Maybe UPAKLoader then .DeviceKeys.DeviceDescription. That's pretty well cached. I've never encountered this SyncSecrets thing it could be fine up to you.

go/client/cmd_account_lockdown.go
```diff
@@ -0,0 +1,187 @@
...
+       - posting signatures,
+       - changing password or e-mail address,
+       - changing profile information or profile picture.
+
+   These actions are still possible using the Keybase client.
+
+   Enabling lockdown mode does not add a link to user's sigchain, it
+   is an extension of Keybase API server access control. Lockdown mode
+   offers protection in case of someone stealing your browser session
+   or account password.`,
```
zapu: @maxtaco do you want to look at help description for this command? thanks

maxtaco: I'd say nuke the last paragraph.  No need to flog ourselves!

PR #13218, opened 2018-08-08
first stab at prototypes for FECNI

maxtaco:
- these are likely going to change a bunch, but maybe this gets us off the ground


maxtaco:
i'm just going to forge ahead with this, and we can circle back later. i want to unblock the team. thanks miles!

PR #13224, opened 2018-08-08
add 2 notifications of team-related changes for UI performance

xgess:
This creates two new notifications to the frontend that something potentially interesting has changed with a team or the user's permissions related thereto. 

The first notification is that something in `TeamListUnverified` may have changed, e.g. renamed, someone left, someone joined, logged in user's role has changed, or one of the settings (profile-promote, showcasing, open/closed) has changed. This is not an exhaustive list of interesting team-related changes, but all of the others are already associated with other notifications and result in other major UI changes (e.g. your account is reset, the team has been deleted, ...) or will also throw a message from gregor that will get covered here (i.e. `team.change`/`membership_changed: true` which seems to be used for possibly-too-many things).

The second notification is that something in `CanUserPerform` may have changed, which basically means the user's role for the team is directly (demoted/promoted) or indirectly (e.g. the only other owner has left) different, or it means that something else has changed which we can't distinguish currently based only on notifications (e.g. an unrelated user has joined the team or left the team).


protocol/avdl/keybase1/notify_ctl.avdl
```diff
@@ -26,6 +26,8 @@ protocol notifyCtl {
     boolean deviceclone;
     boolean chatattachments;
     boolean wallet;
+    boolean unverifiedteamlist;
```
mmaxim: Don't think you need these, they can just show up if you choose `team`

xgess: 👍

PR #13225, opened 2018-08-08
check closeOnSelect on mobile

songgao:


songgao:
thanks for the quick review!

PR #13227, opened 2018-08-08
[KBFS-3230] widget chat container

jzila:
I don't want this merged until we have some chat content for the box, which will come in a separate PR.

Ready for feedback though.


shared/menubar/chat.desktop.js
```diff
@@ -0,0 +1,62 @@
...
+      marginTop: Styles.globalMargins.xtiny,
+      paddingBottom: Styles.globalMargins.xtiny,
+      paddingTop: Styles.globalMargins.xtiny,
+    },
+    isElectron: {
+      marginLeft: Styles.globalMargins.tiny,
+      marginRight: Styles.globalMargins.tiny,
+      paddingLeft: Styles.globalMargins.tiny,
+      paddingRight: Styles.globalMargins.tiny,
+    },
```
songgao: Is there a reason why you split them into common and isElectron? It only runs on desktop.

shared/menubar/index.desktop.js
```diff
@@ -277,133 +203,132 @@ class MenubarRender extends Component<Props, State> {
...
+      right: 0,
+      marginLeft: 'auto',
+      marginRight: 'auto',
+      top: -6,
+    },
+    isElectron: {
+      borderLeft: '6px solid transparent',
+      borderRight: '6px solid transparent',
+      borderBottom: `6px solid ${Styles.globalColors.darkBlue2}`,
+    },
```
songgao: Hmm does flow complain if these weren't in isElectron?

shared/menubar/index.desktop.js
```diff
@@ -277,133 +203,132 @@ class MenubarRender extends Component<Props, State> {
...
+    common: {
+      ...Styles.globalStyles.flexBoxColumn,
+      flex: 1,
+      position: 'relative',
+      marginTop: isDarwin ? 13 : 0,
+      borderTopLeftRadius: Styles.globalMargins.xtiny,
+      borderTopRightRadius: Styles.globalMargins.xtiny,
+      backgroundColor: Styles.globalColors.darkBlue,
+    },
+  }),
```
songgao: These and some below only have `common`, so why not just do this instead?

```
widgetContainer: {
    ... ...
},
```

shared/menubar/index.desktop.js
```diff
@@ -80,95 +54,93 @@ class MenubarRender extends Component<Props, State> {
...
         title: 'Settings',
-        view: this._menuView('Settings', 'iconfont-nav-settings', countMap[settingsTab] || 0),
-        onClick: () => this.props.openApp(settingsTab),
+        view: this._menuView('Settings', 'iconfont-nav-settings', countMap[Tabs.settingsTab] || 0),
+        onClick: () => this.props.openApp(Tabs.settingsTab),
       },
       'Divider',
       ...(this.props.loggedIn ? [{title: 'Open main app', onClick: () => this.props.openApp()}] : []),
-      {title: 'Open folders', onClick: () => this.props.openApp(fsTab)},
+      {title: 'Open folders', onClick: () => this.props.openApp(Tabs.fsTab)},
```
songgao: Should this be "Open Files"?

jzila: The design says "Open folders", but I think you're right, it should be "files".

shared/menubar/index.desktop.js
jzila: Yes

shared/menubar/index.desktop.js
jzila: Mainly because of Flow.

shared/menubar/chat.desktop.js
jzila: Yes, Flow.

shared/menubar/index.desktop.js
chrisnojima: If you use the long form it's compatible

jzila: Long form?

chrisnojima: This is the short hand form. If you do borderLeftWidth: 6   BorderStyle: 'solid' borderLeftColor etc they're the same. Rn doesn't support the short hand where you can define multiple props at the same time

chrisnojima: https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties

jzila:
PR feedback implemented, ready for another look.

shared/menubar/index.desktop.js
```diff
@@ -140,7 +140,7 @@ class MenubarRender extends React.Component<Props, State> {
       },
       'Divider',
       ...(this.props.loggedIn ? [{title: 'Open main app', onClick: () => this.props.openApp()}] : []),
-      {title: 'Open folders', onClick: () => this.props.openApp(Tabs.fsTab)},
+      {title: 'Open files', onClick: () => this.props.openApp(Tabs.fsTab)},
```
songgao: I wonder if this should be `Open Files`. `Files` isn't a word but the name of a tab here, so I don't think it'd contradict to Cécile's preference of not capitalize every word. No strong opinion though, so feel free to leave it!

jzila: I'm going to leave it for now. We can always change it later.

PR #13230, opened 2018-08-09
more rough thoughts on how to structure the FastTeamLoader

maxtaco:
cc: @mlsteele, this captures our convo this afternoon.


maxtaco:
See #13232 instead.

PR #13232, opened 2018-08-09
fastloader skeleton and resolveIDToName implemenation

maxtaco:
- wip

### DIFF
--- a/go/teams/fast.go
+++ b/go/teams/fast.go
@@ -1,10 +1,19 @@
 package teams
 
 import (
+	"fmt"
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/keybase1"
 )
+
+//
+// fast.go
+//
+// Routines for fast-team loading. In fast team loading, we ignore most signatures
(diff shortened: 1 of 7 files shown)


go/teams/fast.go
```diff
@@ -23,13 +34,81 @@ type FastTeamChainLoader struct {
...
+	if err != nil {
+		return res, err
+	}
+	downPointer, ok := parentRes.downPointers[parent.Pair.Seqno]
+	if !ok {
+		return res, NewBadNameError("down pointer not found in parent")
+	}
+	suffix := downPointer.NameComponent
+
+	parentName, err := f.verifyTeamNameViaParentLoad(m, parent.Id, isPublic, parentRes.unverifiedName, parentRes.upPointer)
```
mlsteele: Would it be possible to have `load`'s contract be that it only returns verified names?

maxtaco: That would shuffle everything around. I changed the return types upstream so that there's only one name in the return value. `fastLoadRes` has an unverified name. `FastTeamLoadRes` has a verified name.  The former no longer includes the latter.  Does that solve this issue?

PR #13233, opened 2018-08-09
Self provision if cloned device detected

joshblum:
Provides the functionality to 'self provision' if the current device is cloned. Exposed via the cli with `keybase selfprovision <newdevicename>` and can be exposed in the GUI when the `DeviceCloneNotification` is sent (https://github.com/keybase/client/blob/master/go/service/main.go#L664). 

A follow up PR will handle reboxing the userEK during the provision.


go/engine/selfprovision.go
```diff
@@ -0,0 +1,295 @@
...
+		} else if e.User != nil {
+			m.CDebugf("Error in self provision, reverting to original active device: %v", err)
+			m = m.WithGlobalActiveDevice()
+			salt, err := e.User.GetSalt()
+			if err != nil {
+				m.CDebugf("unable to GetSalt: %v", err)
+				return
+			}
+			// Atomically swap to the new config and active device
+			if err := m.SwitchUserNewConfigActiveDevice(uvOld, e.User.GetNormalizedName(), salt,
```
maxtaco: It's a slight bummer we have to explicitly rollback.  We've avoided rollbacks in other flows by just writing the configuration file as the last thing. Is this not possible for this flow?

go/engine/selfprovision.go
```diff
@@ -0,0 +1,295 @@
...
+	if err := e.makeDeviceKeysWithSigner(m, keys.SigningKey()); err != nil {
+		return err
+	}
+
+	// now store the secrets for our new device
+	encKey, err := m.ActiveDevice().EncryptionKey()
+	if err != nil {
+		return err
+	}
+	if err := e.fetchLKS(m, encKey); err != nil {
```
maxtaco: Is it possible to update `TestPassphraseRecover` in systests to see if this stuff is actually working?  The flow would be to make a device, make a paper key, clone the device, self-provision, and then check that passphrase recover works on the self-provisioned device.

maxtaco: Otherwise, the LKSec stuff is pretty easy to get wrong.  I worry that, for instance, the LKSec system hasn't updated for the new device, and is some mishmash of the old device and the new.

go/engine/login_provision.go
```diff
@@ -1063,65 +1063,65 @@ func (e *loginProvision) cleanup(m libkb.MetaContext) {
...
 	}
-	if cr.GetUID().NotEqual(e.arg.User.GetUID()) {
-		m.CDebugf("loginProvision(verify): config uid %q doesn't match engine uid %q", cr.GetUID(), e.arg.User.GetUID())
+	if cr.GetUID().NotEqual(uid) {
+		m.CDebugf("config uid %q doesn't match engine uid %q", cr.GetUID(), uid)
 	}
 
 	// check session.json is valid
-	e.verifyRegularFile(m, "session", m.G().Env.GetSessionFilename())
+	verifyRegularFile(m, "session", m.G().Env.GetSessionFilename())
```
maxtaco: i'm surprised this still works, since i didn't think we still had this file.  maybe get rid of this?

joshblum: the check always fails here actually (we just log the errors) and i wasn't sure why that was the case. i'll drop the check and fail on the errors

go/engine/selfprovision.go
joshblum: sounds good!

go/engine/selfprovision.go
joshblum: i was having some trouble with this so added it explicitly, let me do some digging as to why the transaction on the config isn't working

maxtaco: what if we don't write the config file in DeviceWrap and just do it at the very end, if all worked?  Does that screw anything else up?

go/engine/selfprovision.go
```diff
@@ -0,0 +1,274 @@
...
+	return nil
+}
+
+// makeDeviceKeysWithSigner creates device keys given a signing key.
+func (e *SelfProvisionEngine) makeDeviceKeysWithSigner(m libkb.MetaContext, signer libkb.GenericKey) error {
+	if err := e.ensureLKSec(m); err != nil {
+		return err
+	}
+
+	ss, err := m.ActiveDevice().SyncSecrets(m)
```
joshblum: is there a better way to get the device type?

maxtaco: upak has it, i believe

go/libkb/active_device.go
```diff
@@ -37,13 +39,13 @@ func (a *ActiveDevice) Dump(m MetaContext, prefix string) {
 		m.CDebugf("%sEncKey: %s", prefix, a.encryptionKey.GetKID())
 	}
 	m.CDebugf("%sPassphraseCache: cacheObj=%v; valid=%v", prefix, (a.passphrase != nil), (a.passphrase != nil && a.passphrase.ValidPassphraseStream()))
-	m.CDebugf("%sPaperKeyCache: %v", prefix, (a.paperKey != nil && a.paperKey.DeviceWithKeys() != nil))
+	m.CDebugf("%ProvisioningKeyCache: %v", prefix, (a.provisioningKey != nil && a.provisioningKey.DeviceWithKeys() != nil))
```
maxtaco: buggy string format here.

go/engine/selfprovision.go
```diff
@@ -0,0 +1,260 @@
...
+	// login context
+	m = m.WithNewProvisionalLoginContextForUserVersionAndUsername(uv, e.G().Env.GetUsername())
+
+	// From this point on, if there's an error, we abort the transaction.
+	defer func() {
+		if tx != nil {
+			tx.Abort()
+		}
+		if err == nil {
+			m = m.CommitProvisionalLogin()
```
maxtaco: Isn't this a noop?

joshblum: the assignment is but the call caches the passphrase stream onto the active device so we still want it. fixed and added a comment to clarify

PR #13235, opened 2018-08-09
protect new keyfinder caches properly

mmaxim:


go/chat/keyfinder.go
```diff
@@ -92,22 +92,48 @@ func (k *KeyFinderImpl) writeKey(key string, v *types.NameInfo) {
 	k.keys[key] = v
 }
 
+func (k *KeyFinderImpl) lookupEncKey(key string) (encItem, bool) {
+	k.Lock()
+	defer k.Unlock()
+	existing, ok := k.encKeys[key]
```
joshblum: could just be return here and `lookupDecKey`

PR #13239, opened 2018-08-09
Automated Iconfont Generation

thebearjew:
Previously, our iconfont (`kb.ttf`) generation process was extraordinarily manual and involved uploading and downloading from icomoon.io three times + several manual edits.

This PR completely automates the formatting and iconfont generation.

There are a couple of important and meaningful changes to design and desktop in this PR.

1. A new Zeplin screen has been created with only the iconfont svgs. This makes iconfont generation much simpler and more consistent.
2. All iconfont Sketch assets were updated to follow the naming convention defined here
`{counter}-kb-iconfont-{name}-{size}.svg`
3. Massive shotout to @adamjspooner who basically remade half of the icons so they woudn't output SVG garbage - THANK YOU
4. `yarn-helpers/font.js` has been simplified to generate the iconfile file and update constants at the same time.

**Notes on storybook thrashing**
This approach to iconfont generation aims to reduce shifting codepoints between subsequent font regenerations. Previously, if one icon file was renamed or moved, then all other codepoints would be shifted. Now, the counter at the beginning of the filename `{counter}-kb-iconfont-{name}-{size}.svg` will determine the generated code point. Now, when files are moved and renamed, all other code points should not shift.

**Files to look at**
`icons/iconfont/README.md`
`desktop/yarn-helper/font.js`


thebearjew:
@keybase/react-hackers bumperino

shared/desktop/yarn-helper/font.js
```diff
@@ -23,183 +17,90 @@ const commands = {
...
-        metadata: {
-          fontFamily: 'kb',
-          majorVersion: 1,
-          minorVersion: 0,
+const baseCharCode = 0xe900
+const iconfontRegex = /^(\d+)-kb-iconfont-(.*)-(\d+).svg$/
+const getSvgFilenames = () =>
+  fs
+    .readdirSync(paths.iconfont)
+    .filter(i => i.match(iconfontRegex))
```
chrisnojima: should just check for length here and ditch the check below

shared/desktop/yarn-helper/font.js
```diff
@@ -23,183 +17,90 @@ const commands = {
...
+    // Need to force Node to interpret the counter as a Number so we don't get 1, 10, 100
+    .sort((x, y) => {
+      const matchX = x.match(iconfontRegex)
+      const matchY = y.match(iconfontRegex)
+      if (!matchX || !matchY || matchX.length !== 4 || matchY.length !== 4) return 0
```
chrisnojima: i think this function could be simpler
```
fs
    .readdirSync(paths.iconfont)
    .map(path => {
        const match = path.match(iconfontRegex)
        return (match && match.length === 4) ? {path, score: match[1]} : null
    })
    .filter(Boolean)
    .sort((x, y) => x.score - y.score)
    .map(({path}) => path.resolve(paths.iconfont, path))

```

shared/desktop/yarn-helper/font.js
```diff
@@ -210,66 +111,108 @@ function updatedFonts() {
...
-        charCode,
-        gridSize: info.gridSize,
-        isFont: true,
-      }
-      charCode++
-    })
+  // Build constants for iconfont svgs
+  const svgFiles = getSvgFilenames()
+  console.log(svgFiles)
+  svgFiles.forEach(path => {
```
chrisnojima: could use reduce instead

shared/desktop/yarn-helper/font.js
```diff
@@ -210,66 +111,108 @@ function updatedFonts() {
...
+    const match = path.match(iconfontRegex)
+    if (!match || match.length !== 4) return
+
+    const counter = Number(match[1])
```
chrisnojima: we can be more modern with es6 here
```
const [, counter, name, size] = match
```

shared/desktop/yarn-helper/font.js
```diff
@@ -210,66 +111,108 @@ function updatedFonts() {
...
-  extension?: string,
-  charCode?: number,
-  require?: any,
-}
+  const iconMeta_ = {
+  ${
+    /* eslint-disable */
+    Object.keys(icons)
+      .map(name => {
+        const icon = icons[name]
```
chrisnojima: should modernize this a bit
```
const meta = [
    isFont: `${icon.isFont}`,
    icon.charCode ?  [`charCode: 0x${icons[name].charCode.toString(16)}`] : [],
    icon.extension ? [`extension: '${icons[name].extension}'`)] : [],
    icon.gridSize ? [`gridSize: ${icons[name].gridSize}`] : [],
    icon.require ?  [`require: require(${icons[name].require})`) : [],
]

return `'${name}': {
    ${meta.join(',\n')}
}`

```

chrisnojima:
lets also add to the readme that you have to scroll down the assets list else zeplin won't export all the assets

chrisnojima:
i think we should throw if there's a malformed filename, like if 
1. there's no count
1. grid isn't 24/16/8

we should also print out the gaps so its clear if something weird is going on

chrisnojima:
@adamjspooner can you sanity check the app to see if the icons look good

adamjspooner:
Will do!

adamjspooner:
It looks like the position of some icons is wrong.

Emoji icon is offset (chat, hover row):
screen shot 2018-08-13 at 10 59 47

Magnifying glass is offset (people tab, search box):
screen shot 2018-08-13 at 10 59 59

Ellipsis is offset (all ellipsis buttons):
screen shot 2018-08-13 at 11 00 42
screen shot 2018-08-13 at 11 01 46

Exit sign person and fire are offset (channel/conversation settings):
screen shot 2018-08-13 at 11 01 26
screen shot 2018-08-13 at 11 01 17

Down and up arrows on files sorting are offset:
screen shot 2018-08-13 at 11 02 57

Plus and insert (?) icons are offset (teams tab):
screen shot 2018-08-13 at 11 03 05

Plus icon is offset (devices and git tabs):
screen shot 2018-08-13 at 11 03 40
screen shot 2018-08-13 at 11 11 55

shared/desktop/yarn-helper/font.js
thebearjew: Just wanted to separate the filenames from the paths because they are needed if different contexts. One is for the icon.constants.js and the other is for passing into webfonts-generator.

chrisnojima: in the first map you could return {path, match} so you have both

chrisnojima: and you run the regex once per, not during the sort

shared/desktop/yarn-helper/font.js
thebearjew: Changed this to just pull from `getSvgNames`

thebearjew:
@keybase/react-hackers this is good for another look @adamjspooner

shared/desktop/yarn-helper/font.js
thebearjew: updated below.

adamjspooner:
Those icons are still a few pixels still too low. I know this is nitpicking, though.

thebearjew:
@keybase/react-hackers victory

shared/desktop/yarn-helper/font.js
```diff
@@ -79,11 +79,14 @@ function updatedFonts() {
          * Important: Align to text font baseline
          * Descent: fontHeight * 0.0625
          * source: https://icomoon.io/#docs/font-metrics
+         *
+         * For mysterious reasons, the descent being set to 1/2 the font height
+         * seesm to align the icon with the text baseline as well as making
```
chrisnojima: seems

chrisnojima:
@adamjspooner wanna check again

adamjspooner:
Those downward offsets are still like a pixel down. It's not a dealbreaker though. I'm okay with them being off by a pixel (if we can't fix it).

chrisnojima:
I think this is close enough that we can leave it and pixel push in styling

adamjspooner:
Totally works for me!

thebearjew:
:shipit:

PR #13244, opened 2018-08-09
For discussion: Users common adapter

buoyad:
Adds a new common adapter `Users`. It takes a function as a child, which it gives some generic information about a list of usernames such as their `fullName` and whether you follow them. It also gives a callback to navigate to the profile and a correctly colored / clickable `Text` instance that can be directly returned by `children`.

Also:  
- Replace instances of `ConnectedUsernames` in the people tab follow notifications, which have been particularly difficult to special case in the past (cc @jzila).
- Move `MaybePopup` out of `common-adapters/index` to it's own file.

I wrote this a couple weeks ago but it was giving a very strange flow error then. It works today hence the PR.

r? @keybase/react-hackers

### DIFF
--- a/shared/common-adapters/index.js
+++ b/shared/common-adapters/index.js
@@ -1,68 +1,4 @@
 // @flow
-import * as React from 'react'
-import Box from './box'
-import PopupDialog from './popup-dialog'
-import {connect, type Dispatch} from '../util/container'
-import {collapseStyles, globalColors, isMobile} from '../styles'
-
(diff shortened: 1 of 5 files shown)


buoyad:
(will fix storyshots if we decide we want to use this)

chrisnojima:
i think we should discuss IRL quick. why not just have this be a connected user, then people who use it can just make a fragment with a map and be done?

PR #13246, opened 2018-08-09
bootstrap 3: config and push cleanup

chrisnojima:
- [x] @zanderz please test the pipe ownership stuff, i moved it around


shared/actions/chat2/index.js
```diff
@@ -2214,26 +2213,38 @@ function* handleSeeingExplodingMessages(action: Chat2Gen.HandleSeeingExplodingMe
   })
 }
 
-const loadStaticConfig = (state: TypedState, action: ConfigGen.BootstrapPayload) =>
+const loadStaticConfig = (state: TypedState) =>
```
chrisnojima: get this on handshake

shared/actions/config/index.js
```diff
@@ -1,210 +1,310 @@
...
 import logger from '../../logger'
-import * as KBFSGen from '../kbfs-gen'
-import * as FsGen from '../fs-gen'
-import * as LoginGen from '../login-gen'
 import * as ConfigGen from '../config-gen'
-import * as TeamsGen from '../teams-gen'
-import * as Constants from '../../constants/config'
-import * as GregorCreators from '../gregor'
-import * as NotificationsGen from '../notifications-gen'
+import * as ChatGen from '../chat2-gen'
```
chrisnojima: marking this so its not skipped, likely best to look at this on its own vs a diff

shared/actions/platform-specific/index.desktop.js
```diff
@@ -0,0 +1,229 @@
+// @flow
+import * as ConfigGen from '../config-gen'
```
chrisnojima: moves of this logic moved from ../platform-specific.desktop.js. also moved some things into here (like pipe ownership check) and removed native only stuff

chrisnojima:

zanderz:
The log shows connection failing correctly with a malicious test pipe:
image

For comparison, a fake pipe with good permissions will show `transport using...`
image

But in the non-owner case, it retries over an over in a loop and takes a long time to tell the user something's wrong. I think at one point, bad pipe ownership failed sooner with a more useful message.

shared/constants/push.js
```diff
@@ -1,17 +1,74 @@
...
-  token: '',
+export const makeInitialState: I.RecordFactory<Types._State> = I.Record({
   hasPermissions: true,
+  showPushPrompt: false,
+  token: '',
+  tokenType: null,
+})
+
+export const normalizePush = (n: any): ?Types.PushNotification => {
+  try {
```
buoyad: why the try/catch?

shared/reducers/config.js
```diff
@@ -3,51 +3,107 @@ import logger from '../logger'
...
+          ? state.deleteIn(['daemonHandshakeWaiters', action.payload.name])
+          : state.setIn(['daemonHandshakeWaiters', action.payload.name], newCount)
+
+      if (action.payload.failedFatal) {
+        return newState.merge({
+          daemonHandshakeFailedReason: action.payload.failedReason || '',
+          daemonHandshakeRetriesLeft: 0,
+        })
+      } else {
+        // Keep the first error
```
buoyad: Why?

shared/reducers/config.js
```diff
@@ -3,51 +3,107 @@ import logger from '../logger'
...
+      return state.merge({
+        daemonError: null,
+        daemonHandshakeFailedReason: '',
+        daemonHandshakeRetriesLeft: Constants.maxHandshakeTries,
+        daemonHandshakeState: 'starting',
+      })
+    case ConfigGen.logoutHandshake:
+      return state.merge({logoutHandshakeWaiters: I.Map()})
+    case ConfigGen.daemonHandshake:
+      return state.set('daemonHandshakeState', 'waitingForWaiters')
```
buoyad: you changed a bunch of these to `merge` below, this one too?

shared/actions/config/index.js
```diff
@@ -1,210 +1,329 @@
...
+  // Switch between login or app routes
+  yield Saga.actionToAction([ConfigGen.loggedIn, ConfigGen.loggedOut], switchRouteDef)
+  // Go to the correct starting screen
+  yield Saga.actionToAction(ConfigGen.daemonHandshakeDone, routeToInitialScreen)
+  // If you start logged in we don't get the incoming call from the daemon so we generate our own here
+  yield Saga.actionToAction(ConfigGen.daemonHandshakeDone, emitInitialLoggedIn)
+
+  // Like handshake but in reverse, ask sagas to do stuff before we tell the server to log us out
+  yield Saga.actionToAction(ConfigGen.logout, startLogoutHandshake)
+  // Give time for all waiters to register and allow no waiters
```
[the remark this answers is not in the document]
chrisnojima: updated the comment to be more explicit
  // Give time for all waiters to register and allow the case where there are no waiters
the idea is the flow is we dispatch asking for people to do something and we continue when we get a callback but if nothing dispatches when we're stuck. so we make out own increment/delay/decrement so if tehre is no new waiters (like on desktop) it'll still work

shared/constants/push.js
chrisnojima: because this is very untrusted input, comes basically raw from push and i never want this to crash us out

shared/reducers/config.js
chrisnojima: i think its better to understand the first thing that went wrong to help debug whats going on, other errors are likely a symptom

shared/reducers/config.js
chrisnojima: k

shared/reducers/config.js
buoyad: Sounds good

PR #13248, opened 2018-08-09
get rid of download popup/sheet/drawer on mobile

songgao:
The way save and "share to other app" works on mobile is to download the file first, then trigger the actual save or share action when the download is done. Currently we show an ugly drawer that shows the download progress. This PR gets rid of that, and instead shows a spinner on the menu item.

For "share to other app", we don't hide the menu on tap any more (KBFS-3144), but instead disable the menu row, and show a spinner. When it's done, the menu will go away. If user taps "Cancel" or anywhere outside the menu, the download is canceled.

For "save", we still hide the menu when user taps it. The plan (KBFS-3153) is to show a toast when the save is actually done, but it seems we don't have a mechanism to trigger a toast from Saga for now, so I'm leaving that out for now. Also undone from the ticket is the spinner on the row, but I wanted to get this out first and see how it feels, and maybe talk to @cecileboucheron about what's next.

The way it handles hiding the menu and tracking downloads is kind of messy, partly because I'm trying to mix stuff into the original menu component. But it seems to work well. Let me know if you folks have a better idea on how this should be implemented.

P.S. with this PR, we are finally free from `ModalLessPopupMenu`! (KBFS-3270)

Video demo: https://song.gao.io/files/nYE34IkPIu6XK2yzznFGnIZet8EIKW3KBAqVsjjjCgRCt0Db/


songgao:
⪮

jzila:
Actually let me test it on android first.

songgao:
Thanks John! This'd end up in admin build, so I figured we could just wait it out ..

songgao:
Decided to merge on kbchat. Here we come admin build!

PR #13249, opened 2018-08-10
handle superseded messages properly for incremental mode in GetThreadNonblock CORE-8523

mmaxim:
There is a problem in the `INCREMENTAL` mode of `GetThreadNonblock` where it does not include messages in the remote result that were modified (by a superseder) from the cached result. This affects edits and deletes of messages that we have cached. Patch does the following to fix the situation:

1.) Change `mergeLocalRemoteThread` to include messages from the remote result that have a `SupersededBy` value that is different than what we sent up for the cached result. 
2.) Fix how `resolveHoles` works in `HybridConversationSource`. Previously, it would try to modify the thread that we fetched from storage in-place. However, it didn't do anything for superseded messages, and so the results we got back from a "holey" cache hit were wrong. Now instead of modifying the thread in `resolveHoles`, we just re-fetch from `Storage` if we get a holey hit and resolve it.
3.) Add a new mode for `basicSupersedesTransform` to include a "hidden" `PLACEHOLDER` message instead of just dropping the message for deleted messages. This allows us to communicate to clients of `GetThreadNonblock` that cached messages have actually been deleted. 
4.) **Bonus Fix**: `PullLocalOnly` had previously been using a `SimpleResultCollector` regardless of the query passed in. This made it such that it returned a subset of what `Pull` would when given the same query. Change this to use a `ResultCollector` derived from the query. 

cc @buoyad

### DIFF
--- a/go/chat/convsource.go
+++ b/go/chat/convsource.go
@@ -660,7 +660,7 @@ func (s *HybridConversationSource) identifyTLF(ctx context.Context, conv types.U
 }
 
 func (s *HybridConversationSource) resolveHoles(ctx context.Context, uid gregor1.UID,
-	thread *chat1.ThreadView, conv chat1.Conversation, reason chat1.GetThreadReason) (err error) {
+	thread *chat1.ThreadView, conv chat1.Conversation, reason chat1.GetThreadReason) (holesFilled int, err error) {
 	defer s.Trace(ctx, func() error { return err }, "resolveHoles")()
 	var msgIDs []chat1.MessageID
 	// Gather all placeholder messages so we can go fetch them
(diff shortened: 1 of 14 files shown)


go/chat/convsource.go
```diff
@@ -165,7 +166,7 @@ func (s *baseConversationSource) postProcessThread(ctx context.Context, uid greg
 
 func (s *baseConversationSource) TransformSupersedes(ctx context.Context,
 	unboxInfo types.UnboxConversationInfo, uid gregor1.UID, msgs []chat1.MessageUnboxed) ([]chat1.MessageUnboxed, error) {
-	transform := newBasicSupersedesTransform(s.G())
+	transform := newBasicSupersedesTransform(s.G(), false)
```
joshblum: on any of these calls when you pass unnamed value in to call this can you add `/* useDeletePlaceholders */` so it's easier to read in the future?

PR #13258, opened 2018-08-10
Store unboxing errors in EK caches

joshblum:
Prevents repeated calls to the server for nonexistent (deleted) generations. If you're loading a conversation with a cold chat cache you will ask for the same delete EK for every message you attempt to unbox. This patch:

- Skips unboxing attempt if a message is already expired
- Modifies the caches of `DeviceEKStorage`, `UserEKBoxStorage, and `TeamEKBoxStorage` to also store error state. When we `DeleteExpired` ephemeral keys we also purge invalid cache entries.
- Handles the `team.newly_added_to_team` notification to purge the `TeamEKBoxStorage` caches when entering/exiting a team so if we have any stored errors and join a team we will fetch the box correctly.


joshblum:
note: assuming these changes are approved will have add the new listener method to kbfs for circle ci to pass

joshblum:
- [ ] PR in kbfs
- [ ] verify notif on server side
- [ ] remove call to notify router on `newly_added_to_team`

PR #13265, opened 2018-08-10
Files: two fixes from logsend tickets

songgao:


songgao:
that was quick; thanks!

PR #13267, opened 2018-08-10
fix saveToCameraRoll type arg

songgao:
The `type` arg [is](https://facebook.github.io/react-native/docs/cameraroll) `enum('photo', 'video')`. See the [ticket](https://keybase.atlassian.net/browse/KBFS-3318) for an Android log. Perhaps it only breaks on Android?


songgao:
Good call; just tested in simulator and made sure `saveAttachmentToCameraRoll` was triggered and resulted in the image being saved to camera roll. Although I'm seeing a irrelevant error in console now; will see if I can get it fixed too

songgao:
OK here it is. I also tried on both desktop and mobile to make sure things still centered.

PR #13268, opened 2018-08-11
reverse sort list before playing it back

mmaxim:
A bad bug here, if you are merging in two or more edits into `Storage` of the same message, the **oldest** edit wins!

### DIFF
--- a/go/chat/storage/storage.go
+++ b/go/chat/storage/storage.go
@@ -2,6 +2,7 @@ package storage
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


mmaxim:
I'll add a test later.

mmaxim:
It should already be sorted coming into `MergeHelper`, but it's in the reverse order that `updateAllSupersededBy` needs. I think maybe my last change made this start showing up, but it's probably been in here forever.

PR #13273, opened 2018-08-13
Rebox userEK during self provision

joshblum:
adds `ephemeralKeyReboxer` to encapsulate some of the reboxing logic to use during provisioning and refactors kex2 to use this as well


joshblum:
fixes a chat but where pairwise mac errors are displayed correctly:

screen shot 2018-08-14 at 11 26 22 am

instead of

screen shot 2018-08-14 at 11 02 36 am

PR #13274, opened 2018-08-13
Fix hanging electron on Linux

jzila:


chrisnojima:
This is just for dev?

jzila:
Yeah, it seems to be fine in prod

PR #13276, opened 2018-08-13
Storybook: payment message popup

buoyad:
There's an issue with the native menu layout that makes the header style a bit weird on mobile. ~I'm not sure what to do about it yet~ I made a PR off of this for it #13300, but everything besides that is ready for review. r? @keybase/react-hackers


buoyad:
@keybase/react-hackers bump! the mobile header styling fix involves touching other `FloatingMenu` implementations, i'll make that a PR off of this

shared/chat/conversation/messages/message-popup/payment/index.js
```diff
@@ -0,0 +1,123 @@
+// @flow
+import * as React from 'react'
+import {toUpper, upperFirst} from 'lodash-es'
+import * as Kb from '../../../../../common-adapters'
+import * as S from '../../../../../styles'
```
nathunsmitty: Are we doing `S` now instead of `Styles`? Wow I'm so behind on what's hip already

shared/chat/conversation/messages/message-popup/payment/index.js
```diff
@@ -0,0 +1,123 @@
...
+import type {Position} from '../../../../../common-adapters/relative-popup-hoc'
+
+type HeaderProps = {|
+  amountNominal: string,
+  balanceChange: string, // may be empty
```
nathunsmitty: Is there a reason why you prefer passing empty strings vs having them be optional properties?

shared/chat/conversation/messages/message-popup/payment/index.stories.js
```diff
@@ -0,0 +1,120 @@
...
+  balanceChangeColor: S.globalColors.red,
+  bottomLine: 'stronghold.com',
+  icon: sendIcon,
+  sender: 'cecileb',
+  topLine: 'you sent',
+  txVerb: 'sent',
+}
+
+const load = () => {
+  Sb.storiesOf('Chat/Conversation/Message popup/Payments', module)
```
nathunsmitty: Could you maybe add a container provider with some top and left margin so we can see the overhanging part of the popup?
screen shot 2018-08-15 at 1 20 54 pm

shared/chat/conversation/messages/index.stories.js
```diff
@@ -4,10 +4,10 @@ import placeholder from './placeholder/index.stories'
 import reactButton from './react-button/index.stories'
 import reactionTooltip from './reaction-tooltip/index.stories'
 import text from './text/index.stories'
-import walletPayment from './wallet-payment/index.stories'
+import accountPayment from './account-payment/index.stories'
 
 const load = () => {
-  ;[chooseEmoji, placeholder, reactButton, reactionTooltip, text, walletPayment].forEach(load => load())
+  ;[chooseEmoji, placeholder, reactButton, reactionTooltip, text, accountPayment].forEach(load => load())
```
nathunsmitty: It may be worth changing the name of the story in `account-payment/index.stories.js` from `Chat/Conversation/Wallet payments` to `Chat/Conversation/Account payments` too

shared/chat/conversation/messages/message-popup/payment/index.js
chrisnojima: we should use Styles i think

shared/chat/conversation/messages/message-popup/payment/index.js
buoyad: This'll be connected during integration, so it's likely that the props will always be supplied. For them not to be something like this would have to happen in `mergeProps`:
```
return {
  ...(balanceChange ? {balanceChange} : {}
  <etc>
}
```
which I think is messy. Another benefit is that this way the prop can only have one falsy value rather than two.

shared/chat/conversation/messages/message-popup/payment/index.js
buoyad: Ok :(

shared/chat/conversation/messages/message-popup/payment/index.stories.js
buoyad: The decorator doesn't wrap the `div#modal-root` this ends up in... maybe it should? I'll see if that's hard

buoyad: I added a custom wrapper instead

PR #13283, opened 2018-08-14
stream video on mobile CORE-8463

mmaxim:
The current video flow involves first downloading the video locally, and then playing it back from the file. This has a number of disadvantages, the primary being it takes up a lot of space, and it is isn't even clear that playing from the encrypted file is fast enough to work at all. Patch does the following:

1.) Adds a new function to `attachments.Store`, called `StreamAsset`. The point of the function is to return an `io.ReadSeeker` that can be fed into `http.ServeContent` to run the video playback.
2.) Add a new class, `signencrypt.decodingReadSeeker` which will bring an `io.ReadSeeker` interface to a signencrypt'd payload. It takes an underlying `io.ReadSeeker` source which will feed the ciphertext into it. `Read` will figure out the "chunk" of the stream that we need to fetch from the `source` and decrypt them. 
3.) Add a LRU cache for use with `signencrypt.decodingReadSeeker` that allow us to keep a fixed amount of data in memory for use with repeated requests for the same chunk. Currently we use 20MB. A typical pattern of requests (from `io.Copy`) is for `Read` requests of 32KB. Each of those will hit the same chunk many times, and so this cache is essential.
4.) Add a minimum number of chunks to `Read` on a cache miss so that we don't make too many calls to the underlying source (AWS S3). Right now this is 4MB, but we probably will want to play around with this going forward. 
5.) Change the frontend flow on mobile to play the video on first tap. We also now download the file before `Save` and `Share` do anything. Also add the size to the duration text.


mmaxim:
cc @malgorithms

go/chat/signencrypt/codec.go
```diff
@@ -479,11 +487,58 @@ func NewDecodingReader(encKey SecretboxKey, verifyKey VerifyKey, signaturePrefix
...
+func chunkFromIndex(index int64) (res chunkSpec) {
+	res.index = index
+	res.ptStart = res.index * DefaultPlaintextChunkLength
+	res.ptEnd = res.ptStart + DefaultPlaintextChunkLength
+	res.cipherStart = res.index * getPacketLen(DefaultPlaintextChunkLength)
+	res.cipherEnd = res.cipherStart + getPacketLen(DefaultPlaintextChunkLength)
+	return res
+}
+
+func getChunkAtOffset(plaintextOffset, plaintextLen int64) (res chunkSpec) {
```
patrickxb: is this used?

mmaxim: Hmm, I guess not!

go/chat/signencrypt/seeker.go
```diff
@@ -0,0 +1,177 @@
...
+		if chunkPlaintext, err = decoder.Write(cipherText); err != nil {
+			return n, err
+		}
+		// We might have some straggling data, so just hit Finish here to potentially pick it up. If it
+		// returns an error, then we just ignore it.
+		if finishPlaintext, err := decoder.Finish(); err == nil {
+			chunkPlaintext = append(chunkPlaintext, finishPlaintext...)
+		}
+		r.writeChunksToCache(chunkPlaintext, prefetchChunks)
+	}
```
patrickxb: I wonder if it would also help to fetch more chunks on a cache hit.  i.e. if you prefetched chunks 0, 1, 2, 3 and just hit on chunk 2, then fetch 4, 5 now while reading 2.

mmaxim: The problem there is it makes every cache hit slow. I was thinking about how I could run this in the background, but it was going to get complicated. So the flow is basically this (on Darwin at least):

1.) Player requests the file to get MIME type, we download 4MB. 
2.) Player requests bytes 0-1 for reasons I don't understand, we hit the 4MB chunk from before.
3.) Player requests most of the rest of the file (but probably not all of it). We start churning through the stream as fast as we can in 4MB increments. The cache hits are so fast it doesn't really matter if we start the download on one of them, since we will eventually hit the end of the cache and start up the download while playback is happening.

mmaxim: I found that the ideal number is dependent on the size of the video as well. I tried streaming a 2GB file last night and it didn't really work. For a bigger video I think it makes sense to use a bigger value for this, I will try that this morning.

patrickxb: I was thinking that the fetch (when there's a hit) would be in the background.  Just a thought, but no need whatsoever to pursue that path.

mmaxim: Yeah I was thinking about how to do it in the background, but the whole story seemed to get more complicated that way. Maybe something we can look at in the future though.

PR #13290, opened 2018-08-14
Change font metrics to fix blurriness and offset

thebearjew:
**Updated**

This is the new font generation process 
1. Read all `svg` files from `client/shared/imags/iconfont/*.svg`
2. Combine into an svg font using `webfonts-generator`
3. Convert to True Type Font via `svg2ttf`
4. Modify the ascent/descent as well as the GASP (grid fitting) table in the final TTF output file using [fontforge](https://fontforge.github.io/en-US/)


thebearjew:
@adamjspooner

shared/common-adapters/icon.desktop.js
```diff
@@ -92,7 +92,7 @@ class Icon extends Component<Props, void> {
           fontWeight: 'normal',
           fontVariant: 'normal',
           textTransform: 'none',
-          lineHeight: 1, // NOT 1px, just 1
+          lineHeight: 'initial', // NOT 1px, just 1
```
buoyad: comment still applies?

thebearjew: Does not

PR #13291, opened 2018-08-14
return cached TeamsListUnverfied on network error

joshblum:
Verified manually that the `Build a team!` button doesn't show once we have the cached state

### DIFF
--- a/go/teams/list_fast.go
+++ b/go/teams/list_fast.go
@@ -1,6 +1,8 @@
 package teams
 
 import (
+	"fmt"
+
 	"golang.org/x/net/context"
 
 	"github.com/keybase/client/go/libkb"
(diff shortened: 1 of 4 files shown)


go/teams/list_fast.go
```diff
@@ -39,12 +41,38 @@ func ListTeamsUnverified(ctx context.Context, g *libkb.GlobalContext, arg keybas
...
+		if found, cerr := g.GetKVStore().GetInto(&teams, cacheKey); cerr != nil || !found {
+			// Nothing we can do here.
+			m.CDebugf("| ListTeamsUnverified unable to get cache item: %v, found: %v", cerr, found)
+			return nil, err
+		}
+	default:
 		return nil, err
 	}
 
+	res := &keybase1.AnnotatedTeamList{
```
joshblum: it seemed better to short circuit here so we don't do the UPAK load for nothing: https://github.com/keybase/client/pull/13291/files#diff-c208ef40eba88b78a63656fe1dbb6ecfR81

PR #13292, opened 2018-08-15
play inline video on desktop for small enough videos

mmaxim:
@keybase/react-hackers 

Makes it so if you click a video in a chat thread it plays it in the thread. Go tells the JS whether that is available for any particular video, and will display the image preview if it isn't.


go/chat/utils/utils.go
```diff
@@ -1080,10 +1080,12 @@ func presentAttachmentAssetInfo(ctx context.Context, g *globals.Context, msg cha
 			info.PreviewUrl = g.AttachmentURLSrv.GetURL(ctx, convID, msg.GetMessageID(), true)
 		}
 		atyp, err := asset.Metadata.AssetType()
-		if err == nil && atyp == chat1.AssetMetadataType_VIDEO && asset.Metadata.Video().DurationMs > 1 {
+		if err == nil && atyp == chat1.AssetMetadataType_VIDEO && asset.Metadata.Video().DurationMs > 1 &&
+			strings.HasPrefix(info.MimeType, "video") {
 			info.VideoDuration = new(string)
 			*info.VideoDuration = formatVideoDuration(asset.Metadata.Video().DurationMs) + ", " +
 				formatVideoSize(asset.Size)
+			info.InlineVideoPlayable = asset.Size < (1<<20)*500
```
chrisnojima: i think the shift operator is hard to understand at a glance, like better to just
```
1024 * 1024 * 500
```

shared/chat/conversation/messages/attachment/image/image-render.desktop.js
```diff
@@ -3,17 +3,66 @@ import * as React from 'react'
...
+  playingVideo: boolean
+
+  constructor(props: Props) {
+    super(props)
+    this.videoRef = React.createRef()
+    this.playingVideo = false
+  }
+
+  onVideoClick = () => {
+    if (!this.playingVideo) {
```
chrisnojima: you should likely check this.videoRef.current isn't null

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -28,28 +28,59 @@ type Props = {
...
+    super(props)
+    this.imageRef = React.createRef()
+  }
+  state = {loaded: false, playingVideo: false}
   _setLoaded = () => this.setState({loaded: true})
+
+  _onClick = () => {
+    if (this.props.inlineVideoPlayable) {
+      this.imageRef.current.onVideoClick()
+      this.setState({playingVideo: !this.state.playingVideo})
```
chrisnojima: should use closure form
```
this.setState(p => ({playingVideo: !p.playingVideo}))
```

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -28,28 +28,59 @@ type Props = {
...
+  constructor(props: Props) {
+    super(props)
+    this.imageRef = React.createRef()
+  }
+  state = {loaded: false, playingVideo: false}
   _setLoaded = () => this.setState({loaded: true})
+
+  _onClick = () => {
+    if (this.props.inlineVideoPlayable) {
+      this.imageRef.current.onVideoClick()
```
chrisnojima: check current isn't null

go/chat/utils/utils.go
mmaxim: We use this all over the codebase.

shared/chat/conversation/messages/attachment/image/index.js
mmaxim: Why?

mmaxim: Oh because of `this`?

go/chat/utils/utils.go
chrisnojima: k

shared/chat/conversation/messages/attachment/image/index.js
chrisnojima: no because setState is async so if your goal is the flip the flag you have to use this form

PR #13299, opened 2018-08-15
fix breadcrumb styling

songgao:
Before and after:

screen shot 2018-08-15 at 11 06 22 am

### DIFF
--- a/shared/fs/header/breadcrumb.desktop.js
+++ b/shared/fs/header/breadcrumb.desktop.js
@@ -2,14 +2,7 @@
 import * as React from 'react'
 import * as Types from '../../constants/types/fs'
 import * as Styles from '../../styles'
-import {
-  Avatar,
-  avatarCastPlatformStyles,
-  Box,
-  Icon,
-  Text,
-  iconCastPlatformStyles,
-} from '../../common-adapters'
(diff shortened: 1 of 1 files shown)


songgao:
Please hold off reviewing. Fixing ...

songgao:
OK ready again!

shared/fs/header/breadcrumb.desktop.js
```diff
@@ -11,91 +18,118 @@ type Props = {
...
+    isElectron: {
+      ...Styles.globalStyles.flexBoxRow,
+      left: 0,
+      right: 0,
+      flexGrow: 1,
+      alignItems: 'center',
+      paddingLeft: 16,
+      paddingRight: 16,
+      height: 24,
+    },
```
jzila: which of these require `isElectron`?

nathunsmitty:
awww yiss
download

Although could this maybe be an iconfont thing? cc @thebearjew

songgao:
@nathunsmitty Thanks for the funny comic 😂I don't think iconfont was broken. This is likely from a previous refactor.

songgao:
@nathunsmitty actually I might have jumped the gun. It seems the download cards are broken too. Hmm.

PR #13300, opened 2018-08-15
Floating menu layout tweaks for native

buoyad:
- Remove header padding from floating menu layout and add to implementation locations where necessary
  - This involved some fs popup headers, cc @songgao if you'd like to take a look
- Add padding below the cancel button on iPhone X to give the gesture bar more room
- Allow all message menus icons to hang over the top on iOS
image

I think my grep caught everything but it's possible it didn't. If anyone sees a place I missed lmk! Also no storyshot changes 🎉  r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/info-panel/menu/index.js
+++ b/shared/chat/conversation/info-panel/menu/index.js
@@ -22,7 +22,13 @@ export type Props = {
 }
 
 const Header = ({teamname, memberCount}: {teamname: string, memberCount: number}) => (
-  <Box style={{...globalStyles.flexBoxColumn, alignItems: 'center', paddingTop: 16}}>
+  <Box
+    style={{
+      ...globalStyles.flexBoxColumn,
+      alignItems: 'center',
+      ...(isMobile ? {paddingBottom: 24, paddingTop: 40} : {paddingTop: 16}),
+    }}
+  >
     <Avatar size={isMobile ? 64 : 48} teamname={teamname} style={{marginBottom: isMobile ? 4 : 2}} />
     <Text type="BodySemibold">{teamname}</Text>
     <Text type="BodySmall">{`${memberCount} member${memberCount !== 1 ? 's' : ''}`}</Text>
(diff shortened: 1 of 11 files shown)


shared/chat/conversation/info-panel/menu/index.js
```diff
@@ -22,7 +22,13 @@ export type Props = {
 }
 
 const Header = ({teamname, memberCount}: {teamname: string, memberCount: number}) => (
-  <Box style={{...globalStyles.flexBoxColumn, alignItems: 'center', paddingTop: 16}}>
+  <Box
+    style={{
```
nathunsmitty: Maybe worth refactoring this so it's not inline/uses `platformStyles`?

shared/chat/manage-channels/delete-channel.js
```diff
@@ -1,17 +1,22 @@
...
 import {Text, Box, Icon, FloatingMenu, type OverlayParentProps, OverlayParentHOC} from '../../common-adapters'
-import {globalStyles, globalColors, globalMargins, platformStyles} from '../../styles'
+import {globalStyles, globalColors, globalMargins, isMobile, platformStyles} from '../../styles'
 
 const PopupHeader = ({channelName}: {channelName: string}) => {
   return (
     <Box
       style={{
         ...globalStyles.flexBoxColumn,
+        ...(isMobile
```
nathunsmitty: same here

shared/chat/conversation/messages/message-popup/header.js
```diff
@@ -50,6 +59,8 @@ const MessagePopupHeader = (props: {
         style={{
           marginBottom: globalMargins.tiny,
           marginTop: !isMobile ? globalMargins.small : 0,
+          position: 'absolute',
+          top: isMobile ? (isAndroid ? 0 : -15) : -25,
```
nathunsmitty: also here to get rid of that double ternary?

PR #13301, opened 2018-08-15
Stellar: send integration

cjb:
@keybase/react-hackers 

Here's the integration for sending XLM.  The architecture is:

* The GUI writes its proposed payment to `state.wallets.buildingPayment`.  The service never writes to this record.
* A saga listens for every update to these entries (this is every keypress as you build the payment) and calls the `buildPayment` RPC with the record.
* The service writes its feedback on the payment to `state.wallets.builtPayment`.  The GUI never writes to this record.  The send button stays disabled until the service tells us we have a `readyToSend`.
* Each component (asset input, send form, confirm screen details, etc) is connected to the store itself and puling its data directly out of these records.
* The encrypted `secretNote` is a HiddenString in the store.
* There's a new notification listener that'll reload our payments map when a new one comes in, and a spinner on the confirm screen send button that'll wait for the payment to be accepted.

XLM-only for the moment, Patrick says that's far more important than adding other assets.  Right now the Send to Stellar Address flow is the working one -- though once the search widget is hooked up for Send to Keybase User (DESKTOP-7557), that'll work too.


shared/actions/wallets.js
```diff
@@ -10,6 +10,43 @@ import logger from '../logger'
 import type {TypedState} from '../constants/reducer'
 import {walletsTab} from '../constants/tabs'
 
+const buildPayment = (state: TypedState) =>
+  RPCTypes.localBuildPaymentLocalRpcPromise({
+    amount: state.wallets.get('buildingPayment').get('amount'),
+    // FIXME: Assumes XLM.
+    // currency: state.wallets.get('buildingPayment').get('currency'),
```
patrickxb: `currency` is not outside assets, but something like USD.  See the avdl:

    string amount, // Opt*. Amount of the asset OR outside currency being sent. Example: "3.005"                      
    // Zero or one of `asset` and `currency` should be set.                                                           
    // If both are null or `asset` is native XLM then XLM are being sent.                                             
    // If `currency` is set then XLM is being sent after converting `amount`.                                         
    // Note: Sending non-XLM assets is not yet supported.                                                             
    union { null, OutsideCurrencyCode } currency,                                                                     
    union { null, Asset } asset,

shared/actions/notifications.js
```diff
@@ -46,6 +47,16 @@ const setupEngineListeners = () => {
     'keybase.1.NotifyBadges.badgeState',
     ({badgeState}, _: any, getState) => [NotificationsGen.createReceivedBadgeState({badgeState})]
   )
+
+  getEngine().setIncomingActionCreators(
+    'stellar.1.notify.paymentNotification',
+    ({accountID}, _: any, getState) => [
+      // Need to reload the account fully so that assets update
+      WalletsGen.createLoadAccounts(),
```
chrisnojima: We can keep this short term but longer term I think these things should dispatch a simple single action and the saga should coordinate what side effects should happen

shared/actions/json/wallets.json
```diff
@@ -61,6 +74,37 @@
...
+      "_description": "Set building currency",
+      "currency": "string",
+    },
+    "setBuildingFrom": {
+      "_description": "Set building from",
+      "from": "string",
+    },
+    "setBuildingPublicMemo": {
+      "_description": "Set building public memo",
+      "publicMemo": "string",
```
chrisnojima: Should be hidden. Not strictly necessary but I think we should mask as much as possible

shared/actions/wallets.js
```diff
@@ -10,6 +10,42 @@ import logger from '../logger'
 import type {TypedState} from '../constants/reducer'
 import {walletsTab} from '../constants/tabs'
 
+const buildPayment = (state: TypedState) =>
+  RPCTypes.localBuildPaymentLocalRpcPromise({
+    amount: state.wallets.get('buildingPayment').get('amount'),
+    // FIXME: Assumes XLM.
+    from: state.wallets.selectedAccount,
+    fromSeqno: '',
+    publicMemo: state.wallets.get('buildingPayment').get('publicMemo'),
```
chrisnojima: We don't need these get calls at all. State.wallets.buildPayment.publicMemo

shared/actions/wallets.js
```diff
@@ -10,6 +10,42 @@ import logger from '../logger'
...
+    from: state.wallets.get('buildingPayment').get('from'),
+    fromSeqno: '',
+    publicMemo: state.wallets.get('buildingPayment').get('publicMemo'),
+    quickReturn: false,
+    secretNote: state.wallets.get('buildingPayment').get('secretNote').stringValue(),
+    to: state.wallets.get('buildingPayment').get('to'),
+    toIsAccountID: !!state.wallets.get('builtPayment').get('toUsername'),
+    worthAmount: '',
+  }, Constants.sendPaymentWaitingKey).then(res =>
+    WalletsGen.createClearBuildingPayment()
```
chrisnojima: We shouldn't chain these actions like this. This should return the one action and those should be side effects

shared/actions/wallets.js
```diff
@@ -158,6 +194,13 @@ function* walletsSaga(): Saga.SagaGenerator<any, any> {
     navigateToAccount
   )
   yield Saga.safeTakeEveryPure(WalletsGen.accountsReceived, maybeSelectDefaultAccount)
+  yield Saga.actionToPromise(WalletsGen.setBuildingAmount, buildPayment)
```
chrisnojima: You should have an array of actions not a bunch of single ones. Do [], buildPayment

shared/constants/wallets.js
```diff
@@ -31,6 +55,31 @@ const makeState: I.RecordFactory<Types._State> = I.Record({
   selectedAccount: Types.noAccountID,
 })
 
+const buildPaymentResultToBuiltPayment = (b: RPCTypes.BuildPaymentResLocal) => {
+  const {
+    amountErrMsg,
```
chrisnojima: We shouldn't do spreads like this. Just use b

cjb:
@chrisnojima Thanks! Made all of these changes.

shared/constants/types/wallets.js
```diff
@@ -50,16 +50,38 @@ export type _Assets = {
...
+  publicMemo: HiddenString,
+  recipientType: ?CounterpartyType,
+  secretNote: HiddenString,
+  to: string,
+}
+
+export type _BuiltPayment = {
+  amountErrMsg: string,
+  banners: ?Array<StellarRPCTypes.SendBannerLocal>,
+  publicMemoErrMsg: string,
```
chrisnojima: lets make these hiddenstring

shared/constants/types/wallets.js
```diff
@@ -50,16 +50,38 @@ export type _Assets = {
...
+  readyToSend: boolean,
+  secretNoteErrMsg: string,
```
chrisnojima: lets make these hiddenstring

shared/wallets/confirm-form/container.js
```diff
@@ -1,11 +1,25 @@
 // @flow
 import ConfirmSend from '.'
+import * as Constants from '../../constants/wallets'
+import * as WalletsGen from '../../actions/wallets-gen'
 import {connect, type TypedState, type Dispatch} from '../../util/container'
 
-const mapStateToProps = (state: TypedState) => ({})
+const mapStateToProps = (state: TypedState) => ({
```
chrisnojima: would improve readability to pull out buildPayment
```
=> {
  const p = state.wallets.buildPayment
  return {
    amount: p.amount,
   ...
  }
}
```

shared/wallets/participants/container.js
```diff
@@ -1,14 +1,33 @@
 // @flow
 import Participants from '.'
+import * as WalletsGen from '../../actions/wallets-gen'
 import {compose, connect, setDisplayName, type TypedState, type Dispatch} from '../../util/container'
 
-const mapStateToProps = (state: TypedState) => ({})
+const mapStateToProps = (state: TypedState) => {
+  // Building section
+  const recipientType = state.wallets.buildingPayment.recipientType
```
chrisnojima: likely the same here

PR #13304, opened 2018-08-15
the only way you can get here is via user interaction

mmaxim:


mmaxim:
Oh wait this code runs for both of them...

chrisnojima:
https://github.com/keybase/client/pull/13308

PR #13306, opened 2018-08-15
Stellar integration in profile page

nathunsmitty:
~~This takes some routing code from #13301 so I think we may want to wait for that to get merged before  merging this PR. Nothing super fancy here, mostly just plumbing through a new prop and some routing stuff.~~

~~One thing to note is that the send/receive modal takes you to the wallets tab, which is I think a result of how the routing is set up currently. I'm not sure if this is the desired behavior or if a user should go back to the profile page after their send/request is done.~~

This works as expected and is ready to go!

Note that there are some z-index complications with the profile screen.

screen shot 2018-09-05 at 12 05 37 pm


shared/profile/container.js
```diff
@@ -52,6 +56,7 @@ class ProfileContainer extends React.PureComponent<EitherProps<Props>> {
 
 const mapStateToProps = (state: TypedState, {routeProps, routeState, routePath}: OwnProps) => {
   const myUsername = state.config.username
+  const myAccountID = WalletConstants.getDefaultAccountID(state)
```
buoyad: Can you add a TODO to get rid of this piping after CORE-8785 is done? that allows us to skip explicitly setting the `from` if we intend to send from the default account.

PR #13310, opened 2018-08-15
fix inbox not loading if you had some random items before a load

chrisnojima:
@keybase/react-hackers more explicit state if we've ever loaded the inbox. I had a metamap that had some values in it but never did a load and then it didn't ever load my inbox


chrisnojima:
@keybase/react-hackers bump

PR #13311, opened 2018-08-15
ReactList Storyshoot mock

nathunsmitty:
This gives us a very basic ReactList mock for storyshots. ReactList gives a bunch of scroll errors when it isn't mocked, and our previous mock just returned `null`.

It turns out rendering a bunch of ReactList items can take some time and some of our stories have 100-200+ items, so we clamp all lists at 10 items.

I ran into a couple unique key errors with this, so I'm also looking into those.


nathunsmitty:
Ok the two components the unique key errors are coming from are the Wallet component and EmojiPicker. I'm working on fixing those in this PR, but let me know if I should split that up.

PR #13315, opened 2018-08-16
add FsError and a component for displaying that

songgao:
This is for when we don't have another way to present user an error, but don't want to blackbar. The component is a red strip at the bottom of the Files screen, with a tooltip (desktop only) that shows error details. If the error is retriable, a "Retry" button is shown. User can also click/tap "x" to dismiss an error.

This PR doesn't push any such errors yet. In the next PR I'm gonna catch a few errors in the FS saga into this, so user don't see blackbar for them and can retry.

With dummy errors it looks like this:

screen shot 2018-08-15 at 5 22 31 pm

screen shot 2018-08-15 at 5 22 38 pm

Not from design, so ugliness is all my fault.


jzila:
cc @malgorithms for awareness that we're doing this currently dev-designed. Y'all might want to tell us how to pretty it up afterwards.

shared/fs/footer/errs.js
```diff
@@ -0,0 +1,104 @@
+// @flow
+import * as React from 'react'
+import * as Styles from '../../styles'
+import {Text, Box, Box2, WithTooltip, Icon, iconCastPlatformStyles} from '../../common-adapters'
```
jzila: `import * as Kb from '../../common-adapters'`

songgao:
> Mostly LGTM. Please also make a ticket for mapping which action are retriable.

Hmm I was actually planning to use the same ticket (thus the `-1` suffix).

malgorithms:
cool thanks for the cc @jzila

PR #13319, opened 2018-08-16
[KBFS-3233] Widget Files Container

jzila:
This is the first step in the widget to get Files in. Not ready to merge until subsequent PRs are implemented. But ready for review anyway.


shared/actions/fs/index.js
```diff
@@ -575,21 +575,25 @@ const editSuccess = (res, action, state: TypedState) => {
 const editFailed = (res, {payload: {editID}}) => Saga.put(FsGen.createEditFailed({editID}))
 
 function* openPathItem(action: FsGen.OpenPathItemPayload): Saga.SagaGenerator<any, any> {
-  const {path, routePath} = action.payload
+  const {path, routePath, openDirectly} = action.payload
```
songgao: I think instead of mixing `openDirectly` into the existing action, it might be better to have a new action, e.g. `openPathItemDirectly`, for two reasons:

1) It'd help keep action handlers simple, and we wouldn't need the if-conditions below;

2) This depends on `state.fs.pathItems`. When the action is dispatched from within the Files tab, it's guaranteed we've already loaded the needed data. But if we are going from the widget, we might not have them yet. So we need a way to load those first (probably just with a stat call). It'd become convoluted if we try to add it here.

WDYT?

shared/actions/fs/index.js
```diff
@@ -575,21 +575,25 @@ const editSuccess = (res, action, state: TypedState) => {
...
-    yield Saga.put(
-      putActionIfOnPath(
-        routePath,
-        navigateAppend([
-          {
-            props: {path},
-            selected: 'folder',
-          },
-        ])
+    const navigate = navigateAppend([{
```
songgao: If you decide to make a separate action for this, we can group the `switchTo` with `navigateAppend` here, rather than needing to dispatch multiple actions in the remote container.

shared/menubar/files-container.desktop.js
```diff
@@ -0,0 +1,42 @@
...
+const mapStateToProps = (state) => ({
+  _tlfRows: [
+    {path: FsTypes.stringToPath('/keybase/team/zila.test/abc')},
+    {path: FsTypes.stringToPath('/keybase/team/zila.test/def')},
+  ],
+})
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  _onSelectPath: (path: FsTypes.Path) => {
+    dispatch(ConfigGen.createShowMain())
```
songgao: It seems `ConfigGen.showMain` is created multiple times. Have you considered dispatching a single action here (and below for onViewAll), and in saga of the main window, do something like `Saga.actionToAction([FsGen.selectPath, FsGen.viewAll], ()=>Saga.put(ConfigGen.createShowMain()))`

shared/actions/fs/index.js
jzila: 1. SGTM, I'll make the change.
2. Right now you're correct, but this won't be true going forward, since we'll be making the necessary calls as part of retrieving history.

shared/menubar/files-container.desktop.js
jzila: It's not really though...the creation only happens if the action is dispatched.

songgao: Sorry, by "created multiple times", I meant this was declared multiple times (here are in `onViewAll` below). The alternative also has the benefit that these on* dispatchers only need to dispatch one action.

shared/actions/fs/index.js
songgao: Cool; thanks!

> Right now you're correct, but this won't be true going forward, since we'll be making the necessary calls as part of retrieving history.

Hmm when we retrieve the history, don't we still need to check `pathItems` to know if a thing is `folder`? I guess I don't know the details enough that I might be imagining something up.

shared/menubar/files-container.desktop.js
jzila: I prefer it this way...why send the code reader to another place if they can see here exactly what is done? The suggestion will end up dispatching 4 actions instead of 3.

shared/actions/fs/index.js
jzila: @songgao I was envisioning that the saga that retrieves history ensures that the relevant path items are updated.

shared/menubar/files-container.desktop.js
```diff
@@ -0,0 +1,29 @@
...
+import * as FsTypes from '../constants/types/fs'
+import * as FsGen from '../actions/fs-gen'
+import {FilesPreview} from './files.desktop'
+import {remoteConnect, compose} from '../util/container'
+
+const mapStateToProps = (state) => ({
+  _tlfRows: [
+    {path: FsTypes.stringToPath('/keybase/team/zila.test/abc')},
+    {path: FsTypes.stringToPath('/keybase/team/zila.test/def')},
+  ],
```
songgao: I know this is just placeholders, but it'd cause re-rendering whenever anything in the store state changes. So maybe move it to `mergeProps` for now?

shared/actions/fs/index.js
```diff
@@ -605,16 +604,33 @@ function* openPathItem(action: FsGen.OpenPathItemPayload): Saga.SagaGenerator<an
...
   )
 }
 
+const openFilesFromWidget = (state: TypedState, action: FsGen.OpenFilesFromWidgetPayload) => {
+  const {path} = action.payload
+  const pathItem = path ? state.fs.pathItems.get(path) : undefined
+  let selected = 'preview'
+  if (!pathItem || pathItem.type === 'folder') {
+    selected = 'folder'
+  }
```
songgao: Slight preference for 

```
const selected = state.fs.pathItems.get(path, Constants.unknownPathItem).type === 'folder'
  ? 'folder'
  : 'preview'
```

shared/menubar/files-container.desktop.js
jzila: This is going to go away in the next PR, and I'm not going to merge this one before having that one approved.

jzila: Also I don't think it'll cause re-rendering, since the shallow equal will stay the same.

songgao: > Also I don't think it'll cause re-rendering, since the shallow equal will stay the same.

Why is it the same? `_tlfRows` is a new array every time this runs.

jzila: Oh true. In any case I'll make sure we do it properly in the follow up. Thanks for calling it out!

PR #13320, opened 2018-08-16
Call identify when fetching ephemeral keys

joshblum:
Adds a call to `identify` to future safe getting ephemeral keys. Currently `identify` was being called when we got the standard encryption keys but that may not always be the case.


mmaxim:
Does this slow things down at all?

joshblum:
nothing noticeable from a quick GUI check but we can keep an eye on it

PR #13325, opened 2018-08-16
Darken inbox background.

adamjspooner:
@keybase/react-hackers, this is ready for review. This darkens the inbox's background (and hover states) so we can remove the border between the inbox and the thread.

cc @cecileboucheron


adamjspooner:
Thanks, @jzila! Will do. I won't merge this until Cécile has time to offer some feedback too.

adamjspooner:
I think I'm sold. I've been using this since yesterday and really like it.

cecileboucheron:
Hmm it looks a bit dark and very blue to me, I think a tone somewhere in between and less saturated would work best? We can create a new "lightGreyBlue" tone maybe. So we can keep blue4 as the hover state.

Also one thing, we should re-export the shh and the lock icons with the new colors:
image
image

cecileboucheron:
I like what Jira does for instance:
image

adamjspooner:
Works for me! I just went with the colors we already have.

adamjspooner:
The only problem I foresee is that `blue4` will be too faint for a discernible hover state.

cecileboucheron:
@adamjspooner ^ What do you think?

adamjspooner:
The contrast between the inbox and thread is really nice. I think the inbox hover state is too faint. Otherwise, I'm going to use this today and see if it grows on me.

cecileboucheron:
@adamjspooner I ended up creating a new `blue6` color for the inbox hover. Let me know what you think.

cecileboucheron:
@adamjspooner In the end I'm creating 2 new swatches: greyBlue1 and greyBlue2, it's simpler. I updated the wallet tab with these 2 colors as well (and removed the right border there too). 
We could update the Settings page as well with the same coloring in the left pane some time.

cecileboucheron:
@keybase/react-hackers ok ready for review!

cecileboucheron:
@jzila Mind taking another look?

Issue #13334, opened 2018-08-17
Android background crash in libkb.GlobalContext.Trace


eliasnaur:
I saw this crash in my adb logcat output today:

```
08-17 15:02:31.837  5276  5276 E oid.apps.photo: Not starting debugger since process cannot load the jdwp agent.
08-17 15:02:31.889  5299  5299 E ndroid.contact: Not starting debugger since process cannot load the jdwp agent.
08-17 15:02:31.896  5287  5287 E id.partnersetu: Not starting debugger since process cannot load the jdwp agent.
08-17 15:02:31.897  5294  5294 E ybase.ossifrag: Not starting debugger since process cannot load the jdwp agent.
08-17 15:02:31.926  5315  5315 E le.android.kee: Not starting debugger since process cannot load the jdwp agent.
08-17 15:02:32.366  5299  5401 E ndroid.contact: The String#value field is not present on Android versions >= 6.0
08-17 15:02:32.799  5294     0 E Go      : panic: runtime error: invalid memory address or nil pointer dereference
08-17 15:02:32.799  5294     0 E Go      : [signal SIGSEGV: segmentation violation code=0x1 addr=0x4 pc=0xd5d88c10]
08-17 15:02:32.799  5294     0 E Go      : 
08-17 15:02:32.799  5294     0 E Go      : goroutine 17 [running, locked to thread]:
08-17 15:02:32.799  5294     0 E Go      : github.com/keybase/client/go/libkb.(*GlobalContext).Trace(0x0, 0xd668b604, 0xe, 0xd74b0db8, 0x9173a03c)
08-17 15:02:32.799  5294     0 E Go      : 	/tmp/go-android/src/github.com/keybase/client/go/libkb/util.go:509 +0x14
08-17 15:02:32.799  5294     0 E Go      : github.com/keybase/client/go/bind.BackgroundSync()
08-17 15:02:32.799  5294     0 E Go      : 	/tmp/go-android/src/github.com/keybase/client/go/bind/keybase.go:354 +0x44
08-17 15:02:32.799  5294     0 E Go      : _/var/folders/4g/mh1d9tz91d3f9p5p716dkdd40000gn/T/gomobile-work-849699548/gomobile_bind.proxykeybase__BackgroundSync()
[...]

PR #12814, opened 2018-07-13
Revert "Revert "Renable background jobs on android (#12780)""

joshblum:
Reverts keybase/client#12804, enabling background jobs on android

PR #13336, opened 2018-08-17
wait for Init in BackgroundSync for Android bkg task startup

mmaxim:
For Android background task wakeup, there is a race between `BackgroundSync` and `Init` it seems. This could cause the app to crash if we tried to `Trace` using a `kbCtx` that hasn't been set to anything yet. Add a new function `waitForInit` that will poll the status of `Init` for `maxTime` duration before failing out or returning. I chose polling since I didn't want to have to mess with a lock or channel and risk a different kind of crash (closing a channel twice or something). 

cc @joshblum

### DIFF
--- a/go/bind/keybase.go
+++ b/go/bind/keybase.go
@@ -516,7 +516,7 @@ func AppWillExit(pusher PushNotifier) {
 // [iOS] returning true will request about ~3mins from iOS to continue execution
 func AppDidEnterBackground() bool {
 	if !isInited() {
-		return
+		return false
 	}
 	defer kbCtx.Trace("AppDidEnterBackground", func() error { return nil })()
 	ctx := context.Background()
--- a/go/bind/keybase.go
+++ b/go/bind/keybase.go
@@ -354,18 +354,30 @@ func Version() string {
 }
 
 func SetAppStateForeground() {
+	if !isInited() {
+		return
+	}
 	defer kbCtx.Trace("SetAppStateForeground", func() error { return nil })()
 	kbCtx.AppState.Update(keybase1.AppState_FOREGROUND)
 }
 func SetAppStateBackground() {
+	if !isInited() {
+		return
+	}
 	defer kbCtx.Trace("SetAppStateBackground", func() error { return nil })()
 	kbCtx.AppState.Update(keybase1.AppState_BACKGROUND)
 }
 func SetAppStateInactive() {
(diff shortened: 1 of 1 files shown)


joshblum:
i wonder why it was crashing here since the latest version have the background sync disabled

mmaxim:
? It looks on to me, and my experience with my phone indicates it is running as well.

joshblum:
we've had it disabled for 2-3 releases so we could zero in on the crasher, https://github.com/keybase/client/pull/12814 reenables it

mmaxim:
Maybe these things stick around. I notice the jobs need to be > 1 to get canceled. So maybe once schedule it just stays there?

joshblum:
ugh yeah that was supposed to be a > 0 i guess so we cancel anything leftover

mmaxim:
Yeah the stack track trace from https://github.com/keybase/client/issues/13334 is from 2.5.1, and it clearly has `BackgroundSync` on the stack, so it is running on his phone.

go/bind/keybase.go
```diff
@@ -350,7 +370,28 @@ func SetAppStateBackgroundActive() {
...
+	if isInited() {
+		return nil
+	}
+	for {
+		select {
+		case <-time.After(200 * time.Millisecond):
+			if isInited() {
+				return nil
+			}
+		case <-time.After(maxDur):
```
maxtaco: Classic bug. This'll never win since you'll keep pushing this off into the future!  You need to make `expireCh := time.After(maxDur)` outside of the loop, and then check `case <-expireCh:`

joshblum:
yeah i guess never stopped running on anyone who had the job initially scheduled and we just didn't make new ones

go/bind/keybase.go
mmaxim: Ugh, nice catch!

maxtaco:
I don't want to bikeshed, but is there an alternate solution where you wait for a channel close, that's closed on init?  Is the problem that you can't be certain that this will be created before you start listening for it to close? Don't feel strongly on this.

mmaxim:
Well I was a little worried since iOS calls `Init` and Android calls `InitOnce`, and I didn't want to have to think about preventing any kind of channel close from happening twice. Might not be a rational fear.

maxtaco:
ok, agreed! channels are like impossible to get right.

PR #13338, opened 2018-08-17
Don't allow isMobile styles to override isAndroid/isIOS

buoyad:
I checked all the statically defined styles, only one needed changing.

- [x] Check `platformStyles` defined inline


buoyad:
@keybase/react-hackers This is ready!

PR #13341, opened 2018-08-17
Add spellchecking.

adamjspooner:
@keybase/react-hackers, this is intended to start a discussion around adding spellchecking. Thoughts?

screen shot 2018-08-17 at 11 19 30

cc @malgorithms


chrisnojima:
we need to discuss

PR #13348, opened 2018-08-17
play inline video on mobile CORE-8609

mmaxim:
@keybase/react-hackers 

This patch makes tapping a video in a chat thread start to play it inline in the thread on mobile, like on desktop. Each inline video is actually a `WebView` control that sources a webpage served by the service. Users can access the fullscreen option on iOS by using the control on the HTML5 viewer, but unfortunately Android users cannot (the fullscreen control doesn't work in the WebView there) (this is also how it works now on Android, it just launches a fullscreen view to play a small video). 

Note this patch includes a fork of the `react-native-wkwebview-reborn` library to allow us to play inline video in the `WKWebView` control on iOS. Let me know if I didn't set that up right.


mmaxim:
This is the change on the webview library https://github.com/keybase/react-native-wkwebview/commit/8812e8cfec2a6959d1df0869b4112769f493c62e

shared/chat/conversation/messages/attachment/image/image-render.native.js
```diff
@@ -1,11 +1,69 @@
 // @flow
 import * as React from 'react'
+import {WebView} from 'react-native'
+import WKWebView from 'react-native-wkwebview-reborn'
 import {NativeImage, NativeDimensions} from '../../../../../common-adapters/native-wrappers.native'
```
chrisnojima: WebView is exported as NativeWebView in native-wrappers.native. outside of common-adapters we shouldn't really import react-native directly

shared/chat/conversation/messages/attachment/image/image-render.native.js
```diff
@@ -1,11 +1,69 @@
...
+  constructor(props: Props) {
+    super(props)
+    this.playingVideo = false
+  }
+
+  onVideoClick = () => {
+    if (!this.webview) {
+      return
+    }
+    const arg = !this.playingVideo ? 'play' : 'pause'
```
chrisnojima: prefer
`const arg = this.playingVideo ? 'pause' : 'play'`

shared/chat/conversation/messages/attachment/image/image-render.native.js
```diff
@@ -1,11 +1,69 @@
...
+    const js = `togglePlay("${arg}")`
+    ;(isIOS ? this.webview.evaluateJavaScript : this.webview.injectJavaScript)(js)
```
chrisnojima: prettier is making this look strange, prefer
```
const runJS = isIOS ? this.webview.evaluateJavasscript : this.webview.injectJavascript
runJS(`togglePlay("${arg}")`)
```

shared/chat/conversation/messages/attachment/image/image-render.native.js
```diff
@@ -1,11 +1,69 @@
...
+    this.playingVideo = !this.playingVideo
+  }
+
+  render() {
+    if (this.props.inlineVideoPlayable && this.props.videoSrc.length > 0) {
+      const uri = this.props.videoSrc + '&poster=' + encodeURIComponent(this.props.src)
+      const source = {uri}
```
chrisnojima: prefer
```
const source = {
  uri: `${this.props.videoSrc}&poster=${encodeURIComponent(this.props.src)}`
}
```

shared/chat/conversation/messages/attachment/image/image-render.native.js
```diff
@@ -1,11 +1,69 @@
...
+      const allLoads = () => {
```
chrisnojima: should just be a method not made inline each time

shared/chat/conversation/messages/attachment/image/image-render.native.js
```diff
@@ -1,11 +1,69 @@
...
+        this.props.onLoad()
+        this.props.onLoadedVideo()
+      }
+      return isIOS ? (
+        <WKWebView
+          ref={ref => {
```
chrisnojima: prefer createRef

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -102,7 +98,9 @@ class ImageAttachment extends React.PureComponent<Props, State> {
         >
           {!!this.props.path && (
             <ImageRender
-              ref={this.imageRef}
+              ref={ref => {
```
chrisnojima: why remove createRef?

mmaxim: It wasn't working on native.

mmaxim: Can't overwrite read-only property or something.

mmaxim:
We talked about the last point in chat, and I'd like to use this on my trip so I'm going to merge. If there is anything else you want to change I can follow up.

PR #13349, opened 2018-08-17
Team chain tool

mlsteele:
Play a team chain back from a [json file](https://github.com/keybase/keybase/pull/2790) and dump the state. Probably doesn't work for subteams yet.

Proof of concept for running as a server barnacle.


zapu:
fascinating. so what's the plan here? could this work as a golang rpc that server could call to validate team chain before committing db changes?

mlsteele:
That's the idea. And perhaps used to fill/backfill additional derived info. Such as whether a link modifies the high set. But probably going to do that a different way. So this will go on the back burner.

PR #13351, opened 2018-08-17
Listen to subteam rename

joshblum:
Based off of https://github.com/keybase/client/pull/13346, so i'll rebase this once that goes in. Depends on the server changes in https://github.com/keybase/keybase/pull/2788 for tests

- Listens to gregor notifications on SubteamRenames and reloads the conversation to get the update. For each conversation that was updated we send a single notification to the frontend so we can support sending updates for multiple topic types.

- Adds the `LookupName` method to the `NameInfoSource` interface since we cannot `LookupID` (via the name) when the subteam name is changing. It might be better to have a  single `LookupNameInfo` method that takes in both and changes the resolution based on memberstype, but i added this as a first stab for correctness.

- The frontend still needs to refresh the conversations based on the gregor notification (and clear the old values out of their store). Even after a DB nuke some old values of the subteam names are shown (unless a new message is added to the thread), there is a bug currently with how the name is calculated.


go/chat/sender.go
```diff
@@ -199,15 +199,28 @@ func (s *BlockingSender) checkConvID(ctx context.Context, conv chat1.Conversatio
...
+		// verify it matches what is on the message itself. If we rename a
+		// subteam the names between the current and reference message will
+		// differ so we cannot rely on that.
+		switch conv.GetMembersType() {
+		case chat1.ConversationMembersType_TEAM:
+			if info, err := CreateNameInfoSource(ctx, s.G(), conv.GetMembersType()).LookupName(ctx,
+				conv.Metadata.IdTriple.Tlfid,
+				conv.Metadata.Visibility == keybase1.TLFVisibility_PUBLIC); err != nil {
+				return err
+			} else if info.CanonicalName != headerQ.TlfName {
```
mmaxim: How does this check work when the rename happens?

joshblum: this depends on the frontend listening to the gregor notif and busting their cache to get the new conversation info loaded since we expect the new `TlfName` on `headerQ`. maybe we should just force this value so we don't have to worry about a race of a pending send and subteam rename coming in?

mmaxim: Nah its ok

go/chat/push.go
```diff
@@ -1010,6 +1010,50 @@ func (g *PushHandler) SetConvSettings(ctx context.Context, m gregor.OutOfBandMes
...
+		}, nil)
+		if err != nil {
+			g.Debug(ctx, "resolve: unable to read conversation: %s", err.Error())
+			return
+		}
+		if len(inbox.Convs) != len(update.ConvIDs) {
+			g.Debug(ctx, "resolve: unable to find all conversations")
+		}
+
+		for _, conv := range inbox.Convs {
```
mmaxim: Why not just send up a list?

joshblum: we could but we'd have to separate each batch to a single topic type so i thought this would be easier

mmaxim: Think we should send up a list, isn't there some common code for topic type stuff? I just worry when the UI gets like a stream of little updates. It usually results in a bunch of little RPC calls into the service which can hurt perf.

PR #13353, opened 2018-08-17
handle errors through FsError and FS saga cleanup

songgao:
With this, we should be showing the red banner from https://github.com/keybase/client/pull/13315 for FS errors most of time.

Also while testing this, I'm noticing the upload banner is still kind of flaky, especially for small files. I'll try to fix it when I get to KBFS-3325.


shared/reducers/fs.js
```diff
@@ -204,11 +203,37 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
-    case FsGen.editFailed:
-      // $FlowFixMe
-      return state.setIn(['edits', action.payload.editID, 'status'], 'failed')
     case FsGen.fsError:
-      return state.setIn(['errors', Constants.makeUUID()], action.payload.error)
+      const {erroredAction, error} = action.payload.error
+      logger.error('error (fs)', erroredAction.type, error)
+      const nextState: Types.State = state.setIn(['errors', Constants.makeUUID()], action.payload.error)
+
+      switch (erroredAction.type) {
```
songgao: @chrisnojima I ended up moving the error stuff for download, upload, and edits here, so saga don't need to worry about what component need information about errors. It also don't require dispatching multiple actions from a single saga action handler. Please let me know if you think there's a better way to do this.

shared/actions/fs/index.js
```diff
@@ -354,7 +355,8 @@ const getWaitDuration = (endEstimate: ?number, lower: number, upper: number): nu
 }
 
 let polling = false
-function* pollSyncStatusUntilDone(): Saga.SagaGenerator<any, any> {
+function* pollSyncStatusUntilDone(action: FsGen.FsActivityPayload): Saga.SagaGenerator<any, any> {
+  console.log('pollllllll')
```
jzila: :fire:

shared/actions/fs/index.js
```diff
@@ -516,13 +523,18 @@ function* _loadMimeType(path: Types.Path, refreshTag?: Types.RefreshTag) {
...
+function* loadMimeType(action: FsGen.MimeTypeLoadPayload) {
+  try {
+    yield Saga.call(_loadMimeType, action.payload.path, action.payload.refreshTag)
+  } catch (error) {
+    yield Saga.put(makeUnretriableErrorHandler(action)(error))
+  }
+}
 
-const commitEdit = (action: FsGen.CommitEditPayload, state: TypedState) => {
+const commitEdit = (state: TypedState, action: FsGen.CommitEditPayload) => {
```
jzila: Since we always return a promise now, you can type the return type here.

chrisnojima: again, not really needed. flow will tell you when it needs to be typed

shared/actions/fs/index.js
songgao: Ah; I only realized in my follow-up PR, so removed it there: https://github.com/keybase/client/pull/13391

songgao: Actually I'll remove it here so it's never in master. Thanks

shared/actions/fs/index.js
songgao: This is used in `actionToPromise` below, so it's not vague to humans. @jzila I'm gonna keep it this way unless you have a strong opinion here.

songgao:
Bump 🕺🏿

PR #13366, opened 2018-08-19
Refactor actions/git

MarcoPolo:
Moves actions/git to a the new coding style. Preferring `actionToAction` and `actionToPromise` over generators.

Adds a git sub reducer, and a couple more git actions to help move away from the generators.

Adds a couple tests for the actions. Just as a start, but happy to add more. (a lot of the additions come from this)

More notes inline.

(also it doesn't seem like I have write access to keybase/client. I can move this so it runs CI tests after I get access)

### DIFF
--- a/shared/actions/__tests__/devices.test.js
+++ b/shared/actions/__tests__/devices.test.js
@@ -267,7 +267,7 @@ describe('navs after revoking', () => {
   it('root of devices on revoke other', () => {
     const {dispatch, getState} = init
     const deviceID = Types.stringToDeviceID('456')
-    dispatch(DevicesGen.createRevoked({deviceID, deviceName: 'a phone', wasCurrentDevice: true}))
+    dispatch(DevicesGen.createRevoked({deviceID, deviceName: 'a phone', wasCurrentDevice: false}))
     expect(getRoute(getState)).toEqual(I.List([Tabs.devicesTab]))
   })
 
--- a/shared/actions/git.js
+++ b/shared/actions/git.js
@@ -2,9 +2,7 @@
 import * as ConfigGen from './config-gen'
 import * as Constants from '../constants/git'
 import * as GitGen from './git-gen'
-import * as WaitingGen from './waiting-gen'
 import * as NotificationsGen from './notifications-gen'
-import * as Entities from './entities'
 import * as I from 'immutable'
 import * as RPCTypes from '../constants/types/rpc-gen'
 import * as RouteTreeTypes from '../constants/types/route-tree'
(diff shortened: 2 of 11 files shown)


shared/actions/__tests__/devices.test.js
```diff
@@ -267,7 +267,7 @@ describe('navs after revoking', () => {
   it('root of devices on revoke other', () => {
     const {dispatch, getState} = init
     const deviceID = Types.stringToDeviceID('456')
-    dispatch(DevicesGen.createRevoked({deviceID, deviceName: 'a phone', wasCurrentDevice: true}))
+    dispatch(DevicesGen.createRevoked({deviceID, deviceName: 'a phone', wasCurrentDevice: false}))
```
MarcoPolo: Just a side fix because this test is for revoke other

shared/actions/git.js
```diff
@@ -2,211 +2,108 @@
...
+    .then((results: ?Array<RPCTypes.GitRepoResult>) =>
+      GitGen.createLoaded(Constants.parseRepos(results || []))
+    )
+    .catch(() => {})
+
+const surfaceGlobalErrors = (_, {payload: {errors}}: GitGen.LoadedPayload) =>
+  Saga.all(errors.map(globalError => Saga.put(ConfigGen.createGlobalError({globalError}))))
+
+// Do we have to call loadgit?
+const createPersonalRepo = (_, action: GitGen.CreatePersonalRepoPayload) =>
```
MarcoPolo: The old code called `loadGit` after it finished the rpc message. It's isn't necessary because the component calls `loadGit` on mount.

If we do want this behavior anyways we could have this dispatch an action like `reposModified` that loadsGit and Navs

shared/actions/git.js
```diff
@@ -234,45 +131,72 @@ function _handleIncomingGregor(action: GitGen.HandleIncomingGregorPayload) {
...
+  return Saga.all(actions)
 }
 
-function _processNavigateToTeamRepo(results: any[]) {
-  const {teamname, repoID} = (results[1]: GitGen.NavigateToTeamRepoPayload).payload
-  const idToInfo: I.Map<string, Types.GitInfo> = results[2]
+const isRepoInfoFresh = (lastLoad: ?number, maxAgeMs: number) => lastLoad && lastLoad - Date.now() < maxAgeMs
+
+// Note: Needs to be sequential (& possibly refire) in order to match the repoID + teamname to a unique id
+const navigateToTeamRepo = (state: TypedState, action: GitGen.NavigateToTeamRepoPayload) => {
```
MarcoPolo: This was a little tricky since it requires data from another action. I went with calling the other action and refiring this action when the data is available. If there's a better solution lmk!

MarcoPolo: I guess another solution would be for the expandedSet to be either the unique `id` or `{repoId: string, teamname: string}`, then derive the unique id in the connector. I'm not convinced that's better, but it would avoid having to do the tricky sequential put/take and refire. Thoughts?

PR #13372, opened 2018-08-19
Switch to rn-fetch-blob

chrisnojima:
Prep for updating to RN 0.57, switch from react-native-fetch-blob to rn-fetch-blob


chrisnojima:
this is on top of the flow stuff so going to wait till thats in

PR #13375, opened 2018-08-19
Hook up provision:switchToGPGSignOnly

MarcoPolo:
`provision.switchToGPGSignOK` is for this case:
```
// If there was an error importing a gpg key into the local
// keyring, tell the user and offer to switch to GPG signing
// with this key.  Return true to switch to GPG signing,
// false to abort provisioning.
```

I do this by reintroducing the gpg choose screen, but removing the import option and showing what the error was. It looks like this:
screen shot 2018-08-19 at 3 42 12 pm

While working on this I found these two issues:
[CORE-8616](https://keybase.atlassian.net/browse/CORE-8616)
[DESKTOP-7583](https://keybase.atlassian.net/browse/DESKTOP-7583)

### DIFF
--- a/shared/actions/provision.js
+++ b/shared/actions/provision.js
@@ -27,6 +27,7 @@ type ValidCallback =
   | 'keybase.1.provisionUi.ProvisionerSuccess'
   | 'keybase.1.provisionUi.chooseDevice'
   | 'keybase.1.provisionUi.chooseGPGMethod'
+  | 'keybase.1.provisionUi.switchToGPGSignOK'
   | 'keybase.1.secretUi.getPassphrase'
 
 const cancelOnCallback = (_: any, response: CommonResponseHandler, __: any) => {
@@ -195,6 +196,7 @@ class ProvisioningManager {
     this._stashResponse('keybase.1.provisionUi.chooseGPGMethod', response)
     return Saga.put(ProvisionGen.createShowGPGPage())
   }
+
   submitGPGMethod = (state: TypedState, action: ProvisionGen.SubmitGPGMethodPayload) => {
     // local error, ignore
     if (state.provision.error.stringValue()) {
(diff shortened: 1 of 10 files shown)


shared/actions/provision.js
```diff
@@ -213,6 +215,27 @@ class ProvisioningManager {
...
 
+  switchToGPGSignOKHandler = (
+    params: RPCTypes.ProvisionUiSwitchToGPGSignOKRpcParam,
+    response: CommonResponseHandler,
+    state: TypedState
+  ) => {
+    this._stashResponse('keybase.1.provisionUi.switchToGPGSignOK', response)
+    return Saga.all([
+      Saga.put(ProvisionGen.createSwitchToGPGSignOnly({importError: params.importError})),
+      Saga.put(ProvisionGen.createShowGPGPage()),
```
chrisnojima: we should just make this a param to the existing showGPGPage action. See the other show* actions which also take payloads. Motivation is the flag will only ever get set by this and its only in regards to this one screen

shared/provision/gpg-sign/index.desktop.js
```diff
@@ -12,43 +12,63 @@ class GPGSign extends Component<Props> {
     return (
       <Container style={styles.container} onBack={() => this.props.onBack()}>
         <RowCSS />
+        {this.props.importError && (
```
chrisnojima: can you explain what happens in this flow. Does this error happen immediately before this screen, or we see this screen, then they click this top button , then they get this error? If so would it be simpler to just keep the original screen and just put an error below the button  (essentially just plumbing through the error) and not do any other kind of state or switching. they can click again if they fixed it, or they'll get the same error again, or they can choose the second option. then the 2 different paths for the mergeProps in the container can go away

PR #13385, opened 2018-08-20
use canonicalized team name

joshblum:
ensures team names are always canonicalized to lowercase when working with them. previously an error of the form:
`▶ ERROR server conversation TLF name mismatch: blah, expected bLaH` could be thrown when performing this check: https://github.com/keybase/client/compare/joshblum/tlfname-CORE-8588?expand=1#diff-560008cd2bdd22d95d33f0194c0cf8e7R198


joshblum:
non-canonical comes from user input, so this https://github.com/keybase/client/pull/13385/files#diff-4fe4d9cec17e3f5479ea802ac589e6c2R185 is really what fixes the problem. `TeamNameFromString` already does the lower casing as it builds parts: https://github.com/keybase/client/pull/13385/files#diff-77a916effb7c2a53cc526b18d3c74361R1985 so the `ToLower` call in `TeamName.String` isn't necessary but i thought would be safer to add incase a team name was manually built in some way and didn't get the conversion in `TeamNameFromString`

PR #13386, opened 2018-08-20
fastloadforchat helper

heronhaye:


protocol/avdl/keybase1/teams.avdl
```diff
@@ -664,6 +664,13 @@ protocol teams {
     boolean allowNameLookupBurstCache; // If it's ok to hit the nameLookup burst cache
   }
 
+  record ChatInfo {
```
maxtaco: this is a lot like `FastTeamLoadRes`.  Can we combine them?

PR #13393, opened 2018-08-21
apicall command fixes and improvements

zapu:
Fixed `apicall` output. Added `--url` flag so people who are used to using API urls in curl or browser can feel right at home. 

`apicall` command should be a good alternative for debugging in lockdown mode.

Example use:
```
» keybase apicall --url "https://keybase.io/_/api/1.0/team/get.json?name=oh_god_the_ai"
▶ INFO Parsed URL as endpoint: "team/get", args: [{Key:name Value:oh_god_the_ai}]
{"box":{"ctext":"1qlj2VYMRjFXTT11HK7Iigb6ww+kC5m3yjHvm34rMWTCc9bcOyi6hEzw3h4wdQwH","generation":2,"nonce":"xHJO2e0aD60m+S9w5 ...💇snip
```

This command was also moved from `devel` to `common` so it's included in prod builds. But because it doesn't define `Usage` field, it's not visible in command list when someone types `keybase` or `keybase help`.

### DIFF
--- a/go/client/cmd_apicall.go
+++ b/go/client/cmd_apicall.go
@@ -91,7 +91,7 @@ func NewCmdAPICall(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Comma
 
 func (c *CmdAPICall) Run() error {
 	if c.parsedHost != "" {
-		if c.parsedHost != c.G().Env.GetServerURI() && c.parsedHost != "https://keybase.io" {
+		if !strings.EqualFold(c.parsedHost, c.G().Env.GetServerURI()) {
 			return fmt.Errorf("Unexpected host in URL mode: %s. This only works for Keybase API.", c.parsedHost)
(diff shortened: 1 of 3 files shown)


go/client/cmd_apicall.go
```diff
@@ -77,11 +81,22 @@ func NewCmdAPICall(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Comma
...
+				Name:  "url",
+				Usage: "Pass full keybase.io URL with query parameters instead of an endpoint.",
+			},
 		},
 	}
 }
 
 func (c *CmdAPICall) Run() error {
+	if c.parsedHost != "" {
+		if c.parsedHost != c.G().Env.GetServerURI() && c.parsedHost != "https://keybase.io" {
```
joshblum: should this be an `||`?

PR #13394, opened 2018-08-21
Devices list fix

cecileboucheron:
@keybase/react-hackers

### DIFF
--- a/shared/devices/row/index.js
+++ b/shared/devices/row/index.js
@@ -39,7 +39,7 @@ const DeviceRow = (props: Props) => {
       onClick={props.showExistingDevicePage}
       icon={<Kb.Icon type={icon} style={Kb.iconCastPlatformStyles(props.isRevoked ? styles.icon : null)} />}
       body={
-        <Kb.Box2 direction="vertical">
+        <Kb.Box2 direction="vertical" style={{justifyContent: 'center', minHeight: 48}}>
           <Kb.Text style={props.isRevoked ? styles.text : null} type="BodySemiboldItalic">
             {props.name}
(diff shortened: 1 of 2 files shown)


shared/devices/row/index.js
```diff
@@ -16,25 +16,23 @@ const DeviceRow = (props: Props) => {
...
-      icon = props.isCurrentDevice ? 'icon-phone-success-48' : 'icon-phone-48'
+      icon = props.isCurrentDevice ? 'icon-phone-success-32' : 'icon-phone-32'
       break
     default:
-      /*::
       declare var ifFlowErrorsHereItsCauseYouDidntHandleAllTypesAbove: (type: empty) => any
-      ifFlowErrorsHereItsCauseYouDidntHandleAllTypesAbove(props.type);
-      */
+      ifFlowErrorsHereItsCauseYouDidntHandleAllTypesAbove(props.type)
       icon = 'icon-paper-key-48'
```
chrisnojima: Why did this change

shared/devices/row/index.js
```diff
@@ -39,7 +39,7 @@ const DeviceRow = (props: Props) => {
       onClick={props.showExistingDevicePage}
       icon={<Kb.Icon type={icon} style={Kb.iconCastPlatformStyles(props.isRevoked ? styles.icon : null)} />}
       body={
-        <Kb.Box2 direction="vertical">
+        <Kb.Box2 direction="vertical" style={{justifyContent: 'center', minHeight: 48}}>
```
cecileboucheron: @chrisnojima This works - though ideally we should not need the justifyContent

PR #13396, opened 2018-08-21
Animate info panel.

adamjspooner:
@keybase/react-hackers, this is ready for review. It adds an animation for sliding on the info panel in a conversation. I'm not sure how to add an animation for sliding it off.


adamjspooner:
@keybase/react-hackers, bump.

chrisnojima:
lets not check in flow-typed files that are entire stubs. it stubs out a ton of files we never need so lets just remove it.
you can also remove this line from package.json
```
    "build-flow-stubs": "flow-typed create-stub",
```
just add the line to flow-interface.js.flow instead

adamjspooner:

adamjspooner:
@chrisnojima, how does this look?

shared/common-adapters/animated.js.flow
```diff
@@ -0,0 +1,8 @@
+// @flow
+import * as React from 'react'
+export type Props = {|
+  children?: Function,
```
chrisnojima: should be something tighter like
```
children: (styles: CrossPlatformStyles) => React.Node
```

shared/chat/conversation/info-panel/container.js
```diff
@@ -156,24 +157,33 @@ class InfoPanelSelector extends React.PureComponent<Props> {
     return isMobile ? (
       <ConnectedInfoPanel onBack={this.props.onBack} conversationIDKey={this.props.conversationIDKey} />
     ) : (
-      <Box onClick={this.props.onBack} style={clickCatcherStyle}>
-        <Box style={panelContainerStyle} onClick={evt => evt.stopPropagation()}>
-          <ConnectedInfoPanel onBack={this.props.onBack} conversationIDKey={this.props.conversationIDKey} />
-        </Box>
-      </Box>
+      <Kb.Box onClick={this.props.onBack} style={clickCatcherStyle}>
```
chrisnojima: this only shows desktop, we should have some example of mobile working also

adamjspooner: I wasn't sure what trivial thing to animate on mobile.

shared/common-adapters/animated.js.flow
adamjspooner: It's different between desktop and native.

PR #13402, opened 2018-08-21
ForceDeleteAll deviceEKs on self provision [CORE-8664]

joshblum:


go/ephemeral/selfprovision_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestEphemeralSelfProvision(t *testing.T) {
+	tc, user := ephemeralKeyTestSetup(t)
+	defer tc.Cleanup()
+
+	ctx := context.TODO()
+	g := tc.G
+	m := libkb.NewMetaContextForTest(tc)
```
zapu: Super minor, but i'd probably do `ctx := m.Ctx()` here to keep it all in one context. On the other hand none of my tests did that, but I feel inspired now

go/ephemeral/selfprovision_test.go
```diff
@@ -0,0 +1,73 @@
...
+		SecretUI:    secUI,
+		LoginUI:     provLoginUI,
+	}
+
+	m = m.WithUIs(uis)
+	libkb.CreateClonedDevice(tc, m)
+	newName := "uncloneme"
+	eng := engine.NewSelfProvisionEngine(g, newName)
+	err = engine.RunEngine2(m, eng)
+	require.NoError(t, err)
```
zapu: should we check that current device is indeed `"uncloneme"` ?

joshblum: yes i'll add this, the engine tests cover that but can't hurt to check here as well

PR #13403, opened 2018-08-21
break require loops

chrisnojima:
This breaks require loops which RN 0.57 points out in the console.warnings now. Loops are allowed but lead to very subtle behavior sometimes so we should avoid it if possible

Also deprecates local-debug-live which we don't use anymore


protocol/bin/flow.js
```diff
@@ -472,9 +472,9 @@ function write(typeDefs, project) {
 
 // This file is auto-generated by client/protocol/Makefile.
 // Not enabled: calls need to be turned on in enabled-calls.json
-import engine from '../../engine'
```
chrisnojima: avoid pulling in engine and engine saga, which pull in some rpc types themselves

shared/chat/inbox/row/chat-filter-row/container.js
```diff
@@ -1,18 +1,18 @@
 // @flow
 import * as Constants from '../../../../constants/chat2'
-import * as Inbox from '../..'
 import * as Types from '../../../../constants/types/chat2'
 import * as Chat2Gen from '../../../../actions/chat2-gen'
 import {isDarwin} from '../../../../constants/platform'
 import {connect, compose, setDisplayName, withProps} from '../../../../util/container'
 import type {TypedState, Dispatch} from '../../../../util/container'
+import type {RowItem, RowItemSmall, RowItemBig} from '../../index.types'
```
chrisnojima: we have a file for just the types already

shared/common-adapters/back-button.native.js
```diff
@@ -2,7 +2,9 @@
 
 import React, {Component} from 'react'
 import {NativeTouchableWithoutFeedback} from './native-wrappers.native'
-import {Badge, Box, Icon, iconCastPlatformStyles} from '.'
+import Badge from './badge'
```
chrisnojima: never import from common-adapters/index as that pulls in everything, including this file

shared/constants/chat2/index.js
```diff
@@ -2,7 +2,6 @@
 import * as I from 'immutable'
 import * as Types from '../types/chat2'
 import * as RPCChatTypes from '../types/rpc-chat-gen'
-import * as Constants from '../../constants/chat2'
```
chrisnojima: Constants... is this file 🔥

chrisnojima:
- [ ] fix jest engine init

PR #13405, opened 2018-08-21
Stellar: chat integration

buoyad:
Some notes:

- The chat messages contain pointers to the relevant request / payment. The accounts infrastructure is used to load & store that data
  - Added loading states to the messages & popup
  - `requestPayment` messages contain the `note` in the message payload 
- Added `paymentIDIsEqual` to check equality between `PaymentID`s, which the service expects frontend to treat as opaque. cc @zapu / @mlsteele 
- The request send button isn't hooked up to the send dialog, I split that to a separate ticket.
- Cancelling requests doesn't currently delete the message, this is pending core support
- Loading payments isn't currently possible with just the information in the chat message, so you have to visit the wallets tab first to load the payment before the message will appear. Payments sent before around last Friday will infini-spinner.

r? @keybase/react-hackers


shared/chat/conversation/messages/message-popup/index.js
```diff
@@ -4,46 +4,88 @@ import * as Types from '../../../../constants/types/chat2'
 import AttachmentMessage from './attachment/container'
 import TextMessage from './text/container'
 import ExplodingMessage from './exploding/container'
+import PaymentMessage from './payment/container'
 import type {Position} from '../../../../common-adapters/relative-popup-hoc'
 
 type Props = {
   attachTo: ?React.Component<any, any>,
-  message: Types.MessageText | Types.MessageAttachment,
+  message:
```
chrisnojima: maybe we should make a type in type for this

shared/chat/conversation/messages/wrapper/wrapper-author/container.js
```diff
@@ -12,7 +12,11 @@ import {isMobile} from '../../../../../constants/platform'
 type OwnProps = {|
   isEditing: boolean,
   measure: null | (() => void),
-  message: Types.MessageText | Types.MessageAttachment,
+  message:
```
chrisnojima: same type could be used here

shared/chat/conversation/messages/wrapper/wrapper-author/container.js
```diff
@@ -62,7 +77,11 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps) => {
   const sequentialUserMessages =
     previous &&
     previous.author === message.author &&
-    (previous.type === 'text' || previous.type === 'deleted' || previous.type === 'attachment')
+    (previous.type === 'text' ||
```
chrisnojima: could move this into a helper in constants

shared/chat/conversation/messages/wrapper/wrapper-timestamp/container.js
```diff
@@ -14,7 +14,14 @@ export type OwnProps = {|
   previous: ?Types.Message,
 |}
 
-const decoratedMessageTypes: Array<Types.MessageType> = ['attachment', 'text', 'systemLeft']
+const showAuthorMessageTypes = ['attachment', 'requestPayment', 'sendPayment', 'text']
+const decoratedMessageTypes: Array<Types.MessageType> = [
```
chrisnojima: could go into constants

PR #13406, opened 2018-08-21
Change how DeviceWrap works w.r.t device switching, cleanup selfprovisioning

zapu:
- We are removing `ConfigWriterTransacter` from `SelfProvisionEngine` because its semantics did not do us any favors, and the code seemed like it does something else than it actually did.
    - `Abort`ing `ConfigWriterTransacter` does not mean the config state will roll back to from before transaction - so this was already confusing. It only meant the config file was not changed. But it would have been changed during next modification, automatically, so this gave us nothing.
    - Fixing that turned out to be difficult mainly because how `UserConfig` is used to cache variables related to currently selected user.
- We already have means of atomically swapping ActiveDevice and saving config: `DeviceWrap.SetActiveDevice`, which calls `mctx.SwitchUserNewConfigActiveDevice` (which calls `mctx.switchUserNewConfig`). It does all that while holding `switchUserMu` lock.
     - Assuming that there are no other device-related config writes during self provisioning, this achieves exactly what previous code wanted to achieve - be able to cancel and retain previous state at any point before "commit" is called.
- So we are also changing `DeviceWrap`  to make `SetActiveDevice` user callable, so it's obvious at which point the switch actually happens. All `DeviceWrap` consumers has been changed according to this new behaviour.
- There is one place - `SetDeviceCloneState` - which sets multiple config variables and wants to do so in transaction. For that we added `Rollback` function in `jsonFileTransaction` which re-reads the config file, hopefully restoring its state to from before the transaction. Note that it won't work with `UserConfig`, but `SetDeviceCloneState` writes to device clone state which is entirely different file (`device_clone.json`).


go/engine/device_wrap.go
```diff
@@ -122,8 +122,8 @@ func (e *DeviceWrap) refreshMe(m libkb.MetaContext) (err error) {
 	return err
 }
 
-func (e *DeviceWrap) setActiveDevice(m libkb.MetaContext) (err error) {
-	defer m.CTrace("DeviceWrap#setActiveDevice", func() error { return err })()
+func (e *DeviceWrap) SetActiveDevice(m libkb.MetaContext) (err error) {
```
joshblum: can you rename to `CommitNewUserConfigAndActiveDevice`? a comment here about when/why this method should be called separately from `Run` would be good too

joshblum: can you rename to `CommitNewUserConfigAndActiveDevice`? a comment here about when/why this method should be called separately from `Run` would be good too

go/engine/paperprovision.go
```diff
@@ -222,11 +222,14 @@ func (e *PaperProvisionEngine) makeDeviceWrapArgs(m libkb.MetaContext) (*DeviceW
 	}, nil
 }
 
-// copied from loginProvision
-// makeDeviceKeys uses DeviceWrap to generate device keys.
+// copied from loginProvision makeDeviceKeys uses DeviceWrap to
```
joshblum: nit: this should be two separate lines or put a period between them since it's two separate thoughts.

go/engine/selfprovision.go
```diff
@@ -95,19 +88,16 @@ func (e *SelfProvisionEngine) Run(m libkb.MetaContext) (err error) {
...
 	// Make new device keys and sign them with current device keys
-	if err = e.provision(m, keys); err != nil {
+	if err := e.provision(m, keys); err != nil {
 		return err
 	}
 
-	// commit the config changes
-	// NOTE we should wrap the Active device change in a transaction as well.
-	if err := tx.Commit(); err != nil {
+	// Finish provisoning by calling SetActiveDevice which sets
```
joshblum: can you put a similar comment for the other callers? just to be explicit so anyone looking in the future understands the reasoning. maybe add something like `This is not revertable, so anything else that fails the provisioning process is best effort since we have commit to this new device`

go/engine/paperprovision.go
```diff
@@ -222,11 +222,14 @@ func (e *PaperProvisionEngine) makeDeviceWrapArgs(m libkb.MetaContext) (*DeviceW
...
+// generate device keys and sets active device.
 func (e *PaperProvisionEngine) makeDeviceKeys(m libkb.MetaContext, args *DeviceWrapArgs) error {
 	eng := NewDeviceWrap(m.G(), args)
-	return RunEngine2(m, eng)
+	if err := RunEngine2(m, eng); err != nil {
+		return err
+	}
+	return eng.SetActiveDevice(m)
```
joshblum: can you move this into the `Run` method as in the self provision engine?

go/engine/signup.go
```diff
@@ -230,6 +230,9 @@ func (s *SignupEngine) registerDevice(m libkb.MetaContext, deviceName string) er
 	if err := RunEngine2(m, eng); err != nil {
 		return err
 	}
+	if err := eng.SetActiveDevice(m); err != nil {
+		return err
+	}
```
joshblum: sorry for another nit since i know this isn't your code, but can you change the signature of `storeSecret` below to not return an error? the function already just warns if it fails and we shouldn't have any errors once we commit to the new active device

go/engine/selfprovision.go
```diff
@@ -95,26 +88,26 @@ func (e *SelfProvisionEngine) Run(m libkb.MetaContext) (err error) {
...
-
 	// Now we can cleanup old deviceEKs and store the new deviceEK, encrypted
 	// by the new globally set active device.
 	if deviceEKStorage := m.G().GetDeviceEKStorage(); deviceEKStorage != nil {
 		if err = deviceEKStorage.ForceDeleteAll(m.Ctx(), e.User.GetNormalizedName()); err != nil {
 			m.CDebugf("unable to remove old ephemeral keys: %v", err)
 		}
 	}
+
+	// Now we can store and encrypt the new deviceEK with the new globally set
```
joshblum: dup comment same above

go/engine/paperprovision.go
```diff
@@ -121,15 +114,12 @@ func (e *PaperProvisionEngine) Run(m libkb.MetaContext) (err error) {
 		return err
 	}
 
-	// commit the config changes
-	if err := tx.Commit(); err != nil {
+	// Finish provisoning by calling SwitchConfigAndActiveDevice. we
```
joshblum: can you copy this to `login_provision` engine and `signup` engine?

zapu:
I've manually tested the provisioning scenarios and also had debugging output in ConfigWriter to ensure only one `save` happened in each of them - so entire config change was contained. The following scenarios were tested:
- Logging in to account without keys.
- Logging in by provisioning through another device (cli mode).
- Selfprovisioning.
- Provisioning using paper key (via `keybase login`).
- Provisioning with synced secret PGP key.

All of these methods call:
```
github.com/keybase/client/go/libkb.(*JSONFile).Save(0xc4200d6690, 0x0, 0x0)
	/home/zapu/go/src/github.com/keybase/client/go/libkb/json.go:150 +0x95
```
exactly once, so if we make sure this is the last thing that happens during provisioning, we can be safe that no one will end up being "half provisioned", or set up in config as someone they can't be (say, because something ultimately failed while creating a device).

In this PR, this is the case in all code paths, except for one:
```
func (e *loginProvision) tryGPG(m libkb.MetaContext) (err error)
```
will call `makeDeviceKeysWithSigner` which leads to config change through `DeviceWrap.SwitchConfigAndActiveDevice`; but then `tryGPG` has a failing path if `libkb.WriteLksSKBToKeyring` fails. I don't know how serious this is. Probably ActiveDevice and config has already been changed, so user sees an error but ends up being provisioned anyway. Which is probably a desirable state to be in.

zapu:
@patrickxb can you take a look? It was supposed to be a simple change in self provisioning but we discovered that config transactions might not have been doing what they seemed, so we went through the code and changed / simplified things.

go/libkb/json.go
```diff
@@ -306,6 +306,17 @@ func (f *jsonFileTransaction) Abort() error {
 	return err
 }
 
+// Rollback reloads config from unchanged config file, bringing its
+// state back to from before the transaction changes. Note that it
+// only works for changes that do not affect UserConfig, which caches
+// values, and has to be reloaded manually.
+func (f *jsonFileTransaction) Rollback() error {
```
patrickxb: does this work if there was no config file before?

zapu: fixed in latest commit, can you take a look again? thank you!

zapu:
@patrickxb what do you think about this one? let me know if there are still doubts - it's a tricky change.

patrickxb:
Remind me: what are we trying to fix here?

go/libkb/json_test.go
```diff
@@ -144,3 +145,78 @@ func TestJsonSetAndGetNull(t *testing.T) {
 	isRet = reader.GetNullAtPath(path)
 	require.True(tc.T, isRet)
 }
+
+func TestJsonTxRollback(t *testing.T) {
```
patrickxb: Did this fail with the existing system?

zapu: yes - we started with similar test when @joshblum and I were first investigating this issue. JSONFile would retain changed data after `tx.Abort()` was called.

patrickxb: As I said in chat, that doesn't really matter since logout happens after any provisioning error.

PR #13413, opened 2018-08-22
fix order of default merge

chrisnojima:
@keybase/react-hackers this actually fixes the nav issue. turns out navigateUp is overridden by the default merge. i mistakenly made the default mergePros which we have to pass do 
`(s,d,o) => ({...s, ...d, ...o})`
vs the actual default
`(s,d,o) => ({...o, ...s, ...d})`

added a comment to nav since thats way too subtle

### DIFF
--- a/shared/app/global-errors/container.js
+++ b/shared/app/global-errors/container.js
@@ -35,5 +35,5 @@ const mapDispatchToProps = dispatch => ({
   },
 })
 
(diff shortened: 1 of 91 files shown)


chrisnojima:
(nav issue is if you swipe left to go back from thread to inbox you can no longer scroll inbox)

chrisnojima:
the typing of that function is whats hard, better for flow to infer it locally

PR #13414, opened 2018-08-22
stellar: Hook up account creation

cjb:
@keybase/react-hackers CC @patrickxb 

We already had an `<EnterName />` component inside `link-existing`, so this adds a new container reusing that display component.  The action itself happens in the style of link-existing, too.


cjb:
@buoyad Thanks!  Made both changes.

PR #13424, opened 2018-08-22
Update typing animation.

adamjspooner:


shared/chat/conversation/input-area/normal/shared.js
```diff
@@ -1,9 +1,54 @@
 // @flow
 import * as React from 'react'
-import {Meta} from '../../../../common-adapters'
-import {globalColors, platformStyles, styleSheetCreate} from '../../../../styles'
+import * as I from 'immutable'
+import * as Kb from '../../../../common-adapters'
+import * as Styles from '../../../../styles'
 import {formatDurationShort} from '../../../../util/timestamp'
 
+export const isTyping = (typing: I.Set<string>) => (
```
chrisnojima: instead of making a helper that returns a component we should just make this a component. aka
```
const IsTyping = ({typing: I.Set<string>}) => <Kb.Text>...
```

chrisnojima: then you can just do 
```
<IsTyping typing={this.props.typing} /> above
```

shared/chat/conversation/input-area/normal/shared.js
```diff
@@ -1,9 +1,54 @@
...
+  <Kb.Text type={Styles.isMobile ? 'BodyTiny' : 'BodySmall'} style={styles.isTypingText}>
+    {typing.size > 0 && (
+      <Kb.Box style={styles.typingIconContainer}>
```
chrisnojima: you can't put a box inside text on mobile

shared/chat/conversation/input-area/normal/shared.js
```diff
@@ -1,9 +1,54 @@
...
+        <Kb.Icon type="icon-typing-24" style={Kb.iconCastPlatformStyles(styles.typingIcon)} />
+      </Kb.Box>
+    )}
+    {typingNames(typing)}
+  </Kb.Text>
+)
+
+export const typingNames = (typing: I.Set<string>) => {
```
chrisnojima: same here, you should be able to use fragments instead of arrays also

adamjspooner:

shared/chat/conversation/input-area/normal/platform-input.native.js
```diff
@@ -174,6 +181,10 @@ class PlatformInput extends Component<PlatformInputProps & OverlayParentProps, S
             visible={this.props.showingMenu}
           />
         )}
+        <IsTyping
+          style={collapseStyles([styles.typing, {opacity: this.props.typing.size > 0 ? 1 : 0}])}
```
chrisnojima: likely want to just do
```
collapseStyles([styles.typing, this.props.typing.size && styles.visibleTyping}])
```

and make styles.typing have opacity: 0 and visibleTyping have opacity:1

PR #13428, opened 2018-08-22
split photo/video upload on android and more cleanups

songgao:


shared/fs/common/path-item-action-container.js
```diff
@@ -37,8 +37,8 @@ const mapDispatchToProps = (dispatch: Dispatch, {path}: OwnProps) => ({
   copyPath: () => dispatch(ConfigGen.createCopyToClipboard({text: Types.pathToString(path)})),
   ...(isMobile
     ? {
-        _saveMedia: () => dispatch(FsGen.createSaveMedia({path})),
-        _shareNative: () => dispatch(FsGen.createShareNative({path})),
+        _saveMedia: () => dispatch(Constants.createSaveMedia({path})),
```
jzila: I'd rather these be `FsGen.createSaveMedia({path, key: Constants.makeDownloadKey(path)})`, to make explicitly clear what the action is named.

songgao: hmm why is using Constants not clear? They are named the same way. I was hoping to make it easier for container since all we really need is `path`.

songgao: Hmm actually I'll try to make a helper function so we can still use FsGen here.

shared/actions/fs/index.js
```diff
@@ -227,8 +210,11 @@ function* monitorDownloadProgress(key: string, opID: RPCTypes.OpID) {
   }
 }
 
-function* download(action: FsGen.DownloadPayload): Saga.SagaGenerator<any, any> {
-  const {path, intent, key} = action.payload
+function* download(
+  action: FsGen.DownloadPayload | FsGen.ShareNativePayload | FsGen.SaveMediaPayload
+): Saga.SagaGenerator<any, any> {
+  const {path, key} = action.payload
```
jzila: @songgao why not just generate the key here from the path?

songgao: `key` is derived from the `path`, and I don't want containers to have to be concerned about it when creating this action. It's really just for internal tracking. Does the latest solution look good to you?

jzila: I guess you need it for the reducer too. Yep the current solution LGTM.

The reason I wanted the `FsGen` creates to be referenced is that you can then assume without looking at the code what the action is named. If it's just in `Constants`, it _could_ have a different name (even if there's a convention to have it be the same).

PR #13430, opened 2018-08-23
Don't switch convo just because we received a new meta

cjb:
@keybase/react-hackers 

Coyne was hitting the following:

* start a new convo with someone you've already talked to

* you are now in a pendingMode convo

* receive a message from anyone

* the convo you just received a message in becomes selected instead of the pending one

This was because `metasReceived` leads to `changeSelectedConversation` in our sagas.  I can't think why this should be the case -- under which circumstance would you want a changed inbox row to select a new convo automatically?

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -1453,7 +1453,6 @@ const changeSelectedConversation = (
 
 const _maybeAutoselectNewestConversation = (
   action:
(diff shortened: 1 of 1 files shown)


cjb:
Switched over to allowing metasReceived to fire off a maybeAutoselectConversation, but only doing so if we're coming from the empty conversation.

PR #13431, opened 2018-08-23
Fix icon clipping on mobile

thebearjew:
I went back to the drawing board w/ this one after realizing that the font metrics intended to resolve anti-aliasing were causing icons to clip on mobile.

After trying to emulate the exact values from the previous font generation, I've come to the conclusion that it is much simpler to generate a font that is pixel aligned, and then resolve the baseline issue using css (`line-height` and `vertical-align` on desktop).

On mobile, the font metrics determine the vertical layout and anti-aliasing. I've tested the font metrics on 5 ios phone versions and 3 android versions (+ physical device. There has not been any anti-aliasing issues and alignment is correct.

Additionally this PR makes some changes to to `fs` styling to update the fixes I did in https://github.com/keybase/client/pull/13290.

@adamjspooner you should expect a maximum of 2px difference in icon positioning on desktop. Please let me know if there is anything worse.


shared/chat/conversation/messages/react-button/index.js
```diff
@@ -204,7 +204,7 @@ export class NewReactionButton extends React.Component<NewReactionButtonProps, N
                 style={iconCastPlatformStyles(
                   Styles.collapseStyles([
                     styles.emojiIconWrapper,
-                    !Styles.isMobile && (this.props.showBorder ? {top: 3} : {top: 1}),
+                    !Styles.isMobile && (this.props.showBorder ? {top: 3} : {top: 0}),
```
buoyad: This looks a little too high next to the message menu button. maybe 1 should stay?

thebearjew:
Going to abandon this one in favor of a different approach I've taken (yet again)

PR #13434, opened 2018-08-23
Add `expectedUID` field to Kex2Provisionee, check it during handleHello

zapu:
When we are provisioning, we always know which user we want to provision as. Make use of that knowledge in Kex2Provisionee, by adding `expectedUID` field. If we get `Hello` message from UID we don't recognize, abort the process.

### DIFF
--- a/go/engine/device_add_test.go
+++ b/go/engine/device_add_test.go
@@ -25,7 +25,8 @@ func TestDeviceAddPUK(t *testing.T) {
 	testDeviceAdd(t, true)
 }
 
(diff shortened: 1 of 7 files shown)


zapu:
@patrickxb can you take a look? thanks!

patrickxb:
Will this work with a pre-expectedUID client?  i.e. if provisionee has this patch, but provisioner doesn't (like if linux goes out with this patch before other platforms).

zapu:
Nothing in the protocol has changed, so I don't see why wouldn't it. Or am I missing something?

PR #13437, opened 2018-08-23
Fix failure to rotate bug

joshblum:
Fixes a bug where a writer of a subteam was unable to perform a rotate when they were not a member of the parent team. We fix this by making the `Admin` section `nil` if we cannot get the permission. This section is required for implicit admin to rotate so it cannot be removed completely.

### DIFF
--- a/go/teams/teams.go
+++ b/go/teams/teams.go
@@ -387,7 +387,7 @@ func (t *Team) ApplicationKeyAtGeneration(ctx context.Context,
 	return ApplicationKeyAtGeneration(t.MetaContext(ctx), t.Data, application, generation)
 }
 
-func (t *Team) Rotate(ctx context.Context) error {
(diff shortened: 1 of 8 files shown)


go/teams/teams.go
```diff
@@ -397,9 +397,11 @@ func (t *Team) Rotate(ctx context.Context) error {
 	// load an empty member set (no membership changes)
 	memSet := newMemberSet()
 
-	admin, err := t.getAdminPermission(ctx, false)
+	// Try to get the admin perms if they are available, if not, proceed anyway
+	var admin *SCTeamAdmin
+	admin, err = t.getAdminPermission(ctx)
 	if err != nil {
-		return err
+		admin = nil
```
mlsteele: Log the error for the next person to debug this. Maybe they will try to combine SBS, subteams, and writer rotates and it will just be too much.

PR #13440, opened 2018-08-24
Add local cache for TeamData of parent chain during subteam loading

zapu:


go/teams/loader.go
```diff
@@ -368,6 +368,10 @@ func (l *TeamLoader) load2(ctx context.Context, arg load2ArgT) (ret *load2ResT,
 	if len(arg.reason) > 0 {
 		traceLabel = traceLabel + " '" + arg.reason + "'"
 	}
+
+	fmt.Printf("(%s calls) TeamLoader::load2 %v, reason: %q, repoll: %t, readSubteamID: %+v\n",
```
joshblum: rm before merge

go/systests/subteams_test.go
```diff
@@ -0,0 +1,49 @@
...
+	bob.leave(subTeamName.String())
+
+	teams.RotateKey(context.TODO(), al.tc.G, *subteamID)
+	teams.RotateKey(context.TODO(), al.tc.G, *subteamID)
+
+	t.Logf("Eve loads team ...")
+	fmt.Printf(":: Eve loads team %s ...\n", subteamID.String())
+
+	eve.loadTeam(subTeamName.String(), false)
+}
```
zapu: note: this test doe snot actually test anything but I used it to count how many times each team is loaded (see printf in `teams/loader.go`)

zapu:
@mlsteele can you take a look? it's related to CORE-8684

zapu:
Thanks, I think issuing a load but making sure it only repolls once is better. I'll go with that.

zapu:
Hm, after trying to go in and rewrite this, I realized that full-blown TeamData caching like we do in this PR has the advantage of short-circuiting `walkUpToAdmin` entirely when `admin.target` team is already in `parentCache`. 

`repollCache` would be much cleaner IMO, but it would still have to walk the chain every time while checking implicit adminships, potentially invoking lots of loads even (but with `forceRepoll: false`).

Also we had an issue in `teams/list.go` before, where we relied on team loader because we assumed subsequent multiple cached loads of same team would be fast, but it turned out adding another cache in there sped it up considerably as well. That being said, I'm starting to be convinced that short-lived, contained (within one "call", in this case load call for a subteam) caches.

@mlsteele let me know if any of the above changes your opinion. I know you were originally fine with either one, but I want us both to be convinced that this is not code smell.

PR #13442, opened 2018-08-24
High skips for all user chains on the client

xgess:
#### For team sigchains
* add a `high_skip` node to the inner section of new team chain links that get sent to the server en route to other members. this uses the high skip values that https://github.com/keybase/client/pull/13387 set on the TeamSigchainState
* also hoist those two values up into the outer section of those team chain links
* add some test assertions that, after posting team chain links to the server, those chain links have the expected high_skip seqnos.

#### For user sigchains
* the same stuff. add high_skip to the inner and outer links. 
* hide it all behind a feature flag

### DIFF
--- a/go/libkb/chain_link.go
+++ b/go/libkb/chain_link.go
@@ -1445,7 +1445,7 @@ func (c ChainLink) AllowStubbing() bool {
 // IsHighUserLink determines whether a chainlink counts as "high" in a user's chain,
 // which is defined as an Eldest link, a link with seqno=1, a link that is Sibkey,
 // PGPUpdate, Revoke, or any link that is revoking.
-func (c ChainLink) IsHighUserLink() (bool, error) {
+func (c ChainLink) IsHighUserLink(mctx MetaContext, uid keybase1.UID) (bool, error) {
 	v2Type, err := c.GetSigchainV2Type()
 	if err != nil {
 		return false, err
@@ -1457,7 +1457,10 @@ func (c ChainLink) IsHighUserLink() (bool, error) {
 		if prevLink == nil {
 			return false, ChainLinkWrongSeqnoError{}
 		}
-		hardcodedEldest = isSubchainStart(&c, prevLink)
+		hardcodedEldest, err = isSubchainStart(mctx, &c, prevLink, uid)
+		if err != nil {
+			return false, err
+		}
 	}
 
 	isFirstLink := v2Type == SigchainV2TypeEldest || c.GetSeqno() == 1 || hardcodedEldest
@@ -1474,8 +1477,8 @@ func (c ChainLink) IsHighUserLink() (bool, error) {
 // subsequent link in the chain (which may not exist yet). This function can
 // only be called after VerifyChain has processed the chainLink, and set
 // c.computedHighSkip.
-func (c ChainLink) ExpectedNextHighSkip() (HighSkip, error) {
-	isHigh, err := c.IsHighUserLink()
+func (c ChainLink) ExpectedNextHighSkip(mctx MetaContext, uid keybase1.UID) (HighSkip, error) {
+	isHigh, err := c.IsHighUserLink(mctx, uid)
 	if err != nil {
 		return HighSkip{}, err
 	}
--- a/go/libkb/chain_link.go
+++ b/go/libkb/chain_link.go
@@ -93,6 +93,7 @@ func (l LinkID) Eq(i2 LinkID) bool {
 type ChainLinkUnpacked struct {
 	prev                               LinkID
 	seqno                              keybase1.Seqno
+	highSkip                           *HighSkip
 	seqType                            keybase1.SeqType
 	ignoreIfUnsupported                SigIgnoreIfUnsupported
 	payloadLocal                       []byte // local track payloads
@@ -226,6 +227,7 @@ type ChainLink struct {
 	unsigned         bool
 	dirty            bool
 	revocationsCache *[]keybase1.SigID
+	computedHighSkip *HighSkip
 
 	unpacked *ChainLinkUnpacked
 	cki      *ComputedKeyInfos
@@ -280,6 +282,10 @@ func (c *ChainLink) getIgnoreIfUnsupportedFromPayload() SigIgnoreIfUnsupported {
 
 func (c *ChainLink) GetIgnoreIfSupported() SigIgnoreIfUnsupported {
 	return c.getIgnoreIfUnsupportedFromPayload()
+}
+
+func (c *ChainLink) getHighSkipFromPayload() *HighSkip {
+	return c.unpacked.highSkip
 }
 
 func (c *ChainLink) IsStubbed() bool {
@@ -490,6 +496,52 @@ func (c *ChainLink) checkAgainstMerkleTree(t *MerkleTriple) (found bool, err err
 	return
 }
 
+func (tmp *ChainLinkUnpacked) parseHighSkipFromPayload(payload []byte) (*HighSkip, error) {
+	hs, dataType, _, err := jsonparser.Get(payload, "high_skip")
+	// high_skip is optional, but must be an object if it exists
+	if err != nil {
+		if err == jsonparser.KeyPathNotFoundError {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	if dataType != jsonparser.Object {
+		return nil, ChainLinkError{fmt.Sprintf("When provided, expected high_skip to be a JSON object, was %v.", dataType)}
+	}
+
+	highSkipSeqnoInt, err := jsonparser.GetInt(hs, "seqno")
+	if err != nil {
+		return nil, err
+	}
+
+	// highSkipHash can either be null (zero-value of a LinkID) or a hexstring.
+	// We call GetString first instead of Get so we only parse the value
+	// twice for the first link.
+	highSkipHashStr, err := jsonparser.GetString(hs, "hash")
+	var highSkipHash LinkID
+	if err != nil {
+		// If there was an error parsing as a string, make sure the value is null.
+		_, dataType, _, getErr := jsonparser.Get(hs, "hash")
+		if getErr != nil {
+			return nil, getErr
+		}
+		if dataType != jsonparser.Null {
+			return nil, ChainLinkError{
+				fmt.Sprintf("high_skip.hash was neither a valid string (%v) nor null.", err.Error()),
+			}
+		}
+	} else {
+		highSkipHash, err = LinkIDFromHex(highSkipHashStr)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	highSkip := NewHighSkip(keybase1.Seqno(highSkipSeqnoInt), highSkipHash)
+	return &highSkip, nil
+}
+
 func (tmp *ChainLinkUnpacked) unpackPayloadJSON(g *GlobalContext, payload []byte) error {
 	if s, err := jsonparser.GetString(payload, "body", "key", "fingerprint"); err == nil {
 		if tmp.pgpFingerprint, err = PGPFingerprintFromHex(s); err != nil {
@@ -523,6 +575,12 @@ func (tmp *ChainLinkUnpacked) unpackPayloadJSON(g *GlobalContext, payload []byte
 			return err
 		}
 	}
+
+	highSkip, err := tmp.parseHighSkipFromPayload(payload)
+	if err != nil {
+		return err
+	}
+	tmp.highSkip = highSkip
 
 	tmp.typ, err = jsonparser.GetString(payload, "body", "type")
 	if err != nil {
@@ -638,11 +696,22 @@ func (c *ChainLink) unpackStubbed(raw string) error {
 	}
 
 	c.id = ol.LinkID()
+
+	// Because the outer link does not have a highSkip parent object, we check
+	// for the nullity of highSkipSeqno to see if highSkip should be set, since
+	// a null highSkipHash is valid when specifying highSkip=0.
+	var highSkipPtr *HighSkip
+	if ol.HighSkipSeqno != nil {
+		highSkip := NewHighSkip(*ol.HighSkipSeqno, *ol.HighSkipHash)
+		highSkipPtr = &highSkip
+	}
+
 	c.unpacked = &ChainLinkUnpacked{
 		prev:                ol.Prev,
 		seqno:               ol.Seqno,
 		seqType:             ol.SeqType,
 		ignoreIfUnsupported: ol.IgnoreIfUnsupported,
+		highSkip:            highSkipPtr,
 		sigVersion:          ol.Version,
 		outerLinkV2:         ol,
 		stubbed:             true,
@@ -973,13 +1042,14 @@ func (c *ChainLink) verifyPayloadV2() error {
 	prev := c.getPrevFromPayload()
 	curr := c.getPayloadHash()
 	ignoreIfUnsupported := c.getIgnoreIfUnsupportedFromPayload()
-	linkType, err := c.GetSigchainV2Type(SigIgnoreIfUnsupported(ignoreIfUnsupported))
+	linkType, err := c.GetSigchainV2TypeFromInner(SigIgnoreIfUnsupported(ignoreIfUnsupported))
 	if err != nil {
 		return err
 	}
 	seqType := c.getSeqTypeFromPayload()
+	highSkip := c.getHighSkipFromPayload()
 
-	if err := ol.AssertFields(version, seqno, prev, curr, linkType, seqType, ignoreIfUnsupported); err != nil {
+	if err := ol.AssertFields(version, seqno, prev, curr, linkType, seqType, ignoreIfUnsupported, highSkip); err != nil {
 		return err
 	}
 
@@ -1016,6 +1086,10 @@ func (c *ChainLink) getSeqnoFromPayload() keybase1.Seqno {
 
 func (c *ChainLink) GetSeqno() keybase1.Seqno {
 	return c.unpacked.seqno
+}
+
+func (c *ChainLink) GetHighSkip() *HighSkip {
+	return c.unpacked.highSkip
 }
 
 func (c *ChainLink) GetSigID() keybase1.SigID {
@@ -1164,7 +1238,7 @@ func (c *ChainLink) verifyLinkV2() error {
 	return c.verifyPayloadV2()
 }
 
-func (c *ChainLink) GetSigchainV2Type(ignoreIfUnsupported SigIgnoreIfUnsupported) (SigchainV2Type, error) {
+func (c *ChainLink) GetSigchainV2TypeFromInner(ignoreIfUnsupported SigIgnoreIfUnsupported) (SigchainV2Type, error) {
 	if c.unpacked == nil || c.unpacked.typ == "" {
 		return SigchainV2TypeNone, errors.New("chain link not unpacked")
 	}
@@ -1179,6 +1253,22 @@ func (c *ChainLink) GetSigchainV2TypeFromV2Shell() (SigchainV2Type, error) {
 		return SigchainV2TypeNone, errors.New("GetSigchainV2TypeFromV2Shell: chain link has no v2 shell")
 	}
 	return c.unpacked.outerLinkV2.LinkType, nil
+}
+
+// GetSigchainV2Type is a helper function for getting a ChainLink's type. If it
+// is a v2 link (that may or may not be stubbed), return the type from the
+// outer link, otherwise from the inner link.
+func (c *ChainLink) GetSigchainV2Type() (SigchainV2Type, error) {
+	if c.unpacked == nil {
+		return SigchainV2TypeNone, errors.New("chain link is not unpacked")
+	}
+	if c.unpacked.outerLinkV2 == nil && c.unpacked.typ == "" {
+		return SigchainV2TypeNone, errors.New("chain inner link type is not unpacked, and has no v2 shell")
+	}
+	if c.unpacked.outerLinkV2 != nil {
+		return c.GetSigchainV2TypeFromV2Shell()
+	}
+	return c.GetSigchainV2TypeFromInner(c.GetIgnoreIfSupported())
 }
 
 func (c *ChainLink) checkServerSignatureMetadata(ckf ComputedKeyFamily) (ret keybase1.KID, err error) {
@@ -1350,4 +1440,50 @@ func (c ChainLink) AllowStubbing() bool {
 		return false
 	}
 	return c.unpacked.outerLinkV2.LinkType.AllowStubbing()
+}
+
+// IsHighUserLink determines whether a chainlink counts as "high" in a user's chain,
+// which is defined as an Eldest link, a link with seqno=1, a link that is Sibkey,
+// PGPUpdate, Revoke, or any link that is revoking.
+func (c ChainLink) IsHighUserLink() (bool, error) {
+	v2Type, err := c.GetSigchainV2Type()
+	if err != nil {
+		return false, err
+	}
+
+	hardcodedEldest := false
+	if c.GetSeqno() > 1 {
+		prevLink := c.parent.GetLinkFromSeqno(c.GetSeqno() - 1)
+		if prevLink == nil {
+			return false, ChainLinkWrongSeqnoError{}
+		}
+		hardcodedEldest = isSubchainStart(&c, prevLink)
+	}
+
+	isFirstLink := v2Type == SigchainV2TypeEldest || c.GetSeqno() == 1 || hardcodedEldest
+	isNewHighLink := isFirstLink ||
+		v2Type == SigchainV2TypeRevoke ||
+		v2Type == SigchainV2TypeWebServiceBindingWithRevoke ||
+		v2Type == SigchainV2TypeCryptocurrencyWithRevoke ||
+		v2Type == SigchainV2TypeSibkey ||
+		v2Type == SigchainV2TypePGPUpdate
+	return isNewHighLink, nil
+}
+
+// ExpectedNextHighSkip returns the expected highSkip of the immediately
+// subsequent link in the chain (which may not exist yet). This function can
+// only be called after VerifyChain has processed the chainLink, and set
+// c.computedHighSkip.
+func (c ChainLink) ExpectedNextHighSkip() (HighSkip, error) {
+	isHigh, err := c.IsHighUserLink()
+	if err != nil {
+		return HighSkip{}, err
+	}
+	if isHigh {
+		return NewHighSkip(c.GetSeqno(), c.id), nil
+	}
+	if c.computedHighSkip == nil {
+		return HighSkip{}, NewUserReverifyNeededError("Expected to have already computed this link's HighSkip, but it was not computed.")
+	}
+	return *c.computedHighSkip, nil
 }
--- a/go/libkb/chain_link_v2.go
+++ b/go/libkb/chain_link_v2.go
@@ -198,6 +198,12 @@ type OuterLinkV2 struct {
 	// - it can be stubbed for non-admins
 	// - it cannot be stubbed for admins
 	IgnoreIfUnsupported SigIgnoreIfUnsupported `codec:"ignore_if_unsupported"`
+	// -- Links exist in the wild that are missing fields below this line too.
+	// If not provided, both of these are nil, and highSkip in the inner link is set to nil.
+	// Note that a link providing HighSkipSeqno == 0 and HighSkipHash == nil is valid
+	// (and mandatory) for an initial link.
+	HighSkipSeqno *keybase1.Seqno `codec:"high_skip_seqno"`
+	HighSkipHash  *LinkID         `codec:"high_skip_hash"`
 }
 
 func (o OuterLinkV2) Encode() ([]byte, error) {
@@ -234,7 +240,83 @@ type SigHasRevokes bool
 
 func (b SigIgnoreIfUnsupported) Bool() bool { return bool(b) }
 
+func encodeOuterLink(
+	m MetaContext,
+	v1LinkType LinkType,
+	seqno keybase1.Seqno,
+	innerLinkJSON []byte,
+	prevLinkID LinkID,
+	hasRevokes SigHasRevokes,
+	seqType keybase1.SeqType,
+	ignoreIfUnsupported SigIgnoreIfUnsupported,
+	highSkip *HighSkip,
+) ([]byte, error) {
+	var encodedOuterLink []byte
+
+	currLinkID := ComputeLinkID(innerLinkJSON)
+
+	v2LinkType, err := SigchainV2TypeFromV1TypeAndRevocations(string(v1LinkType), hasRevokes, ignoreIfUnsupported)
+	if err != nil {
+		return encodedOuterLink, err
+	}
+
+	// When 2.3 links are mandatory, it will be invalid for highSkip == nil,
+	// so the featureflag check will be removed and the nil check will result
+	// in an error.
+	allowHighSkips := m.G().Env.GetFeatureFlags().HasFeature(EnvironmentFeatureAllowHighSkips)
+	if allowHighSkips && highSkip != nil {
+		highSkipSeqno := &highSkip.Seqno
+		highSkipHash := &highSkip.Hash
+		outerLink := OuterLinkV2{
+			Version:             2,
+			Seqno:               seqno,
+			Prev:                prevLinkID,
+			Curr:                currLinkID,
+			LinkType:            v2LinkType,
+			SeqType:             seqType,
+			IgnoreIfUnsupported: ignoreIfUnsupported,
+			HighSkipSeqno:       highSkipSeqno,
+			HighSkipHash:        highSkipHash,
+		}
+		encodedOuterLink, err = outerLink.Encode()
+	} else {
+		// This is a helper struct. When the code for Sigchain 2.3
+		// is released, it is possible some clients will still post 2.2
+		// links, i.e., without high_skip information. Due to a bug
+		// in Keybase's fork of go-codec, omitempty does not work
+		// for arrays. So, we send up the serialization of the
+		// appropriate struct depending on whether we are making a 2.3 link.
+		// When 2.3 links are mandatory, this struct can be deleted.
+		encodedOuterLink, err = MsgpackEncode(
+			struct {
+				_struct             bool                   `codec:",toarray"`
+				Version             int                    `codec:"version"`
+				Seqno               keybase1.Seqno         `codec:"seqno"`
+				Prev                LinkID                 `codec:"prev"`
+				Curr                LinkID                 `codec:"curr"`
+				LinkType            SigchainV2Type         `codec:"type"`
+				SeqType             keybase1.SeqType       `codec:"seqtype"`
+				IgnoreIfUnsupported SigIgnoreIfUnsupported `codec:"ignore_if_unsupported"`
+			}{
+				Version:             2,
+				Seqno:               seqno,
+				Prev:                prevLinkID,
+				Curr:                currLinkID,
+				LinkType:            v2LinkType,
+				SeqType:             seqType,
+				IgnoreIfUnsupported: ignoreIfUnsupported,
+			})
+	}
+
+	if err != nil {
+		return encodedOuterLink, err
+	}
+
+	return encodedOuterLink, err
+}
+
 func MakeSigchainV2OuterSig(
+	m MetaContext,
 	signingKey GenericKey,
 	v1LinkType LinkType,
 	seqno keybase1.Seqno,
@@ -243,24 +325,10 @@ func MakeSigchainV2OuterSig(
 	hasRevokes SigHasRevokes,
 	seqType keybase1.SeqType,
 	ignoreIfUnsupported SigIgnoreIfUnsupported,
+	highSkip *HighSkip,
 ) (sig string, sigid keybase1.SigID, linkID LinkID, err error) {
-	currLinkID := ComputeLinkID(innerLinkJSON)
 
-	v2LinkType, err := SigchainV2TypeFromV1TypeAndRevocations(string(v1LinkType), hasRevokes, ignoreIfUnsupported)
-	if err != nil {
-		return sig, sigid, linkID, err
-	}
-
-	outerLink := OuterLinkV2{
-		Version:             2,
-		Seqno:               seqno,
-		Prev:                prevLinkID,
-		Curr:                currLinkID,
-		LinkType:            v2LinkType,
-		SeqType:             seqType,
-		IgnoreIfUnsupported: ignoreIfUnsupported,
-	}
-	encodedOuterLink, err := outerLink.Encode()
+	encodedOuterLink, err := encodeOuterLink(m, v1LinkType, seqno, innerLinkJSON, prevLinkID, hasRevokes, seqType, ignoreIfUnsupported, highSkip)
 	if err != nil {
 		return sig, sigid, linkID, err
 	}
@@ -426,6 +494,10 @@ func SigchainV2TypeFromV1TypeTeams(s string) (ret SigchainV2Type, err error) {
 	return ret, err
 }
 
+func mismatchError(format string, arg ...interface{}) error {
+	return SigchainV2MismatchedFieldError{fmt.Sprintf(format, arg...)}
+}
+
 func (o OuterLinkV2) AssertFields(
 	version int,
 	seqno keybase1.Seqno,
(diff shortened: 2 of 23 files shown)


heronhaye:
WIP

todo:
- more tests
- many clarifications

go/libkb/kbsig.go
```diff
@@ -329,6 +349,26 @@ func (arg ProofMetadata) ToJSON(m MetaContext) (ret *jsonw.Wrapper, err error) {
 	ret.SetKey("seqno", jsonw.NewInt64(int64(seqno)))
 	ret.SetKey("prev", prev)
 
+	// If this is a standard user link, arg.Me will be provided. It is not
+	// provided during a provisionee's KEX and for team sigs, where it is
+	// overridden. It is expected to be provided by PerUserKey and
+	// Stellar proofs as well.
+	var hPrevInfo HPrevInfo
+	if arg.Me != nil {
```
heronhaye: As far as I can tell `arg.Me` is only `nil` during a KEX. Is this correct?

heronhaye: They're also nullable for PerUserKeyProof and StellarProof types but I'm not sure why (they're functions that take in *User instead of methods on User later in kbsig.go)

go/libkb/sig_chain.go
```diff
@@ -370,6 +378,29 @@ func (sc *SigChain) VerifyChain(m MetaContext) (err error) {
 		curr.chainVerified = true
 	}
 
+	newHPrevInfo := NewInitialHPrevInfo()
```
heronhaye: The main for loop in VerifyChain (folded away right now) traverses a user's sigchain backwards, I can't see any reason for that. HPrevInfo computing code needs to traverse forwards, so I can save a pass if I can reverse the traversal in the first for loop in VerifyChain.

go/libkb/sig_chain.go
```diff
@@ -211,11 +216,14 @@ func (sc *SigChain) VerifiedChainLinks(fp PGPFingerprint) (ret ChainLinks) {
 	return ret
 }
 
-func (sc *SigChain) Bump(mt MerkleTriple) {
+func (sc *SigChain) Bump(mt MerkleTriple, isHighDelegator bool) {
 	mt.Seqno = sc.GetLastKnownSeqno() + 1
 	sc.G().Log.Debug("| Bumping SigChain LastKnownSeqno to %d", mt.Seqno)
 	sc.localChainTail = &mt
```
heronhaye: Is `localChainTail` only used for bumping pointers during multisigs?

go/libkb/sig_chain.go
```diff
@@ -48,6 +48,11 @@ type SigChain struct {
 	idVerified        bool
 	loadedFromLinkOne bool
```
heronhaye: `loadedFromLinkOne` seems like dead code. But I'm not sure how I would be calculating hPrevs if this was ever set `true`. Should I remove it?

go/teams/chain.go
```diff
@@ -719,7 +733,7 @@ type checkInnerLinkResult struct {
 // Does not modify `prevState` but returns a new state.
 func (t *teamSigchainPlayer) addInnerLink(
 	prevState *TeamSigChainState, link *ChainLinkUnpacked, signer SignerX,
-	isInflate bool) (
+	isInflate bool, skipComputingHPrevInfo bool) (
```
heronhaye: @mlsteele I'm adding this flag to addInnerLink because the inflate was messing up the computed values. Not super elegant but if we're not going to call addInnerLink out of order otherwise, it might be ok. What do you think?

go/libkb/kbsig.go
```diff
@@ -329,6 +349,26 @@ func (arg ProofMetadata) ToJSON(m MetaContext) (ret *jsonw.Wrapper, err error) {
...
+		hPrevInfo = arg.Me.GetLastKnownHPrevInfo()
+	} else {
+		hPrevInfo = *arg.HPrevInfoOverride
```
heronhaye: This is also kind of fragile. For example, if for some reason KEX changes to add more links than just the sibkey/subkey, this would be wrong and we would need to end up sending hPrevInfo pointers through KEX.

go/teams/chain.go
mlsteele: Seems like `skipComputingHPrevInfo == isInflate` so how about just use `isInflate`? That should hold as long as it remains impossible to inflate high links.

heronhaye: Well, the issue I had was the player was run and lastKnownHprevs were computed. Then inflate was run on a middle link and it tried to compare the link's provided hPrevInfo with what was computed (which is accurate for *new* links only).

But yes, they are equal. Since we can maintain that addInnerLink is never called otherwise I'll use isInflate instead and add a comment.

go/libkb/sig_chain.go
maxtaco: maybe it use to go backwards to verify the start of the current subchain.

go/libkb/kbsig.go
maxtaco: lots of other parts of this code that fills in signature bodies requires non-`nil` arg.Me

go/libkb/chain_link.go
```diff
@@ -492,6 +497,24 @@ func (tmp *ChainLinkUnpacked) unpackPayloadJSON(g *GlobalContext, payload []byte
 		}
 	}
 
+	if _, dataType, _, err := jsonparser.Get(payload, "hprev_info"); err == nil && dataType == jsonparser.Object {
```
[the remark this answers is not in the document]
maxtaco: I don't want `hprev_info` to make it into the sigchain.  Can you call this `hprev` in the output json?

go/libkb/chain_link.go
```diff
@@ -604,11 +627,22 @@ func (c *ChainLink) unpackStubbed(raw string) error {
 	}
 
 	c.id = ol.LinkID()
+
+	// Because the outer link does not have a hPrevInfo parent object, we check
+	// for the nullity of hPrevSeqno to see if hPrevInfo should be set, since
+	// a null hPrevHash is valid even when specifying hPrevInfo.
```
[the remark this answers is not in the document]
maxtaco: btw, it's still not too late to make a substructure for hprev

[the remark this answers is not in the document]
maxtaco: it's still not too late to add a substructure

go/libkb/kbsig.go
heronhaye: I wanted to make sure the only case where it is nil is during KEX so I can handle all the cases

go/libkb/kbsig.go
```diff
@@ -252,6 +252,22 @@ func remoteProofToTrackingStatement(s RemoteProofChainLink, base *jsonw.Wrapper)
 	return nil
 }
 
+type HPrevInfo struct {
```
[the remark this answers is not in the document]
heronhaye: So `keybase1.LinkID` is a string and `libkb.LinkID` is bytes. My understanding is the first is used for teams and the second elsewhere. If I put it in AVDL should I make a new AVDL LinkID type that is bytes? Or use the string type?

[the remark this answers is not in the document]
maxtaco: OK, that makes sense.

heronhaye:
I like the idea where clients can do whatever until a certain day when we say they must send hprev_info. Otherwise, if a user upgrades only on one device and posts an hprevinfo sig they'll be forced to upgrade their other devices too.

Sequence might be

1. Run backfills for teams and users
2. Merge in checking (optional) hprev infos on kbweb
3.1. Enable hprev support on go client 
3.2. Enable hprev support on webclient
4. (in the future) Enforce new sigs must be >2.3 on server

go/libkb/chain_link.go
```diff
@@ -1303,3 +1342,26 @@ func (c ChainLink) AllowStubbing() bool {
...
+	v2Type, err := c.GetSigchainV2Type(c.GetIgnoreIfSupported())
+	if err != nil {
+		v2Type, err = c.GetSigchainV2TypeFromV2Shell()
+		if err != nil {
+			return false, ChainLinkError{fmt.Sprintf("Could not determine linkType of chain link.")}
+		}
+	}
+
+	isFirstLink := v2Type == SigchainV2TypeEldest || int(c.GetSeqno()) == int(1)
+	isNewHighLink := isFirstLink ||
```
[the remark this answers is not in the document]
heronhaye: The code that computes it for teams needs info from the player so this can't handle it. I'll change the name to signify it's only for userchains.

go/libkb/kbsig.go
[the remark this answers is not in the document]
heronhaye: The issue is that `keybase1.LinkID` has type string, and `libkb.LinkID` has type bytes. The `libkb.LinkID` has a few nice functions for equality checking/exim in `chain_link.go`, and if I made a new `"LinkIDBytes"` type in AVDL I'd have to wrap those, which seems unnecessary/bloated for a type that's only going to be used in `HPrevInfo`.

What if we made a `keybase1.HPrevInfo` using a `keybase1.LinkID` but only used it for export? And make a corresponding `Export()` function in `libkb/rpc_exim.go`?

go/libkb/kbsig.go
heronhaye: I just wanted to make sure that KEX was the only case where `arg.Me=nil`, so I could cover all the cases

go/libkb/chain_link.go
```diff
@@ -1303,3 +1342,26 @@ func (c ChainLink) AllowStubbing() bool {
...
 	return c.unpacked.outerLinkV2.LinkType.AllowStubbing()
 }
+
+// This logic is defined at
+// https://keybase.atlassian.net/wiki/spaces/GEN/pages/499318820/Sigchain+Playback+Performance+Ideas
+// under Discussion. Note that the SigchainV2Type converter covers all
+// the cases where a link is revoking but it is not a revoke type link.
+func (c ChainLink) IsHighLink() (bool, error) {
+	v2Type, err := c.GetSigchainV2Type(c.GetIgnoreIfSupported())
+	if err != nil {
```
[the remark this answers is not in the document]
heronhaye: Good point. I made a helper function to help with that and return proper errors.

go/libkb/chain_link.go
```diff
@@ -1303,3 +1342,26 @@ func (c ChainLink) AllowStubbing() bool {
...
+		v2Type, err = c.GetSigchainV2TypeFromV2Shell()
+		if err != nil {
+			return false, ChainLinkError{fmt.Sprintf("Could not determine linkType of chain link.")}
+		}
+	}
+
+	isFirstLink := v2Type == SigchainV2TypeEldest || int(c.GetSeqno()) == int(1)
+	isNewHighLink := isFirstLink ||
+		v2Type == SigchainV2TypeRevoke ||
+		v2Type == SigchainV2TypeWebServiceBindingWithRevoke ||
```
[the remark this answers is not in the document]
heronhaye: The kbweb PR rejects invalid hprevs if given.

go/libkb/chain_link.go
```diff
@@ -1134,6 +1208,22 @@ func (c *ChainLink) GetSigchainV2TypeFromV2Shell() (SigchainV2Type, error) {
 	return c.unpacked.outerLinkV2.LinkType, nil
 }
 
+// Helper function for getting a ChainLink's type. If it is a v2 link (that may
```
maxtaco: use Go-style commenting where you begin the comment with the name of the function.

go/libkb/chain_link.go
```diff
@@ -1303,3 +1393,23 @@ func (c ChainLink) AllowStubbing() bool {
 	}
 	return c.unpacked.outerLinkV2.LinkType.AllowStubbing()
 }
+
+// This logic is defined at
```
maxtaco: same thing about Go-style comments.  Also, not great to link to a private document in public code.  Can you put in the important parts of that doc here?

go/libkb/kbsig.go
maxtaco: Shouldn't the logic go the other way... To always use HprevInfoOverride if provided? And if not fallback to arg.Me.  And if not, then return an error?

go/libkb/kbsig.go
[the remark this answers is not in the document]
maxtaco: But it's not used anywhere, is it?

go/libkb/sig_chain.go
```diff
@@ -370,6 +381,29 @@ func (sc *SigChain) VerifyChain(m MetaContext) (err error) {
 		curr.chainVerified = true
 	}
 
+	newHPrevInfo := NewInitialHPrevInfo()
+	for i := 0; i < len(sc.chainLinks); i++ {
+		curr := sc.chainLinks[i]
+
+		if hPrevInfo := curr.GetHPrevInfo(); hPrevInfo != nil {
```
maxtaco: Can you provide a comment about it being an error if null, at some point?  How will we enforce that? I.e., how will we know when to error if an hprev is expected but isn't there?

go/libkb/sig_chain.go
```diff
@@ -370,6 +381,29 @@ func (sc *SigChain) VerifyChain(m MetaContext) (err error) {
 		curr.chainVerified = true
 	}
 
+	newHPrevInfo := NewInitialHPrevInfo()
+	for i := 0; i < len(sc.chainLinks); i++ {
```
maxtaco: This will be a major performance regression.  Looking at the full sigchain loading code. note that `VerifyChain` is called several times, but short-circuits due to `curr.chainVerified`.  Your code is just going to plow over the whole chain repeatedly.  I think it needs to be better.

maxtaco: You can use the same structure as what's already there to solve this issue btw. I would recommend attaching an hprev to those links that don't supply it explicitly, so you can pick up traversal where you left off.

And *this* is why we go backwards.  I think you need to go backwards too.

go/libkb/kbsig.go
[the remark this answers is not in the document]
heronhaye: `keybase1.HPrevInfo` is not used anywhere currently

go/libkb/sig_chain.go
heronhaye: This function will have to play old links too, right? Would we need to do something like set a feature-flag and reject the link if the feature-flag is on, it doesn't have an hprev_info and it's ctime is after some "enforce hprev_infos" day?

go/libkb/kbsig.go
heronhaye: Override is a bad name, it used to actually override before I put the multisig handling into sigchain_bump. But now exactly one of arg.Me and hPrevInfoOverride should be provided. I'll change the name and error if both or neither are provided.

go/libkb/chain_link.go
```diff
@@ -458,6 +463,53 @@ func (c *ChainLink) checkAgainstMerkleTree(t *MerkleTriple) (found bool, err err
...
+		} else {
+			return nil, err
+		}
+	}
+
+	if dataType != jsonparser.Object {
+		return nil, ChainLinkError{fmt.Sprintf("When provided, expected high_skip to be a JSON object, was %v.", dataType)}
+	}
+
+	hPrevSeqnoInt, err := jsonparser.GetInt(payload, "high_skip", "seqno")
```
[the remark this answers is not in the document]
patrickxb: It should be fine.  There might be a slight speedup by changing line 467 to 

    hs, dataType, _, err := jsonparser.Get(payload, "high_skip")

Then this line 481

    hPrevSeqnoInt, err := jsonparser.GetInt(hs, "seqno")

and elsewhere below where you are getting things inside `high_skip`.

Best way to tell would be to benchmark it.  I'll add a comment on the whole PR about the sigchain benchmark tool I wrote when we worked on optimizations earlier.

patrickxb:
I wrote `client/go/tools/sigchain` to help benchmark sigchain loading.  It might be helpful to use it for this work.  It has support for memory and cpu profiling.

go/libkb/sig_chain.go
maxtaco: Ok, simple fix!  Instead of `i := 0`, just do `i := <when the above loop stopped>`

Sorry for the churn...

go/libkb/sig_chain.go
```diff
@@ -370,6 +381,29 @@ func (sc *SigChain) VerifyChain(m MetaContext) (err error) {
...
+
+		isHigh, err := curr.IsHighUserLink()
+		if err != nil {
+			return err
+		}
+		if isHigh {
+			newHPrevInfo = NewHPrevInfo(curr.GetSeqno(), curr.id)
+		}
+	}
+	sc.hPrevInfo = newHPrevInfo
```
maxtaco: i would set this field on each link, rather than on the chain

go/libkb/kbsig.go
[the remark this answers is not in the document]
heronhaye: While doing benchmarking I noticed something interesting. Having that export function makes loading/verifying/storing the sigchain about 20% faster. I'm not sure exactly where it is called but I guess the default export is slow. So we might want to consider making exports for other objects as well (will do some more benchmarking to see if it helps).

heronhaye:
Note: I ended up going with the "forward loop after verify". If we change that logic it should be in another PR I think.

Here are some basic benchmarks with patrick's tool. I use Chris' sigchain up to link 1181--I was having difficulty inserting hPrevInfos into Sigchain v2 links and generating the sig properly.

Here's what's weird: My code runs faster than master (I've rebased) for even a sigchain without any hPrevInfos. I think this is probably due to a bug and I'm currently trying to find out why.

**New code, chain without any hPrevInfos**
sig chain load time: 220.659368ms
sig chain load time: 215.105665ms
sig chain load time: 185.585235ms
sig chain load time: 185.113498ms
sig chain load time: 181.354496ms
sig chain load time: 182.001488ms

**New code, chain specifies hPrevInfos at every link**
sig chain load time: 209.048189ms
sig chain load time: 187.376278ms
sig chain load time: 180.46583ms
sig chain load time: 183.717237ms
sig chain load time: 180.633384ms
sig chain load time: 180.596296ms


**Master, chain specifies hPrevInfos but is not parsed**
sig chain load time: 220.743641ms
sig chain load time: 199.590767ms
sig chain load time: 194.840192ms
sig chain load time: 200.595857ms
sig chain load time: 198.960194ms
sig chain load time: 196.902234ms

maxtaco:
I suggested over KB messenger that @heronhaye take a count of thenumber of link unpacks, to make sure they haven't changed.

go/libkb/chain_link_v2_test.go
```diff
@@ -110,3 +111,51 @@ func TestOuterLinkV2WithMetadataPointerContainerDecode(t *testing.T) {
 	err = MsgpackDecode(&o, bytes)
 	requireErrorHasSuffix(t, errCodecDecodeSelf, err)
 }
+
+func serdeOuterLink(t *testing.T, hPrevSeqno *keybase1.Seqno, hPrevHash LinkID, addHPrevHash bool) OuterLinkV2 {
+	o := OuterLinkV2{Version: 2, Seqno: keybase1.Seqno(2), Prev: nil, Curr: nil, LinkType: 1, SeqType: 1, IgnoreIfUnsupported: false}
```
xgess: can you swap out these integers with the named constants, i.e. `LinkType: 1` with `LinkType: SigchainV2TypeEldest` and `SeqType: 1` with `SeqType: keybase1.SeqType_PUBLIC`?

go/libkb/chain_link_v2_test.go
```diff
@@ -110,3 +111,51 @@ func TestOuterLinkV2WithMetadataPointerContainerDecode(t *testing.T) {
...
+}
+
+func TestHPrevBackwardsCompatibility(t *testing.T) {
+	o := serdeOuterLink(t, nil, nil, false)
+	require.True(t, o.HPrevSeqno == nil)
+	require.True(t, o.HPrevHash == nil)
+	require.NoError(t, hPrevInfoMatch(o, nil))
+
+	seqno := keybase1.Seqno(3)
+	o = serdeOuterLink(t, &seqno, nil, false)
```
xgess: this is kind of a weird test. you're creating a new outer link with a seqno of 2 but that has a previous high sequence number of 3? maybe the serdeOuterLink helper should also take seqno as a parameter so the numbers can at least advance in the same direction.

go/libkb/chain_link_v2_test.go
```diff
@@ -110,3 +111,51 @@ func TestOuterLinkV2WithMetadataPointerContainerDecode(t *testing.T) {
...
+	require.True(t, *o.HPrevSeqno == 3)
+	require.True(t, o.HPrevHash == nil)
+	hPrevInfo := NewHPrevInfo(keybase1.Seqno(3), nil)
```
xgess: this is the good example in the test, but we should never actually have an hPrevInfo that has a valid seqno and a nil hash, correct? that seems very unexpected to me. this whole test is actually pretty deep into implementation details. could be the kind of thing that hampers our ability to refactor and clean-up code later. i'll try to figure out if there's a more behavioral way to test that the hprev stuff is backwards-compatible.

go/systests/team_tx_test.go
```diff
@@ -149,7 +149,7 @@ func TestTeamTxDependency(t *testing.T) {
 
 	tx := teams.CreateAddMemberTx(teamObj)
 	tx.AddMemberByUsername(context.Background(), tracy.username, keybase1.TeamRole_READER)
-	tx.AddMemberByUsername(context.Background(), bob.username, keybase1.TeamRole_WRITER)
+	tx.AddMemberByUsername(context.Background(), bob.username, keybase1.TeamRole_ADMIN)
```
xgess: @heronhaye maybe also add the new thing you're testing here (high info) to the description above, since it's not really obvious that this test is now also testing something else.

go/systests/team_tx_test.go
```diff
@@ -181,13 +181,14 @@ func TestTeamTxDependency(t *testing.T) {
...
 	tx.AddMemberByAssertion(context.Background(), fmt.Sprintf("%s@rooter", tracy.username), keybase1.TeamRole_WRITER)
-	tx.AddMemberByUsername(context.Background(), bob.username, keybase1.TeamRole_WRITER)
+	tx.AddMemberByUsername(context.Background(), bob.username, keybase1.TeamRole_ADMIN)
 
 	payloads = tx.DebugPayloads()
 	require.Equal(t, 3, len(payloads))
 
 	err = tx.Post(libkb.NewMetaContextForTest(*ann.tc))
 	require.NoError(t, err)
+
```
xgess: whitespace

go/teams/sig.go
```diff
@@ -267,20 +287,33 @@ func precheckLinkToPost(ctx context.Context, g *libkb.GlobalContext,
...
+		state, err = playSigItem(ctx, g, me, state, sigItem)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func playSigItem(ctx context.Context, g *libkb.GlobalContext, me keybase1.UserVersion, state *TeamSigChainState, sigMultiItem libkb.SigMultiItem) (*TeamSigChainState, error) {
```
mlsteele: I'd suggest putting `precheck` in the name like `playPrecheckSigItem` so that nobody's tempted to call this for anything else.

go/libkb/chain_link.go
```diff
@@ -458,6 +464,53 @@ func (c *ChainLink) checkAgainstMerkleTree(t *MerkleTriple) (found bool, err err
...
+
+	// hPrevHash can either be null (zero-value of a LinkID) or a hexstring.
+	// We call GetString first instead of Get so we only parse the value
+	// twice for the first link.
+	hPrevHashStr, err := jsonparser.GetString(hs, "hash")
+	var hPrevHash LinkID
+	if err != nil {
+		// If there was an error parsing as a string, make sure the value is null.
+		_, dataType, _, getErr := jsonparser.Get(hs, "hash")
+		if err != nil {
```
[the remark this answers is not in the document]
heronhaye: I wanted to return the GetString error if both failed (in line 500). But yes it should be getErr here instead.

go/teams/sig.go
```diff
@@ -287,20 +287,33 @@ func precheckLinkToPost(ctx context.Context, g *libkb.GlobalContext,
...
 		role, err := state.GetUserRole(me)
 		if err != nil {
 			role = keybase1.TeamRole_NONE
 		}
 		isAdmin = role.IsAdminOrAbove()
-
-		// As an optimization, AppendChainLink consumes its state.
-		// We don't consume our state parameter.
-		// So clone state before we pass it along to be consumed.
 		state = state.DeepCopyToPtr()
```
xgess: i don't think we need to deep copy again inside this method. the one in precheckLinksToPost is to protect the caller of precheckLinksToPost from having the state change underneath it. if precheckLinksToPost is going to be the only thing calling playSigItem then i think we can just leave it as a pointer and update it with each pass and actually not even have to return it from this method.

go/libkb/chain_link.go
```diff
@@ -1303,3 +1394,40 @@ func (c ChainLink) AllowStubbing() bool {
 	}
 	return c.unpacked.outerLinkV2.LinkType.AllowStubbing()
 }
+
+// isHighUserLink determines whether a chainlink counts as "high" in a user's chain,
```
xgess: `IsHighUserLink`

go/libkb/kbsig.go
```diff
@@ -329,6 +359,27 @@ func (arg ProofMetadata) ToJSON(m MetaContext) (ret *jsonw.Wrapper, err error) {
...
+	}
+
+	hPrevInfoObj := jsonw.NewDictionary()
+	hPrevInfoObj.SetKey("seqno", jsonw.NewInt64(int64(hPrevInfo.Seqno)))
+	if hash := hPrevInfo.Hash; hash != nil {
+		hPrevInfoObj.SetKey("hash", jsonw.NewString(hash.String()))
+	} else {
+		hPrevInfoObj.SetKey("hash", jsonw.NewNil())
+	}
+	ret.SetKey("high_skip", hPrevInfoObj)
```
mlsteele: Would it be better to skip this section if `hPrevInfo` is default-valued? I think it would be default-valued in the case of a team loaded from the previous version of the cache.

mlsteele: Also what about putting hprev writes behind a feature flag?

go/libkb/chain_link_v2.go
```diff
@@ -197,6 +197,11 @@ type OuterLinkV2 struct {
 	// - it can be stubbed for non-admins
 	// - it cannot be stubbed for admins
 	IgnoreIfUnsupported SigIgnoreIfUnsupported `codec:"ignore_if_unsupported"`
+	// -- Links exist in the wild that are missing fields below this line too.
+	// If not provided, both of these are nil, and hPrevInfo in the inner link is set to nil.
+	// Note that a link providing HPrevSeqno != nil and HPrevHash == nil is valid for an initial link.
```
mlsteele: Why is `HPrevSeqno != nil and HPrevHash == nil` valid? Is it required? I think it would be ideal if there were only one possible representation for an initial link (enforced by server).

xgess: huh. yes. i agree. an initial link (team root and subteam head) should always be `HPrevSeqno==0` and `HPrevHash==nil`. and if we're going to force a full reload on pre-existing teams before new links can be posted, then these should be the only cases that have HPrevHash of nil moving forward.

heronhaye: Yes, if hash is null only seqno=0 is valid. This is checked in VerifyChain, the outerlink stuff just checks that inner/outer match.

xgess:
talking with @heronhaye, here's what's outstanding...

on this PR
1. stop actually sending `high_skip`s to the server. we can do this with the new feature flag implementation or by wrapping it in `if false` to stop adding complexity to an already very large piece of work, and then replace it with a feature flag in a fast-follow. 
2. start tracking high links on the team_sigchain_state as they come in. this was removed from an earlier commit and just needs to get put back.
3. get the server PR https://github.com/keybase/keybase/pull/2806 merged and deployed.
4. manual testing against pre-existing and new user/team chains on our deployed server code (with the feature flag off and on).


on the server PR (https://github.com/keybase/keybase/pull/2806) which validates any `high_skip`s that are on links...
1. maybe a little more manual testing on user sigchains, but otherwise it's ready for more eyes. kthx!


otherwise
1. finish backfill - might be worth discussing if it makes sense to go 100% of the way if we only *need* one pre-high_skip-high-link per sigchain. already brought this up with @mlsteele. 
2. set up feature flag on the server. off for everyone.
3. investigate how to build a reader/loader for teams that respects skips. we've been talking about what this will need to look like. what will have to be added to the server, how to fallback to the normal sigchain, etc. we'll make a design doc soon, but it likely won't be trivial. 
4. same for users. teams first because we think that will maybe be easier. i wouldn't be surprised if we change our minds an even number of times. 


think through roll out
1. are we going to get a new link type where an admin can sign for all the high links? @heronhaye mentioned it, but i was unfamiliar. maybe we can assume this will be there when we're building the reader so we can (1) take full advantage of it, and (2) reduce the scope of cases the reader will need to be able to handle. 
2. other rollout concerns like verifying that the clients and the server 100% agree on which are high links, do we want different feature flags for user chains and team chains, etc. all things we'll dig into. :)

go/libkb/chain_link_v2_test.go
```diff
@@ -110,3 +111,51 @@ func TestOuterLinkV2WithMetadataPointerContainerDecode(t *testing.T) {
...
 	requireErrorHasSuffix(t, errCodecDecodeSelf, err)
 }
+
+func serdeOuterLink(t *testing.T, hPrevSeqno *keybase1.Seqno, hPrevHash LinkID, addHPrevHash bool) OuterLinkV2 {
+	o := OuterLinkV2{Version: 2, Seqno: keybase1.Seqno(20), Prev: nil, Curr: nil, LinkType: SigchainV2TypeEldest, SeqType: keybase1.SeqType_PUBLIC, IgnoreIfUnsupported: false}
+	if hPrevSeqno != nil {
+		o.HPrevSeqno = hPrevSeqno
+	}
+	if addHPrevHash {
+		o.HPrevHash = hPrevHash
```
patrickxb: this doesn't compile

go/libkb/kbsig.go
```diff
@@ -520,6 +589,10 @@ func MakeSig(
 	case KeybaseSignatureV2:
 		prevSeqno := me.GetSigChainLastKnownSeqno()
 		prevLinkID := me.GetSigChainLastKnownID()
+		hPrevInfo, err := me.GetExpectedNextHPrevInfo()
```
patrickxb: you should use a different variable name than `err` here since the outer `err` is used, so this will screw up the error handling.

heronhaye:
There's still some more work to do before merging with making sure the upgrade process is smooth. I can change the variable name but I don't understand the issue--is it wrong or just confusing?

go/libkb/kbsig.go
xgess: just tested. can confirm. didn't realize go's order of precedence in the scope. very good to know.

go/teams/chain.go
```diff
@@ -1594,9 +1628,43 @@ func (t *teamSigchainPlayer) addInnerLink(
...
+	highSkipVerifyDesired := !isInflate && !precheck
+	// If presented with an hPrevInfo, and we are unable to verify it,
+	// but would like to, throw a GreenLinkError to force a repoll.
+	if hPrevInfo != nil && highSkipVerifyDesired && reverifyNeeded {
+		fmt.Printf("THROWING GREEN LINK ERROR! %v\n", payload.Seqno)
+		return res, GreenLinkError{msg: "Did not have information to compute expected team high skip; reverifying."}
+	}
+	// We only run addInnerLink out of order when inflating, and so we skip
+	// computing high skips during it.
+	// updates high prev data, so we skip it. If hPrevInfo is presented and
```
xgess: this comment could be clearer if it didn't look like it was mid-way through a copy-paste job

go/libkb/user.go
```diff
@@ -639,12 +653,12 @@ func (u *User) localDelegatePerUserKey(perUserKey keybase1.PerUserKey) error {
 	return nil
 }
 
-func (u *User) SigChainBump(linkID LinkID, sigID keybase1.SigID) {
-	u.SigChainBumpMT(MerkleTriple{LinkID: linkID, SigID: sigID})
+func (u *User) SigChainBump(linkID LinkID, sigID keybase1.SigID, isHighDelegator bool) {
```
patrickxb: could you add a comment explaining what `isHighDelegator` means?

go/libkb/kbsig.go
```diff
@@ -252,6 +252,32 @@ func remoteProofToTrackingStatement(s RemoteProofChainLink, base *jsonw.Wrapper)
 	return nil
 }
 
+type HPrevInfo struct {
+	Seqno keybase1.Seqno
+	Hash  LinkID
+}
```
patrickxb: Could you add a comment describing what the `HPrevInfo` type is?  I can't tell from the name.  I'm guessing `H` is for high?

go/teams/rename.go
```diff
@@ -75,12 +75,13 @@ func RenameSubteam(ctx context.Context, g *libkb.GlobalContext, prevName keybase
 
 		g.Log.CDebugf(ctx, "RenameSubteam make sigs")
 		renameSubteamSig, err := generateRenameSubteamSigForParentChain(
-			g, me, deviceSigningKey, parent.chain(), subteam.ID, newName, admin)
+			libkb.NewMetaContext(ctx, g), g, me, deviceSigningKey, parent.chain(), subteam.ID, newName, admin)
```
patrickxb: you should only need to call `libkb.NewMetaContext` once here.  and also, there's no need to also pass in `g` if you are passing in a `libkb.MetaContext`.

go/teams/rename.go
```diff
@@ -75,12 +75,13 @@ func RenameSubteam(ctx context.Context, g *libkb.GlobalContext, prevName keybase
...
 		if err != nil {
 			return err
 		}
 
 		renameUpPointerSig, err := generateRenameUpPointerSigForSubteamChain(
+			libkb.NewMetaContext(ctx, g),
```
patrickxb: see above.

there's no need to also pass in `g` if you are passing in a `libkb.MetaContext`.

go/teams/rename.go
```diff
@@ -115,7 +116,7 @@ func RenameSubteam(ctx context.Context, g *libkb.GlobalContext, prevName keybase
 	})
 }
 
-func generateRenameSubteamSigForParentChain(g *libkb.GlobalContext, me libkb.UserForSignatures, signingKey libkb.GenericKey, parentTeam *TeamSigChainState, subteamID keybase1.TeamID, newSubteamName keybase1.TeamName, admin *SCTeamAdmin) (item *libkb.SigMultiItem, err error) {
+func generateRenameSubteamSigForParentChain(m libkb.MetaContext, g *libkb.GlobalContext, me libkb.UserForSignatures, signingKey libkb.GenericKey, parentTeam *TeamSigChainState, subteamID keybase1.TeamID, newSubteamName keybase1.TeamName, admin *SCTeamAdmin) (item *libkb.SigMultiItem, err error) {
```
patrickxb: remove `g *libkb.GlobalContext` parameter.  use `m.G()` where necessary.

go/teams/rename.go
```diff
@@ -176,7 +183,7 @@ type chainPair struct {
 	subteam *TeamSigChainState
 }
 
-func generateRenameUpPointerSigForSubteamChain(g *libkb.GlobalContext, me libkb.UserForSignatures, signingKey libkb.GenericKey, teams chainPair, newSubteamName keybase1.TeamName, admin *SCTeamAdmin) (item *libkb.SigMultiItem, err error) {
+func generateRenameUpPointerSigForSubteamChain(m libkb.MetaContext, g *libkb.GlobalContext, me libkb.UserForSignatures, signingKey libkb.GenericKey, teams chainPair, newSubteamName keybase1.TeamName, admin *SCTeamAdmin) (item *libkb.SigMultiItem, err error) {
```
patrickxb: remove g *libkb.GlobalContext parameter. use m.G() where necessary.

protocol/avdl/keybase1/common.avdl
```diff
@@ -332,4 +332,9 @@ protocol Common {
     int x1;
     int y1;
   }
+
+  record HPrevInfo {
```
patrickxb: this doesn't look like it is used anywhere.  delete it?

go/teams/create.go
```diff
@@ -425,7 +428,7 @@ func makeRootTeamSection(teamName string, teamID keybase1.TeamID, members SCTeam
 	return teamSection, nil
 }
 
-func generateNewSubteamSigForParentChain(g *libkb.GlobalContext, me libkb.UserForSignatures, signingKey libkb.GenericKey, parentTeam *TeamSigChainState, subteamName keybase1.TeamName, subteamID keybase1.TeamID, admin *SCTeamAdmin) (item *libkb.SigMultiItem, err error) {
+func generateNewSubteamSigForParentChain(m libkb.MetaContext, g *libkb.GlobalContext, me libkb.UserForSignatures, signingKey libkb.GenericKey, parentTeam *TeamSigChainState, subteamName keybase1.TeamName, subteamID keybase1.TeamID, admin *SCTeamAdmin) (item *libkb.SigMultiItem, err error) {
```
patrickxb: remove `g` parameter.  use `m.G()`.

go/libkb/kbsig.go
heronhaye: We're going to change this to HighSkip before merging

go/libkb/chain_link_v2_test.go
```diff
@@ -110,3 +111,57 @@ func TestOuterLinkV2WithMetadataPointerContainerDecode(t *testing.T) {
...
+
+	o := serdeOuterLink(t, tc, nil, nil)
+	require.True(t, o.HighSkipSeqno == nil)
+	require.True(t, o.HighSkipHash == nil)
+	require.NoError(t, highSkipMatch(o, nil))
+
+	seqno := keybase1.Seqno(3)
+	o = serdeOuterLink(t, tc, &seqno, nil)
+	require.True(t, *o.HighSkipSeqno == 3)
+	require.True(t, o.HighSkipHash == nil)
```
xgess: this can't actually happen though: a highskipseqno of 3 and a highskiphash of nil. i'd argue we should use realistic cases for our happypath tests even if the code being tested doesn't verify it.

go/libkb/kbsig.go
```diff
@@ -520,7 +586,13 @@ func MakeSig(
 	case KeybaseSignatureV2:
 		prevSeqno := me.GetSigChainLastKnownSeqno()
 		prevLinkID := me.GetSigChainLastKnownID()
+		highSkip, highSkipErr := me.GetExpectedNextHighSkip()
+		if highSkipErr != nil {
+			return sig, sigID, linkID, highSkipErr
+		}
+		/* TODO Should this be nullable for kex..? */
```
xgess: this is fixed elsewhere now, yes?

go/teams/chain.go
```diff
@@ -1602,9 +1636,43 @@ func (t *teamSigchainPlayer) addInnerLink(
...
+	highSkipVerifyDesired := !isInflate && !precheck
+	// If presented with an highSkip, and we are unable to verify it
+	// but would like to, throw a CantVerifyNewLinkAttrsError.
+	if highSkipVerifyDesired && reverifyNeeded {
+		t.G().GetLog().Debug("throwing CantVerifyNewLinkAttrsError for high_skip")
+		return res, NewCantVerifyNewLinkAttrsError(payload.Seqno)
+	}
+	// We only run addInnerLink out of order when inflating, and so we skip
+	// computing high skips during it.
+	// updates high prev data, so we skip it. If highSkip is presented and
```
xgess: `high prev` should be `high skip`

go/teams/loader.go
```diff
@@ -437,6 +444,15 @@ func (l *TeamLoader) load2InnerLockedRetry(ctx context.Context, arg load2ArgT) (
...
 		ret = l.storage.Get(ctx, arg.teamID, arg.public)
+		// as part of rollout of signchain v2.3, if a reverify is needed
+		// then treat it as a cache miss and ask for a full reload.
+		if ret != nil {
+			teamSigChainState := TeamSigChainState{inner: ret.Chain}
+			if teamSigChainState.ReverifyNeededForHighSkips() {
+				arg.forceFullReload = true
+				ret = nil
+			}
+		}
```
mlsteele: Move to `load2DecideRepoll` and return `discardCache` if possible. Arg doesn't need to be modified as it's not used later.

But also.. the difference between all this cache adapting and just busting the cache. Is it just that teams with 1 sigchain link are not busted?

go/teams/chain_parse.go
```diff
@@ -159,6 +159,23 @@ func (link *SCChainLink) UnmarshalPayload() (res SCChainLinkPayload, err error)
 	return res, err
 }
 
+type HighSkipStr struct {
+	Seqno keybase1.Seqno
+	Hash  *string
+}
```
mlsteele: Does `Str` mean this struct is a struct, not string? Could name SCHighSkip.

go/teams/loader.go
xgess: def agree about moving the code and using the infrastructure that's already there.

go/teams/chain_parse.go
mlsteele: Probably want to serialize with lowercase fields with `json:` tags

go/teams/loader.go
mlsteele: oh yeah forgot about that, that it's already repairing itself. Nice!

go/teams/sig.go
```diff
@@ -262,61 +281,69 @@ func seqTypeForTeamPublicness(public bool) keybase1.SeqType {
...
+		state, err = playSigItem(ctx, g, me, state, sigItem, true)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func playSigItem(ctx context.Context, g *libkb.GlobalContext, me keybase1.UserVersion, state *TeamSigChainState, sigMultiItem libkb.SigMultiItem, precheck bool) (*TeamSigChainState, error) {
```
mlsteele: Comment that this eats the state. Like `AppendChainLink`
```
// `state` is moved into this function. There must exist no live references into it from now on.
```

go/teams/chain.go
```diff
@@ -79,6 +79,39 @@ func (t TeamSigChainState) GetLatestHighLinkID() keybase1.LinkID {
...
+func (t TeamSigChainState) GetHighSkipIfValid() (*libkb.HighSkip, error) {
+	if !t.ReverifyNeededForHighSkips() {
+		highSkip, err := t.GetHighSkip()
+		if err != nil {
+			return nil, err
+		}
+		return &highSkip, nil
+	}
+	return nil, nil
+}
```
mlsteele: For a state at seqno 1 loaded from an old cache, I think this could return default values but it should return seqno=1.

go/teams/chain.go
```diff
@@ -1602,9 +1645,35 @@ func (t *teamSigchainPlayer) addInnerLink(
...
-		res.newState.inner.LastHighLinkID = link.LinkID().Export()
-		res.newState.inner.LastHighSeqno = link.Seqno()
+
+	// We only run addInnerLink out of order when inflating, and so we skip
+	// computing high skips during it.
+	// updates high skip data, so we skip it. If highSkip is presented and
+	// either we don't need to reverify or we want to check it, check it.
+	highSkip := payload.HighSkip
+	if highSkipVerifyDesired {
+		if highSkip != nil && !reverifyNeeded {
```
mlsteele: I think `&& !reverifyNeeded` is redundant here. May want to remove it or if you're feeling real cautious error out instead of skipping verification.

go/teams/chain.go
```diff
@@ -1602,9 +1645,35 @@ func (t *teamSigchainPlayer) addInnerLink(
...
+				}
+				if *highSkip.Hash != string(lastHighLinkID) {
+					return res, fmt.Errorf("Expected HighSkipHash=%s, got %s", string(lastHighLinkID), *highSkip.Hash)
+				}
+			}
+		}
+		if isHighLink {
+			res.newState.inner.LastHighLinkID = link.LinkID().Export()
+			res.newState.inner.LastHighSeqno = link.Seqno()
+		}
```
mlsteele: You might pull this green section into its own function to avoid adding to the behemoth? I don't care though.

go/teams/chain.go
```diff
@@ -1602,9 +1645,35 @@ func (t *teamSigchainPlayer) addInnerLink(
...
 		}
 	}
-	if isHighLink {
-		res.newState.inner.LastHighLinkID = link.LinkID().Export()
-		res.newState.inner.LastHighSeqno = link.Seqno()
+
+	// We only run addInnerLink out of order when inflating, and so we skip
+	// computing high skips during it.
+	// updates high skip data, so we skip it. If highSkip is presented and
+	// either we don't need to reverify or we want to check it, check it.
```
mlsteele: This comment is jumbled. May also want to add that the new link's high skip slot is always optional.

mlsteele:
Teams part looks good!

xgess:
@heronhaye what do we want to verify before merging this? let's block off some time and push it over the finish line.

heronhaye:
Miles' PR has a good list

    modern resets

    revocations

    web_service_binding that are eldest (danl 7)

    web_service_binding that are eldest on non-current subchains that are missing from the db

xgess:
for convenience, here is that commit: https://github.com/keybase/client/commit/d75f95b9918a726698ea86877f86f2230b5fb8ac

go/libkb/kbsig.go
```diff
@@ -267,6 +293,10 @@ type ProofMetadata struct {
 	SeqType             keybase1.SeqType
 	MerkleRoot          *MerkleRoot
 	IgnoreIfUnsupported SigIgnoreIfUnsupported
+	// HighSkipFallback is used for teams to provide team chain highSkips and
```
xgess: might want to remove this comment since it's not relevant anymore

xgess: we can also do it in a later PR if it's the only thing outstanding

go/libkb/sig_chain.go
```diff
@@ -809,9 +867,9 @@ func (c ChainLinks) omittingNRightmostLinks(n int) ChainLinks {
 
 // VerifySigsAndComputeKeys iterates over all potentially all incarnations of the user, trying to compute
 // multiple subchains. It returns (bool, error), where bool is true if the load hit the cache, and false otherwise.
-func (sc *SigChain) VerifySigsAndComputeKeys(m MetaContext, eldest keybase1.KID, ckf *ComputedKeyFamily) (bool, error) {
+func (sc *SigChain) VerifySigsAndComputeKeys(m MetaContext, eldest keybase1.KID, ckf *ComputedKeyFamily, uid keybase1.UID) (bool, error) {
```
xgess: i dont totally understand why we need to pass around the `uid` here as well, but just spoke with @heronhaye and NAB

PR #13443, opened 2018-08-24
Stellar: getPaymentDetailsLocal changes for chat

zapu:
- Change `getPaymentDetailsLocal` to accept nullable `accountID` parameter. Chat previews do not know account IDs, so if they want unformatted message (because service will not know whether it's a debit or credit without knowing `accountID`), they can pass `nil` as `accountID`.
- Add `TxID` to `SendPaymentResLocal`. No reason not to, service already had it, and it helped me with debugging. I'll also be fine without that change if anyone objects.
- Add `TestPaymentDetailsEmptyAccId` test that simulates how chat would pull payment details with nil `accountID`.


zapu:
cc @buoyad

go/stellar/stellarsvc/frontend.go
```diff
@@ -1282,6 +1288,7 @@ func (s *Server) SendPaymentLocal(ctx context.Context, arg stellar1.SendPaymentL
 	return stellar1.SendPaymentResLocal{
 		KbTxID:  sendRes.KbTxID,
 		Pending: sendRes.Pending,
+		TxID:    sendRes.TxID,
```
patrickxb: @buoyad do you want this?

zapu: I can remove this. I needed it for debugging one time while working on tests for this PR.

patrickxb: If you want to keep it, how about returning a PaymentID?

buoyad: Frontend only uses PaymentID

patrickxb: Ok, @zapu can you either remove or change to PaymentID?  thanks.

zapu: will do, thanks!

PR #13457, opened 2018-08-27
Stellar: make FormatAmount removes trailing zeroes in 8-digit precision mode

zapu:
This is a rather bold one.

It changes `stellar.FormatAmount` to remove trailing zeroes, which can also remove fractional part entirely. E.g. `123.4560000` would be formatted as `123.456`, or `123.0000000` as `123`.

This behavior replaces `libkb.StellarSimplifyAmount`. Callers can now use `stellar.FormatAmount` and not make a decision about how number should be formatted.

Since we are dealing with money here, maybe it's desired to keep at least `n` digits after point, even if they are trailing zeroes. FWIW the `twoPrecision` mode is untouched, so formatting `12.10` will still end up being `12.10`. Let's discuss.

The alternative is to add another argument to `FormatAmount` to trigger this behavior. I'm fine with both but I'm weakly leaning towards the one in this PR.


zapu:
cc @buoyad

patrickxb:
Can you fill me in on the reason for this PR?

zapu:
There was a report that returning a description like that to GUI is not wanted:
```
{
   "amount" : "3.1",
   "amountDescription" : "3.1000000 USD",
   "currency" : "USD",
}
```
and Miles said that

> FormatAmount should be taught how to do this and StellarSimplifyAmount removed

so I went with that. But it turned out to be a fair number of changes and I'm not sure if we want to truncate zeroes everywhere. I'm fine with still using `StellarSimplifyAmount` or something in chosen places, but on the other hand maybe it's good if an app is consistent in how it displays numbers.

zapu:
Superseded by https://github.com/keybase/client/pull/13471

PR #13458, opened 2018-08-27
Correctly deselect blocked conversations

thebearjew:
### Problem

Conversations: 

```
A: (most recent)
B: (2 minutes ago)
```

When attempting to block conversation `A` , `A` would be reselected and the conversation thread for `A` would remain fixed on the screen until clicking on another conversation.

```
A: (most recent)
```

If `A` is the only conversation, then after blocking the thread will remain stuck with no way to clear it.
The user will have to completely restart the Keybase GUI and service.

screen shot 2018-08-27 at 9 53 53 am

### Solution

When `metaDelete` action was coming in to the Saga, we would bail early and not auto-select the newest conversation after removal. This update prevents bailing if the action is `findNewestConversation` or `selectSomethingElse`.

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -1489,7 +1489,11 @@ const _maybeAutoselectNewestConversation = (
     action.payload.conversationIDKey === selected
   ) {
     // Intentional fall-through -- force select a new one
-  } else if (Constants.isValidConversationIDKey(selected)) {
+  } else if (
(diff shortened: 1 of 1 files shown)


chrisnojima:
this doesn't feel like the right solution to me. the blocked item should be removed, which should kick off the reselection process if it was the one selected

thebearjew:
@chrisnojima

thebearjew:
@chrisnojima Okay sounds good.

PR #13460, opened 2018-08-27
set global limit on number of parts that can be simultaneously uploaded CORE-8604

mmaxim:
If a user on a connection with some kind of bottleneck on it tries to upload a ton of data, Keybase can hose that connection by attempting too many part uploads at once. This is only in play if the user uploads a lot of files at once (which skirts around the 10 wide pipeline in `S3Stores). Patch does the following:

1.) Adds a new type `s3UploadPipeliner` which is global to the `attachments` package. It keeps track of how many part uploads are in flight, and will block requests that go over the limit.
2.) Change `putMultiPipeline` to start a goroutine per block, but be limited by the global `s3UploadPipeliner`.

cc @malgorithms @maxtaco

### DIFF
--- a/go/chat/attachments/s3.go
+++ b/go/chat/attachments/s3.go
@@ -8,12 +8,60 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sync"
 
 	"github.com/keybase/client/go/chat/s3"
 	"github.com/keybase/client/go/chat/types"
 	"github.com/keybase/client/go/protocol/chat1"
 	"golang.org/x/sync/errgroup"
 )
+
+const s3PipelineMaxWidth = 25
+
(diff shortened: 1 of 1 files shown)


go/chat/attachments/s3.go
```diff
@@ -138,17 +186,22 @@ func (a *S3Store) putMultiPipeline(ctx context.Context, r io.Reader, size int64,
...
+				defer s3UploadPipeline.Complete()
 				if err := a.uploadPart(ectx, task, b, previous, previousParts, multi, retCh); err != nil {
 					return err
 				}
-			}
-			return nil
-		})
-	}
-
+				return nil
```
patrickxb: you could just `return a.uploadPart(...)`

PR #13463, opened 2018-08-27
Stellar send to another account integration

nathunsmitty:
Send money from one account to another. I also tried to get some style bugs in the send form and confirm form, most notably adding ellipses to long account names/ids.


shared/actions/wallets.js
```diff
@@ -182,6 +179,10 @@ const navigateToAccount = (
     // Link existing failed, don't nav
     return
   }
+  return Saga.put(Route.navigateUp())
```
nathunsmitty: Since we need the create new and link existing popups to return to the send form if triggered from the send form, I think the best way to handle this is to have them navigate up vs to the accounts page?

nathunsmitty:
@keybase/react-hackers ready for review!

shared/actions/wallets.js
nathunsmitty: Something to note about this as its currently implemented though is that it will navigate you to the new linked or created account after the send form closes. Not sure if this is the desired behavior.

PR #13465, opened 2018-08-27
WIP unhide status bar when exiting video in chat

buoyad:
- [ ] Test on all platforms

### DIFF
--- a/shared/chat/conversation/attachment-video-fullscreen/index.native.js
+++ b/shared/chat/conversation/attachment-video-fullscreen/index.native.js
@@ -2,10 +2,18 @@
 import React from 'react'
 import {WebView, Box, ClickableBox, Text} from '../../../common-adapters'
 import {globalColors, platformStyles, globalMargins, globalStyles} from '../../../styles'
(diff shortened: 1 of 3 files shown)


shared/chat/conversation/attachment-video-fullscreen/index.native.js
```diff
@@ -2,10 +2,18 @@
...
+import {
+  NativeStatusBar,
+  OverlayParentHOC,
+  type OverlayParentProps,
+} from '../../../common-adapters/mobile.native'
 import type {Props} from './index.types'
 
 class _VideoFullscreen extends React.Component<Props & OverlayParentProps> {
+  componentWillUnmount() {
+    NativeStatusBar.setHidden(false)
```
chrisnojima: should we store the value coming in, just in case its not hidden / etc?

buoyad: React native `StatusBar` doesn't have a way to query whether it's hidden. I don't think this would do much harm, a little routing within the app should fix it if this leads to a bad state.

chrisnojima: K

PR #13466, opened 2018-08-27
no more default title for image/video attachments CORE-8692

mmaxim:
@keybase/react-hackers  cc @malgorithms 

Patch does the following:

1.) Remove `fileName` from the default entry in the "Title" text box when uploading attachments. 
2.) Change the image attachment container to not sub in `fileName` if no title is present.
3.) Change inbox snippet for title-less image/videos to indicate the type of the attachment.
4.) Change around the upload box on desktop to look better.
5.) **Bonus Fix:** saving videos broke on mobile after my streaming changes, and I didn't realize until I tried to test this PR, so fix that up.


shared/constants/chat2/message.js
```diff
@@ -922,6 +926,12 @@ export const messageExplodeDescriptions: Types.MessageExplodeDescription[] = [
   {text: 'Never explode (turn off)', seconds: 0},
 ].reverse()
 
+export const getFilename = (path: string) => {
```
chrisnojima: maybe you should use helpers from constants/fs for this

mmaxim: Eh, this works well enough for what it needs to do. It was just hanging out in some container before.

chrisnojima: but you can just call getLocalPathName  right?

mmaxim: ok

PR #13468, opened 2018-08-28
Stellar: refactor formatting, change FormatAmount to keep decimal in some cases

zapu:
Formatting function bas been moved to `stellar` package into separate file, same with tests.

When in 7-digit precision mode and formatting adds a thousands separator, do not truncate all zeroes in case of 7-zero fractional part, but keep two zeroes instead.

This has been pointed out by Chris Coyne. If we add a thousands separator, we should keep the decimal point as well so it's not ambiguous.


zapu:
let's close this for now, I'm working on a PR that does not move everything around and therefore is more manageable.

PR #13469, opened 2018-08-28
cannot skip HomeScreenTodoType_NONE

joshblum:
changes to client tests to match server change: https://github.com/keybase/keybase/pull/2817


joshblum:
tests passed locally, merging so others can rebase

PR #13470, opened 2018-08-28
Don't black bar for 'EOF from server' error

buoyad:
r? @keybase/react-hackers


songgao:
Does this EOF error resolve itself after a while?

buoyad:
@songgao yeah, they happen when the service restarts, and are usually transient. Before this you'd see the reachability screen and then a black bar on reconnecting. This removes the black bar.

songgao:
Cool; thanks. I'll be able close some KBFS ticket too then!

shared/reducers/config.js
```diff
@@ -112,6 +113,16 @@ export default function(state: Types.State = initialState, action: ConfigGen.Act
       const {globalError} = action.payload
       if (globalError) {
         logger.error('Error (global):', globalError)
+        if (
```
chrisnojima: i think we should move this into util/error and have some kind of transform there. it can return null if its this case and we'll ignore it

buoyad: if we move it to `convertToError`, the calling saga won't be able to tell that the RPC failed. is that what we want?

chrisnojima: we don't need to use convertToError per se, i'm just saying we should move this kind of logic to util/error

buoyad: @chrisnojima got it, will do

buoyad:
@chrisnojima I moved the check to a new function `isErrorTransient`, PTAL

PR #13472, opened 2018-08-28
Include senderID, recipientID in MessageSendPayment

patrickxb:
cc @buoyad


protocol/avdl/chat1/local.avdl
```diff
@@ -174,6 +174,8 @@ protocol local {
 
   record MessageSendPayment {
     stellar1.PaymentID paymentID;
+    stellar1.AccountID senderID;
+    stellar1.AccountID recipientID;
```
mlsteele: RecipientID is optional? Frontend can handle when it's blank?

patrickxb: It is always filled in currently.  There's just a nil check in the code on the Recipient structure so we don't crash.

patrickxb: It will be the relay account id

mlsteele: That would be weird. That means if you receive a relay payment and then call `getPaymentDetailsLocal(relay account, payment id)` you'd get back delta=decrease.

patrickxb: Then i'll just leave it blank for relay payments and frontend will have to deal.

@zapu made a change to make an empty account id work for payment details.

PR #13476, opened 2018-08-28
Allow backspacing across lines in terminal

heronhaye:
Linux Demo: https://asciinema.org/a/BfvzbZnYK4imPwKR0d4wVOzrg
Probably want to try on Mac and Windows too. I did some basic smoketesting and everything looks OK.

I don't think we can get support for Ctrl-U across lines, seems it only supports color codes.

We might be able to modify it to support an arbitrary width/height. By default its 80x24 with no option to set it anywhere, but width and height are internal variables. At that point it may be better to use another terminal library.

Also, if your terminal is smaller than 80 chars wide or you resize it, it's completely broken, but that was true before as well.

### DIFF
--- a/go/client/terminal.go
+++ b/go/client/terminal.go
@@ -55,12 +55,18 @@ func (t *Terminal) PromptPassword(s string) (string, error) {
 	if err := t.open(); err != nil {
 		return "", err
 	}
+	if t.escapeWrites {
+		s = terminalescaper.Clean(s)
+	}
 	return t.engine.PromptPassword(s)
 }
 
 func (t *Terminal) Prompt(s string) (string, error) {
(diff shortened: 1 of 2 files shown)


heronhaye:
@patrickxb, are you familiar with this part?

patrickxb:
doesn't this undo @AMarcedone work?  I never reviewed that stuff.  @maxtaco?

heronhaye:
https://github.com/keybase/client/pull/12208

mmaxim:
I reverted this to fix kbfs-server tests.

mmaxim:
Also did anyone address @patrickxb's question?

https://github.com/keybase/client/pull/13476#issuecomment-416747003

PR #13477, opened 2018-08-28
TeamWritePermDeniedError

joshblum:
if you tried to modify git settings for a team where you were a writer or below you used to get:
`▶ ERROR DB error (code 2625)`
now it says:
`▶ ERROR permission denied to modify team`

is there an easier way to pipe the error message from the server directly?


patrickxb:
I'm not sure what you mean by easier.  you could supply the code in the api arg in the AppStatusCodes field, then it will get handled.  Is that what you mean?

joshblum:
The server returns an error message but it's not show outside of devel, was wondering if there was a way to get it to show without a client change.. `AppStatusCodes` is a little simpler than before though

PR #13484, opened 2018-08-29
decrement pipeCheckFail handshake wait on success

zanderz:
It seems there was no matching decrement for this handshake wait.
This makes it work on Windows, but DESKTOP-7675 could still be causing problems too.


shared/actions/platform-specific/index.desktop.js
```diff
@@ -132,16 +132,19 @@ const checkRPCOwnership = () =>
...
-              if (result.isOwner) {
-                resolve()
+              const resultjs = JSON.parse(stdout.toString())
+              if (resultjs.isOwner) {
+                resolve(resultjs)
                 return
               }
               logger.info(`pipeowner check result: ${stdout.toString()}`)
-              reject(new Error('pipeowner check failed'))
+              throw new Error('pipeowner check failed')
```
zanderz: Would this do it - decrement the handshake wait in the catch{} below?

Generally I struggled trying to invoke the handhsakewait inside the promise, it just didn't seem to work. Open to suggestions.

PR #13485, opened 2018-08-29
undo wait on success for windows pipe fail

chrisnojima:
@keybase/react-hackers @zanderz this is what we want

### DIFF
--- a/shared/actions/platform-specific/index.desktop.js
+++ b/shared/actions/platform-specific/index.desktop.js
@@ -152,7 +152,7 @@ const checkRPCOwnership = () =>
       yield Saga.put(
         ConfigGen.createDaemonHandshakeWait({
           failedFatal: true,
-          failedReason: e.message,
+          failedReason: e.message || 'windows pipe owner fail',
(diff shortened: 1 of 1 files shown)


chrisnojima:
(i didn't test this, can you @zanderz )

shared/actions/platform-specific/index.desktop.js
```diff
@@ -142,6 +142,12 @@ const checkRPCOwnership = () =>
             })
           })
       )
+      yield Saga.put(
```
zanderz: Doesn't this need to be done conditionally? What if result.isOwner is false (the security hole the pipe check is meant to cover)?

chrisnojima: there is no catch handler on this promise, so when it throws the catch below catches and it goes to the error flow. you can see this if you just change the promise to
```
      yield Saga.call(
        () =>
          new Promise((resolve, reject) => {
            reject('MOCK')
          })
      )

```

you'll see the failedFatal handshake fires

PR #13488, opened 2018-08-29
Fixed font-size in mobile more menu

cecileboucheron:
@keybase/react-hackers


shared/settings/nav/settings-item.js
```diff
@@ -27,7 +27,7 @@ export default function SettingsItem(props: SettingsItemProps) {
         />
       )}
       <Text
-        type={'BodySmallSemibold'}
+        type={'BodySemibold'}
```
chrisnojima: don't wrap in {}, just 'BodySemibold'

PR #13490, opened 2018-08-29
Stellar: show send form when clicking send button from chat

buoyad:
Uses `setBuilding*` to set up the values in the form. Since our inputs aren't controlled, values won't be filled but corresponding errors will show when they should. Also adds a call to `loadAccounts` on login. r? @keybase/react-hackers

### DIFF
--- a/shared/wallets/send-form/container.js
+++ b/shared/wallets/send-form/container.js
@@ -1,11 +1,16 @@
 // @flow
 import SendForm from '.'
+import * as WalletsGen from '../../actions/wallets-gen'
 import {connect, type TypedState} from '../../util/container'
 
 const mapStateToProps = (state: TypedState) => ({})
 
 const mapDispatchToProps = (dispatch, {navigateUp}) => ({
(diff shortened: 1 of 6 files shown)


shared/wallets/send-form/container.js
```diff
@@ -1,11 +1,16 @@
...
+import * as WalletsGen from '../../actions/wallets-gen'
 import {connect, type TypedState} from '../../util/container'
 
 const mapStateToProps = (state: TypedState) => ({})
 
 const mapDispatchToProps = (dispatch, {navigateUp}) => ({
-  onClose: () => dispatch(navigateUp()),
+  onClose: () => {
+    dispatch(navigateUp())
+    dispatch(WalletsGen.createClearBuildingPayment())
```
chrisnojima: maybe better to handle this stuff as a saga. if you look at signup/login/ etc they watch for nav up and cancel things

buoyad:
@chrisnojima I added an `abandonPayment` action to handle the navigation. If that looks ok feel free to merge

PR #13491, opened 2018-08-29
Stellar Integration: Search Keybase user in send form

nathunsmitty:
If you want to test this out, my keybase username is nathunsmitty. 😆 

@keybase/react-hackers r?


shared/search/user-input/container.js
```diff
@@ -148,20 +150,25 @@ export type Props = _Props & {
 }
 
 const ConnectedUserInput = compose(
-  connect(mapStateToProps, mapDispatchToProps, (s, d, o) => ({...o, ...s, ...d})),
+  connect(mapStateToProps, mapDispatchToProps, (s, d, o: OwnProps) => ({...o, ...s, ...d})),
```
nathunsmitty: Was getting some errors here if I didn't type ownProps.

shared/wallets/confirm-form/participants/container.js
```diff
@@ -17,6 +17,7 @@ const mapStateToProps = (state: TypedState) => {
     fromAccountName: '',
     fromAccountContents: '',
     recipientUsername: built.toUsername,
+    recipientFullName: 'placeholder', // TODO: get a way to get full names from usernames
```
nathunsmitty: One issue I kept running into during this PR was receiving a user's full name, as in the mockups. It'd be really nice to have a user's full name returned by the built payment RPC, but I'm not sure if there'd be a better place to get that info?

shared/wallets/search/index.js
```diff
@@ -1,50 +0,0 @@
-// @flow
```
nathunsmitty: The search story got merged into the `SendForm/participants` story if you're looking for it.

shared/wallets/send-form/participants/search.js
```diff
@@ -0,0 +1,82 @@
...
+const styles = Styles.styleSheetCreate({
+  row: {
+    paddingBottom: 0,
+  },
+  list: Styles.platformStyles({
+    isElectron: {
+      position: 'absolute',
+      top: 100,
+      zIndex: 4,
+      backgroundColor: Styles.globalColors.white,
```
nathunsmitty: The styling for the search overlay could probably be improved, I'm not totally satisfied with the `top: 100` but couldn't think of a better way to have the search results appear below the header and search input with maintaining the absolute positioning. Open to suggestions!

shared/wallets/send-form/participants/to-field.js
```diff
@@ -109,32 +125,34 @@ class ToField extends React.Component<ToFieldProps, ToFieldState> {
             onChanged={this.onDropdownChange}
             items={items}
             selected={
-              this.state.selectedAccount ? (
-                <SelectedEntry account={this.state.selectedAccount} user={this.props.user} />
+              this.state.selectedRecipient instanceof Object ? (
+                <SelectedEntry account={this.state.selectedRecipient} user={this.props.user} />
               ) : (
-                undefined
+                <DropdownText key="placeholder-select" text="Pick another account" />
```
nathunsmitty: Add some placeholder text I missed in the #13463

shared/wallets/send-form/participants/search.js
buoyad: Adding `position: relative` to the wrapping `Box2` might let you get rid of the `top` here.

shared/wallets/send-form/participants/to-field.js
buoyad: This placeholder text isn't rendering for me

shared/wallets/send-form/participants/to-field.js
```diff
@@ -42,45 +56,46 @@ class ToField extends React.Component<ToFieldProps, ToFieldState> {
...
+    }
+  }
 
+  render() {
     let component
 
-    if (this.props.recipientType === 'keybaseUser' && this.props.recipientUsername) {
+    // There are a few different ways the participants form can look:
+    // Case 1: A user has been set, so we display their name and avatar
+    // We can only get this case when the recipient is a stellar address or searched user, not another account.
```
buoyad: wait so, does this happen for a stellar address or does the case on L138 happen?

shared/wallets/send-form/participants/to-field.js
nathunsmitty: Hmm what do you see when you open the send form for send to another account?

shared/common-adapters/name-with-icon.js
```diff
@@ -82,7 +84,7 @@ const NameWithIconVertical = (props: Props) => {
         )}
         {!!props.username && (
           <ConnectedUsernames
-            clickable={true}
+            clickable={!props.onClick}
```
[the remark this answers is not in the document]
nathunsmitty: But what if that's not the desired behavior? I want to give `NameWithIcon` to open the tracker, not the profile. If `clickable={true}` then it'll navigate to the profile and open the tracker.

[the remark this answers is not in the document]
chrisnojima: we can just make
{
  onClick: () => void | 'profile' | 'tracker'
}

shared/wallets/send-form/participants/search.js
nathunsmitty: `position: relative` doesn't work because then the list would be part of the flexbox and we want to overlay the flexbox 🙁

shared/wallets/send-form/participants/to-field.js
nathunsmitty: It depends on if the address typed in is associated with a keybase user. If yes, then the username should replace the address [as per the mockup](https://zpl.io/aXz335K). If not, then we just show the address input (L138)

shared/wallets/confirm-form/participants/container.js
```diff
@@ -21,6 +21,7 @@ const mapStateToProps = (state: TypedState) => {
     recipientAccountAssets: recipientAccount.balanceDescription,
     recipientAccountName: recipientAccount.name || recipientAccount.accountID,
     recipientUsername: built.toUsername,
+    recipientFullName: 'placeholder', // TODO: get a way to get full names from usernames
```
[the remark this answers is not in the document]
buoyad: I think falling back to an empty string is ok. `infoMap` is sourced from users you've interacted with in chat, so there's a good chance it'll hit in this case

shared/wallets/send-form/participants/search.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  closeResultsList = () => this.setState({displayResultsList: false})
+
+  render() {
+    return (
+      <React.Fragment>
+        <ParticipantsRow heading="To" style={styles.row} headingStyle={styles.rowHeading}>
+          <Box2 direction="vertical" fullWidth={true}>
+            <UserInput
```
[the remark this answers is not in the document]
nathunsmitty: The (x) is actually a part of the UserInput component itself. I'll add a prop to hide it.

Issue #13450, opened 2018-08-25
Send/recieve messages heavily delayed after enabling auto delete, and setting messages to explode..


restingmurderface:
I’m using Keybase in iOS 11.4.1, on an iPhone 6s.  Yesterday after clearing my conversation history, setting everything to auto delete after thirty days, and future messages to auto explode after three days, the app got rather vexed and slowed right down. Both my partner and I are now both having problems sending and receiving messages from one another in chat. We get notifications, but messages won’t load, or they take forever to load. Messages either take too long to send, or don’t send at all and we have to submit multiple times to get them to go through. 

I turned off auto delete, exploding messages, logged out, logged back in, uninstalled and reinstalled, and the problem is still there. It’s basically unusable right now, what am I doing wrong? My internet connection is absolutely fine and super fast, so that’s not the problem here. What did I break and how do I fix it? 

Please and thank you in advance.

restingmurderface:
Anyone?

mmaxim:
This is a problem with how we handle the delete history message. It seems like maybe on your older device the problem is magnified. We will have a fix in the next release, sorry! For now, the best workaround is probably just create a new conversation with the person (through a team might be the easiest way). Sorry about that!

restingmurderface:
Thanks for the info - I’ll look out for the new release and try the new conversation thing.

mmaxim:
New release will hopefully be out in the next week, so the fix should hopefully land soon. I'll be working on the fix in the next day or so, and will link the PR here.

PR #13494, opened 2018-08-29
respect the max fetch num when scanning local chat data CORE-8698

mmaxim:
If we have a thread with a bunch of delete history supersededBy messages (say like 10000), then we will fetch all 10K of them out and run them through the supersedes transformer. This is really slow, so cap the number of messages we can pull from local storage to `maxFetchNum` (which mimics the server). 

This should solve: https://github.com/keybase/client/issues/13450


mmaxim:
Yeah would have to be 1000 individually deleted messages, and like you said, that is already the case when fetching from the server.

PR #13495, opened 2018-08-29
Check previous isUserMessage instead of current

buoyad:
This fixes the author wrapper not showing up correctly for user messages that follow system messages. Previously,
```
previous && previous.author === message.author && Constants.isUserMessage(message)
```
this could be true if `previous` was a system message written by the current user.

Example before:
image

Example after:
image

r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/wrapper/wrapper-author/container.js
+++ b/shared/chat/conversation/messages/wrapper/wrapper-author/container.js
@@ -71,7 +71,7 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps) => {
   const {message, previous} = stateProps
 
   const sequentialUserMessages =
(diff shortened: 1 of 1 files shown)


shared/chat/conversation/messages/wrapper/wrapper-author/container.js
```diff
@@ -71,7 +71,7 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps) => {
   const {message, previous} = stateProps
 
   const sequentialUserMessages =
-    previous && previous.author === message.author && Constants.isUserMessage(message)
+    previous && previous.author === message.author && Constants.isUserMessage(previous)
```
chrisnojima: do we actually want both?

buoyad: Yeah

PR #13496, opened 2018-08-29
[WIP KBFS-3234] Widget File Row

jzila:
Not yet ready for review--will be ready once a follow-up is posted.


shared/common-adapters/usernames-remote-container.js
```diff
@@ -0,0 +1,34 @@
...
+const mergeProps = (stateProps, {_onUsernameClicked}, ownProps) => {
+  const following = new Set(stateProps.following)
+  const userData = ownProps.usernames
+    .map(username => ({
+      broken: stateProps.broken.trackerState === 'error',
+      following: following.has(username),
+      username,
+      you: stateProps.username === username,
+    }))
+    // .filter(u => !ownProps.skipSelf || !u.you)
```
chrisnojima: 🔥

shared/constants/fs.js
```diff
@@ -251,13 +251,17 @@ const itemStylesKeybase = {
...
 }
 
-const getIconSpecFromUsernames = (usernames: Array<string>, me?: ?string) => {
-  if (usernames.length === 1) {
-    return makeAvatarPathItemIconSpec(usernames[0])
-  } else if (usernames.length > 1) {
-    return makeAvatarsPathItemIconSpec(usernames.filter(username => username !== me))
-  }
-  return makeBasicPathItemIconSpec('iconfont-question-mark', unknownTextColor)
+const getIconSpecFromUsernames = (usernames?: ?Array<string>, me?: ?string) => {
```
chrisnojima: does it really have to have so many optional params?

shared/menubar/remote-proxy.desktop.js
```diff
@@ -67,6 +69,8 @@ const mergeProps = stateProps => ({
   loggedIn: stateProps.loggedIn,
   username: stateProps.username,
   conversations: stateProps.conversations,
+  broken: stateProps.broken,
```
chrisnojima: broken x 2?

shared/menubar/remote-proxy.desktop.js
```diff
@@ -67,6 +69,8 @@ const mergeProps = stateProps => ({
...
+  following: stateProps._following.toJS(),
```
chrisnojima: yeah we usually won't want to use toJS() since its recursive

shared/menubar/files-container.desktop.js
```diff
@@ -18,10 +30,21 @@ const mapDispatchToProps = dispatch => ({
 
 const mergeProps = (stateProps, dispatchProps) => ({
   onViewAll: dispatchProps.onViewAll,
-  tlfRows: stateProps._tlfRows.map(c => ({
-    onSelectPath: () => dispatchProps._onSelectPath(c.path),
-    path: FsTypes.pathToString(c.path),
-  })),
+  tlfRows: stateProps._tlfRows.map(c => {
+    const {participants, teamname} = FsUtil.tlfToParticipantsOrTeamname(FsTypes.pathToString(c.tlf))
```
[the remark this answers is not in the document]
jzila: I'm confused about this suggestion. Why wouldn't we want to pass the data to the component in a presentable form? Guidance is to make the component as dumb as possible.

shared/menubar/files-container.desktop.js
```diff
@@ -18,10 +30,21 @@ const mapDispatchToProps = dispatch => ({
...
+    return {
+      onSelectPath: () => dispatchProps._onSelectPath(c.tlf),
+      tlf: FsTypes.pathToString(c.tlf),
+      // Default to private visibility--this should never happen though.
+      tlfType: FsTypes.getPathVisibility(c.tlf) || 'private',
+      writer: c.writer,
+      participants: participants || [],
+      teamname: teamname || '',
+      iconSpec,
+      timestamp: TimestampUtil.formatTimeForConversationList(c.timestamp),
```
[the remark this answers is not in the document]
jzila: It's not really presentation, still content. And as above, I prefer to make the component as dumb as possible.

shared/menubar/remote-proxy.desktop.js
jzila: I don't follow. We can't serialize an immutable map/array, so `toJS()` makes sense here.

chrisnojima: we usually prefer toArray() or toObject(). toJS will return a totally mixed object which flow will leave you entirely untyped

jzila: But if we have an array of `I.RecordOf<Types._SomeType>`, then how will those serialize?

chrisnojima: yes, ultimately it will have to be serialized as plain js objects but its likely better to do something like
```
.toArray().map() or something vs toJS()
```

jzila: Understood, I'll do that then.

jzila: Actually, for `following`, that seems like overkill...

jzila: Yeah, `.toArray()` will just work here.

shared/menubar/remote-proxy.desktop.js
jzila: WDYM? This is `mergeProps`, where the previous one is in `mapStateToProps`.

jzila:
I'm going to implement all the PR feedback in the last PR, so I don't need to do the merge dance.

shared/constants/fs.js
jzila: I'll pass in an empty array instead.

jzila:
After all is done, I will also add in a refactor so that `GetNewestConvMetas` is in `mergeProps` instead of `mapStateToProps`.

shared/menubar/files-container.desktop.js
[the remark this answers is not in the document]
songgao: It's redundant information passed around, and it's really just one utility function call, which I think can still count as dumb. It's not like the component has to deal with immutable data structures, etc.

I think this is a similar situation where we make menu item action handlers in container, but populate actual menu item text presentation in components.

shared/menubar/remote-proxy.desktop.js
chrisnojima: oh github made this look like the same object

shared/menubar/files-container.desktop.js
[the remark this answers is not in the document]
jzila: I think I'm going to keep it the way it is. The data needs to be calculated anyway, and it doesn't make a difference for performance whether it happens here or there. And it's cleaner to populate the component directly from a prop rather than generating it. Finally, `participants` and `teamname` can both be undefined, so we're saving not just a function call but an `|| []`/`|| ''` check.

shared/menubar/files-container.desktop.js
```diff
@@ -18,10 +30,21 @@ const mapDispatchToProps = dispatch => ({
...
-    path: FsTypes.pathToString(c.path),
-  })),
+  tlfRows: stateProps._tlfRows.map(c => {
+    const {participants, teamname} = FsUtil.tlfToParticipantsOrTeamname(FsTypes.pathToString(c.tlf))
+    const iconSpec = FsConstants.getIconSpecFromUsernamesAndTeamname([c.writer], null, stateProps._username)
+    return {
+      onSelectPath: () => dispatchProps._onSelectPath(c.tlf),
+      tlf: FsTypes.pathToString(c.tlf),
+      // Default to private visibility--this should never happen though.
+      tlfType: FsTypes.getPathVisibility(c.tlf) || 'private',
```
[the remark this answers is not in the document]
jzila: That said, we don't need to be calling `FsTypes.pathToString(c.tlf)` twice.

shared/menubar/files-container.desktop.js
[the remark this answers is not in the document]
songgao: Performance is not my main concern. You have condition on `tlfType` in the component, and based on `tlfType` either use `participants` or `teamname`. `participants` and `teamname` have to be defined based on `tlfType` values, so the three things are tangled together with mental constraints that are not enforced by either flow or the language itself. So it's not clear from first look at the Props type how those would work with each other.

That being said, I think this is minor and I'm fine with the way you have now.

shared/menubar/remote-proxy.desktop.js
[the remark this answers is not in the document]
songgao: oops typo s/with/without

shared/menubar/files-container.desktop.js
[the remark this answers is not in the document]
jzila: I can probably refactor it to use `usernames.js`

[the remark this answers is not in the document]
songgao: I'd feel bad if my comment prompted you to spend too much time on this, so please feel free to ignore this if it becomes complicated!

jzila:
@chrisnojima need a thumb on this.

PR #13500, opened 2018-08-29
fix upload bug found by brayoh

songgao:


shared/actions/fs/index.js
```diff
@@ -619,6 +619,13 @@ function* openPathItem(action: FsGen.OpenPathItemPayload): Saga.SagaGenerator<an
   )
 }
 
+const uploadFanOut = (state: TypedState, action: FsGen.UploadsPayload) =>
+  Saga.sequentially(
```
chrisnojima: why sequentially vs all?

shared/actions/fs/index.js
```diff
@@ -619,6 +619,13 @@ function* openPathItem(action: FsGen.OpenPathItemPayload): Saga.SagaGenerator<an
...
+    action.payload.localPaths.map(localPath =>
```
chrisnojima: why do this, why not just handle it natively instead of converting it ?

songgao: You mean the existing `upload` handler should just take a `localPaths` array instead of just a single `localPath`?

songgao: Each individual upload would be a separate RPC call though, so I figured doing a fan out here is cleaner.

shared/actions/fs/index.js
songgao: I didn't pick sequentially over all consciously. I guess I just didn't use `all` before and assumed `sequentially` was preferred. Do we prefer `all` when order doesn't matter?

chrisnojima: yes, do things in parallel

shared/actions/fs/index.js
chrisnojima: then why bother with this action at all, just emit the individual ones upstream

songgao: I was trying to keep actionToPromise there (see the changes in `shared/actions/fs/platform-specific.desktop.js` below). If we need all these multiple actions, it seems like we'd need to use something like `takeEvery` there?

songgao:
bumpy ping!

PR #13501, opened 2018-08-29
Fix icon positioning round 2

thebearjew:
- Resolves icons clipping on mobile
- Uses the exact font metrics and css from the original icon font
- Is exactly 1px below the old icons.

- Some CSS changes need to be made in the following places (old CSS)
    - [x] Files `Breadcrumb` component
    - [x] Profile ellipses `...`
    - [x] Reactjis icon


thebearjew:
@keybase/react-hackers

shared/common-adapters/icon.desktop.js
```diff
@@ -92,8 +92,7 @@ class Icon extends Component<Props, void> {
           fontWeight: 'normal',
           fontVariant: 'normal',
           textTransform: 'none',
-          lineHeight: 'initial',
-          verticalAlign: 'bottom',
+          lineHeight: 1,
```
thebearjew: This was always the correct CSS to apply to the icon-font elements. `line-height: number` is a scalar of the element's `font-size`. So for a small icon we render with `font-size: 8px` and `line-height: 1` will make that icon's line height `8px` as well.

With the descent calculations in `font.js` this should always position the icon-font within 1px of the baseline of the text it is rendered next to.

shared/desktop/yarn-helper/font.js
```diff
@@ -236,6 +236,14 @@ const setFontMetrics = () => {
...
    */
   const kbTtf = path.resolve(paths.fonts, 'kb.ttf')
+  // Nav icons need to be shifted more because the grid size is 24.
+  // Without shifting to -(64 + 21) the nav icons will be aligned on
+  // a half pixel which will cause blurriness.
+  const icon24 = getSvgNames(true)
+    .filter(({size}) => size === '24')
+    .map(({filePath}) => `'${filePath.replace('.svg', '')}'`)
+  const icon24First = icon24[0]
+  const icon24Last = icon24[icon24.length - 1]
```
thebearjew: The discovery here was that the larger icons `kb-iconfont-nav-*-.svg` needed to move moved down an additional amount relative to the other icons.

Comparing the original icon-font to this one, the alignment of all glyphs match exactly now.

shared/common-adapters/icon.desktop.js
thebearjew: https://icomoon.io/#docs/font-metrics

thebearjew:
@keybase/react-hackers This needs another look so I can get them icons looking 💯

PR #13509, opened 2018-08-29
Retry on errors but without blackbar

joshblum:
changes the return signature for  `boxer.UnboxMessage` so we don't return errors but instead put the message in the thread with a `TRANSIENT` error type. `GetThreadNonblockLocal` and `ConversationRetry` then scan the thread for any transient errors so we know if we need to fetch/refetch the thread. this change gives us the benefit of no blackbars with retry for the broken messages.

Are there any other places we should attempt a retry? should we do the same for the `FullInboxRetry`? In particular `UnboxMessages` *wont* return an error for failed unboxing where it would before, not sure if there are any other relevant callers to attempt a retry.

cc @keybase/react-hackers i could use an assist here with formatting the error message. currently these failures look like: 
screen shot 2018-08-29 at 5 55 48 pm
I'd like it just to say the error message without the `Failure to send:` or `Cancel`/`Retry` buttons but was having some trouble tracing how to thread this through correctly

### DIFF
--- a/go/chat/boxer.go
+++ b/go/chat/boxer.go
@@ -322,6 +322,9 @@ func (b *Boxer) UnboxMessage(ctx context.Context, boxed chat1.MessageBoxed, conv
 	}
 
 	unboxed, ierr := b.unbox(ctx, boxed, conv.GetMembersType(), encryptionKey, ephemeralSeed)
+	if ierr == nil {
+		ierr = b.checkInvariants(ctx, conv.GetConvID(), boxed, unboxed)
+	}
 	if ierr != nil {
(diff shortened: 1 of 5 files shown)


mmaxim:
How are you getting the send error to show up?

joshblum:
always threw a random error why trying to get the ephemeral keys so it would show while testing

mmaxim:
I don't think that frontend display is an actual problem, you are just getting an unboxing error during send, which should basically never happen.

joshblum:
these messages were previously sent and read, just did a db nuke so we'd refetch and attempt unbox. I think `failed to send` always shows if you're the author

joshblum:
filed DESKTOP-7692 re error message formatting

PR #13519, opened 2018-08-30
Reselect conversation when leaving a team

thebearjew:
This prevents the currently selected team conversation from being stuck in the thread view, much in the same way that #13458 handled reselecting after a blocked conversation.

The fix is to always reselect after leaving a team, which we did not do before because the fall-through logic was not true. Instead we bailed early because the currently selected conversation (now old as the user has left the team), was valid and we skipped re-selection.

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -1469,12 +1469,13 @@ const _maybeAutoselectNewestConversation = (
   state: TypedState
 ) => {
   const selected = Constants.getSelectedConversation(state)
+  const selectedExists = Constants.isExistingSelectedConversation(state)
   if (action.type === Chat2Gen.metaDelete) {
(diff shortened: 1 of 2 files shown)


shared/actions/chat2/index.js
```diff
@@ -1469,12 +1469,13 @@ const _maybeAutoselectNewestConversation = (
   state: TypedState
 ) => {
   const selected = Constants.getSelectedConversation(state)
+  const selectedExists = Constants.isExistingSelectedConversation(state)
```
chrisnojima: could be
```
let selected =  Constants.getSelectedConversation(state)
if (!state.chat2.metaMap.get(selected)) {
  selected= Constants.noConversationIDKey
}
```

thebearjew:
@keybase/react-hackers

PR #13523, opened 2018-08-30
show an error message if new conversation fails CORE-8697

mmaxim:
@keybase/react-hackers 

If `NewConversationLocal` returns an error, put up an error message on the screen with instructions instead of putting up the global error. This restores the behavior of the old new conversation flow.


mmaxim:
bump

PR #13529, opened 2018-08-30
WIP: type incoming call map type rpcs

chrisnojima:
- [x] pull types into map and reference it with $PropertyType
- [x] remove exported rpcparam types
- [ ] fixup tracker using dispatch / getstate in incoming call map
- [ ] test changes


protocol/bin/flow.js
```diff
@@ -428,24 +390,41 @@ export type Long = number
...
+    .join('\n')
+
+  const messageTypes = `\nexport type MessageTypes = {|
+${messageTypesData}
+|}`
+
   const data = [
-    ...[...consts, ...types, ...messageResponse].sort(),
-    ...messageParams.sort(),
+    messageTypes,
```
chrisnojima: best to look at the diff of the generated to get the gist of the changes here. We basically make a map of method => metaData, then the helpers now just pull that metadata using $PropertyType<CallMap, 'inParams'> to extract the type info. This makes the helpers not actually reference any of the types directly at all and its very clear what all the related data is in the map on top

shared/actions/chat2/index.js
```diff
@@ -38,7 +38,7 @@ const inboxRefresh = (
 ) => {
   const username = state.config.username || ''
 
-  const onUnverified = function({inbox}: RPCChatTypes.ChatUiChatInboxUnverifiedRpcParam) {
+  const onUnverified = function({inbox}) {
```
chrisnojima: All these types went away as it actually is typed now. Before it was `any` and you'd have to say what the type was yourself

shared/actions/chat2/index.js
```diff
@@ -565,169 +564,153 @@ const reactionUpdateToActions = (info: RPCChatTypes.ReactionUpdateNotif) => {
   return [Chat2Gen.createUpdateReactions({conversationIDKey, updates})]
 }
 
+// This is to simplify the changes that setIncomingCallMap created. Could clean this up and remove this
+const arrayOfActionsToSequentially = actions =>
+  Saga.call(Saga.sequentially, (actions || []).map(a => Saga.put(a)))
+
 // Handle calls that come from the service
 const setupEngineListeners = () => {
-  engine().setIncomingActionCreators(
```
chrisnojima: this is the big change. instead of 
`setIncomingActionCreators('method', () => action) // n times`
its now
`setIncomingCallMap({method: () => effect})`
which matches exactly what you do with the engine saga helpers

chrisnojima:
old coverage: 
screen shot 2018-08-31 at 8 50 48 pm

new coverage:
screen shot 2018-08-31 at 8 51 06 pm

chrisnojima:

chrisnojima:
@keybase/react-hackers bump, we should get this in asap after the release

shared/actions/notifications.js
```diff
@@ -43,15 +43,15 @@ const setupEngineListeners = () => {
...
-    NotificationsGen.createReceivedBadgeState({badgeState})
-  )
-
-  getEngine().setIncomingActionCreators('stellar.1.notify.paymentNotification', refreshPayments)
-  getEngine().setIncomingActionCreators('stellar.1.notify.paymentStatusNotification', refreshPayments)
+  getEngine().setIncomingCallMap({
+    'keybase.1.NotifyBadges.badgeState': ({badgeState}) =>
+      Saga.put(NotificationsGen.createReceivedBadgeState({badgeState})),
+    'stellar.1.notify.paymentNotification': refreshPayments,
+    'stellar.1.notify.paymentStatusNotification': refreshPayments,
```
chrisnojima: @cjb note: this correctly fails flow as there is no accountID param in the paymentStatusNotification call

shared/actions/notifications.js
```diff
@@ -46,13 +45,9 @@ const setupEngineListeners = () => {
   getEngine().setIncomingCallMap({
     'keybase.1.NotifyBadges.badgeState': ({badgeState}) =>
       Saga.put(NotificationsGen.createReceivedBadgeState({badgeState})),
-    'stellar.1.notify.paymentNotification': refreshPayments,
```
chrisnojima: @cjb i'm going to just mark this as a flowissue so it can be fixed later. also we shouldn't be putting any random stuff into notifications, should go into the correct actions folder (wallets)

PR #13545, opened 2018-08-31
add out-of-date banner in widget

songgao:
This PR adds two RPCs in service and use it in frontend to show the out-of-date banner in the widget:

* add `getUpdateInfo` for figuring out if we are up to date, out of date, or critically out of date.
* add `startUpdateIfNeeded` to start an update

"out of date" is determined when the updater concludes there's an update available for this installer ID. This means that admins will be "out of date" everyday before updating. "critically out of date" is determined when API server sends back a `X-Keybase-Client-Upgrade-To` [header](https://github.com/keybase/client/blob/c78256d40909dae08741e48599046ee95394ec28/go/libkb/api.go#L535). Originally I had a version where I used the `X-Keybase-Client-Unsupported` header for "critical", but realized later that the "Upgrade-To" is probably more appropriate. @maxtaco Please let me know if "Unsupported" header should be used here instead.

So both added RPC will need to call the updater binary. `getUpdateInfo` is easy as it just needs to start the process running `updater need-upgrade` (added [here](https://github.com/keybase/go-updater/pull/171)) and wait for it to finish. `startUpdateIfNeeded` on the other hand, needs to spawn the updater process in background like what we do when we fork out the `service` process. So in order to reuse existing code, I moved the platform-dependent spawning code from `client` to `libcmdline` and exported the function.

I tested with frontend code on the `getUpdateInfo`, and it seems to have worked well. I couldn't figure out a way to test `startUpdateIfNeeded` though, as the RPC isn't in there with existing clients, and replacing the `keybase` binary seems to affect what updater perceives about the running version somehow, possibly through some config file. So I was hoping to test this again when we have some admin releases out, and by installing an out-of-date admin build that already has the RPCs in.


songgao:
This is ready!

songgao:
bump

shared/actions/platform-specific/index.desktop.js
```diff
@@ -242,14 +245,31 @@ const sendKBServiceCheck = (state: TypedState, action: ConfigGen.DaemonHandshake
   }
 }
 
+const checkOutOfDate = () =>
```
chrisnojima: Let's name this something like startOutOfDateCheckLoop or something to make it clear it runs forever

shared/menubar/out-of-date.js
```diff
@@ -0,0 +1,54 @@
...
+const OutOfDate = ({outOfDate, critical, updateNow}: Props) =>
+  outOfDate && (
+    <Kb.Box2
+      style={critical ? styles.boxCritical : styles.boxNonCritical}
+      fullWidth={true}
+      centerChildren={true}
+      direction="vertical"
+    >
+      <Kb.Text type="BodySemibold" style={critical ? styles.textCritical : styles.textNonCritical}>
+        Your Keybase app is {critical && 'critically'} out of date.
```
chrisnojima: You'll end up with two spaces this way if not critical

songgao:
Sounds good; will do.

shared/actions/platform-specific/index.desktop.js
```diff
@@ -242,14 +245,34 @@ const sendKBServiceCheck = (state: TypedState, action: ConfigGen.DaemonHandshake
   }
 }
 
+const startOutOfDateCheckLoop = () =>
+  Saga.call(function*() {
+    while (1) {
+      const {status, message} = yield Saga.call(RPCTypes.configGetUpdateInfoRpcPromise)
```
chrisnojima: this needs to be wrapped i a try / catch or on error the loop will die

PR #13554, opened 2018-09-02
use iOS app group shared directory for config dir, and migrate old devices CORE-8738

mmaxim:
On iOS, app extensions run in their own processes, and cannot access the files from the container app.  Apple provides a "shared container" filesystem that is linked on the organization "app group". In order for the share extension to be logged in when used, our config files need to be in this shared location. Patch does the following:

1.) Add a new notion of "mobile shared home" to the whole env system and `HomeFinder`. If the `Config` specifies a mobile shared home, the `Darwin` `HomeFinder` will use it for `ConfigDir`. 
2.) Thread the mobile shared home directory down from native into Go during `Init`. If the device had previously been provisioned, we will copy the files from app storage to shared storage. Note: we also copy all the stored EK info as well. cc @joshblum

### DIFF
--- a/go/erasablekv/erasable_kv_store.go
+++ b/go/erasablekv/erasable_kv_store.go
@@ -44,7 +44,7 @@ const noiseSuffix = ".ns"
 const storageSubDir = "eraseablekvstore"
 
 func getStorageDir(g *libkb.GlobalContext, subDir string) string {
-	return filepath.Join(g.Env.GetDataDir(), storageSubDir, subDir)
+	return filepath.Join(g.Env.GetConfigDir(), storageSubDir, subDir)
 }
 
 type ErasableKVStore interface {
--- a/go/erasablekv/erasable_kv_store.go
+++ b/go/erasablekv/erasable_kv_store.go
@@ -44,7 +44,12 @@ const noiseSuffix = ".ns"
 const storageSubDir = "eraseablekvstore"
 
 func getStorageDir(g *libkb.GlobalContext, subDir string) string {
-	return filepath.Join(g.Env.GetConfigDir(), storageSubDir, subDir)
+	base := g.Env.GetDataDir()
+	// check for iOS
+	if runtime.GOOS == "darwin" && g.GetAppType() == libkb.MobileAppType {
+		base = g.Env.GetConfigDir()
+	}
+	return filepath.Join(base, storageSubDir, subDir)
 }
 
 type ErasableKVStore interface {
--- a/shared/react-native/ios/Keybase/AppDelegate.m
+++ b/shared/react-native/ios/Keybase/AppDelegate.m
@@ -90,18 +90,8 @@ const BOOL isDebug = NO;
 - (BOOL) maybeMigrateDirectory:(NSString*)source dest:(NSString*)dest {
   NSError* error = nil;
   NSFileManager* fm = [NSFileManager defaultManager];
-  
-  // Check to see if dest is empty, and if so, let's migrate from source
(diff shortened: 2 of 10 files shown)


go/erasablekv/erasable_kv_store.go
```diff
@@ -44,7 +44,12 @@ const noiseSuffix = ".ns"
 const storageSubDir = "eraseablekvstore"
 
 func getStorageDir(g *libkb.GlobalContext, subDir string) string {
-	return filepath.Join(g.Env.GetDataDir(), storageSubDir, subDir)
+	base := g.Env.GetDataDir()
+	// check for iOS
+	if runtime.GOOS == "darwin" && g.GetAppType() == libkb.MobileAppType {
```
mmaxim: @joshblum this is a little janky, but this is what I came up with to make it so the EKs come along for the migration ride on iOS. `DataDir` == `ConfigDir` except on XDG platforms.

shared/react-native/ios/Keybase/AppDelegate.m
```diff
@@ -87,6 +87,60 @@ - (void) createBackgroundReadableDirectory:(NSString*) path setAllFiles:(BOOL)se
...
+  [self createBackgroundReadableDirectory:appKeybasePath setAllFiles:YES];
+  [self createBackgroundReadableDirectory:sharedKeybasePath setAllFiles:YES];
+  [self createBackgroundReadableDirectory:appEraseableKVPath setAllFiles:YES];
+  [self addSkipBackupAttributeToItemAtPath:appKeybasePath];
+  [self addSkipBackupAttributeToItemAtPath:sharedKeybasePath];
+  
+  if (![self maybeMigrateDirectory:appKeybasePath dest:sharedKeybasePath]) {
+    return home;
+  }
+  NSString* sharedEraseableKVPath =  [NSString stringWithFormat:@"%@/Library/Application Support/Keybase/eraseablekvstore/device-eks", sharedHome];
```
joshblum: why not move this up to the other path definitions?

shared/react-native/ios/Keybase/AppDelegate.m
```diff
@@ -96,29 +150,25 @@ - (void) setupGo
...
-  NSString * logPath = [@"~/Library/Caches/Keybase" stringByExpandingTildeInPath];
-  NSString * serviceLogFile = skipLogFile ? @"" : [logPath stringByAppendingString:@"/ios.log"];
-
-  // Make keybasePath if it doesn't exist
-  [self createBackgroundReadableDirectory:keybasePath setAllFiles:YES];
-  [self addSkipBackupAttributeToItemAtPath:keybasePath];
-
+  NSString* home = NSHomeDirectory();
+  NSString* sharedHome = [[[NSFileManager defaultManager] containerURLForSecurityApplicationGroupIdentifier:@"group.keybase"] relativePath];
+  sharedHome = [self setupHome:home sharedHome:sharedHome];
```
joshblum: should we log some error or warning if `home == sharedHome`? we don't really expect it to happen right?

mmaxim: Shouldn't happen.

shared/react-native/ios/Keybase/AppDelegate.m
mmaxim: I originally had a check to see if the shared dir was empty before running this copy, and so this directory was breaking that. I removed it since I realized that if the app is killed during this migration it would hose your install.

PR #13564, opened 2018-09-04
Less right-margin on Name w/ icon

cecileboucheron:
@keybase/react-hackers 
cc @adamjspooner


shared/common-adapters/name-with-icon.js
```diff
@@ -180,20 +181,20 @@ const TextOrComponent = ({
 const styles = styleSheetCreate({
   fullWidthText: platformStyles({isElectron: {width: '100%', whiteSpace: 'nowrap', display: 'unset'}}),
   fullWidthTextContainer: platformStyles({isElectron: {width: '100%', textAlign: 'center'}}),
-  hAvatarStyle: {marginRight: 16},
+  hAvatarStyle: {marginRight: isMobile ? globalMargins.small : globalMargins.tiny},
```
nathunsmitty: May be better to use `platformStyles` here

shared/.gitignore
```diff
@@ -0,0 +1,2 @@
+../browser/keybase-chrome-1.10.14.zip
+../browser/keybase-firefox-1.10.14.zip
```
nathunsmitty: I think these lines should be in our root .gitignore file instead of a separate one in shared/. Someone else could probably say better than I could, though.

cecileboucheron: oops.

cecileboucheron:
@keybase/react-hackers @adamjspooner bump ^

adamjspooner:
Taking a look now.

PR #13568, opened 2018-09-04
Handle "Build a Team" showing up after RPC timeout

thebearjew:
@keybase/react-hackers

### DIFF
--- a/shared/chat/inbox/row/build-team/container.js
+++ b/shared/chat/inbox/row/build-team/container.js
@@ -6,7 +6,7 @@ import type {TypedState, Dispatch} from '../../../../util/container'
 import BuildTeam from '.'
 
 const mapStateToProps = (state: TypedState) => ({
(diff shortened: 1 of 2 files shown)


shared/chat/inbox/row/build-team/container.js
```diff
@@ -6,8 +6,7 @@ import type {TypedState, Dispatch} from '../../../../util/container'
 import BuildTeam from '.'
 
 const mapStateToProps = (state: TypedState) => ({
-  loaded: state.teams.loaded,
-  showBuildATeam: state.teams.teamnames.size === 0,
+  showBuildATeam: !state.chat2.metaMap.count(m => m.teamType !== 'adhoc'),
```
buoyad: `metaMap.some` seems more appropriate here. Also, could you store the `metaMap` in `stateProps` and move the check to `mergeProps`?

PR #13571, opened 2018-09-04
Copy Text: Select text on copy icon click

nathunsmitty:
Selects the text in the CopyText component when you hit the copy icon. Also some refactoring of CopyText.

[See it in action here](http://localhost:6006/?selectedKind=Common%2FCopy%20text&selectedStory=Basic&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel).

@keybase/react-hackers 
cc @cecileboucheron


cecileboucheron:
👍

shared/common-adapters/copy-text.js
```diff
@@ -1,27 +1,13 @@
 // @flow
 import * as React from 'react'
 import * as ConfigGen from '../actions/config-gen'
-import {Box2} from './box'
-import Icon from './icon'
-import Button from './button'
-import ButtonBar from './button-bar'
-import Text from './text'
-import Toast from './toast'
+import {Box2, Icon, Button, ButtonBar, Text, Toast} from '.'
```
MarcoPolo: I think we want to avoid this since it's a circular dependency

nathunsmitty: Good catch! I was getting an error when I tried to add the `HOCTimers` component to this line. Guess circular dependencies ruin my attempts to reduce lines 🙃

PR #13585, opened 2018-09-05
Don't cut off descenders in inbox filter + snippets

buoyad:
Just wrasslin' with styling


buoyad:

shared/chat/inbox/index.desktop.js
```diff
@@ -170,8 +171,8 @@ const _containerStyle = {
   backgroundColor: globalColors.blueGrey,
   contain: 'strict',
   height: '100%',
-  maxWidth: 260,
-  minWidth: 260,
+  maxWidth: inboxWidth,
```
nathunsmitty: is there a reason why we're setting min and max width as opposed to just width?

shared/chat/inbox/row/small-team/index.js
```diff
@@ -134,30 +149,28 @@ class SmallTeam extends React.PureComponent<Props, State> {
...
-  container: {flexShrink: 0, height: RowSizes.smallRowHeight},
+  container: {flexShrink: 0, height: InboxSizes.smallRowHeight},
   conversationRow: {
     ...globalStyles.flexBoxColumn,
     flexGrow: 1,
     height: '100%',
     justifyContent: 'center',
     paddingLeft: 8,
     paddingRight: 8,
+    width: InboxSizes.inboxWidth - 48 - 16,
```
nathunsmitty: Where's the - 48 - 16 coming from? A comment would be helpful here

shared/chat/inbox/index.desktop.js
buoyad: I'm not sure, I didn't want to modify these styles, just to factor `260` out into a constant

shared/chat/inbox/row/small-team/index.js
buoyad: Hmm, this doesn't seem to be doing anything actually.. I'll get rid of it

buoyad:
ok @nathunsmitty, updated with your comments. PTAL

PR #13589, opened 2018-09-05
Show error for reusing device names and add logging to provisioning

buoyad:
We had split the error handling for reusing device names, some in the reducer and some in the view container. The reducer set the error, but since the container disallowed submitting the error was never shown. This consolidates the handling in the reducer so the user can hit 'Submit' and will see an error if the device name has been used. r? @keybase/react-hackers

### DIFF
--- a/shared/reducers/provision.js
+++ b/shared/reducers/provision.js
@@ -58,11 +58,11 @@ export default function(state: Types.State = initialState, action: ProvisionGen.
         error: initialState.error,
       })
     case ProvisionGen.submitDeviceName:
-      if (state.existingDevices.indexOf(action.payload.name) !== -1) {
+      if (state.existingDevices.find(ed => ed.toLowerCase() === action.payload.name.toLowerCase())) {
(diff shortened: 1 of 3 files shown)


buoyad:
Two other things that made this nastier:

- The code that checked for a reused device name didn't account for case, so you could get through to the code page with an invalid name
- Tests covered the reducer / saga for this case, but since the handling was in the view the tests could pass

shared/reducers/provision.js
```diff
@@ -58,11 +58,11 @@ export default function(state: Types.State = initialState, action: ProvisionGen.
         error: initialState.error,
       })
     case ProvisionGen.submitDeviceName:
-      if (state.existingDevices.indexOf(action.payload.name) !== -1) {
+      if (state.existingDevices.find(ed => ed.toLowerCase() === action.payload.name.toLowerCase())) {
```
chrisnojima: should just to lower case this once and not over and over in the loop

PR #13590, opened 2018-09-05
Actually copy secret key on account removal

nathunsmitty:
Turns out there was a bug where your account ID was being copied instead. Most of this code is copied from the export secret key screen; it loads the secret key and ensures that's copied instead.


nathunsmitty:
@keybase/react-hackers

PR #13591, opened 2018-09-05
Removed inbox dividers

cecileboucheron:


cecileboucheron:
@adamjspooner mind taking a look?

adamjspooner:
Taking a look now.

shared/chat/inbox/row/big-team-header/index.js
```diff
@@ -80,6 +80,7 @@ const styles = styleSheetCreate({
   },
   team: platformStyles({
     common: {
+      color: globalColors.black_60,
```
adamjspooner: I prefer `black_75`, but it's not a deal breaker.

cecileboucheron: I like it. `black_60` is BodySmallSemiboldSecondaryLink and consistent with our clickable section labels.

PR #13592, opened 2018-09-05
Make 'Please tell us' on black bar work when logged out

buoyad:
Adds a check on `state.config.loggedIn` and navigates with login routes rather than app routes. r? @keybase/react-hackers

### DIFF
--- a/shared/app/global-errors/container.js
+++ b/shared/app/global-errors/container.js
@@ -7,6 +7,7 @@ import {feedbackTab} from '../../constants/settings'
 import * as RouteTreeGen from '../../actions/route-tree-gen'
 
 const mapStateToProps = state => ({
+  _loggedIn: state.config.loggedIn,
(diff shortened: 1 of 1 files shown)


shared/app/global-errors/container.js
```diff
@@ -18,22 +19,38 @@ const mapDispatchToProps = dispatch => ({
...
+      )
+    } else {
+      dispatch(RouteTreeGen.createNavigateAppend({path: ['feedback']}))
+    }
   },
 })
 
-const Connected = connect(mapStateToProps, mapDispatchToProps, (s, d, o) => ({...o, ...s, ...d}))(GlobalError)
+const mergeProps = (stateProps, dispatchProps, ownProps) => {
+  return {
```
chrisnojima: should just use the shorthand and remove the 'return'

PR #13595, opened 2018-09-05
Add Wallets/Create Account story

nathunsmitty:
I'm working on another PR that touches the account name, so I figured it'd be nice to have a story for this screen, even if it looks pretty similar to other screens.

Story is [here](http://localhost:6006/?selectedKind=Wallets&selectedStory=Create%20account&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel).

r? @keybase/react-hackers


shared/wallets/create-account/index.js
```diff
@@ -4,20 +4,20 @@ import * as Kb from '../../common-adapters'
...
 import {EnterName} from '../link-existing/index'
 
-type Props = {
+type Props = {|
   createNewAccountError: string,
   error: string,
   name: string,
   nameValidationState: ValidationState,
-  onBack: () => void,
+  onBack?: () => void,
```
nathunsmitty: `onBack` isn't passed in the container at all currently, so I don't think it really makes sense to have it be a required prop. From the mockups, sometimes the back button is there and sometimes it isn't, depending on how it's opened.

shared/wallets/link-existing/index.js
```diff
@@ -171,7 +171,7 @@ const EnterKey = (props: EnterKeyProps) => (
 type EnterNameProps = {
   error: string,
   name: string,
-  onBack: ?() => void,
+  onBack?: () => void,
```
nathunsmitty: The same thing goes for entering a name in the link existing popup – sometimes there is a back button and sometimes there isn't. Seems like it'd be better to make it optional instead of maybe to be consistent.

PR #13601, opened 2018-09-06
more traditional feature flag system

maxtaco:
- ask the server if it's live or not, and cache it for as long as the server says

### DIFF
--- a/go/libkb/features.go
+++ b/go/libkb/features.go
@@ -2,6 +2,8 @@ package libkb
 
 import (
 	"strings"
+	"sync"
+	"time"
 )
 
 type Feature string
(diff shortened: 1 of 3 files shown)


maxtaco:
quick review on this, it's pretty simple

maxtaco:
cc @heronhaye, this system I prefer....

heronhaye:
lgtm, is there server support as well?

maxtaco:
No, just a flag to not send down signatures, which was in there for a while.

go/libkb/features.go
```diff
@@ -33,3 +35,109 @@ func (set FeatureFlags) Admin() bool {
 func (set FeatureFlags) Empty() bool {
 	return len(set) == 0
 }
+
+type featureSlot struct {
+	sync.Mutex
```
mlsteele: Wouldn't locktab be better?

maxtaco: this seemed simpler. i can change it

patrickxb:
approved

PR #13606, opened 2018-09-06
Show spinner in pending chat if no users have been selected

buoyad:
r? @keybase/react-hackers


buoyad:
Hmm, seems like that spinner issue has already been fixed

PR #13608, opened 2018-09-06
detect transient error on validating pairwise macs

joshblum:


joshblum:
looked through other times we return permanent errors, they all seem legit since it's for failed validation checks on static data (so we won't get transient api errors)

PR #13610, opened 2018-09-06
Enforce log pattern merkle skips

heronhaye:


go/libkb/merkle_client.go
```diff
@@ -394,6 +395,34 @@ func GetNodeHashVoid(w *jsonw.Wrapper, nhp *NodeHash, errp *error) {
 	}
 }
 
+func computeSetBitsBigEndian(x int) []int {
+	var ret []int
+	bit := 1
+	for i := 0; i <= int(math.Ceil(math.Log2(float64(x)))); i++ {
```
maxtaco: pull this (floating point) out of the loop so it's only performed once, not every time through the loop.  And plus, would we prefer a discrete log2 that just uses bitshifting? (Or maybe something crazy like this: https://stackoverflow.com/questions/11376288/fast-computing-of-log2-for-64-bit-integers)

maxtaco: Use `uints` or `uint64s`?

go/libkb/merkle_client.go
```diff
@@ -394,19 +395,36 @@ func GetNodeHashVoid(w *jsonw.Wrapper, nhp *NodeHash, errp *error) {
 	}
 }
 
-func computeSetBitsBigEndian(x int) []int {
-	var ret []int
-	for bit := 1; bit <= x; bit *= 2 {
+func computeSetBitsBigEndian(x uint) []uint {
+	if x == 0 {
+		return nil
+	} else if x == 1 {
```
maxtaco: no `else`

go/libkb/merkle_client.go
```diff
@@ -394,19 +395,36 @@ func GetNodeHashVoid(w *jsonw.Wrapper, nhp *NodeHash, errp *error) {
...
+		return []uint{1}
+	}
+	// Allocate maximum array size necessary
+	high := int(math.Ceil(math.Log2(float64(x))))
+	ret := make([]uint, high)
```
maxtaco: it might be easier to `ret = make([]uint, 0, high)`, and then just append?

heronhaye: ah, I didn't know about this. Thanks!

PR #13613, opened 2018-09-06
Enforce https if requested with protocol or in cmdline option

heronhaye:


go/externals/services.go
```diff
@@ -19,6 +19,10 @@ func (e externalServicesCollection) Register(st libkb.ServiceType) {
 	}
 }
 
+func (e externalServicesCollection) ManuallyRegister(key string, st libkb.ServiceType) {
```
heronhaye: possibly should be better named, couldn't think of one.

mlsteele:
image

heronhaye:
Isn't that intended?

mlsteele:
I doubt it. But rather than figure that out how about this one?


This is after clicking these buttons.
1
image

2
image

3
image

heronhaye:
@mlsteele this should fix the gui. waiting for nojima to get back to make sure I didn't break everything

mlsteele:
Cool! Consider the service half approved. I think anyone on `@keybase/react-hackers` should be able to review this. However, CI says you have a flow error first.

go/externals/proof_support_web.go
```diff
@@ -103,7 +106,14 @@ func (t WebServiceType) NormalizeRemoteName(m libkb.MetaContext, s string) (ret
 		err = libkb.NewWebUnreachableError(host)
 		return
 	}
-	if len(prot) > 0 && prot == "https" && found != "https:" {
+	if len(t.scheme) > 0 && len(prot) > 0 && prot != t.scheme {
```
akalin-keybase: probably don't need all the length checks here and below, right?

shared/actions/profile/proofs.js
```diff
@@ -72,6 +72,9 @@ function _addProof(action: ProfileGen.AddProofPayload) {
       actions.push(Saga.put(_registerBTC()))
       break
     // flow needs this for some reason
+    case 'web':
+      actions.push(Saga.call(_addServiceProof, 'web'))
```
akalin-keybase: don't think you need to special-case this, you could just `case 'web':` below `case 'https':`

shared/profile/confirm-or-pending/shared.js
```diff
@@ -71,6 +71,7 @@ export function propsForPlatform(props: Props): Object {
           props.username
         } doesn't load over https. If you get a real SSL certificate (not self-signed) in the future, please replace this proof with a fresh one.`,
       }
+    case 'web':
```
akalin-keybase: can you put this below `case 'https':` to be consistent with the other places

shared/profile/post-proof/shared.js
```diff
@@ -114,7 +114,8 @@ export function propsForPlatform(props: Props): MoreProps {
       }
     case 'http':
     case 'https':
-      const root = `${props.platform}://${props.platformUserName}`
+    case 'web':
+      const root = props.platformUserName
```
akalin-keybase: seems like for http and https, we want to have `${props.platform}://${props.platformUserName}`, but for web, we want to give four URLs, two for https and two for http right? (with possibly some additional copy to say that https is preferred)

shared/profile/prove-enter-username/index.stories.js
```diff
@@ -34,7 +34,7 @@ const load = () => {
     .add('Bitcoin', () => <ProveEnterUsername {...props} platform="btc" />)
     .add('Bitcoin Disabled', () => <ProveEnterUsername {...props} platform="btc" canContinue={false} />)
     .add('DNS', () => <ProveEnterUsername {...props} platform="dns" />)
-    .add('Website', () => <ProveEnterUsername {...props} platform="http" />)
+    .add('Website', () => <ProveEnterUsername {...props} platform="web" />)
```
akalin-keybase: let's have one line for http, https, and web here

shared/profile/prove-website-choice/container.js
```diff
@@ -10,7 +10,7 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
   // both with a preference for https
   onCancel: () => dispatch(ProfileGen.createCancelAddProof()),
   onOptionClick: choice =>
-    dispatch(ProfileGen.createAddProof({platform: choice === 'file' ? 'https' : 'dns'})),
+    dispatch(ProfileGen.createAddProof({platform: choice === 'file' ? 'web' : 'dns'})),
```
akalin-keybase: you can remove the comment above now i think

akalin-keybase: would be nice to have an option for the user to specify http or https, although that could be a separate ticket

akalin-keybase:
Oh, also can you merge/rebase on master and fix any flow errors?

shared/profile/post-proof/shared.js
heronhaye: In the URL for where to post the proof, I thought it'd be clearer to just give back what the user specified: `example.com` if they provided it, or `https://example.com` if they provided that. The service RPC doesn't really give back the normalized protocol or domain name, so we'd need to do some regex over the proof instructions (or better, add to the service function result). I can do the service change if you think that's better.

akalin-keybase:
So I think adding the `web` case in the last few places will fix the flow errors. btw, you can do `yarn test stories -u` to have it update the storyshots file.

go/externals/proof_service_web.go
```diff
@@ -52,9 +52,17 @@ func (rc *WebChecker) CheckStatus(m libkb.MetaContext, h libkb.SigHint, pcm libk
...
+	scheme string
+}
 
-func (t WebServiceType) AllStringKeys() []string     { return []string{"web", "http", "https"} }
+func (t WebServiceType) AllStringKeys() []string {
+	if t.scheme == "" {
+		return []string{"web"}
+	}
+	return []string{t.scheme}
+}
```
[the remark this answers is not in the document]
heronhaye: It's not being used anywhere and it's pretty old (from 2014). Should I just get rid of it?

heronhaye:
ok cool, thanks @akalin. For some reason flow was giving me no errors even though I was missing those switch cases.

akalin-keybase:
Yeap, flow doesn't mandate that all cases of an option type are present in a switch

heronhaye:
Is this ok to merge @akalin-keybase?

go/externals/proof_service_web.go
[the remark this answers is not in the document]
akalin-keybase: Ah yeah, just get rid of it then

shared/common-adapters/platform-icon.js
```diff
@@ -34,6 +34,7 @@ function _specsForMobileOrDesktop() {
     http: {icon: isMobile ? 'icon-website-64' : 'icon-website-48'},
     https: {icon: isMobile ? 'icon-website-64' : 'icon-website-48'},
     dnsOrGenericWebSite: {icon: isMobile ? 'icon-website-64' : 'icon-website-48'},
+    web: {icon: isMobile ? 'icon-website-64' : 'icon-website-48'},
```
akalin-keybase: please move this right below https

PR #13618, opened 2018-09-06
fallback to in-memory db if leveldb is not configured (with an option) CORE-8742

mmaxim:
On iOS, multiple copies of the share extension can be running at once against the same LevelDB. Once one has it locked, then others won't be able to use it and the app won't start. Provide an option for falling back to an in-memory database if we can't use LevelDB for some reason. This option is only set in the iOS share extension.

### DIFF
--- a/go/libkb/globals.go
+++ b/go/libkb/globals.go
@@ -516,19 +516,9 @@ func (g *GlobalContext) configureDiskCachesLocked() (err error) {
 	g.LocalChatDb = NewJSONLocalDb(NewLevelDb(g, g.Env.GetChatDbFilename))
 
 	epick := FirstErrorPicker{}
-	epick.Push(g.LocalDb.ForceOpen())
-	epick.Push(g.LocalChatDb.ForceOpen())
-	if err = epick.Error(); err != nil {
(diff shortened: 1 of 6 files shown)


go/libkb/globals.go
```diff
@@ -508,17 +508,27 @@ func (g *GlobalContext) FlushCaches() {
...
 
 	epick := FirstErrorPicker{}
-	epick.Push(g.LocalDb.Open())
-	epick.Push(g.LocalChatDb.Open())
-	return epick.Error()
+	epick.Push(g.LocalDb.ForceOpen())
+	epick.Push(g.LocalChatDb.ForceOpen())
+	if err = epick.Error(); err != nil {
+		// Failed to open leveldb. If we are configured such that using the in-memory cache is an
+		// option, then let's sub that in now.
```
patrickxb: is it possible one worked and the other didn't?  i.e. no error with LocalDb but and error with LocalChatDb?

mmaxim: I'm not sure how, I thought it would be simpler though if they were all on the same backend.

PR #13619, opened 2018-09-06
[KBFS-3123]: 1 file uploading: show the file name

brayoh:
#### Description of task:
We currently say "Encrypting and uploading 1 files" but let's show the file name when there's only 1 file being uploaded: "Encrypting and uploading Screenshot_2018.png"

#### Relevant Screenshots:
screen shot 2018-09-06 at 22 50 00

### DIFF
--- a/shared/fs/footer/upload-container.js
+++ b/shared/fs/footer/upload-container.js
@@ -45,7 +45,7 @@ const mergeProps = ({_uploads, _pathItems}, {debugToggleShow}) => {
     // since journal status comes a bit slower, and merging the two causes
     // flakes on our perception of overall upload status.
     files: filePaths.size,
-    fileName: filePaths.size === 1 ? filePaths.first() : undefined,
+    fileName: filePaths.size === 1 ? Types.getPathName(filePaths.first() || Types.stringToPath('')) : null,
     endEstimate: enableDebugUploadBanner ? _uploads.endEstimate + 32000 : _uploads.endEstimate,
     totalSyncingBytes: _uploads.totalSyncingBytes,
     debugToggleShow,
--- a/shared/fs/footer/upload-container.js
+++ b/shared/fs/footer/upload-container.js
@@ -4,6 +4,7 @@ import * as Types from '../../constants/types/fs'
 import {compose, connect, setDisplayName, type TypedState} from '../../util/container'
 import Upload from './upload'
 import UploadCountdownHOC, {type UploadCountdownHOCProps} from './upload-countdown-hoc'
+import {unknownPathItem} from '../../constants/fs'
 
 const mapStateToProps = (state: TypedState) => ({
   _uploads: state.fs.uploads,
@@ -35,16 +36,16 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
 })
 
 const mergeProps = ({_uploads, _pathItems}, {debugToggleShow}) => {
-  const folders = _pathItems.filter(path => path.type === 'folder')
-  const uploadPaths = _uploads.syncingPaths // get array of paths
-  const filePaths = uploadPaths.filter(item => !folders.get(item))
+  const filePaths = _uploads.syncingPaths.filter(
+    path => _pathItems.get(path, unknownPathItem).type !== 'folder'
+  )
 
   return ({
     // We just use syncingPaths rather than merging with writingToJournal here
     // since journal status comes a bit slower, and merging the two causes
     // flakes on our perception of overall upload status.
-    files: _uploads.syncingPaths.size,
-    filePaths,
+    files: filePaths.size,
+    fileName: filePaths.size === 1 ? filePaths.first() : undefined,
     endEstimate: enableDebugUploadBanner ? _uploads.endEstimate + 32000 : _uploads.endEstimate,
     totalSyncingBytes: _uploads.totalSyncingBytes,
     debugToggleShow,
--- a/shared/fs/footer/upload-container.js
+++ b/shared/fs/footer/upload-container.js
@@ -36,22 +36,15 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
 
 const mergeProps = ({_uploads, _pathItems}, {debugToggleShow}) => {
   const folders = _pathItems.filter(path => path.type === 'folder')
-  const uploadPaths = _uploads.syncingPaths.toJS() // get array of paths
-  const filePaths = []
-
-  uploadPaths.map(uPath => {
-    if (!folders.has(uPath)) {
-      //  path is not part of the folders
-      filePaths.push(uPath)
-    }
-  })
+  const uploadPaths = _uploads.syncingPaths // get array of paths
+  const filePaths = uploadPaths.filter(item => !folders.get(item))
 
   return ({
     // We just use syncingPaths rather than merging with writingToJournal here
     // since journal status comes a bit slower, and merging the two causes
     // flakes on our perception of overall upload status.
     files: _uploads.syncingPaths.size,
-    filePaths: filePaths,
+    filePaths,
     endEstimate: enableDebugUploadBanner ? _uploads.endEstimate + 32000 : _uploads.endEstimate,
     totalSyncingBytes: _uploads.totalSyncingBytes,
     debugToggleShow,
--- a/shared/fs/footer/upload-container.js
+++ b/shared/fs/footer/upload-container.js
@@ -39,10 +39,10 @@ const mergeProps = ({_uploads, _pathItems}, {debugToggleShow}) => {
   const uploadPaths = _uploads.syncingPaths.toJS() // get array of paths
   const filePaths = []
 
-  uploadPaths.map(path => {
-    if (!folders.has(path)) {
+  uploadPaths.map(uPath => {
+    if (!folders.has(uPath)) {
       //  path is not part of the folders
-      filePaths.push(path)
+      filePaths.push(uPath)
     }
   })
 
--- a/shared/fs/footer/upload-container.js
+++ b/shared/fs/footer/upload-container.js
@@ -7,6 +7,7 @@ import UploadCountdownHOC, {type UploadCountdownHOCProps} from './upload-countdo
 
 const mapStateToProps = (state: TypedState) => ({
   _uploads: state.fs.uploads,
+  _pathItems: state.fs.pathItems,
 })
 
 // NOTE flip this to show a button to debug the upload banner animations.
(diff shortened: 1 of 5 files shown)


jzila:
@brayoh nice! Looks like `yarn flow` didn't pass, so you have a few small fixes to do from that.

Also for anyone else taking a look here, [?w=1](https://github.com/keybase/client/pull/13619/files?w=1) is very helpful.

shared/fs/footer/upload-container.js
```diff
@@ -33,16 +34,29 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
   debugToggleShow: getDebugToggleShow(dispatch),
 })
 
-const mergeProps = ({_uploads}, {debugToggleShow}) =>
-  ({
+const mergeProps = ({_uploads, _pathItems}, {debugToggleShow}) => {
+  const folders = _pathItems.filter(path => path.type === 'folder')
```
songgao: This would end up going through everything. `_pathItems` is a map and you can call `.get` on it to retrieve any item. Why not go through everything in `_uploads.syncingPaths` and check against `_pathItems` to see if anything is a folder?

shared/fs/footer/upload-container.js
```diff
@@ -33,16 +34,29 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
...
+  const uploadPaths = _uploads.syncingPaths.toJS() // get array of paths
```
songgao: I don't think this is needed. You can call `filter` directly on `_uploads.syncingPaths` which is a Set: https://facebook.github.io/immutable-js/docs/#/Set/filter

brayoh:
@jzila Thanks, let me check it out and make the fixed

shared/fs/footer/upload-container.js
```diff
@@ -33,16 +34,29 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
...
+  const filePaths = []
+
+  uploadPaths.map(path => {
+    if (!folders.has(path)) {
+      //  path is not part of the folders
+      filePaths.push(path)
+    }
+  })
```
songgao: Several things:

1. I think we'd want the `files` prop to only include files anyway, and `syncingPaths` has both files and folders now. So for example, if you are uploading `/keybase/team/blah/a/b.txt`, `syncingPaths` would include both `/keybase/team/blah/a` and `/keybase/team/blah/a/b.txt`. In this case, ideally we'd want `files` to be 1 instead of 2.

2. I think we should show the file name instead of file path. There's `Constants.getPathName` that can be useful.

3. The component only need to know about file name if it's the only one. So `filePaths` should be a single string `fileName`.

If you take my suggestion above about using `filter` on `syningPaths`, then we'd get a `uploadPaths` that only includes files. So `files` can just be the size of it, and `fileName` would be either null or undefined, or the only thing from `uploadPaths`.

shared/fs/footer/upload.desktop.js
```diff
@@ -37,37 +37,50 @@ const realCSS = `
...
+  )
+}
 
 const stylesText = {
   color: globalColors.white,
 }
 
+const stylesTextElipsis = {
+  color: globalColors.white,
+  maxWidth: '60vw',
```
songgao: Hmm is there an alternative to using `vw` here? I haven't see it used anywhere in our code base.

brayoh: The only option is to use pixels, percentage won't work.

shared/fs/footer/upload-container.js
brayoh: I don't think using`.get` will retrieve the folders in `pathItems`

songgao: What I meant was getting rid of this `folders` entirely, but instead have something like this below:

```
const filePaths = _uploads.syncingPaths.filter(path => _pathItems.get(path, Constants.unknownPathItem).type !== 'folder')

File: shared/fs/footer/upload-container.js [hunk 5]
```diff
@@ -33,16 +34,22 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
...
+  const folders = _pathItems.filter(path => path.type === 'folder')
+  const uploadPaths = _uploads.syncingPaths // get array of paths
+  const filePaths = uploadPaths.filter(item => !folders.get(item))
+
+  return ({
     // We just use syncingPaths rather than merging with writingToJournal here
     // since journal status comes a bit slower, and merging the two causes
     // flakes on our perception of overall upload status.
     files: _uploads.syncingPaths.size,
+    filePaths,
```
songgao: As I mentioned in the previous comment, I think we'd need to turn path into a name here. Also we only need one filename here. WDYT about something like this?

```
fileName: filePaths.size() === 1 ? filePaths.first() : undefined,
```

shared/fs/footer/upload-container.js
```diff
@@ -33,16 +34,22 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
...
+const mergeProps = ({_uploads, _pathItems}, {debugToggleShow}) => {
+  const folders = _pathItems.filter(path => path.type === 'folder')
+  const uploadPaths = _uploads.syncingPaths // get array of paths
+  const filePaths = uploadPaths.filter(item => !folders.get(item))
+
+  return ({
     // We just use syncingPaths rather than merging with writingToJournal here
     // since journal status comes a bit slower, and merging the two causes
     // flakes on our perception of overall upload status.
     files: _uploads.syncingPaths.size,
```
songgao: Did you mean to leave this as size of syncing paths? I was thinking we could use `filePaths.size` here.

shared/fs/footer/upload-container.js
```diff
@@ -33,16 +34,22 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
   debugToggleShow: getDebugToggleShow(dispatch),
 })
 
-const mergeProps = ({_uploads}, {debugToggleShow}) =>
-  ({
+const mergeProps = ({_uploads, _pathItems}, {debugToggleShow}) => {
+  const folders = _pathItems.filter(path => path.type === 'folder')
+  const uploadPaths = _uploads.syncingPaths // get array of paths
```
songgao: Hmm why reassigning it to a different name? Also, you can check [this file](https://github.com/keybase/client/blob/master/shared/constants/types/fs.js) for types. `syncingPaths` is an `I.Set` here.

shared/fs/footer/upload-container.js
brayoh: Cool, let me change that

shared/fs/footer/upload-container.js
brayoh: Let me try that

shared/fs/footer/upload.desktop.js
```diff
@@ -35,28 +35,40 @@ const realCSS = `
...
   transition: all .3s ${easing};
 }
+.text-overflow {
+  color: ${globalColors.white};
+  max-width: 60vw;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  text-align: center;
+}
```
songgao: Is there a way to do this in the `styles` object below? I think we prefer to not use real CSS whenever possible.

I'm not good a CSS, but for example would `max-width: '100%'` work here?

brayoh: I'd have to change a lot for that to work, `max-width: 100%` won't work since `flexbox` styles override `position` styles

brayoh: Also, the reason I moved this from the object below is because of a flow error, turns outs the only way to make `text-overflow` work is by passing the css directly

chrisnojima: we should not move things into the 'real' css blocks because of this. The reason this doesn't pass on flow is because its not actually cross platform with RN. This is what the platformStyles helper is for

brayoh: Thanks for pointing that out, how do I address this to make it work for cross-platform?

brayoh: Thanks, let me add that.

brayoh:
@songgao I've implemented the changes

brayoh:
Sorry,  I somehow missed out on that

brayoh:
I've made the changes @jzila

shared/fs/footer/upload.desktop.js
```diff
@@ -37,26 +38,46 @@ const realCSS = `
...
+      whiteSpace: 'nowrap',
+      overflow: 'hidden',
+      textAlign: 'center',
+      textOverflow: 'ellipsis',
+    },
+    isElectron: {
+      overflowWrap: 'ellipsis',
+    },
+  }),
+})
```
jzila: Usually we put the styles on the bottom of the page, before the exports. That is primarily so that a reader of the code sees relevant code when they open a file, rather than a bunch of styles.

shared/fs/footer/upload.desktop.js
```diff
@@ -37,26 +38,46 @@ const realCSS = `
...
-)
+const styles = Styles.styleSheetCreate({
+  textOverflow: Styles.platformStyles({
+    common: {
+      color: globalColors.white,
+      maxWidth: '60%',
+      whiteSpace: 'nowrap',
+      overflow: 'hidden',
+      textAlign: 'center',
+      textOverflow: 'ellipsis',
```
jzila: This should be in `isElectron`.

shared/fs/footer/upload.desktop.js
```diff
@@ -37,26 +38,46 @@ const realCSS = `
...
+    },
+    isElectron: {
+      overflowWrap: 'ellipsis',
```
jzila: Do we need `overflowWrap`?

shared/fs/footer/upload.desktop.js
```diff
@@ -37,26 +38,46 @@ const realCSS = `
...
-      </Box>
-    </CSSTransition>
-  </React.Fragment>
-)
+const styles = Styles.styleSheetCreate({
+  textOverflow: Styles.platformStyles({
+    common: {
+      color: globalColors.white,
+      maxWidth: '60%',
+      whiteSpace: 'nowrap',
```
jzila: This should be in `isElectron`.

shared/fs/footer/upload.js.flow
```diff
@@ -4,9 +4,10 @@ import * as React from 'react'
 export type UploadProps = {
   showing: boolean,
   files: number,
+  fileName: string | null,
```
songgao: Not that you have to change, but I think this could also be `fileName: ?string`, which is equivalent to what you have. Another form you might see in the codebase is `fileName?: string`, or `fileName?: ?string`. The difference is, if the `?` is after the field name, it means the field is optional, i.e. can either be the type indicated, or `undefined`. If the `?` is right before the type, it means the field has to present, but can either be the type indicated, or `null`. Similarly, `fileName?: ?string` means `fileName` can be either a string, `null`, or `undefined`.

brayoh: Thanks, let me resolve this and the comments above

shared/fs/footer/upload.desktop.js
brayoh: Yes so that if the text is overflowing, it shows the `ellipsis dots`

shared/fs/footer/upload.desktop.js
brayoh: Let me change

brayoh:
@jzila My tests are working locally but they are failing on the build, any idea why?

shared/fs/footer/upload.desktop.js
jzila: Oops nvmind you're right. Maybe the entry should be named something other than the same thing as a css class, to reduce confusion.

shared/fs/footer/upload.js.flow
songgao: Under @taruti 's prompt I'm realizing that I was being inaccurate here when I said `fileName: ?string` was equivalent to `fileName: string | null`. `fileName: ?string` actually covers both `null` and `undefined`. `fileName?` indicates that it's optional. Please see [Maybe types](https://flow.org/en/docs/types/primitives/#toc-maybe-types) and [Optional](https://flow.org/en/docs/types/primitives/#toc-optional-object-properties) stuff in the Flow doc.

Not suggesting any change here, but since I gave mis-information here I wanted to make sure to correct it. Sorry about that!

PR #13621, opened 2018-09-06
Add z index to RelativePopup

nathunsmitty:
This fixes the `Toast` being hidden behind a `PopupDialog` issue.

Components that use RelativePopup:
- FloatingMenu
- FloatingBox
- Toast
- FloatingPicker

Let me know if my grepping missed anything. I think we want all of these components to be above PopupDialog Elements (Dropdown, MaybePopup).


nathunsmitty:
closing this out in favor of another solution

nathunsmitty:
turns out this was the right solution all along!

PR #13623, opened 2018-09-06
Use payment info bundled wth chat message or notification to render chat messages

buoyad:
A `UIPaymentInfo` object will come either as a field on the message if the service has cached it, or later on in a notification if not. We'll store this on the message object in `messageMap`. I couldn't get rid of the call to load wallets on login because we still need the default accountID to open the send form. r? @keybase/react-hackers


shared/actions/chat2/index.js
```diff
@@ -742,6 +742,27 @@ const setupEngineListeners = () => {
       `ChatHandler: got NotifyChat.ChatSetConvSettings with no valid minWriterRole for convID ${conversationIDKey}. The local version may be out of date.`
     )
   })
+  engine().setIncomingActionCreators(
+    'chat.1.NotifyChat.ChatPaymentInfo',
+    (notif: RPCChatTypes.NotifyChatChatPaymentInfoRpcParam) => {
+      const conversationIDKey = Types.conversationIDToKey(notif.convID)
+      const paymentInfo = Constants.uiPaymentInfoToChatPaymentInfo(notif.info)
+      if (!paymentInfo) {
+        // This should never happen
```
cjb: If it shouldn't ever happen, perhaps a black bar in the case that it does? Otherwise we just won't notice the logger.error()

buoyad: Sounds good

PR #13624, opened 2018-09-06
Add generic proof support to assertion language

joshblum:
First pass for adding generic proofs support for assertions. Specifically:

- `keybase id foo@mastodon.social` attempt to search for `foo` on the server and fail there (in dev), the server needs to know about generic services
- sets the stage for loading configurations from the server (coming next) by registering any generic configs we find


go/bind/keybase.go
```diff
@@ -160,7 +160,7 @@ func Init(homeDir string, mobileSharedHome string, logFile string, runModeStr st
 
 	kbCtx = libkb.NewGlobalContext()
 	kbCtx.Init()
-	kbCtx.SetServices(externals.GetServices())
+	kbCtx.SetServices(externals.NewExternalServices(kbCtx))
```
joshblum: while we're doing this refactor can we rename the module? i've always found it a bit confusing... maybe `proof_services`?

go/chat/utils/utils.go
```diff
@@ -112,17 +112,17 @@ func AggRateLimits(rlimits []chat1.RateLimit) (res []chat1.RateLimit) {
 // ReorderParticipants based on the order in activeList.
 // Only allows usernames from tlfname in the output.
 // This never fails, worse comes to worst it just returns the split of tlfname.
-func ReorderParticipants(ctx context.Context, g libkb.UIDMapperContext, umapper libkb.UIDMapper,
+func ReorderParticipants(m libkb.MetaContext, g libkb.UIDMapperContext, umapper libkb.UIDMapper,
```
mlsteele: I think `mctx` rather than `m` is preferred by the more-letters club

go/engine/identify2_test.go
```diff
@@ -1100,26 +1100,31 @@ func TestResolveAndCheck(t *testing.T) {
 	evilResolver := evilResolver{goodResolver, "t_alice", tracyUID}
 
 	var tests = []struct {
-		s string
-		e error
-		r libkb.Resolver
+		s       string
+		e       error
+		useEvil bool
```
joshblum: `go vet` complained about copying locks hence the changes here

go/engine/identify2_test.go
```diff
@@ -1100,26 +1100,31 @@ func TestResolveAndCheck(t *testing.T) {
...
+		{"t_tracy", nil, false},
+		{"t_tracy+" + string(tracyUID) + "@uid", nil, false},
+		{"tacovontaco@twitter+t_tracy@rooter+foobunny@github", libkb.UnmetAssertionError{}, false},
+		{"foobunny@github", libkb.ResolutionError{}, false},
+		{"foobunny", libkb.NotFoundError{}, false},
+		{"foobunny+foobunny@github", libkb.NotFoundError{}, false},
+		{"t_alice", libkb.UIDMismatchError{}, true},
+		{"t_alice+t_tracy@rooter", libkb.UnmetAssertionError{}, true},
+		{"t_alice+" + string(aliceUID) + "@uid", libkb.UnmetAssertionError{}, true},
+		// NOTE: Generic proofs are a WIP, this should change to a
```
joshblum: but this stuff is new

go/engine/identify2_test.go
```diff
@@ -1100,26 +1100,31 @@ func TestResolveAndCheck(t *testing.T) {
...
+		// NotFoundError and a success case as the implementation proceeds
+		{"foobunny@mastodon.social", libkb.AppStatusError{}, false},
+		// TODO set this up for success!
+		{"t_alice@mastodon.social", libkb.AppStatusError{}, false},
```
mlsteele: seems easy to forget about this, is there a ticket number that could be written next to the todo?

go/externals/proof_support_generic.go
```diff
@@ -0,0 +1,97 @@
...
+type GenericServiceType struct {
+	libkb.BaseServiceType
+	conf       keybase1.GenericServiceConfig
+	usernameRe *regexp.Regexp
+}
+
+func NewGenericServiceType(conf keybase1.GenericServiceConfig) GenericServiceType {
+	return GenericServiceType{
+		conf:       conf,
+		usernameRe: regexp.MustCompile(conf.UsernameRe),
```
joshblum: will this be a pain to share between iced/go?

go/externals/proof_support_generic.go
```diff
@@ -0,0 +1,97 @@
...
+
+func NewGenericChecker(p libkb.RemoteProofChainLink) (*GenericChecker, libkb.ProofError) {
+	return &GenericChecker{p}, nil
+}
+
+func (rc *GenericChecker) GetTorError() libkb.ProofError { return nil }
+
+func (rc *GenericChecker) CheckStatus(m libkb.MetaContext, h libkb.SigHint, _ libkb.ProofCheckerMode, pvlU libkb.PvlUnparsed) libkb.ProofError {
+	// TODO will have to use the `check_url`/`check_path` config values to verify the proof.
+	return nil
```
mlsteele: Can this also `return fmt.Errorf("Not implemented")`? If that breaks then this is fine for now.

go/externals/services.go
```diff
@@ -4,37 +4,86 @@
...
+		return
+	}
 
-func GetServices() libkb.ExternalServicesCollector {
-	return externalServices
+	loader := NullConfigLoader{}
+	configs, err := loader.LoadAll()
+	if err != nil {
+		// TODO we probably don't want it to work this way..
+		panic(fmt.Sprintf("Unable to load proofs: %v", err))
```
mlsteele: heh yeah. Panic makes me nervous. Log and proceed as if configs was empty? Add ticket number to TODO

protocol/avdl/keybase1/prove_common.avdl
```diff
@@ -86,4 +86,22 @@ protocol proveCommon {
...
+    boolean enabled;
+    boolean isDevel;
+    string displayName;
+    string domain;
+    union { null, GenericServiceGroup } group;
+    string usernameRe;
+    string prefillUrl;
+    string checkUrl;
+    array<string> checkPath;
+  }
```
mlsteele: I feel like GenericServiceConfig since it's in `keybase1.` is a pretty abstract name since generic could be anything and service also means service. Maybe something like ProofServiceConfig, ProofServiceParams?

go/chat/utils/utils.go
joshblum: hehe sounds good

go/engine/identify2_test.go
joshblum: nice thing is the test will fail since the expected status will change once it's fix but down for adding it

go/bind/keybase.go
mlsteele: For the sake of diff readability that should be its own PR. If packages can handle underscores that seems nice

go/externals/proof_support_generic.go
mlsteele: Oh I was looking for but missed this.. I don't think this should use `MustCompile` since it makes it too easy for the server to crash all the clients. The client may want to do other validation against the server's response. Perhaps a validation function that, if it fails, just removes the service from the list would be good.

protocol/avdl/keybase1/prove_common.avdl
joshblum: yes better

go/externals/proof_support_generic.go
mlsteele: I think that versioning and compatibility will need to be addressed so in that way it will be an unavoidable pain for both sides to have a representation.

go/engine/identify2_test.go
mlsteele: ok if you're sure it won't get left sounds good

go/externals/proof_support_generic.go
joshblum: makes sense, i'll leave as is and drop a todo when we implement the loading

joshblum:
@mlsteele ready for another look, will open a PR in kbfs to address the signature change to pass `libkb.GlobalContext` through

mlsteele:
Oh that may be the reason this didn't have G in the first place. They probably don't have a G.

joshblum:
opened https://github.com/keybase/kbfs/pull/1789/ to handle the global context issues.

go/externals/proof_service_parameterized.go
```diff
@@ -0,0 +1,105 @@
...
+func (t ParamProofServiceType) GetTypeName() string                { return t.conf.Domain }
+
+func (t ParamProofServiceType) RecheckProofPosting(tryNumber int, status keybase1.ProofStatus, _ string) (warning *libkb.Markup, err error) {
+	return t.BaseRecheckProofPosting(tryNumber, status)
+}
+func (t ParamProofServiceType) GetProofType() string { return t.BaseGetProofType(t) }
+
+func (t ParamProofServiceType) CheckProofText(text string, id keybase1.SigID, sig string) (err error) {
+	return t.BaseCheckProofTextFull(text, id, sig)
+}
```
mlsteele: I don't really get what this is for. Do you think it would be better to leave it erroring out for now?

joshblum: sgtm

go/externals/proof_service_parameterized.go
```diff
@@ -0,0 +1,105 @@
...
+	return &ParamProofChecker{l}
+}
+
+func (t ParamProofServiceType) IsDevelOnly() bool { return t.conf.IsDevel }
+
+func (t ParamProofServiceType) FormatProofText(m libkb.MetaContext, ppr *libkb.PostProofRes) (res string, err error) {
+	//url := strings.replace(t.conf.PrefillUrl, usernameKey, username, 1)
+	//url = strings.replace(url, sigHashKey, ppr.ID, 1)
+	//return libkb.FmtMarkup(fmt.Sprintf("Please visit <strong>%s</strong>, to complete the proof", url))
+	// TODO
```
mlsteele: For future reference whenever this interpolation gets implemented copying `func substituteExact` from `pvl/helpers.go` could be a good way.

go/externals/services.go
```diff
@@ -4,37 +4,130 @@
...
+	}
+	for _, st := range staticServices {
+		p.register(st)
+	}
+	return &p
+}
+
+func (p *staticProofServices) register(st libkb.ServiceType) {
+	for _, k := range st.AllStringKeys() {
+		p.collection[k] = st
```
mlsteele: do you need `useDevelProofCheckers || !v.IsDevelOnly()` here too?

go/externals/services.go
```diff
@@ -4,37 +4,130 @@
...
+
+	loader := NullConfigLoader{}
+	proofTypes, err := loader.LoadAll()
+	if err != nil {
+		// TODO CORE-8655
+		p.G().Log.CDebugf(context.TODO(), "unable to load configs: %v", err)
+		return
+	}
+	for _, params := range proofTypes {
+		p.register(NewParamProofServiceType(params))
```
mlsteele: do you need useDevelProofCheckers || !v.IsDevelOnly() here too?

go/libkb/globals.go
```diff
@@ -1023,14 +1024,14 @@ func (g *GlobalContext) LogoutSelfCheck() error {
 }
 
 func (g *GlobalContext) MakeAssertionContext() AssertionContext {
-	if g.Services == nil {
+	if g.proofServices == nil {
```
mlsteele: need lock?

go/libkb/globals.go
```diff
@@ -1023,14 +1024,14 @@ func (g *GlobalContext) LogoutSelfCheck() error {
...
 		return nil
 	}
-	return MakeAssertionContext(g.Services)
+	return MakeAssertionContext(g.proofServices)
 }
 
-func (g *GlobalContext) SetServices(s ExternalServicesCollector) {
-	g.Services = s
+func (g *GlobalContext) SetProofServices(s ExternalServicesCollector) {
+	g.proofServices = s
```
mlsteele: need lock?

go/externals/proof_service_parameterized.go
joshblum: added as note

go/externals/services.go
joshblum: ah good catch

go/libkb/globals.go
joshblum: yes

go/libkb/globals.go
joshblum: yes

joshblum:
re the `*Static` methods i made the main constructors private so only the `NormalizeSocialAssertionStatic` and `AssertionParseAndOnlyStatic` methods are callable from `externals`. The code is used for more than test cases in kbfs since they just want to verify syntactical correctness of the assertion (but not that the service specified in the assertion is valid necessarily), i agree the naming could be better, but not sure what would be best. any ideas?

PR #13650, opened 2018-09-07
Build payment 'fromPrimaryAccount' arg

mlsteele:
Add `buildPaymentLocalArg.fromPrimaryAccount ` and `BuildPaymentResLocal.from`.

@buoyad


shared/actions/wallets.js
```diff
@@ -18,6 +18,7 @@ const buildPayment = (state: TypedState) =>
   RPCTypes.localBuildPaymentLocalRpcPromise({
     amount: state.wallets.buildingPayment.amount,
     // FIXME: Assumes XLM.
+    fromPrimaryAccount: false,
```
mlsteele: @buoyad this should be backwards compatible until someone gets to it

PR #13656, opened 2018-09-08
ignore link 98 of ajar's sigchain, which was a bad PGP update

maxtaco:
- we'll work on fixing the server bug that allowed this update
- clients rightly reject this signature
- tested to allow ajar to function properly post-reset


ajartille:
Pre-reset chats/teams working again, both old and new messages. Appreciate the weekend replies and super quick action to hotfix. Thanks 1e6 & kudos.

PR #13661, opened 2018-09-10
run team auditor on slow team loads

maxtaco:
- the team auditor more or less drops into place, but the one tricky aspect here is that we need to plumb through the "merkle head seqno", which is the last known merkle seqno when the team head was fabricated.
- we don't want to blow away loader caches, so if it's not there in the state, we backfill it, and check server replies against the previously verified chain link ID for seqno=1
- on a fresh load, we just load it from disk
- test that makes sure the back-fill of merkle head works


maxtaco:
@mlsteele this was a pretty easy integration and should hopefully be an easy review!

go/teams/audit.go
```diff
@@ -41,11 +41,23 @@ var mobileParams = AuditParams{
 	LRUSize:               500,
 }
 
+var devParams = AuditParams{
```
maxtaco: Don't want to slow down CI anymore.

go/teams/loader.go
```diff
@@ -1402,3 +1408,61 @@ func (l *TeamLoader) NotifyTeamRename(ctx context.Context, id keybase1.TeamID, n
...
+	}
+	if !headLink.LinkID().Eq(expectedLink) {
+		return ret, NewInvalidLink(headLink, "wrong head link hash: %s != %s", headLink.LinkID, expectedLink)
+	}
+	if headLink.isStubbed() {
+		return ret, NewInvalidLink(headLink, "got a stubbed head link, but wasn't expecting that")
+	}
+	headMerkle := headLink.inner.Body.MerkleRoot.ToMerkleRootV2()
+	state.HeadMerkle = &headMerkle
+	return headMerkle.Seqno, nil
```
mlsteele: Does anything here check that headLink.inner matches headLink.outer?

maxtaco: ah good point, i just assumed the unpack function did that, but maybe that was a bad assumption

maxtaco: Great catch on that one!

maxtaco:
thanks for the review!

PR #13662, opened 2018-09-10
Add routing for createNewAccount and linkExisting to Settings

nathunsmitty:
Previously, if you tried to create a new account or link an existing account from the settings, an error would occur due to the popups not existing under the settings route. This fixes that!


shared/wallets/routes.js
```diff
@@ -46,19 +46,16 @@ const walletChildren = {
         component: ConfirmForm,
         tags: makeLeafTags({layerOnTop: !isMobile}),
       },
-      linkExisting,
```
nathunsmitty: Also did some alphabetical sorting for eslint

nathunsmitty:
@buoyad yes!

shared/actions/json/wallets.json
```diff
@@ -40,11 +40,13 @@
     "createNewAccount": {
       "_description": "Add a new wallet to your account",
       "name": "string",
+      "show?": "boolean",
```
nathunsmitty: Since sometimes we want to navigate to our newly created account (creating a wallet from the wallet list) and other times just navigate up (creating a wallet from the send form), it seems like the best way to handle this is to pass in a prop from `routeProps` then pass that through the actions that handle the navigation.

I'm not a super big fan of the duplicated `show` prop in both the `create` and `created` forms of the new account and link account actions, but idk a better way to handle it

shared/wallets/create-account/container.js
```diff
@@ -13,8 +13,9 @@ const mapStateToProps = (state: TypedState) => ({
   waiting: anyWaiting(state, Constants.createNewAccountWaitingKey),
 })
 
-const mapDispatchToProps = (dispatch: Dispatch, {name, navigateUp}) => ({
-  _onCreateAccount: (name: string) => dispatch(WalletsGen.createCreateNewAccount({name})),
+const mapDispatchToProps = (dispatch: Dispatch, {navigateUp, routeProps}) => ({
+  _onCreateAccount: (name: string) =>
+    dispatch(WalletsGen.createCreateNewAccount({name, show: routeProps.get('show')})),
```
buoyad: can we rename `show` to `showOnCreation`? a bit ambiguous as it is

PR #13667, opened 2018-09-10
[KBFS-3235] Connect File Data

jzila:
Also PR feedback for #13496.

### DIFF
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -81,14 +81,12 @@ const loadUserFileEdits = (state: TypedState, action) =>
       const tlfUpdates = Constants.userTlfHistoryRPCToState(writerEdits || [])
       // Ensure we stat all path items for all updates, including paths to
       // navigate through to get to the updated files.
-      Saga.all(tlfUpdates.reduce((acc, u) =>
-        acc.concat(Types
-          .getPathElements(u.path)
-          .map((e, i, a) =>
-            Saga.put(FsGen.createFilePreviewLoad({
-              path: Types.getPathFromElements(a.slice(0, i + 1)),
-            })))
-        ), []))
+      Saga.all(tlfUpdates.reduce((acc: I.Map<Types.Path, any>, u) =>
+        Types.getPathElements(u.path).reduce((acc, e, i, a) => {
+          const path = Types.getPathFromElements(a.slice(0, i + 1))
+          return acc.set(path, Saga.put(FsGen.createFilePreviewLoad({path})))
+        }, acc), I.Map()).toList().toArray()
+      )
       return FsGen.createUserFileEditsLoaded({
         tlfUpdates: tlfUpdates,
       })
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -79,9 +79,16 @@ const loadUserFileEdits = (state: TypedState, action) =>
   RPCTypes.SimpleFSSimpleFSUserEditHistoryRpcPromise()
     .then(writerEdits => {
       const tlfUpdates = Constants.userTlfHistoryRPCToState(writerEdits || [])
-      Saga.all(tlfUpdates.map(u =>
-        Saga.put(FsGen.createFilePreviewLoad({path: u.path}))
-      ))
+      // Ensure we stat all path items for all updates, including paths to
+      // navigate through to get to the updated files.
+      Saga.all(tlfUpdates.reduce((acc, u) =>
+        acc.concat(Types
+          .getPathElements(u.path)
+          .map((e, i, a) =>
+            Saga.put(FsGen.createFilePreviewLoad({
+              path: Types.getPathFromElements(a.slice(0, i + 1)),
+            })))
+        ), []))
       return FsGen.createUserFileEditsLoaded({
         tlfUpdates: tlfUpdates,
       })
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -77,20 +77,20 @@ const filePreview = (state: TypedState, action) =>
 
 const loadUserFileEdits = (state: TypedState, action) =>
   RPCTypes.SimpleFSSimpleFSUserEditHistoryRpcPromise()
-    .then(writerEdits => {
-      const tlfUpdates = Constants.userTlfHistoryRPCToState(writerEdits || [])
+    .then(writerEdits => Constants.userTlfHistoryRPCToState(writerEdits || []))
+    .then(tlfUpdates =>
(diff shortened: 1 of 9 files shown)


shared/menubar/files-container.desktop.js
```diff
@@ -1,50 +1,32 @@
...
+  loadTlfUpdates: dispatchProps.loadTlfUpdates,
   userTlfUpdates: stateProps._userTlfUpdates.map(c => {
-    const {participants, teamname} = FsUtil.tlfToParticipantsOrTeamname(FsTypes.pathToString(c.tlf))
+    const tlf = FsTypes.pathToString(c.tlf)
+    const {participants, teamname} = FsUtil.tlfToParticipantsOrTeamname(tlf)
     const iconSpec = FsConstants.getIconSpecFromUsernamesAndTeamname([c.writer], null, stateProps._username)
     return {
       onSelectPath: () => dispatchProps._onSelectPath(c.tlf),
-      tlf: FsTypes.pathToString(c.tlf),
+      tlf,
```
songgao: Why not just pass the original `tlf` to component?

shared/menubar/files-container.desktop.js
```diff
@@ -60,6 +42,22 @@ const mergeProps = (stateProps, dispatchProps) => ({
...
+type TlfUpdateHocProps = {|
+  loadTlfUpdates: () => void,
+  userTlfUpdates: Array<UserTlfUpdateRowProps>,
+|}
+
+const TlfUpdateHoc = (ComposedComponent: React.ComponentType<any>) =>
+  class extends React.PureComponent<TlfUpdateHocProps> {
+    componentDidMount() {
+      this.props.loadTlfUpdates()
+    }
```
songgao: If the content relies on explicit `loadTlfUpdates`, it seems we might also need to call it in `componentDidUpdate`.

shared/menubar/files.desktop.js
```diff
@@ -46,14 +43,55 @@ const FileUpdate = (props: FileUpdateProps) => (
...
+
+const FileUpdatesHoc = (ComposedComponent: React.ComponentType<any>) =>
+  class extends React.PureComponent<FileUpdatesProps, FileUpdatesState> {
+    state = {
+      isShowingAll: false,
+    }
+    render() {
+      return <ComposedComponent {...this.props} onShowAll={() => this.setState({isShowingAll: !this.state.isShowingAll})} isShowingAll={this.state.isShowingAll} />
+    }
+  }
```
songgao: Hmm is there a reason you didn't just make `FileUpdatesShowAll` into a class and do these there? It seems clear that way.

shared/menubar/files.desktop.js
```diff
@@ -46,14 +43,55 @@ const FileUpdate = (props: FileUpdateProps) => (
...
+      <FileUpdate key={u.name} {...u} tlfType={props.tlfType} />
     ))}
+    {props.updates.length > defaultNumFileOptionsShown && (
+      // $FlowIssue ¯\_(ツ)_/¯
+      <FileUpdatesShowAll onShowAll={props.onShowAll} isShowingAll={props.isShowingAll} numUpdates={props.updates.length} />
+    )}
   </Kb.Box2>
 )
 
+const ComposedFileUpdates = compose(FileUpdatesHoc)(FileUpdates)
```
songgao: This could just be `const ComposedFileUpdate = FileUpdatesHoc(FileUpdates)`. Also, it seems `onShowAll` and `isShowingAll` is only used in `FileUpdatesShowAll`, so if you choose to use HOC, it seems the HOC can wrap around `FileUpdatesShowAll` to avoid plumbing these through `FileUpdates`.

shared/menubar/files.desktop.js
```diff
@@ -79,12 +117,12 @@ const UserTlfUpdateRow = (props: UserTlfUpdateRowProps) => (
...
         </Kb.Text>
       </Kb.Box2>
-      <FileUpdates updates={props.updates} moreUpdateCount={props.moreUpdateCount} />
+      <ComposedFileUpdates updates={props.updates} tlfType={props.tlfType} />
     </Kb.Box2>
   </Kb.Box2>
 )
 
-export const FilesPreview = ({onViewAll, userTlfUpdates}: FilesPreviewProps) => (
+export const FilesPreview = ({onShowAll, userTlfUpdates}: FilesPreviewProps) => (
```
songgao: `onShowAll` is not used here.

shared/menubar/files-container.desktop.js
jzila: Because the component needs it as a string.

shared/menubar/files.desktop.js
jzila: Yeah, could do that.

shared/menubar/files.desktop.js
jzila: Nope, `isShowingAll` is also used in `FileUpdates`: https://github.com/keybase/client/pull/13667/files#diff-41bacb263b756a2ef2365b2ef779dd93R83

shared/menubar/files-container.desktop.js
songgao: What I was trying to ask was why doesn't component just take a `Types.Path` instead.

shared/menubar/files-container.desktop.js
jzila: Hmm, that's called _all the time_. I'll have to make that part respond to notifications.

jzila: Will do in a separate PR.

shared/menubar/files.desktop.js
jzila: I couldn't get it to work without a HOC.

shared/actions/fs/index.js
```diff
@@ -77,15 +77,20 @@ const filePreview = (state: TypedState, action) =>
...
+    .then(tlfUpdates =>
+      // Ensure we stat all path items for all updates, including paths to
+      // navigate through to get to the updated files.
+      Promise.all(tlfUpdates.reduce((acc: I.Map<Types.Path, any>, u) =>
+        Types.getPathElements(u.path).reduce((acc, e, i, a) => {
+          const path = Types.getPathFromElements(a.slice(0, i + 1))
+          return acc.set(path, () => Saga.put(FsGen.createFilePreviewLoad({path})))
+        }, acc), I.Map()).toList().toArray()
+      )
+      .then(() => FsGen.createUserFileEditsLoaded({
```
jzila: @chrisnojima will this work to trigger all the actions appropriately?

PR #13670, opened 2018-09-10
dont allow inbox call while logged out

chrisnojima:
@keybase/react-hackers

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -36,6 +36,10 @@ const inboxRefresh = (
   action: Chat2Gen.InboxRefreshPayload | Chat2Gen.LeaveConversationPayload,
   state: TypedState
 ) => {
+  if (!state.config.loggedIn) {
+    return
+  }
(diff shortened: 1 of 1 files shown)


shared/actions/chat2/index.js
```diff
@@ -36,6 +36,10 @@ const inboxRefresh = (
   action: Chat2Gen.InboxRefreshPayload | Chat2Gen.LeaveConversationPayload,
   state: TypedState
 ) => {
+  if (!state.config.loggedIn) {
+    return
+  }
+
```
songgao: At the end of this function we are still [returning something](https://github.com/keybase/client/pull/13670/files#diff-62d6b6a8baeb0bb4a699ed1965bcbf00R63). There's not action creator for the event that uses this. Is the return at the end still necessary?

Same for the `unboxRows` below.

chrisnojima: ? there's a safeTakeEveryPure

PR #13674, opened 2018-09-10
Prompt to uninstall outdated Dokan in the GUI

zanderz:
Note that this replaces https://github.com/keybase/client/pull/13648 , which was inferior because it tried to do the registry operations separately on the GUI side instead of having the service do it as part of the fuse status info. The uninstaller still has to be invoked from js though.


shared/actions/fs/platform-specific.desktop.js
```diff
@@ -323,12 +344,13 @@ function* platformSpecificSaga(): Saga.SagaGenerator<any, any> {
   yield Saga.safeTakeEvery(FsGen.fuseStatus, fuseStatusSaga)
   yield Saga.safeTakeEveryPure(FsGen.fuseStatusResult, fuseStatusResultSaga)
   yield Saga.safeTakeEveryPure(FsGen.installKBFS, RPCTypes.installInstallKBFSRpcPromise, installKBFSSuccess)
-  yield Saga.safeTakeEveryPure(FsGen.uninstallKBFSConfirm, uninstallKBFSConfirm, uninstallKBFSConfirmSuccess)
   yield Saga.actionToAction(FsGen.openAndUpload, openAndUpload)
   if (isWindows) {
     yield Saga.safeTakeEveryPure(FsGen.installFuse, installDokanSaga)
+    yield Saga.safeTakeEveryPure(FsGen.uninstallKBFSConfirm, uninstallDokanSaga)
```
jzila: This isn't pure. You can use `actionToPromise` though, with a tiny refactor of `uninstallDokanSaga`.

shared/actions/fs/platform-specific.desktop.js
```diff
@@ -288,10 +288,31 @@ function installCachedDokan() {
...
   return Saga.call(installCachedDokan)
 }
 
+function* uninstallDokanSaga() {
+  const state: TypedState = yield Saga.select()
+  const uninstallString = Constants.kbfsUninstallString(state)
+  if (uninstallString) {
+    yield Saga.call(uninstallDokan, uninstallString)
+  }
+}
```
jzila: If you refactor this a bit, you can use `actionToPromise` below.
```javascript
const uninstallDokanPromise = (state: TypedState) => {
  const uninstallString = Constants.kbfsUninstallString(state)
  if (uninstallString) {
    return uninstallDokan(uninstallString)
  }
  return new Promise(resolve => resolve())
}
```

go/install/fuse_status_windows.go
```diff
@@ -51,6 +54,52 @@ func detectDokanDll(dokanPath string, log Log) bool {
 	return exists
 }
 
+// Read all the uninstall subkeys and find the ones with DisplayName starting with "Dokan Library"
+// and containing "Bundle"
+func findDokanUninstall(wow64 bool) (result string) {
+	var access uint32 = registry.ENUMERATE_SUB_KEYS | registry.QUERY_VALUE
+	// Assume this is build 32 bit, so we need this flag to see 64 but registry
```
jzila: s/but/bit

shared/fs/banner/fileui-banner/index.js
```diff
@@ -27,6 +29,11 @@ const Banner = ({kbfsEnabled, showBanner, onInstall, onDismiss, openInFileUI}: P
     alignItems: 'center',
     position: 'relative',
   }
+  const promptText = kbfsOutdated ? ( dokanUninstall ?
```
jzila: Not sure you need the parenthesis. Usually we format these like:
```javascript
const promptText = kbfsOutdated
  ? dokanUninstall
    ? 'A newer version...'
    : 'A newer version...'
  : 'Get access to your files...'
```

shared/actions/fs/platform-specific.desktop.js
zanderz: good idea, thanks

go/install/fuse_status_windows.go
zanderz: done

go/install/fuse_status_windows.go
```diff
@@ -51,6 +54,52 @@ func detectDokanDll(dokanPath string, log Log) bool {
...
+		return
+	}
+	for _, name := range names {
+		subKey, err := registry.OpenKey(k, name, registry.QUERY_VALUE)
+		if err != nil {
+			fmt.Printf("Error %s opening subkey %s\n", err.Error(), name)
+		}
+
+		displayName, _, err := subKey.GetStringValue("DisplayName")
+		if err == nil && strings.HasPrefix(displayName, "Dokan Library") && strings.Contains(displayName, "Bundle") {
```
jzila: Are we swallowing the error here? Would be clearer if it was:
```go
// Some package scoped region above:
dokanRegexp := regexp.MustCompile("^Dokan Library.*Bundle")
...
if err != nil {
    fmt.Printf("...")
    continue
}
if !dokanRegexp.MatchString(displayName) {
    continue
}
```

zanderz: The error isn't worth logging

shared/fs/banner/fileui-banner/index.js
zanderz: missed this one. ok.

shared/actions/fs/platform-specific.desktop.js
zanderz: actually, I found I wanted to get the fuse status again after the synchronous uninstall, so I changed it back except not pure.

zanderz:
I tried to not use a generator by various combinations of `actionToPromise`, `actionToAction`, using `chiled_process.exec` instead of `execSync`, etc., and none of it worked so I put it back this way. Again, all it needs to do is get a string from the store, invoke it as an external process, then do an action when finished.
@chrisnojima

shared/actions/fs/platform-specific.desktop.js
```diff
@@ -287,10 +287,32 @@ function installCachedDokan() {
   })
 }
 
+function uninstallDokan(uninstallString: string): Promise<boolean> {
+  return new Promise((resolve, reject) => {
+    logger.info('Invoking dokan uninstaller')
+    try {
+      execSync(uninstallString, {windowsHide: true})
```
chrisnojima: we shouldn't ever make sync calls, the entire ui will just hang if we do this

jzila: Yeah that's what we were talking about in chat.

chrisnojima: i think this could be something like
```
const uninstallDokanSaga = (state: TypedState) => {
    const uninstallString = Constants.kbfsUninstallString(state)
    if (!uninstallString) {
        return 
    }
    logger.info('Invoking dokan uninstaller')
    return new Promise(resolve => {
        try {
            exec(uninstallString, {windowsHide: true}, resolve)
        }  catch (e) {
            logger.error('uninstallDokan caught', err)
            resolve()
        }
    }).then(() => FsGen.createFuseStatus())
}

...
    yield Saga.actionToPromise(FsGen.uninstallKBFSConfirm, uninstallDokanSaga)

```

zanderz: huh, I thought that was the whole point of a Promise, to synchronize lengthy things.

Anyway, thanks Nojima, passing resolve as the callback is what I was missing, I guess my promise wasn't ever resolving when i tried it that way.

chrisnojima: this is incorrect. Promises are just a way to control async stuff.  Please read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises

PR #13676, opened 2018-09-10
fix t.forEach error when enabling finder integration

songgao:


shared/actions/fs/platform-specific.desktop.js
```diff
@@ -322,7 +321,7 @@ function* platformSpecificSaga(): Saga.SagaGenerator<any, any> {
   yield Saga.safeTakeEveryPure(FsGen.openInFileUI, openInFileUISaga)
   yield Saga.safeTakeEvery(FsGen.fuseStatus, fuseStatusSaga)
   yield Saga.safeTakeEveryPure(FsGen.fuseStatusResult, fuseStatusResultSaga)
-  yield Saga.safeTakeEveryPure(FsGen.installKBFS, RPCTypes.installInstallKBFSRpcPromise, installKBFSSuccess)
```
songgao: @jzila Here's the bug: `RPCTypes.installInstallKBFSRpcPromise` accepts [two parameters](https://github.com/keybase/client/blob/a8b0cff5946e1438c24a06c4799e3161d0760296/shared/constants/types/rpc-gen.js#L817): `params` and `waitingKey`. `Saga.safeTakeEveryPure` [passes `state` as the second parameter](https://github.com/keybase/client/blob/a8b0cff5946e1438c24a06c4799e3161d0760296/shared/util/saga.js#L112). So here basically we end up passing the entire `state` as `waitingKey`, causing it to error [here](https://github.com/keybase/client/blob/a8b0cff5946e1438c24a06c4799e3161d0760296/shared/reducers/waiting.js#L13)

songgao: also @chrisnojima any idea why flow missed this?

chrisnojima: yes, so the flow definitions for safeTakeEveryPure are a little crappy.
because the worker is defined as
```
  pureWorker: ((action: A, state: TypedState) => any) | ((action: A) => any),
```
it seemingly treats this function which takes `(action, somethingElse)` as compatible as the second part `(action:A) => any`
if we instead change the type to
```
  pureWorker: (action: A, state: TypedState) => any,
```
it actually does catch this error

screen shot 2018-09-10 at 8 40 34 pm

songgao: Ah that makes sense. Thanks! I guess that why actionToAction/actionToPromise just always takes a function with two parameters and use `state` as first one.

PR #13677, opened 2018-09-10
fix installer to look for kext in correct location

songgao:


packaging/desktop/kbfuse.sh
```diff
@@ -15,7 +15,7 @@ cd $dir
 client_dir="$dir/../.."
 fuse_dir="$client_dir/osx/Fuse"
 tmp_dir="/tmp/desktop-kbfuse"
-installer_url="https://prerelease.keybase.io/darwin-package/KeybaseInstaller-1.1.63-darwin.tgz"
+installer_url="https://prerelease.keybase.io/darwin-package/KeybaseInstaller-1.1.64-darwin.tgz"
```
jzila: Is this something we'll have to update all the time?

songgao: Only when we change the installer or upgrade kbfuse, I think. In this case the fix is in the installer, and 1.1.63 is already in admin build, so it'd need be bump.

PR #13679, opened 2018-09-11
Nojima/desktop better take every type

chrisnojima:
@keybase/react-hackers makes the type stronger to catch the forEach issue song ran into.
cc: @songgao

### DIFF
--- a/shared/util/saga.js
+++ b/shared/util/saga.js
@@ -98,22 +98,15 @@ function actionToAction<A, E>(pattern: RS.Pattern, f: (state: TypedState, action
 // It can also return all([put(action1), put(action2)]) to dispatch multiple actions
 function safeTakeEveryPure<A, R, FinalEffect, FinalErrorEffect>(
   pattern: RS.Pattern,
-  pureWorker: ((action: A, state: TypedState) => any) | ((action: A) => any),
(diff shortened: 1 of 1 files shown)


shared/util/saga.js
```diff
@@ -98,22 +98,15 @@ function actionToAction<A, E>(pattern: RS.Pattern, f: (state: TypedState, action
 // It can also return all([put(action1), put(action2)]) to dispatch multiple actions
 function safeTakeEveryPure<A, R, FinalEffect, FinalErrorEffect>(
   pattern: RS.Pattern,
-  pureWorker: ((action: A, state: TypedState) => any) | ((action: A) => any),
+  pureWorker: (action: A, state: TypedState) => any,
```
jzila: Would it work properly if you did `(action: A, state?: TypedState)`?

chrisnojima:
I don't think there's any value in doing this arity thing

shared/util/saga.js
songgao: I assume that would mean a different function signature for `pureWorker`, and all users of `safeTakeEveryPure` that needs the `state` would need to test if it's `undefined` or not.

chrisnojima:
it means we alwasy pass state, even if its not used. its not expensive to get it so lets just do it always

PR #13687, opened 2018-09-11
fix paper key creation

chrisnojima:
@keybase/react-hackers


shared/actions/devices.js
```diff
@@ -25,7 +25,7 @@ const requestPaperKey = () =>
           phrase,
         }: RPCTypes.LoginUiDisplayPaperKeyPhraseRpcParam) =>
           Saga.put(DevicesGen.createPaperKeyCreated({paperKey: new HiddenString(phrase)})),
-        'keybase.1.loginUi.promptRevokePaperKeys': (_, response, __) => {
```
chrisnojima: all this arity stuff is too subtle. i'll fix this after the release

PR #13693, opened 2018-09-11
iOS share extension CORE-8742

mmaxim:
**Go**
Main challenges to overcome were the hard memory limits, the separation of the extension and main app, and the reality that multiple extension processes can be running simultaneously (combined with LevelDB only being usable by a single process). 

1.) Add a new file to `bind` for handling extension functionality, called `extension.go`. All the important functionality is in here, `ExtensionInit`, `ExtensionGetInbox`, and `ExtensionPostXXX`. 
2.) `ExtensionInit` starts up the minimal amount of stuff to send chat messages.
3.) `ExtensionGetInbox` reads the shared inbox from the shared container directory.
4.)  Saves the last conversation sent into, so we can initialize into that on subsequent share attempts.
5.) Add option to disable attachment multi upload and enable for the extension.
6.) Add option to set the HTTP server start port, and set it to something different in the extension.
7.) Don't attempt to lock a PID file on iOS only. It isn't really useful on mobile and blocks multiple extensions from running.
8.) Add ability to receive an attachment preview from the caller in byte form. 

**Native**
Main idea here is to use the `SLComposeServiceViewController` to drive the Go functions from above. 

1.) Add the extension project to the Keybase iOS workspace, and set up entitlements so that it uses app groups to get at shared data.
2.) Add `ShareViewController` to implement  `SLComposeServiceViewController` base class. 
3.) Add `ConversationViewController` for selecting which conversations to send into.
4.) Add `InitFailedViewController` to display when we are unable to initialize Go (likely the user is logged out).
5.) `ShareViewController` will call `ExtensionInit` and display a wait state until it returns. Once it does, we attempt to populate the conversation field with the last conversation sent to. If nothing, we prompt for something. The user then sees a table view of their inbox from `ExtensionGetInbox`. Upon the user hitting `Post`, we attempt to figure out what kind of chat message to send given what is given by the base class. Once we have the best message, we invoke the fitting `ExtensionPostXXX` function to send it.
6.) On success we display a local push indicating it worked, the same as we do with an error. 
7.) Add native functions for calculating previews for images and videos. The Go versions of these functions used too much memory to be usable in the extension.


go/chat/attachments/preview.go
```diff
@@ -99,17 +99,19 @@ func previewVideoBlank(ctx context.Context, g *globals.Context, log utils.DebugL
...
 	if err != nil {
 		return nil, err
 	}
 
 	width, height := previewDimensions(img.Bounds())
 
-	// nfnt/resize with NearestNeighbor is the fastest I've found.
-	preview := resize.Resize(width, height, img, resize.Bicubic)
+	log.Debug(ctx, "previewImage: resizing image: bounds: %s", img.Bounds())
+	preview := resize.Resize(width, height, img, resize.NearestNeighbor)
```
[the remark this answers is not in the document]
mmaxim: That's a mistake, I tried it to see if it would work (it didn't). I have a local change here I forgot to commit, sorry!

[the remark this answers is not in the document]
malgorithms: oh yeah nearest neighbor almost always looks really bad when scaling images.  cool, just saw @mmaxim 's comment that it was a mistake.

go/chat/attachments/s3.go
```diff
@@ -87,7 +87,13 @@ func (a *S3Store) PutS3(ctx context.Context, r io.Reader, size int64, task *Uplo
 	region := a.regionFromParams(task.S3Params)
 	b := a.s3Conn(task.S3Signer, region, task.S3Params.AccessKey).Bucket(task.S3Params.Bucket)
 
-	if size <= minMultiSize {
+	multiPartUpload := size <= minMultiSize
+	if multiPartUpload && a.env.GetAttachmentDisableMulti() {
```
joshblum: `||`?

go/bind/extension.go
```diff
@@ -0,0 +1,531 @@
...
+	var msg string
+	if err != nil {
+		msg = fmt.Sprintf("We could not send your %s. Please try from the Keybase app.", typ)
+	} else {
+		msg = fmt.Sprintf("Your %s was shared successfully.", typ)
+	}
+	pusher.LocalNotification("extension", msg, -1, "default", strConvID, "chat.extension")
+}
+
+func ExtensionPostJPEG(strConvID, name string, public bool, membersType int,
```
patrickxb: (discussed in chat that it might be a good idea to also have ExtensionPostPNG if the original is a PNG file)

go/chat/attachments/s3.go
```diff
@@ -87,7 +87,13 @@ func (a *S3Store) PutS3(ctx context.Context, r io.Reader, size int64, task *Uplo
 	region := a.regionFromParams(task.S3Params)
 	b := a.s3Conn(task.S3Signer, region, task.S3Params.AccessKey).Bucket(task.S3Params.Bucket)
 
-	if size <= minMultiSize {
+	multiPartUpload := size <= minMultiSize
```
patrickxb: `multiPartUpload := size > minMultiSize` ???

shared/react-native/ios/KeybaseShare/ShareViewController.m
```diff
@@ -0,0 +1,301 @@
...
+  for (NSItemProvider* a in attachments) {
+    if (cond(a)) {
+      [res addObject:a];
+    }
+  }
+  return res;
+}
+
+- (BOOL)isRealURL:(NSItemProvider*)item {
+  // "file URLs" also have type "url", but we want to treat them as files, not text.
```
joshblum: can this be `isTextURL` or something i think it's less confusing

go/bind/extension.go
```diff
@@ -0,0 +1,531 @@
...
+var extensionInitMu sync.Mutex
+
+func ExtensionIsInited() bool {
+	extensionInitMu.Lock()
+	defer extensionInitMu.Unlock()
+	return extensionInited
+}
+
+func ExtensionInit(homeDir string, mobileSharedHome string, logFile string, runModeStr string,
+	accessGroupOverride bool, externalDNSNSFetcher ExternalDNSNSFetcher, nvh NativeVideoHelper) (err error) {
```
joshblum: why do we need `ExternalDNSNSFetcher` and `NativeVideoHelper` if we just pass in `NULL` from native?

go/chat/attachments/s3.go
mmaxim: This is probably why the tests are failing, big mistake on my part, thanks!

go/bind/extension.go
mmaxim: It's for Android, I can just call into native directly from Go on iOS.

mmaxim: Although I guess for the extension there is no Android so I could get rid of this.

go/chat/attachments/s3.go
mmaxim: I think `&&` is correct. I wanted to only print that line if we turn that variable off because of the config. So if `multiPartUpload` is true, and we disabled it, then it logs that line and sets it to false.

go/bind/extension.go
mmaxim: Filed as another ticket, right now PNG works, but it will have the wrong MIME type (and a JPEG preview).

shared/react-native/ios/KeybaseShare/ShareViewController.m
mmaxim: How about `isWebURL`?

go/chat/attachments/s3.go
joshblum: ah gotcha i guess because the condition was flipped i was thinking we wouldn't properly set `multiPartUpload = false` but the fix makes it make sense

shared/react-native/ios/KeybaseShare/ShareViewController.m
joshblum: sounds good!

shared/react-native/ios/KeybaseShare/ConversationViewController.m
```diff
@@ -0,0 +1,134 @@
...
+     [NSLayoutConstraint constraintWithItem:av
+                                  attribute:NSLayoutAttributeCenterY
+                                  relatedBy:NSLayoutRelationEqual
+                                     toItem:self.tableView
+                                  attribute:NSLayoutAttributeCenterY
+                                 multiplier:1 constant:0]
+     ]
+   ];
+  
+  dispatch_async(dispatch_get_main_queue(), ^{
```
chrisnojima: we might want to trampoline this, lets discuss

shared/react-native/ios/KeybaseShare/ShareViewController.m
```diff
@@ -0,0 +1,301 @@
...
+}
+
+- (void)createImagePreview:(NSURL*)url resultCb:(void (^)(int,int,int,int,NSData*))resultCb  {
+  UIImage* original = [UIImage imageWithData:[NSData dataWithContentsOfURL:url]];
+  CFURLRef cfurl = CFBridgingRetain(url);
+  CGImageSourceRef is = CGImageSourceCreateWithURL(cfurl, nil);
+  NSDictionary* opts = [[NSDictionary alloc] initWithObjectsAndKeys:
+                        (id)kCFBooleanTrue, (id)kCGImageSourceCreateThumbnailWithTransform,
+                        (id)kCFBooleanTrue, (id)kCGImageSourceCreateThumbnailFromImageIfAbsent,
+                        [NSNumber numberWithInt:640], (id)kCGImageSourceThumbnailMaxPixelSize,
```
chrisnojima: you should be able to use `@640`

shared/react-native/ios/KeybaseShare/ShareViewController.m
```diff
@@ -0,0 +1,301 @@
...
+    // Just bail out of here if nothing was selected
+    [self maybeCompleteRequest:YES];
+    return;
+  }
+  NSArray* items = [self getSendableAttachments];
+  if ([items count] == 0) {
+    [self maybeCompleteRequest:YES];
+    return;
+  }
+  for (int i = 0; i < [items count]; i++) {
```
chrisnojima: could use https://developer.apple.com/documentation/foundation/nsarray/1415846-enumerateobjectsusingblock

PR #13699, opened 2018-09-11
WIP: change attachment api to reduce render thrashing

chrisnojima:


chrisnojima:
@keybase/react-hackers

chrisnojima:
danny found a bug where toast isn't working, i'm debugging now

chrisnojima:
@buoyad @cjb new commit fixes the copy text toast. 2 problems
1. i had to change the floating box componentDidUpdate to calc the targetRect in a smarter way
2. changed the copyText to not rerender while its popping up the toast, else the refs would get nulled and set again and the wrong thing would happen

shared/fs/common/path-item-action.js
```diff
@@ -203,7 +203,7 @@ const PathItemAction = (props: Props & OverlayParentProps) => {
       <FloatingMenu
         closeOnSelect={false}
         containerStyle={styles.floatingContainer}
-        attachTo={props.attachmentRef}
+        attachTo={props.getAttachmentRef}
```
jzila: @chrisnojima This doesn't work--it's broken in master.
```
TypeError: this.props.attachTo is not a function
```

shared/fs/header/add-new.js
```diff
@@ -122,7 +122,7 @@ const AddNew = (props: AddNewProps & OverlayParentProps) => {
           )}
         </ClickableBox>
         <FloatingMenu
-          attachTo={props.attachmentRef}
+          attachTo={props.getAttachmentRef}
```
jzila: Broken here too.

PR #13700, opened 2018-09-11
Git load users better

mlsteele:
Listing a git repo sometimes failed if it was last written by a deleted user.

Reproed. It's nondeterministic and seemingly rare. I don't know why it is nondeterministic. But this seems to fix it.

### DIFF
--- a/go/git/get.go
+++ b/go/git/get.go
@@ -230,7 +230,9 @@ func getMetadataInnerSingle(ctx context.Context, g *libkb.GlobalContext,
 	}
 
 	// Load UPAKs to get the last writer username and device name.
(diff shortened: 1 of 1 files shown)


go/git/get.go
```diff
@@ -230,7 +230,9 @@ func getMetadataInnerSingle(ctx context.Context, g *libkb.GlobalContext,
 	}
 
 	// Load UPAKs to get the last writer username and device name.
-	lastWriterUPAK, _, err := g.GetUPAKLoader().LoadV2(libkb.NewLoadUserArg(g).WithUID(responseRepo.LastModifyingUID))
+	lastWriterUPAK, _, err := g.GetUPAKLoader().LoadV2(libkb.NewLoadUserArgWithContext(ctx, g).
+		WithUID(responseRepo.LastModifyingUID).
+		WithPublicKeyOptional())
```
mlsteele: I think that this missing `PublicKeyOptional` was leading to the `No public key found` error. Trap argument

PR #13720, opened 2018-09-12
[KBFS-3150] Mobile: copy should say "Create new folder", not "New folder"

brayoh:
#### Description of task:
In the mobile action sheet, when tapping on the "+ Add" button, it should say "Create new folder", not just "New folder".

#### Relevant Screenshots:
screen shot 2018-09-12 at 22 35 20

### DIFF
--- a/shared/fs/header/add-new.js
+++ b/shared/fs/header/add-new.js
@@ -74,7 +74,7 @@ const propsToMenuItems = (props: AddNewProps) => {
       icon: 'iconfont-upload',
     })
   props.pickAndUploadPhoto && props.pickAndUploadVideo && items.push('Divider')
-  items.push({title: 'Create New folder', onClick: props.newFolderRow, icon: 'iconfont-folder-new'})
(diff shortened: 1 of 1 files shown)


shared/fs/header/add-new.js
```diff
@@ -74,7 +74,7 @@ const propsToMenuItems = (props: AddNewProps) => {
       icon: 'iconfont-upload',
     })
   props.pickAndUploadPhoto && props.pickAndUploadVideo && items.push('Divider')
-  items.push({title: 'New folder', onClick: props.newFolderRow, icon: 'iconfont-folder-new'})
+  items.push({title: 'Create New folder', onClick: props.newFolderRow, icon: 'iconfont-folder-new'})
```
nathunsmitty: I think we may want to lowercase the "N" in "New" though so the text is `Create new folder`

brayoh: Cool, let me change that

PR #13726, opened 2018-09-12
make it harder to have an inbox with a stale UID in it

mmaxim:


joshblum:
note to check the ephemeral purger/conv loader for proper shutdown

PR #13727, opened 2018-09-12
get rid of GlobUI

maxtaco:
- this should fix the crash we saw in keybase/keybase-issues#3278


zapu:
`push` still crashes for me:
```
» kb push
Running `go install github.com/keybase/client/go/keybase`...Done.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x70 pc=0x156d5af]
```

keybase-travis:
Ok thanks, didn’t get to test that yet.  Hmmm.

keybase-travis:
Hmmm it’s no surprise, likely the pointer isn’t set yet. I will just change
it to use fmt.printf

zapu:
I can do it and commit to this branch while I'm on this.

zapu:
BTW Github formats your responses as coming from "keybase-travis" and it's funny and scary at the same time.

what if AI is already sending patches to software on Github? 🛸

PR #13730, opened 2018-09-13
refactor mobile push to avoid a type of error CORE-8817

mmaxim:
If the user get an audible push notification from an edit (because an @ message of them was edited into it), then we can get a black bar if that notification is clicked. The reason is that the logic for unboxing a message on an audible notification for caching purposes still attempts to compute the local notification text for a silent notification. That fails for an edit, and the user is presented with a black bar on Android.


go/chat/mobilepush.go
```diff
@@ -0,0 +1,182 @@
...
+	return &MobilePush{
+		Contextified: globals.NewContextified(g),
+		DebugLabeler: utils.NewDebugLabeler(g.GetLog(), "MobilePush", false),
+	}
+}
+
+func (h *MobilePush) AckNotificationSuccess(ctx context.Context, pushIDs []string) {
+	defer h.Trace(ctx, func() error { return nil }, "AckNotificationSuccess")()
+	// Get session token
+	nist, _, err := h.G().ActiveDevice.NISTAndUID(ctx)
```
joshblum: can this get conn/nist situation get wrapped up in a helper? didn't we need this in the extension as well?

go/chat/mobilepush.go
```diff
@@ -0,0 +1,182 @@
...
+	defer conn.Shutdown()
+
+	// Make remote successful call on our ad hoc conn
+	cli := chat1.RemoteClient{Cli: NewRemoteClient(h.G(), conn.GetClient())}
+	if err = cli.RemoteNotificationSuccessful(ctx,
+		chat1.RemoteNotificationSuccessfulArg{
+			AuthToken:        gregor1.SessionToken(nist.Token().String()),
+			CompanionPushIDs: pushIDs,
+		}); err != nil {
+		h.Debug(ctx, "AckNotificationSuccess: failed to invoke remote notification success: %",
```
joshblum: `%v", err)`

mmaxim: I think this is fixed in my extension PR, its been like this forever,

go/chat/mobilepush.go
mmaxim: Sure

mmaxim:
@joshblum PTAL

PR #13739, opened 2018-09-13
auditor should work on a stale team

maxtaco:
- it could be that the team we have locally is stale, and the merkle roots will show links off the end of our
  chain. in this case, the audit is still legit, don't crash out

### DIFF
--- a/go/teams/audit.go
+++ b/go/teams/audit.go
@@ -232,13 +232,6 @@ func (a *Auditor) doPostProbes(m libkb.MetaContext, history *keybase1.AuditHisto
 		return maxMerkleProbe, nil
 	}
 
-	// missedTeamSeqnos keeps track of all of the team seqnos that were advertised in the merkle
-	// tree, but that we didn't have data for. This can happen if we have a stale view of the team,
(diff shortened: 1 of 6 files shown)


go/teams/audit.go
```diff
@@ -244,7 +251,8 @@ func (a *Auditor) doPostProbes(m libkb.MetaContext, history *keybase1.AuditHisto
 		if tuple.team > keybase1.Seqno(0) {
 			expectedLinkID, ok := chain[tuple.team]
 			if !ok {
-				return maxMerkleProbe, NewAuditError("team chain doesn't have a link for seqno %d, but expected one", tuple.team)
+				missedTeamSeqnos[tuple.team] = struct{}{}
+				continue
```
mlsteele: What's wrong with using maxChainSeqno?
```
if !ok {
  if tuple.team > maxChainSeqno {
    // Team seqnos found in the merkle tree may be ahead of what we have locally.
    continue
  }
  return maxMerkleProbe, NewAuditError("team chain doesn't have a link for seqno %d, but expected one", tuple.team)	
}
```

maxtaco:
@mlsteele now with a test

PR #13744, opened 2018-09-13
Removed paddings on mobile message box

cecileboucheron:
@keybase/react-hackers


shared/chat/conversation/input-area/normal/platform-input.native.js
```diff
@@ -335,7 +335,7 @@ const styles = styleSheetCreate({
   },
   container: {
     ...globalStyles.flexBoxRow,
-    alignItems: 'flex-end',
+    alignItems: 'center',
```
chrisnojima: are you sure you want this. does this look on when there's multiple lines?

cecileboucheron: `actionButton` has its own `alignSelf` so we're good.

PR #13748, opened 2018-09-14
Use action parameter for wallet send from field

zanderz:


zanderz:
createSetBuildingFrom has a from field, and it gets routed to here. I watched the from field come in in the debugger.

buoyad:
@zanderz oh right, I had a brain fart. The reducers put all those fields in the store, so I think this should use `state.wallets.buildingPayment.from` if it's not empty and pass in `fromPrimaryAccount: true` if it is empty

zanderz:
I have a couple of questions - doesn't the buildPayment RPC get called once for each of the 6 actions? is that intended? Also, can we be sure the reducer happens to put `from` in the state before the buildPayment promise is invoked with `from` ?

buoyad:
Yeah, the reducers run first so `from` will be there. This gets called when the user updates any field on the send form.

zanderz:
so, is this ok now?

buoyad:
merging this for use in the remove load call PR

PR #13753, opened 2018-09-14
[KBFS-3267] "EMPTY" tag renders off screen on mobile devices

brayoh:
#### Description of task:
when opening a private folder between multiple people and then backing out.
On mobile
    Go to a shared folder that is empty
    Click on the shared folder (goes into an empty folder view)
    Click the back button to the list of private folders view
    The empty tag will render at the end of the folder participants (off screen on mobile)

#### Relevant Screenshots:
screen shot 2018-09-14 at 16 45 12


shared/fs/row/common.js
```diff
@@ -39,6 +39,9 @@ const rowText = Styles.platformStyles({
     textOverflow: 'ellipsis',
     whiteSpace: 'nowrap',
   },
+  isMobile: {
+    maxWidth: '75%',
```
jzila: I'm not sure this will scale well for different screen sizes, since the `EMPTY` tag is a fixed width.

brayoh: The best way to scale out is using percentage instead of pixels, the percentage will occupy the specified percentage of the screen unlike pixels which vary according to the device screen.

brayoh: In the docs, they advice you use points or percentages https://facebook.github.io/react-native/docs/layout-props#maxwidth

brayoh: screen shot 2018-09-14 at 20 41 22

I've tried on the various screen sizes, it's working as expected

chrisnojima:
i agree that we likely don't want percentage here. there's a fixed amount of space on the right so likely the layout of the container needs to be constrained differently

brayoh:
@chrisnojima I'm a bit lost, are you suggesting we use pixels?

chrisnojima:
Maybe i'm misunderstanding what this ticket is trying to accomplish. Are the screenshots with it fixed? What does it look like when its not fixed. This is my assumption:
You come in and the [emtpy] tag is too far to the right
So this fix is to make the row take 75% of the screen width to give yourself some room

I'm saying if you use a percentage  you're just going to have this 25% gap, instead of a correctly sized gap for the [empty] tag. So, I'm advocating for a fix which just leaves that amount of space for the empty tag if the tag exists

brayoh:
Let me make the necessary changes

brayoh:
@chrisnojima I've made the changes, please have a look and let me know if it's okay for me to go ahead and merge the PR

PR #13754, opened 2018-09-14
Add hover states for shh and lock icons.

adamjspooner:


shared/chat/inbox/row/big-team-channel/index.js
```diff
@@ -41,6 +49,8 @@ class BigTeamChannel extends PureComponent<Props> {
               styles.channelBackground,
               this.props.isSelected && styles.selectedChannelBackground,
             ])}
+            onMouseLeave={() => this.setState({isHovered: false})}
```
chrisnojima: should just make a this._onMouseLeave

shared/chat/inbox/row/big-team-channel/index.js
```diff
@@ -25,13 +25,21 @@ type Props = {
...
+
 const HoverBox = isMobile
   ? Box
   : glamorous(Box)({
       ':hover': {backgroundColor: globalColors.blueGrey2},
     })
 
-class BigTeamChannel extends PureComponent<Props> {
+class BigTeamChannel extends PureComponent<Props, State> {
+  state = {
```
chrisnojima: if we're doing this we likely dont need Hoverbox anymore?

adamjspooner: Using `state` for the hover state is still buggy. I'd rather continue using `glamorous` until we/I can work out a better solution. This is just to get the icon hovers working. I'm open to suggestions for making it better.

shared/chat/inbox/row/filter-big-team-channel/index.js
```diff
@@ -19,17 +21,41 @@ type Props = {
...
 }
 
-class FilterBigTeamChannel extends PureComponent<Props> {
+type State = {
+  isHovered: boolean,
+}
+
+const HoverBox = isMobile
+  ? Box
+  : glamorous(Box)({
```
chrisnojima: note we don't need this anymore with this pr (https://github.com/keybase/client/pull/13870). There are generated classNames for hover so you can just add `className='.hover_color_blueGrey2'`

adamjspooner:
@chrisnojima, that class name only changes text color. Should we add the same for background color?

adamjspooner:

adamjspooner:
Totally agreed. I added an issue for our next sprint.

PR #13757, opened 2018-09-14
use react-window for inbox

chrisnojima:
@keybase/react-hackers this switches us to react-window for inbox (from react-list). I think ultimately we want to move to react-window.

I did go down the route of re-adding will-change: transform to the react-list but that also caused issues before so i went down the react-window route

There are serious inbox perf issues w/ sending data over the ipc channel (@jzila  we should discuss)

If you review this please test this out in a bunch of different scenarios


mmaxim:
Is this to address that error I was talking about from last night?

chrisnojima:
initially i went in to investigate that, i couldn't repro it but i did notice some severe degradations in performance. So this is to address that but it could also affect that log

buoyad:
@chrisnojima something was wrong with my packager so I wasn't running the new code when reviewing. The floating divider isn't updating on scroll, you have to click on the inbox to get it to update

buoyad:
Actually it's just taking a couple seconds to settle after scrolling. Row highlighting and the pointer cursor go away after scrolling, then everything updates a second or two later

PR #13765, opened 2018-09-14
don't load wallets at login

zanderz:
is it as simple as this?


chrisnojima:
@buoyad ^

buoyad:
@keybase/react-hackers this is ready!

shared/actions/wallets.js
```diff
@@ -14,13 +14,14 @@ import {getPath} from '../route-tree'
...
 import {getEngine} from '../engine'
+import {anyWaiting} from '../constants/waiting'
 
 const buildPayment = (state: TypedState, action: any) =>
   RPCTypes.localBuildPaymentLocalRpcPromise({
     amount: state.wallets.buildingPayment.amount,
     // FIXME: Assumes XLM.
-    fromPrimaryAccount: !state.wallets.buildingPayment.from,
-    from: state.wallets.buildingPayment.from,
+    fromPrimaryAccount: state.wallets.buildingPayment.from === Types.noAccountID,
```
zanderz: this was my change, based on your suggestion, but should it now be `state.wallets.builtPayment`?

buoyad: `builtPayment` will always be empty when the send form is first opened, though. the idea is to pass the input from the form to `BuildPaymentLocal` and then use the resolved fields in `SendPaymentLocal`. In this case in `BuildPaymentLocal` the form says `fromPrimaryAccount: true`, we get the address which we put into `builtPayment.from`, which we then send off to `SendPaymentLocal`

zanderz:
it won't let me click the review button 'cause I'm the original author, but - lgtm

PR #13777, opened 2018-09-17
Update deps: fix teams add contacts

chrisnojima:
- [x] test desktop
- [x] prod desktop build
- [x] test ios
- [x] test android


shared/teams/invite-by-email/container.native.js
```diff
@@ -217,6 +217,7 @@ export default compose(
     // If contacts or _pendingInvites changes, recalculate the props on the contact rows.
     withPropsOnChange(['contacts', 'loadingInvites', '_pendingInvites'], props => {
       // Create static contact row props here
+      const knownIDs = new Set()
```
chrisnojima: just don't allow dupes. i had some contacts where i had their email in multiple times and the old code doens't account for this. in general this code is overly complicated i think.
cc: @buoyad

buoyad: sounds good. and agreed, maybe we could ticket a refactor? especially since it could be reused for upcoming conversations with phone numbers / emails.

chrisnojima:

PR #13779, opened 2018-09-17
Abstract PvlSource to MerkleStore

joshblum:
Abstracts `PvlSource` into a generic `MerkleStore` where the client can fetch some data from the server and verify the hash is what we expect in the merkle root. This paves the way for parameterized proofs since we store the hash known configurations in the merkle root. The original code is unchanged besides naming except for:

- bug where we were not respecting the `dbVersion` variable correctly
- removed async db write (we don't do this elsewhere, i don't think it's a performance concern)
- dbKey is abstracted to be different for each store based on the `tag` the store is initialized with


go/merklestore/store.go
```diff
@@ -0,0 +1,346 @@
...
+	// This is the hash we are being instructed to use.
+	hash := keybase1.MerkleStoreKitHash(s.getRootHash(*root))
+
+	if hash == "" {
+		return "", "", NewMerkleStoreError("merkle root has empty %s hash: %v", s.tag, seqnoWrap(root.Seqno()))
+	}
+
+	// Use in-memory cache if it matches
+	if fromMem := s.memGet(hash); fromMem != nil {
+		m.CDebugf("MerkleStore: mem cache hit, using hash: %s", hash)
```
mlsteele: Assertion parsing will probably be a lot more happening than proof checking. Might want to punt this to VDL. Also add s.tag.

go/merklestore/store.go
```diff
@@ -0,0 +1,346 @@
...
+	s.dbSet(m.BackgroundWithLogTags(), hash, kitJSON)
+
+	m.CDebugf("MerkleStore: using hash: %s", hash)
+	return kitJSON, hash, nil
+}
+
+type merkleStoreServerRes struct {
+	Status  libkb.AppStatus         `json:"status"`
+	KitJSON keybase1.MerkleStoreKit `json:"kit_json"`
+}
```
mlsteele: I didn't realize the applications were so similar that this could be shared. Nice.

go/merklestore/store.go
```diff
@@ -0,0 +1,346 @@
...
+	// arbitrarily t_alice.  If t_alice is removed, this path will break.
+	uid := s.G().GetMyUID()
+	if len(uid) == 0 {
+		// Use t_alice's uid.
+		uid = libkb.TAliceUID
+	}
+	q.Add("uid", libkb.UIDArg(uid))
+	_, err := s.G().MerkleClient.LookupUser(m, q, nil)
+	return err
+}
```
mlsteele: yeck. Maybe there's a new function on MerkleClient to accomplish the same thing? @maxtaco

go/merklestore/init.go
```diff
@@ -0,0 +1,21 @@
...
+	tag := "pvl"
+	endpoint := "merkle/pvl"
+	getRootHash := func(root libkb.MerkleRoot) string {
+		return root.PvlHash()
+	}
+	kitFilename := g.Env.GetPvlKitFilename()
+	s := NewMerkleStore(g, tag, endpoint, kitFilename, supportedVersion, getRootHash)
+	g.SetPvlSource(s)
+	return s
+}
```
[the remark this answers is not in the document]
joshblum: it's already changed in the next PR! came to the same realization heh https://github.com/keybase/client/pull/13791

go/merklestore/store.go
joshblum: added `s.tag` but don't follow what we should be punting? are you saying we should parse the json here so we cache the parsed value?

go/merklestore/store.go
joshblum: yeah this is unfortunate. if there is something existing that is better we can swap it in or open a ticket to do the same but won't block this from merging

joshblum:
going to apply the feedback changes to the next PR https://github.com/keybase/client/pull/13791

go/merklestore/store.go
mlsteele: I meant CDebugf -> VDL.CLogf(ctx, VLog0,

joshblum: gotcha will change

PR #13780, opened 2018-09-17
client support for force-repoll mode

maxtaco:
- when we get a gregor message of type team.force_repoll, then force a repoll on every team load since we no longer expect pushed cache invalidations
- a test in systests to make sure this bit is getting set properly

### DIFF
--- a/go/systests/teams_test.go
+++ b/go/systests/teams_test.go
@@ -1538,3 +1538,38 @@ func TestTeamBustResolverCacheOnSubteamRename(t *testing.T) {
 		require.Error(t, err)
 	}
 }
+
+func TestForceRepollState(t *testing.T) {
+	tt := newTeamTester(t)
+	defer tt.cleanup()
+
+	tt.addUser("onr")
+	tt.addUser("wtr")
+
+	_, err := tt.users[0].tc.G.API.Post(libkb.APIArg{
+		Endpoint:    "test/big_state_cutoff",
+		SessionType: libkb.APISessionTypeREQUIRED,
+		Args: libkb.HTTPArgs{
+			"cutoff": libkb.I{Val: 1},
+		},
+	})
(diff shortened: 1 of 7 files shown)


go/systests/teams_test.go
```diff
@@ -1538,3 +1538,38 @@ func TestTeamBustResolverCacheOnSubteamRename(t *testing.T) {
...
+	defer tt.cleanup()
+
+	tt.addUser("onr")
+	tt.addUser("wtr")
+
+	_, err := tt.users[0].tc.G.API.Post(libkb.APIArg{
+		Endpoint:    "test/big_state_cutoff",
+		SessionType: libkb.APISessionTypeREQUIRED,
+		Args: libkb.HTTPArgs{
+			"cutoff": libkb.I{Val: 1},
```
mlsteele: this would make concurrent tests real weird. Could pass team ID?

maxtaco: only takes effect for the user posting!  so it should be fine for concurrent (if we don't use the user in other tests, which we don't!)

go/teams/loader.go
```diff
@@ -242,6 +250,10 @@ func (l *TeamLoader) load1(ctx context.Context, me keybase1.UserVersion, lArg ke
 		mungedForceRepoll = true
 		mungedWantMembers = nil
 	}
+	if !mungedForceRepoll && l.InForceRepollMode(ctx) {
+		l.G().Log.CDebugf(ctx, "Must force-repoll in force-repoll mode")
+		mungedForceRepoll = true
```
mlsteele: is this used? Anyway throw this in `load2DecideRepoll` instead of munge

maxtaco: cool thanks!

maxtaco: done

PR #13786, opened 2018-09-17
test case that exercises CORE-8859

maxtaco:
- fails without the (server-side) fix
- succeeds with it


maxtaco:
```// See CORE-8859, there was a bug that showed up when we loaded the subteam first and then
// the parent, since when we loaded the subteam, we were in "subteam reader" mode, and
// due to a previous server bug, didn't get boxes and prevs back when in subteam reader mode.
// Then, when we tried to access a box in the parent, we would fail. Test that it works.```

PR #13790, opened 2018-09-17
one more level of try

chrisnojima:
these inner calls aren't wrapped in try so they could fail


chrisnojima:
@keybase/react-hackers

PR #13794, opened 2018-09-17
this isnt allowed on ios anymore

chrisnojima:
@keybase/react-hackers fix crash on ios


chrisnojima:
Yes. Same as android now

PR #13797, opened 2018-09-18
black bar when clicking on header of currency list

zanderz:


zanderz:
Also: dropdown doesn't close when you click elsewhere, shouldn't it?

PR #13798, opened 2018-09-18
Don't send participants for teams

jzila:
@chrisnojima I think with the memoized object it should cache properly so it hits the shallow equal appropriately.


chrisnojima:
i'm merging this

PR #13802, opened 2018-09-18
Don't assert name during decryption

joshblum:
cc @maxtaco


maxtaco:
sorry about the bug here.  what do you want to do about feature flagging FTL on in devel?

logtm

joshblum:
enabled with https://github.com/keybase/keybase/pull/2879

PR #13803, opened 2018-09-18
protect virtualized list, update yarn lock

chrisnojima:
@keybase/react-hackers this stops some yellowboxes
also yarn.lock is updated to a newer babel core


shared/common-adapters/virtualized-list.native.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+import * as React from 'react'
+import {VirtualizedList} from 'react-native'
+
+class NativeVirtualizedList extends React.Component<React.ElementProps<VirtualizedList>> {
+  _mounted = true
```
buoyad: Is it safe to make this `true` before `componentDidMount`? Seems like it could replace one cause of yellow boxes with another.

chrisnojima: in practice yes since the callbacks never happen before componentDidMount, essentially if the component is alive its ok, if its on its way to being dead then its not, so we can safely init it here. Maybe it should be called _safeToCallSetState or something but we do this everywhere

chrisnojima: we do this pretty inconsistently: https://github.com/keybase/client/search?p=1&q=_mounted&unscoped_q=_mounted

chrisnojima: let me just change them all

chrisnojima:
@buoyad updated

PR #13805, opened 2018-09-18
Build with production tags in CI

joshblum:
will help prevent accidentally skipping a build specific definition.


joshblum:
sgtm!

PR #13814, opened 2018-09-18
Fix retry on prev pointers

joshblum:
Relaxes the prevpointer condition and beefs up the retry loop. in particular:

- if we get to the last page of the conversation we will nuke our local cache and retry (at most one times)
- if we never find any prevs, we allow the message to be sent so the conversation doesn't get into a crash loop.

### DIFF
--- a/go/chat/convsource.go
+++ b/go/chat/convsource.go
@@ -286,6 +286,10 @@ func (s *RemoteConversationSource) Clear(ctx context.Context, convID chat1.Conve
 	return nil
 }
 
+func (s *RemoteConversationSource) ClearLocalCache(ctx context.Context, convID chat1.ConversationID, uid gregor1.UID) error {
+	return nil
+}
+
 func (s *RemoteConversationSource) GetMessages(ctx context.Context, conv types.UnboxConversationInfo,
 	uid gregor1.UID, msgIDs []chat1.MessageID, threadReason *chat1.GetThreadReason) ([]chat1.MessageUnboxed, error) {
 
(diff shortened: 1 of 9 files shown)


go/chat/convsource.go
```diff
@@ -932,6 +936,10 @@ func (s *HybridConversationSource) Clear(ctx context.Context, convID chat1.Conve
 	return s.storage.ClearAll(ctx, convID, uid)
 }
 
+func (s *HybridConversationSource) ClearLocalCache(ctx context.Context, convID chat1.ConversationID, uid gregor1.UID) error {
```
mmaxim: What is the difference between this and `ClearAll`?

joshblum: seemed more efficient to call nuke rather than going through all of the messages of a conversation, maybe `ClearAll` should just call `MaybeNuke` instead of `clearUpthrough`

go/chat/storage/storage.go
```diff
@@ -313,7 +313,13 @@ func (h *HoleyResultCollector) Holes() int {
 	return h.holes
 }
 
-func (s *Storage) MaybeNuke(ctx context.Context, force bool, err Error, convID chat1.ConversationID,
+func (s *Storage) MaybeNukeLocked(ctx context.Context, force bool, err Error, convID chat1.ConversationID, uid gregor1.UID) Error {
```
mmaxim: I don't like this name, usually you put `Locked` on the end if you already have the lock. I would make the lower case name have this.

go/chat/convsource.go
mmaxim: I think that would be better, I'd rather not have two functions with effectively the same name in here.

mmaxim: Although `ClearAll` has the advantage it actually removes the data right there. Maybe just call that, I don't think there is a real efficiency problem with calling it.

go/chat/storage/storage.go
joshblum: sgtm!

go/chat/convsource.go
joshblum: oops just saw this, what do you mean by `removes the data right there`?

mmaxim: All nuke does is take out the "block index", whereas `ClearAll` kills the messages too.

joshblum: ah gotcha

PR #13816, opened 2018-09-18
Optimize ipc layer

chrisnojima:
This PR adds are serialization layer to the ipc traffic. So now we use the proxy component to figure out at a high level which props might change. Then we have another map which figures out what to send over the wire. this lets us keep things immutable for longer and really only send a true diff across the wire.


shared/actions/pinentry.js
```diff
@@ -21,7 +21,7 @@ function setupEngineListeners() {
       })
   })
 
-  engine().setIncomingCallMap({
+  engine().setCustomResponseIncomingCallMap({
```
chrisnojima: this fixes pinentry not working on master, we actually handle response here

shared/chat/inbox/container/remote.js
```diff
@@ -1,22 +1,13 @@
 // @flow
-import shallowEqual from 'shallowequal'
-import * as Container from '../../../util/container'
 import * as Constants from '../../../constants/chat2'
 import * as Styles from '../../../styles'
 import * as SmallTeam from '../row/small-team'
 import * as ChatTypes from '../../../constants/types/chat2'
+import type {TypedState} from '../../../constants/reducer'
+import Flags from '../../../util/feature-flags'
+import memoize from 'memoize-one'
```
chrisnojima: we use memoize one instead of reselect

shared/chat/inbox/container/remote.js
```diff
@@ -1,22 +1,13 @@
...
-const getMetaMap = (state: Container.TypedState) => [state.chat2.metaMap, state]
-
-const createShallowEqualSelector = Container.createSelectorCreator(Container.defaultMemoize, shallowEqual)
-
-// Get conversations
-const getMetas = Container.createSelector([getMetaMap], ([metaMap, state]) => [
-  metaMap.filter((meta, id) => Constants.isValidConversationIDKey(id)),
-  state,
-])
+export const maxShownConversations = Flags.fileWidgetEnabled ? 3 : 7
```
chrisnojima: we only show 3 if files is on so match that

shared/constants/config.js
```diff
@@ -21,7 +21,7 @@ export const teamFolder = (team: string) => `${defaultKBFSPath}${defaultTeamPref
 export const makeState: I.RecordFactory<Types._State> = I.Record({
   appFocused: true,
   appFocusedCount: 0,
-  avatars: {}, // Can't be an I.Map since it's used by remotes
+  avatars: I.Map(),
```
chrisnojima: this is immutable now! 🎆

shared/desktop/remote/store.desktop.js
```diff
@@ -42,45 +42,28 @@ class RemoteStore {
     sendToMainWindow('remoteWindowWantsProps', props.windowComponent, props.windowParam)
   }
 
-  // Some shared inner components needs immutable structures (Avatar), we can likely fix this longer term but for now lets just
-  // map these types back to immutable so the components aren't aware we're doing this over-the-wire store stuff which requires
-  // things to not be immutable. We have very little stuff in remote windows so i think this is simpler than some larger overhaul
-  // to enable embedded connected components
-  _makeImmutable = (props: any) => {
```
chrisnojima: this kinda weird thing is now part of the avatars, where it belongs

shared/desktop/remote/sync-avatar-props.desktop.js
```diff
@@ -19,6 +19,48 @@ type Props = {
   windowParam: string,
 }
 
+export const serialize = {
+  avatars: (v: any, o: any) => {
+    if (!v) return undefined
+    const toSend = v.filter((sizes, name) => {
+      return !o || sizes !== o.get(name)
+    })
+    return toSend.isEmpty() ? undefined : toSend.toJS()
```
chrisnojima: we only send changed avatars

shared/desktop/remote/sync-props.desktop.js
```diff
@@ -5,69 +5,117 @@
...
+        // Do a shallow equal
+        if (debugSerializer) {
+          console.log('[Serializer]: ---------------------------------')
         }
-        return map
-      }, {})
-      this._lastProps = childProps
-      return toSend
-    }
+        Object.keys(newProps).forEach(k => {
```
chrisnojima: we do a shallow equal and only send different sub things through the serializer layer

shared/chat/inbox/container/remote.js
```diff
@@ -26,47 +17,56 @@ export type RemoteConvMeta = $Diff<
...
+const metaMapToFirstValues = memoize(metaMap =>
+  metaMap
+    .partialSort(maxShownConversations, (a, b) => b.timestamp - a.timestamp)
+    .valueSeq()
+    .toArray()
 )
 
-// Just to cache the sorted values
-const GetNewestConvMetas = createShallowEqualSelector([getSortedConvMetas], map => map)
+// A hack to store the state so we can convert at the last possible minute. This is a lot simpler than plumbing this all the way through
```
jzila: lol I was trying to avoid doing this.

shared/chat/inbox/container/remote.js
```diff
@@ -26,47 +17,56 @@ export type RemoteConvMeta = $Diff<
...
-        youAreReset: m.membershipType === 'youAreReset',
-        youNeedToRekey,
-      }
-    }).toArray()
+// To cache the list
+const valuesCached = memoize((...vals) => vals.map(v => v))
+
+const metaMapToFirstValues = memoize(metaMap =>
+  metaMap
+    .partialSort(maxShownConversations, (a, b) => b.timestamp - a.timestamp)
```
jzila: Why did you remove the filter? We're going to need to add to it anyway, so please leave it in.

shared/chat/inbox/container/remote.js
chrisnojima: this is just a short term thing / very isolated. i think ideally we'd keep this in the store so we don't have to recompute it needlessly

jzila: The memoization should address having to recompute it.

chrisnojima:
@keybase/react-hackers bump

shared/chat/inbox/container/remote.js
jzila: Also where is `partialSort` coming from? I don't see it in Immutable's Map, just `immutable-sorted`, which we don't seem to be using.

jzila: Nvmind, we're using `immutable-sorted` in our `package.json` as the `immutable` package.

shared/chat/inbox/container/remote.js
```diff
@@ -26,47 +17,57 @@ export type RemoteConvMeta = $Diff<
...
+    .filter((_, id) => Constants.isValidConversationIDKey(id))
```
buoyad: I think this should go before `partialSort`

shared/chat/inbox/container/remote.js
```diff
@@ -26,47 +17,57 @@ export type RemoteConvMeta = $Diff<
...
+    isInWidget: true,
+    isMuted: m.isMuted,
+    isSelected: false,
+    // excluding onSelectConversation
+    participantNeedToRekey,
+    participants: m.teamname
+      ? []
+      : Constants.getRowParticipants(m, _username)
+          .toArray()
+          .slice(0, 2),
```
buoyad: Why the `slice` here?

shared/desktop/remote/sync-props.desktop.js
```diff
@@ -5,69 +5,117 @@
 import * as React from 'react'
 import * as SafeElectron from '../../util/safe-electron.desktop'
 
+// set this to true to see details of the serialization process
+const debugSerializer = __DEV__ && false
```
buoyad: Could we put this in `local-debug`

shared/local-debug.desktop.js
```diff
@@ -3,7 +3,7 @@ import {jsonDebugFileName} from './constants/platform.desktop'
 import {noop} from 'lodash-es'
 
 // Set this to true if you want to turn off most console logging so you can profile easier
-const PERF = false
+let PERF = false
```
buoyad: It'd be nice to add this inside `config` so the `local-debug` controls are more consistent.

shared/menubar/remote-container.desktop.js
```diff
@@ -58,10 +58,13 @@ const mapDispatchToProps = dispatch => ({
...
   },
 })
-const mergeProps = (stateProps, dispatchProps, ownProps) => ({
-  ...stateProps,
-  ...dispatchProps,
-  showUser: () => dispatchProps._showUser(stateProps.username),
-  ...ownProps,
-})
+
+const mergeProps = (stateProps, dispatchProps, ownProps) => {
```
buoyad: ?

shared/fs/remote-container.js
```diff
@@ -2,17 +2,17 @@
...
   timestamp: number,
+  tlf: Types.Path,
   updates: Array<Types.Path>,
+  writer: string,
 }
 
-const GetRowsFromTlfUpdates = (tlfUpdates: Types.UserTlfUpdates): Array<RemoteTlfUpdates> => tlfUpdates.toArray().map(t => ({
-  tlf: t.path,
-  writer: t.writer,
+const GetRowsFromTlfUpdate = (t: Types.TlfUpdate): RemoteTlfUpdates => ({
```
buoyad: I notice this is still sending when nothing has changed. Should we add a memoization layer?

shared/chat/inbox/container/remote.js
chrisnojima: i did this on purpose to not go through the entire list twice

shared/chat/inbox/container/remote.js
chrisnojima: we only need to show 2 people (its for the avatars only)

shared/desktop/remote/sync-props.desktop.js
chrisnojima: i doubt anyone will ever use this so its not a general thing you can control externally. we do this in a bunch of places for not-often-used settings

shared/fs/remote-container.js
chrisnojima: really the serializer needs to be updated. kbfs team was working on this while i was doing stuff so i didn't bother

shared/menubar/remote-container.desktop.js
chrisnojima: confused about what the question is

shared/chat/inbox/container/remote.js
jzila: But the filter can result in removal of items, which will cause there to be fewer than `maxShownConversations`.

chrisnojima: in practice this never happens as the timestamps of the non valid conversations are zero so they won't ever be included if there are real convos. if there are real convos they'll show up first

jzila: That's true right now, but it won't be true as we add more filters.

shared/fs/remote-container.js
jzila: @buoyad it shouldn't be...that function is only called in `mergeProps`.

shared/chat/inbox/container/remote.js
chrisnojima: what other filters would we add? the point of this exercise is to speed this thing up

jzila: https://keybase.atlassian.net/browse/KBFS-3331

jzila: I guess you could add the filter to the `partialSort` lambda function.

chrisnojima: if we need to filter beforehand we can but we don't in this code so it doesn't need to move. i prefer to not add complexity for some future case

shared/menubar/remote-container.desktop.js
buoyad: Why remove the shorthand?

shared/chat/inbox/container/remote.js
buoyad: It also truncates the list of participants for me

chrisnojima: oh i guess its for usernames too, k

shared/fs/remote-container.js
buoyad: @jzila `mergeProps`? It's in `serializer` [here](https://github.com/keybase/client/blob/b98e9e4760d33dc0fb8793242ae2ebc4bd5d6eab/shared/menubar/remote-serializer.desktop.js#L32). When I have `debugSerializer` on I see lists for key `fileRows` output to console

PR #13817, opened 2018-09-18
remove old KBFS code in GUI

songgao:
`constant/types/folders.js` is still there as it's used in a couple places in code. Some of them don't seem to be in use any more, but I'm not confident enough to just remove them myself. Will comment inline with questions.


shared/profile/index.desktop.js
```diff
@@ -249,6 +249,8 @@ class ProfileRender extends PureComponent<Props, State> {
       }
     }
 
+    // TODO/songgao: is it intended that this prop is still here? The prop is
+    // not provided in the container at all.
```
songgao: I think this is the last major use of `constants/types/folders.js`. It seems to intend to show a list of folders by the user, but is not visible in the app now. Do we still want it shown?

songgao: Ugh I'm realizing I accidentally deleted a previous comment: I think this is the last major use of `constants/types/folders.js`. It seems to intend to show a list of user's folders in the profile page, but isn't actually effective since the container doesn't provide the `tlfs` prop at all, thus the app currently doesn't show it either. Do we still want it shown or is it safe to remove this?

songgao:
Thanks Steve! Before merging, I'll wait for a day or so hoping to resolve my question about the Profile use of `constants/types/folders.js`

chrisnojima:
@malgorithms

malgorithms:
If I'm understanding this issue correctly from a quick discussion with @chrisnojima it sounds like we should have a new ticket to get folders of interest back on profiles, assuming that code needs to be re-written. it really would be nice for KBFS, if, when you look at someone's profile, you see folders of interest.

this'll be especially valuable when we get to the new back button programming and make profile visiting easier/faster inside the app.

songgao:
Sounds good. Thanks folks! I'll create a ticket for that and merge this as-is (i.e. including the unused `tlfs` code) in case when we work on that ticket we can re-use some of the react component code.

PR #13818, opened 2018-09-18
fix opening non-keybase path when fuse is disabled

songgao:
This fixes opening chat attachments and downloaded stuff from KBFS, when FUSE integration is not enabled.

Depends on https://github.com/keybase/client/pull/13817


songgao:
Bump! This will likely all change when we do KBFS-3432, but it's a simple change so might as well get in.

PR #13820, opened 2018-09-19
Fix info panel notifications

buoyad:
r? @keybase/react-hackers


shared/chat/conversation/info-panel/notifications/container.js
```diff
@@ -105,38 +105,39 @@ export default compose(
     }
   ),
   lifecycle({
-    componentWillReceiveProps(nextProps) {
-      // Same as the store?
+    componentDidUpdate(prevProps) {
```
buoyad: `nextProps` -> `this.props`
`this.props` -> `prevProps`

chrisnojima:
I think we should do that separately   It's not really cookie cutter and we shouldn't rush it

PR #13821, opened 2018-09-19
Removed debug logging of file descriptor

tjps:
The `Using logFd = ` message occurs frequently in syslog.  It was added in 0951eb8 but appears to be a debugging remnant.  

Posting this PR to see if there are any objections to removing it.


maxtaco:
cc @keybase/react-hackers

PR #13829, opened 2018-09-19
[KBFS-3331] Rows with uninteresting activity jumping to top of chat widget

brayoh:
#### Description of task:
The calculation of a row's position in the chat widget should be the max of:

   - when you last sent into that chat
   - when there was last activity in the chat you should've been notified of

for the second point, if you're in a big general channel, you don't want all activity, but if someone at-mentions you, odds are you'll have notifications on for that, so it should increase its score.

this means the things you want to get to will most likely be at the top - -chats with things you want to read and/or you just wrote in and want to get back to.

### DIFF
--- a/shared/chat/inbox/container/remote.js
+++ b/shared/chat/inbox/container/remote.js
@@ -20,7 +20,7 @@ export type RemoteConvMeta = $Diff<
 // To cache the list
 const valuesCached = memoize((...vals) => vals.map(v => v))
 
-const metaMapToFirstValues = memoize(metaMap =>
+const metaMapToFirstValues = memoize((metaMap, _lastState) =>
   metaMap
     .filter((meta, id) => {
       if (Constants.isValidConversationIDKey(id)) {
@@ -58,7 +58,7 @@ const metaMapToFirstValues = memoize(metaMap =>
 let _lastState: TypedState
 export const conversationsToSend = (state: TypedState) => {
   _lastState = state
-  return valuesCached(...metaMapToFirstValues(state.chat2.metaMap))
+  return valuesCached(...metaMapToFirstValues(state.chat2.metaMap, _lastState))
 }
 
 export const serialize = (m: ChatTypes.ConversationMeta): RemoteConvMeta => {
--- a/shared/chat/inbox/container/remote.js
+++ b/shared/chat/inbox/container/remote.js
@@ -50,7 +50,7 @@ const metaMapToFirstValues = memoize((metaMap, _lastState) =>
         return false
       }
     })
-    .partialSort(maxShownConversations, (a, b) => b.timestamp - a.timestamp)
+    .partialSort(maxShownConversations * 10, (a, b) => b.timestamp - a.timestamp)
     .valueSeq()
     .toArray()
 )
--- a/shared/chat/inbox/container/remote.js
+++ b/shared/chat/inbox/container/remote.js
@@ -23,7 +23,30 @@ const valuesCached = memoize((...vals) => vals.map(v => v))
 const metaMapToFirstValues = memoize(metaMap =>
   metaMap
     .partialSort(maxShownConversations, (a, b) => b.timestamp - a.timestamp)
-    .filter((_, id) => Constants.isValidConversationIDKey(id))
+    .filter((meta, id) => {
+      if (Constants.isValidConversationIDKey(id)) {
+        if (meta.teamType === 'adhoc') {
+          // DM's
+          return !meta.isMuted ? true : false
+        } else if (meta.teamType === 'big') {
+          // channels
+          if (!meta.notificationsGlobalIgnoreMentions) {
+            // @here and @mention not ignored
+            if (meta.notificationsDesktop === 'onAnyActivity') {
+              return true
+            } else if (meta.notificationsDesktop === 'onWhenAtMentioned') {
+              // check if user has been mentioned in the message
+              let snippet = meta.snippet.toLowerCase()
+              return snippet.indexOf(`@${_username}`) === 0 || snippet.indexOf('you') === 0
+            }
+            return false
+          }
+          return false
+        }
+        return false
+      }
+      return false
+    })
     .valueSeq()
     .toArray()
 )
--- a/shared/chat/inbox/container/remote.js
+++ b/shared/chat/inbox/container/remote.js
@@ -24,30 +24,31 @@ const metaMapToFirstValues = memoize((metaMap, _lastState) =>
   metaMap
     .filter((meta, id) => {
       if (Constants.isValidConversationIDKey(id)) {
-        if (meta.teamType === 'adhoc') {
-          // DM's
-          return !meta.isMuted
-        } else if (meta.teamType === 'big') {
-          // channels
-          if (!meta.notificationsGlobalIgnoreMentions) {
-            // @here and @mention not ignored
-            if (meta.notificationsDesktop === 'onAnyActivity') {
-              return true
-            } else if (meta.notificationsDesktop === 'onWhenAtMentioned') {
-              // check if user has been mentioned in the message
-              let snippet = meta.snippet.toLowerCase()
-              return (
-                (!!_lastState.config.username && snippet.indexOf(`@${_lastState.config.username}`) === 0) ||
-                snippet.indexOf('you') === 0
-              )
-            }
-            return false
-          }
-          return false
+        return false
+      }
+      if (meta.teamType === 'adhoc') {
+        // DM's
+        return !meta.isMuted
+      }
+      if (!meta.teamType === 'big') {
+        return false
+      }
+
+      // channels
+      if (!meta.notificationsGlobalIgnoreMentions) {
+        // @here and @mention not ignored
+        if (meta.notificationsDesktop === 'onAnyActivity') {
+          return true
+        } else if (meta.notificationsDesktop === 'onWhenAtMentioned') {
+          // check if user has been mentioned in the message
+          let snippet = meta.snippet.toLowerCase()
+          return (
+            (!!_lastState.config.username && snippet.indexOf(`@${_lastState.config.username}`) === 0) ||
+            snippet.indexOf('you') === 0
+          )
         }
         return false
       }
-      return false
     })
     .partialSort(maxShownConversations, (a, b) => b.timestamp - a.timestamp)
     .valueSeq()
--- a/shared/chat/inbox/container/remote.js
+++ b/shared/chat/inbox/container/remote.js
@@ -36,7 +36,10 @@ const metaMapToFirstValues = memoize(metaMap =>
             } else if (meta.notificationsDesktop === 'onWhenAtMentioned') {
               // check if user has been mentioned in the message
               let snippet = meta.snippet.toLowerCase()
-              return snippet.indexOf(`@${_username}`) === 0 || snippet.indexOf('you') === 0
+              return (
+                (!!_laststate.config.username && snippet.indexOf(`@${_laststate.config.username}`) === 0) ||
+                snippet.indexOf('you') === 0
+              )
             }
             return false
           }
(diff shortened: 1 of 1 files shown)


shared/chat/inbox/container/remote.js
```diff
@@ -13,10 +14,45 @@ const getMetaMap = (state: Container.TypedState) => [state.chat2.metaMap, state]
...
+  const prioritizedMetas = metaMap.sort((a, b) => {
+    if (a.teamType === 'adhoc' && b.teamType === 'adhoc') {
+      // chats first
+      if (
+        Constants.getHasUnread(state, b.conversationIDKey) ||
+        Constants.getHasUnread(state, a.conversationIDKey)
+      ) {
+        // sort unread message by time
+        return b.timestamp - a.timestamp
+      }
```
songgao: Could you explain the logic here? It seems even if we don't hit the condition, we'd end up at the [return after the conditions](https://github.com/keybase/client/pull/13829/files#diff-2e93add81efa99eb09c839d315c22ae9R51) below.

Perhaps you meant to check if only one of them has unread messages, in which case we'd put the unread at the beginning?

shared/chat/inbox/container/remote.js
```diff
@@ -13,10 +14,45 @@ const getMetaMap = (state: Container.TypedState) => [state.chat2.metaMap, state]
...
+        return b.timestamp - a.timestamp
+      } else if (
+        (a.snippet.toLowerCase().indexOf('you') === 0 &&
+          moment().diff(new Date(b.timestamp), 'minutes') < 30) ||
+        (b.snippet.toLowerCase().indexOf('you') === 0 && moment().diff(new Date(a.timestamp), 'minutes') < 30)
+      ) {
+        // last activity in the chat was less than 30 minutes ago
+        return b.timestamp - a.timestamp
+      }
+    }
```
songgao: Seems like we are missing an `else` here which means one of them is adhoc and the other is big team, where we'd want to put the adhoc one first.

shared/yarn.lock
```diff
@@ -13,9 +13,9 @@
   version "1.1.8"
   resolved "https://registry.yarnpkg.com/@ant-design/icons/-/icons-1.1.8.tgz#a95dbe1e32928500ab130fb35f77e2ca447ec0aa"
 
-"@babel/cli@7.1.0":
-  version "7.1.0"
-  resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.1.0.tgz#a9429fd63911711b0fa93ae50d73beee6c42aef8"
+"@babel/cli@^7.0.0":
```
jzila: You should use `yarn modules` instead of `yarn install`. It's an alias for `yarn install --pure-lockfile`. And please rebase.

shared/chat/inbox/container/remote.js
```diff
@@ -13,10 +14,45 @@ const getMetaMap = (state: Container.TypedState) => [state.chat2.metaMap, state]
...
+        // sort unread message by time
+        return b.timestamp - a.timestamp
+      }
+    } else if (a.teamType === 'big' || b.teamType === 'big') {
+      // channels
+      // check if user has been mentioned in the message
+      if (a.snippet.indexOf(`@${_username}`) === 0 || b.snippet.indexOf(`@${_username}`) === 0) {
+        return b.timestamp - a.timestamp
+      } else if (
+        (a.snippet.toLowerCase().indexOf('you') === 0 &&
```
songgao: Hmm I'd guess there would be a better way to check if the latest message is sent by `you`. Same for the @-mentions above, by looking at stuff from the store.

shared/chat/inbox/container/remote.js
```diff
@@ -13,10 +14,45 @@ const getMetaMap = (state: Container.TypedState) => [state.chat2.metaMap, state]
...
+          moment().diff(new Date(b.timestamp), 'minutes') < 30) ||
+        (b.snippet.toLowerCase().indexOf('you') === 0 && moment().diff(new Date(a.timestamp), 'minutes') < 30)
+      ) {
+        // last activity in the chat was less than 30 minutes ago
+        return b.timestamp - a.timestamp
+      }
+    }
+    // no recent activity no messages
+    return b.timestamp - a.timestamp
```
jzila: You always return `b.timestamp - a.timestamp`, so effectively all the code above does nothing.

shared/chat/inbox/container/remote.js
brayoh: It's working progress, I'm still trying to figure out how to get it right, I raised the PR so that I can get your assitance

shared/chat/inbox/container/remote.js
brayoh: Let me look through the store and see how it's looking

shared/yarn.lock
brayoh: Cool, let me do that, Thanks for the info

shared/chat/inbox/container/remote.js
brayoh: I'm still trying to see the best way to get it right, let me reimplement and see if I'll come up with something better

shared/chat/inbox/container/remote.js
```diff
@@ -9,6 +12,43 @@ import memoize from 'memoize-one'
...
+const createShallowEqualSelector = Container.createSelectorCreator(Container.defaultMemoize, shallowEqual)
+
+// Get conversations
+const getMetas = Container.createSelector([getMetaMap], ([metaMap, state]) => {
+  const _username = state.config.username
+
+  const prioritizedMetas = metaMap.filter(meta => {
+    if (meta.teamType === 'adhoc') {
+      // DM's
+      return Constants.getHasUnread(state, meta.conversationIDKey) ? true : false
```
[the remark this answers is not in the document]
brayoh: Cool, let me change that.

shared/chat/inbox/container/remote.js
```diff
@@ -17,14 +57,44 @@ export type RemoteConvMeta = $Diff<
...
+        snippet: m.snippet,
+        snippetDecoration: m.snippetDecoration,
+        subColor: styles.subColor,
+        teamname: m.teamname,
+        timestamp: Constants.timestampToString(m),
+        usernameColor: styles.usernameColor,
+        youAreReset: m.membershipType === 'youAreReset',
+        youNeedToRekey,
+      }
+    })
```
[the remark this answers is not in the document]
brayoh: Let me have a look

shared/chat/inbox/container/remote.js
```diff
@@ -17,14 +57,44 @@ export type RemoteConvMeta = $Diff<
   {onSelectConversation: () => void}
 >
 
-// To cache the list
-const valuesCached = memoize((...vals) => vals.map(v => v))
-
-const metaMapToFirstValues = memoize(metaMap =>
-  metaMap
-    .partialSort(maxShownConversations, (a, b) => b.timestamp - a.timestamp)
-    .filter((_, id) => Constants.isValidConversationIDKey(id))
```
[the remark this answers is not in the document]
brayoh: Cool

shared/chat/inbox/container/remote.js
```diff
@@ -9,6 +12,43 @@ import memoize from 'memoize-one'
...
+        if (meta.notificationsDesktop === 'onAnyActivity') {
+          return true
+        } else if (meta.notificationsDesktop === 'onWhenAtMentioned') {
+          // check if user has been mentioned in the message
+          return meta.snippet.indexOf(`@${_username}`) === 0
+            ? true
+            : meta.snippet.toLowerCase().indexOf('you') === 0 &&
+              moment().diff(new Date(meta.timestamp), 'minutes') < 30
+              ? true
+              : false
```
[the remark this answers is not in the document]
brayoh: Cool

shared/chat/inbox/container/remote.js
```diff
@@ -23,7 +23,30 @@ const valuesCached = memoize((...vals) => vals.map(v => v))
 const metaMapToFirstValues = memoize(metaMap =>
   metaMap
     .partialSort(maxShownConversations, (a, b) => b.timestamp - a.timestamp)
-    .filter((_, id) => Constants.isValidConversationIDKey(id))
+    .filter((meta, id) => {
```
jzila: The filter needs to be moved above the `partialSort`.

brayoh: Okay, let me do that.

shared/chat/inbox/container/remote.js
```diff
@@ -22,8 +22,31 @@ const valuesCached = memoize((...vals) => vals.map(v => v))
...
+        } else if (meta.teamType === 'big') {
+          // channels
+          if (!meta.notificationsGlobalIgnoreMentions) {
+            // @here and @mention not ignored
+            if (meta.notificationsDesktop === 'onAnyActivity') {
+              return true
+            } else if (meta.notificationsDesktop === 'onWhenAtMentioned') {
+              // check if user has been mentioned in the message
+              let snippet = meta.snippet.toLowerCase()
+              return snippet.indexOf(`@${_username}`) === 0 || snippet.indexOf('you') === 0
```
jzila: since `_username` is gone now, this needs to be
```javascript
return (!!_laststate.config.username && snippet.indexOf(`@${_laststate.config.username}`) === 0) || snippet.indexOf('you') === 0
```

jzila:
@chrisnojima mind taking a final look? We're relying on `_lastState`, just wanted to make sure that was kosher for now.

shared/chat/inbox/container/remote.js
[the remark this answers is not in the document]
chrisnojima: so i think we should avoid these types of deeply nested functions. you can collapse this with just returning early, something like
```
    .filter((meta, id) => {
      if (!Constants.isValidConversationIDKey(id)) {
        return false
      }
      if (meta.teamType === 'adhoc') {
        return !meta.isMuted
      } 
      if (meta.teamType !== 'big') {
        return false
      }
      // channels
      if (meta.notificationsGlobalIgnoreMentions) {
        return false
      }
      // @here and @mention not ignored
      if (meta.notificationsDesktop === 'onAnyActivity') {
        return true
      } 
      if (meta.notificationsDesktop === 'onWhenAtMentioned') {
          // check if user has been mentioned in the message
          let snippet = meta.snippet.toLowerCase()
          return (
            (!!_lastState.config.username && snippet.indexOf(`@${_lastState.config.username}`) === 0) ||
            snippet.indexOf('you') === 0
          )
      }
      return false
    })
```
(check that logic cause it could be wrong)

shared/chat/inbox/container/remote.js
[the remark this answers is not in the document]
chrisnojima: it might be faster to just do a partialSort of some multiple, then do the filter. we'l be doing a lot of extra work to filter on a thing we're just going to throw away. I'd assume we'd be safe with maxShownConversations*3 or something. for instance i have 4000 convos, so i'd rather just go through 15 or 100 or something

jzila:
I'm fine with doing a filter on a subset of ~`maxShownConversations*10`.

brayoh:
@chrisnojima Thanks for the feedback I'll implement it then update the PR

shared/chat/inbox/container/remote.js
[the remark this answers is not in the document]
chrisnojima: lets do this ('im fine with *100)

shared/chat/inbox/container/remote.js
```diff
@@ -20,10 +20,38 @@ export type RemoteConvMeta = $Diff<
...
+          let snippet = meta.snippet.toLowerCase()
+          return (
+            (!!_lastState.config.username && snippet.indexOf(`@${_lastState.config.username}`) === 0) ||
+            snippet.indexOf('you') === 0
+          )
+        }
+        return false
+      }
+    })
+    .slice(maxShownConversations)
```
jzila: This should be `.slice(0, maxShownConversations)`

PR #13836, opened 2018-09-19
Add RPC for set-mobile-only, CLI command

patrickxb:


protocol/avdl/stellar1/local.avdl
```diff
@@ -327,6 +327,9 @@ protocol local {
   KeybaseRequestID makeRequestLocal(int sessionID, string recipient, union { null, Asset } asset, 
     union { null, OutsideCurrencyCode } currency, string amount, string note);
 
+  void setAccountMobileOnlyLocal(int sessionID, AccountID accountID);
```
mlsteele: take bool to unset?

go/systests/stellar_client_test.go
```diff
@@ -469,3 +469,26 @@ func (s *stellarRetryClient) MarkAsReadLocal(ctx context.Context, arg stellar1.M
...
+func (s *stellarRetryClient) SetAccountMobileOnlyLocal(ctx context.Context, arg stellar1.SetAccountMobileOnlyLocalArg) error {
+	var err error
+	for i := 0; i < retryCount; i++ {
+		err = s.cli.SetAccountMobileOnlyLocal(ctx, arg)
+		if err == nil {
+			break
+		}
+	}
+	return err
+}
```
mlsteele: I've never seen stellarRetryClient before but why retry rpcs that hit app and stellard but not horizon?

patrickxb: well, clients don't know what is going to hit horizon...

protocol/avdl/stellar1/local.avdl
patrickxb: don't want to expose unset quite yet, as only mobile devices would be able to unset and it might be more complicated than an RPC call...

mlsteele: Not sure why it would be more complicated. It might be annoying to add later since the argument `enabled bool` would default to false. But if you want ok

mlsteele:
Hopefully `set-mobile-only` is hidden on prod clients

mlsteele:
You should probably be blocked from doing this if you have no mobile devices. And warn if your mobile devices have not been used in a while.

patrickxb:
The server blocks it in production

patrickxb:
Yes, there will be more restrictions going forward.

PR #13839, opened 2018-09-19
Show search suggestions on the Stellar 'To Keybase user' form

cjb:
@keybase/react-hackers


zanderz:
LGTM, I tried it. But wouldn't it have had the same effect to just set focus to the "To" field when the dialog opens?

cjb:
@zanderz Hm, that doesn't sound right to me -- I don't think there is any connection between focus and performing a searchSuggestions.  Might be wrong though!

PR #13832, opened 2018-09-19
Post generic social proof

joshblum:
builds on #13830, will rebase once that is in.

adds the ability to post a generic social proof, coming next will be parsing the link once it's in the sig chain. this will include changing `LastWriterWins=false` for generic social proofs (unlike other social proofs, this controls the `You already have a proof for xxx; overwrite?` prompt)

the current behavior is:

```
$ kb prove gubble.social josh
Running `go install github.com/keybase/client/go/keybase`... Done.
▶ WARNING Running in devel mode
Please click on the following link to post to GubbleSocial:

Please visit http://localhost:3000/gubble_social?kb_username=josh&sig_hash=9608a876d6d0d753ea392dd1b1636d42444ac158f07c39ff09d4791b6a7e4de60f to complete the proof.

Check GubbleSocial now? [Y/n] y
▶ ERROR RPC error 703: Error: No actor found for row: {"proof_id":"89044789dd2147153d69c510","sig_id":"9608a876d6d0d753ea392dd1b1636d42444ac158f07c39ff09d4791b6a7e4de60f","uid":"1b4f0e9851971998e732078544c96b19","proof_type":1002,"check_data_json":"{\"entropy\":\"dlMXFfTa/UZHLGwjYaprLZ6r\",\"name\":\"gubble.social\",\"username\":\"josh\"}","state":0,"status":0,"retry_count":0,"hard_fail_count":0,"proof_text_check":"","last_check":"2018-09-19T19:44:26.000Z","remote_id":"","api_url":null,"human_url":null,"eldest_seqno":9} (code 703)
```
in the tests we verify that the checking will fail until the hunter goes in on the server, we can post without issue if we don't check though

```
$ kb prove gubble.social joshua
Running `go install github.com/keybase/client/go/keybase`... Done.
▶ WARNING Running in devel mode
You already have a proof for josh@gubble.social; overwrite? [y/N] y
Please click on the following link to post to GubbleSocial:

Please visit http://localhost:3000/gubble_social?kb_username=joshua&sig_hash=32a9f4ffcb4631712f3310ceffab3add308ab3073154407862d282fa05eaee9a0f to complete the proof.

Check GubbleSocial now? [Y/n] n
▶ WARNING not exportable error: Proof wasn't available; we'll keep trying (libkb.ProofNotYetAvailableError)
▶ ERROR Proof wasn't available; we'll keep trying
```

these changes depend on server the server changes here: https://github.com/keybase/keybase/pull/2887

PR #13846, opened 2018-09-20
parse generic social sigchain links

joshblum:
builds on #13832, will rebase once that is in

id fails since the checker is not implemented yet

```
$ kb id test1 -s
Running `go install github.com/keybase/client/go/keybase`... 


Done.
▶ INFO Identifying test1
✔ "josh" on gubble.social failed: Not implemented (code=200)
✔ "joshua" on gubble.social failed: Not implemented (code=200)
✔ "test1" on rooter: http://localhost:3000/_/api/1.0/rooter/test1/3756f515f46b328c7c1c670ea5154303.json
```

### DIFF
--- a/go/externals/proof_service_generic_social.go
+++ b/go/externals/proof_service_generic_social.go
@@ -31,7 +31,8 @@ func NewGenericSocialProofChecker(p libkb.RemoteProofChainLink) (*GenericSocialP
 func (rc *GenericSocialProofChecker) GetTorError() libkb.ProofError { return nil }
 
 func (rc *GenericSocialProofChecker) CheckStatus(m libkb.MetaContext, h libkb.SigHint, _ libkb.ProofCheckerMode, pvlU keybase1.MerkleStoreEntry) libkb.ProofError {
-	// TODO will have to use the `check_url`/`check_path` config values to verify the proof.
+	// TODO CORE-8658 will have to use the `check_url`/`check_path` config
+	// values to verify the proof.
 	return libkb.NewProofError(keybase1.ProofStatus_BASE_HARD_ERROR, "Not implemented")
 }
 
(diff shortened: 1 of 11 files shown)


go/libkb/id_table.go
```diff
@@ -141,6 +141,7 @@ type RemoteProofChainLink interface {
 	ComputeTrackDiff(tl *TrackLookup) TrackDiff
 	GetProofType() keybase1.ProofType
 	ProofText() string
+	RequiresHint() bool
```
joshblum: not sure if this is the right move since we should be getting hints back from the server once the hunter is in..

go/externals/proof_service_generic_social.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (t GenericSocialProofServiceType) MakeProofChecker(l libkb.RemoteProofChainLink) libkb.ProofChecker {
+	return &GenericSocialProofChecker{l}
+}
+
+func (t GenericSocialProofServiceType) IsDevelOnly() bool { return false }
+
+// We can post multiple generic social proofs to a single service.
+func (t GenericSocialProofServiceType) LastWriterWins() bool { return false }
```
mlsteele: You can only have only live proof for each unique value of `GetTypeName`. Isn't LastWriterWins exactly what you want? Same as github?

go/libkb/id_table.go
```diff
@@ -242,32 +246,22 @@ func (s *SocialProofChainLink) insertIntoTable(tab *IdentityTable) {
 func (s *SocialProofChainLink) ToDisplayString() string {
 	return s.username + "@" + s.service
 }
-func (s *SocialProofChainLink) LastWriterWins() bool      { return true }
+func (s *SocialProofChainLink) LastWriterWins() bool {
+	return !s.isGeneric
```
mlsteele: same comment about LastWriterWins

PR #13851, opened 2018-09-20
Fix data race in request/sendPayment messages

buoyad:
Previously on thread load we

1. Receive `requestPayment` and `sendPayment` messages with no info attached -> store in `messageMap`
2. On notification, update in `messageMap` to add the info

I noticed that sometimes on first thread load some messages would never populate. It turns out that's because step 2 can happen before step 1. In that case, since `messageMap` doesn't have the message, we would drop the notification and the info would never be populated.

This separates the info in the message body from the one in the notification. In particular,

- Adds a map in the store `accountsInfoMap` to store info that comes in through notifications. Message records aren't populated by the notifications anymore, this map is
- Changes getters to check `accountsInfoMap` before looking in the message

As a bonus this also tweaks the message loading text.
r? @keybase/react-hackers

### DIFF
--- a/shared/constants/chat2/message.js
+++ b/shared/constants/chat2/message.js
@@ -29,40 +29,40 @@ export const getMessageID = (m: RPCChatTypes.UIMessage) => {
   }
 }
 
-export const getRequestMessageInfo = (state: TypedState, message: Types.MessageRequestPayment) => {
-  if (message.requestInfo) {
-    return message.requestInfo
-  }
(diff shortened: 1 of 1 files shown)


chrisnojima:
Can we discuss this. The original idea was to not disconnect these things and it seems like it's been re disconnected

buoyad:
@zanderz it would be possible, but the data could be blown away by later thread loads (likely the one we're racing with)

@chrisnojima yeah sure

PR #13860, opened 2018-09-20
Removed italic on Me things + tweaked chat message layout

cecileboucheron:
@keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/account-payment/index.js
+++ b/shared/chat/conversation/messages/account-payment/index.js
@@ -10,7 +10,7 @@ import {
   Text,
   type IconType,
 } from '../../../../common-adapters'
-import {globalColors, platformStyles, styleSheetCreate} from '../../../../styles'
+import {globalColors, globalMargins, platformStyles, styleSheetCreate} from '../../../../styles'
 
 export type Props = {|
   action: string,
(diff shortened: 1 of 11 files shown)


shared/chat/conversation/messages/wrapper/wrapper-author/index.js
```diff
@@ -281,9 +285,7 @@ const styles = styleSheetCreate({
     alignSelf: 'flex-start',
     marginBottom: 0,
   },
-  usernameYou: {
-    ...globalStyles.italic,
-  },
+  usernameYou: {},
```
songgao: Looks like this could just be removed?

shared/devices/row/index.js
```diff
@@ -58,9 +58,7 @@ const styles = Styles.styleSheetCreate({
       textDecorationLine: 'line-through',
       textDecorationStyle: 'solid',
     },
-    isElectron: {
-      fontStyle: 'italic',
-    },
+    isElectron: {},
```
songgao: Same here; I think we can get rid of this line.

shared/chat/conversation/messages/account-payment/index.js
```diff
@@ -34,7 +34,11 @@ const AccountPayment = (props: Props) => {
   ) : (
     <React.Fragment>
       <Box2 direction="horizontal" fullWidth={true} style={styles.headingContainer}>
-        <Box2 direction="horizontal" gap="xtiny" style={styles.headingContainer}>
+        <Box2
+          direction="horizontal"
+          gap="xtiny"
+          style={{...styles.headingContainer, marginBottom: globalMargins.xtiny}}
```
cecileboucheron: @chrisnojima Flow checks are complaining about this here, what am I doing wrong?

buoyad: Switch this object out for `collapseStyles([styles.headingContainer, {marginBottom: ...}])`, flow should stop complaining.

cecileboucheron:
@songgao Can you take another look please?

PR #13864, opened 2018-09-20
deprecate componentWillReceiveProps and componentWillMount

zanderz:
replacing componentWillReceiveProps() with:
1. getDerivedStateFromProps() for state changes, and
2. componentDidUpdate() for data changes and side effects

as per https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#migrating-from-legacy-lifecycles


zanderz:
I was wondering, would exercising this require duplicating an error sending a chat? I tried all the basic features I think (user mention, team mention, exploding header, message disposition) but maybe not all the states

shared/chat/conversation/messages/wrapper/chat-send.js
```diff
@@ -76,19 +69,40 @@ class SendIndicator extends React.Component<Props, State> {
...
+      this._setStatus('error')
     } else if (this.props.sent) {
       // previously sent message
       this._setVisible(false)
     }
   }
 
-  componentWillReceiveProps(nextProps: Props) {
-    if (nextProps.failed && !this.props.failed) {
+  static getDerivedStateFromProps(props, state) {
```
chrisnojima: is there a reason why this can't just use componentDidUpdate and not getDerivedStateFromProps?

shared/teams/team/settings-tab/retention/index.js
```diff
@@ -141,17 +145,32 @@ class RetentionPicker extends React.Component<Props, State> {
     this._init()
   }
 
-  componentWillReceiveProps(nextProps: Props) {
+  static getDerivedStateFromProps(props, state) {
```
chrisnojima: can we just use componentDidUpdate?

shared/chat/conversation/messages/wrapper/chat-send.js
zanderz: No, I thought that was the upgrade path recommended in the article.

shared/teams/team/settings-tab/retention/index.js
zanderz: Ok

chrisnojima: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

PR #13869, opened 2018-09-21
Show account public keys in more places

buoyad:
Creates a new common component `SmallAccountID` that initially displays `XXXXXXXX...XXXXXXXX` and expands and becomes selectable on click. 
smallaccountid
This adds `SmallAccountID` to the wallet header and sender/recipient in transaction details. Also adds the accountID in wallet settings in a `CopyText`. r? @keybase/react-hackers


buoyad:
@keybase/react-hackers bump

chrisnojima:
i guess my reply didn't go through. i believe the UX is that its supposed to expand on hover, not click
cc: @cecileboucheron

cecileboucheron:
This could work too I think. Especially because it'll be that exact behavior on mobile. 
@malgorithms @adamjspooner Thoughts?

buoyad:
@malgorithms and I talked a bit about this last week - a pro of this approach is we don't have to handle the case where the user click-and-drags outside the hover area and causes the ID to collapse

malgorithms:
yeah, I prefer click, rather than hover

shared/wallets/wallet/header/container.js
```diff
@@ -10,7 +10,7 @@ const mapStateToProps = (state: TypedState) => {
   return {
     accountID: selectedAccount.accountID,
     isDefaultWallet: selectedAccount.isDefault,
-    keybaseUser: state.config.username,
+    keybaseUser: state.config.username || '',
```
chrisnojima: username is actually always a string now, if its empty its ''

PR #13870, opened 2018-09-21
Perf 1-6 merged down

chrisnojima:
This leverages css more to speed up common-adapters/icon

To render the entire icon sheet in storybook:
Master: ~300ms
This PR: ~100ms!

This was part of my exploration of using emotion. Emotion is around 2x faster than glamorous so we should start to use it but going even lower level is faster still. We likely want to use emotion for smaller components but things like icon/text/box/etc should be as fast as they can be since they're everywhere

@keybase/react-hackers


adamjspooner:
LGTM!

cecileboucheron:
👍

PR #13872, opened 2018-09-21
2: text to css

chrisnojima:
This switches text to use css to increase perf


chrisnojima:
@keybase/react-hackers bump

chrisnojima:
cc: @cecileboucheron @adamjspooner to sanity check

cecileboucheron:
Shouldn't we try to use our globalColors constants instead of hard-coded rgb values?

chrisnojima:
this is a real css file so i can't pull in those vars. we can make the pipeline more complicated later i guess but it didn't seem worth it right now

cecileboucheron:
Ok. We'll have to remember to change this file if we change our colors.

chrisnojima:
sure, also the text colors change very very rarely

shared/common-adapters/text.css
```diff
@@ -0,0 +1,295 @@
+/* Used by text on desktop to speed up perf */
```
buoyad: I found a text type that's missing here (`BodyTinyBold`). Do we have a way to make sure this doesn't diverge with `text.meta.*`?

chrisnojima: longer term we could generate this , but i'm keeping this simple for now

buoyad:
There are some things that use text's `getStyle` (mostly input components afaict), so they're only getting partial styles on this branch

shared/common-adapters/text.css
chrisnojima: added it back in

chrisnojima:
@buoyad i added back getStyle but its only called by external components, not text. we should reconcile that better soon

MarcoPolo:
What about generating these at load like icons? Not super important but might avoid some of the issues mentioned

chrisnojima:
Yeah lets ticjet  this

shared/common-adapters/text.desktop.js
```diff
@@ -62,31 +62,70 @@ class Text extends Component<Props> {
...
+  const textDecoration = meta.isLink && backgroundMode !== 'Normal' ? {textDecoration: 'underline'} : null
+
+  return {
+    ...colorStyle,
+    ...lineClampStyle,
+    ...clickableStyle,
+    ...selectableStyle,
+    ...textDecoration,
+  }
+}
```
MarcoPolo: Couldn't these also be handled by css?

PR #13879, opened 2018-09-22
new kex2 secret type, as recommended by NCC Audit

maxtaco:
- 9 words, not 8
- use 2^10 scrypt cost always (on mobile or desktop)
- use UID as salt
- make a test and get it to work
- we'll still generate legacy phrases for a few releases, and then cut over


patrickxb:
Have you tried this on an older mobile device?

PR #13884, opened 2018-09-23
improve new folder naming hints when user creates multiple at same time

songgao:
so that user gets following as default hint:

New Folder
New Folder 1
New Folder 2

instead of:

New Folder
New Folder
New Folder

### DIFF
--- a/shared/reducers/fs.js
+++ b/shared/reducers/fs.js
@@ -173,9 +173,9 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
 
       const existingNewFolderNames = new Set(
         state.edits
-          .valueSeq()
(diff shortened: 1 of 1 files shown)


shared/reducers/fs.js
```diff
@@ -170,8 +170,20 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
         console.warn(`bad parentPath: ${parentPathItem.type}`)
         return state
       }
+
+      const existingNewFolderNames = new Set(
+        state.edits
+          .valueSeq()
```
chrisnojima: Maybe faster to filter first then map then use toSet to get the immutable set vs the js set

PR #13885, opened 2018-09-23
load fuse status on start so we don't show fuse banner 1st time in Files

songgao:
Without this change, we only load fuse status in componentDidMount, and
we'd end up showing the fuse banner for a split of a second the first
time user enters the Files tab.

### DIFF
--- a/shared/actions/fs/platform-specific.desktop.js
+++ b/shared/actions/fs/platform-specific.desktop.js
@@ -1,6 +1,7 @@
 // @flow
 import * as I from 'immutable'
 import * as FsGen from '../fs-gen'
+import * as ConfigGen from '../config-gen'
 import * as Saga from '../../util/saga'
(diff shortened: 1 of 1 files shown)


songgao:
Hmm I wouldn't say this is super cheap. The RPC handler does a stat call, then tries to find out about the kernel extension, then calls `/sbin/mount` to confirm the mount point. No real networking is involved though. Does it sound like something we'd like to avoid at startup?

chrisnojima:
you can merge this and we can evaluate it later

jzila:
I think it's cheap from the event loop's perspective. Most of the expense is in system calls.

songgao:
@chrisnojima ping! OK if I merge this for now? We can always optimize it later if needed I guess.

songgao:
damnit GH didn't refresh the content for me. Thanks Nojima and John; merging now!

PR #13886, opened 2018-09-23
fix upload banner height

songgao:
Otherwise, it's entirely constrained by the content height unless it's hitting the `maxHeight`, making the upload banner very short and without margin.

### DIFF
--- a/shared/fs/footer/upload.desktop.js
+++ b/shared/fs/footer/upload.desktop.js
@@ -85,6 +85,7 @@ const stylesBox = {
   alignItems: 'center',
   justifyContent: 'center',
   maxHeight: height,
+  height,
 }
 
 export default Upload


shared/fs/footer/upload.desktop.js
```diff
@@ -85,6 +85,7 @@ const stylesBox = {
   alignItems: 'center',
   justifyContent: 'center',
   maxHeight: height,
+  height,
```
jzila: doesn't this make `maxHeight` redundant?

songgao: The original change from Brian was to change from `height` to `maxHeight`, so I'm guessing it `maxHeight` was needed for forcing it into a single line and show the ellipsis and kept both.

PR #13887, opened 2018-09-23
don't make upload banner show up for folder creation

songgao:
@brayoh 's work [here](https://github.com/keybase/client/pull/13619) for KBFS-3123 mostly took care of KBFS-3122 as well. There's just this race condition where when we hear from journal after a new folder is created, and before we get the result of the `folderList` RPC call triggered by `editSuccess`, we don't know yet from `pathItems` that it's a folder. Without this commit we might show user the upload banner since the type is `unknown`. This PR makes the upload container look at `state.fs.edits` as well to eliminate this race.

### DIFF
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -188,6 +188,14 @@ function* folderList(
       ...entries.map(direntToPathAndPathItem),
     ]
     yield Saga.put(FsGen.createFolderListLoaded({pathItems: I.Map(pathItems), path: rootPath}))
+    if (action.type === FsGen.editSuccess) {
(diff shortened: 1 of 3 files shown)


shared/actions/fs/index.js
```diff
@@ -188,6 +188,14 @@ function* folderList(
       ...entries.map(direntToPathAndPathItem),
     ]
     yield Saga.put(FsGen.createFolderListLoaded({pathItems: I.Map(pathItems), path: rootPath}))
+    if (action.type === FsGen.editSuccess) {
+      // Note that we discard the Edit metadat here rather than immediately
```
chrisnojima: Metadata typo

PR #13888, opened 2018-09-23
fix breadcrumb dropdown order

songgao:
`w=1` is your friend


songgao:
Hmm you mean the container should give the component a reversed array for dropdown items?

PR #13894, opened 2018-09-24
update to framed-msgpack-rpc 1.1.15

mmaxim:
Tried it on mobile and it worked.


mmaxim:
Hmm actually Android didn't work.

mmaxim:
@chrisnojima this now passes CI, PTAL

PR #13897, opened 2018-09-24
use depth prepass check in client

heronhaye:
(changed to streaming)

### DIFF
--- a/go/libkb/api.go
+++ b/go/libkb/api.go
@@ -4,6 +4,7 @@
 package libkb
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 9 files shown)


go/libkb/api.go
```diff
@@ -330,7 +331,14 @@ func doRequestShared(m MetaContext, api Requester, arg APIArg, req *http.Request
 	}
 
 	if wantJSONRes {
-		reader := newCountingReader(internalResp.Body)
+		var buf bytes.Buffer
+		bodyTee := io.TeeReader(internalResp.Body, &buf)
```
heronhaye: not totally sure if this is the right idea performance-wise. all tests pass though.

maxtaco: :+1:

PR #13901, opened 2018-09-24
fix pinentry

chrisnojima:
@keybase/react-hackers


cjb:
@chrisnojima When did it break, is it broken in latest public?

chrisnojima:
it broke right after the release, so only affects linux ppl

PR #13902, opened 2018-09-24
background ek deletion

joshblum:
debugging scripts to help pinpoint query problems of ek generation


joshblum:
from various log sends the deletion on mobile can be a bottleneck so i'm moving it to the background.  @maxtaco going to ditch the creation scripts for now, ptal at the current changes

PR #13903, opened 2018-09-24
Use square icons.

adamjspooner:
@keybase/react-hackers, this is ready for review. It uses square icons to fix the share sheet bug and removes some icons that weren't in use.


chrisnojima:
lemme do a test build of this to ensure it passes itunes

PR #13908, opened 2018-09-25
store the list of TLFIDs associated with a team instead of just latest CORE-8930

mmaxim:
Fixes a problem where KBFS can insert new TLFID links into a team chain that already has one. Chat will use the first one if we need to lookup the team ID associated with a TLFID (we probably should have just made this something we got from the chat server), and if verifying that the team matches the TLFID of a conversation will use all of them. All other RPCs continue to use the latest TLF ID on the team.


go/teams/storage.go
```diff
@@ -89,7 +89,7 @@ type DiskStorage struct {
 }
 
 // Increment to invalidate the disk cache.
-const diskStorageVersion = 9
+const diskStorageVersion = 10
```
mmaxim: Note this bumps the team cache

PR #13914, opened 2018-09-25
5: react-window for inbox

chrisnojima:
Switches to react-window for inbox

### DIFF
--- a/shared/chat/inbox/index.desktop.js
+++ b/shared/chat/inbox/index.desktop.js
@@ -33,12 +33,6 @@ class Inbox extends PureComponent<Props, State> {
       listRowsResized = true
     }
 
-    // divider size changed
-    // if (prevProps.showSmallTeamsExpandDivider !== this.props.showSmallTeamsExpandDivider) {
-    // // just reset all sizes, this should be rare
-    // listRowsResized = true
-    // }
-
     // filter / not filter
     if (!!prevProps.filter !== !!this.props.filter) {
       listRowsResized = true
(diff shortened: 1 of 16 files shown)


chrisnojima:

shared/chat/inbox/index.desktop.js
```diff
@@ -24,15 +25,27 @@ class Inbox extends PureComponent<Props, State> {
...
+    let listRowsResized = false
     if (prevProps.smallTeamsExpanded !== this.props.smallTeamsExpanded) {
-      setTimeout(() => {
-        this._updateShowFloating()
-      }, 500)
+      listRowsResized = true
+    }
+
+    // divider size changed
+    // if (prevProps.showSmallTeamsExpandDivider !== this.props.showSmallTeamsExpandDivider) {
```
buoyad: 🔥

shared/chat/inbox/index.desktop.js
```diff
@@ -150,15 +147,22 @@ class Inbox extends PureComponent<Props, State> {
...
+                <VariableSizeList
+                  height={height}
+                  width={width}
+                  ref={this._setRef}
+                  onItemsRendered={this._onItemsRendered}
+                  itemCount={this.props.rows.length}
+                  itemSize={this._itemSizeGetter}
+                  estimatedItemSize={56}
+                >
+                  {({index, style}) => this._itemRenderer(index, style)}
```
MarcoPolo: I wonder if it would be even faster if we used PureComponent here. Probably a bit of work to move the itemRenderer out.

chrisnojima: No time is spent here basically

PR #13915, opened 2018-09-25
6: More inbox optimizations for scrolling

chrisnojima:
- [x] dont load team operations until info menu opens up
- [x] smarter memoize for inbox small rows
- [x] cache avatars for 30 minutes
- [x] tapping inbox items caused the entire list to rerender

### DIFF
--- a/shared/chat/conversation/info-panel/menu/index.js
+++ b/shared/chat/conversation/info-panel/menu/index.js
@@ -13,6 +13,8 @@ export type Props = {
   memberCount: number,
   teamname: string,
   visible: boolean,
+  hasCanPerform: boolean,
+  loadOperations: () => void,
   onAddPeople: () => void,
   onHidden: () => void,
   onInvite: () => void,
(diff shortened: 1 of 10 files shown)


shared/chat/conversation/info-panel/menu/index.js
```diff
@@ -33,61 +35,70 @@ const Header = ({teamname, memberCount}: {teamname: string, memberCount: number}
...
-            <Kb.Text style={styles.text} type={Styles.isMobile ? 'BodyBig' : 'Body'}>
-              {props.manageChannelsTitle}
-            </Kb.Text>
-            {props.badgeSubscribe && <Kb.Box style={styles.badge} />}
-          </Kb.Box>
-        ),
-      }
+class InfoPanelMenu extends React.Component<Props> {
+  componentDidDUpdate(prevProps: Props) {
+    if (this.props.hasCanPerform && this.props.visible !== prevProps.visible) {
```
MarcoPolo: What's the reason for checking the diff on visibility?

chrisnojima: Because by default it's not visible. Only on hover. It's still mounted though. So as you scroll around it'd call this (for no reason). Now it'll only call when you click the menu

PR #13916, opened 2018-09-25
Enforce 8 character minimum on *new* and *updated* passphrases.

heronhaye:
Not too much work...old 6-char passwords still work fine in both terminal and GUI.

Password strength meters are generally useless in my experience or at worst, mislead users into thinking weak passwords are strong: https://nakedsecurity.sophos.com/2015/03/02/why-you-cant-trust-password-strength-meters/.

Passphrase generation as Thomas suggested might be cool in the GUI. There is the risk of users mixing up paperkeys and passphrases, but if the new signup flow in the GUI doesn't include paperkeys then I think we should add it (in another ticket).

### DIFF
--- a/go/engine/signup_test.go
+++ b/go/engine/signup_test.go
@@ -274,7 +274,7 @@ func TestSignupPassphrases(t *testing.T) {
 	tc := SetupEngineTest(t, "signup")
 	defer tc.Cleanup()
 	CreateAndSignupFakeUserWithPassphrase(tc, "pass", "123456789012")
-	CreateAndSignupFakeUserWithPassphrase(tc, "pass", "123456")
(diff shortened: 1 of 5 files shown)


heronhaye:
@keybase/react-hackers 

Could I get a review? The intent is to increase password length from 6 to 8 for new and updated passwords, but allow old shorter passwords for login.

dabura667:
We should also check against pwned passwords API.
https://haveibeenpwned.com/API/v2#PwnedPasswords

There is a suuuper simple example of how to use the API here:

Site: https://junderw.github.io/pwcheck.html
Source:
https://github.com/junderw/junderw.github.io/blob/83578d65ea8a3f5b8725b7bf48941356354e6637/pwcheck.html#L19-L54

PR #13917, opened 2018-09-25
min term width

heronhaye:
while I figure out how to run kbfs docker tests...this should fix master.

### DIFF
--- a/go/minterm/minterm.go
+++ b/go/minterm/minterm.go
@@ -79,7 +79,7 @@ func (m *MinTerm) getNewTerminal(prompt string) (*terminal.Terminal, error) {
 	term := terminal.NewTerminal(m.getReadWriter(), prompt)
 	a, b := m.Size()
 	if a < 80 {
-		a, b = 80, 24
+		a = 80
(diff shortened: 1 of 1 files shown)


go/minterm/minterm.go
```diff
@@ -78,6 +78,9 @@ func (m *MinTerm) fdIn() int { return int(m.termIn.Fd()) }
 func (m *MinTerm) getNewTerminal(prompt string) (*terminal.Terminal, error) {
 	term := terminal.NewTerminal(m.getReadWriter(), prompt)
 	a, b := m.Size()
+	if a < 80 {
+		a, b = 80, 24
+	}
```
zapu: is there a reason not to do height independently?

heronhaye: I was thinking someone could realistically have a short terminal, say in a tiling wm.

zapu: height probably doesn't matter anyway, but we likely aren't calculating the right thing here for example when someone resizes their terminal to 70x60 for some reason

PR #13920, opened 2018-09-25
fix bug where FilePreviewLoaded overrides loaded into pending

songgao:
Otherwise opening widget would override some `FolderPathItem` into pending sometimes, and we'd show placeholders in the Files tab.

### DIFF
--- a/shared/reducers/fs.js
+++ b/shared/reducers/fs.js
@@ -7,6 +7,19 @@ import * as Types from '../constants/types/fs'
 
 const initialState = Constants.makeState()
 
+const coalesceFolderUpdate = (
+  original: Types.FolderPathItem,
+  updated: Types.FolderPathItem
+): Types.FolderPathItem =>
+  // We don't want to override a loaded folder into pending, because otherwise
+  // next time user goes into that folder we'd show placeholders. We also don't
+  // want to simply use the original PathItem, since it's possible some
+  // metadata has updated. So use the new item, but reuse children and
+  // progress.
+  original.progress === 'loaded' && updated.progress === 'pending'
+    ? updated.withMutations(u => u.set('children', original.children).set('progress', 'loaded'))
+    : updated
+
 export default function(state: Types.State = initialState, action: FsGen.Actions) {
   switch (action.type) {
     case FsGen.resetStore:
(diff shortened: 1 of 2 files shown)


shared/reducers/fs.js
```diff
@@ -7,14 +7,31 @@ import * as Types from '../constants/types/fs'
...
+  original: Types.FolderPathItem,
+  updated: Types.FolderPathItem
+): Types.FolderPathItem =>
+  // We don't want to override a loaded folder into pending, because otherwise
+  // next time user goes into that folder we'd show placeholders. We also don't
+  // want to simply use the original PathItem, since it's possible some
+  // metadata has updated. So use the new item, but reuse children and
+  // progress.
+  original.progress === 'loaded' && updated.progress === 'pending'
+    ? updated.withMutations(u => u.set('children', original.children).set('progress', 'loaded'))
```
jzila: please put these `.set` calls on separate lines.

shared/reducers/fs.js
```diff
@@ -7,14 +7,31 @@ import * as Types from '../constants/types/fs'
...
   switch (action.type) {
     case FsGen.resetStore:
       return initialState
     case FsGen.filePreviewLoaded:
       return state.updateIn(['pathItems', action.payload.path], (original: Types.PathItem) => {
         const {meta} = action.payload
-        if (!original || original.type !== 'file' || meta.type !== 'file') {
+        if (original.type === 'folder' && meta.type === 'folder') {
+          const c = coalesceFolderUpdate(original, meta)
+          console.log({songgao: 'coalesce', original, meta, c})
```
jzila: :fire:

shared/reducers/fs.js
```diff
@@ -7,14 +7,31 @@ import * as Types from '../constants/types/fs'
...
+
 export default function(state: Types.State = initialState, action: FsGen.Actions) {
   switch (action.type) {
     case FsGen.resetStore:
       return initialState
     case FsGen.filePreviewLoaded:
       return state.updateIn(['pathItems', action.payload.path], (original: Types.PathItem) => {
         const {meta} = action.payload
-        if (!original || original.type !== 'file' || meta.type !== 'file') {
+        if (original.type === 'folder' && meta.type === 'folder') {
```
jzila: What about when the meta type is `unknown`?

shared/reducers/fs.js
songgao: Hmm no matter what I try, my linter keeps merging them back into single line. I guess linter doesn't do that when the line is too short?

jzila: Your linter is too aggressive. Can you make it not rewrite the files?

PR #13926, opened 2018-09-26
populate current wallet name when editing

zanderz:


shared/wallets/wallet/settings/container.js
```diff
@@ -17,7 +17,7 @@ import * as WalletsGen from '../../../actions/wallets-gen'
 const mapStateToProps = (state: TypedState, {routeProps}) => {
   const accountID = routeProps.get('accountID')
   const account = Constants.getAccount(state, accountID)
-  const name = account.name || accountID || account.accountID
+  const name = account.name || Constants.getAccountName(account) || accountID || account.accountID
```
chrisnojima: i think we decided this should always be account.name and we're going to make it such that new wallets always have to have one

shared/wallets/wallet/settings/popups/rename-account/container.js
```diff
@@ -37,7 +42,7 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => ({
 
 export default compose(
   withStateHandlers(
-    {name: ''},
+    {name: '__INITIAL_PLACEHOLDER_NAME__'},
```
chrisnojima: you can just init this with the name. 
```
{name: props => props.name}
```

zanderz: i assumed the props wouldn't be available there - don't know why I didn't just try it.

zanderz: that didn't work. I changed it to use `null`, since `''` wouldn't work.

shared/wallets/wallet/settings/container.js
zanderz: ok

PR #13929, opened 2018-09-26
Use a new selectableText prop to allow text highlighting only on the transaction details page

cjb:
@keybase/react-hackers 

We previously tried to stop clicks on a transaction row going through to the transaction details page if they were on a selectable area, but having areas that are only sometimes "real" clicks is too tricky.  Instead, always have the text unselectable on the transaction row, but always selectable on the transaction-details view.

Also, initialize timestamps to null rather than 0, since 0 is a valid epoch timestamp.


shared/wallets/transaction/index.js
```diff
@@ -252,16 +249,12 @@ export const TimestampLine = (props: TimestampLineProps) => {
     )
   }
   if (!props.timestamp) {
-    return (
-      <Text type="BodySmall">
-        {props.relative ? 'Pending' : "The Stellar network hasn't confirmed your transaction."}
```
zanderz: what was the relative/pending thing?

cjb: Holdover from before we switched the timestamp format to be the same as Chat -- after doing that we can just use the same text on both screens.

PR #13936, opened 2018-09-26
Workaround box2 injecting too many gaps and disable chat to send form on mobile

buoyad:
`Box2` was injecting a gap after `WalletsIcon`, which has a wrapper that renders nothing if we shouldn't see it. I worked around it by manually adding the gaps where they're needed. I also disabled navigating to the send form on mobile as no work has been done there. r? @keybase/react-hackers


shared/chat/conversation/messages/account-payment/container.js
```diff
@@ -82,6 +82,10 @@ const mapStateToProps = (state, ownProps: OwnProps) => {
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
   onSend: () => {
+    if (Styles.isMobile) {
```
chrisnojima: lets just make a dummy component in the route and show a `<Text> TODO` kind of thing

PR #13941, opened 2018-09-26
retry home a few times before blackbar

maxtaco:


maxtaco:
Oh, you think the RPC is just dead inflight. Maybe....

PR #13942, opened 2018-09-26
fix blackbar when going to Files tab from folder icon

songgao:
My fix from yesterday causes blackbar some times. This should fix that.

### DIFF
--- a/shared/reducers/fs.js
+++ b/shared/reducers/fs.js
@@ -27,6 +27,13 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
     case FsGen.filePreviewLoaded:
       return state.updateIn(['pathItems', action.payload.path], (original: Types.PathItem) => {
         const {meta} = action.payload
+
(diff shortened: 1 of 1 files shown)


shared/reducers/fs.js
```diff
@@ -27,6 +27,13 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
     case FsGen.filePreviewLoaded:
       return state.updateIn(['pathItems', action.payload.path], (original: Types.PathItem) => {
         const {meta} = action.payload
+
+        // Since updateIn passes `original` in as `any`, it may not be an
+        // actual PathItem.
+        if (!original) {
+          return meta
```
jzila: Will this work to set it even if it doesn't exist?

songgao: Yup:

```
> (new I.Map({})).updateIn(['a'], old=>1)
Map { "a": 1 }
```

PR #13946, opened 2018-09-26
remote whitelist config

chrisnojima:
This allows the daemon to give us a configuration object at runtime. In order for this to work sanely with the current architecture this info has to be known at startup ASAP, therefore any changes to this are saved after a login to a file and that file is read on app start once.
Additionally, due to some complexity with the inability on mobile to do this kind of reading synchronously in the JS side we must do this sync reading on the native side and pass the data through the bridge before the app starts. We leverage the existing constants load to do this.


chrisnojima:
- [ ] core support from @maxtaco

shared/actions/config/index.js
```diff
@@ -360,6 +361,17 @@ const allowLogoutWaiters = (_, action: ConfigGen.LogoutHandshakePayload) =>
     ),
   ])
 
+const updateServerConfig = (state: TypedState) =>
+  Saga.call(function*() {
+    // TODO
+    // const serverConfig = yield Saga.call(RPCTypes.API)
```
chrisnojima: when core is done we plumb it through here, for now its hardcoded

shared/app/rpc-stats.js
```diff
@@ -233,9 +184,5 @@ const styles = styleSheetCreate({
   },
 })
 
-// We only use username and pull stats on a timer. Don't want the stats gathering to affect redux at all
-const mapStateToProps = (state: TypedState) => ({
-  username: state.config.username,
-})
-
-export default connect(mapStateToProps, () => ({}), (s, d, o) => ({...o, ...s, ...d}))(RpcStats)
+const TheRPCStats = printRPCStats ? RpcStats : () => null
```
chrisnojima: we can now drive this from the serverconfig

shared/app/server-config.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+import * as File from '../util/file'
+import {serverConfigFileName} from '../constants/platform'
+import logger from '../logger'
+
+export async function updateServerConfigLastLoggedIn(username: string, serverConfig: Object) {
```
chrisnojima: manages writing this data from the JS side. first we read the old data and merge in the new data

shared/react-native/ios/Keybase/AppDelegate.m
```diff
@@ -62,23 +62,17 @@ - (void) setupGo
                                                    } error:&err];
 }
 
-#ifdef SYSTRACING
```
chrisnojima: this is long deprecated

chrisnojima:
- [ ] features payload seems incorrect: c  @maxtaco

chrisnojima:

PR #13952, opened 2018-09-27
Use content-disposition to control file previews

taruti:
Client side of https://github.com/keybase/kbfs/pull/1794

### DIFF
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -464,6 +464,12 @@ function* ignoreFavoriteSaga(action: FsGen.FavoriteIgnorePayload): Saga.SagaGene
   }
 }
 
+// Return a header till first semicolon in lower case.
+const headerTillSemiLower = (header: string): string => {
+  const idx = header.indexOf(';')
+  return (idx > -1 ? header.slice(0, idx) : header).toLowerCase()
+}
+
 // Following RFC https://tools.ietf.org/html/rfc7231#section-3.1.1.1 Examples:
 //   text/html;charset=utf-8
 //   text/html;charset=UTF-8
@@ -471,9 +477,9 @@ function* ignoreFavoriteSaga(action: FsGen.FavoriteIgnorePayload): Saga.SagaGene
 //   text/html; charset="utf-8"
 // The last part is optional, so if `;` is missing, it'd be just the mimetype.
 const extractMimeFromContentType = (contentType, disposition: string): Types.Mime => {
-  const ind = contentType.indexOf(';')
-  const mimeType = (ind > -1 ? contentType.slice(0, ind) : contentType).toLowerCase()
-  return Constants.makeMime({mimeType, displayPreview: disposition !== 'attachment'})
+  const mimeType = headerTillSemiLower(contentType)
+  const displayPreview = headerTillSemiLower(disposition) !== 'attachment'
+  return Constants.makeMime({mimeType, displayPreview})
 }
 
 const getMimeTypePromise = (localHTTPServerInfo: Types._LocalHTTPServer, path: Types.Path) =>
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -470,23 +470,24 @@ function* ignoreFavoriteSaga(action: FsGen.FavoriteIgnorePayload): Saga.SagaGene
 //   Text/HTML;Charset="utf-8"
 //   text/html; charset="utf-8"
 // The last part is optional, so if `;` is missing, it'd be just the mimetype.
-const extractMimeTypeFromContentType = (contentType: string): string => {
+const extractMimeTypeFromContentType = (contentType, disposition: string): Types.Mime => {
   const ind = contentType.indexOf(';')
-  return (ind > -1 ? contentType.slice(0, ind) : contentType).toLowerCase()
+  const mimeType = (ind > -1 ? contentType.slice(0, ind) : contentType).toLowerCase()
+  return Constants.makeMime({mimeType, displayPreview: disposition !== 'attachment'})
 }
 
 const getMimeTypePromise = (localHTTPServerInfo: Types._LocalHTTPServer, path: Types.Path) =>
   new Promise((resolve, reject) =>
     getContentTypeFromURL(
       Constants.generateFileURL(path, localHTTPServerInfo),
-      ({error, statusCode, contentType}) => {
+      ({error, statusCode, contentType, disposition}) => {
         if (error) {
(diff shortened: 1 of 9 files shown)


shared/actions/fs/index.js
```diff
@@ -462,23 +462,24 @@ function* ignoreFavoriteSaga(action: FsGen.FavoriteIgnorePayload): Saga.SagaGene
 //   Text/HTML;Charset="utf-8"
 //   text/html; charset="utf-8"
 // The last part is optional, so if `;` is missing, it'd be just the mimetype.
-const extractMimeTypeFromContentType = (contentType: string): string => {
+const extractMimeTypeFromContentType = (contentType, disposition: string): Types.Mime => {
```
songgao: Would you mind either renaming `extractMimeTypeFromContentType` into `extractMimeFromContentType`, or renaming `Types.Mime` into `Types.MimeType` so they are consistent?

shared/actions/fs/index.js
```diff
@@ -462,23 +462,24 @@ function* ignoreFavoriteSaga(action: FsGen.FavoriteIgnorePayload): Saga.SagaGene
...
   const ind = contentType.indexOf(';')
-  return (ind > -1 ? contentType.slice(0, ind) : contentType).toLowerCase()
+  const mimeType = (ind > -1 ? contentType.slice(0, ind) : contentType).toLowerCase()
+  return Constants.makeMime({mimeType, displayPreview: disposition !== 'attachment'})
```
songgao: I don't think this would cover this form:

```
Content-Disposition: attachment; filename="filename.jpg"
```

Also, it seems better to filter the other way around, i.e. `disposition === 'inline`, so that we don't accidentally display non-inline and non-attachment stuff inline. WDYT?

shared/constants/fs.js
```diff
@@ -573,18 +578,21 @@ export const userTlfHistoryRPCToState = (
...
-  if (mimeType.startsWith('image/')) {
-    return 'image'
-  }
-  if (mimeType.startsWith('audio/') || mimeType.startsWith('video/')) {
-    return 'av'
-  }
-  if (mimeType === 'application/pdf') {
-    return 'pdf'
+export const viewTypeFromMimeType = (mime: ?Types.Mime): Types.FileViewType => {
+  if (mime != null && mime.displayPreview) {
```
songgao: We should use `!==` here. Also, I think `if (mime && mime.displayPreview)` would also work here.

shared/constants/fs.js
```diff
@@ -651,7 +659,7 @@ export const syntheticEventToTargetRect = (evt?: SyntheticEvent<>): ?ClientRect
 // shouldUseOldMimeType determines if mimeType from newItem should reuse
 // what's in oldItem.
 export const shouldUseOldMimeType = (oldItem: Types.FilePathItem, newItem: Types.FilePathItem): boolean => {
-  if (oldItem.mimeType === '' || newItem.mimeType !== '') {
+  if (oldItem.mimeType === null || newItem.mimeType !== null) {
```
songgao: Slight preference for `if (!oldItem.mimeType || !new.mimeType)`

Same for other places (e.g. in `shared/fs/filepreview/view-container.js`) if you agree.

shared/fs/filepreview/view-container.js
```diff
@@ -28,7 +28,7 @@ const mapStateToProps = (state: TypedState, {path}: Props) => {
   const _pathItem = state.fs.pathItems.get(path, Constants.makeFile())
   return {
     _serverInfo: state.fs.localHTTPServerInfo,
-    mimeType: _pathItem.type === 'file' ? _pathItem.mimeType : '',
+    mimeType: _pathItem.type === 'file' ? _pathItem.mimeType : null,
     isSymlink: _pathItem.type === 'symlink',
   }
```
songgao: Hmm this must have been my fault originally. But since you're here, would you mind replacing `mimeType` and `isSymlink` with `_pathItem`, and do those in `mergeProps`? The reason is `mapStateToProps should be as fast as possible, as long as the change doesn't cause re-rendering more. In this case, being contingent on a single `pathItem` is infrequent enough I think.

shared/actions/fs/index.js
taruti: Done

shared/actions/fs/index.js
taruti: Old server provides no disposition, so I'd like to default to more previews if there is a weird mismatch.

taruti: Will handle more complex dispositions correctly.

shared/constants/fs.js
taruti: Flow is not happy with those. Iirc because undefined vs null.

shared/constants/fs.js
taruti: Doesn't "!foo" mean "!(foo == true)" with all the conversions, perhaps keep '!==' there?

shared/constants/fs.js
songgao: Hmm but with `mime: ?Mime,` above, `mimeType` can't be `undefined`., but either `null` or `Types.Mime`. Any chance flow issue was on a different function signature?

songgao: I think `if (mime && mime.displayPreview)` would be the best here, but it's really a nit ...

shared/actions/fs/index.js
songgao: > Old server provides no disposition, so I'd like to default to more previews if there is a weird mismatch.

If we merge both together, this should be super rare. Also, we could probably temporarily default to `inline` if it's empty until more releases have gone out. According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition it can also be `form-data`. On the other hand we don't use any of those on the Go side so it's probably fine. It'd be nice to have a comment either here or on the Go side explaining it though.

shared/constants/fs.js
taruti: "!==" does not work due to Flow pointing out "Property displayPreview is missing in null or undefined" (?Mime includes undefined), but using the "if(mime ..." works, will use that.

shared/constants/fs.js
taruti: Ok, I'll do it that way.

shared/constants/fs.js
songgao: Whoa I had no idea `?Mime` covers `undefined`. Thanks for pointing it out!

taruti:
Rebasing now.

PR #13953, opened 2018-09-27
Tweak message spacing.

adamjspooner:
@keybase/react-hackers, this adjusts the message spacing ever so slightly. There were too many nested paddings/margins causing different kinds of messages to have different levels of spacing. This cleans that up.


adamjspooner:
For instance, take a look at a message that has reactions.

PR #13955, opened 2018-09-27
patch NCC-KB2018-006

maxtaco:


maxtaco:
@mlsteele ready for another look

go/libkb/sig_chain.go
```diff
@@ -545,8 +544,17 @@ func isSubchainStart(currentLink *ChainLink, prevLink *ChainLink) bool {
 	if !currentLink.ToEldestKID().Equal(prevLink.ToEldestKID()) {
 		return true
 	}
+
 	// case 4
-	return hardcodedResets[currentLink.unpacked.sigID]
+	found, _, err := currentLink.checkSpecialLinksTable(hardcodedResets, uid, "harcoded resets")
+	if err != nil {
+		m.CWarningf("Error in isSubchainStart, will be ignored: %s", err.Error())
+		return false
```
mlsteele: why ignore rather than bail out?

maxtaco: i was going down that path but then i'd have to change a lot of callers that are in boolean conjunctions.  I can do it though, but it does mitigate the attack as is.

maxtaco: (a lot = ~3)

go/libkb/sig_chain.go
```diff
@@ -962,15 +978,14 @@ func (l *SigChainLoader) LoadLinksFromStorage() (err error) {
...
+		l.M().CDebugf("| short-circuting LoadLinksFromStorage: LoadLastLinkIDFromStorage returned err=%v", err)
+		if mt == nil {
 			l.M().CDebugf("| mt (MerkleTriple) nil result from load last link ID from storage")
-		} else if mt.LinkID == nil {
+		}
+		if mt != nil && mt.LinkID == nil {
 			l.M().CDebugf("| mt (MerkleTriple) from storage has a nil link ID")
 		}
-		return
+		return nil
```
mlsteele: this used to return the error if there was one from LoadLastLinkIDFromStorage

PR #13956, opened 2018-09-27
Yarn prettier helper

chrisnojima:
This updates prettier and adds a script that runs it on all files but is smarter about blacklist/whitelist
commit 1 is the helper
commit 2 is updating prettier
commit 3 is running prettier


chrisnojima:
@keybase/react-hackers this updates prettier

PR #13959, opened 2018-09-27
Mock timestamp for stories

MarcoPolo:
Fixes issue where Jenkins will fail because it's either faster or slower than your computer
@keybase/react-hackers


shared/people/follow-notification/index.stories.js
```diff
@@ -5,11 +5,15 @@ import * as Sb from '../../stories/storybook'
 import FollowNotification, {type Props} from '.'
 import moment from 'moment'
 
+// Mock moment.now(), since moment uses this for everything
+// $ForceType
+moment.now = () => 1538066670959
```
chrisnojima: we shouldn't do it this way. we already have mocks for storybook, see the timestamp mocks etc

MarcoPolo:
@chrisnojima updated

PR #13969, opened 2018-09-27
always load the team name when unboxing any team based chat 

mmaxim:


xgess:
can confirm this fixes the bug

mmaxim:
Yeah I have tests for that offline check, so I think it’s probably ok, but will manually test extensively as well

PR #13972, opened 2018-09-27
'See transaction details' in sendPayment message menu

buoyad:
- Changes `stellar1.PaymentID` to a string typedef
- Adds accountID and paymentID to `chat1.UIPaymentInfo`
- Adds an opaque type `PaymentID` with mapping functions to/from the service type
- Changes the structure of the payments store from a flat list to a nested map (cc @cjb)
  - Old: `AccountID -> List<Payment>`
  - New `AccountID -> PaymentID -> Payment`
- Changes reducers/getters/containers to work with the above
- Adds loading state to tx details and account list pane
r? @keybase/react-hackers


go/stellar/loader.go
```diff
@@ -203,17 +203,17 @@ func (p *Loader) runRequests() {
 func (p *Loader) loadPayment(id stellar1.PaymentID) {
 	ctx := context.Background()
 	s := getGlobal(p.G())
-	details, err := s.remoter.PaymentDetails(ctx, id.TxID.String())
+	details, err := s.remoter.PaymentDetails(ctx, id.String())
```
patrickxb: This isn't correct.  Remoter needs the transaction id, not the payment id.

go/stellar/stellarsvc/frontend.go
```diff
@@ -393,7 +393,7 @@ func (s *Server) GetPaymentDetailsLocal(ctx context.Context, arg stellar1.GetPay
 	}
 
 	oc := stellar.NewOwnAccountLookupCache(ctx, s.G())
-	details, err := s.remoter.PaymentDetails(ctx, arg.Id.TxID.String())
+	details, err := s.remoter.PaymentDetails(ctx, arg.Id.String())
```
patrickxb: Again, this needs to be a transaction id, not a payment id.

protocol/avdl/stellar1/local.avdl
```diff
@@ -102,7 +102,7 @@ protocol local {
   record PaymentsPageLocal {
     array<PaymentOrErrorLocal> payments;
     union { null, PageCursor } cursor;
-    union { null, PaymentID } oldestUnread;  // draw orange line underneath this payment
+    PaymentID oldestUnread;  // draw orange line underneath this payment
```
patrickxb: This loses some information.  If null, then nothing unread.

go/client/chat_cli_rendering.go
```diff
@@ -510,7 +510,7 @@ func formatSendPaymentMessage(g *libkb.GlobalContext, body chat1.MessageSendPaym
 		g.Log.CDebugf(ctx, "GetWalletClient() error: %s", err)
 		return "[error getting payment details]"
 	}
-	details, err := cli.PaymentDetailCLILocal(ctx, body.PaymentID.TxID.String())
+	details, err := cli.PaymentDetailCLILocal(ctx, body.PaymentID.String())
```
patrickxb: I think this should be transaction id too.

buoyad:
@keybase/react-hackers bump

shared/constants/wallets.js
```diff
@@ -326,6 +338,32 @@ const paymentToYourRoleAndCounterparty = (
   }
 }
 
+// Update payment, take all new fields except any that contain the default value
+const emptyPayment = makePayment()
+// $FlowIssue thinks we don't have toJS()
+const keys = Object.keys(emptyPayment.toJS())
```
chrisnojima: why not just get the keys? `.keySeq().toArray()`

shared/constants/wallets.js
```diff
@@ -190,7 +190,19 @@ const paymentResultToPayment = (w: RPCTypes.PaymentOrErrorLocal) => {
   if (!w.payment) {
     return makePayment({error: w.err})
   }
-  const p = w.payment
+  return makePayment(rpcPaymentToPaymentCommon(w.payment))
+}
+
+const paymentDetailResultToPayment = (p: RPCTypes.PaymentDetailsLocal) => {
```
chrisnojima: prefer `=> makePayment(...`

shared/constants/wallets.js
```diff
@@ -326,6 +338,32 @@ const paymentToYourRoleAndCounterparty = (
   }
 }
 
+// Update payment, take all new fields except any that contain the default value
+const emptyPayment = makePayment()
+// $FlowIssue thinks we don't have toJS()
+const keys = Object.keys(emptyPayment.toJS())
+const updatePayment = (oldPayment: Types.Payment, newPayment: Types.Payment): Types.Payment => {
+  const res = oldPayment.withMutations(paymentMutable => {
```
chrisnojima: why not `merge`

shared/constants/wallets.js
```diff
@@ -347,19 +385,21 @@ const getDisplayCurrencies = (state: TypedState) => state.wallets.currencies
 const getDisplayCurrency = (state: TypedState, accountID?: Types.AccountID) =>
   state.wallets.currencyMap.get(accountID || getSelectedAccount(state), makeCurrency())
 
-const getPayments = (state: TypedState, accountID?: Types.AccountID) =>
-  state.wallets.paymentsMap.get(accountID || getSelectedAccount(state))
+const getPayments = (state: TypedState, accountID?: Types.AccountID) => {
+  const map = state.wallets.paymentsMap.get(accountID || getSelectedAccount(state), null)
```
chrisnojima: just return w/o saving map

shared/constants/wallets.js
```diff
@@ -347,19 +385,21 @@ const getDisplayCurrencies = (state: TypedState) => state.wallets.currencies
...
+  return map
+}
 
-const getPendingPayments = (state: TypedState, accountID?: Types.AccountID) =>
-  state.wallets.pendingMap.get(accountID || getSelectedAccount(state))
+const getPendingPayments = (state: TypedState, accountID?: Types.AccountID) => {
+  const map = state.wallets.pendingMap.get(accountID || getSelectedAccount(state), null)
```
chrisnojima: ditto

shared/reducers/wallets.js
```diff
@@ -26,19 +26,17 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
...
-          })
-        )
+      return state.updateIn(['paymentsMap', action.payload.accountID], (payments = I.Map()) =>
+        Constants.updatePaymentMap(payments, [action.payload.payment])
       )
     case WalletsGen.paymentsReceived:
       return state
-        .setIn(['paymentsMap', action.payload.accountID], I.List(action.payload.payments))
-        .setIn(['pendingMap', action.payload.accountID], I.List(action.payload.pending))
+        .updateIn(['paymentsMap', action.payload.accountID], (paymentsMap = I.Map()) =>
```
chrisnojima: prefer
```
.updateIn(['paymentsMap', action.payload.accountID], I.Map(), paymentsMap =>
```

shared/reducers/wallets.js
```diff
@@ -26,19 +26,17 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
...
+          Constants.updatePaymentMap(paymentsMap, action.payload.payments)
+        )
+        .updateIn(['pendingMap', action.payload.accountID], (pendingMap = I.Map()) =>
```
chrisnojima: ditto

shared/wallets/transaction-details/index.js
```diff
@@ -164,23 +168,31 @@ class TransactionDetails extends React.Component<Props> {
   }
 
   render() {
-    const {sender, receiver} = propsToParties(this.props)
```
chrisnojima: likely simpler to just make a wrapper that does this and renders 2 things. the loader and the previous version of this

shared/wallets/wallet/container.js
```diff
@@ -47,14 +51,18 @@ const mergeProps = (stateProps, dispatchProps) => {
...
+  if (!paymentsMap) {
     return ['notLoadedYet']
   }
-  if (payments.count() === 0) {
+  if (paymentsMap.size === 0) {
     return ['noPayments']
   }
-  return payments.map(p => ({paymentID: p.id, status: p.statusSimplified})).toArray()
-}
+  const payments = paymentsMap.toList()
```
chrisnojima: why call `toList`

shared/reducers/wallets.js
buoyad: I got turned off of `notSetValue` making this PR because of [this](https://facebook.github.io/immutable-js/docs/#/Map/updateIn):

> If the `updater` function returns the same value it was called with, then no change will occur. This is still true if `notSetValue` is provided.
> For code using ES2015 or later, using `notSetValue` is discourged in favor of function parameter default values.

We use `null` to mean not loaded and an empty `Map` to mean loaded and empty, so that was causing a problem

shared/constants/wallets.js
buoyad: I don't think that works with `Record`s

chrisnojima: `.toSeq().keySeq().toArray()`

shared/wallets/wallet/container.js
buoyad: I want to throw out the keys and use the list of values from here on, `toList` seemed like a good way to do that

buoyad:
thanks @chrisnojima, addressed feedback PTAL

PR #13973, opened 2018-09-28
3: box2 css

chrisnojima:
vs the base pr, initial mount of the thread storybook (which has a lot of nested box2s)

root:
692.25
671.28
546.73

pr:
681.86
655.16
505.03


shared/chat/conversation/messages/wrapper/wrapper-timestamp/index.js
```diff
@@ -92,7 +92,7 @@ class _WrapperTimestamp extends React.Component<Props & OverlayParentProps, Stat
                 underlayColor: Styles.globalColors.blue5,
               }
             : {})}
-          direction="column"
+          direction="vertical"
```
chrisnojima: glamorous makes us lose all typing

shared/common-adapters/box.desktop.js
```diff
@@ -73,38 +56,9 @@ class Box2 extends React.Component<Box2Props> {
...
-    alignItems: 'stretch',
-    justifyContent: 'flex-start',
-  },
-  hrbox: {
-    ...globalStyles.flexBoxRowReverse,
-    alignItems: 'stretch',
-    justifyContent: 'flex-start',
-  },
-}
+const vBoxGap = (key, gap) => <div key={key} className={`box2_gap_vertical_${gap}`} />
```
chrisnojima: note these are just functions and not components anymore. if we do this we actually don't make these wrapped components w/ lifecycles so we can reduce some overhead here

chrisnojima:
@keybase/react-hackers

chrisnojima:
@keybase/react-hackers bump

PR #13976, opened 2018-09-28
Factor out fetching display currency settings

zapu:
`stellar.GetAccountDisplayCurrency` will fetch current display currency for accountID, but return "USD" if fetched currency is NULL. This has been done manually in two places and forgotten to be done in one.

### DIFF
--- a/go/stellar/stellar.go
+++ b/go/stellar/stellar.go
@@ -1110,8 +1110,29 @@ func DeleteAccount(m libkb.MetaContext, accountID stellar1.AccountID) error {
 	return remote.Post(m.Ctx(), m.G(), nextBundle)
 }
 
-func GetCurrencySetting(mctx libkb.MetaContext, remoter remote.Remoter, accountID stellar1.AccountID) (res stellar1.CurrencyLocal, err error) {
+const defaultCurrencySetting = "USD"
+
+// GetAccountDisplayCurrency gets currency setting from the server, and it
+// returned currency is empty (NULL in database), then default "USD" is used.
+// When creating a wallet, client always sets default currency setting. Also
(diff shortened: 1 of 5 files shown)


go/stellar/stellar.go
```diff
@@ -64,7 +65,7 @@ func CreateWallet(ctx context.Context, g *libkb.GlobalContext) (created bool, er
 		g.Log.CErrorf(ctx, "We've just posted a bundle that's missing PrimaryAccount: %s", err)
 		return false, err
 	}
-	if err := remote.SetAccountDefaultCurrency(ctx, g, primary.AccountID, "USD"); err != nil {
+	if err := remote.SetAccountDefaultCurrency(ctx, g, primary.AccountID, DefaultCurrencySetting); err != nil {
```
patrickxb: Let's remove this block.  

I don't think we should save this when wallets are created, let the user set it.  Confirmed with @malgorithms to remove it.

zapu: It's purely technical, though, isn't it? it will still return the default "USD" for non-set accounts.

Also it was put there as preparation for when we can detect region/locale and set it accordingly - https://keybase.atlassian.net/browse/CORE-8071

zapu: or do we want to support "outside-currency-less" accounts? both are fine by me, I just want to ensure we are not forgetting anything that we discussed prior

patrickxb: Yes, but we want to know that the user didn't choose USD, we are just using that as a default if they have nothing set.

patrickxb: That way we can change the default to EUR or whatever and it will work fine.

PR #13977, opened 2018-09-28
fix text css 2

chrisnojima:
@keybase/react-hackers this fixes a couple of missing css styles. the root cause of this is cause the text storybook wasn't updated when we added some styles. i just added a new programatic section which lists them all. in order to support this i added an object that captures TextType so we can reference that in code and uses $Keys in flow


chrisnojima:
@buoyad fixed

PR #13980, opened 2018-09-28
Add currentWorth to stellar1.PaymentLocal, stellar1.PaymentDetailsLocal

patrickxb:


go/stellar/transform.go
```diff
@@ -311,5 +337,18 @@ func newPaymentLocal(txID stellar1.TransactionID, ctime stellar1.TimeMs, amount
...
+		outsideAmount, err := stellarnet.ConvertXLMToOutside(amount, exchRate.Rate)
+		if err != nil {
+			return nil, err
+		}
+		currentWorth, err := FormatCurrency(mctx.Ctx(), mctx.G(), outsideAmount, exchRate.Currency)
+		if err != nil {
+			return nil, err
+		}
+		loc.CurrentWorth = currentWorth
+		loc.CurrentWorthCurrency = string(exchRate.Currency)
```
mlsteele: i suspect that this conversion should not be critical to this function and that the frontend should be able to handle when they're missing

patrickxb: that's my assumption as well

mlsteele: and yet here we are with `return nil, err`

patrickxb: I can change it...the conversion and format errors i thought would be more our coding fault as opposed to a network error, so surfacing the error would be ok because we would catch it.

PR #13981, opened 2018-09-28
Check that buildingPayment hasn't changed before using builtPayment result

buoyad:
This stashes the input to `buildPaymentLocal` in `BuiltPaymentReceivedPayload` and checks that it hasn't changed in the reducer before accepting the output. r? @keybase/react-hackers


shared/reducers/wallets.js
```diff
@@ -17,10 +17,12 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
...
       return state.setIn(['assetsMap', action.payload.accountID], I.List(action.payload.assets))
     case WalletsGen.builtPaymentReceived:
-      return state.set(
-        'builtPayment',
-        state.get('builtPayment').merge(Constants.makeBuiltPayment(action.payload.build))
-      )
+      return action.payload.forBuildingPayment === state.buildingPayment
+        ? state.set(
+            'builtPayment',
+            state.get('builtPayment').merge(Constants.makeBuiltPayment(action.payload.build))
```
chrisnojima: can just be `state.builtPayment.merge`

PR #13983, opened 2018-09-28
update to stellarnet for new amount parsing consolidation

xgess:
integrate https://github.com/keybase/stellarnet/pull/26, a new, more strict, more consolidated API for converting stellar amounts between strings and int64s and xdr.Int64s. And I'll revendor once that stellarnet PR is merged. 

Also, it looks like stellarnet hasn't been revendored in a while, so there are some other changes in here too, but they look pretty additive / safe to me.


patrickxb:
could you add a test for the exponential parsing?

xgess:
@patrickxb done

PR #13985, opened 2018-09-28
Fix view of permanently errored messages sent by current user

buoyad:
This fixes something I noticed in #13972.
image
Both these errors are permanent, but since the second one was sent by me the message thinks we can retry. This adds an additional check on `submitState` to show the cancel/retry buttons. Messages that failed in the outbox are `pending`, and permanent errors have no `submitState`. r? @keybase/react-hackers


mmaxim:
I guess the check for text or attachment is why the errors for your delete didn't show up.

PR #13993, opened 2018-09-28
create/link account in the same layer from SendForm

songgao:
As a bonus point, now if user creates a new account or link existing, from the SendForm, after it's done it'd automatically select the newly added account.

It's still kind of messy, but it works! I'd be happy to clean it up but I'd like to have some eyes on it to make sure I'm on the right track. @chrisnojima @buoyad


chrisnojima:
@cjb or @buoyad should take a look as they've been in this code more than i have

shared/wallets/link-existing/index.js
```diff
@@ -56,6 +56,13 @@ class LinkWallet extends React.Component<LinkWalletProps, LinkWalletState> {
     if (this.props.nameValidationState === 'valid' && this.state.view === 'name') {
       this.props.onClearErrors()
       this.props.onDone()
+      // This is for when we are showing this from a SendForm.
+      //
+      // songgao: Note that we are switching back to the root of SendForm
+      // before we have finished creating the account, but the critical point
```
buoyad: Could we pipe a spinner to the "to" dropdown while a new account is being created?

songgao:
@buoyad thanks for looking! Please take another look when you get a chance. The tricky part was that the spinner needs to show up until 1) `setBuildingTo` is done; and 2) `loadAccounts` has finished so the new account ID is in there. I started with trying to sequencing both in Saga but it felt weird, so I ended up checking in container if the `toAccount` exists, and show the spinner if not.

songgao:
Bump!

shared/wallets/send-form/body/index.js
```diff
@@ -20,17 +22,19 @@ const Spinner = () => (
...
-    {isProcessing && <Spinner />}
-    {bannerInfo && <Banner />}
-    <Participants />
-    <AssetInput />
-    <Divider />
-    <NoteAndMemo />
-    <Footer isRequest={isRequest} />
-  </Box2>
-)
+const Body = (props: Props) => {
```
buoyad: Could still use the shorthand here

shared/wallets/send-form/participants/dropdown.js
```diff
@@ -6,24 +6,32 @@ import {AccountEntry} from '../../common'
...
 
 // The display of the selected account in the dropdown.
-export const SelectedEntry = ({account, user, ...props}: SelectedEntryProps) => (
+export const SelectedEntry = ({account, spinner, user, ...props}: SelectedEntryProps) => (
   <Kb.Box2 {...props} direction="horizontal" centerChildren={true} gap="tiny" fullWidth={true}>
+    {spinner && (
+      <Kb.Icon style={Kb.iconCastPlatformStyles(styles.spinner)} type="icon-progress-grey-animated" />
+    )}
     <Kb.Avatar size={16} username={user} />
     <Kb.Text type="BodySemibold" style={styles.text}>
```
buoyad: This flashes `NOACCOUNTID` while the spinner is up, could we hide it?

shared/wallets/send-form/participants/container.js
```diff
@@ -23,6 +23,7 @@ const mapStateToProps = (state: TypedState) => {
     .toArray()
   let fromAccount
   let toAccount
+  let showSpinner = state.wallets.waitingForSetBuildingTo
```
buoyad: `Constants.linkExistingWaitingKey` and `Constants.createNewAccountWaitingKey` exist, could this instead be an `anyWaiting` on those keys and the one for `loadAccounts`?

buoyad: I'm a little worried about having `waitingForSetBuildingTo` as a separate flag. If one of the RPCs in the createAccount -> setBuildingTo fails it could get into the wrong state

shared/wallets/send-form/index.js
```diff
@@ -1,54 +1,76 @@
...
+  }
+  render() {
+    switch (this.state.currentScreen) {
+      case 'root':
+        return (
+          <SendFormRoot
+            isRequest={this.props.isRequest}
+            onClose={this.props.onClose}
+            onLinkAccount={this.linkExisting}
+            onCreateNewAccount={this.createNewAccount}
```
buoyad: After creating a new account it gets selected in the UI behind the send form, but the account details never load on closing the form. It should load the details, but I think it'd also be fine to not select the newly created account when created from the send form

shared/wallets/send-form/participants/container.js
songgao: Will do; I guess I just didn't read enough about existing types.

shared/wallets/send-form/index.js
songgao: Hmm sorry I'm not sure I understand correctly. The new account is selected because of [this action](https://github.com/keybase/client/pull/13993/files#diff-314fc39c85ff1807227a38a2848d7851R283) which is dispatched when [created or linked](https://github.com/keybase/client/blob/9d4c6568298b6ba71fd473f404b7bd7d6c18143c/shared/actions/wallets.js#L417) happens, which also triggers `loadAccounts` [here](https://github.com/keybase/client/blob/9d4c6568298b6ba71fd473f404b7bd7d6c18143c/shared/actions/wallets.js#L381). So it seems to me if the form is closed early, we wouldn't select. If we select, it must have been a successful creation or linking, which would also definitely trigger loading account details.

buoyad: Yeah, that all makes sense. I just mean when it's selected in the background, the assets / tx history in the background perma-spinner even after closing the dialog:
dev_keybase

Just checked and this happens on master too, we can leave it for another ticket if you want.

songgao:
🎉

PR #13999, opened 2018-10-01
Show all channels in chat mention hud

buoyad:
Adds a loading state to the channel mention hud and a call to `TeamsGen.getChannels` on mount if we haven't loaded it before. The teams `teamNameToChannelInfos` doesn't have any cache busting, so this can get out of date if channels are removed/created. I'll make tickets for a notification for this and we can decide how to refresh the data / integrate with this PR. r? @keybase/react-hackers 
channelhudload


chrisnojima:
we discussed IRL. i think we need to just have this info and sync and not try and load this on starting up the hud

PR #14006, opened 2018-10-01
stricter checks for version numbers

maxtaco:
- check that payload JSON version matches the outer link version for v=2
- check that what is actually signed matches JSON or Msgpack by looking at the first byte


maxtaco:
I ran a script over the prod data, all non-PGP signatures pass this check

maxtaco:
@mlsteele I experimented a bunch, both today and yesterday, with not setting `tmp.sigVersion` until it's verified against the payload, but that is a big refactor of the code.  I'd prefer to keep this PR small.

PR #14014, opened 2018-10-01
Limit length of publicMemo and secretNote

buoyad:
Adds hard length limits to secretNote and publicMemo input fields and lines showing how many characters are remaining.
image
Lengths are hardcoded to 500 and 28, lifted from [here](https://github.com/keybase/client/blob/b6b7470/go/stellar/stellarsvc/frontend.go#L953-L968). r? @keybase/react-hackers


shared/wallets/send-form/note-and-memo/index.stories.js
```diff
@@ -4,17 +4,34 @@ import * as Sb from '../../../stories/storybook'
...
     ))
     .add('Input with a memo error', () => (
-      <NoteAndMemo {...commonProps} memoError="There's something wrong with your memo." />
+      <NoteAndMemo {...commonProps} publicMemoError="There's something wrong with your memo." />
+    ))
+    .add('Prefilled', () => (
+      <NoteAndMemo
+        {...commonProps}
+        secretNote="This is a prefilled secret note"
+        publicMemo="This is a prefilld pblc memo"
```
cjb: thes wods pelt rong :)

buoyad: hah yeah, i was surprised by how short the public memo limit is in practice

PR #14017, opened 2018-10-01
fix Fs badge position on mobile and some story improvements

songgao:
This is actually way smaller than it seems. Each commit should be individually reviewable. The first one (b77cf7c) is probably the largest but it's mostly mechanical changes.

### DIFF
--- a/shared/fs/common/index.stories.js
+++ b/shared/fs/common/index.stories.js
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react'
 import * as Constants from '../../constants/fs'
+import {isMobile} from '../../constants/platform'
 import * as Sb from '../../stories/storybook'
 import * as Styles from '../../styles'
 import * as Types from '../../constants/types/fs'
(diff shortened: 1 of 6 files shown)


shared/fs/common/index.stories.js
```diff
@@ -20,18 +21,25 @@ const pathItemActionPopupProps = (path: Types.Path) => {
...
-        title: 'menu item',
-        onClick: Sb.action('onClick'),
-      },
-    ],
+    ...(isMobile
+      ? {
+          saveMedia: Sb.action('saveMedia'),
+          shareNative: Sb.action('shareNative'),
+        }
+      : {}),
```
jzila: what happened to `menuItems`?

shared/fs/row/common.js
```diff
@@ -73,8 +73,8 @@ const pathItemActionIcon = {
 
 const badgeContainer = {
   position: 'absolute',
-  left: Styles.isMobile ? -24 : 24,
-  top: Styles.isMobile ? -20 : -1,
+  left: Styles.isMobile ? -28 : 24,
+  top: Styles.isMobile ? -4 : -1,
```
jzila: did you test this on both iOS and Android?

shared/fs/common/index.stories.js
songgao: They were gone long ago; we now construct them in the component. Container still decides whether an item shows up in the menu, but by either populating an action or not, instead of constructing menu items directly.

shared/fs/row/common.js
songgao: I tested on iOS. You mean Android would have different layout?

jzila: I'm not sure. I originally came up with these offsets via trial and error on the iOS simulator, so I'm not sure how they could be wrong.

songgao: Perhaps they were just affected by some other changes. Anyway I'm building android now and will try those stories on Android

songgao:
As mentioned on chat, I'm merging without testing separately on Android since I can't get it running with some moderate effort. That is, Android won't look any different than iOS.

PR #14018, opened 2018-10-01
basic search with indexer

joshblum:
basic search implementation using the indexer. left some notes for potential optimizations but would like to get profiling setup so we can compare and decide where to focus our efforts.

vendors in the library: https://github.com/deckarep/golang-set for the set functionality


go/chat/search/indexer.go
```diff
@@ -157,13 +176,245 @@ func (i *Indexer) Remove(ctx context.Context, convID chat1.ConversationID, uid g
...
+	if opts.AfterContext > MaxContext || opts.AfterContext < 0 {
+		opts.AfterContext = MaxContext
+	}
+
+	tokens := tokenize(query)
+	if tokens == nil {
+		return nil, nil
+	}
+	// used to highlight the matching text.
+	queryRe, err := regexp.Compile(strings.Join(tokens, ".*"))
```
joshblum: This won’t work as is, will fix tomorrow

PR #14019, opened 2018-10-01
Add an optional IdentifyBehavior parameter to SimpleFSStat

jzila:


jzila:
Circle will fail until keybase/kbfs#1834 is in.

PR #14020, opened 2018-10-01
[KBFS-3446] Allow specifying identifyBehavior for filePreview

jzila:


jzila:
@songgao yep, same default as before.

PR #14028, opened 2018-10-02
upak-lite-based teamloader

heronhaye:
Almost entirely just shuffling code around and fighting go's "type system". Possibly I made some bad design decisions wrt interfaces, let me know! I made an attempt to get away with UPAKLites in other places too if possible, could expand that with some more work.

todo
- [x] fix old tests
- [x] test that if upaklite loader fails, should fall back to full loader and still work in LoadKey
- [x] test for equivalence between full and lite FindKID
- [x] test fails to find lowkey when specified

### DIFF
--- a/go/client/cmd_upak.go
+++ b/go/client/cmd_upak.go
@@ -82,7 +82,7 @@ func (c *cmdUPAK) Run() error {
 		if v != keybase1.UPAKVersion_V2 {
 			return fmt.Errorf("didn't get UPAK v2")
 		}
-		upk2, _ := res.V2().FindKID(c.kid)
+		upk2, _ := libkb.FindKID(res.V2(), c.kid)
 		if upk2 == nil {
 			return fmt.Errorf("key %s wasn't found", c.kid)
 		}
--- a/go/libkb/upak_loader.go
+++ b/go/libkb/upak_loader.go
@@ -20,14 +20,16 @@ type UPAKLoader interface {
 	Load(arg LoadUserArg) (ret *keybase1.UserPlusAllKeys, user *User, err error)
 	LoadV2(arg LoadUserArg) (ret *keybase1.UserPlusKeysV2AllIncarnations, user *User, err error)
 	LoadLite(arg LoadUserArg) (ret *keybase1.UPKLiteV1AllIncarnations, err error)
+	LoadV2OrLite(arg LoadUserArg) (ret *keybase1.UPKLiteAllIncarnationsInterface, user *User, err error)
 	CheckKIDForUID(ctx context.Context, uid keybase1.UID, kid keybase1.KID) (found bool, revokedAt *keybase1.KeybaseTime, deleted bool, err error)
 	LoadUserPlusKeys(ctx context.Context, uid keybase1.UID, pollForKID keybase1.KID) (keybase1.UserPlusKeys, error)
-	LoadKeyV2(ctx context.Context, uid keybase1.UID, kid keybase1.KID) (*keybase1.UserPlusKeysV2, *keybase1.UserPlusKeysV2AllIncarnations, *keybase1.PublicKeyV2NaCl, error)
+	LoadKey(ctx context.Context, uid keybase1.UID, kid keybase1.KID, includeLowKeys bool) (*keybase1.UPKLiteInterface, *keybase1.UPKLiteAllIncarnationsInterface, *keybase1.PublicKeyV2NaCl, error)
 	Invalidate(ctx context.Context, uid keybase1.UID)
 	LoadDeviceKey(ctx context.Context, uid keybase1.UID, deviceID keybase1.DeviceID) (upak *keybase1.UserPlusAllKeys, deviceKey *keybase1.PublicKey, revoked *keybase1.RevokedKey, err error)
 	LoadUPAKWithDeviceID(ctx context.Context, uid keybase1.UID, deviceID keybase1.DeviceID) (*keybase1.UserPlusKeysV2AllIncarnations, error)
 	LookupUsername(ctx context.Context, uid keybase1.UID) (NormalizedUsername, error)
 	LookupUsernameUPAK(ctx context.Context, uid keybase1.UID) (NormalizedUsername, error)
+	LookupUsernameUPAKLite(ctx context.Context, uid keybase1.UID) (NormalizedUsername, error)
 	LookupUID(ctx context.Context, un NormalizedUsername) (keybase1.UID, error)
 	LookupUsernameAndDevice(ctx context.Context, uid keybase1.UID, did keybase1.DeviceID) (username NormalizedUsername, deviceName string, deviceType string, err error)
 	ListFollowedUIDs(ctx context.Context, uid keybase1.UID) ([]keybase1.UID, error)
@@ -252,6 +254,26 @@ func (u *CachedUPAKLoader) PutUserToCache(ctx context.Context, user *User) error
 	return err
 }
 
+func (u *CachedUPAKLoader) LoadV2OrLite(arg LoadUserArg) (*keybase1.UPKLiteAllIncarnationsInterface, *User, error) {
+	var upak keybase1.UPKLiteAllIncarnationsInterface
+	var user *User
+	if arg.upakLite {
+		upakLite, err := u.LoadLite(arg)
+		if err != nil || upakLite == nil {
+			return nil, nil, err
+		}
+		upak = *upakLite
+	} else {
+		upakV2, u, err := u.LoadV2(arg)
+		if err != nil || upakV2 == nil {
+			return nil, nil, err
+		}
+		upak = *upakV2
+		user = u
+	}
+	return &upak, user, nil
+}
+
 func (u *CachedUPAKLoader) LoadLite(arg LoadUserArg) (*keybase1.UPKLiteV1AllIncarnations, error) {
 	m, tbs := arg.m.WithTimeBuckets()
 	arg.m = m
@@ -471,6 +493,7 @@ func (u *CachedUPAKLoader) LoadV2(arg LoadUserArg) (*keybase1.UserPlusKeysV2AllI
 	m, tbs := arg.m.WithTimeBuckets()
 	arg.m = m
 	defer tbs.Record("CachedUPAKLoader.LoadV2")()
+
 	return u.loadWithInfo(arg, nil, nil, true)
 }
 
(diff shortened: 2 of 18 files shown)


go/client/cmd_upak.go
```diff
@@ -82,7 +82,7 @@ func (c *cmdUPAK) Run() error {
 		if v != keybase1.UPAKVersion_V2 {
 			return fmt.Errorf("didn't get UPAK v2")
 		}
-		upk2, _ := res.V2().FindKID(c.kid)
+		upk2, _ := libkb.FindKID(res.V2(), c.kid)
```
heronhaye: needed to do this because go doesn't let you implement a "default method" for an interface

go/libkb/upak_loader.go
```diff
@@ -252,6 +254,32 @@ func (u *CachedUPAKLoader) PutUserToCache(ctx context.Context, user *User) error
 	return err
 }
 
+func (u *CachedUPAKLoader) LoadV2OrLite(arg LoadUserArg) (*keybase1.UPKLiteAllIncarnationsInterface, *User, error) {
```
heronhaye: this function is terrible. couldn't figure out how to make it better.

go/libkb/upak_loader.go
```diff
@@ -525,11 +554,13 @@ func (u *CachedUPAKLoader) LoadUserPlusKeys(ctx context.Context, uid keybase1.UI
...
 }
 
-// LoadKeyV2 looks through all incarnations for the user and returns the incarnation with the given
+// LoadKey looks through all incarnations for the user and returns the incarnation with the given
 // KID, as well as the Key data associated with that KID. It picks the latest such
 // incarnation if there are multiple.
-func (u *CachedUPAKLoader) LoadKeyV2(ctx context.Context, uid keybase1.UID, kid keybase1.KID) (ret *keybase1.UserPlusKeysV2,
-	upak *keybase1.UserPlusKeysV2AllIncarnations, key *keybase1.PublicKeyV2NaCl, err error) {
+// It first tries to load a UPAKLite, but if this fails, fallsback to full UPAK loader
+func (u *CachedUPAKLoader) LoadKey(ctx context.Context, uid keybase1.UID, kid keybase1.KID) (ret *keybase1.UPKLiteInterface,
```
heronhaye: (the important part - loading a uid/kid for a team sigchain link)

go/protocol/keybase1/extras.go
```diff
@@ -2625,3 +2611,59 @@ func (se *SelectorEntry) UnmarshalJSON(b []byte) error {
...
+	return u.DeviceKeys
+}
+
+func (u UpkLiteV1AllIncarnations) GetCurrent() UPKLiteInterface {
+	return u.Current
+}
+
+func (u UpkLiteV1AllIncarnations) GetPastIncarnations() []UPKLiteInterface {
+	ret := make([]UPKLiteInterface, len(u.PastIncarnations))
+	for idx, incarnation := range u.PastIncarnations {
```
heronhaye: also pretty annoying, but in Go if Foo implements FooInterface, []Foo is still not a []FooInterface

go/teams/loader.go
```diff
@@ -823,7 +823,7 @@ func (l *TeamLoader) userPreload(ctx context.Context, links []*ChainLinkUnpacked
 			for uid := range uidSet {
 				wg.Add(1)
 				go func(uid keybase1.UID) {
-					_, _, err := l.G().GetUPAKLoader().LoadV2(
+					_, err := l.G().GetUPAKLoader().LoadLite(
```
heronhaye: Only preload lite chains. Not sure if we want admins to preload the entire chain, but then I guess we don't know if we're an admin before loading.

go/libkb/upak_loader.go
```diff
@@ -20,14 +20,16 @@ type UPAKLoader interface {
 	Load(arg LoadUserArg) (ret *keybase1.UserPlusAllKeys, user *User, err error)
 	LoadV2(arg LoadUserArg) (ret *keybase1.UserPlusKeysV2AllIncarnations, user *User, err error)
 	LoadLite(arg LoadUserArg) (ret *keybase1.UpkLiteV1AllIncarnations, err error)
+	LoadV2OrLite(arg LoadUserArg) (ret *keybase1.UPKLiteAllIncarnationsInterface, user *User, err error)
```
heronhaye: could have a better name

go/libkb/upak_loader.go
```diff
@@ -525,11 +554,14 @@ func (u *CachedUPAKLoader) LoadUserPlusKeys(ctx context.Context, uid keybase1.UI
 	return up, nil
 }
 
-// LoadKeyV2 looks through all incarnations for the user and returns the incarnation with the given
+// LoadKey looks through all incarnations for the user and returns the incarnation with the given
 // KID, as well as the Key data associated with that KID. It picks the latest such
 // incarnation if there are multiple.
-func (u *CachedUPAKLoader) LoadKeyV2(ctx context.Context, uid keybase1.UID, kid keybase1.KID) (ret *keybase1.UserPlusKeysV2,
-	upak *keybase1.UserPlusKeysV2AllIncarnations, key *keybase1.PublicKeyV2NaCl, err error) {
+// It first tries to load a UPAKLite, but if this fails, fallsback to full UPAK loader
```
xgess: this isn't true anymore

go/teams/loader.go
xgess: do we actually want to start doing this?

go/teams/member_set.go
```diff
@@ -134,6 +134,17 @@ func loadUPAK2(ctx context.Context, g *libkb.GlobalContext, uid keybase1.UID, fo
...
 }
 
+func loadUPAKLite(ctx context.Context, g *libkb.GlobalContext, uid keybase1.UID, forcePoll bool) (ret *keybase1.UpkLiteV1AllIncarnations, err error) {
+	defer g.CTrace(ctx, fmt.Sprintf("loadUPAKLite(%s)", uid), func() error { return err })()
+
+	arg := libkb.NewLoadUserArg(g).WithNetContext(ctx).WithUID(uid).WithPublicKeyOptional()
+	if forcePoll {
+		arg = arg.WithForcePoll(true)
+	}
+	upak, err := g.GetUPAKLoader().LoadLite(arg)
```
xgess: i can't remember exactly, but i dont think the lite loading really respects the forcePoll request. or maybe just not as much.

xgess:
this feels like a half-rollout. in not a great way. we should definitely talk to @maxtaco about it, but my instinct is that we should get this to a state where it works, it's protected by some explicit tests, but it's not actually being used by anything in prod. and that will let us put it down temporarily without doing any additional caching and performance work until we're ready. with that in mind, i'd suggest we change all of the existing tests and behavior to `includeLowKeys = true` (behaviorally what it is before this PR). and then add another couple of tests that verify equivalence of the things we care about between the full and temporary loads across the interface (e.g. LoadKey yields the same device key for current and past incarnations, .... this should probably even be very easy if you do it in upak_loader_lite_test.go). it makes me nervous that some of the existing tests were swapped to `includeLowKeys = false` and some were kept at true. that may give us some confidence that this works as expected, but it exposes some additional risk that our tests and our code are deviating elsewhere. 

and when we're ready to start cutting things onto LoadLite, we can do the code and the corresponding tests together. or if we decide this was all a huge mistake because we don't need it or it introduces some new risk, it'll be easier to rip it all out together.

@heronhaye what do you think?

go/teams/member_set.go
heronhaye: I don't think it does at all until we do caching

heronhaye:
Yes, that was the plan (to merge this in with LoadKey always doing a full UPAK load). I agree, the old tests should be switched back to `includeLowKeys=true` for now even if they don't need it.

maxtaco:
Let’s discuss before merge. On Friday

maxtaco:
Where is it being feature-flagged off?

heronhaye:
It isn't right now, I kind of wanted to make sure team loads work in all the other tests. To disable lite loading, we can either ignore the includeLowKeys flag in LoadKey and always do a full load or flip the flag at https://github.com/keybase/client/pull/14028/files#diff-597f9f77e0dda25d828b4bbb4767fcbeR45.

I prefer that we do both.

maxtaco:
CAn we guard all of the upak-lite based loading with a runmode != Prod flag?

go/libkb/upak_loader.go
```diff
@@ -544,34 +570,64 @@ func (u *CachedUPAKLoader) LoadKeyV2(ctx context.Context, uid keybase1.UID, kid
 	// It should be that a ForcePoll is good enough, but in some rare cases,
 	// people have cached values for previous pre-reset user incarnations that
 	// were incorrect. So clobber over that if it comes to it.
-	attempts := []LoadUserArg{
-		argBase,
-		argBase.WithForcePoll(true),
-		argBase.WithForceReload(),
+	var attempts []LoadUserArg
+	if includeLowKeys {
```
maxtaco: `|| u.G().Env.GetRunMode() == RunModProd`

go/libkb/upak_loader.go
```diff
@@ -682,6 +738,16 @@ func (u *CachedUPAKLoader) LookupUsernameUPAK(ctx context.Context, uid keybase1.
 	return ret, err
 }
 
+// LookupUsername should use a map later on when caching is done, but for now it doesn't.
+func (u *CachedUPAKLoader) LookupUsernameUPAKLite(ctx context.Context, uid keybase1.UID) (NormalizedUsername, error) {
```
maxtaco: does this need to be feature-flagged off?

maxtaco:
I'm not in favor of doing a refactor of the existing code here, as suggested above.  I'd be OK to merge the code in so that it doesn't rot, as long as production doesn't change it all.  I'm also OK to leave tests on, as long as we're still testing what we run in prod. I.e., I don't want to test only UPAKLite in CI and run UPKs in prod.

maxtaco:
I think it if takes more than an hour to fix this up for a merge into master, then we should just keep the code in a branch.  I don't want to invest more time on Sigchain v2.3 right now, especially if that work might introduce security bugs.

heronhaye:
Nope, should be easy. So I'll make it so it's not possible to load a upak lite in prod. All tests have "master behavior" except for the ones I added which test LoadKey with UPAK lites.

maxtaco:
ok, thanks, i'll take a look when that's up

heronhaye:
Nope, should be easy. So I'll make it so it's not possible to load a upak lite in prod. All tests have "master behavior" except for the ones I added which test LoadKey with UPAK lites.

heronhaye:
uh, not sure what I did. Rebased and `git push -f`d but it close the PR for some reason.

heronhaye:
new pr at https://github.com/keybase/client/pull/14246...

PR #14030, opened 2018-10-02
7: dont render reactions row on no reactions

chrisnojima:
@keybase/react-hackers skip reactions row if no reactions

storybook thread:
base pr: 343ms
this pr: 256ms


chrisnojima:
yes they're all stacked

PR #14044, opened 2018-10-03
audit perf: feature-flag audits off during testing period

maxtaco:
We had a load problem caused by team audits today, and we'd like the
ability to toggle them off from the server-side while we load-test the
feature. We'll remove this feature flag shortly


maxtaco:
needs https://github.com/keybase/keybase/pull/2955

PR #14050, opened 2018-10-03
don't cache transient ephemeral errors

joshblum:
only cache ephemeral key errors if they are permanent (key is missing/unable to be unboxed). prevents transient (network/api) errors from being permanently cached. bumps the db version of `UserEKBoxStorage` and `TeamEKBoxStorage` so any existing bad values are retried.

### DIFF
--- a/go/ephemeral/user_ek_box_storage.go
+++ b/go/ephemeral/user_ek_box_storage.go
@@ -7,9 +7,12 @@ import (
 	"sync"
 	"time"
 
+	"github.com/keybase/client/go/erasablekv"
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/keybase1"
 )
+
+const userEKBoxStorageDBVersion = 3
 
 type userEKBoxCacheItem struct {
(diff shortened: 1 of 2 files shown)


go/ephemeral/user_ek_box_storage.go
```diff
@@ -100,17 +101,17 @@ func (s *UserEKBoxStorage) Get(ctx context.Context, generation keybase1.EkGenera
...
 	}
 
-	// We don't have anything in our cache, fetch from the server
-	s.Unlock() // release the lock while we fetch
-	return s.fetchAndStore(ctx, generation)
+	defer s.Unlock() // release the lock after we unbox
+	if cacheItem.HasError() {
+		return userEK, cacheItem.Error()
+	}
+	return s.unbox(ctx, generation, cacheItem.UserEKBoxed)
```
mlsteele: this didn't change behavior I think

joshblum: yeah i changed this so it matched the team storage more closely

PR #14052, opened 2018-10-03
Stellar: Show banners on send/confirm screens, show errors on sending

cjb:
@keybase/react-hackers


shared/wallets/confirm-form/container.js
```diff
@@ -7,21 +7,42 @@ import {connect, type TypedState} from '../../util/container'
...
     waitingKey: Constants.sendPaymentWaitingKey,
     yourUsername: state.config.username,
   }
 }
 
 const mapDispatchToProps = (dispatch, {navigateUp}) => ({
-  onBack: () => dispatch(navigateUp()),
-  onClose: () => dispatch(navigateUp()),
+  onBack: () => {
+    dispatch(WalletsGen.createClearErrors())
```
chrisnojima: another pattern we have is doing this in the saga. you have a saga listener for navigate up and if its a stellar path you'd clear the error. https://github.com/keybase/client/blob/master/shared/actions/provision.js#L517

shared/wallets/send-form/body/container.js
```diff
@@ -1,11 +1,21 @@
 // @flow
 import Body from '.'
 import {compose, connect, setDisplayName, type TypedState} from '../../../util/container'
+import {bannerLevelToBackground} from '../../../constants/wallets'
 
-const mapStateToProps = (state: TypedState) => ({})
+const mapStateToProps = (state: TypedState) => ({
+  banners: (state.wallets.builtPayment.banners || []).map(banner => ({
```
chrisnojima: likely better to do this in mergeProps so you don't thrash (even with the new empty array each time)

shared/wallets/confirm-form/container.js
cjb: Fixed via 03ef461

shared/wallets/send-form/body/container.js
cjb: Fixed via 4ba6b67

cjb:
@chrisnojima Thanks! Ready for another look.

PR #14054, opened 2018-10-03
no need to keep visiable state separately in WithTooltip anymore

songgao:
suggested by Danny


songgao:
Oh cool thanks for the heads up! If you can remove the `visible` in your PR, I'll close this one.

buoyad:
Ah, I merged my PR already. Want to do that with this one?

songgao:
Yup!

PR #14061, opened 2018-10-03
Show the account name in the export secret key dialog

cjb:
@keybase/react-hackers 

Show the account name (or the ID if that doesn't exist) in the export secret key dialog.


shared/wallets/export-secret-key/index.js
```diff
@@ -19,7 +22,11 @@ export default class ExportSecretKeyPopup extends React.Component<Props> {
   render() {
     const header = (
       <React.Fragment>
-        <Text type="BodySmallSemibold">{this.props.username}’s account</Text>
+        {this.props.walletName ? (
```
chrisnojima: all wallets have a name now right?

cjb: Yeah, I believe so, just thought I'd be safe, dunno if previously-created wallets are getting names.

PR #14063, opened 2018-10-03
Faster repeated buildPayment

mlsteele:
buildPayment was slower each time it was called repeatedly. This was mostly because the calls were stacking up in identify2 which wasn't using its cache since it's in resolveAndCheck mode.

Two fixes
- cancel all but the most recent buildPayment RPC
- add a 1 minute cache in front of LookupRecipient

context.Canceled errors are now exported with SCCanceled

Now LookupRecipient hits the cache for subsequent calls. There are still other remote RPCs buildPayment is hammering but hopefully we have another ticket for those.

`keybase script buildpayment` runs buildPayment a bunch of times quickly with some pauses and prints out timing.

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -2,6 +2,7 @@
 import * as Constants from '../constants/wallets'
 import * as Types from '../constants/types/wallets'
 import * as RPCTypes from '../constants/types/rpc-stellar-gen'
+import * as RPCTypesKb from '../constants/types/rpc-gen'
 import * as Saga from '../util/saga'
 import * as WalletsGen from './wallets-gen'
 import * as Chat2Gen from './chat2-gen'
@@ -15,6 +16,7 @@ import {walletsTab} from '../constants/tabs'
 import flags from '../util/feature-flags'
 import {getEngine} from '../engine'
 import {anyWaiting} from '../constants/waiting'
+import {RPCError} from '../util/errors'
 
 const buildPayment = (state: TypedState, action: any) =>
   RPCTypes.localBuildPaymentLocalRpcPromise({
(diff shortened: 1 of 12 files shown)


shared/actions/wallets.js
```diff
@@ -29,12 +31,20 @@ const buildPayment = (state: TypedState, action: any) =>
...
+        build: Constants.buildPaymentResultToBuiltPayment(build),
+        forBuildingPayment: state.wallets.buildingPayment,
+      })
+    )
+    .catch(error => {
+      if (error instanceof RPCError && error.code === RPCTypesKb.constantsStatusCode.sccanceled) {
+        // ignore cancellation
+      } else {
+        throw error
+      }
```
mlsteele: I get this (with a logger.info in place of the "ignore" comment above). Want to confirm that those values aren't used.

image

mlsteele:
@keybase/react-hackers [one file](https://github.com/keybase/client/pull/14063/files#diff-314fc39c85ff1807227a38a2848d7851) :)

shared/actions/wallets.js
```diff
@@ -2,6 +2,7 @@
 import * as Constants from '../constants/wallets'
 import * as Types from '../constants/types/wallets'
 import * as RPCTypes from '../constants/types/rpc-stellar-gen'
+import * as RPCTypesKb from '../constants/types/rpc-gen'
```
chrisnojima: to follow naming conventions it should be

```
import * as RPCStellarTypes from '../constants/types/rpc-stellar-gen'
import * as RPCTypes from '../constants/types/rpc-gen'
```

mlsteele:
Flipped names. Conflict prone 🤞

PR #14064, opened 2018-10-03
Change the shape of buttons and inputs.

adamjspooner:
@cecileboucheron, would you mind checking this branch out and reviewing it? I also reduced the border size on the reaction button because `2px` felt too large.


shared/teams/team/settings-tab/retention/index.js
```diff
@@ -35,7 +36,7 @@ export type Props = {
 }
 
 type State = {
-  saving: boolean,
+  sraving: boolean,
```
cecileboucheron: is this intended?

adamjspooner: Nope! Thanks for catching that.

adamjspooner:

PR #14071, opened 2018-10-04
Add --json mode to "keybase chat list-channels"

zapu:


go/client/cmd_chat_listchannels.go
```diff
@@ -52,19 +59,26 @@ func (c *CmdChatListChannels) Run() error {
...
 
-	ui.Printf("Listing channels on %s:\n\n", c.tlfName)
-	for _, c := range listRes.Convs {
-		convLine := fmt.Sprintf("#%s", c.Channel)
-		if c.Headline != "" {
-			convLine += fmt.Sprintf(" [%s]", c.Headline)
+	if c.json {
+		b, err := json.Marshal(listRes)
+		if err != nil {
+			return err
```
zapu: diff makes this really ugly but all that happened here is new `if c.json` branch, and rest moved to `else` block

mmaxim: Maybe it can just early out of the json block to avoid the indent?

PR #14073, opened 2018-10-04
clear avatar cache on user change

chrisnojima:
steps to repro on master

1. log in
1. view user A
1. switch to a different user
1. view user A, see a grey avatar

@keybase/react-hackers cc: @mmaxim


chrisnojima:
@keybase/react-hackers bump: 10 line change

shared/common-adapters/avatar.js
```diff
@@ -156,6 +165,7 @@ const _sharedAskForUserData = new SharedAskForUserData()
 
 const mapStateToProps = (state: TypedState, ownProps: OwnProps) => {
   const name = ownProps.username || ownProps.teamname
+  _sharedAskForUserData._checkLoggedIn(name)
```
mmaxim: This is kind of confusing to me, we could be passing in a team name to check against the current username?

chrisnojima: this is us checking if our own username has changed (aka we logged in / out)

chrisnojima: oh this got mutated, needs to be state.config.username

mmaxim: K that makes more sense, thanks!

chrisnojima: updated tnx!

PR #14076, opened 2018-10-04
Wallets on mobile part 1

chrisnojima:


chrisnojima:
@keybase/react-hackers this gets wallets working on mobile. its not all the way done but i don't want a huge pr since there's so many other tickets going on at the same time so lets get this going piece by piece

PR #14078, opened 2018-10-04
Stellar: cancel cancelable payments and consolidate paymentsMap and pendingMap

buoyad:
Adds `CancelPaymentLocal` RPC and hooks it up to cancel buttons on the tx list and details screens.
image
image
^ tooltip on hovering 'Cancelable'

Also consolidates `pendingMap` into `paymentsMap` to simplify lookups and remove some `status`s we were passing around as props. r? @keybase/react-hackers


shared/chat/conversation/messages/account-payment/container.js
```diff
@@ -30,16 +30,29 @@ const mapStateToProps = (state, ownProps: OwnProps) => {
...
+      if (paymentInfo.delta !== 'none') {
+        sign = paymentInfo.delta === 'increase' ? '+' : '-'
+      }
+      let balanceChangeColor = Styles.globalColors.black
+      if (paymentInfo.delta !== 'none') {
+        balanceChangeColor =
+          paymentInfo.delta === 'increase' ? Styles.globalColors.green2 : Styles.globalColors.red
+      }
+      if (pending) {
+        balanceChangeColor = Styles.globalColors.black_20
```
chrisnojima: we really should move this into some helper

shared/chat/conversation/messages/account-payment/container.js
```diff
@@ -30,16 +30,29 @@ const mapStateToProps = (state, ownProps: OwnProps) => {
...
+      }
       return {
-        action: paymentInfo.worth ? 'sent Lumens worth' : 'sent',
+        action: paymentInfo.worth ? `${verb} lumens worth` : verb,
```
chrisnojima: Lumens

shared/wallets/wallet/container.js
```diff
@@ -51,18 +68,7 @@ const mergeProps = (stateProps, dispatchProps) => {
...
-  if (paymentsMap.size === 0) {
-    return ['noPayments']
-  }
-  const payments = paymentsMap.toList()
-  return payments // payments always have a time
-    .sort((p1, p2) => (p2.time && p1.time && p2.time - p1.time) || 0)
-    .map(p => ({paymentID: p.id, status: p.statusSimplified}))
-    .toArray()
-})
+const sortAndStripTimestamps = (p: Array<{paymentID: Types.PaymentID, timestamp: number}>) =>
```
chrisnojima: we ok to lose the memoize?

buoyad: I think so, I don't notice any perf difference with it gone and we can always add it back in if we need it

buoyad:
thanks @chrisnojima, addressed feedback PTAL

PR #14080, opened 2018-10-04
Fix Windows release script

cjb:
r? @zanderz


zanderz:
I added 1 more commit so the target is amd64, is that ok? Matches https://github.com/keybase/keybase/pull/2940

PR #14087, opened 2018-10-04
Wrap jsonparser to give more informative error messages.

heronhaye:


go/jsonparserw/jsonparserw.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+
+	"github.com/buger/jsonparser"
+	"github.com/pkg/errors"
+)
+
+func wrapError(data []byte, keys []string, err error) error {
+	return errors.Wrap(err, fmt.Sprintf("jsonparserw error in data %s with keys %v", "[redacted]", keys))
```
maxtaco: i can't think of when we'd ever not redact, so maybe just keep the keys in there and get rid of `[redacted]`?

heronhaye: For devel. Also I figured we're only using jsonparser for public sigchains right now so it might be fine, but would need some better naming so we don't accidentally use it for sensitive data.

maxtaco: too dangerous!

PR #14090, opened 2018-10-04
don't force refresh when using FTL to lookup team names

maxtaco:
- The sticking point here was CI, which we hacked to pass
- Hack CI by just polling for the right update

### DIFF
--- a/go/chat/subteam_rename_test.go
+++ b/go/chat/subteam_rename_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/keybase/client/go/chat/types"
 	"github.com/keybase/client/go/chat/utils"
 	"github.com/keybase/client/go/protocol/chat1"
 	"github.com/keybase/client/go/protocol/keybase1"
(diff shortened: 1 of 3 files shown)


maxtaco:
CC: @mmaxim

go/chat/subteam_rename_test.go
```diff
@@ -143,11 +144,46 @@ func TestChatSubteamRename(t *testing.T) {
...
 		case <-time.After(2 * time.Second):
 		}
-		ib, err := tc.ChatG.InboxSource.Read(context.TODO(), users[0].User.GetUID().ToBytes(), nil, true, &chat1.GetInboxLocalQuery{
-			ConvIDs: u1ExpectedUpdates,
-		}, nil)
-		require.NoError(t, err)
-		require.True(t, len(ib.Convs) >= len(u1ExpectedUpdates))
+
+		pollIB := func() *types.Inbox {
+			ib, err := tc.ChatG.InboxSource.Read(context.TODO(), users[0].User.GetUID().ToBytes(), nil, true, &chat1.GetInboxLocalQuery{
```
joshblum: what happens if you try to send before the gregor notif comes in?

PR #14093, opened 2018-10-05
[KBFS-3433] get folders back to Profile

songgao:


shared/profile/folders/container.js
```diff
@@ -0,0 +1,120 @@
+// @flow
+import * as React from 'react'
+import * as FsGen from '../../actions/fs-gen'
+import * as FsTypes from '../../constants/types/fs'
+import flags from '../../util/feature-flags'
+import {compose, connect, setDisplayName} from '../../util/container'
+import Folders from './index'
```
chrisnojima: prefer `'.'`

shared/profile/folders/container.js
```diff
@@ -0,0 +1,120 @@
...
+    componentDidMount() {
+      this.props.loadTlfs()
+    }
+    render() {
+      return <ComposedComponent {...this.props} />
+    }
+  }
+
+export default compose(
+  flags.admin
```
chrisnojima: why have this be only admin

shared/profile/folders/container.js
```diff
@@ -0,0 +1,120 @@
...
+                dispatchProps._openInFilesTab(FsTypes.stringToPath(`/keybase/private/${tlfName}`)),
+              isPublic: false,
+              isSelf: false,
+              text: `private/${tlfName}`,
+            })),
+        ],
+  loadTlfs: dispatchProps.loadTlfs,
+})
+
+const TlfsLoadingHoc = (ComposedComponent: React.ComponentType<any>) =>
```
chrisnojima: why have a hoc, just make a simple component. its not going to be reused

shared/profile/folders/container.js
songgao: @cecileboucheron thought we should let admins play with it first, get a feel about it, before releasing to the public, as it might look bad and we might want to change layout. Should this use a separate flags as well?

songgao:
@chrisnojima ping! Do you want this under a different tag?

songgao:
Added the flag anyway. Please let me know if it should instead still be `admin`, and I'll revert the last commit before merging.

PR #14096, opened 2018-10-05
fix missing desktop import

chrisnojima:
@keybase/react-hackers native got this stuff changed but desktop still needs to export it. we could just not export this and have platform saga handle it (likely better) but this is fine for now

cc: @mmaxim


chrisnojima:
if you have a .js.flow file thats what flow 'sees'. if we (ever) get to move forward with the per-platform flow stuff i started it'd catch this and we'd ditch the .js.flow files

chrisnojima:
FYI hot server does catch this

PR #14097, opened 2018-10-05
lower wait time for service check

mmaxim:
These wait times guarantee we sit around for 900ms on every startup, we can lower it and likely be fine.


mmaxim:
It always wait the polling interval before doing anything, so we are stuck here for 900ms every time. You think checking every 100ms is too aggressive at startup? Doesn't seem like it to me.

PR #14100, opened 2018-10-05
profile search indexing cli

joshblum:
adds a dev only command to index the inbox or a conversation for performance testing.

seems like the main bottleneck was reading/writing the index for each message esp as the index grew so i added the `BatchAdd` functionality. dropped msg metadata from the index as well to further decrease the size, if a message matches we can apply the search filters when we pull the full message from storage.


go/chat/search/indexer.go
```diff
@@ -160,7 +154,7 @@ func (idx *Indexer) Add(ctx context.Context, convID chat1.ConversationID, uid gr
 // Remove tokenizes the message content and updates/removes index keys for each token.
 func (idx *Indexer) Remove(ctx context.Context, convID chat1.ConversationID, uid gregor1.UID,
 	msg chat1.MessageUnboxed) (err error) {
-	defer idx.Trace(ctx, func() error { return err }, "Indexer.Remove")()
```
joshblum: probably want a batch remove, but can see how hte convsource integration plays out

PR #14102, opened 2018-10-05
Phone numbers RPC

zapu:


go/client/commands_common.go
```diff
@@ -69,6 +69,7 @@ func GetCommands(cl *libcmdline.CommandLine, g *libkb.GlobalContext) []cli.Comma
 		NewCmdVerify(cl, g),
 		NewCmdVersion(cl, g),
 		newCmdWallet(cl, g),
+		NewCmdPhoneNumber(cl, g),
```
patrickxb: should this go in `devel` for now?

zapu: you're right, thanks

PR #14103, opened 2018-10-05
button for mobile to profile as well

joshblum:
we may not want to merge this at all or may want to hide it, but we can do some test builds with it


joshblum:
closing in favor of https://github.com/keybase/client/pull/14170

PR #14105, opened 2018-10-05
Rework copy text and fix crash

chrisnojima:
@keybase/react-hackers this makes the layout of copytext a little better and fixes the crash on click on mobile. im' making a small pr for just that to cherry pick as a hot fix as this is too big


chrisnojima:
@keybase/react-hackers

PR #14107, opened 2018-10-05
don't hit sig/get if we don't have to

maxtaco:
- the merkle tree might say our cached copy is up-to-date, and we would have been hitting the sig/get endpoint just to get a nil update, so eliminate the RT

### DIFF
--- a/go/teams/ftl.go
+++ b/go/teams/ftl.go
@@ -415,6 +415,7 @@ func stateHasKeys(m libkb.MetaContext, shoppingList *shoppingList, arg fastLoadA
 	if arg.NeedLatestKey && !state.LoadedLatest {
 		m.CDebugf("latest was never loaded, we need to load it")
 		shoppingList.needRefresh = true
+		shoppingList.needLatest = true
 		fresh = false
 	}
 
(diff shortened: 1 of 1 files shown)


maxtaco:
cc @mmaxim

go/teams/ftl.go
```diff
@@ -1102,16 +1121,22 @@ func makeState(arg fastLoadArg, s *keybase1.FastTeamData) *keybase1.FastTeamData
...
 
-	groceries, err := f.loadFromServerWithRetries(m, arg, shoppingList)
+	groceries, err := f.loadFromServerWithRetries(m, arg, state, shoppingList)
 	if err != nil {
 		return nil, err
 	}
 
+	if groceries == nil {
+		m.CDebugf("FastTeamChainLoader#refresh: our state was fresh according to the Merkle tree")
+		return nil, nil
```
mlsteele: why not return `state, nil`?

maxtaco: So the result isn't re-cached.

mlsteele: wicked fast

maxtaco: I used to be threading through a bool to reflect freshness, but this seemed a little cleaner

PR #14111, opened 2018-10-05
badge the devices tab for new_device messages from gregor

xgess:
1. update badge state with `new_device` messages from gregor (see: https://github.com/keybase/keybase/pull/2969)
2. use the count to badge the devices tab
3. dismiss those messages after going to the devices tab (this logic was copy-pasta from the git tab)


patrickxb:
Do you think once the user goes to the device tab, they should see which devices are new?

xgess:
@patrickxb maybe. i'd probably need help with that. or maybe it should be a desktop ticket. if we want to do that, i can beef this up to pass along more than just a count of the new devices. that's easy enough.

xgess:
just spoke to chris. i'm going to push a list of the new devices into the badge state instead of just a count. and then either remove the UI badging and unbadging logic or see if i can get a little `new` label to show up.

PR #14113, opened 2018-10-05
Provisioning fixes

chrisnojima:
- [x] better logging of responses
- [x] maybeCancelProvision used to submit a cancel and wait, but now it'll nav you all the way back
- [x] better logging when initial provision call finishes, marking singleton as done
- [x] final error page is always layered on top of root and not appended to so clicking back puts you back to the right place
- [x] use better cancel method. fixes clicking back on paperkey screen not exiting login flow

### DIFF
--- a/shared/constants/provision.js
+++ b/shared/constants/provision.js
@@ -8,7 +8,7 @@ import HiddenString from '../util/hidden-string'
 export const waitingKey = 'provision:waiting'
 
 // Do NOT change this. This is the value used by the daemon also so this way we can ignore it when they do it / when we do
-export const cancelDesc = 'kex canceled by caller'
+export const cancelDesc = 'Input canceled'
 
 export const makeState: I.RecordFactory<Types._State> = I.Record({
   codePageIncomingTextCode: new HiddenString(''),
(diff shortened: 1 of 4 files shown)


chrisnojima:
@keybase/react-hackers @cjb @patrickxb

shared/constants/provision.js
```diff
@@ -8,7 +8,7 @@ import HiddenString from '../util/hidden-string'
 export const waitingKey = 'provision:waiting'
 
 // Do NOT change this. This is the value used by the daemon also so this way we can ignore it when they do it / when we do
-export const cancelDesc = 'kex canceled by caller'
+export const cancelDesc = 'Input canceled'
```
cjb: confirm that the comment's still correct and the daemon changed their version?

chrisnojima: yes, works on pass and paperkey

patrickxb: we still have an error that has that string in it, but i'm not sure what `cancelDesc` is for.

patrickxb: kex2/base.go:var ErrCanceled = errors.New("kex canceled by caller")

chrisnojima: in my testing, we send this to cancel the login process when clicking back. it seems to function correctly on every step so i think we're good

PR #14114, opened 2018-10-05
Use getBalanceChangeColor in message menu

buoyad:
This is also calculated in `Transaction`, but that component uses `Role` and `pending` to figure it out. Those props are also used elsewhere and are derived from the same info `getBalanceChangeColor` uses, so I didn't think it was necessary to switch over. r? @keybase/react-hackers


shared/chat/conversation/messages/message-popup/payment/container.js
```diff
@@ -81,12 +81,15 @@ const sendMergeProps = (stateProps, dispatchProps, ownProps: SendOwnProps) => {
     }
   }
   const {_you: you} = stateProps
+  let sign = ''
```
chrisnojima: this should also likely go into some helper, we do it all over

PR #14128, opened 2018-10-07
fix conversation unread state in widget

songgao:
We were comparing only the `ConversationMeta` when determining whether to send over a conversation or not. However, unread and badging status are `unreadMap` and `badgeMap` respectively, rather than in `ConversationMeta`. So if the unread status is the only thing updated, we wouldn't send over the new thing.

Instead, plumb through `hasUnread` and `hasBadge` early so we we can inspect them in the remote serializer.

cc @patrickxb

### DIFF
--- a/shared/chat/inbox/container/remote.js
+++ b/shared/chat/inbox/container/remote.js
@@ -18,10 +18,10 @@ export type RemoteConvMeta = $Diff<
 >
 
 // To cache the list
-const valuesCached = memoize((state, ...vals) =>
+const valuesCached = memoize((badgeMap, unreadMap, ...vals) =>
   vals.map(v => ({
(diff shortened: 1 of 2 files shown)


songgao:
I think it's working properly. The only issue is it sends identical message twice for each "mark as read" event. I think this is because irrelevant fields in `ConversationMeta` changes, such as `trustedState` changing from `requesting` to `trusted`. Not sure how to eliminate that easily. Perhaps merge for now until this whole thing changes again?

songgao:
@chrisnojima ping!

chrisnojima:
In the serialize function you get old and new so you can send only if something has changed as far as the widget has concerned

songgao:
@chrisnojima PTAL; is this what you meant?

PR #14142, opened 2018-10-08
Make impteam with invite type:phone

zapu:


zapu:
@mlsteele I think that's rougly what's going to go in.

Refactoring `SocialAssertion` was a dead end because how it's used everywhere, not only in client but also in kbfs. So let's just leave it for now.

I added some support for distinguishing server-trust and client-verified assertions though in `libkb/assertion.go`. The idea is to recognize these and skip *identify* part when resolving+identifying.

go/libkb/assertion.go
```diff
@@ -281,6 +288,8 @@ func (a AssertionHTTP) CollectUrls(v []AssertionURL) []AssertionURL        { ret
 func (a AssertionHTTPS) CollectUrls(v []AssertionURL) []AssertionURL       { return append(v, a) }
 func (a AssertionDNS) CollectUrls(v []AssertionURL) []AssertionURL         { return append(v, a) }
 func (a AssertionFingerprint) CollectUrls(v []AssertionURL) []AssertionURL { return append(v, a) }
+func (a AssertionPhoneNumber) CollectUrls(v []AssertionURL) []AssertionURL { return append(v, a) }
+func (a AssertionPhoneNumber) IsServerTrust() bool                         { return true }
```
mlsteele: seems like this should cascade up and infect AND and ORs

zapu:
Looks like it's only for implicit teams: `verifyResolveResult` knows to skip `IsServerTrust` `ResolveResults`, but it's only called from `ResolveImplicitTeamDisplayName`.

zapu:
We can also change `ResolveImplicitTeamSetUntrusted` instead and make it do more work to resolve phone numbers, and leave the "main" resolve alone.

protocol/avdl/keybase1/common.avdl
```diff
@@ -292,23 +292,24 @@ protocol Common {
...
+    "phone" or "email").
+
+    If the assertion is for social network, resolving an assertion requires
     that the user posts a Keybase proof on the asserted service as the asserted
     user.
+
+    For server trust assertion, we have to trust the server.
+
+    TODO: SocialAssertion is historical name for this type, but it's not
+    correct since we do phone/email assertions as well.
```
mlsteele: is there a plan to get this todo todone?

mlsteele: ok seems like we need to decide one way or the other whether phone numbers can live here and not have this as a todo. might want to ask kbfs

go/libkb/assertion.go
mlsteele: oh this is a method on AssertionURL, not AssertionExpression as I had thought. nevermind

zapu: oh ok, I forgot to ask you about that. thanks!

protocol/avdl/keybase1/common.avdl
zapu: @maxtaco can you take a look?

The problem is that we had to shoehorn servertrust phone type invites into `SocialAssertion` record type, to avoid having to change any of the implicit team related types. Refactoring this would require bunch of changes in both client and KBFS.

maxtaco: It shouldn't be too bad to change the name of the type since it doesn't affect the wire format at all. And then KBFS can change their types independently.

strib: FYI CircleCI will likely fail if you change this without using type aliases, so it'll require a double-repo merge.  But it's fine with me!

zapu: should we do that as part of this PR or make a new one?

zapu: @mlsteele OK to merge and then work on a refactor?

Is there anything else I should fix in this PR? Thanks!

PR #14144, opened 2018-10-08
use metacontext instead of context for teams Storage

maxtaco:


go/teams/loader.go
```diff
@@ -434,6 +435,7 @@ func (l *TeamLoader) load2InnerLockedRetry(ctx context.Context, arg load2ArgT) (
 	ctx, tbs := l.G().CTimeBuckets(ctx)
 	tracer := l.G().CTimeTracer(ctx, "TeamLoader.load2ILR", teamEnv.Profile)
 	defer tracer.Finish()
+	mctx := libkb.NewMetaContext(ctx, l.G())
```
mlsteele: Things get tricky when both `ctx` and `mctx` are in scope. See L616. No bug yet. How about going all the way and making this function take `mctx` so that no `ctx` var is in scope?

maxtaco: omg that would take a lot of work

mlsteele: `s/ctx/mctx.Ctx()` and passthrough methods for `CTimeBuckets` and `CTimeTracer`

mlsteele: but sure another thing could be to not bind `mctx` and just do `ret = l.storage.Get(libkb.NewMetaContext(...)...)`

maxtaco: took the easy way out

PR #14145, opened 2018-10-08
notify for new and revoked devices

xgess:
* start receiving `device.new` and `device.revoked` messages from gregor (see: keybase/keybase#2969)
* push the relevant device ids into the badge state
* use the total number of unique device ids as a badge count on the devices tab
* dismiss those messages after leaving the devices tab
* toggle open the `Revoked devices` if there are any that are "new."

Here's a screenshot immediately after I created a device and revoked a different one.
screen shot 2018-10-08 at 5 57 57 pm


xgess:
would love any of your thoughts @keybase/react-hackers. thanks!

shared/devices/container.js
```diff
@@ -4,12 +4,14 @@ import * as DevicesGen from '../actions/devices-gen'
...
 import * as RouteTree from '../actions/route-tree'
 import * as Constants from '../constants/devices'
+import * as I from 'immutable'
 import {compose, connect, setDisplayName, safeSubmitPerMount} from '../util/container'
 import {partition} from 'lodash-es'
 
 const mapStateToProps = state => ({
   _deviceMap: state.devices.deviceMap,
   waiting: Constants.isWaiting(state),
+  newlyChangedItemIds: state.devices.getIn(['isNew'], []),
```
chrisnojima: we actually don't want the fallback item here, else it'll make a new array each time which causes this method to thrash. also if its not a prop thats going directly through mergeProps we should prefix with an underscore.
```
_newlyChangedItemIds: state.devices.isNew
```

shared/devices/container.js
```diff
@@ -40,6 +42,10 @@ type OwnProps = void
 
 function mergeProps(stateProps, dispatchProps, ownProps: OwnProps) {
   const [revoked, normal] = splitAndSortDevices(stateProps._deviceMap)
+  const revokedItems = revoked.map(deviceToItem)
+  const newlyRevokedIds = I.Set(revokedItems.map(d => d.key)).intersect(
+    I.Set(stateProps.newlyChangedItemIds || [])
```
chrisnojima: now you don't need to wrap this as its a I.Set already

shared/devices/index.js
```diff
@@ -49,6 +54,10 @@ class Devices extends React.PureComponent<Props, State> {
     )
 
   render() {
+    if (this.props.hasNewlyRevoked && !this.state.toggledExplicitly) {
```
chrisnojima: don't call setState in render. just initialize the state correctly on line 35. we then don't need this toggledExplicitly state

shared/devices/index.stories.js
```diff
@@ -27,6 +27,7 @@ const provider = Sb.createPropProviderWithCommon({
     firstItem: deviceID === '1',
     isCurrentDevice: deviceID === '1',
     isRevoked: !['1', '2', '3'].includes(deviceID),
```
chrisnojima: you likely want to add a story for this so we can see it and ensure it doesn't regress

shared/devices/row/container.js
```diff
@@ -26,16 +27,12 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => ({
...
 
-export default compose(
-  connect(
-    mapStateToProps,
-    mapDispatchToProps,
-    mergeProps
-  ),
-  setDisplayName('DeviceRow')
-)(DeviceRow)
+export default compose(connect(mapStateToProps, mapDispatchToProps, mergeProps), setDisplayName('DeviceRow'))(
```
chrisnojima: you likely don't have prettier installed?

shared/devices/container.js
xgess: 👍

shared/devices/container.js
xgess: 👍

shared/devices/row/container.js
xgess: lol. i thought it was prettier that added this to my diff.

shared/devices/index.js
xgess: i had trouble with this. the toggle looked like it was opening and then immediately closing when the component first rendered. :| 
i'll give it another shot.

shared/devices/index.stories.js
xgess: 👍

shared/devices/index.js
xgess: ok. i think i got it.

shared/devices/container.js
xgess: actually put it back to the un-underscored because i thought it would be better to have the devices component tell the device rows whether they're new instead of having each device row figure it out.

chrisnojima: but the prop is not emitted from mergeProps, therefore its really only for mapStateToProps to pass to mergeProps, so it should use an underscore. Are you thinking of a different prop? or am i missing something

shared/devices/row/container.js
chrisnojima: if you're using vscode the team has been complaining that its not doing the right thing re: prettier since the last update

xgess: i am. i can change this back to what it was. /shrug

shared/devices/container.js
chrisnojima: cool

shared/devices/index.js
```diff
@@ -35,18 +38,28 @@ class Devices extends React.PureComponent<Props, State> {
...
+      return (
+        <RevokedHeader
+          key="revokedHeader"
+          expanded={this.state.revokedExpanded}
+          onToggleExpanded={this._toggleExpanded}
+        />
+      )
+    } else {
+      const newItem = this.props.newlyChangedItemIds.has(item.id)
+      return <DeviceRow key={item.id} deviceID={item.id} firstItem={index === 0} isNew={newItem} />
```
chrisnojima: so the common pattern is to not plumb through this kind of thing to the rows. instead the row has its own connector and just just pull its own state out itself

xgess: oh. ok. i can do that.

xgess:
@mmaxim @chrisnojima i pushed up a couple of changes since yall approved. just wanna make sure those are green before merging:
1. pushed a bunch of dismissal logic into go. 
2. fixed a bug in `notifications.js` where a device could get an undismissable badge

PR #14148, opened 2018-10-08
WIP: Remove mouse handlers for hover styles.

adamjspooner:


adamjspooner:
@chrisnojima, better, worse, or about the same?

shared/styles/index.desktop.js
```diff
@@ -133,6 +133,14 @@ export const initDesktopStyles = () => {
     }
   }, '')
   style.appendChild(document.createTextNode(css))
+  style.appendChild(
```
chrisnojima: this is for us doing programmatically created css, if its static like this just make a .css file

shared/chat/inbox/row/avatars.js
```diff
@@ -19,37 +19,33 @@ function rowBorderColor(idx: number, isLastParticipant: boolean, backgroundColor
...
+  isMuted ? (
+    <Icon type={isSelected ? 'icon-shh-active-24' : 'icon-shh-24'} style={avatarIconStyle} />
+  ) : isLocked ? (
+    <Icon type={isSelected ? 'icon-addon-lock-active-22' : 'icon-addon-lock-22'} style={avatarIconStyle} />
+  ) : null
+
+const MutedIcon = ({isLocked, isMuted}) =>
+  isMuted ? (
+    <>
+      <Icon className="muted_icon" type="icon-shh-16" style={avatarIconStyle} />
```
chrisnojima: i don't think we want to render 3 icons and just hide them, you can have one icon. maybe let me make an example since there isn't one currently

adamjspooner: Thanks!

chrisnojima: so i started doing this but its a lot of code and its likely not worth it, i think we should keep what we had before

adamjspooner: So, just leave it with the mouse handlers?

chrisnojima: yeah. in order for this to be easier i think how we handle css needs to be better

adamjspooner: Sounds good. Thanks for the feedback!

PR #14149, opened 2018-10-08
client/protocol: SimpleFS RPC for Symlink

jakob223:
Fixes KBFS-3443. Needs to be merged in conjunction with the corresponding client PR.


go/client/cmd_simplefs_symlink.go
```diff
@@ -0,0 +1,74 @@
...
+// ParseArgv gets the required path argument for this command.
+func (c *CmdSimpleFSSymlink) ParseArgv(ctx *cli.Context) error {
+	nargs := len(ctx.Args())
+	if nargs != 2 {
+		return errors.New("ln requires exactly 2 arguments")
+	}
+
+	var err error
+	var sources []keybase1.Path
+	sources, c.dest, err = parseSrcDestArgs(c.G(), ctx, "ln")
```
strib: I'm not sure about running the `target` through this call.  It makes whatever path is there absolute, so the user wouldn't be able to make a relative symlink.  So for example, I was playing around and did this:
```sh
keybase$ KEYBASE_RUN_MODE=prod ./keybase fs ln ../tmp /keybase/private/strib/tmp/ln3
keybase$ ls -l /keybase/private/strib/tmp/ln3
lrwx------ 1 strib root 51 Oct  8 15:15 /keybase/private/strib/tmp/ln3 -> /home/strib/go/src/github.com/keybase/client/go/tmp
```

That seems pretty weird.  On the other hand, I can see an argument for it working that way, since you might be running the command from some directory and using tab-completion to figure out the right target.  However, I think that's outweighed by the fact that the user might inadvertently be exposing their local directory structure in the symlink.

So my opinion is that we should only be parsing the link part, and we should leave the target part alone.  And then, in the protocol, the target can just be a string, rather than a Path.  What do you think?

strib:
Also, the jenkins tests failed here because you need to update the simplefs mock used in the tests with the new method.

go/client/cmd_simplefs_symlink.go
```diff
@@ -0,0 +1,74 @@
...
+type CmdSimpleFSSymlink struct {
+	libkb.Contextified
+	src  keybase1.Path
+	dest keybase1.Path
+}
+
+// NewCmdSimpleFSSymlink creates a new cli.Command.
+func NewCmdSimpleFSSymlink(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	return cli.Command{
+		Name:         "ln",
```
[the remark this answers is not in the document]
jakob223: Yeah, ran it by Chris and Max, they were cool with it.

go/client/cmd_simplefs_symlink.go
jakob223: Yeah, I think making it a string makes sense. Will change to that.

go/client/cmd_simplefs_symlink.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/keybase/cli"
+	"github.com/keybase/client/go/libcmdline"
+	"github.com/keybase/client/go/libkb"
+	keybase1 "github.com/keybase/client/go/protocol/keybase1"
+)
+
+// CmdSimpleFSSymlink is the 'fs ln' command.
+type CmdSimpleFSSymlink struct {
+	libkb.Contextified
+	src  string
```
strib: I'd call these fields `target` and `link` for clarity.

protocol/avdl/keybase1/simple_fs.avdl
```diff
@@ -250,6 +250,12 @@ protocol SimpleFS {
    */
   void simpleFSCopy(OpID opID, Path src, Path dest);
 
+
+  /**
+   Make a symlink of file or directory
+   */
+  void simpleFSSymlink(string src, Path dest);
```
strib: Same here, `target` and `link`.

PR #14151, opened 2018-10-08
Fix blackbar when hitting back on device add

chrisnojima:
this fixes clicking back when adding a new device. 
Core sends us back 2 possible error values when we cancel so we need to ignore both. previously we ignored one, then switched to the other, now we handle both

### DIFF
--- a/shared/actions/__tests__/provision.test.js
+++ b/shared/actions/__tests__/provision.test.js
@@ -89,7 +89,7 @@ describe('provisioningManagerProvisioning', () => {
       expect(response.result).not.toHaveBeenCalled()
       expect(response.error).toHaveBeenCalledWith({
         code: RPCTypes.constantsStatusCode.scinputcanceled,
-        desc: Constants.cancelDesc,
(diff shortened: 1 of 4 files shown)


chrisnojima:

shared/actions/__tests__/provision.test.js
```diff
@@ -89,7 +89,7 @@ describe('provisioningManagerProvisioning', () => {
       expect(response.result).not.toHaveBeenCalled()
       expect(response.error).toHaveBeenCalledWith({
         code: RPCTypes.constantsStatusCode.scinputcanceled,
-        desc: Constants.cancelDesc,
+        desc: 'Input canceled',
```
buoyad: Can we keep this as a constant?

chrisnojima: i think its ok for this test to know about the value. i don't want to export the individual constants out

buoyad: Oh I misread this as the main action file. Sounds good

PR #14168, opened 2018-10-09
Position exploding meta.

adamjspooner:
@keybase/react-hackers, this is ready for review. It moves the exploding meta to the bottom of the container on mobile. It moves it to line up with the first line of the message when a username is present on desktop.


chrisnojima:
do you want some more padding by the ... in this state?
screen shot 2018-10-09 at 1 28 17 pm

adamjspooner:
Thanks, @chrisnojima. This is ready for another look.

PR #14170, opened 2018-10-09
Chat index profiling

chrisnojima:
less surface area than https://github.com/keybase/client/pull/14103
@joshblum


chrisnojima:
@keybase/react-hackers bump

PR #14173, opened 2018-10-09
Stellar: move amount err msg to input area and style send button box

buoyad:
Go side is adding a period at the end of this string: `"Your available to send is *%s XLM*."` 

r? @keybase/react-hackers


shared/wallets/send-form/available/index.js
```diff
@@ -1,26 +1,38 @@
...
-  <Box2 direction="vertical">
-    {!!props.amountErrMsg && (
-      <Text type="Body" style={styles.text}>
-        {props.amountErrMsg}
-      </Text>
-    )}
-  </Box2>
-)
+const Available = (props: Props) => {
+  // This will only work to apply one custom style.
```
buoyad: I couldn't find a way to tell `Markdown` that bold is supposed to become extrabold. I decided against extending it since we're moving to simple markdown.

shared/wallets/send-form/footer/index.js
```diff
@@ -13,13 +12,32 @@ type Props = {
...
+    >
       {!!props.worthDescription && (
-        <Kb.Text style={styles.worthDescription} type="BodySmall">
-          This is <Kb.Text type="BodySmallExtrabold">{props.worthDescription}</Kb.Text>
-        </Kb.Text>
+        <Kb.Box2 direction="horizontal">
+          <Kb.Text style={styles.worthDescription} type="BodySmall">
+            This is <Kb.Text type="BodySmallExtrabold">{props.worthDescription}</Kb.Text>.
+          </Kb.Text>
+          <Kb.Icon
```
chrisnojima: we should hide this if it doesn't actually work, just comment it out

PR #14175, opened 2018-10-09
param proof test update

joshblum:
test updates to go in once https://github.com/keybase/keybase/pull/2991


joshblum:
manually tested, merging so tests pass against new server images

PR #14179, opened 2018-10-10
Stellar: integrate account badges in the nav and account rows

cjb:
@keybase/react-hackers 

This links up with @akalin-keybase's PR #14120 -- this PR handles displaying account badges (badgeState written to a new `store.wallets.unreadPaymentsMap`), and that PR provides the mechanism for clearing them (`markAsRead` on the badged payments).

There's a difference in display between desktop and mobile: desktop uses an orange dot, while mobile badges the row with the number of unread payments.  The app itself is badged (on desktop only) with the total number of unread payments, plus the total number of chat messages etc.

Also some attribute alphabetizing.

### DIFF
--- a/shared/actions/json/wallets.json
+++ b/shared/actions/json/wallets.json
@@ -7,8 +7,7 @@
   ],
   "actions": {
     "abandonPayment": {
-      "_description":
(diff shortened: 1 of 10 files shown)


cjb:
Thanks for the review! Agree with both comments.

PR #14184, opened 2018-10-10
enforce note max length

joshblum:
enforce and test max length for encrypted notes


go/chat/msgchecker/constants.go
```diff
@@ -23,7 +23,7 @@ const (
 	BoxedSystemMessageBodyMaxLength         = 5000
 	BoxedDeleteHistoryMessageBodyMaxLength  = 200
 	BoxedSendPaymentMessageBodyMaxLength    = 200
-	BoxedRequestPaymentMessageBodyMaxLength = 500
+	BoxedRequestPaymentMessageBodyMaxLength = 600
```
patrickxb: I'm confused...why is this in chat?  Isn't this the maximum length of the encrypted note in a payment?

joshblum: fixed!

PR #14187, opened 2018-10-10
6-liner feature flag out-of-date banner

songgao:


chrisnojima:
what is the plan here? are you working on this and we'll turn it on or ?

songgao:
@chrisnojima Coyne requested to not release this until it's more responsive. Right now it just runs `keybase update check` behind the scene when "update now" is clicked. I have a ticket for next sprint to make it turn into something else and wait on the download until the update prompt pops up. Feature flagging is in case this gets released before that's done.

chrisnojima:
lets just add its own feature flag and not tie it to admin

chrisnojima:
lets just add its own feature flag and not tie it to admin

PR #14190, opened 2018-10-10
Resolve phone numbers + SBS

zapu:
- Add `func resolvePhoneNumberViaServerLookup` in `libkb/resolve.go` that will resolve phone numbers using `user/phone_numbers_search` endpoint.
- Add `TeamInviteCategory_PHONE` to take `// nothing to verify, need to trust the server` branch in SBS notification handler.
- Detect `ResolutionErrorRateLimited` error when resolving users during creation of implicit team and abort, preventing us from creating an implicit team with number that's potentially already claimed, but we can't know about this.


zapu:
cc @heronhaye

go/libkb/assertion.go
```diff
@@ -522,7 +522,7 @@ func (a AssertionSocial) ToLookup() (key, value string, err error) {
 }
 
 func (a AssertionPhoneNumber) ToLookup() (key, value string, err error) {
-	return "phone", a.Value, nil
+	return "phone", "+" + a.Value, nil
```
mlsteele: We probably want to do as much canonicalization as possible to phone numbers to avoid unnecessary weird conflicts. Not sure where or if this is already done. Wouldn't want different impteams with "1083475713" and "108 347 5713" and "108-347-5713".

zapu: It's done on the server side right now:
https://github.com/keybase/keybase/blob/master/lib/checkers.iced#L218-L226

mlsteele: Cool. Also the client should get the canonicalized version into impteam invitation sigs

go/teams/resolve.go
```diff
@@ -178,7 +194,7 @@ func verifyResolveResult(ctx context.Context, g *libkb.GlobalContext, resolvedAs
 	}
 
 	if resolvedAssertion.ResolveResult.IsServerTrust() && g.Env.GetRunMode() != libkb.ProductionRunMode {
-		g.Log.CDebugf(ctx, "Trusting the server on assertion: %q (server trust - no way for clients to verify)", resolvedAssertion.Assertion.String())
+		g.Log.CWarningf(ctx, "Trusting the server on assertion: %q (server trust - no way for clients to verify)", resolvedAssertion.Assertion.String())
```
mlsteele: This would mean users see log lines when attached to the CLI while the service is doing just about anything related to impteams with phone numbers wouldn't it?

zapu: good point

zapu: the idea was to see this warning when doing
```
 ~ » kb chat send "48784578263@phone" "hello"
▶ WARNING not exportable error: invalid team name: team names must be between 2 and 16 characters long (*errors.errorString)
Creating a new PRIVATE CHAT conversation
 ~ »
```

but this does not seem to be the case. Maybe let's just change it back to debug?

go/libkb/resolve.go
```diff
@@ -418,6 +425,62 @@ func (r *ResolverImpl) resolveTeamViaServerLookup(m MetaContext, au AssertionURL
...
+	arg.Args["phone_number"] = S{Val: val}
+	arg.AppStatusCodes = []int{SCOk}
+
+	var lookup phoneNumberLookup
+	if err := m.G().API.GetDecode(arg, &lookup); err != nil {
+		if appErr, ok := err.(AppStatusError); ok {
+			switch appErr.Code {
+			case SCInputError:
+				err = ResolutionError{Input: input, Msg: err.Error(), Kind: ResolutionErrorInvalidInput}
+			// TODO: Phone number search rate limiting code goes here.
```
zapu: @mlsteele this one is being done in another ticket, left a TODO so it's easier to point someone where to look about changing code in the client

mlsteele: sounds good. Can be nice to have the ticket number number in the todo

PR #14193, opened 2018-10-10
Prime TeamName -> TeamID cache resolutions as a side effect of fetching favorites

maxtaco:
When a user does an `ls` in `/keybase/team`, it maps to a `GetFavorites` call,
and then, in common cases, an immediate call to `Resolve3` on every team in the
list. With this PR, we plumb the team ID back from the server (it was already being
sent down) and then prime the resolver cache with these resolutions. It should hopefully
immediately return in this case, rather than incurring a server round trip.

### DIFF
--- a/go/engine/favorite_list.go
+++ b/go/engine/favorite_list.go
@@ -54,12 +54,35 @@ func (f *FavoritesAPIResult) GetAppStatus() *libkb.AppStatus {
 	return &f.Status
 }
 
+func (e *FavoriteList) cacheFolder(m libkb.MetaContext, folder keybase1.Folder) {
+	name, err := keybase1.TeamNameFromString(folder.Name)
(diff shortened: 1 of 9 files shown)


maxtaco:
cc @strib

maxtaco:
cc @chrisnojima, I had to tinker with flow here

maxtaco:
it Go, if it's optional, we just get an empty string.  but i can do it if you want me to....

chrisnojima:
I think if it's really optional we should mark it as such. Then we don't need to plumb all these empty strings and it's more accurate. We can leave it as is but that's the ideal I think

maxtaco:
ok, i'll fix it.

maxtaco:
@mmaxim this ok?

Issue #14139, opened 2018-10-08
KBFS key generation too low


gwenall:
`ERROR CODE 218 - error unboxing chat message (transient): KBFS key generation too low: 1 < 3 in method chat.1.local.getThreadNonBlock`

Facing this error from mid-June on a 1 to 1 chat  where messages cannot be sent nor received from my end to a specific user. 
however I don't face any issue chatting to any new or existing contact
Tried to clear the entire conversation with this specific user but face the same issue.

Keybase Version 2.7.2-20181003163314+89c6c11dc6

mmaxim:
Try doing `keybase db nuke`, that should hopefully fix your situation (this clears some local caches).

gwenall:
Thanks for the quick reply  @mmaxim 
I tried the db nuke, but,even with an empty chat, the very same error came up whenever I tried to send any messages to this particular contact
My log ID is:   ce96d72977c15098cc26241c

mmaxim:
Ok I will take a look, but in the meantime if you want to work around the problem you can create a team with just you and the person and can talk in there. It will work just the same as the normal chat. Sorry!

gwenall:
Thank you

PR #14198, opened 2018-10-11
add more chat permanent errors for missing keys CORE-9134

mmaxim:
This can happen in some bizarre upgrade scenarios, so let's not error out the whole thread with a "transient" error and label these key errors as permanent.


mmaxim:
Addresses: https://github.com/keybase/client/issues/14139

PR #14206, opened 2018-10-11
jenkins test yarn.lock and build-actions

chrisnojima:
- [x] ensure yarn.lock test works
- [x] ensure build-actions test works


chrisnojima:
@keybase/react-hackers this is ready. the first commit is the updates to the test file and changes to package.json and an actions.json file
commit 2 fixes the package.json and that test passes
commit 3 fixes teh actions.json file and that test passes

PR #14209, opened 2018-10-11
Stellar: Move wallet tab between Teams and Devices in desktop's main nav

cjb:
@keybase/react-hackers 

* Use one source of truth instead of two (render, hotkeys) for desktop tab order
* Calculate the hotkey instead of hardcode, since we're no longer just adding one to the end for wallets
* Move the Wallets tab inbetween Teams and Devices

### DIFF
--- a/shared/app/nav.desktop.js
+++ b/shared/app/nav.desktop.js
@@ -11,7 +11,6 @@ import * as Tabs from '../constants/tabs'
 import {switchTo} from '../actions/route-tree'
 import {connect, type TypedState} from '../util/container'
 import {globalStyles} from '../styles'
-import flags from '../util/feature-flags'
 import RpcStats from './rpc-stats'
 
 type Props = {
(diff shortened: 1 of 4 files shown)


shared/app/nav.desktop.js
```diff
@@ -22,16 +21,11 @@ type Props = {
...
-  '2': Tabs.chatTab,
-  '3': Tabs.fsTab,
-  '4': Tabs.teamsTab,
-  '5': Tabs.devicesTab,
-  '6': Tabs.gitTab,
-  '7': Tabs.settingsTab,
-  ...(flags.walletsEnabled ? {'8': Tabs.walletsTab} : {}),
-}
+
+const hotkeyTabMap = Tabs.desktopTabOrder.reduce((tabMap, tab, index) => {
```
cjb: Anyone want to try to find a shorter way to do this? :)  `const hotkeyTabMap = {...Tabs.desktopTabOrder}` is close -- but it's 0-indexed instead of 1-indexed.

chrisnojima: 1. you can keep it zero based and just change it in the _onHotkey to add the offset
1. you could just do something like `{...['', ...Tabs.desktopTabOrder]}` but thats not very readable

PR #14210, opened 2018-10-11
reverts commit 3d549c370eb597948f622d315c4e12e74b6aff01.

maxtaco:
- Revert "remove strict check for now, to avoid link cache bust"
- check validity of link IDs when looking up special links
- introduce a disk_version field


maxtaco:
@mmaxim giving this a second chance.  i think it failed so miserably before because i didn't accommodate people caching old links.  now i do.

maxtaco:
(let's let this bake before we push out linux though)

go/libkb/chain_link.go
```diff
@@ -221,6 +221,7 @@ type ChainLink struct {
 	Contextified
 	parent            *SigChain
 	id                LinkID
+	diskVersion                        int
```
mmaxim: This is kind of weird whitespace here, how did `gofmt` let this happen?

PR #14211, opened 2018-10-11
better sign out / sign in

chrisnojima:
- [x] handles last username better
- [x] reloads inbox on log in always
- [x] remove reset checking logic which (in __DEV__) runs the reducer differently than on prod which we should never do

@keybase/react-hackers


shared/actions/chat2/index.js
```diff
@@ -2659,7 +2659,7 @@ function* chat2Saga(): Saga.SagaGenerator<any, any> {
   yield Saga.safeTakeEveryPure(Chat2Gen.openFolder, openFolder)
 
   // On bootstrap lets load the untrusted inbox. This helps make some flows easier
```
buoyad: `On login`

PR #14212, opened 2018-10-11
Preserve pending/history split from service

buoyad:
Adds a `section` field to payments and uses it to put them in the right places. r? @keybase/react-hackers

### DIFF
--- a/shared/wallets/wallet/container.js
+++ b/shared/wallets/wallet/container.js
@@ -20,7 +20,8 @@ const mapDispatchToProps = (dispatch, {navigateAppend, navigateUp}) => ({
   navigateAppend,
   navigateUp,
   _onLoadMore: accountID => dispatch(WalletsGen.createLoadMorePayments({accountID})),
-  _onMarkAsRead: (accountID, mostRecentID) => dispatch(WalletsGen.createMarkAsRead({accountID, mostRecentID})),
+  _onMarkAsRead: (accountID, mostRecentID) =>
+    dispatch(WalletsGen.createMarkAsRead({accountID, mostRecentID})),
 })
 
 const mergeProps = (stateProps, dispatchProps) => {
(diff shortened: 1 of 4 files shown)


shared/wallets/wallet/container.js
```diff
@@ -40,12 +41,12 @@ const mergeProps = (stateProps, dispatchProps) => {
   let mostRecentID
   stateProps.payments &&
```
chrisnojima: could just use lodash.partition

buoyad: Thanks, done :+1:

shared/wallets/wallet/container.js
```diff
@@ -36,29 +37,18 @@ const mergeProps = (stateProps, dispatchProps) => {
...
-        history.push({paymentID: p.id, timestamp: p.time})
-      } else if (p.section === 'pending') {
-        if (!pending) {
-          pending = []
-        }
-        pending.push({paymentID: p.id, timestamp: p.time})
-      }
-    })
-  if (history) {
+  const paymentsList = stateProps.payments && stateProps.payments.toList().toArray()
```
chrisnojima: can't it just be
```
const mapped = (stateProps.payments?.toList().toArray())?.map(p => ({paymentID: p.id, timestamp: p.time}))
const [history, pending] = partition(mapped, p => p.section === 'history')
```

buoyad: Then you'd need to add `section` to the map too

chrisnojima: ok we can keep it as-is but it could be something like. cause splitPayments should never be null and you don't need to check the param to partition
```
const paymentToItem = p => ({paymentID: p.id, timestamp: p.time})
const [history, pending] = partition(stateProps.payments?.toList().toArray(), p => p.section === 'history').map(arr => arr.map(paymentToItem)
```

chrisnojima: or more like your existing code
```
  const [_history, _pending] = partition(paymentsList, p => p.section === 'history')
  const mapItem = p => ({paymentID: p.id, timestamp: p.time})
  let history = _history.map(mapItem)
  const pending =_pending.map(mapItem)
```

buoyad: @chrisnojima done

PR #14213, opened 2018-10-11
use classNames and fix blue highlight on filtered conversations

chrisnojima:
- [x] arrow keys on jump to conversation now has the correct background color
- [x] added classNames helper to Styles and applied it everywhere


chrisnojima:

chrisnojima:
cc: @adamjspooner @cecileboucheron for usage of the new helper (just search `classNames(` )

adamjspooner:
Thanks!

PR #12094, opened 2018-05-29
Implement GetPaymentsLocal

patrickxb:
This does not handle the "pagination" part of the ticket yet, but thought it would be good to get this part of it in now.

PR #14216, opened 2018-10-11
Add PaymentStatus_Canceled and use for relay payments that have been yanked

buoyad:


patrickxb:
This looks fine to me, but I'm having deja vu here...pretty sure we had it like this in a branch once and @mlsteele had a reason not to do it like this.

@mlsteele what do you think?

mlsteele:
This looks good.

@patrickxb I did give you [hard time](https://github.com/keybase/client/pull/12094#discussion_r192221180) about something related. But I think that was only about non-completed relays. Whereas this change only affects completed relays.

buoyad:
Thanks all!

PR #14218, opened 2018-10-11
delete incoming messages if they're of type delete

chrisnojima:
this is a very small fix to fix missing timestamps on messages that come after a deleted message. I think we could discuss handling of deletes differently in the future but i think thats a larger change that would require some care. there are flows where we get placeholders and we need to replace them with delete (or clean up in a different way). Short term (this pr) if we load a thread and get a deleted message we just toss it immediately and act as if it didn't exist. 

@keybase/react-hackers


mmaxim:
Have you actually seen this happen? I don't think it should be possible.

chrisnojima:
Yes 100%. Load Keybase#design ( see the jira ticket)

mmaxim:
Ok please hold off on merging for a second, I want to try and understand what is going on here.

chrisnojima:
- [ ] waiting on mike

chrisnojima:
@mmaxim updated, can you test the flow you mentioned. i tried and it seemed to work but we should be more sure

PR #14219, opened 2018-10-11
BuildRequestLocal

songgao:


protocol/avdl/stellar1/local.avdl
```diff
@@ -335,6 +335,59 @@ protocol local {
...
+  // Check but do not send a request.
+  // Arguments are optional unless marked [required] or otherwise described.
+  // However, a coherent payment is required in order for res.readyToSend to returned as true.
+  BuildRequestResLocal buildRequestLocal(
+    int sessionID,
+
+    // Recipient username, acctID, or assertion.
+    // Examples: "", "russel", "GBQTE2V7Y356TFBZL6YZ2PA3KIILNSAAQRV5C7MVWS22KQTS4EMK7I46"
+    string to,
+    boolean toIsAccountID, // [required]
```
mlsteele: can you request from an account ID? If not don't need toIsAccountID

protocol/avdl/stellar1/local.avdl
```diff
@@ -335,6 +335,59 @@ protocol local {
...
+
+    string amount, // Amount of the asset OR outside currency being sent. Example: "3.005"
```
mlsteele: "being sent" -> "being requested"
handful of copy paste things like this

protocol/avdl/stellar1/local.avdl
```diff
@@ -335,6 +335,59 @@ protocol local {
...
+    // Zero or one of `asset` and `currency` should be set.
+    // If both are null or `asset` is native then XLM are being sent.
+    // If `currency` is set then XLM is being sent after converting `amount`.
+    // Note: Sending non-XLM assets is not yet supported.
+    union { null, OutsideCurrencyCode } currency,
+    union { null, Asset } asset,
+
+    string secretNote   // Encrypted for the sender, and recipient if possible.
```
mlsteele: for requests secretNote is always visible to the recipient. Ends up in chat message.

protocol/avdl/stellar1/local.avdl
```diff
@@ -335,6 +335,59 @@ protocol local {
...
+    string toErrMsg;
+
+    // Optional. `arg.to` may be an account ID that resolves to a user.
+    // If this is returned, caller could switch to (to=res.toUsername,toIsAccountID=false).
+    string toUsername;
+
+    // Optional. Error to show with the amount field.
+    // Example: "russel doesn't accept *BTC/Stronghold.com*\nPlease pick another asset."
+    // Example: "Your available to send is *128.4567890 XLM*."
+    string amountErrMsg;
```
mlsteele: neither example makes sense for requests

protocol/avdl/stellar1/local.avdl
songgao: by account ID you mean a stellar address right? I don't see that in design, so looks like at least not for now. So I'll remove it for now.

protocol/avdl/stellar1/local.avdl
```diff
@@ -336,6 +336,57 @@ protocol local {
...
+    boolean readyToRequest;
+
+    // Optional. Error to show with the To field.
+    // All errMsg fields are in 'keybase chat language' markdown.
+    // Example: "This stellar address is incorrect"
+    string toErrMsg;
+
+    // Optional. `arg.to` may be an account ID that resolves to a user.
+    // If this is returned, caller could switch to (to=res.toUsername,toIsAccountID=false).
+    string toUsername;
```
mlsteele: delete this return value

protocol/avdl/stellar1/local.avdl
```diff
@@ -336,6 +336,57 @@ protocol local {
     boolean pending;
   }
 
+  // Check but do not send a request.
+  // Arguments are optional unless marked [required] or otherwise described.
+  // However, a coherent payment is required in order for res.readyToSend to returned as true.
+  BuildRequestResLocal buildRequestLocal(
+    int sessionID,
+
+    // Recipient username, acctID, or assertion.
```
mlsteele: s/Recipient/Requestee

protocol/avdl/stellar1/local.avdl
```diff
@@ -336,6 +336,57 @@ protocol local {
...
+    // Examples: "", "russel", "GBQTE2V7Y356TFBZL6YZ2PA3KIILNSAAQRV5C7MVWS22KQTS4EMK7I46"
+    string to,
+
+    string amount, // Amount of the asset OR outside currency being requested. Example: "3.005"
+    // Zero or one of `asset` and `currency` should be set.
+    // If both are null or `asset` is native then XLM are being requested.
+    // If `currency` is set then XLM is being requested after converting `amount`.
+    // Note: Sending non-XLM assets is not yet supported.
```
mlsteele: s/Sending/Requesting
It may actually be supported, but that would be a weird asymmetry, so whatever you want.

protocol/avdl/stellar1/local.avdl
```diff
@@ -336,6 +336,57 @@ protocol local {
...
+    int sessionID,
+
+    // Recipient username, acctID, or assertion.
+    // Examples: "", "russel", "GBQTE2V7Y356TFBZL6YZ2PA3KIILNSAAQRV5C7MVWS22KQTS4EMK7I46"
+    string to,
+
+    string amount, // Amount of the asset OR outside currency being requested. Example: "3.005"
+    // Zero or one of `asset` and `currency` should be set.
+    // If both are null or `asset` is native then XLM are being requested.
+    // If `currency` is set then XLM is being requested after converting `amount`.
```
mlsteele: I think that if real USD is requested then the sender's form will be pre-filled in USD. So I don't think this sentence is helping. Probably remove it.

go/stellar/stellarsvc/frontend.go
```diff
@@ -1253,6 +1253,102 @@ func (s *Server) CreateWalletAccountLocal(ctx context.Context, arg stellar1.Crea
...
+		})
+	} else {
+		bpaArg.From = &primaryAccountID
+	}
+
+	amountX := s.buildPaymentAmountHelper(ctx, bpc, bpaArg)
+	res.AmountErrMsg = amountX.amountErrMsg
+	res.WorthDescription = amountX.worthDescription
+	res.WorthInfo = amountX.worthInfo
+	readyChecklist.amount = amountX.haveAmount
```
mlsteele: this unnecessarily dies if it can't show the price in XLM, but that's probably worth having less code 👍

songgao: Before this line it was failing the check all the time, so I just added it without know much of what's behind the scene. What should the desired behavior be here? Should `readyToRequest` maybe not depend on `readyChecklist.amount`?

mlsteele: There's nothing behind the scenes of readyChecklist. That's just some booleans. readyToRequest is supposed to mean that passing the args to `makeRequestLocal` would likely succeed. So `readyChecklist.amount = amountX.haveAmount` is good. I'd just leave it. I was just mentioning that if you put 3 USD and it failed to figure out how many XLM that was that would cause readyToSend to be false which is not strictly necessary.

protocol/avdl/stellar1/local.avdl
```diff
@@ -336,6 +336,57 @@ protocol local {
     boolean pending;
   }
 
+  // Check but do not send a request.
+  // Arguments are optional unless marked [required] or otherwise described.
+  // However, a coherent payment is required in order for res.readyToSend to returned as true.
```
mlsteele: s/readyToSend/readyToRequest

PR #14220, opened 2018-10-11
Use the user's default currency when choosing which one to send

zanderz:


shared/reducers/wallets.js
```diff
@@ -41,7 +41,9 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
     case WalletsGen.displayCurrenciesReceived:
       return state.set('currencies', I.List(action.payload.currencies))
     case WalletsGen.displayCurrencyReceived:
-      return state.setIn(['currencyMap', action.payload.accountID], action.payload.currency)
+      return state
+        .setIn(['currencyMap', action.payload.accountID], action.payload.currency)
+        .set('buildingPayment', state.get('buildingPayment').merge({currency: action.payload.currency.code}))
```
chrisnojima: use `.update`

zanderz: like this?

zanderz:
ready again after some confusing merging

shared/reducers/wallets.js
```diff
@@ -50,7 +50,9 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
     case WalletsGen.displayCurrenciesReceived:
       return state.set('currencies', I.List(action.payload.currencies))
     case WalletsGen.displayCurrencyReceived:
-      return state.setIn(['currencyMap', action.payload.accountID], action.payload.currency)
+      return state
+        .update('currencyMap', c => c.set(action.payload.accountID, action.payload.currency))
+        .set('building', state.get('building').merge({currency: action.payload.currency.code}))
```
chrisnojima: this can use update also

zanderz: ok

PR #14221, opened 2018-10-12
change PostEditNonblock to take either message ID or outbox ID CORE-9152

mmaxim:
Patch aims to solve the problem of editing pending messages behaving strangely in current builds (normally generates two sends of the same message, slightly altered). cc @chrisnojima @buoyad 

1.) Add a new parameter to `PostEditNonblock`, called `Target` which can be used to specify either a message or outbox ID. The frontend just aims to send whatever it has for the message being edited.
2.) If we get a message ID, then just operate as normal. If we get an outbox ID, we attempt to update the message in the outbox. If we find the outbox ID in the outbox, we update the contents.
3.) If we can't find the outbox ID in the outbox, we fetch the thread and try to find a message with that outbox ID on it, and post an edit.
4.) If we find the outbox ID, but when we update we don't do anything, then do the fetch thread thing as well.
5.) Add a new method to `Deliverer`, called `UpdateInFlight`, which will update the outbox item as well as cancel any outstanding send (in case we are updating the thing we are sending). Note this phase isn't perfect, since we are trying to solve a race between the UI, service, and server.


shared/reducers/chat2.js
```diff
@@ -214,6 +214,18 @@ const messageMapReducer = (messageMap, action, pendingOutboxToOrdinal) => {
...
     }
+    case Chat2Gen.pendingMessageWasEdited: {
+      const {conversationIDKey, ordinal, text} = action.payload
+      return messageMap.updateIn(
+        [conversationIDKey, ordinal],
+        message =>
+          !message || message.type !== 'text'
+            ? message
+            : message.withMutations(m => {
+                m.set('text', text)
```
chrisnojima: This can just be message.set('text', text)

shared/reducers/chat2.js
```diff
@@ -820,6 +832,7 @@ const rootReducer = (
     case Chat2Gen.messageDelete:
     case Chat2Gen.messageEdit:
     case Chat2Gen.messageWasEdited:
+    case Chat2Gen.pendingMessageWasEdited:
```
chrisnojima: ??

mmaxim: What's wrong here?

chrisnojima: Is that the same key?

mmaxim: The first one is in `metaMapReducer` the second is in `rootReducer`

mmaxim:
Ok @joshblum this is ready for another look with the new method I described.

go/chat/sender.go
```diff
@@ -437,6 +438,44 @@ func (s *BlockingSender) checkTopicNameAndGetState(ctx context.Context, msg chat
 	return topicNameState, nil
 }
 
+func (s *BlockingSender) resolveOutboxIDEdit(ctx context.Context, uid gregor1.UID,
+	convID chat1.ConversationID, msg *chat1.MessagePlaintext) error {
+	if msg.SupersedesOutboxID == nil {
+		return nil
+	}
+	s.Debug(ctx, "resolveOutboxIDEdit: resolving edit: outboxID: %s", msg.SupersedesOutboxID)
+	typ, err := msg.MessageBody.MessageType()
```
joshblum: small nit: can you add a `MessageBody.IsEdit()` method in extras? just cleaner to read, maybe avdl should just generate this..

go/chat/storage/outbox.go
```diff
@@ -351,27 +351,43 @@ func (o *Outbox) RetryMessage(ctx context.Context, obid chat1.OutboxID,
 	return res, nil
 }
 
-func (o *Outbox) UpdateMessage(ctx context.Context, replaceobr chat1.OutboxRecord) error {
```
joshblum: revert these changes? seem unused

mmaxim: I kind of like them, I reverted the changes to `MessageDeliverer` though.

go/chat/sender.go
mmaxim: Sure

mmaxim: Actually I kind of like having two different kinds of error here...

joshblum: ok we can leave as is

PR #14231, opened 2018-10-12
hook indexer into conv source

joshblum:
patch does the following:

- calls `Indexer.Add` on incoming messages. this handles adding search terms for new messages, and updating the index on EDITs
- calls `Indexer.Remove` on any deletes (DELETE, expunge, DELETEHISTORY, exploding)
- factors `ChatHelper` to take in a `chat1.GetThreadReason` so we can call it directly instead of making another helper, we can add more reasons, tried to make a best effort for existing calls.

the following PR will address filling in the missing index on search (in the background or syncronously depending on search flags) and output index percentage filled


go/chat/storage/storage.go
```diff
@@ -540,6 +546,9 @@ func (s *Storage) updateAllSupersededBy(ctx context.Context, convID chat1.Conver
 
 	// queue asset deletions in the background
 	s.assetDeleter.DeleteAssets(ctx, uid, convID, allAssets)
+	// queue search index update in the background
+	go s.G().Indexer.Remove(ctx, convID, uid, allPurged)
```
mmaxim: Can we put the goroutine behavior inside this guy? I like to return a channel as a callback from these kinds of function to give the caller the option to wait or not.

go/chat/storage/storage.go
```diff
@@ -437,14 +437,19 @@ func (s *Storage) MergeHelper(ctx context.Context,
 		}
 	}
 
+	// queue search index update in the background
+	go s.G().Indexer.Add(ctx, convID, uid, msgs)
```
mmaxim: Same here

PR #14233, opened 2018-10-12
WIP fix logger?

chrisnojima:


akalin-keybase:
Hmmm...I still see a stack in the console on desktop. Tried to set breakpoints but it wouldn't fire -- not sure where it's coming from!

chrisnojima:
Restart hot sever

chrisnojima:
Restart hot server

akalin-keybase:
Yeap, that fixed it, thanks

PR #14234, opened 2018-10-12
use selected currency when sending

zanderz:
Turns out the amount is always supposed to be in lumens, based on the command line example. So we needed the converted amount without any currency code or symbols.


mmaxim:
cc @patrickxb

go/stellar/stellarsvc/frontend.go
```diff
@@ -993,6 +993,7 @@ func (s *Server) BuildPaymentLocal(ctx context.Context, arg stellar1.BuildPaymen
 		} else {
 			res.AmountFormatted = amountFormatted
 		}
+		res.WorthAmount = amountX.amountOfAsset
```
patrickxb: Please add tests for this value to TestBuildPaymentLocal.  

I also think that if we are going to return WorthAmount, we should return WorthCurrency too.

Also, what is AmountFormatted that you added in the other PR used for?

zanderz: That was just to do e.g. `.5` => `0.5`. I put a comment in the go side but not in the avdl, I'll put it there too.

PR #14240, opened 2018-10-14
format image/video attachments with title on the bottom

mmaxim:
@cecileboucheron @adamjspooner @malgorithms 

So I brought up the idea of moving the caption of the image to the bottom of the preview in #design the other day, and I was just curious how it could play out. I put together this PR today to try it out, and I think it is a nice improvement. I attached a screenshot to this PR, but it would probably be most useful to just try out the branch and see what you think. This is more in the style of WhatsApp, which I think has a nice presentation for picture attachments. This also fixes up some of the absolute positioning elements (like the download button, which is in the screenshot being fixed). 

image

### DIFF
--- a/shared/chat/conversation/messages/attachment/image/index.js
+++ b/shared/chat/conversation/messages/attachment/image/index.js
@@ -81,8 +81,7 @@ class ImageAttachment extends React.PureComponent<Props, State> {
           <Kb.Box
             style={Kb.iconCastPlatformStyles(
               Styles.collapseStyles([
-                styles.loading,
-                !this.state.loaded && styles.spinner,
+                styles.backgroundContainer,
                 {
                   width: this.props.width + 6,
                   minHeight: !this.state.loaded ? this.props.height : 0,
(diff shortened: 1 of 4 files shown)


shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -78,70 +78,92 @@ class ImageAttachment extends React.PureComponent<Props, State> {
...
           <Kb.Box
             style={Kb.iconCastPlatformStyles(
               Styles.collapseStyles([
-                styles.loading,
-                !this.state.loaded && styles.spinner,
+                styles.backgroundContainer,
                 {
-                  height: this.props.height,
-                  width: this.props.width,
+                  width: this.props.width + 6,
```
adamjspooner: Can we add a comment for the `+ 6`?

mmaxim: Yeah the point is to make the div behind the photo just show up as a small border.

shared/chat/conversation/messages/attachment/image/index.js
```diff
@@ -255,21 +278,14 @@ const styles = Styles.styleSheetCreate({
...
-  }),
   title: Styles.platformStyles({
     isElectron: {
       wordBreak: 'break-word',
+      padding: 5,
+      display: 'block',
+    },
+    isMobile: {
+      padding: 5,
     },
```
adamjspooner: You could throw the padding into the `common` object if they're going to be the same. It could probably stand a few more pixels, too. Maybe `Styles.globalMargins.tiny`?

adamjspooner:
This is a great improvement!

shared/chat/conversation/messages/attachment/image/index.js
adamjspooner: On that note, can we either:
1. Increase the size of the container's border radius, or
2. Decrease the size of the image's border radius?

We want the spacing around the radius to be even. Currently, it gets fatter at the apex.

mmaxim: My last commit I think fixes this, I did both 1 and 2.

malgorithms:
I like this

mmaxim:
Ok I'll just merge and you all can play with it, easy to revert or tweak if we don't like it after a bit.  Thanks!

PR #14242, opened 2018-10-14
fix clickable area on screens with tooltips

songgao:
@buoyad I brought these back. Although they do follow scrolling now, it seems when they are not visible they still have weird positions, and block clickable areas in various places. Also even if their positions are correct, they'd still block some clickable area around the target that triggered. So it seems better to just not mount it when the mouse of out.

This fixes the bug where when user is at Files tab, tab buttons' clickable areas became smaller.

### DIFF
--- a/shared/common-adapters/with-tooltip.desktop.js
+++ b/shared/common-adapters/with-tooltip.desktop.js
@@ -21,7 +21,7 @@ class WithTooltip extends React.Component<Props, State> {
     this.setState({mouseIn: true})
   }
   _onMouseLeave = () => {
-    this.setState({mouseIn: false})
+    this.setState({mouseIn: false, visible: false})
   }
   _setAttachmentRef = attachmentRef => (this._attachmentRef = attachmentRef)
   componentDidUpdate(prevProps: Props, prevState: State) {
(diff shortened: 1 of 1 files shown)


shared/common-adapters/with-tooltip.desktop.js
```diff
@@ -22,6 +24,16 @@ class WithTooltip extends React.Component<Props, State> {
...
   _setAttachmentRef = attachmentRef => (this._attachmentRef = attachmentRef)
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    if (!prevState.mouseIn && this.state.mouseIn) {
+      // Set visible after Toast is mounted, to trigger transition on opacity.
+      // Note that we aren't doing anything to make ease out work. We don't
+      // keep Toast mounted all the time (which would make this unnecessary)
+      // because in that case the position of the overlay can be messed up when
+      // not visible, causing ghost unclickable areas.
+      this.setState({visible: true})
+    }
```
buoyad: Can we set `visible` to false in the opposite condition? I think not doing that is causing the fade-in to only happen once per mount

songgao: good catch!

PR #14246, opened 2018-10-15
upak lite teamloader 2

heronhaye:
Failing tests are due stellar testnet outage. The "disable upak lite in prod" commit shows the change I made since you last looked at it.

### DIFF
--- a/go/systests/teams_test.go
+++ b/go/systests/teams_test.go
@@ -1061,7 +1061,7 @@ func TestTeamSignedByRevokedDevice(t *testing.T) {
 
 	t.Logf("bob should see alice's key is revoked")
 	{
-		_, _, pubKey, err := bob.tc.G.GetUPAKLoader().LoadKeyV2(context.TODO(), alice.uid, revokedKID)
+		_, _, pubKey, err := bob.tc.G.GetUPAKLoader().LoadKey(context.TODO(), alice.uid, revokedKID, false)
 		require.NoError(t, err)
 		t.Logf("%v", spew.Sdump(pubKey))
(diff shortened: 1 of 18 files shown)


go/teams/chatmessaging.go
```diff
@@ -58,12 +58,12 @@ func SendChatInviteWelcomeMessage(ctx context.Context, g *libkb.GlobalContext, t
 		return false
 	}
 
-	inviterName, err := g.GetUPAKLoader().LookupUsername(ctx, inviter)
+	inviterName, err := g.GetUPAKLoader().LookupUsernameUPAKLite(ctx, inviter)
```
maxtaco: how does this not use upak lite in prod?

go/systests/teams_test.go
```diff
@@ -1061,7 +1061,7 @@ func TestTeamSignedByRevokedDevice(t *testing.T) {
 
 	t.Logf("bob should see alice's key is revoked")
 	{
-		_, _, pubKey, err := bob.tc.G.GetUPAKLoader().LoadKeyV2(context.TODO(), alice.uid, revokedKID)
+		_, _, pubKey, err := bob.tc.G.GetUPAKLoader().LoadKey(context.TODO(), alice.uid, revokedKID, true)
```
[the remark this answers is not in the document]
heronhaye: There's no LoadKeyV2 anymore, just a generalized LoadKey for both. Since includeLowKeys=true, it goes down the UPAK path.

go/teams/chatmessaging.go
heronhaye: You're right, my bad. I think in addition checking the run mode, the code should only load UPAKLites if a specific feature flag is set that is only set during some tests. Does that sound ok (in addition to fixing these)?

PR #14247, opened 2018-10-15
update smaller deps

chrisnojima:
updating smaller deps before updating rn to 573


chrisnojima:
actually hold up

chrisnojima:
@keybase/react-hackers this is ready again

PR #14248, opened 2018-10-15
Surya/core 9144/stellar formatting drop 00s

heronhaye:


patrickxb:
The ticket only says $0.00 should be $0.

heronhaye:
I asked Danny on the ticket and he said
>The ticket this split from says "Remove decimal digits if it's just zeros", cc @cecileboucheron 
which I intepreted as any value on the integer part.

patrickxb:
Can we confirm with design first?  @malgorithms @cecileboucheron @adamjspooner 

Seems weird to drop it to me.  Especially if I type in 5.00.

heronhaye:
I agree personally, I like the consistency of double decimals everywhere

malgorithms:
hmm... I think there are 3 imaginable ways we'd want to display amounts:

1. 2-digit mode. That should always show the `.00` which this PR removed. I agree with you both, I think.

2. full length always. Not sure we would use this anywhere, but `1 XLM` would be `1.0000000000 XLM` and `1234.123 XLM` would be `1,234.1230000000 XLM`.  (or whatever - not sure how many digits). Pretty much the only time to use this is when right justifying a table of numbers.

3. smart-precision mode. this is where we drop the decimal when possible (which we're trying to do here), and add extra decimals when needed to avoid ambiguity. `1 XLM` would stay `1 XLM` and `1.2 XLM` would stay `1.2 XLM`, but `1234 XLM` would have to be printed as `1,234.0 XLM`. and `1.234 XLM` would be printed as `1.2340 XLM`.

heronhaye:
@malgorithms Could you explain what you mean by "ambiguity"? Large enough to have a comma? Is the worry for countries that switch commas and decimal points?

malgorithms:
yeah, I think we want to avoid any issue where people don't know if we're using the thousands and decimal separators of their locale or using ours. So, this is dangerous: `1.123` .... is it 1123 or is it 1.12300000 ?

But anyone who sees `1.1` knows that it's `1.1`, so if we're printing `0` to `0` we should also be printing `1.2` as `1.2` and `1.23` as `1.23`. but `1.234` needs to become `1.2340`. Basically we append extra decimals until it's no longer ambiguous.

I don't think the logic here has to be that hard to produce good results. Unless we plan on full decimal everwhere (what I put as option 2) but that seems sucky.

heronhaye:
Okay, it seems like Miles already did this work with libkb.StellarSimplifyAmount, I'll just make FormatCurrency call that.

go/stellar/stellarsvc/service_test.go
```diff
@@ -763,7 +763,7 @@ func TestRequestPaymentOutsideCurrency(t *testing.T) {
 	require.Nil(t, details.Asset)
 	require.NotNil(t, details.Currency)
 	require.Equal(t, stellar1.OutsideCurrencyCode("USD"), *details.Currency)
-	require.Equal(t, "$8.20 USD", details.AmountDescription)
+	require.Equal(t, "$8.2 USD", details.AmountDescription)
```
patrickxb: This is what we want?  $8.2 USD?

heronhaye: I just talked to coyne, I misunderstood. The ticket should be dropped in favor of always doing two decimal points for USD like we were doing already.

malgorithms: yeah, sorry if I wasn't clear.

 I was trying to say in my long post above there are 3 _imaginable_ ways we'd want to format something. In the 2-digit case it should always be 2-digits; I'm agreeing with you two about fiats always getting decimals and I just relayed that to Cécile and Adam.

In the smart digit case it shouldn't use any unnecessary zeroes.

And in the full digit case (which I'm not sure we'll ever use for something like XLM) it would give full precision always.

PR #14258, opened 2018-10-15
change clipboard paste attachment send to use a temp file managed by service CORE-8306

mmaxim:
@keybase/react-hackers 

The point of this patch is to have the JS use a temp file that is managed by the service when handling an attachment send via pasting from the clipboard. Currently JS creates its own file, but it doesn't have any mechanism for cleaning this off the disk. Having the service manage this is also cleaner.


mmaxim:
bump

shared/chat/conversation/normal/container.js
```diff
@@ -20,6 +20,8 @@ const mapStateToProps = (state, {conversationIDKey}) => {
 }
 
 const mapDispatchToProps = dispatch => ({
+  _onPaste: (conversationIDKey: Types.ConversationIDKey, data: Buffer) =>
+    dispatch(Chat2Gen.createAttachmentPasted({conversationIDKey, data})),
   _onAttach: (conversationIDKey: Types.ConversationIDKey, paths: Array<string>) =>
     dispatch(
       RouteTree.navigateAppend([{props: {conversationIDKey, paths}, selected: 'attachmentGetTitles'}])
```
MarcoPolo: Isn't this going to be broken? since paths here is an array of strings? Did you grep do see if anywhere else is navigating to `attachmentGetTitles`.

Since path is passed as a routeProp there really isn't any type checking to make sure that the `paths` you pass in to the route prop when you navigate is the correct type

shared/chat/conversation/attachment-get-titles/container.js
```diff
@@ -9,7 +9,10 @@ import {navigateUp} from '../../../actions/route-tree'
 import {type RouteProps} from '../../../route-tree/render-route'
 import type {PathToInfo} from '.'
 
-type OwnProps = RouteProps<{paths: Array<string>, conversationIDKey: Types.ConversationIDKey}, {}>
+type OwnProps = RouteProps<
+  {paths: Array<Types.PathAndOutboxID>, conversationIDKey: Types.ConversationIDKey},
```
MarcoPolo: I feel like this should be changes to be called pathAndOutboxIDs instead of paths, to reflect the type better.

mmaxim:
Thanks @MarcoPolo! I made some changes and double checked all flows, and I think it is in good shape. Also made the change @chrisnojima and I discussed this morning where JS no longer actually writes this file.

PR #14259, opened 2018-10-15
Autofocus asset input on opening send form

buoyad:
Had to switch around some `replaceEntity`s in search actions so we could use `showingSearchSuggestions` to cancel focusing the to field input.
r? @keybase/react-hackers


chrisnojima:
does this affect regular search in various places

buoyad:
@chrisnojima No not as far as I can tell. I tested profile/create new convo/add to team and they work as expected

PR #14262, opened 2018-10-15
[KBFS-3364] Files header action

jzila:
Plus a bit of a refactor for style cleanness. Actual code changes in https://github.com/keybase/client/pull/14262/files/2669a153165ed2c8f63549af78ff9e0eb1c851cd..b66dde4aeb4da11e7c58f225f3f5707b94363e23


songgao:
LGTM! I'm not using the "review" approval so it doesn't get cleared from the review queue just in case you want to block on a real FE review.

I remember getting feedback on not having connected stuff in `index.js` but `container.js` instead. Although I think in this case to make it easily importable, I personally prefer to have them in `index.js` (the way you have).

jzila:
@songgao thanks, I think your review is good enough since this only touches KBFS stuff.

PR #14263, opened 2018-10-15
fix "update now"; show "Updating ..."; check again after update

songgao:
The "check again" part is because user could hit "Ignore", and nothing would get updated. In this case we should not show the out-of-date banner any more. So ask service again and eventually update the store to reflect the fact that no update is needed anymore.

Tagging @keybase/kbfs-hackers for the first commit in Go, and @keybase/react-hackers for the second commit in GUI.

### DIFF
--- a/go/service/config.go
+++ b/go/service/config.go
@@ -17,7 +17,6 @@ import (
 
 	"github.com/keybase/client/go/engine"
 	"github.com/keybase/client/go/install"
-	"github.com/keybase/client/go/libcmdline"
 	"github.com/keybase/client/go/libkb"
 	keybase1 "github.com/keybase/client/go/protocol/keybase1"
 	"github.com/keybase/go-framed-msgpack-rpc/rpc"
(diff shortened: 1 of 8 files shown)


chrisnojima:
I don't think this is how this should work. If there's an update we should show it. If they ignore we should let them override the snooze

mmaxim:
Agree, I'm not sure why this snooze can't be overridden, but we should change it. I feel like it should only affect the auto updating aspect of the updater, but if I hit `keybase update check` or this out of date banner I should get the update regardless of my snooze state.

songgao:
There's `KEYBASE_UPDATER_FORCE` env var but I don't think it differentiates snoozed update or real update. This is server controlled. I don't know where the code is but can try to dig out.

We can add a parameter for this banner as well as the `keybase update check` CLI so it ignores the snooze in these places, but still have the auto updater (if it's enabled) follow the snooze.

mmaxim:
Yeah I know the server controls it, maybe @maxtaco can you point us in the right direction.

songgao:
Cool thanks. I'm closing this PR so people don't start reviewing it. Will put it up again when it's ready!

PR #14266, opened 2018-10-16
Disable buildpayment accountID->username resolution

mlsteele:
Remove built.toUsername. That field was used only for a canned feature where an account ID could morph in the gui into a username.

There is a related bug that will be fixed in an upcoming PR where when the service is told to send to an account ID it actually sends to a user.


shared/wallets/confirm-form/participants/container.js
```diff
@@ -8,21 +8,20 @@ const mapStateToProps = state => {
...
-  const recipientFullName = userInfo ? userInfo.fullname : ''
+  let recipientFullName = userInfo ? userInfo.fullname : ''
   const fromAccount = getAccount(state, built.from)
 
-  const recipientType = build.recipientType
   let recipientStellarAddress
   let recipientAccountIsDefault
   let recipientAccountName
   let recipientAccountAssets
   if (recipientType === 'keybaseUser') {
```
buoyad: I think you can remove this block entirely

mlsteele: which block?

buoyad: Can replace
```
if (recipientType === 'keybaseUser') {
    if (build.to.includes('@')) {
      // this is an sbs assertion
      recipientFullName = ''
    }
} else
```
with
``` 
if (recipientType !== 'keybaseUser')
```
I don't think we'll ever find a fullName in `userInfo` for sbs, and if we do I don't think it'd be inaccurate to show.

mlsteele: ah nice thanks

PR #14267, opened 2018-10-16
Fix 'retry' not appearing on failed messages sometimes

buoyad:
The submitState is:

* 'Pending' when the failed outbox record comes with a fresh thread load
* 'Failed' when the failure comes from a `ChatActivityType.failedMessage`

r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/wrapper/wrapper-author/container.js
+++ b/shared/chat/conversation/messages/wrapper/wrapper-author/container.js
@@ -88,7 +88,7 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps) => {
     failureDescription = message.errorReason
     if (stateProps.isYou && ['pending', 'failed'].includes(message.submitState)) {
       // This is a message still in the outbox, we can retry/edit to fix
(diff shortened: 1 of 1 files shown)


shared/chat/conversation/messages/wrapper/wrapper-author/container.js
```diff
@@ -86,7 +86,7 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps) => {
   let isErrorFixable = false
   if ((message.type === 'text' || message.type === 'attachment') && message.errorReason) {
     failureDescription = message.errorReason
-    if (stateProps.isYou && message.submitState === 'pending') {
+    if (stateProps.isYou && ['pending', 'failed'].includes(message.submitState)) {
       // This is a message still in the outbox, we can retry/edit to fix
       failureDescription = stateProps.isYou ? `Failed to send: ${message.errorReason}` : message.errorReason
```
mmaxim: Why do we check `isYou` again here? Is the first check for it wrong?

buoyad: Hm no, just an oversight. Thanks!

PR #14271, opened 2018-10-16
Move wallet search to floatingbox

chrisnojima:
mobile needs the search results to be in its own list and not on top how desktop currently implements it

### DIFF
--- a/shared/wallets/common/participants-row/index.js
+++ b/shared/wallets/common/participants-row/index.js
@@ -15,6 +15,11 @@ type Props = {|
 
 // A row for use in Participants components; provides a blue heading to the left of the content.
 class ParticipantsRow extends React.Component<Props> {
+  static defaultProps = {
+    bottomDivider: true,
+    headingAlignment: 'Left',
+  }
+
   render() {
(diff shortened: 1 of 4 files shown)


chrisnojima:
@keybase/react-hackers

shared/wallets/common/participants-row/index.js
```diff
@@ -14,32 +14,41 @@ type Props = {|
...
-      <Kb.Box style={styles.childContainer}>
-        <Kb.Box style={styles.childFillContainer}>{props.children}</Kb.Box>
-      </Kb.Box>
-    </Kb.Box2>
-    {props.bottomDivider && (
-      <Kb.Divider style={props.dividerColor ? {backgroundColor: props.dividerColor} : {}} />
-    )}
-  </React.Fragment>
-)
+class ParticipantsRow extends React.Component<Props> {
```
MarcoPolo: Might as well use PureComponent?

chrisnojima: ok but likely doesn't help as its driven by children which change all the time

PR #14275, opened 2018-10-16
enable zooming on desktop

chrisnojima:
@malgorithms should we just turn this on?


malgorithms:
hell yeah

chrisnojima:
@keybase/react-hackers this turns on command +/- for zooming in desktop. use at your own risk

adamjspooner:
Is there a `cmd+0` (return to normal) option?

chrisnojima:
yes that works

PR #14276, opened 2018-10-16
fix FTL bug with loading a parent team having been let in

maxtaco:
- the bug sequence we saw was: (1) user FTLs foo.bar; (2) gets
let into foo; and then (3) FTLs foo. This previously died with
a FastTeamError "no unverified key seed found at generation 1."
- fix by just doing a full reload in that case, setting ftl_low=0
- test that repros it

### DIFF
--- a/go/teams/ftl.go
+++ b/go/teams/ftl.go
@@ -203,6 +203,7 @@ type fastLoadArg struct {
 	downPointersNeeded []keybase1.Seqno
 	readSubteamID      keybase1.TeamID
 	needLatestName     bool
+	forceReset         bool
 }
 
 // needChainTail returns true if the argument mandates that we need a reasonably up-to-date chain tail,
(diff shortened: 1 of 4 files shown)


go/teams/ftl.go
```diff
@@ -221,7 +222,25 @@ func (f *FastTeamChainLoader) load(m libkb.MetaContext, arg fastLoadArg) (res *f
 	lock := f.locktab.AcquireOnName(m.Ctx(), m.G(), arg.ID.String())
 	defer lock.Release(m.Ctx())
 
-	return f.loadLocked(m, arg)
+	return f.loadLockedWithRetries(m, arg)
+}
+
+// loadLockedWithRetries attempts two loads of the team. If the first iteration returns an FTLMissingSeedError,
+// we'll blast through the cache and attempt a full reload a second time. Then that's for all the marbles.
```
heronhaye: was not familiar with this idiom! :+1:

heronhaye: was not familiar with this idiom! :+1:

PR #14294, opened 2018-10-17
[WIP KBFS-3515] Add file deletion

jzila:


shared/actions/fs/index.js
```diff
@@ -672,6 +672,17 @@ const letResetUserBackIn = ({payload: {id, username}}: FsGen.LetResetUserBackInP
 
 const letResetUserBackInResult = () => undefined // Saga.put(FsGen.createLoadResets())
 
+const deleteFile = (state, {payload: {path}}: FsGen.DeleteFilePayload) => new Promise(() => {
```
songgao: Why an additional promise? Would something like this work?

```
const deleteFile = (state, {payload: {path}}: FsGen.DeleteFilePayload) => 
    RPCTypes.SimpleFSSimpleFSRemoveRpcPromise({
        opID: Constants.makeUUID(),
        ....
    })
```

shared/actions/fs/index.js
```diff
@@ -672,6 +672,17 @@ const letResetUserBackIn = ({payload: {id, username}}: FsGen.LetResetUserBackInP
...
+  const opID = Constants.makeUUID()
+  return RPCTypes.SimpleFSSimpleFSRemoveRpcPromise({
+    opID,
+    path: {
+      PathType: RPCTypes.simpleFSPathType.kbfs,
+      kbfs: Constants.fsPathToRpcPathString(path),
+    },
+  })
```
songgao: I think we need a `.catch(makeRetriableErrorHandler(action)` here. Also need to update `Constants.erroredActionToMessage` to translate error for this action.

songgao:
Also I'm realizing `SimpleFSRemove` is an async operation, so we might need to call `SimpleFSWait` at some point.

jzila:
> Also I'm realizing SimpleFSRemove is an async operation, so we might need to call SimpleFSWait at some point.

Yeah I'm not sure. In my testing it worked perfectly.

strib:
I think calling Wait is required for async operations, otherwise kbfsfuse will leak memory.

jzila:
OK cool thanks. I'll fix it and ask for re-review then.

shared/actions/fs/index.js
jzila: Sure, although given that I'll need to add a `Wait`, seems like it's going to need a bit of a refactor anyway.

jzila:
@songgao ready for another look

shared/actions/fs/index.js
```diff
@@ -672,6 +672,19 @@ const letResetUserBackIn = ({payload: {id, username}}: FsGen.LetResetUserBackInP
...
+const deleteFile = (state, action: FsGen.DeleteFilePayload) => new Promise(() => {
+  const opID = Constants.makeUUID()
+  return RPCTypes.SimpleFSSimpleFSRemoveRpcPromise({
+    opID,
+    path: {
+      PathType: RPCTypes.simpleFSPathType.kbfs,
+      kbfs: Constants.fsPathToRpcPathString(action.payload.path),
+    },
+  })
+    .then(() => RPCTypes.SimpleFSSimpleFSWaitRpcPromise({opID}))
```
songgao: Hmm I'm not sure this would work. The thing that's resolved is passed into `Saga.put`. This would probably need to be a generator function.

jzila: Seems to work just fine: 
screen shot 2018-10-17 at 12 51 19 pm

songgao: Oh I think I understand now. This becomes another promise and it just keeps resolving. Sorry for the silly comment!

shared/actions/fs/index.js
```diff
@@ -672,6 +672,19 @@ const letResetUserBackIn = ({payload: {id, username}}: FsGen.LetResetUserBackInP
 
 const letResetUserBackInResult = () => undefined // Saga.put(FsGen.createLoadResets())
 
+const deleteFile = (state, action: FsGen.DeleteFilePayload) => new Promise(() => {
```
songgao: Mind removing this `new Promise(` layer?

PR #14309, opened 2018-10-18
Check nil obr

joshblum:
hit this in log `8bc6c38b64b9b3d00d50fb1c`


joshblum:
discussed offline: there was a race when multiple retry requests were issued

PR #14311, opened 2018-10-18
more mobile send: send from profile

chrisnojima:
this enables send from profile and fixes some random things i saw


shared/people/container.js
```diff
@@ -28,7 +27,7 @@ const mapStateToProps = state => ({
   waiting: WaitingConstants.anyWaiting(state, getPeopleDataWaitingKey),
 })
 
-const mapDispatchToProps = dispatch => ({
+const mapDispatchToProps = (dispatch, {navigateAppend}) => ({
```
chrisnojima: if you click search too quickly you can get a redbox

shared/actions/wallets.js
```diff
@@ -235,6 +235,7 @@ const loadSendAssetChoices = (state: TypedState, action: WalletsGen.LoadSendAsse
   })
 
 const loadDisplayCurrency = (state: TypedState, action: WalletsGen.LoadDisplayCurrencyPayload) =>
+  action.payload.accountID !== Types.noAccountID &&
```
chrisnojima: we can call this with noAccountID and get a black bar

shared/react-native/ios/Keybase/NativeLogger.m
```diff
@@ -27,20 +27,22 @@ + (BOOL)requiresMainQueueSetup
...
-  AppDelegate *appDel = (AppDelegate *)[[UIApplication sharedApplication] delegate];
-  DDFileLogger *fileLogger = (DDFileLogger *)appDel.fileLogger;
-  NSArray<NSString *> *paths = [[fileLogger logFileManager] sortedLogFilePaths];
-  for (NSString *path in paths) {
-    NSString *fileContents = [NSString stringWithContentsOfFile:path encoding:NSUTF8StringEncoding error:NULL];
-    for (NSString *line in [fileContents componentsSeparatedByCharactersInSet:[NSCharacterSet newlineCharacterSet]]) {
-      NSRange range = [line rangeOfString:[NSString stringWithFormat:@"%@%@: ", tagPrefix, tagName]];
-      if (range.location != NSNotFound) {
-        [lines addObject:[line substringFromIndex:range.location + range.length]];
+  dispatch_async(dispatch_get_main_queue(), ^{
```
chrisnojima: this likely needs to only happen in the main queue. access to the AppDelegate for sure but likely this is all bad to touch from multiple threads

chrisnojima:
@keybase/react-hackers

shared/react-native/ios/Keybase/NativeLogger.m
mmaxim: Maybe better for this to have its own queue?

shared/actions/wallets.js
zanderz: would isValidAccountID be better?

chrisnojima: unclear, there's just one value so it doesn't seem necessary

shared/react-native/ios/Keybase/NativeLogger.m
chrisnojima: @mmaxim done

chrisnojima:
you aren't supposed to touch app delegate from a non-ui thread, so we bounce to that thread to get it, then bounce to a background thread to do the log dump

PR #14312, opened 2018-10-18
Support for making/posting/fetching stellar account bundles with mobile-only flag

patrickxb:
This is not the full feature, but the client code that can make a bundle for just one account, encrypt/encode it and post it to the server, fetch it and decode/decrypt.

TestMakeAccountMobileOnlyOnRecentMobile is an example of what the code can do at this point.


protocol/avdl/stellar1/bundle.avdl
```diff
@@ -77,4 +78,52 @@ protocol bundle {
...
+    Hash prev; // SHA256 of previous msgpack(EncryptedAccountBundle)
+    AccountID accountID;
+    AccountMode mode;
+  }
+
+  record AccountBundleSecretV1 {
+    Hash visibleHash; // SHA256 of msgpack(AccountBundleVisibleV1)
+    AccountID accountID;
+    array<SecretKey> signers;
+    string name;
```
mlsteele: I don't this this name field should exist. The name should be stored in 1 place not two. And that place should probably be in BundleSecretEntry.name since you'd want to be able to view and change the name of a mobile-only bundle from a desktop device.

patrickxb:
With your proposal, how would the server check that the client had access to the secret key at revision 4 before posting a new bundle with secret key at revision 5 (and mode == USER)?

i.e. if revision 4 was mobile only, couldn't a desktop device just post some secret key at revision 5 with mode==USER and put the hash in to the main bundle and post it?  server would accept it.

mlsteele:
Good point. Amend the proposal to be that `BundleVisibleEntry` (not `BundleSecretEntry`) should have a field that stores the hash of the ciphertext of the account bundle. That way the server can ensure that non-mobile clients don't change the account bundle hashes.

But also the server shouldn't let a non-mobile client change the mode of an account from MOBILE to USER.

patrickxb:
Ok, that seems reasonable to have the hash in BundleVisibleEntry.

This would require posting the "mega" bundle with the "account" bundle every time, which isn't necessarily bad, but different.

@maxtaco do you want to look at this and the discussion before I change anything?

maxtaco:
@patrickxb thanks for asking, no, just go ahead...

patrickxb:
@mlsteele please take another look.  bundle.avdl changed much like we discussed in chat, the tests from before pass (CI needs the api server changes in https://github.com/keybase/keybase/pull/3036).

Again, this doesn't attempt to support a full migration from V1 to V2, that will be next.

go/stellar/remote/remote.go
```diff
@@ -147,6 +148,55 @@ func Post(ctx context.Context, g *libkb.GlobalContext, clearBundle stellar1.Bund
...
+// PostBundleRestricted encrypts and uploads a restricted bundle to the server.
+func PostBundleRestricted(ctx context.Context, g *libkb.GlobalContext, bundle *stellar1.BundleRestricted) (err error) {
+	defer g.CTraceTimed(ctx, "Stellar.PostBundleRestricted", func() error { return err })()
+
+	pukGen, pukSeed, err := getLatestPuk(ctx, g)
+	if err != nil {
+		return err
+	}
+
+	// XXX fix this
```
mlsteele: xxx

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+	for acctID, acctBundle := range b.AcctBundles {
+		benc.AcctBundles[acctID] = acctBundle.EncB64
+	}
+
+	return benc
+}
+
+// BundleEncodedB64 contains all the encoded fields for communicating
+// with the api server to post and get account bundles.
+type BundleEncodedB64 struct {
```
mlsteele: B64 makes it sounds like it's a string type. name `BundleEncoded`?

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+	ws, err := AccountWithSecret(a, accountID)
+	if err != nil {
+		return err
+	}
+	if ws.Mode == stellar1.AccountMode_USER {
+		return ErrNoChangeNecessary
+	}
+	a.Revision++
+	a.Prev = a.OwnHash
+	a.OwnHash = nil
```
mlsteele: BundleRestricted is very similar to Bundle. So there's a bunch of duplicated code like these assignments in MakeAllDevices and MakeMobileOnly. What about just using Bundle and then using bundle.Advance for these?

go/stellar/remote/remote.go
patrickxb: thanks

patrickxb:
I didn't want to mess with the existing endpoints so that the two systems can coexist temporarily.

go/stellar/acctbundle/acctbundle.go
patrickxb: But it's different (BundleRestricted).  

I felt it was simpler to leave Bundle alone than try to make it work with existing bundles and new bundles.

protocol/avdl/stellar1/bundle.avdl
```diff
@@ -15,48 +21,155 @@ protocol bundle {
...
+    V2_2,
+    V3_3,
+    V4_4,
+    V5_5,
+    V6_6,
+    V7_7,
+    V8_8,
+    V9_9,
+    V10_10
+  }
```
mlsteele: I bet the account bundle versioning will be pretty intimately tied to BundleVersion so it might make sense to use `BundleVersion`. But this is fine too

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+
+	return x, nil
+}
+
+// NewFromBundle creates a BundleRestricted from a Bundle.
+func NewFromBundle(bundle stellar1.Bundle) (*stellar1.BundleRestricted, error) {
+	r := &stellar1.BundleRestricted{
+		Revision:       bundle.Revision,
+		Prev:           bundle.Prev,
+		AccountBundles: make(map[stellar1.AccountID]stellar1.AccountBundle),
```
mlsteele: want ownhash for posting a good prev with the upgrade?

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+	parent, parentVersion, err := unboxParent(encBundle, hash, encodedBundle.VisParent, puk)
+	if err != nil {
+		return nil, 0, err
+	}
+	parent.AccountBundles = make(map[stellar1.AccountID]stellar1.AccountBundle)
+	for _, parentEntry := range parent.Accounts {
+		if acctEncB64, ok := encodedBundle.AcctBundles[parentEntry.AccountID]; ok {
+			acctBundle, err := decodeAndUnboxAcctBundle(m, finder, acctEncB64, parentEntry)
+			if err != nil {
+				// XXX keep going?
```
mlsteele: xxx
I'd recommend not keep going. If you keep going the client is liable to post another bundle that has even less of whatever got corrupted.

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+				return nil, 0, err
+			}
+			if acctBundle != nil {
```
mlsteele: This seems scary. Dropping account bundle when the version is unrecognized. Seems like old clients could throw out accounts from new clients. Is that part of the plan since account bundles can survive without updates or something?

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+	puk, err := finder.SeedByGeneration(m, eab.Gen)
+	if err != nil {
+		return nil, err
+	}
+	ab, version, err := unbox(eab, hash, puk)
+	if err != nil {
+		return nil, err
+	}
+
+	// XXX version???
```
mlsteele: xxx
idk
but seems like eab.version should be asserted

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+// Does not check invariants.
+func accountEncrypt(bundle stellar1.AccountBundleSecretVersioned, pukGen keybase1.PerUserKeyGeneration, puk libkb.PerUserKeySeed) (res stellar1.EncryptedAccountBundle, resB64 string, err error) {
+	// Msgpack (inner)
+	clearpack, err := libkb.MsgpackEncode(bundle)
+	if err != nil {
+		return res, resB64, err
+	}
+
+	// Derive key
+	symmetricKey, err := puk.DeriveSymmetricKey(libkb.DeriveReasonPUKStellarBundle)
```
mlsteele: do we need all new derivation reasons for each new format?

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+	var visibleV2 stellar1.BundleVisibleV2
+	err = libkb.MsgpackDecode(&visibleV2, visiblePack)
+	if err != nil {
+		return empty, err
+	}
+	return merge(secretV2, visibleV2)
+}
+
+// decryptParent decrypts an encrypted parent bundle with the provided puk.
+func decryptParent(encBundle stellar1.EncryptedBundle, puk libkb.PerUserKeySeed) (stellar1.BundleSecretVersioned, error) {
```
mlsteele: this func is a dup of `bundle.Decrypt` with different error messages

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+		return nil, 0, err
+	}
+
+	var bundleOut stellar1.AccountBundle
+	switch version {
+	case stellar1.AccountBundleVersion_V1:
+		secretV1 := versioned.V1()
+		bundleOut = stellar1.AccountBundle{Signers: secretV1.Signers}
+	default:
+		return nil, 0, err
```
mlsteele: this is an unrecognized version should error right?

patrickxb:
@mlsteele I believe I addressed all your comments.  Can you take a look?  Thanks!

go/stellar/acctbundle/acctbundle.go
mlsteele: which already has tests in boxer_test.go

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,683 @@
...
+	for i, acctEntry := range visibleV2.Accounts {
+		secret, ok := a.AccountBundles[acctEntry.AccountID]
+		if !ok {
+			continue
+		}
+		ab, err := accountBoxAndEncode(acctEntry.AccountID, secret, pukGen, puk)
+		if err != nil {
+			return nil, err
+		}
+		if ab != nil {
```
mlsteele: seems like `ab` isn't nil-able so better crash than skip

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,683 @@
...
+			}
+
+			parent.AccountBundles[parentEntry.AccountID] = *acctBundle
+		}
+	}
+
+	return parent, parentVersion, nil
+}
+
+func decodeAndUnboxAcctBundle(m libkb.MetaContext, finder PukFinder, encB64 string, parentEntry stellar1.BundleEntryRestricted) (*stellar1.AccountBundle, error) {
```
mlsteele: `parentEntry` unused. Meant to check account ID and hash of encrypted account bundle?

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,683 @@
...
+		bundleOut = stellar1.AccountBundle{Signers: secretV1.Signers}
+	case stellar1.AccountBundleVersion_V2,
+		stellar1.AccountBundleVersion_V3,
+		stellar1.AccountBundleVersion_V4,
+		stellar1.AccountBundleVersion_V5,
+		stellar1.AccountBundleVersion_V6,
+		stellar1.AccountBundleVersion_V7,
+		stellar1.AccountBundleVersion_V8,
+		stellar1.AccountBundleVersion_V9,
+		stellar1.AccountBundleVersion_V10:
```
mlsteele: What are all these future version numbers for? There's no forward-compatible decoder for bundles

go/stellar/acctbundle/acctbundle.go
patrickxb: I added it.

go/stellar/acctbundle/acctbundle.go
patrickxb: They give an opportunity to give a different error when a newer version comes in to an old client.

go/stellar/acctbundle/acctbundle.go
patrickxb: thanks, fixing.

go/stellar/acctbundle/acctbundle.go
patrickxb: changed to use bundle.Decrypt

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,666 @@
...
+	return stellar1.BundleRestricted{
+		Revision: visible.Revision,
+		Prev:     visible.Prev,
+		Accounts: convertVisibleAccounts(visible.Accounts),
+		/*
+			AccountID: visible.AccountID,
+			Mode:      visible.Mode,
+			Signers:   secret.Signers,
+			Name:      secret.Name,
+		*/
```
mlsteele: `secret` unused, so missing name

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,666 @@
...
+	ab, _, err := unbox(eab, hash, puk)
+	if err != nil {
+		return nil, err
+	}
+	if ab.AccountID != parentEntry.AccountID {
+		return nil, errors.New("account bundle and parent entry account ID mismatch")
+	}
+	if !libkb.SecureByteArrayEq(ab.OwnHash, parentEntry.EncAcctBundleHash) {
+		return nil, errors.New("account bundle and parent entry hash mismatch")
+	}
```
mlsteele: Probably better to compare this hash earlier at L382 before decrypting. No reason to decrypt something that has the wrong hash.

mlsteele:
I promise those are the last 2 comments. I think this could use a pass to make the function names crystal clear but that can come way later

go/stellar/acctbundle/acctbundle.go
patrickxb: whoops...fixed.

go/stellar/acctbundle/acctbundle.go
patrickxb: done

patrickxb:
Thanks for all the reviewing!

go/stellar/remote/remote.go
```diff
@@ -168,6 +168,7 @@ func PostBundleRestricted(ctx context.Context, g *libkb.GlobalContext, bundle *s
 	section["visible_parent"] = boxed.VisParentB64
 	section["version_parent"] = boxed.FormatVersionParent
 	section["account_bundles"] = boxed.AcctBundles
+	section["miniversion_parent"] = 2
```
mlsteele: miniversion can go away now cause a v2 bundle can't have the miniversion 1 problem

go/stellar/stellarsvc/service_test.go
```diff
@@ -757,6 +758,199 @@ func TestRequestPaymentOutsideCurrency(t *testing.T) {
...
+
+	// tinker with it
+	acctBundle.Revision = 4
+	err = remote.PostBundleRestricted(context.Background(), tc.G, acctBundle)
+	require.Error(t, err)
+}
+
+// TestMakeAccountMobileOnlyOnRecentMobile imports a new secret stellar key, then
+// makes it mobile only.  The subsequent fetch fails because it is
+// a recently provisioned mobile device.  After 14 days, the fetch works.
```
xgess:
```suggestion
// a recently provisioned mobile device.  After 7 days, the fetch works.
```

go/stellar/remote/remote.go
xgess: @mlsteele it looks like this is here because the server is sharing parsing and validation code between the two different bundle formats. it's actually only validated as being `2` if it's set. so removing this line would be fine. any idea why patrick would have added it 11 days ago though?

go/stellar/acctbundle/acctbundle.go
```diff
@@ -0,0 +1,673 @@
...
+// with the api server to post and get account bundles.
+type BundleEncodedB64 struct {
+	EncParent           string                        `json:"encrypted_parent"` // base64 msgpacked Enc
+	VisParent           string                        `json:"visible_parent"`
+	FormatVersionParent stellar1.AccountBundleVersion `json:"version_parent"`
+	AcctBundles         map[stellar1.AccountID]string `json:"account_bundles"`
+}
+
+// BoxAndEncode encrypts and encodes a BundleRestricted object.
+func BoxAndEncode(a *stellar1.BundleRestricted, pukGen keybase1.PerUserKeyGeneration, puk libkb.PerUserKeySeed) (*BoxedEncoded, error) {
```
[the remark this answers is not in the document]
xgess: done

go/stellar/remote/remote.go
xgess: executive decision. miles and i have decided to remove this.

xgess:
@mlsteele anything else you think we need to do here before we start thinking about migration?

PR #14318, opened 2018-10-18
move saving route to config. 

chrisnojima:
In order to do this this also makes the sagas get initialized right after the engine is setup (else we start the root sagas and those can touch engine things before its initialized).


.eslintrc
```diff
@@ -56,6 +56,7 @@
         "jsx-quotes": "off",
         "no-duplicate-imports": "off",
         "no-mixed-operators": "off",
+        "no"-unused-expressions": ["error",  {"allowShortCircuit": true}],
```
chrisnojima: lets us do stuff like
```
blah?.whatever?.()
```

shared/actions/platform-specific/index.native.js
```diff
@@ -175,7 +174,10 @@ const updateChangedFocus = (action: ConfigGen.MobileAppStatePayload) => {
   return Saga.put(ConfigGen.createChangedFocus({appFocused}))
 }
 
-const clearRouteState = () => Saga.call(AsyncStorage.removeItem, 'routeState')
+const clearRouteState = () =>
```
chrisnojima: use config.json api

shared/actions/platform-specific/index.native.js
```diff
@@ -213,7 +220,11 @@ function* loadStartupDetails() {
   let startupLink = ''
   let startupTab = null
 
-  const routeStateTask = yield Saga.fork(AsyncStorage.getItem, 'routeState')
+  const routeStateTask = yield Saga.fork(() =>
```
chrisnojima: the previous code figures this all out in one shot very early so i had to change how we init sagas (below)

shared/actions/platform-specific/push.native.js
```diff
@@ -302,13 +302,14 @@ function* initialPermissionsCheck(): Saga.SagaGenerator<any, any> {
   } else {
     const shownNativePushPromptTask = yield Saga.fork(askNativeIfSystemPushPromptHasBeenShown)
     const shownMonsterPushPromptTask = yield Saga.fork(() =>
-      RPCTypes.configGetValueRpcPromise({path: `ui.${monsterStorageKey}`}).catch(() => false)
+      RPCTypes.configGetValueRpcPromise({path: `ui.${monsterStorageKey}`})
```
chrisnojima: better to convert it here vs later on

shared/app/index.native.js
```diff
@@ -29,13 +29,15 @@ class Keybase extends Component<any> {
 
     if (!global.keybaseLoaded) {
       global.keybaseLoaded = true
-      this.store = configureStore()
+      const {store, runSagas} = configureStore()
```
chrisnojima: configureStore returns 2 things now

shared/store/configure-store.js
```diff
@@ -116,6 +116,8 @@ export default function configureStore() {
     })
   }
 
-  runSagas()
```
chrisnojima: don't run this itself, just return it

chrisnojima:
@keybase/react-hackers

chrisnojima:
@keybase/react-hackers bump

chrisnojima:
@keybase/react-hackers bump (not for this release)
- [ ] after the release

chrisnojima:
@keybase/react-hackers bump. lint is failing cause of a fix in another pr so this can be looked at

PR #14324, opened 2018-10-18
stop node thread from knowing about react

chrisnojima:
@keybase/react-hackers this removes some dependencies between node thread and things like react


chrisnojima:
good idea, done

PR #14325, opened 2018-10-18
Persist that last currency set was XLM

zanderz:
Before, the result of the rpc to get the user's display currency was put directly into the building payment by the reducer. With this change, the send-form sets it after checking what the default should be.

Have not checked the persistence of lastSentXLM on mobile yet.

Seems to fix the requested send currency.


zanderz:
Tested on both ios and desktop @keybase/react-hackers

shared/actions/platform-specific/index.native.js
```diff
@@ -216,7 +218,8 @@ function* loadStartupDetails() {
   const routeStateTask = yield Saga.fork(AsyncStorage.getItem, 'routeState')
   const linkTask = yield Saga.fork(Linking.getInitialURL)
   const initialPush = yield Saga.fork(getStartupDetailsFromInitialPush)
-  const [routeState, link, push] = yield Saga.join(routeStateTask, linkTask, initialPush)
+  const lastSentXLMTask = yield Saga.fork(AsyncStorage.getItem, 'lastSentXLM')
```
chrisnojima: we're deprecating AsyncStorage (there's a pr open for this now) so lets not use it

shared/desktop/app/app-state.desktop.js
```diff
@@ -55,6 +56,7 @@ export default class AppState {
       height: 600,
       isFullScreen: null,
       isMaximized: null,
+      lastSentXLM: false,
```
chrisnojima: we shouldn't store this here. mobile and this should just use the config.json writing api in go

chrisnojima: https://github.com/keybase/client/pull/14318

shared/actions/platform-specific/index.desktop.js
```diff
@@ -81,6 +82,10 @@ const writeElectronSettingsNotifySound = (_: any, action: ConfigGen.SetNotifySou
   action.payload.writeFile &&
   SafeElectron.getIpcRenderer().send('setAppState', {notifySound: action.payload.sound})
 
+const writeLastSentXLM = (_: any, action: SetLastSentXLMPayload) =>
```
chrisnojima: the reason state is passed in first is cause the saga runs after the reducer, hence `state.wallet.lastSentXLM` already has the value in it, so you just can do
```
const writeLastSentXML = state => 
```

shared/actions/platform-specific/index.native.js
zanderz: ok, done

shared/desktop/app/app-state.desktop.js
zanderz: got it

shared/actions/platform-specific/index.desktop.js
zanderz: But we need the writeFile value too, right? to know whether this one is supposed to be persisted?

shared/actions/config/index.js
```diff
@@ -390,6 +391,24 @@ const updateServerConfig = (state: TypedState) =>
     }
   })
 
+const writeLastSentXLM = (_: any, action: SetLastSentXLMPayload) =>
+  Saga.call(function*() {
```
zanderz: not certain about doing it this way, it quiets down flow at least

zanderz:
ok ready for 1 more look

shared/actions/config/index.js
chrisnojima: just do actionToPromise. lastSentXLM is store in the store right, just grab it from there, you don't need the action at all

shared/actions/config/index.js
```diff
@@ -390,6 +391,24 @@ const updateServerConfig = (state: TypedState) =>
...
+    if (action.payload.writeFile) {
+      yield Saga.call(RPCTypes.configSetValueRpcPromise, {
+        path: 'stellar.lastSentXLM',
+        value: {isNull: false, b: action.payload.lastSentXLM},
+      })
+    }
+  })
+
+const readLastSentXLM = () =>
```
chrisnojima: use actionToPromise here as well

zanderz:
ok switched to actionToPromise() @keybase/react-hackers

shared/actions/config/index.js
```diff
@@ -390,6 +391,23 @@ const updateServerConfig = (state: TypedState) =>
     }
   })
 
+const writeLastSentXLM = (_: any, action: SetLastSentXLMPayload) => {
```
chrisnojima: the value is in the state already, just do
```
const writeLastSentXLM = (state) => {
...
value: {isNull:false, b: state.wallet.lastSentXLM}
```

zanderz: so you are saying to just get rid of writeFile and just write it every time? I was following the pattern of other stuff saved in app-state.json. Without the writeFile marker, it will re-write what it reads out on startup - maybe no big deal.

shared/actions/config/index.js
```diff
@@ -390,6 +391,23 @@ const updateServerConfig = (state: TypedState) =>
     }
   })
 
+const writeLastSentXLM = (state: TypedState, action: SetLastSentXLMPayload) => {
```
chrisnojima: can use the shorthand ```=> action.payload.writeFile...```
should also catch, we don't care if this fails so lets not black bar

shared/reducers/wallets.js
```diff
@@ -59,8 +59,9 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
         if (action.payload.accountID) {
           stateMutable.update('currencyMap', c => c.set(action.payload.accountID, action.payload.currency))
         }
+        const lastSentXLM = state.get('lastSentXLM')
```
chrisnojima: this can be just `state.lastSentXLM`

PR #14327, opened 2018-10-19
refactor rows in Files into its own component

songgao:
This is is purely a refactor PR, and *shouldn't* change any behaviors. This would be used in the move/copy modal dialog for Files. Since it moves stuff around, I was hoping to get this in first by its own to hopefully avoid conflicts.


songgao:
bump!

songgao:
⪮

shared/fs/index.js
```diff
@@ -2,156 +2,89 @@
...
-        return (
-          <WrapRow key="">
-            <Text type="BodySmallError">This should not happen.</Text>
-          </WrapRow>
-        )
-    }
-  }
-
   render() {
+    console.log({songgao: 'Files', path: this.props.path, props: this.props})
```
chrisnojima: 🔥

shared/fs/index.js
```diff
@@ -2,156 +2,89 @@
...
+          <Kb.Box style={styles.emptyContainer}>
+            <Kb.Text type="BodySmall">This is an empty folder.</Kb.Text>
+          </Kb.Box>
+        }
+      />
     )
     return (
-      <Box style={styleOuterContainer}>
-        <Box style={stylesContainer}>
+      <Kb.Box style={styles.outerContainer}>
```
chrisnojima: why not box2

shared/fs/row/rows-container.js
```diff
@@ -0,0 +1,193 @@
...
+
+const mergeProps = (s, d, o: OwnProps) => ({
+  items: getItemsFromStateProps(s, o.path, o.sortSetting),
+  routePath: o.routePath,
+  ifEmpty: o.ifEmpty,
+})
+
+export default compose(
+  FilesLoadingHoc,
+  // $FlowIssue @jzila/@songgao lots of exposed flow issues here
```
chrisnojima: this should get looked at at some point (not in this pr)

shared/fs/row/rows.js
```diff
@@ -0,0 +1,102 @@
...
+    )
+  }
+}
+
+const rowHeight = Styles.isMobile ? 64 : 40
+
+const styles = Styles.styleSheetCreate({
+  rowContainer: {
+    ...Styles.globalStyles.flexBoxColumn,
+    height: rowHeight,
```
chrisnojima: you should be able to do flexShrink: 0

shared/fs/index.js
songgao: This is the fire that warmed up the code freeze 😎

shared/fs/row/rows.js
songgao: Could you explain what `flexShrink` does here? I think the height was added for mobile for stable height; I don't remember for sure though.

chrisnojima: if you set a height and flexShrink: 0 it shouldn't reduce the height (get rid of minHeight/maxHeight)

songgao:
@chrisnojima ready for another 👀

shared/fs/index.js
chrisnojima: you should switch to box2 for all these and get rid of most of these styles

shared/fs/index.js
```diff
@@ -43,7 +42,7 @@ class Files extends React.PureComponent<FolderProps> {
     )
     return (
       <Kb.Box style={styles.outerContainer}>
-        <Kb.Box style={styles.container}>
+        <Kb.Box2 direction="vertical" fullHeight={true}>
```
chrisnojima: you can remove the container style

songgao:
@chrisnojima ping! 🏓

PR #14329, opened 2018-10-19
Stellar: Onboarding

cjb:
@keybase/react-hackers CC @mlsteele 

Lots of changes here!  Which makes many variations to test.

* Nav tab-bar tabs can now have NEW badges associated with them.  The one for Wallets is hooked up to Gregor -- either clicking on the tab, or the $ in the Chat input area will dismiss the NEW badge.

* The service now keeps track of whether we've accepted the Stellar disclaimer.  If we haven't, we don't have a wallet yet.  The new `getWalletSettings` action is fired on startup and finds out whether it's been accepted, this ends up in the store as `state.wallets.acceptedDisclaimer`.

* If we go to the Wallets tab, or the Send form inside of Chat, or the Transaction details page from Chat, and we haven't accepted the disclaimer, we render the `<Onboarding />` component on the Wallets tab instead.

* `<Onboarding />` asks whether we want to end up on the wallet list or the Link Existing view, after accepting the disclaimer, then prompts us to accept it after a five second delay.  Accepting it should show a spinner while our wallet is created.

* If we reject it or close the modal, we go back to whichever tab we came from.  We weren't storing the tab we were previously on, so now `state.routeTree.previousTab` exists to do that.  It's a little complicated on mobile, where the Wallets routes are underneath the Settings tab, so there we just back out to the root of the Settings tab.

* If we accept the disclaimer, we call `getWalletSettings` again to update `state.wallets.acceptedDisclaimer`, and then our views start redirecting to the right places.  If we'd selected "Link Existing" before accepting, we should end up there.

* There's still some UX cleanup needed: e.g. the buttons that are on the onboarding purple background need a new color style, there should be a gradient hinting that you can scroll down the disclaimer text, and the "What is Stellar?" link.  These can come in a separate PR.


cjb:
Oh, a note on testing: ideally you want a user that is in the whitelist (so, can create a wallet) which does not *already* have a wallet (because existing wallets have acceptedDisclaimer=true preset), and the way to get one of those is to make a test user and ask @mlsteele to add it to the whitelist.

shared/constants/types/wallets.js
```diff
@@ -4,6 +4,9 @@ import * as I from 'immutable'
 import HiddenString from '../../util/hidden-string'
 import * as StellarRPCTypes from './rpc-stellar-gen'
 
+// When accepting the Stellar disclaimer, next path after acceptance
+export type NextScreen = 'linkExisting' | 'openWallet'
```
buoyad: Could this be named something more specific?

shared/reducers/route-tree.js
```diff
@@ -202,8 +202,13 @@ export default function routeTreeReducer(state: Types.State = initialState, acti
     }
   }
 
+  // When moving tabs, record the old tab so we can go back to it later.
+  const oldTab = routeState && routeState.get('selected')
+  const newTab = newRouteState && newRouteState.get('selected')
+  const previousTab = oldTab && newTab && oldTab !== newTab && oldTab
```
buoyad: This is kind of hard to read, couldn't you just set this to `oldTab`?

shared/wallets/onboarding/disclaimer.js
```diff
@@ -0,0 +1,198 @@
...
+class Disclaimer extends React.Component<DisclaimerProps, DisclaimerState> {
+  state = {secondsLeft: 5}
+  timer: ?TickerID = null
+
+  _setSecondsLeft = (secondsLeft: number) => this.setState({secondsLeft})
+
+  tick = () => {
+    this.setState({secondsLeft: this.state.secondsLeft - 1}, () => {
+      if (this.state.secondsLeft === 0 && this.timer) {
+        removeTicker(this.timer)
```
buoyad: Need one of these in `componentWillUnmount` too

shared/wallets/onboarding/intro.js
```diff
@@ -0,0 +1,76 @@
...
+    </WalletPopup>
+  )
+}
+
+const styles = Styles.styleSheetCreate({
+  bodyText: {color: Styles.globalColors.white, marginBottom: Styles.globalMargins.small, textAlign: 'center'},
+  container: {backgroundColor: Styles.globalColors.purple2, padding: Styles.globalMargins.medium},
+  headerText: {
+    color: Styles.globalColors.white,
+    marginBottom: Styles.globalMargins.medium,
```
buoyad: :sheep: this wraps on small screens, `textAlign: 'center'`?
image

shared/wallets/onboarding/disclaimer.js
```diff
@@ -0,0 +1,198 @@
...
+
+type DisclaimerState = {|
+  secondsLeft: number,
+|}
+
+class Disclaimer extends React.Component<DisclaimerProps, DisclaimerState> {
+  state = {secondsLeft: 5}
+  timer: ?TickerID = null
+
+  _setSecondsLeft = (secondsLeft: number) => this.setState({secondsLeft})
```
buoyad: is this used anywhere?

cjb:
@buoyad says (in chat) that merging now and finishing review later once it's in master sounds good, so I'm gonna do that now and kick some builds.

PR #14332, opened 2018-10-19
Make keybase.desktop Linux shortcut not kill Keybase by default

aljelly:
When you install Keybase on Linux it runs in the background on system boot. The `run_keybase` kills all the Keybase processes and starts them up again, so if Keybase is running it restarts it.

When I click the Keybase shortcut, I know Keybase is already running in the background and I'd expect it to just show the GUI and not restart Keybase. However with the current `keybase.desktop` shortcut on Linux, when you click on the shortcut, it runs `run_keybase`. Because this command involves killing all Keybase processes, it can be problematic and surprising, as I've found on one occasion.

The biggest problem is, what if you're uploading something to KBFS? If you've moved the files to KBFS, and then say someone's sent you a message on Keybase, and you go to click the shortcut in the app launcher to see it, Keybase will restart and the files you were uploading will be lost. Even if you're not uploading anything, it takes slightly longer for the Keybase GUI to show up when you click the shortcut (and it has to load too) because it's restarting the whole thing. This is especially a problem because some distros are either planning to or already have removed the systray. I recently upgraded to elementary OS Juno and they've already done this, so the only way I can open Keybase in a user friendly way is to click on the launcher.

The PR basically changes the launcher to point to the Keybase GUI executable instead. It now just brings up the Electron UI. If the Keybase GUI is already up, it does nothing. I don't think this would create any problems, unless there are any circumstances where the Keybase GUI executable won't get installed to `/opt/keybase/Keybase`. As a bonus, showing the GUI is slightly quicker than it was previously.

`run_keybase` is still there in the form of a "Restart" action (which more accurately represents what it does) which (at least on my distro) shows as an option when you right click the icon.

Thank you for Keybase, by the way!

### DIFF
--- a/packaging/linux/keybase.desktop
+++ b/packaging/linux/keybase.desktop
@@ -1,8 +1,13 @@
 [Desktop Entry]
 Name=Keybase
-Exec=run_keybase
+Exec=/opt/keybase/Keybase
 Icon=keybase
 Terminal=false
 Type=Application
 Categories=Network;
 StartupNotify=false
+Actions=restart;
+
+[Desktop Action restart]
+Name=Restart
+Exec=run_keybase


eli-schwartz:
That's sort of broken, since the keybase GUI depends on the daemonized kbfs and keybase helper processes and this change would result in them never being started at all.

The appropriate way to deal with this is, of course, that the run_keybase script should stop trying to do altogether too many things at once. I solved this by replacing run_keybase with my own wrapper script https://git.archlinux.org/svntogit/community.git/tree/trunk/keybase-gui?h=packages/keybase and shipping this for my distro package.

PR #14335, opened 2018-10-19
ignore sig_id sent down from server

maxtaco:
- this is a follow on to an earlier bug with signature chain replay, in which a badLinks table was indexed by a sketchy sigID
- we left sigIDs in for the way-less-risky badWhitespace links, but it still can leak out into other parts of the program.
- this PR removes all untrusted `sig_id`s sent down from the server, and will only read into unpacked.sigID on retrieval of objects from disk
- otherwise, we derive the `sigID` from the verifiable data (the signature)

### DIFF
--- a/go/libkb/chain_link.go
+++ b/go/libkb/chain_link.go
@@ -167,54 +167,54 @@ var badChainLinks = map[keybase1.LinkID]SpecialChainLink{
 // hash in subsequent chainlinks.  Caused by bad code on 15 Sep 2015.
 const whitespaceIssue20150915 = "Bad whitespace stripping on 15 Sep 2015"
 
-var badWhitespaceChainLinks = map[keybase1.SigID]string{
-	"595a73fc649c2c8ccc1aa79384e0b3e7ab3049d8df838f75ef0edbcb5bbc42990f": whitespaceIssue20150915,
-	"e256078702afd7a15a24681259935b48342a49840ab6a90291b300961669790f0f": whitespaceIssue20150915,
-	"30831001edee5e01c3b5f5850043f9ef7749a1ed8624dc703ae0922e1d0f16dd0f": whitespaceIssue20150915,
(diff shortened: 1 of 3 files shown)


go/libkb/chain_link.go
```diff
@@ -909,6 +899,19 @@ func (c *ChainLink) Unpack(m MetaContext, trusted bool, selfUID keybase1.UID, pa
 		if i, err := jsonparserw.GetInt(packed, "disk_version"); err == nil {
 			c.diskVersion = int(i)
 		}
+
+		// It is not acceptable to digest sig_id from the server, but we do derived it
```
patrickxb: derived -> derive

go/libkb/sig_chain.go
```diff
@@ -522,7 +522,7 @@ const resetReason = "hardcoded reset"
 
 var hardcodedResets = map[keybase1.LinkID]SpecialChainLink{
 	"f6dae096194690cfee8974b0e10a99ecac2cc8e4f9383516a1f626f614e566e0": SpecialChainLink{UID: keybase1.UID("2d5c41137d7d9108dbdaa2160ba7e200"), Seqno: keybase1.Seqno(11), Reason: resetReason},
-	"8d7c1a0c99186f972afc5d3624aca2f88ddc3a5dbf84e826ef0b520c31a78aa3": SpecialChainLink{UID: keybase1.UID("f1c263462dd526695c458af924977719"), Seqno: keybase1.Seqno(18), Reason: resetReason},
+	"8d7c1a0c99186f972afc5d3624aca2f88ddc3a5dbf84e826ef0b520c31a78aa3": SpecialChainLink{UID: keybase1.UID("f1c263462dd526695c458af924977719"), Seqno: keybase1.Seqno(8), Reason: resetReason},
```
maxtaco: I couldn't ID this user due to this typo.

PR #14336, opened 2018-10-19
desktop notifications for reactions to your own messages CORE-9184

mmaxim:


mmaxim:
@malgorithms @cecileboucheron @adamjspooner  This turns on desktop notifications for reactions to your own messages. It say "<username> reacted to your message." in the toast. Any opinion on getting this in for Monday? Do you want to try it out a little before turning it on for everyone?

go/chat/utils/utils.go
```diff
@@ -889,13 +906,20 @@ func GetDesktopNotificationSnippet(conv *chat1.ConversationLocal, currentUsernam
...
 	if err != nil || !msg.IsValid() {
 		return ""
 	}
 	mvalid := msg.Valid()
+	var snippet string
 	if !mvalid.IsEphemeral() {
-		snippet, _ := GetMsgSnippet(msg, *conv, currentUsername)
+		switch msg.GetMessageType() {
+		case chat1.MessageType_REACTION:
+			snippet = "reacted to your message."
```
adamjspooner: Do we want to show what they reacted with?

adamjspooner:
This is great! I don't have a strong opinion about getting it in by Monday. I'm game if it works out.

go/chat/utils/utils.go
mmaxim: Turned out to be harder than I expected. We store something like :+1: in the actual message text for the reaction, but only the JS code knows how to turn that into an actual emoji. If we like this, we can try and plumb all that through.

adamjspooner: Ah, gotcha. Knowing how they reacted would be super helpful (so you don't have to go look at the message just to see the emoji), but it's not a deal-breaker.

mmaxim: I can trivially put the colon in there, but I wasn't sure if people would like that.

mmaxim:
https://github.com/keybase/keybase/pull/3044 is a PR on the server side for mobile push notifications for these as well.

go/chat/utils/utils.go
joshblum: we have a library vendored for this actually for the cli https://github.com/keybase/client/blob/master/go/client/chat_cli_rendering.go#L686

go/chat/utils/utils.go
```diff
@@ -486,6 +488,21 @@ func FilterExploded(expunge *chat1.Expunge, msgs []chat1.MessageUnboxed, now tim
 	return res
 }
 
+func GetReaction(msg chat1.MessageUnboxed) (string, error) {
```
joshblum: not needed?

go/chat/utils/utils.go
```diff
@@ -360,20 +360,22 @@ func IsVisibleChatMessageType(messageType chat1.MessageType) bool {
...
-	if len(atMentions) > 0 || chanMention != chat1.ChannelMention_NONE {
+	switch messageType {
+	case chat1.MessageType_EDIT:
+		// an edit with atMention or channel mention should generate notifications
+		if len(atMentions) > 0 || chanMention != chat1.ChannelMention_NONE {
+			return true
+		}
+	case chat1.MessageType_REACTION:
+		// effect of this is all reactions will notify if they are sent to a person that
+		// is notified for any messages in the conversation
```
joshblum: some changes are missing here but are in the server pr: https://github.com/keybase/keybase/pull/3044/files#diff-0c41f3030b7365bba4b704fc82318c25R1358

joshblum:
fwiw +1 for monday!

go/chat/utils/utils.go
mmaxim: Those are changes that I made in some old client PR that never got vendored into the server, but are now coming for the ride.

go/chat/utils/utils.go
mmaxim: Kept it for the future when the reaction is included in the toast.

go/chat/utils/utils.go
joshblum: ah makes sense just wanted to double check

go/chat/utils/utils.go
```diff
@@ -916,7 +918,12 @@ func GetDesktopNotificationSnippet(conv *chat1.ConversationLocal, currentUsernam
 	if !mvalid.IsEphemeral() {
 		switch msg.GetMessageType() {
 		case chat1.MessageType_REACTION:
-			snippet = "reacted to your message."
+			reaction, err := GetReaction(msg)
```
joshblum: should we do a similar check for TEXT to make sure the body type matches the header type? otherwise we can ditch that method to just do `Valid().MessageBody().Reaction()` directly

mmaxim: I guess.

PR #14326, opened 2018-10-18
SendForm: route to screens on mobile and display in Root on desktop

buoyad:
- Add 'choose-asset' and 'confirm' to possible screens
- Route on mobile; show directly on desktop
r? @keybase/react-hackers

PR #14338, opened 2018-10-19
Add openSendRequestForm action to prepare & navigate to the form

buoyad:
Requires #14326 

* Adds an action `openSendRequestForm` that takes all the possible `setBuilding*` params and dispatches the appropriate actions and load calls. Switches all places we nav to a send/request form to use that action.
* Makes `GetDisplayCurrencyLocal` param `accountID` optional and use the default accountID if none is given.

r? @keybase/react-hackers

### DIFF
--- a/protocol/avdl/stellar1/local.avdl
+++ b/protocol/avdl/stellar1/local.avdl
@@ -355,7 +355,7 @@ protocol local {
 
     string amount, // Amount of the asset OR outside currency being requested. Example: "3.005"
 
-    // Exactly one of `asset` and `currency` should be set.
+    // At most one of `asset` and `currency` should be set.
     union { null, Asset } asset,
     // Note: Requesting non-XLM assets is not yet supported. (Miles: It may
     // actually be supported, but that would be a weird asymmetry)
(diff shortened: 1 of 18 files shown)


protocol/avdl/stellar1/local.avdl
```diff
@@ -201,7 +201,8 @@ protocol local {
   // OutsideCurrencyCode examples: "USD", "EUR". Has to be one of
   // supported currencies, returned from getDisplayCurrenciesLocal.
   void changeDisplayCurrencyLocal(int sessionID, AccountID accountID, OutsideCurrencyCode currency);
-  CurrencyLocal getDisplayCurrencyLocal(int sessionID, AccountID accountID);
+  // Gets display currency for primary account if no accountID is given.
+  CurrencyLocal getDisplayCurrencyLocal(int sessionID, union { null, AccountID } accountID);
```
zapu: @patrickxb OK with API change?

patrickxb: It seems fine, but just so I know, what path in the frontend needs to call this when it doesn't have an account id?

buoyad: @patrickxb it's for prefilling the send form with the default display currency. When coming from chat / profile we haven't loaded any account IDs. When coming from the wallets tab we'll still pass in `accountID`

patrickxb: Ok.  

Isn't GetWalletAccountsLocal() called pretty early on when the app starts up?  Then you would have the primary account ID.

buoyad: We did briefly call that on startup but we don't anymore

zanderz:
LGTM

PR #14341, opened 2018-10-19
add infrastructure for handling phone-related messages from gregor

xgess:
see https://github.com/keybase/keybase/pull/3042 for where these messages are coming from

* receive and dismiss the three phone messages from gregor
* fire notifications for each

i made them three separate ones because i figured the UI would want to handle them in potentially very different ways. mainly superseded. if that one comes in, you probably really want to know about it.  but i'm happy to collapse this down into a single notification with the type of change that happened.


go/service/phone_numbers.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/keybase/client/go/gregor"
+	"github.com/keybase/client/go/libkb"
+	gregor1 "github.com/keybase/client/go/protocol/gregor1"
+	keybase1 "github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/go-framed-msgpack-rpc/rpc"
+
+	"golang.org/x/net/context"
+)
+
+type PhoneNumbersHandler struct {
```
zapu: if we have a new file and new type for handling phone numbers, can we also move 
```
func (h *UserHandler) AddPhoneNumber(ctx context.Context, arg keybase1.AddPhoneNumberArg)
func (h *UserHandler) VerifyPhoneNumber(ctx context.Context, arg keybase1.VerifyPhoneNumberArg)
func (h *UserHandler) GetPhoneNumbers(ctx context.Context, sessionID int)
```
here, from `service/user.go`

protocol/avdl/keybase1/phone_numbers.avdl
```diff
@@ -0,0 +1,19 @@
...
+
+    record PhoneNumberVerifiedMsg {
+        @jsonkey("phone")
+        string phoneNumber;
+    }
+
+    record PhoneNumberSupersededMsg {
+        @jsonkey("phone")
+        string phoneNumber;
+    }
```
zapu: BTW you can also just do one msg for all 3, like `record PhoneNumberMsg` or something like that. Your call.

xgess: yeah i mentioned that in the PR description. i initially did that, and thought the front end might want to treat them very differently (e.g. add phone number, don't care. superseded phone number: notify right away). i'm extremely indifferent.

go/service/phone_numbers.go
xgess: excellent idea.

go/client/rpc.go
```diff
@@ -145,6 +145,14 @@ func GetUserClient(g *libkb.GlobalContext) (cli keybase1.UserClient, err error)
 	return
 }
 
+func GetPhoneNumbersClient(g *libkb.GlobalContext) (cli keybase1.PhoneNumbersClient, err error) {
+	var rcli *rpc.Client
+	if rcli, _, err = GetRPCClientWithContext(g); err == nil {
+		cli = keybase1.PhoneNumbersClient{Cli: rcli}
+	}
+	return
+}
```
zapu: man, can you imagine how neat this file could be if we had generics or any sort of templating in go? one can dream

go/systests/phone_number_test.go
```diff
@@ -110,3 +110,97 @@ func TestInvalidPhoneNumberResolve(t *testing.T) {
...
+	require.NoError(t, err)
+	cli2 := &client.CmdVerifyPhoneNumber{
+		Contextified: libkb.NewContextified(ann.tc.G),
+		PhoneNumber:  phone,
+		Code:         code,
+	}
+	err = cli2.Run()
+	require.NoError(t, err)
+	ann.drainGregor()
+	require.Equal(t, annListener.verifiedPhones, expectedNotification)
```
zapu: maybe one thing if you still have this branch checked out somewhere (if not just hit merge):
I'd add something like 
```
require.Equal(t, annListener.supersededPhones, []keybase1.PhoneNumber{})
```
to ensure other notifications did not sneak in unexpected.

xgess: 👍

go/client/rpc.go
xgess: lol. it would not need to be a whole file.

PR #14342, opened 2018-10-19
phone/email home screen - client

heronhaye:
More or less done but we need new icons from design. And the routes are not great either because you can't actually do any of these things in mobile. So this is blocked until later.


zapu:
cc @keybase/react-hackers

zapu:
are there reasons not to merge this? I was a bit anxious after last TODO badge problems but after looking into TODO system again for "noemail" todos, I think we are safe.

chrisnojima:
Why merge this if it's not done. We're releasing Monday

zapu:
ok

heronhaye:
We don't have icons

PR #14347, opened 2018-10-21
fix the out-of-date banner in widget

songgao:
* Tagging @keybase/kbfs-hackers for the first commit in Go:

Instead of just calling `libcmdline.SpawnDetachedProcess`, use `exec.Command` and call `Start()` on macOS. The former doesn't work with the updater; the process just dies. Perhaps it has to do with the GUI prompt. The latter works, and I've tested to make sure the process continues to run even after the parent process dies.

Unfortunately the same doesn't apply on Windows. If the cmd window running the parent process is closed, the spawned process exits too. So on Windows, keep using the old SpawnDetachedProcess. I haven't tried on Windows, but plan to do so after this is merged and an admin build is out.

* Tagging @keybase/react-hackers for the second commit:

Introduce an `updating` flag in the `outOfDate` object and use that to determine if we should show the "Update Now" link or "Updating ..." text.

* Misc

Note that `updater check` and `updater need-update` ignores the snooze now, so even if user has clicked "Ignore" in the prompt, the banner still shows, and clicking "Update Now" will trigger the updater prompt.


go/install/install_windows.go
```diff
@@ -328,3 +331,28 @@ func getCachedPackageModifyString(log Log) (string, error) {
 	log.Debug("getCachedPackageModifyString: " + `SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall` + bundleKey + "displayName " + displayName + ", publisher " + publisher)
 	return "", errors.New("no cached package path found")
 }
+
+// StartUpdateIfNeeded starts to update the app if there's one available. It
+// calls `updater check` internally so it ignores the snooze.
+func StartUpdateIfNeeded(ctx context.Context, log logger.Logger) error {
```
zanderz: Windows part LGTM, here - this was just moved over, right?

songgao: Moved over, and added the `Wait()`. But yeah, it's mostly the same code for Windows. I still need to test it on Windows after the build is out and see if update actually works.

songgao:
@keybase/react-hackers still need a thumb on the react part; thanks!

PR #14348, opened 2018-10-22
add left/right hotkeys in attachment fullscreen view to cycle images/videos

mmaxim:
@keybase/react-hackers 

Adds a new feature where users can hit the left/right arrow keys when in fullscreen attachment mode to move through images/videos on desktop. cc @adamjspooner @malgorithms @cecileboucheron. This is based partly off the work that @adamjspooner started. If you all like this we could add arrows to the design, and we could see how hard it would be to work on mobile too. 

**Go**
1.) Add a new type `attachments.Gallery` which provides the `NextMessage` method. This method will take a message, a direction, and whether or not to limit to only image/video and produce the next image in the given direction.
2.) The only complicated thing here is the "forward in time" scan, which is slightly awkward in the paging system we have.
3.) Fix a bug in the block engine storage where it would scan past the start point for a "prev" style scan. The server would not do this, and so without the new `minID` parameter the behavior is inconsistent between the local storage and the server database storage. We didn't notice this because nothing really uses this scan direction now.

**JS**
1.) Change the attachment fullscreen view to pull the attachment directly from the store instead of the route props.
2.) Add new actions to control which attachment the fullscreen view is pointing at. Hook them into the new service RPC for getting the next attachment.
3.) Change some of the existing actions to pass `Message` objects around instead of `convID,ordinal` pairs since the attachment message we pull in the fullscreen view might not be in the message map. This enables all the download/show in finder stuff to keep working.
4.) Upgrade the fullscreen image viewer to show video (could drop the other video component I put in), and respond to left/right keys.

Demo:
demo


go/chat/attachments/gallery.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+		if !opts.ImagesOnly {
+			return true
+		}
+		md := body.Attachment().Object.Metadata
+		atyp, err := md.AssetType()
+		if err != nil {
+			return false
+		}
+		return atyp == chat1.AssetMetadataType_IMAGE || atyp == chat1.AssetMetadataType_VIDEO
```
joshblum: we should be checking `ImagesOnly` here too right? maybe
```go
switch atype {
case chat1.AssetMetadataType_IMAGE:
      return true
case chat1.AssetMetadataType_VIDEO:
      return !opts.ImageOnly
default:
     return false
```

go/chat/attachments/gallery.go
```diff
@@ -0,0 +1,116 @@
...
+				Recent: true,
+			})
+			tv, err := g.G().ConvSource.Pull(ctx, convID, uid, chat1.GetThreadReason_GENERAL,
+				&chat1.GetThreadQuery{
+					MessageTypes: []chat1.MessageType{chat1.MessageType_ATTACHMENT},
+				}, pagination)
+			if err != nil {
+				return res, err
+			}
+			if len(tv.Messages) > 0 && tv.Messages[0].GetMessageID() != msgID {
```
joshblum: if `tv.Messages[0].GetMessageID() == msgID` why do we skip this whole page?

go/chat/attachments/gallery.go
```diff
@@ -0,0 +1,116 @@
...
+			return true
+		}
+		md := body.Attachment().Object.Metadata
+		atyp, err := md.AssetType()
+		if err != nil {
+			return false
+		}
+		return atyp == chat1.AssetMetadataType_IMAGE || atyp == chat1.AssetMetadataType_VIDEO
+	}
+	if opts.BackInTime {
```
joshblum: i think these blocks would be easier to read if `opts.BackInTime` just determined the pagination (maybe `Pull` could  take a `reversed` param so the loops could just iterate over the response?

go/chat/attachments/gallery.go
```diff
@@ -0,0 +1,116 @@
...
+	utils.DebugLabeler
+
+	PrevStride, NextStride int
+}
+
+func NewGallery(g *globals.Context) *Gallery {
+	return &Gallery{
+		Contextified: globals.NewContextified(g),
+		DebugLabeler: utils.NewDebugLabeler(g.GetLog(), "Attachments.Gallery", false),
+		NextStride:   10,
```
joshblum: why the difference?

go/chat/attachments/gallery.go
mmaxim: Video counts, I just made the param name shorter. "Image" means anything the fullscreen viewer can see.

go/chat/attachments/gallery.go
mmaxim: Probably not needed, I put it in there to guard against the case that I fixed in storage_blockengine.go

go/chat/attachments/gallery.go
mmaxim: I can swing by to discuss this.

go/chat/attachments/gallery.go
mmaxim: And this

cecileboucheron:
Can't wait to have this in the app.

mmaxim:
@joshblum PTAL

go/chat/attachments/gallery.go
```diff
@@ -29,88 +31,83 @@ func NewGallery(g *globals.Context) *Gallery {
...
+func (g *Gallery) eligibleNextMessage(msg chat1.MessageUnboxed, opts NextMessageOptions) bool {
+	if !msg.IsValid() {
+		return false
+	}
+	body := msg.Valid().MessageBody
+	typ, err := body.MessageType()
+	if err != nil || typ != chat1.MessageType_ATTACHMENT {
+		return false
+	}
+	if !opts.ImagesOnly {
```
joshblum: `MediaOnly`?

mmaxim:
I'm going to change it to take a list of asset types.

mmaxim:
ok @keybase/react-hackers this is ready to look at when you get a chance.

shared/actions/chat2/index.js
```diff
@@ -1711,7 +1710,57 @@ function* attachmentDownload(action: Chat2Gen.AttachmentDownloadPayload) {
...
+    convID: Types.keyToConversationID(conversationIDKey),
+    messageID,
+    backInTime,
+    assetTypes: [RPCChatTypes.localAssetMetadataType.image, RPCChatTypes.localAssetMetadataType.video],
+    identifyBehavior: RPCTypes.tlfKeysTLFIdentifyBehavior.chatGui,
+  })
+  if (nextAttachmentRes.message) {
+    return Constants.uiMessageToMessage(state, conversationIDKey, nextAttachmentRes.message)
+  }
+  yield Saga.put(Chat2Gen.createAttachmentFullscreenSelection({message: currentFullscreen}))
```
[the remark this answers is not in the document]
mmaxim: I guess.

shared/actions/chat2/index.js
```diff
@@ -2694,6 +2725,7 @@ function* chat2Saga(): Saga.SagaGenerator<any, any> {
   yield Saga.safeTakeEvery(Chat2Gen.attachmentDownload, attachmentDownload)
   yield Saga.safeTakeEvery(Chat2Gen.attachmentsUpload, attachmentsUpload)
   yield Saga.safeTakeEvery(Chat2Gen.attachmentPasted, attachmentPasted)
+  yield Saga.safeTakeEvery(Chat2Gen.attachmentFullscreenNext, attachmentFullscreenNext)
```
chrisnojima: we should move to using actionToAction (in this case) and actionToPromise in simpler cases

shared/actions/chat2/index.js
```diff
@@ -1711,7 +1710,34 @@ function* attachmentDownload(action: Chat2Gen.AttachmentDownloadPayload) {
 
   // Download it
   const destPath = yield Saga.call(downloadFilePath, message.fileName)
-  yield Saga.call(downloadAttachment, destPath, conversationIDKey, message, ordinal)
+  yield Saga.call(downloadAttachment, destPath, message)
+}
+
+function* attachmentFullscreenNext(action: Chat2Gen.AttachmentFullscreenNextPayload) {
```
chrisnojima: if this is actionToAction you get state passed in and you don't need to call select

shared/chat/conversation/attachment-fullscreen/index.desktop.js
```diff
@@ -56,18 +59,35 @@ class _Fullscreen extends React.Component<Props & OverlayParentProps, State> {
...
-              <OrientedImage
-                src={this.props.path}
-                style={this.props.isZoomed ? styleImageZoom : styleImageFit}
-                onLoad={this._setLoaded}
-              />
+              {!this.props.isVideo ? (
+                <OrientedImage
+                  src={this.props.path}
+                  style={this.props.isZoomed ? styleImageZoom : styleImageFit}
+                  onLoad={() => this._setLoaded(this.props.path)}
```
chrisnojima: this isn't safe, you need to wrap this in a check for _mounted

shared/reducers/chat2.js
```diff
@@ -265,24 +265,30 @@ const messageMapReducer = (messageMap, action, pendingOutboxToOrdinal) => {
...
-        if (!message || message.type !== 'attachment') {
+      return messageMap.updateIn(
+        [action.payload.message.conversationIDKey, action.payload.message.ordinal],
+        message => {
+          if (!message || message.type !== 'attachment') {
+            return message
+          }
+          const path = action.error ? '' : action.payload.path
           return message
+            .set('downloadPath', path)
```
chrisnojima: prefer using merge in this type of case

shared/reducers/chat2.js
```diff
@@ -835,9 +841,30 @@ const rootReducer = (
...
+      if (
+        state.attachmentFullscreenMessage &&
+        state.attachmentFullscreenMessage.conversationIDKey === message.conversationIDKey &&
+        state.attachmentFullscreenMessage.id === message.id &&
+        message.type === 'attachment'
+      ) {
+        nextState = nextState.set('attachmentFullscreenMessage', message.set('downloadPath', path))
+      }
+      return nextState.withMutations(s => {
+        s.set('metaMap', metaMapReducer(state.metaMap, action))
```
chrisnojima: can use merge and remove withMutations/set

shared/reducers/chat2.js
mmaxim: Ok, all I did was change the indent here, I didn't actually change this.

shared/reducers/chat2.js
mmaxim: I just copy/pasted this up from how it was being handled before.

shared/chat/conversation/attachment-fullscreen/index.desktop.js
mmaxim: I'm not sure I understand why, what is unsafe here? Isn't this just the same as before but with the path being passed through?

chrisnojima: so i think we should clean up if we're bothering to touch some of this stuff so, even if its not something you changed per se we likely want to update it. onLoad can be called at any point, even after this thing is unmounted so its never safe to allow that with setState calls

shared/actions/chat2/index.js
mmaxim: I'm not sure how to use it here, the action that would be returns from one of these `actionToAction` handlers depends on the result of an RPC call. That is, I need to call the RPC, get its response, and then dispatch the correct action depending on the result. I'm not sure how to do that in something like `Saga.sequentially` or anything like that that wouldn't make this code harder to read.

shared/actions/chat2/index.js
chrisnojima: you can still return a full blown generator in a call
```
const attachmentFullscreenNext = (state: TypedState, action: Chat2Gen.AttachmentFullscreenNextPayload) => Saga.call(function*() {
...
})
```

mmaxim: If I do that then what do I gain over what is there now? Just having the state parameter?

chrisnojima: eventually we'll deprecate the safeTake*. this newer api helps enforce better practices but still gives you an escape hatch if you need to do more complicated async ops

mmaxim:
Ok made some changes.

PR #14354, opened 2018-10-22
plumb through the resolved conversationIDkey and set an isPending flag

chrisnojima:
@keybase/react-hackers this fixes you trying drag/drop or paste an image into a conversation from the new convo flow.

Steps to repro:
1. create new convo
1. select a person who you have an existing convo with
1. drag file or paste image

nothing happens

The reason this happens is we do bookkeeping of the convo and we treat it as a special value PENDING if we're in the middle of building the convo. We also want previews to work so we have a way to resolve the pending to an existing conversation (if it exists)

So before the conversationIDKey in this case would flow down as PENDING, then all the consumers of it would have to check if its pending and act accordingly. But really only a few pieces are pending-aware (the header shows the search bar instead, the input bar does different things).

So now we pass the resolved conversationIDKey so most things act like you're in a regular convo and we additionally pass a new flag `isPending` which tells components it really is a search state

### DIFF
--- a/shared/chat/conversation/container.js
+++ b/shared/chat/conversation/container.js
@@ -12,6 +12,7 @@ import Rekey from './rekey/container'
 
 type SwitchProps = {
   conversationIDKey: Types.ConversationIDKey,
+  isPending: boolean,
   type: 'error' | 'noConvo' | 'rekey' | 'youAreReset' | 'normal' | 'rekey',
 }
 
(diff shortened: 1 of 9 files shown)


chrisnojima:
cc: @mmaxim

zanderz:
LGTM. Just want to make sure the comparison is still valid in inbox/new-conversation/container.js:
`isSelected: conversationIDKey === Constants.pendingConversationIDKey,`

PR #14356, opened 2018-10-22
update electron to 305 and remove applescript. move back to using electrons method for login items

chrisnojima:
@keybase/react-hackers

### DIFF
--- a/shared/desktop/app/app-state.desktop.js
+++ b/shared/desktop/app/app-state.desktop.js
@@ -133,103 +133,10 @@ export default class AppState {
       logger.info('Skipping auto login state change due to dev env. ')
       return
     }
-    // Comment this out if you want to test auto login stuff
 
-    const isDarwin = process.platform === 'darwin'
-    const isWindows = process.platform === 'win32'
-    // Electron has a bug where app.setLoginItemSettings() to false fails!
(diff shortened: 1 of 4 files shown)


zanderz:
could get rid of `getDarwinAppName` too if this works

shared/desktop/app/app-state.desktop.js
```diff
@@ -133,105 +133,12 @@ export default class AppState {
...
-            }
-          )
-        } catch (e) {
-          logger.info('Error setting apple startup prefs: ', e)
-        }
-      }
-    } catch (e) {
-      logger.info('Error setting apple startup prefs: ', e)
+    if (process.platform === 'darwin' || process.platform === 'win32') {
+      SafeElectron.getApp().setLoginItemSettings({openAtLogin: !!this.state.openAtLogin})
```
buoyad: Can we log this?

zanderz:
I also tested on macos too

chrisnojima:
it skips on dev, you can comment that out

zanderz:
no I mean even after commenting out our __DEV__ test. If we're shipping with this I'll just sanity check this part of the build first.

zanderz:
Test installer with this merged into master correctly sets startup key

PR #14362, opened 2018-10-22
Add cancelable / canceled states to relay payment views

buoyad:
There are still some smaller things to do but I'm going to split those off because this is getting big.

- Adds two `PaymentStatus`s, `claimable` and `canceled` to represent cancelable/canceled respectively.
- Adds `{PaymentLocal, PaymentDetailsLocal}.originalToAssertion` to store who the payment was originally for in the case that `status === 'canceled'`. Also make the payment look more like a direct payment in general
- Makes canceled payments look more like direct payments (with strikethroughs)
- Adds cancel buttons / explainers in chat messages and transaction rows
- Adds claim buttons leading to disclaimer for receiving users that haven't accepted yet
- Tweaks wording/icons/colors for these pretty much everywhere

r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/account-payment/index.js
+++ b/shared/chat/conversation/messages/account-payment/index.js
@@ -48,13 +48,13 @@ const AccountPayment = (props: Props) => {
       <Box2
         direction="horizontal"
         fullWidth={true}
-        style={collapseStyles([styles.alignItemsCenter, styles.flexWrap])}
+        style={collapseStyles([
+          styles.alignItemsCenter,
+          styles.flexWrap,
+          {marginBottom: globalMargins.xtiny},
+        ])}
       >
-        <Box2
-          direction="horizontal"
-          gap="xtiny"
-          style={collapseStyles([styles.alignItemsCenter, {marginBottom: globalMargins.xtiny}])}
-        >
+        <Box2 direction="horizontal" gap="xtiny" gapEnd={true} style={styles.alignItemsCenter}>
           <Icon type={props.icon} color={globalColors.purple2} fontSize={12} />
           <Text
             type="BodySmall"
@@ -66,8 +66,8 @@ const AccountPayment = (props: Props) => {
             </Text>
             {props.pending ? '...' : '.'}
           </Text>
-          {props.canceled && <Text type="BodySmall">CANCELED</Text>}
         </Box2>
+        {props.canceled && <Text type="BodySmall">CANCELED</Text>}
         {!!props.balanceChange && (
           <Box2 direction="horizontal" style={styles.marginLeftAuto}>
             <Text


protocol/avdl/stellar1/local.avdl
```diff
@@ -127,8 +128,8 @@ protocol local {
     PaymentID id;
     TransactionID txID;             // For relay payments, the funding tx.
     TimeMs time;
-    PaymentStatus statusSimplified; // we are working on making this correct for relay payments...
-    string statusDescription;       // "pending", "completed", "error"
+    PaymentStatus statusSimplified;
+    string statusDescription;       // "pending", "completed", "error", "claimable", "canceled"
```
mlsteele: same q about duplicate fields

protocol/avdl/stellar1/local.avdl
```diff
@@ -67,8 +67,8 @@ protocol local {
   record PaymentLocal {
     PaymentID id;                   // For relay payments, the funding tx.
     TimeMs time;
-    PaymentStatus statusSimplified; // we are working on making this correct for relay payments...
-    string statusDescription;       // "pending", "completed", "error"
+    PaymentStatus statusSimplified;
+    string statusDescription;       // "pending", "completed", "error", "claimable", "canceled"
```
mlsteele: It would be cleaner to just have PaymentStatus. Is there a reason to have duplicate fields?

go/stellar/transform.go
```diff
@@ -212,6 +212,16 @@ func transformPaymentRelay(mctx libkb.MetaContext, acctID stellar1.AccountID, p
 		loc.ToType = stellar1.ParticipantType_STELLAR
 		loc.ToUsername = ""
 		loc.ToAccountName = ""
+		if loc.StatusSimplified == stellar1.PaymentStatus_CANCELED {
```
mlsteele: would slightly prefer `p.Claim.TxStatus == stellar1.TransactionStatus_SUCCESS` so the ordering of assignments doesn't matter

go/stellar/transform.go
```diff
@@ -212,6 +212,16 @@ func transformPaymentRelay(mctx libkb.MetaContext, acctID stellar1.AccountID, p
...
+			// canceled payment. blank out toAssertion and stow in originalToAssertion
+			// set delta to what it would have been had the payment completed
+			loc.ToAssertion = ""
+			loc.OriginalToAssertion = p.ToAssertion
+			loc.Delta = stellar1.BalanceDelta_INCREASE
+			if acctID == p.FromStellar {
+				loc.Delta = stellar1.BalanceDelta_DECREASE
+			}
+		}
```
mlsteele: Why would you want delta to be what could have but definitely won't happen? If you canceled a relay back into the same account intuitively the delta is NONE as net no money has moved.

Btw I think it's possible to send a relay from primary A, change your primary to B, and cancel it. And it will have gone from A to B. Just a random thought.

buoyad: The views for `status == 'claimable'` and `status == 'canceled'` are very similar. We use the delta to figure out some signs and icons (+/-, sending/receiving), which should be the same in the two cases.

I'm not sure your example will work like that - [`CancelPaymentLocal`](https://github.com/keybase/client/blob/master/go/stellar/stellarsvc/frontend.go#L484) uses `relay.FromStellar` for `into`.

protocol/avdl/stellar1/local.avdl
buoyad: @mlsteele / @zanderz I'm fine with removing it, there's some more JS changes too so I'd rather do it separately

go/stellar/transform.go
mlsteele: So if a user deletes an account that they sent an unclaimed relay from then they lose the money even though technically there's a box for them with the secret key?

protocol/avdl/stellar1/local.avdl
mlsteele: Separately sounds great

go/stellar/transform.go
buoyad: Well, if they're claiming the payment from the UI then they haven't deleted the account. They'd have to use the CLI which falls back to the default account ID

mlsteele: I mean if someone sends and then deletes the sending account. There's no way for them to reclaim the money. Seems unfortunate. Sort of makes sense.

protocol/avdl/stellar1/local.avdl
```diff
@@ -93,9 +93,10 @@ protocol local {
     string    fromUsername;    // Optional. Filled if KEYBASE.
 
     union { null, AccountID } toAccountID; // Optional.
-    string    toAccountName; // Optional. Filled if toType = OWNACCOUNT
-    string    toUsername;    // Optional. Filled if KEYBASE.
-    string    toAssertion;   // Optional. Filled if SBS.
+    string    toAccountName;               // Optional. Filled if toType = OWNACCOUNT
+    string    toUsername;                  // Optional. Filled if KEYBASE.
+    string    toAssertion;                 // Optional. Filled if SBS.
```
mlsteele: "Filled if SBS" is no longer accurate since a canceled relay to an SBS will be blanked.

go/stellar/transform.go
mlsteele: yes

PR #14349, opened 2018-10-22
Enable file widget for all

jzila:

PR #14363, opened 2018-10-22
 fix widget snapshots

songgao:
This didn't come up before likely because it was feature flagged off and not actually included in the storyshots. Then when we turned it on today in https://github.com/keybase/client/pull/14349 , Jenkins didn't run because GH was still f'ed.


jzila:
I ran `yarn test -u` myself, and it didn't reveal anything. Weird.

And yes we chose to merge without Jenkins deliberately.

PR #14370, opened 2018-10-23
teams: invalidate KBFS Favorites cache more

jakob223:
Invalidate the KBFS favorites cache when:
 - user is added to a team
 - user is removed from a team
 - a team is renamed
 - a team is deleted
There are two other cases not included here:
 - when a user changes their favorites - that is already implemented
 - when SBS is resolved on an implicit team: see https://github.com/keybase/keybase/pull/3059

### DIFF
--- a/go/teams/handler.go
+++ b/go/teams/handler.go
@@ -3,6 +3,7 @@ package teams
 import (
 	"context"
 	"encoding/base64"
+	"encoding/hex"
 	"errors"
 	"fmt"
 
@@ -182,6 +183,17 @@ func sweepOpenTeamResetAndDeletedMembers(ctx context.Context, g *libkb.GlobalCon
 	return needRepoll, err
 }
 
+func refreshKBFSFavoritesCache(g *libkb.GlobalContext) error {
+	// TODO: determine if there is a better place to put this code
+	uid := g.GetMyUID().ToBytes()
+	kbUID, err := keybase1.UIDFromString(hex.EncodeToString(uid))
+	if err != nil {
+		return err
+	}
+	g.NotifyRouter.HandleFavoritesChanged(kbUID)
+	return nil
+}
+
 func handleChangeSingle(ctx context.Context, g *libkb.GlobalContext, row keybase1.TeamChangeRow, change keybase1.TeamChangeSet) (err error) {
 	change.KeyRotated = row.KeyRotated
 	change.MembershipChanged = row.MembershipChanged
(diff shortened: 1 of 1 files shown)


strib:
Why can't we just rely on the gregor messages from the server to trigger the notification in all these cases?  Seems like the server-based favorites list should be the source of truth, and so it should send a message everytime an action like this is committed, right?

go/teams/handler.go
```diff
@@ -182,6 +183,17 @@ func sweepOpenTeamResetAndDeletedMembers(ctx context.Context, g *libkb.GlobalCon
 	return needRepoll, err
 }
 
+func refreshKBFSFavoritesCache(g *libkb.GlobalContext) error {
+	// TODO: determine if there is a better place to put this code
+	uid := g.GetMyUID().ToBytes()
+	kbUID, err := keybase1.UIDFromString(hex.EncodeToString(uid))
```
maxtaco: Why doesn't `kbUID := g.GetMyUID()` just work?

jakob223:
> Why can't we just rely on the gregor messages from the server to trigger the notification in all these cases? Seems like the server-based favorites list should be the source of truth, and so it should send a message everytime an action like this is committed, right?

I chose to implement it this way in order to avoid adding gregor messages where we already have them. In many of these cases (as far as I can tell), the server doesn't explicitly know which users would need to be notified, it just uses a complicated SQL query to delete the appropriate rows from the favorites table.

This decision was partially motivated by my understanding the client code a lot better than the server code and being more confident about modifying the right place. I could be convinced that it's the wrong decision.

Currently the gregor message for KBFS favorites is called exactly when a favorite is created or destroyed; except in the case of SBS resolution* we could even just build and modify the cache locally, only requesting it on login/logout.

*we could do that too, but it would be a little more complicated.

strib:
> I chose to implement it this way in order to avoid adding gregor messages where we already have them

Ok, that's fine.  I'm a little worried about races (i.e., if the server sends out this other kind of message before it actually updates the favorites list, and the user fetches the list before the update happens too, we might have some confusing results).  But if you're pretty sure that's not a big issue, I'm ok with this way.

go/teams/handler.go
jakob223: Whoops, I didn't realize `GetMyUID()` returns a `keybase1.UID`.

jakob223:
>  But if you're pretty sure that's not a big issue, I'm ok with this way.

I looked through all the places where the server is both updating favorites and sending a gregor notification, and in all cases the favorites update happens first.

strib:
Sweet, thank you!  I'm going to add patrick as a reviewer here too, since I don't know anything about these parts of the client code.

jakob223:
I claim this can be merged whenever, as long as it happens before https://github.com/keybase/kbfs/pull/1870 - any reason I shouldn't merge it now?

PR #14371, opened 2018-10-23
pgp push/pull private commands

maxtaco:
- This patch introduces two new commands: `keybase pgp push-private` and `keybase pgp pull-private`. They export/import private keys from GnuGP's keychain and import/export to KBFS
- Storage is in keybase/private/you/.keys/pgp/*
- This allows simple and safe use of PGP keys across your devices, that's as strong as your device keys (rather than just your password, in the old model).
- There are no tests, since they woul dhave to be docker tests, and this is a 2-point timebox. If this feature becomes more successful, we can add tests.


go/client/cmd_pgp_pull_private.go
```diff
@@ -0,0 +1,92 @@
...
+				Usage: "force pull, and don't prompt for confirmation",
+			},
+		},
+		Action: func(c *cli.Context) {
+			cl.ChooseCommand(&CmdPGPPullPrivate{Contextified: libkb.NewContextified(g)}, "pull-private", c)
+		},
+		Description: `"keybase pgp pull-private" exports the given private keys from the GnuPG
+   keychain and pus them to KBFS, at /keybase/private/<you>/.keys/pgp. Here,
+   they are protected with your Keybase device keys, and cannot be broken
+   via brute-force passphrase guessing.`,
```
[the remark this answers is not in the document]
maxtaco: whoops, i didn't update that description. fixed, thanks

go/client/cmd_pgp_push_private.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return ret, nil
+}
+
+func (v *CmdPGPPushPrivate) Run() (err error) {
+
+	if !v.force {
+		dui := v.G().UI.GetDumbOutputUI()
+		dui.Printf(
+			ColorString(v.G(), "bold", "PLEASE READ THIS CAREFULLY -- PRIVATE KEYS ARE AT STAKE!") + "\n" +
```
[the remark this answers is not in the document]
maxtaco: hmmm weird, i guess it's busted. for my account it doesn't work, since i have two keys. one sec...

maxtaco:
Thanks for the review. Let's get it into master. It sometimes works, so the eager Linux users can play around with it.

zapu:
I agree, let's merge

go/engine/pgp_push_private.go
```diff
@@ -0,0 +1,218 @@
...
+	})
+	return err
+}
+
+func (e *PGPPushPrivate) remove(m libkb.MetaContext, fs *keybase1.SimpleFSClient, file string) (err error) {
+	opid, err := fs.SimpleFSMakeOpid(m.Ctx())
+	if err != nil {
+		return err
+	}
+	defer fs.SimpleFSClose(m.Ctx(), opid)
```
strib: Despite what the avdl documentation says, I suspect this is returning an error, since for a `Remove` call there is no handle that needs to be closed.  The `Wait` below should be enough to finish the operation.

go/engine/pgp_push_private.go
```diff
@@ -0,0 +1,218 @@
...
+		Cli: rpc.NewClient(xp, libkb.NewContextifiedErrorUnwrapper(m.G()), nil),
+	}, nil
+}
+
+func (e *PGPPushPrivate) mkdir(m libkb.MetaContext, fs *keybase1.SimpleFSClient, path string) (err error) {
+	opid, err := fs.SimpleFSMakeOpid(m.Ctx())
+	if err != nil {
+		return err
+	}
+	defer fs.SimpleFSClose(m.Ctx(), opid)
```
strib: Same here, making a directory doesn't leave a handle hanging around that needs to be cleaned up.  (Yes, this interface kind of sucks.)

maxtaco: Should I `Wait`?

go/engine/pgp_push_private.go
maxtaco: I think I did see an error in my logs, now that you mention it.

go/engine/pgp_push_private.go
maxtaco: BTw, for this, I'm doing what's in `cmd_simplefs_mkdir.go`

strib: You don't need to `Wait` either; it will error too.  My earlier statement wasn't totally accurate: you only need to wait when using an opID to calls that are asynchronous.  Making a directory is not one of these (though I realize that's totally not obvious.)

I guess it's a (harmless) bug that the `fs mkdir` does a close.

PR #14372, opened 2018-10-23
cleanup platform file

chrisnojima:
I needed this file to be cleaned up for another PR. Before each path was figured out all over the place, instead we just group it by platform 
also removed a bunch of things that are only used in  one place

### DIFF
--- a/shared/constants/platform.desktop.js
+++ b/shared/constants/platform.desktop.js
@@ -1,24 +1,20 @@
 // @flow
 import path from 'path'
 
-const isMobile = false
-const isAndroid = false
-const isIOS = false
-const isLargeScreen = true
-const isSimulator = false
-const isIPhoneX = false
-const isDeviceSecureAndroid: boolean = false
-const isAndroidNewerThanM: boolean = false
-const isAndroidNewerThanN: boolean = false
+const platform = process.platform
+export const isMobile = false
+export const isAndroid = false
+export const isIOS = false
+export const isLargeScreen = true
+export const isIPhoneX = false
+export const isElectron = true
+export const isDarwin = platform === 'darwin'
+export const isWindows = platform === 'win32'
+export const isLinux = platform === 'linux'
+export const isAndroidNewerThanN = false
 
-const isElectron = true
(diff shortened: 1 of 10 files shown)


shared/constants/platform.desktop.js
```diff
@@ -27,196 +23,79 @@ if (__DEV__ && !__STORYBOOK__) {
   console.log(`Run mode: ${runMode}`)
 }
 
-const envedPathLinux = {
```
chrisnojima: this file is mostly moved around, the output should be the same except for one case (see below)

shared/constants/platform.desktop.js
```diff
@@ -27,196 +23,79 @@ if (__DEV__ && !__STORYBOOK__) {
...
-
-function findDataRoot(): string {
-  switch (process.platform) {
-    case 'darwin':
-      return `${homeEnv}/Library/Application Support/${envedPathOSX[runMode]}/`
-    case 'linux':
-      const linuxDefaultRoot = `${homeEnv}/.local/share`
-      return `${process.env['XDG_DATA_HOME'] || linuxDefaultRoot}/${envedPathLinux[runMode]}/`
-    case 'win32':
-      return `${process.env['LOCALAPPDATA'] || ''}\\Keybase\\`
```
chrisnojima: for every other path on every platform we always use Keybase${suffix} except this, which seems like a bug, so in the new code it does this also

shared/constants/platform.native.js
```diff
@@ -2,85 +2,42 @@
 import {Dimensions, Platform, NativeModules} from 'react-native'
 import {cachesDirectoryPath} from '../util/file.native'
 
-// Modules from the native part of the code. Differently named on Android/iOS
-const nativeBridge = NativeModules.KeybaseEngine ||
-  NativeModules.ObjcEngine || {
```
chrisnojima: Long ago we made renamed ObjcEngine to KeybaseEngine

chrisnojima:
@keybase/react-hackers 
cc: @zanderz for windows changes

PR #14373, opened 2018-10-23
ephemeral notification bad network

joshblum:
splits the `EphemeralPurgeNotifInfo` into two notifications, previously on a bad network these notifications could be delayed.

cc @buoyad


shared/constants/chat2/meta.js
```diff
@@ -110,7 +110,12 @@ export const updateMeta = (
   meta: Types.ConversationMeta
 ): Types.ConversationMeta => {
   // Older/same version and same state?
-  if (meta.inboxVersion <= old.inboxVersion && meta.trustedState === old.trustedState) {
+  if (meta.inboxVersion === old.inboxVersion) {
+    return old.withMutations(m => {
```
chrisnojima: we prefer
```
old.merge({
  snippet: meta.snippet,
  snippetDecoration: meta.snippetDecoration
})
```

PR #14376, opened 2018-10-23
fix selection changing a bunch as you mouse over and cleanup bad recompose calls

chrisnojima:
@keybase/react-hackers this fixes the weird mouse behavior around user mentions and removes the usage of recompose in this component


chrisnojima:
@keybase/react-hackers bump

PR #14377, opened 2018-10-23
schema/config version

joshblum:
pending https://github.com/keybase/keybase/pull/3060


joshblum:
merging so ci doesn't break once the server PR is out

PR #14380, opened 2018-10-24
move/copy on desktop: components and story

songgao:
depends on https://github.com/keybase/client/pull/14327


songgao:
bump!

songgao:
need ... re.. view ...

shared/fs/destination-picker/index.desktop.js
```diff
@@ -0,0 +1,129 @@
+// @flow
+import * as React from 'react'
+import * as Types from '../../constants/types/fs'
+import * as Constants from '../../constants/fs'
+import * as Styles from '../../styles'
+import * as Kb from '../../common-adapters'
+import Rows from '../row/rows-container'
+import * as F from '../common'
```
jzila: `import * as FsCommon`

shared/fs/destination-picker/index.desktop.js
```diff
@@ -0,0 +1,129 @@
...
+import * as R from '../row/common'
```
jzila: `import * as RowCommon`

shared/fs/row/rows.js
```diff
@@ -10,59 +10,80 @@ import Tlf from './tlf-container'
...
 
 type Props = {
   items: Array<Types.RowItem>,
   routePath: I.List<string>,
+  inDestinationPicker?: boolean,
   ifEmpty?: ?React.Node,
 }
 
-export const WrapRow = ({children}: {children: React.Node}) => (
+export const WrapRow = ({children, noDivider}: {children: React.Node, noDivider?: boolean}) => (
```
jzila: Can you make the divider a positive case instead of a negative case? i.e. `showDivider`

I know this means you'll have to plumb it through the existing code, but at least it'll be easier to read overall.

songgao: Good point. Although I'm realizing we seem to always want the divider. So I'll probably just revert this part.

songgao:
Thanks @jzila ! I'll implement the feedback in https://github.com/keybase/client/pull/14430 as it does some renames (unsplit desktop/split) and it'd be annoying to deal with conflicts.

PR #14381, opened 2018-10-24
move/copy dialog on desktop: redux and saga

songgao:
depends on https://github.com/keybase/client/pull/14380


shared/actions/fs/index.js
```diff
@@ -685,6 +685,38 @@ const deleteFile = (state, action: FsGen.DeleteFilePayload) => {
...
+      Types.pathConcat(
+        state.fs.moveOrCopy.destinationParentPath,
+        Types.getPathName(state.fs.moveOrCopy.sourceItemPath)
+      )
+    ),
+  },
+})
+
+const moveOrCopy = (state, action: FsGen.MovePayload | FsGen.CopyPayload) => {
+  const opID = Constants.makeUUID()
```
chrisnojima: why not just make the const here and remove the function , aka
```
const moveOrCopy = (state, action: FsGen.MovePayload | FsGen.CopyPayload) => {
  const params = {opID:Constants.makeUUID() , src: { ... /* code form stateToMoveOrCopyParams*/
  ...
  return action.type === FsGen.move ? RPCTypes.SimpleFSSimpleFSMoveRpcPromise(params) : RpcTypes....
```
nothing else will use stateToMoveOrCopyParams right

shared/fs/destination-picker/container.js
```diff
@@ -0,0 +1,69 @@
...
+      .writable
+)
+
+const destinationAndSourceAreInSameTLF = memoize(stateProps => {
+  if (!stateProps._moveOrCopy.sourceItemPath) {
+    return false
+  }
+  const src = Types.getPathElements(stateProps._moveOrCopy.sourceItemPath)
+  const dst = Types.getPathElements(stateProps._moveOrCopy.destinationParentPath)
+  return src.length >= 3 && dst.length >= 3 && src[1] === dst[1] && src[2] === dst[2]
```
chrisnojima: surprised there isn't a Constants helper for this that'll give you the tlf and you can just compare it (even as a string)

shared/fs/destination-picker/index.js.flow
```diff
@@ -0,0 +1,15 @@
+// @flow
```
chrisnojima: does this really need a split mobile/desktop impl?

shared/actions/fs/index.js
songgao: Ah yeah; good catch. I was probalby thinking two approaches at the same time

shared/fs/destination-picker/index.js.flow
songgao: Maybe not; I was gonna do it like this for now, and try and see if I can get away without a split when I get to the mobile (will get to it right after these are merged).

songgao:
@chrisnojima ready for another look thanks!

songgao:
ping 🏓

shared/actions/fs/index.js
chrisnojima: ^ above i move opID in there too so you don't need 2 copies

songgao:
@chrisnojima PTAL thanks!

shared/reducers/fs.js
```diff
@@ -142,10 +142,13 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
     case FsGen.journalUpdate: {
       const {syncingPaths, totalSyncingBytes, endEstimate} = action.payload
-      return state
-        .setIn(['uploads', 'syncingPaths'], I.Set(syncingPaths))
-        .setIn(['uploads', 'totalSyncingBytes'], totalSyncingBytes)
-        .setIn(['uploads', 'endEstimate'], endEstimate || undefined)
+      return (
+        state
+          // $FlowFixMe
+          .setIn(['uploads', 'syncingPaths'], I.Set(syncingPaths))
```
chrisnojima: Is this needed with new immutable?

songgao: You mean the $FlowFixMe part? It came up when I added more `setIn`s. And since it didn't complain as it does when $FlowFixMe is unnecessary, after a rebase, I assume it means this is still needed. 

Or do you mean something else?

songgao: I'm gonna merge this for now, and will address more feedbacks in the next PR.

PR #14394, opened 2018-10-24
Fix to field clearing out when you come back from currency selection

cjb:
@keybase/react-hackers 

We were unintentionally using it as an uncontrolled input.


chrisnojima:
actually we should make this impossible

PR #14395, opened 2018-10-24
Fix KEYBASE_FEATURES env for ui

MarcoPolo:
@keybase/react-hackers If anything is in the featureFlagsOverride it ignores the KEYBASE_FEATURES env. Which I'd imagine is not intentional


zanderz:
isn't that what override means though?

chrisnojima:
honestly we shold just drop teh env vars and just use the json

PR #14399, opened 2018-10-24
switch to mainline immutable rc-10

chrisnojima:
- [x] type image stubs so we can require images as strings correctly
- [x] fix some wallet typing
- [x] fix some fs typing
- [x] change to OrderedSet and just sort.

We do lose partialSort and SortedSet but i think we should likely handle this differently. Instead we should just keep the newest 10 messages and keep that in the store. then the remote and inbox can just use that and not have to sort all the time


shared/constants/fs.js
```diff
@@ -602,7 +602,10 @@ export const isMedia = (pathItem: Types.PathItem): boolean =>
   pathItem.type === 'file' && ['image', 'av'].includes(viewTypeFromMimeType(pathItem.mimeType))
 
 const slashKeybaseSlashLength = '/keybase/'.length
-export const generateFileURL = (path: Types.Path, localHTTPServerInfo: ?Types._LocalHTTPServer): string => {
+export const generateFileURL = (
+  path: Types.Path,
+  localHTTPServerInfo: ?$ReadOnly<Types._LocalHTTPServer>
```
chrisnojima: we should avoid exposing these _ types actually but this is the smallest change that fixes this for now

chrisnojima:
cc: @jzila @songgao

MarcoPolo:
🔥 all the `// $FlowIssue`

chrisnojima:
yeah!

PR #14411, opened 2018-10-25
remove error from non-error actions

chrisnojima:
This'll reduce a bunch of noise from the actions log
@keybase/react-hackers


chrisnojima:
@keybase/react-hackers bump

PR #14418, opened 2018-10-25
dep mkdirp

chrisnojima:
@keybase/react-hackers removes mkdirp


chrisnojima:
in the past i think. if i set a random log name it does make the path correctly

zanderz:
:thumbsup:

PR #14427, opened 2018-10-25
load display currency when account is loaded

zanderz:
...so it doesn't switch in the payment window.
This also keeps the currency empty if there's an outstanding call to retrieve displayCurrency. Should there be a spinner too?


shared/actions/wallets.js
```diff
@@ -215,11 +221,17 @@ const loadAccounts = (
     | ConfigGen.LoggedInPayload
 ) =>
   !actionHasError(action) &&
-  RPCStellarTypes.localGetWalletAccountsLocalRpcPromise().then(res =>
-    WalletsGen.createAccountsReceived({
+  RPCStellarTypes.localGetWalletAccountsLocalRpcPromise().then(res => {
+    // load the display currency of each wallet, now that we have the IDs
+    action.type === WalletsGen.loadAccounts &&
```
buoyad: This doesn't do anything (`map`s but doesn't do anything with the result). You could make a saga for `accountsReceived` that spawns those `loadDisplayCurrency` actions.

shared/wallets/wallet/settings/container.js
```diff
@@ -36,7 +36,6 @@ const mapDispatchToProps = (dispatch, {routeProps, navigateUp, navigateAppend})
   },
   _refresh: () => {
     dispatch(WalletsGen.createLoadDisplayCurrencies())
-    dispatch(WalletsGen.createLoadDisplayCurrency({accountID: routeProps.get('accountID')}))
```
buoyad: I think we should keep this call so we know it's up to date when showing this screen.

shared/actions/wallets.js
zanderz: But it does call `createLoadDisplayCurrency` - is there something else it needs to do?


VM2363:1 console.trace

  | (anonymous) | @ | VM2363:1
-- | -- | -- | --
  | createLoadDisplayCurrency | @ | wallets-gen.js:458
  | action.type._wallets_gen__WEBPACK_IMPORTED_MODULE_5__.loadAccounts.res.map.account | @ | wallets.js:204
  | _constants_types_rpc_stellar_gen__WEBPACK_IMPORTED_MODULE_2__.localGetWalletAccountsLocalRpcPromise.then.res | @ | wallets.js:204
  | _tickCallback | @ | internal/process/next_tick.js:68
  | Promise.then (async) |   |  
  | loadAccounts | @ | wallets.js:202
  | runCallEffect | @ | proc.js:523
  | runEffect | @ | proc.js:445

shared/wallets/wallet/settings/container.js
zanderz: ok

shared/actions/wallets.js
zanderz: I double checked and I'm certain it worked, but I changed it like you suggest.

buoyad: I don't think it could have, `actionToPromise` will `put` one action returned from the promise. Before, this was like
```
res.map(a => createLoadDisplayCurrency(...))
return createAccountsReceived(...)
```
so the `accountsReceived` will be dispatched but the mapped `loadDisplayCurrencies` don't do anything.

shared/wallets/send-form/asset-input/container.js
```diff
@@ -4,12 +4,13 @@ import * as WalletsGen from '../../../actions/wallets-gen'
...
 import * as Constants from '../../../constants/wallets'
+import {anyWaiting} from '../../../constants/waiting'
 
 const mapStateToProps = state => {
   const accountID = state.wallets.selectedAccount
   const currency = state.wallets.building.currency
-
-  const displayUnit = Constants.getCurrencyAndSymbol(state, currency)
+  const currencyWaiting = anyWaiting(state, Constants.loadDisplayCurrencyWaitingKey)
+  const displayUnit = currencyWaiting ? '' : Constants.getCurrencyAndSymbol(state, currency)
```
buoyad: maybe this should blank when `getCurrencyAndSymbol` returns 'XLM'. to avoid the flicker in every time the send form opens

PR #14433, opened 2018-10-26
Check file is not dir before running

joshblum:
keeps the error out of the logs and just echoed back to the user:

```
$ kb encrypt -i canaryd -o canary2.enc tester
▶ ERROR A directory is not a valid file
```

affects `decrypt` and the `keybase pgp` equivalents


joshblum:
i initially had some work to do that i think i hit the main areas but didn't want to stifle non-sensitive error messages that would make the UX hard to manage. we could kill the error that is `CTrace`d so the user sees it but not the logs.. may make debugging difficult for logsend though that's why i went with this practical approach

Issue #14432, opened 2018-10-26
macOS Ugly Error Message


fargozhu:
OS version: macOS 10.14 (18A391)
Keybase version: 2.8.0-20181023171250+c7287208b0

![screenshot 2018-10-26 at 10 17 40](https://user-images.githubusercontent.com/904062/47557603-63647700-d911-11e8-80a6-c7ebbeb7328b.png)

maxtaco:
CC @keybase/react-hackers

chrisnojima:
investigating

chrisnojima:
PR up, thanks!

PR #14434, opened 2018-10-26
fix flow issue and incorrect object.keys wrapping

chrisnojima:
fixes https://github.com/keybase/client/issues/14432
@keybase/react-hackers

### DIFF
--- a/shared/common-adapters/avatar.js
+++ b/shared/common-adapters/avatar.js
@@ -109,8 +109,6 @@ class SharedAskForUserData {
   _userQueue = {}
   _userLastReq = {}
   _username = ''
-  _teamQueue: Object = {}
-  _userQueue: Object = {}
 
   // call this with the current username
   _checkLoggedIn = username => {
@@ -143,6 +141,8 @@ class SharedAskForUserData {
       }
       return false
     })
+    this._teamQueue = {}
+    this._userQueue = {}
     if (usernames.length) {
       this._dispatch(ConfigGen.createLoadAvatars({usernames}))
     }
--- a/shared/common-adapters/avatar.js
+++ b/shared/common-adapters/avatar.js
@@ -109,6 +109,8 @@ class SharedAskForUserData {
   _userQueue = {}
   _userLastReq = {}
   _username = ''
+  _teamQueue: Object = {}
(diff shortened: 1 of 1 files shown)


mmaxim:
Would this have caused other problem with loading avatars? We have two tickets on core about how they failing to load in some circumstances.

chrisnojima:
Mmm unlikely

shared/common-adapters/avatar.js
```diff
@@ -109,6 +109,8 @@ class SharedAskForUserData {
   _userQueue = {}
   _userLastReq = {}
   _username = ''
+  _teamQueue: Object = {}
+  _userQueue: Object = {}
```
mmaxim: Both of these already in there, this one just has a type on it. Delete the two above it?

shared/common-adapters/avatar.js
```diff
@@ -143,12 +143,10 @@ class SharedAskForUserData {
       }
       return false
     })
-    this._teamQueue = {}
```
mmaxim: I don't get why this is removed. How does this ever get cleared?

mmaxim:
One general concern about this entire avatar strategy here is what happens if the service fails to load an avatar. It seems like that failure is going to be stuck on the frontend for 30 minutes. There isn't any kind of feedback into this `SharedAskForUserData` of whether or not the requests succeeded, it just marks the time for when it made the request. 

Maybe we could reconsider this whole thing, and go back to just hitting the service every time (which does have a story for a failed avatar load). I'd be curious to know just what kind of speed this thing brings, given the complexity it brings to the table.

mmaxim:
The cache also only really seems to affect the case where I am just scrolling the inbox up and down (probably an unlikely user interaction). The first load is still hitting the service for everything (given its been 30 minutes), so I am still sending a call per avatar if I am interacting with the app in the common case. Since that is true, why not just request every time and eliminate a whole class of bugs from ever happening?

chrisnojima:
i'd prefer to not discuss architecture discussions in a bugfix pr

mmaxim:
K I'm going to put up a different PR with some change I think will help here.

PR #14435, opened 2018-10-26
Always show currency codes

patrickxb:
The result of buildpayment is admittedly confusing, but this PR should make it so frontend doesn't need to decipher as much.

`worthDescription` is still used for the "This is $0.20 USD" at the bottom of the build payment, request payment dialog.

But there are three new fields that should make it clearer:  displayAmountXLM, displayAmountFiat, sendingIntentionXLM.  They make the confirm dialog a lot easier to understand.

This also has a small fix for the chat message "worth" so that it has a currency code as well.

cc @malgorithms this fixes CORE-9325 which was marked high priority...requires releases to fix.


zanderz:
Shouldn't the command line also use these new fields? maybe a separate ticket?

patrickxb:
CLI has different RPCs and currently doesn't show ambiguous data, so i'm not too concerned about it.

    2018/10/25 15:53
    1000 CLP (6.1736364 XLM)
    zanderz -> patrick
    Note: It was chilean pesos, amirite?
    
    2018/10/26 11:42
    0.1 USD (0.4308003 XLM)
    chris -> patrick

PR #14437, opened 2018-10-26
make inbox unboxing interruptible, and more aggressively limit parallelism CORE-8968

mmaxim:
Patch does the following:

1.) Refactor "localizing" inbox items into a its own file.
2.) Make `InboxSource` have a single `localizePipeline` per instance.
3.) Make `localizePipeline` interruptible.
4.) Have `GetThreadNonblock` interrupt it so that when we fetch a thread there is no contention for resources.


go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+	}
+	return res
+}
+
+func (b *baseLocalizer) getConvs(inbox types.Inbox, maxLocalize *int) []chat1.Conversation {
+	convs := utils.PluckConvs(inbox.ConvsUnverified)
+	if maxLocalize == nil {
+		return convs
+	}
+	return convs[:*maxLocalize]
```
joshblum: should check the length here or could out of bounds panic

go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+	maxLocalize *int) (res []chat1.ConversationLocal, err error) {
+	defer b.Trace(ctx, func() error { return err }, "Localize")()
+	inbox = b.filterSelfFinalized(ctx, inbox)
+	convs := b.getConvs(inbox, maxLocalize)
+	b.baseLocalizer.pipeline.queue(ctx, uid, convs, b.localizeCb)
+	if err != nil {
+		return res, err
+	}
+	res = make([]chat1.ConversationLocal, len(convs))
+	indexMap := make(map[string]int)
```
joshblum: why does the ordering matter here?

go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+		localizeCb:    localizeCb,
+	}
+}
+
+func (b *blockingLocalizer) Localize(ctx context.Context, uid gregor1.UID, inbox types.Inbox,
+	maxLocalize *int) (res []chat1.ConversationLocal, err error) {
+	defer b.Trace(ctx, func() error { return err }, "Localize")()
+	inbox = b.filterSelfFinalized(ctx, inbox)
+	convs := b.getConvs(inbox, maxLocalize)
+	b.baseLocalizer.pipeline.queue(ctx, uid, convs, b.localizeCb)
```
joshblum: should this be returning an error on the queue being full? we check `err` below but it's from the return value

go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+	filteredInbox := b.filterInboxRes(ctx, inbox, uid)
+	// Send inbox over localize channel
+	b.localizeCb <- types.AsyncInboxResult{
+		InboxRes: &filteredInbox,
+	}
+	// Spawn off localization into its own goroutine and use cb to communicate with outside world
+	bctx := BackgroundContext(ctx, b.G())
+	go func() {
+		b.Debug(bctx, "Localize: starting background localization: convs: %d", len(inbox.ConvsUnverified))
+		b.baseLocalizer.pipeline.queue(bctx, uid, b.getConvs(inbox, maxLocalize), b.localizeCb)
```
joshblum: check and log error? (if this returns an error as noted above)

go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+	retCh chan types.AsyncInboxResult) {
+	defer s.Trace(ctx, func() error { return nil }, "queue")()
+	s.Lock()
+	defer s.Unlock()
+	job := newLocalizerPipelineJob(ctx, s.G(), uid, convs, retCh)
+	job.ctx, job.cancelFn = context.WithCancel(BackgroundContext(ctx, s.G()))
+	select {
+	case s.jobQueue <- job:
+	default:
+		s.Debug(ctx, "queue: dropped job, queue full")
```
joshblum: return this as an err?

go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+func (s *localizerPipeline) resume(ctx context.Context) bool {
+	defer s.Trace(ctx, func() error { return nil }, "resume")()
+	s.Lock()
+	defer s.Unlock()
+	s.suspendWg.Done()
+	return false
+}
+
+func (s *localizerPipeline) isContextCanceled(err error) bool {
+	if err == nil {
```
joshblum: can we drop this check and let it fall through to the end?

go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+					s.Debug(job.ctx, "localizeLoop: re-enqueuing canceled job")
+					s.jobQueue <- job.retry(s.G())
+				}
+				if job.closeIfDone() {
+					s.Debug(job.ctx, "localizeLoop: all job tasks complete")
+				}
+			}()
+			select {
+			case <-doneCh:
+				job.cancelFn()
```
joshblum: why do we call this here?

go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+
+	return foundMsgs, chat1.ConversationErrorType_NONE, nil
+}
+
+func (s *localizerPipeline) getMinWriterRoleInfoLocal(ctx context.Context, info *chat1.ConversationMinWriterRoleInfo) *chat1.ConversationMinWriterRoleInfoLocal {
+	if info == nil {
+		return nil
+	}
+	username := ""
+	name, err := s.G().GetUPAKLoader().LookupUsername(ctx, keybase1.UID(info.Uid.String()))
```
joshblum: realize this is just a mv but should we use uid mapper here instead for performance?

go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+	select {
+	case s.jobQueue <- job:
+	default:
+		s.Debug(ctx, "queue: dropped job, queue full")
+	}
+	s.Debug(ctx, "queue: len: %d", len(s.jobQueue))
+}
+
+func (s *localizerPipeline) clearQueue() {
+	s.jobQueue = make(chan *localizerPipelineJob, 500)
```
joshblum: can you make this configurable and drop it to 1-2 in tests to see if any bad behavior is triggered?

go/chat/localizer.go
```diff
@@ -0,0 +1,952 @@
...
+		conv  chat1.Conversation
+		index int
+	}
+
+	eg, ctx := errgroup.WithContext(ctx)
+	convCh := make(chan job)
+	pending := localizeJob.getPending()
+	retCh := make(chan chat1.ConversationID, len(pending))
+
+	if len(pending) == 0 {
```
joshblum: check before making `retCh`

go/chat/localizer.go
mmaxim: I'll just make `queue` block if the buffer is full, think it is ok.

mmaxim:
ok @joshblum I think that addresses your current comments.

PR #14438, opened 2018-10-26
hasAcceptedDisclaimer RPC fast path

mlsteele:
Narrowed `getWalletSettingsLocal` to `hasAcceptedDisclaimerLocal` which has one job, to return whether the user has accepted the disclaimer, caching when possible, but with no loss of serializability.

Calling `hasAcceptedDisclaimerLocal` when the user _has_ accepted the disclaimer will result in 1 server RPC ever per install. Calling `hasAcceptedDisclaimerLocal` when the user has not accepted the disclaimer will still result in a server RPC each time, since you'd want to pick up an acceptance from another device.

The RPC can still error out on network conditions. But in the case where the user has viewed the wallet tab on the device before that will not happen.


go/stellar/global.go
```diff
@@ -163,6 +173,81 @@ func (s *Stellar) UpdateUnreadCount(ctx context.Context, accountID stellar1.Acco
...
+func (s *Stellar) hasAcceptedDisclaimer(ctx context.Context) (bool, error) {
+	log := func(format string, args ...interface{}) {
+		s.G().Log.CDebugf(ctx, "Stellar.hasAcceptedDisclaimer "+format, args...)
+	}
+	uv, err := s.G().GetMeUV(ctx)
+	if err != nil {
+		return false, err
+	}
+	s.disclaimerLock.Lock()
+	s.disclaimerLock.Unlock()
```
patrickxb: you need to move this after 194

patrickxb:
Where does ` s.disclaimerAccepted` get set?

go/stellar/global.go
mlsteele: yeah oops meant this to be deferred

go/stellar/global.go
```diff
@@ -163,6 +173,83 @@ func (s *Stellar) UpdateUnreadCount(ctx context.Context, accountID stellar1.Acco
...
+}
+
+func (s *Stellar) informAcceptedDisclaimerLocked(ctx context.Context) (err error) {
+	defer s.G().CTraceTimed(ctx, "Stellar.informAcceptedDisclaimer", func() error { return err })()
+	uv, err := s.G().GetMeUV(ctx)
+	if err != nil {
+		return err
+	}
+	// Store memory
+	s.disclaimerAccepted = &uv
```
mlsteele: disclaimerAccepted gets set here

mlsteele:
hm broke gui during rebase

go/stellar/global.go
```diff
@@ -163,6 +173,83 @@ func (s *Stellar) UpdateUnreadCount(ctx context.Context, accountID stellar1.Acco
...
+
+func (s *Stellar) informAcceptedDisclaimer(ctx context.Context) {
+	s.disclaimerLock.Lock()
+	defer s.disclaimerLock.Unlock()
+	_ = s.informAcceptedDisclaimerLocked(ctx)
+}
+
+func (s *Stellar) informAcceptedDisclaimerLocked(ctx context.Context) (err error) {
+	defer s.G().CTraceTimed(ctx, "Stellar.informAcceptedDisclaimer", func() error { return err })()
+	uv, err := s.G().GetMeUV(ctx)
```
patrickxb: There's a very small chance uv could change in between your checks above and this line.  But it is there...

I'm fine with it as `keybase logout && keybase login someoneelse` would be very hard to squeeze into that time window, but you could pass uv in here...

PR #14439, opened 2018-10-26
Don't return an error if no output

patrickxb:


patrickxb:
I get this running service locally

{code}
▶ ERROR Error calling updater: strconv.ParseBool: parsing "": invalid syntax
▶ ERROR Error calling updater: strconv.ParseBool: parsing "": invalid syntax
{code}

patrickxb:
You could use `cmd.CombinedOutput()` to get stderr too....

songgao:
But the exit code is 1 if it errorred, which should've caused CombinedOutput() to error right?

heronhaye:
are we going to do this? or should we close

PR #14441, opened 2018-10-26
filter out info banners in the confirm form

zanderz:


chrisnojima:
we dont' want to do this kind of filtering

PR #14446, opened 2018-10-26
fix broken send dialog in save lastXLM

zanderz:
will figure it out later

### DIFF
--- a/shared/actions/json/wallets.json
+++ b/shared/actions/json/wallets.json
@@ -222,7 +222,8 @@
     },
     "sentPayment": {
       "_description": "Successfully sent a payment",
-      "kbTxID": "HiddenString"
+      "kbTxID": "HiddenString",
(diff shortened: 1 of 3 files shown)


cjb:
It's okay with me to add the LastSentXLM payload to the SentPayment action, and then just fire SentPayment.  That seems like a short change that keeps everything working.

zanderz:
Thanks @cjb , missed that comment. I see what you're saying but one possible issue now is if there is a send error. Still an improvement until we do something bigger, like Nojima suggests.

cjb:
@zanderz That seems okay -- if the feature is to remember that you last sent XLM, and there was a send error, then you didn't actually last send XLM. :)

zanderz:
Doesn't there have to be an action in order for the actual writing to take place, in actions/config/index.js? Or would that one just listen for sentPayment?

PR #14447, opened 2018-10-26
Markdown fixes

MarcoPolo:
Fixes these issues with the new markdown:

1. null conversation ID. Fixes the createChannel regex to return null instead of the regex /#()\b/
1. Weird boundaries with `_` and `*` i.e. ```under_score the first, `under_score the second` ``` was parsed wrong
1. Fix mobile link styling

Any issues I missed? @mmaxim @chrisnojima 

@keybase/react-hackers


chrisnojima:
Common URL's   Like google.com etc

chrisnojima:
Plus 1 as thumbs up. See chat history in frontend

chrisnojima:
Was the link styling desktop also?

chrisnojima:
I'm talking without the http prefix.  I can try again later but it didn't work on desktop when I tried

MarcoPolo:
ah I just repro'd. It works if you have just the url, but fails if there is text in front.
i.e.:


google.com works

this won't work google.com

I can't fix it right now, but the fix involves just having the text break at every space. The text rigth now is too greedy and eating the beginning of the link which doesn't let the link regex match.

MarcoPolo:
don't merge yet, the changes to bold are kinda broken

MarcoPolo:
alright, changes to bold are fixed.

mmaxim:
I think that is all I know about, seems good to merge modulo CI

mmaxim:
Merge away!

shared/markdown/regex.js
```diff
@@ -1,5 +1,29 @@
...
+)
 
-export {tldExp}
+const commonTlds = [
```
chrisnojima: why have 2 of these?

shared/markdown/generate-emoji-parser.js
```diff
@@ -2,6 +2,7 @@
 import fs from 'fs'
 import path from 'path'
 import emojiData from 'emoji-datasource'
+import {escapeRegExp} from 'lodash'
```
chrisnojima: lodash-es

chrisnojima: (i'm not exactly sure why this doens't fail lint)

buoyad:
Channel mentions aren't styled / clickable on this branch
image

shared/common-adapters/markdown/shared.js
```diff
@@ -77,7 +77,7 @@ function createMentionRegex(meta: ?MarkdownMeta): ?RegExp {
 }
 
 function createChannelRegex(meta: ?MarkdownMeta): ?RegExp {
-  if (!meta || !meta.mentionsChannelName) {
+  if (!meta || !meta.mentionsChannelName || meta.mentionsChannelName.isEmpty) {
```
buoyad: `.isEmpty` -> `.isEmpty()`

MarcoPolo: nice catch! I'll fix the clickable thing in another branch

MarcoPolo: O actually, this might fix your clickable issue

shared/markdown/generate-emoji-parser.js
MarcoPolo: lodash-es doesn't work here

chrisnojima: K

PR #14450, opened 2018-10-27
always update if trusted state changes

mmaxim:
cc @joshblum 

Always update to new meta payload if the trusted state changes. This fixes a problem where channel names don't load if the data doesn't come up in local metadata (recent regression noticed by @xgess).


joshblum:
thx!

PR #14455, opened 2018-10-29
Remove unused suppressions

buoyad:
@keybase/react-hackers not sure how these got here, cc @songgao? Will merge once CI passes.


songgao:
Sorry I *just* saw this. I guess flow got smarter after an upgrade where it started to be able to figure that out?

PR #14461, opened 2018-10-29
missing .mpack file: RemoteSettingsRepairman

zanderz:
The "repairman" skipped files it thought were in both locations, but failed to identify that case when deleting the old version.


zanderz:
Also: should we just remove this repairman? it is only for clients older than Feb 2017

maxtaco:
I think we should remove it

zanderz:
ok removed @maxtaco  (ci is broken in master right?)

maxtaco:
I don't know, you'd have to ask...

PR #14465, opened 2018-10-29
make sure pending messages don't get removed CORE-9281

mmaxim:
I think the following scenario is possible, I have no proof that it can actually happen:

1.) Be on a bad connection
2.) Get a thread stale notification from the service.
3.) Invoke `GetThreadNonblock` on the service for new messages.
4.) Compose a message and hit send
5.) Insert the message into the store
6.) Get the thread callback from `GetThreadNonblock` which does not contain the just sent message and be in clear the thread mode.
7.) Blow away the sent message in the thread view.
8.) Call `PostTextLocalNonblock`. There is nothing happening from here to put the pending message back until it gets sent by the service and calls back with the real version of it.

To combat this, try to insert the new pending message before and after the call to `PostTextLocalNonblock`. Also make sure we don't upgrade a sent message with a pending one.


shared/actions/chat2/index.js
```diff
@@ -1277,7 +1277,7 @@ const messageSend = (action: Chat2Gen.MessageSendPayload, state: TypedState) =>
 
   logger.info('[MessageSend]', 'non-empty text?', text.stringValue().length > 0)
 
-  return Saga.sequentially([addMessage, postText])
+  return Saga.sequentially([addMessage, postText, addMessage])
```
mmaxim: The reason we do this twice is to "undo" the removal of the first `addMessage` by a callback from `GetThreadNonblock` that was live while we dispatched these actions. The change to `upgradeMessage` complements this by making it safe to send in the message with the same `outboxID` as one that may be in the store as a sent message already.

chrisnojima: i this this is very confusing, so at least we have to add a comment why we'd do this but really this is cuase we don't have the bookkeeping to fix this race so we're injecting it in twice, which seems like a way to paper over the symptom

mmaxim: I agree it's not ideal, but given the options seems like the easiest way to handle the situation. Added a comment.

PR #14467, opened 2018-10-29
don't render markdown memo if it's an empty string

buoyad:
This broke with new markdown. r? @keybase/react-hackers


chrisnojima:
lets actually change markdown memo itself to render nothing and not have to wrap it all the time

buoyad:
@chrisnojima done, PTAL

PR #14475, opened 2018-10-30
Start the 2.10.0 cycle

cjb:
@chrisnojima @mmaxim @maxtaco 

Let's go arbitrary version numbers!

### DIFF
--- a/go/libkb/version.go
+++ b/go/libkb/version.go
@@ -4,4 +4,4 @@
 package libkb
 
 // Version is the current version (should be MAJOR.MINOR.PATCH)
-const Version = "3.0.0"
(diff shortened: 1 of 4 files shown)


mmaxim:
I was thinking it would be cool to roll up to 3.0.0 when the wallet is out for everyone. Maybe 2.10?

cjb:
@mmaxim I didn't want to think about whether we have any accidental lexical comparisons, e.g. 2.10.0 < 2.2.0.  What'd you think?

mmaxim:
Think it is unlikely, I'm not aware of anything that uses this number other than the updater which does the normal semver comparison.

PR #14478, opened 2018-10-30
fix border on really remove account dialog

buoyad:
@keybase/react-hackers should we add an `overflow: 'hidden'` on the dialog container? I think that'd knock out a bunch of these issues. Not sure if there are still any places where we need to render outside the dialog


chrisnojima:
seems fine to me

PR #14480, opened 2018-10-30
Nix search package

joshblum:
refactors `search` package into the `chat` package in prep for index syncing.

a bunch of stuff is chat package specific `NewBlockingSender`, `NewBoxer`, `NewConversation` etc, seems easier to do this than hit import nightmare. separated so the diff is easier when we do the syncing

### DIFF
--- a/go/chat/search_constants.go
+++ b/go/chat/search_constants.go
@@ -1,4 +1,4 @@
-package search
+package chat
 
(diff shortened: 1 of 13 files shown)


joshblum:
true but i think it will be easier to have all the logic in a single place

PR #14486, opened 2018-10-30
Factor download to attachemnts pkg

joshblum:


go/chat/attachments/downloader.go
```diff
@@ -0,0 +1,85 @@
...
+		sink = f
+	} else {
+		if sink, err = os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600); err != nil {
+			return "", nil, err
+		}
+	}
+	return filename, sink, nil
+}
+
+func DownloadAttachmentLocal(ctx context.Context, g *globals.Context, uid gregor1.UID,
```
mmaxim: Drop the `Local`

mmaxim: I would just call this `Download`

PR #14487, opened 2018-10-30
fixes for people

chrisnojima:
- [x] better typing
- [x] less render thrash

I noticed this while working on other stuff on the people tab

@keybase/react-hackers


shared/people/container.js
```diff
@@ -1,21 +1,48 @@
...
+}
+
+class LoadOnMount extends React.PureComponent<Props> {
   componentDidMount() {
     this.props.getData(false)
   }
 
+  _onSearch = () => this.props.onSearch()
+  _getData = (markViewed?: boolean) => this.props.getData(markViewed)
+  _onClickUser = (username: string) => this.props.onClickUser(username)
```
buoyad: why not pass these functions through directly?

buoyad: oh, so `People` won't need to rerender if the functions change 👍

shared/people/container.js
```diff
@@ -39,9 +66,9 @@ const mapDispatchToProps = (dispatch, {navigateAppend}) => ({
 
 const mergeProps = (stateProps, dispatchProps) => {
   return {
-    newItems: stateProps._newItems.toJS(),
-    oldItems: stateProps._oldItems.toJS(),
-    followSuggestions: stateProps.followSuggestions.toJS(),
+    newItems: stateProps._newItems,
+    oldItems: stateProps._oldItems,
```
buoyad: we should get rid of these underscored names

chrisnojima: done

PR #14494, opened 2018-10-30
update vendored go-winio

zanderz:
to include our second PR against microsoft/go-winio: https://github.com/Microsoft/go-winio/pull/101


zanderz:
Note this did seem to have fixed the user's crash after all

maxtaco:
CI seems like a CI bug.

PR #14496, opened 2018-10-30
Use black_75_on_white so emojis don't become transparent

MarcoPolo:
@keybase/react-hackers @cecileboucheron

### DIFF
--- a/shared/wallets/send-form/note-and-memo/index.js
+++ b/shared/wallets/send-form/note-and-memo/index.js
@@ -213,7 +213,7 @@ const styles = Styles.styleSheetCreate({
     flex: 1,
   },
   input: {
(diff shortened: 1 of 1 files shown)


adamjspooner:
Do you want to include the fix for `.text_Body` too?

MarcoPolo:
@adamjspooner are you talking about every case of `Body` type in text tags? I think that's a much bigger change and shouldn't be included here. This is just for the stellar send form

adamjspooner:
Yeah, the emoji in chat messages are transparent as well. I wasn't sure if you wanted to clean that up with this PR.

MarcoPolo:
Yeah that's a different issue. Fixed here: https://github.com/keybase/client/pull/14495

PR #14498, opened 2018-10-31
routing fixes for moveOrCopy and gray out bad destinations

songgao:
There are some comments in `fs/routes.js`, but basically:

* split the root of fsTab (which was originally just folder view) into a bare route node without component, but with two children `folder` and `destinationPicker`. It has `defaultSelected: 'folder'` so when we go there it still shows the folder view;
* for `destinationPicker`, i.e. the moveOrCopy dialog, only store an index in `routeProps`, but bookkeep in `state.fs.moveOrCopy`, which has a `destinationParentPath: I.List<Path>` for current showing path in the moveOrCopy dialog, and a `sourceItemPath: Path` for the moving/copying target.
  * For desktop, since we only have one layer of `destinationPicker`, so `state.fs.moveOrCopy.destinationParentPath` always has at most one element inside, and `routeProps.get('index')` is always `0`. When user navigates in moveOrCopy dialog, we just dispatch actions that sets the element at index 0 in the `I.List` to different paths.
  * For mobile, since we need to make swipe actions work, different layers of `destinationPicker`s need to be mounted, with `routeProps.get('index')` being 0, 1, ... . `state.fs.moveOrCopy.destinationParentPath` keeps a `Path` for each mounted layer of `destinationPicker`.

With this setup, on mobile user can:

* swipe back (i.e. from left edge of screen to right) to go to parent folder;
* being deep inside folders, but still able to tap "Cancel" to go back to whatever folder they were viewing when they initiated "Move or Copy" from a menu.

depends on: https://github.com/keybase/client/pull/14430

### DIFF
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -5,6 +5,7 @@ import * as FsGen from '../fs-gen'
 import * as I from 'immutable'
 import * as RPCTypes from '../../constants/types/rpc-gen'
 import * as Saga from '../../util/saga'
+import * as SettingsConstants from '../../constants/settings'
 import * as Tabs from '../../constants/tabs'
 import engine from '../../engine'
 import * as NotificationsGen from '../notifications-gen'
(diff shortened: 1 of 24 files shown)


songgao:
⪮

shared/actions/fs/index.js
```diff
@@ -724,6 +725,48 @@ const moveOrCopy = (state, action: FsGen.MovePayload | FsGen.CopyPayload) => {
   )
 }
 
+const moveOrCopyOpen = isMobile
```
chrisnojima: maybe easier to read if its just sep functions
```
const moveOrCopyOpenDesktop...
const moveOrCopyOpenNative...
...
yield Saga.actionToAction(FsGen.moveOrCopyOpen, isMobile ? moveOrCopyOpenNative : moveOrCopyOpenDesktop)
```

songgao:
@chrisnojima ready for another look; thanks!

PR #14499, opened 2018-10-31
Show a blue background around payments we've just been told about, until you tab away

cjb:
@keybase/react-hackers 

* When we're told about a new payment via the notifications hooks (such as `stellar.1.notify.paymentStatusNotification`), add it to a per-account `newPayments` map in the store.  
* When we're rendering the transaction list, use a blue background color for payments that are in this map -- this replaces the previous logic of "payment is pending or unread".
* When we leave the wallets tab for another tab, and we were on the transaction list for an account, remove any payments in the `newPayments` map for that account so they stop being blue.

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -569,23 +569,11 @@ const maybeClearErrors = (state: TypedState) => {
 }
 
 const maybeClearNewTxs = (action: RouteTreeGen.SwitchToPayload, state: TypedState) => {
-  const list = I.List(action.payload.path)
-  const root = list.first()
-  const settingsPath = getPath(state.routeTree.routeState, [Tabs.settingsTab])
-  const walletsPath = getPath(state.routeTree.routeState, [Tabs.walletsTab])
+  const rootTab = I.List(action.payload.path).first()
   // If we're leaving from the Wallets tab, and the Wallets tab route
   // was the main transaction list for an account, clear new txs.
   // FIXME: The hardcoded routes here are fragile if routes change.
-  if (
-    (!isMobile &&
-      root !== Tabs.walletsTab &&
-      state.routeTree.previousTab === Tabs.walletsTab &&
-      walletsPath.get(1) === 'wallet') ||
-    (isMobile &&
-      root !== Tabs.settingsTab &&
-      state.routeTree.previousTab === Tabs.settingsTab &&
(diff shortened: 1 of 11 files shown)


shared/actions/wallets.js
```diff
@@ -561,6 +568,31 @@ const maybeClearErrors = (state: TypedState) => {
...
 
+const maybeClearNewTxs = (action: RouteTreeGen.SwitchToPayload, state: TypedState) => {
+  const list = I.List(action.payload.path)
+  const root = list.first()
+  const settingsPath = getPath(state.routeTree.routeState, [Tabs.settingsTab])
+  const walletsPath = getPath(state.routeTree.routeState, [Tabs.walletsTab])
+  // If we're leaving from the Wallets tab, and the Wallets tab route
+  // was the main transaction list for an account, clear new txs.
+  // FIXME: The hardcoded routes here are fragile if routes change.
+  if (
```
chrisnojima: we'll make all this a lot simpler in the future , but for now i think it might be clearer to encode it more like (untested)

```
// outside this function, maybe in Constants?
const rootWalletTab = isMobile ? Tabs.settingsTab : Tabs.walletsTab
const isLookingAtWallet = routeState => getPath(routeState, rootWalletTab).get(isMobile ? 2 : 1) === 'wallet'

// then this turns into
if (rootTab !== Constants.rootWalletTab && Constants.isLookingAtWallet(lastRoute)) {

}
```

also, with the existing logic does the walletsPath.get(1) === 'wallet' work? isn't this the current path and not the previous one?

cjb: @chrisnojima Constants change looks good.

Yeah, it works because this is on switchTab, so the path itself isn't changing, just which tab is selected.  We don't record the previous path anywhere -- the reducer runs before the saga and blows it away -- so looking at the previous path isn't easy right now.

If we'd like to expire the newPayments when you nav inside the wallets tab (e.g. go to transactions details and back, or click on a different account in the wallets list), as well as when you change tabs, I think the best way to get there might be to just add the expiry to those two click handlers, rather than hook into every navigateAppend/navigateTo and try to tell what the previous path was somehow.  What do you think?

cjb:
@chrisnojima Pushed another change, to the wallet-row container -- now we manually clear payments when moving between accounts in the account list, too.  That seems a good UX overall.

PR #14503, opened 2018-10-31
Vendor go-crypto

zapu:
Tested on android simulator, fixes the user with problematic key


zapu:
finally passed, @maxtaco quick R on revendor?

PR #14505, opened 2018-10-31
remove filter stories

chrisnojima:
@keybase/react-hackers node 10/11 differ in how they return stuff from this external module's function so its not worth having it be in the stories

### DIFF
--- a/shared/chat/conversation/messages/react-button/emoji-picker/index.stories.js
+++ b/shared/chat/conversation/messages/react-button/emoji-picker/index.stories.js
@@ -1,31 +1,12 @@
 // @flow
 import * as React from 'react'
-import {Box, Box2, NewInput} from '../../../../../common-adapters'
+import {Box} from '../../../../../common-adapters'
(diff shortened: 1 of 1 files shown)


shared/chat/conversation/messages/react-button/emoji-picker/index.stories.js
```diff
@@ -1,31 +1,12 @@
...
-  }
-}
-
-const load = () => {
-  const story = storiesOf('Chat/Emoji picker', module)
+const load = () =>
+  storiesOf('Chat/Emoji picker', module)
     .addDecorator(story => <Box style={{height: 400, overflow: 'hidden', width: 300}}>{story()}</Box>)
     .add('Default', () => <ChooseEmoji onChoose={action('onChoose')} width={300} />)
-    .add('Custom filter', () => <WithFilter />)
```
buoyad: was this failing too? I find this one useful sometimes

chrisnojima: k put back

PR #14506, opened 2018-10-31
unfurling scraper part 1 CORE-9243

mmaxim:
Patch does the following:

1.) New package called `unfurl` which will be the place where all the unfurl specific code will land.
2.) Add a new basic scraper that just looks for OpenGraph tags for metadata.
3.) Add a tester program to demo the scraper.
4.) Add some easy test cases.
5.) Vendor in the following libraries:
- https://github.com/gocolly/colly (and deps)
- golang.org/x/net/publicsuffix


go/chat/unfurl/scraper_test.go
```diff
@@ -0,0 +1,124 @@
...
+			require.True(t, e.Description == nil || *e.Description == *r.Description)
+			require.True(t, (e.ImageUrl == nil && r.ImageUrl == nil) || (e.ImageUrl != nil && r.ImageUrl != nil))
+			require.True(t, e.ImageUrl == nil || *e.ImageUrl == *r.ImageUrl)
+			require.True(t, (e.FaviconUrl == nil && r.FaviconUrl == nil) || (e.FaviconUrl != nil && r.FaviconUrl != nil))
+			require.True(t, e.FaviconUrl == nil || *e.FaviconUrl == *r.FaviconUrl)
+		default:
+			require.Fail(t, "unknown unfurl typ")
+		}
+	}
+	testCase("wsj0", chat1.NewUnfurlRawWithGeneric(chat1.UnfurlGenericRaw{
```
joshblum: can you add `.html` to all these files? just helps with default rendering on github/editors

go/chat/unfurl/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	return parsed.Hostname(), nil
+}
+
+func GetDomain(uri string) (res string, err error) {
+	parsed, err := url.Parse(uri)
+	if err != nil {
+		return res, err
+	}
+	hostname := parsed.Hostname()
```
joshblum: use `GetHostname` here?

go/chat/unfurl/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	if len(hostname) == 0 {
+		return res, errors.New("no hostname")
+	}
+	return publicsuffix.EffectiveTLDPlusOne(hostname)
```
joshblum: if`publicsuffix.EffectiveTLDPlusOne("foo.bar.golang.org") -> "golang.org"` why not split that on `.` and just return `golang`? would make a `switch` in `ClassifyDomain` possible and we can drop the `IsDomain` function

PR #14511, opened 2018-10-31
fill search indexer in background

joshblum:
Adds a background loop the the `Indexer` to queue up a small number of jobs on the `ConvLoader` to slowly build the index in the background.


go/service/main.go
```diff
@@ -378,15 +378,16 @@ func (d *Service) purgeOldChatAttachmentData() {
...
+	kuid := d.G().Env.GetUID()
+	if !kuid.IsNil() {
+		uid := kuid.ToBytes()
 		g := globals.NewContext(d.G(), d.ChatG())
 		g.MessageDeliverer.Start(context.Background(), uid)
 		g.ConvLoader.Start(context.Background(), uid)
 		g.FetchRetrier.Start(context.Background(), uid)
 		g.EphemeralPurger.Start(context.Background(), uid)
 		g.InboxSource.Start(context.Background(), uid)
+		g.Indexer.Start(chat.IdentifyModeCtx(context.Background(), keybase1.TLFIdentifyBehavior_CHAT_SKIP, nil), uid)
```
joshblum: was getting hit with `ERROR: invalid context with no chat metadata` in tests but couldn't use within the `search` package because of the circular dependency. is there some other workaround that is cleaner?

PR #14512, opened 2018-10-31
Cap allowed decimal places in send form amount input

buoyad:
Caps to 7 for XLM and 2 for anything else. There was a bug where if you've entered 7 decimals & then changed to a display currency, typing chars into the input was impossible because the value was always invalid. I fixed this by clearing out the amount when the display currency changes. Another option would be to truncate the input when this happens but I figured we might want to do this anyway (cc @malgorithms). r? @keybase/react-hackers

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -328,6 +328,12 @@ const loadDisplayCurrency = (state: TypedState, action: WalletsGen.LoadDisplayCu
   )
 }
 
+const displayCurrencyReceived = (state: TypedState, action: WalletsGen.DisplayCurrencyReceivedPayload) => {
+  if (action.payload.setBuildingCurrency) {
+    return Saga.put(WalletsGen.createSetBuildingCurrency({currency: action.payload.currency.code}))
+  }
+}
+
 const changeDisplayCurrency = (state: TypedState, action: WalletsGen.ChangeDisplayCurrencyPayload) =>
   RPCStellarTypes.localChangeDisplayCurrencyLocalRpcPromise(
     {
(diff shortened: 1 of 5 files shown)


malgorithms:
I'd lean towards just truncating the input if someone changes the currency, rather than clearing the input. 95 out of 100 times when someone types an input in and then switches the currency, they won't have put too many digits in. in the other 5 cases, we'll just be stripping out the meaningless data.

buoyad:
@malgorithms Sounds good to me

shared/reducers/wallets.js
```diff
@@ -59,11 +59,6 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
         if (action.payload.accountID) {
           stateMutable.update('currencyMap', c => c.set(action.payload.accountID, action.payload.currency))
         }
-        if (action.payload.setBuildingCurrency) {
```
buoyad: Left these action changes in because I think it's best to only have this set in one place

shared/actions/wallets.js
```diff
@@ -328,6 +328,12 @@ const loadDisplayCurrency = (state: TypedState, action: WalletsGen.LoadDisplayCu
   )
 }
 
+const displayCurrencyReceived = (state: TypedState, action: WalletsGen.DisplayCurrencyReceivedPayload) => {
+  if (action.payload.setBuildingCurrency) {
```
chrisnojima: Can just be an &&

shared/wallets/send-form/asset-input/index.js
```diff
@@ -4,10 +4,32 @@ import * as Kb from '../../../common-adapters'
 import * as Styles from '../../../styles'
 import Available from '../available/container'
 
+const isValidAmount = (amt, numDecimalsAllowed) => {
+  if (!isNaN(+amt) || amt === '.') {
+    // This is a valid number. Now check the number of decimal places
+    const split = amt.includes('.') && amt.split('.')
+    if (!split) {
```
chrisnojima: You can always split. If none it should be size 1

shared/wallets/send-form/asset-input/index.js
```diff
@@ -4,10 +4,32 @@ import * as Kb from '../../../common-adapters'
 import * as Styles from '../../../styles'
 import Available from '../available/container'
 
+const isValidAmount = (amt, numDecimalsAllowed) => {
+  if (!isNaN(+amt) || amt === '.') {
```
chrisnojima: Prefer number cast vs using +

shared/wallets/send-form/asset-input/index.js
```diff
@@ -4,10 +4,32 @@ import * as Kb from '../../../common-adapters'
...
+    const decimal = split[split.length - 1]
+    if (decimal.length <= numDecimalsAllowed) {
+      return true
+    }
+  }
+  return false
+}
+
+const truncateAmount = (amt, numDecimalsAllowed) => {
+  const num = +amt
```
chrisnojima: Same here

buoyad:
@chrisnojima addressed comments PTAL

shared/reducers/wallets.js
zanderz: Thing is, lastSentXLM isn't applied anywhere else, so it's broken now

buoyad: whoops! Making new pr...

PR #14520, opened 2018-10-31
Don't show the ellipsis action menu on chat message for pending/relay send txs

cjb:
@keybase/react-hackers


shared/constants/chat2/message.js
```diff
@@ -1067,6 +1067,28 @@ export const enoughTimeBetweenMessages = (
...
+  switch (message.type) {
+    case 'text':
+    case 'attachment':
+    case 'requestPayment':
+      return true
+    case 'sendPayment': {
+      // Is the payment pending?
+      if (
+        message &&
+        message.paymentInfo &&
```
buoyad: The `paymentInfo` won't always be on the message. This should use `Constants.getPaymentMessageInfo` to get the info.

cjb:
@buoyad Thanks, how's this look?

cjb:
@buoyad Oh right, thanks!  How's this?

PR #14522, opened 2018-11-01
fix bad trackbreak caching in identify2

maxtaco:
- the issue was that if alice tracked bob, and bob revoked a proof, that failed track didn't make it into and out of the cache through some code paths
- repro test case, and bug fix


maxtaco:
cc: @mmaxim

PR #14526, opened 2018-11-01
Make secret key warning more ominous.

adamjspooner:


shared/wallets/export-secret-key/index.js
```diff
@@ -51,16 +51,14 @@ export default class ExportSecretKeyPopup extends React.Component<Props> {
           style={iconCastPlatformStyles(styles.icon)}
         />
         {!Styles.isMobile && header}
+        <Box2 direction="horizontal" style={styles.warningContainer}>
+          <Text backgroundMode="Information" type="BodySmallSemibold" style={styles.warningText}>Only paste your secret key in 100% safe places. A hacker with this key could steal your Stellar&nbsp;account.</Text>
```
buoyad: :sheep: `A hacker` -> `Anyone` might be scarier, it's so easy anyone can do it.

adamjspooner: Good point! I'll change and merger. 👍

PR #14529, opened 2018-11-01
remove hash from signtool command

zanderz:
verified this results in signing with correct updated cert on both sf and ny machines


zanderz:
@maxtaco winbotny is good after this

PR #14533, opened 2018-11-01
Add InlineBanner common adapter and use to show warning with public memo

buoyad:
cc @adamjspooner should the copy be changed to your comment?
> Only visit URLs you know to be safe. Hackers might mislead you into giving them your private Stellar key (and all your money).

r? @keybase/react-hackers


shared/common-adapters/inline-banner.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+import * as React from 'react'
+import Box2 from './box'
+import Text, {type Background} from './text'
+import * as Styles from '../styles'
+
+// Show a banner inline in a layout
+//
+// This adapter is a WIP as of 11/1/2018, add props and functionality as new
```
chrisnojima: i don't think this comment adds much value, this is true of everything

buoyad: I can remove it. I thought we typically wouldn't want to add things to common adapters except as a last resort, so this comment could overrule that until it's baked a while longer.

chrisnojima: i think we likely don't want this as a common then

buoyad:
@chrisnojima I removed the adapter

PR #14538, opened 2018-11-01
tokenize v3

joshblum:


go/chat/search/utils.go
```diff
@@ -45,20 +45,19 @@ func tokenize(msgText string) []string {
...
-		stripped := stripExpr.ReplaceAllString(token, "")
-		tokenSet.Add(stripped)
-		stemmed := porterstemmer.StemWithoutLowerCasing([]rune(stripped))
-		tokenSet.Add(string(stemmed))
+		stripped := stripExpr.Split(token, -1)
+		for _, s := range stripped {
+			tokenSet.Add(s)
+			stemmed := porterstemmer.StemWithoutLowerCasing([]rune(s))
+			tokenSet.Add(string(stemmed))
+		}
```
heronhaye: did you also want to add the original token replaced with nothing? (I can't think of why but maybe there was a reason for the original implementation?)

joshblum: this change hits that case as well since we always add the token before the stemming/further splitting (https://github.com/keybase/client/pull/14538/files#diff-5094981fcefe1fa4ea02538383595577R49)

 so: `"@joshblum"` -> [`"@joshblum"`, `"joshblum"`] and `"joshblum@twitter"` -> `["joshblum@twitter", "joshblum", "twitter"]`

PR #14540, opened 2018-11-01
unfurling: store favicon and image as Asset on Unfurl and upload to S3 CORE-9302

mmaxim:
Patch does the following:

1.) Introduces `Packager` whose responsibility is to gather assets needed for displaying an unfurl and turn then into `chat1.Asset`.
2.) Run `PreprocessAsset` on the image and favicon from a URL to make sure they are a sane size, and to get meta information.
3.) Uploads the assets using `attachments.Store`.


go/chat/unfurl/packager.go
```diff
@@ -0,0 +1,133 @@
...
+		maxAssetSize: 2000000,
+	}
+}
+
+func (p *Packager) assetFilename(url string) string {
+	toks := strings.Split(url, "/")
+	if len(toks) > 0 {
+		return toks[len(toks)-1]
+	}
+	return "unknown.jpg"
```
joshblum:
```suggestion
	return "preview.jpg"
```

joshblum: (unless this is not user facing then idc)

go/chat/unfurl/packager.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (p *Packager) assetFromURL(ctx context.Context, url string, uid gregor1.UID,
+	convID chat1.ConversationID) (res chat1.Asset, err error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return res, err
+	}
+	defer resp.Body.Close()
+	dat, err := ioutil.ReadAll(resp.Body)
```
joshblum: can we use `resp.ContentLength` to abort before reading this into memory?

go/chat/unfurl/packager.go
mmaxim: No one should ever see this, its only here in case we can't MIME detect from the content of the URL.

PR #14546, opened 2018-11-01
Lower lvldb file limit to 48

joshblum:
Drops the max number of open files for levelDB so it's 96 across the chat/normal dbs. for OSX the default is 256 files so this gives us a little more headroom. if we do run out of files it's pretty fatal to the application running since errors start getting thrown everywhere. I tried picking through the log where this happened without finding any root culprit, wasn't able to reproduce locally either. saw a max of ~ 120 files with `lsof -p $kbpid`. 

cc @mmaxim


mmaxim:
LGTM, awaiting @maxtaco

PR #14547, opened 2018-11-01
Add a timestamp. Remove a timestamp.

adamjspooner:
@keybase/react-hackers, this is ready for review. It does two things:

1. It adds a timestamp (`4:20 PM` next to usernames in chat).
1. It removes the timestamp when the orange line is present.


shared/chat/conversation/messages/wrapper/wrapper-timestamp/index.js
```diff
@@ -71,7 +71,7 @@ class _WrapperTimestamp extends React.Component<Props & OverlayParentProps, Stat
     return (
       <Box style={styles.container}>
         {props.orangeLineAbove && <Box style={styles.orangeLine} />}
-        {!!props.timestamp && <Timestamp timestamp={props.timestamp} />}
+        {!!props.timestamp && !props.orangeLineAbove && <Timestamp timestamp={props.timestamp} />}
```
chrisnojima: better to just not set timeststamp if there's an orange line here: https://github.com/keybase/client/blob/master/shared/chat/conversation/messages/wrapper/wrapper-timestamp/container.js#L50

adamjspooner:

mmaxim:
I think the orange line is also still flakey, I click into threads periodically and just don't get one, but I guess that can be a separate PR.

shared/chat/conversation/messages/wrapper/wrapper-author/index.js
```diff
@@ -158,7 +158,7 @@ const RightSide = props => {
               isBroken={props.isBroken}
               onClick={props.onAuthorClick}
             />
-            <Text type="BodyTiny">{formatTimeForConversationList(props.timestamp)}</Text>
+            <Text type="BodyTiny">{moment(props.timestamp).format('h:mm A')}</Text>
```
chrisnojima: if we need another format we should just add it to util/timestamp (so it can be mocked also). we really don't want the existing one?

adamjspooner: Will do. I could exit early from the existing one or add a small helper. The existing one's logic changes the timestamp if it's older than a day.

PR #14549, opened 2018-11-01
Stellar lookup identify rejigger

mlsteele:
The expected behavior after this patch is:
For the CLI, a full identify is done and if it fails your send fails.
For the GUI, a trusted resolution is done, but no proofs other than those in your assertion have any bearing on whether you can send. Tracking is also not used.

The part about running an actual identify for gui sends is coming later. This patch is just recovery after max's fix exposed bugs here.

### DIFF
--- a/go/stellar/stellar.go
+++ b/go/stellar/stellar.go
@@ -343,17 +343,11 @@ func LookupRecipient(m libkb.MetaContext, to stellarcommon.RecipientInput, isCLI
 		return res, err
 	}
 
-	idRes, err := identifyRecipient(m, string(to), isCLI)
+	maybeUsername, err := lookupRecipientAssertion(m, string(to), isCLI)
 	if err != nil {
 		return res, err
 	}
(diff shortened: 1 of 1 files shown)


go/stellar/stellar.go
```diff
@@ -897,48 +889,65 @@ func localizePayment(ctx context.Context, g *libkb.GlobalContext, p stellar1.Pay
...
+			IdentifyBehavior: keybase1.TLFIdentifyBehavior_CLI,
+		}
 	}
 
 	eng := engine.NewResolveThenIdentify2(m.G(), &arg)
-	err := engine.RunEngine2(m, eng)
+	err = engine.RunEngine2(m, eng)
 	if err != nil {
-		// Ignore these errors
+		// These errors means no resolution was found.
```
patrickxb: means -> mean

patrickxb: (no need for full CI redo just for this)

PR #14554, opened 2018-11-01
gregorHandler: handle in-band KBFS favorites messages

jakob223:
Changed favorites messages to be in-band on the server (see https://github.com/keybase/keybase/pull/3086 ) - this makes the corresponding changes in the service.

We now keep track of which messages we have seen, since we'll see them for a whole week until they expire.

As part of this change, we also invalidate the favorites cache in the client automatically after one week so that these messages can expire. A PR on the KBFS repo for that is forthcoming.

### DIFF
--- a/go/service/kbfs_favorites_handler.go
+++ b/go/service/kbfs_favorites_handler.go
@@ -1,0 +1,85 @@
+// Handlers for ephemeral-related gregor messages
+
+package service
+
+import (
+	"encoding/hex"
+	"fmt"
+	"github.com/keybase/client/go/protocol/keybase1"
+	"strings"
+
+	"golang.org/x/net/context"
+
+	"github.com/keybase/client/go/gregor"
+	"github.com/keybase/client/go/libkb"
+	"github.com/keybase/client/go/protocol/gregor1"
+)
+
+const kbfsFavoritesHandlerName = "kbfsFavoritesHandler"
+
+type kbfsFavoritesHandler struct {
+	libkb.Contextified
+
+	// handled stores the set of requests that this client has processed so
+	// that we don't handle a request more than once.
+	handled map[string]bool
+}
+
+var _ libkb.GregorInBandMessageHandler = (*ekHandler)(nil)
+
+func newKBFSFavoritesHandler(g *libkb.GlobalContext) *kbfsFavoritesHandler {
+	return &kbfsFavoritesHandler{
+		Contextified: libkb.NewContextified(g),
+		handled:      make(map[string]bool),
+	}
+}
+
+func (r *kbfsFavoritesHandler) Create(ctx context.Context, cli gregor1.IncomingInterface,
+	category string, item gregor.Item) (bool, error) {
+	switch category {
+	case "kbfs.favorites":
+		return true, r.favoritesChanged(ctx, cli, item)
+	default:
+		if strings.HasPrefix(category, "kbfs.") {
+			return false, fmt.Errorf("unknown KBFS category: %q", category)
+		}
+		return false, nil
+	}
+}
+
+func (r *kbfsFavoritesHandler) Dismiss(ctx context.Context, cli gregor1.IncomingInterface, category string, item gregor.Item) (bool, error) {
+	return false, nil
+}
+
+func (r *kbfsFavoritesHandler) IsAlive() bool {
+	return true
+}
+
+func (r *kbfsFavoritesHandler) Name() string {
+	return kbfsFavoritesHandlerName
+}
+
+func (r *kbfsFavoritesHandler) favoritesChanged(ctx context.Context,
+	cli gregor1.IncomingInterface,
+	item gregor.Item) error {
+	r.G().Log.CDebugf(ctx, "kbfsFavoritesHandler: kbfs."+
+		"favorites received")
+
+	// check whether we have seen this message ID before.
+	// We can't dismiss the message even if we've handled it because we want
+	// all clients to see it.
+	msgID := item.Metadata().MsgID().String()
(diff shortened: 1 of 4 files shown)


go/service/kbfs_favorites_handler.go
```diff
@@ -0,0 +1,85 @@
+// Handlers for ephemeral-related gregor messages
```
strib: Probably want the copyright notice at the top of this file.

go/service/kbfs_favorites_handler.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+const kbfsFavoritesHandlerName = "kbfsFavoritesHandler"
+
+type kbfsFavoritesHandler struct {
+	libkb.Contextified
+
+	// handled stores the set of requests that this client has processed so
+	// that we don't handle a request more than once.
+	handled map[string]bool
```
strib: I don't know anything about gregor handlers, so maybe this is a standard practice, but it seems like this map can grow without bound.  Memory leak?

Also, it's not protected by a lock, but maybe that's fine if these handlers are only ever invoked by a single goroutine.

patrickxb: You definitely shouldn't do this.

Instead, locally dismiss the message.  I'll put a comment where that goes.

go/service/kbfs_favorites_handler.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil
+	}
+	r.handled[msgID] = true
+
+	kbUID, err := keybase1.UIDFromString(hex.EncodeToString(
+		item.Metadata().UID().Bytes()))
+	if err != nil {
+		return err
+	}
+	r.Contextified.G().NotifyRouter.HandleFavoritesChanged(kbUID)
```
patrickxb: Add something like this here:

        // We will locally dismiss for now so that each client only plays them once:                
        if err := r.G().GregorDismisser.LocalDismissItem(ctx, item.Metadata().MsgID()); err != nil {                                                                                             
                r.G().Log.CDebugf(ctx, "failed to locally dismiss something jakob edit me: %s", err)    
        }   

If I'm understanding what you are trying to do.

patrickxb: Actually, what are you trying to do?  Maybe you should fully dismiss the item here.

go/service/kbfs_favorites_handler.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// check whether we have seen this message ID before.
+	// We can't dismiss the message even if we've handled it because we want
+	// all clients to see it.
+	msgID := item.Metadata().MsgID().String()
+	if _, ok := r.handled[msgID]; ok {
+		// we have already handled this message
+		return nil
+	}
+	r.handled[msgID] = true
```
patrickxb: remove all this

go/service/kbfs_favorites_handler.go
```diff
@@ -0,0 +1,85 @@
...
+
+	kbUID, err := keybase1.UIDFromString(hex.EncodeToString(
+		item.Metadata().UID().Bytes()))
+	if err != nil {
+		return err
+	}
```
patrickxb: kbUID := keybase1.UID(item.Metadata().UID().String())

go/service/kbfs_favorites_handler.go
jakob223: I didn't realize you could locally dismiss a message. If that's possible, it's what I should be doing.

patrickxb:
Why in-band messages?

go/service/kbfs_favorites_handler.go
maxtaco: If you dismiss on one device, they get dismissed for all, so just beware!

patrickxb: Not with a local dismiss.

jakob223:
> Why in-band messages?

So that a user whose device is off can turn it back on and not have to re-fetch favorites unless they've received an invalidation.

jakob223:
@patrickxb I believe I've addressed all of your comments. Can you look again?

go/service/kbfs_favorites_handler.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/keybase/client/go/protocol/gregor1"
+)
+
+const kbfsFavoritesHandlerName = "kbfsFavoritesHandler"
+
+type kbfsFavoritesHandler struct {
+	libkb.Contextified
+}
+
+var _ libkb.GregorInBandMessageHandler = (*ekHandler)(nil)
```
patrickxb: did you mean to use *kbfsFavoritesHandler here?  instead of ekHandler

patrickxb:
> > Why in-band messages?
> 
> So that a user whose device is off can turn it back on and not have to re-fetch favorites unless they've received an invalidation.

Won't this mean the gregor state will get really large?

go/service/kbfs_favorites_handler.go
jakob223: yes I did.. copy-paste mishap.

jakob223:
> > > Why in-band messages?
> > 
> > 
> > So that a user whose device is off can turn it back on and not have to re-fetch favorites unless they've received an invalidation.
> 
> Won't this mean the gregor state will get really large?

I don't have a conception as to how big gregor state normally is. I don't think users should be getting favorites invalidations very often though - how many do you imagine would bloat the gregor state?

jakob223:
CI is failing but tests are passing for me locally.. I told Jenkins to run again.

patrickxb:
you might want to try rebasing.

PR #14555, opened 2018-11-02
[KBFS-3241] Sharing to KBFS in iOS

jzila:
Still some cleanup to go, but the basic bones are there and they work.

### DIFF
--- a/shared/react-native/ios/KeybaseShare/FilesViewController.m
+++ b/shared/react-native/ios/KeybaseShare/FilesViewController.m
@@ -10,7 +10,6 @@
 #import "keybase/keybase.h"
 
 @interface FilesViewController ()
-@property UISearchController* searchController;
 @property NSArray* unfilteredInboxItems; // the entire inbox
 @property NSArray* filteredInboxItems; // inbox items that are filtered by the search bar
 @end
@@ -21,12 +20,7 @@
   [super viewDidLoad];
   
   self.preferredContentSize = CGSizeMake(self.view.frame.size.width, 2*self.view.frame.size.height); // expand
-  self.searchController = [[UISearchController alloc] initWithSearchResultsController:nil];
-  self.searchController.searchResultsUpdater = self;
-  self.searchController.hidesNavigationBarDuringPresentation = false;
-  self.searchController.dimsBackgroundDuringPresentation = false;
   self.definesPresentationContext = YES;
-  [self.tableView setTableHeaderView:self.searchController.searchBar];
   
   // show this spinner on top of the table view until we have parsed the inbox
   UIActivityIndicatorView* av = [[UIActivityIndicatorView alloc] initWithActivityIndicatorStyle:UIActivityIndicatorViewStyleGray];
@@ -117,22 +111,6 @@
 - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath {
   NSDictionary* folder = [self getItemAtIndex:indexPath];
   [self.delegate folderSelected:folder]; // let main view controller know we have something
-}
-
-- (void)updateSearchResultsForSearchController:(UISearchController *)searchController {
-  NSString* term = [searchController.searchBar.text lowercaseString];
-  if ([term length] == 0) {
-    // reset on blank search bar
-    [self setFilteredInboxItems:self.unfilteredInboxItems];
-  } else {
-    NSPredicate* pred = [NSPredicate predicateWithBlock:^BOOL(id obj, NSDictionary* bindings) {
-      NSDictionary* item = obj;
-      return [item[@"Name"] containsString:term];
-    }];
-    NSArray* filtered = [self.unfilteredInboxItems filteredArrayUsingPredicate:pred];
-    [self setFilteredInboxItems:filtered];
-  }
-  [self.tableView reloadData];
 }
 
 @end
--- a/shared/react-native/ios/KeybaseShare/FilesViewController.m
+++ b/shared/react-native/ios/KeybaseShare/FilesViewController.m
@@ -10,9 +10,11 @@
 #import "keybase/keybase.h"
 
 @interface FilesViewController ()
-@property NSArray* path; // the path we are currently showing
+@property NSString* path; // the path we are currently showing
 @property NSArray* directoryEntries; // the directory entries at the current path
 @end
+
+NSString* const UpOneLevel = @".. [Up one level]";
 
 @implementation FilesViewController
 
@@ -21,9 +23,11 @@
   
   self.preferredContentSize = CGSizeMake(self.view.frame.size.width, 2*self.view.frame.size.height); // expand
   self.definesPresentationContext = YES;
+  [self setPath:@"/"];
   
   // show this spinner on top of the table view until we have parsed the inbox
   UIActivityIndicatorView* av = [[UIActivityIndicatorView alloc] initWithActivityIndicatorStyle:UIActivityIndicatorViewStyleGray];
+  [av setTag:self.view.tag+1];
   [self.view addSubview:av];
   [av setTranslatesAutoresizingMaskIntoConstraints:NO];
   [av setHidesWhenStopped:YES];
@@ -45,11 +49,15 @@
      ]
    ];
   
+  [self dispatchFilesBrowser];
+}
+
+- (void)dispatchFilesBrowser {
+  UIActivityIndicatorView* av = [self.view viewWithTag:self.view.tag + 1];
   dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
     NSError* error = NULL;
-    [self setPath:[NSArray new]];
     [self setDirectoryEntries:[NSArray new]];
-    NSString* jsonFiles = KeybaseExtensionListPath(@"/", &error); // returns the path list in JSON format
+    NSString* jsonFiles = KeybaseExtensionListPath(self.path, &error); // returns the path list in JSON format
     if (jsonFiles == nil) {
       dispatch_async(dispatch_get_main_queue(), ^{
         NSLog(@"failed to get files: %@", error);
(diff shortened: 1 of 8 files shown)


jzila:
@mmaxim mind giving it a try on a real device with some photos? I haven't tried bigger things yet.

jzila:
I tested with videos too and they work well.

mmaxim:
Ok will try today

jzila:
Let's get this across the finish line so we can test on admin builds. I unfortunately don't have a real device to test with.

shared/react-native/ios/KeybaseShare/FilesViewController.m
```diff
@@ -0,0 +1,222 @@
...
+                                                                attribute:NSLayoutAttributeCenterY
+                                                               multiplier:1 constant:0]
+                                   ]
+   ];
+  
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    NSError* error = NULL;
+    [self setDirectoryEntries:[NSArray new]];
+    NSString* jsonFiles = KeybaseExtensionListPath(self.path, &error); // returns the path list in JSON format
+    if (jsonFiles == nil) {
```
songgao: Should this cover `"null"` too since the Go side of `ExtensionListPath` never intentionally returns `nil`.

songgao: Also do we wanna check `&error` somewhere?

shared/react-native/ios/KeybaseShare/ShareViewController.m
```diff
@@ -329,6 +336,74 @@ - (void)processItem:(NSItemProvider*)item lastItem:(BOOL)lastItem {
...
+  };
+  
+  // If we get an image in the form of a UIImage, we first write it to a temp file
+  // and run it through the normal file sharing code path.
+  NSItemProviderCompletionHandler imageHandler = ^(UIImage* image, NSError* error) {
+    NSString* guid = [[NSProcessInfo processInfo] globallyUniqueString];
+    NSString* filename = [NSString stringWithFormat:@"%@%@.jpg", NSTemporaryDirectory(), guid];
+    NSURL* url = [NSURL fileURLWithPath:filename];
+    
+    NSData* imageData = UIImageJPEGRepresentation(image, 0.7);
```
songgao: I know chat does the same, but am curious if 0.7 is the JPEG quality and if we should have picked a higher value.

shared/react-native/ios/KeybaseShare/ShareViewController.m
```diff
@@ -329,6 +336,74 @@ - (void)processItem:(NSItemProvider*)item lastItem:(BOOL)lastItem {
...
+  } else if ([item hasItemConformingToTypeIdentifier:@"public.image"]) {
+    [item loadItemForTypeIdentifier:@"public.image" options:nil completionHandler:fileHandler];
+  } else if ([item hasItemConformingToTypeIdentifier:@"public.file-url"]) {
+    [item loadItemForTypeIdentifier:@"public.file-url" options:nil completionHandler:fileHandler];
+  } else if ([item hasItemConformingToTypeIdentifier:@"public.text"]) {
+    [item loadItemForTypeIdentifier:@"public.text" options:nil completionHandler:textHandler];
+  } else if ([item hasItemConformingToTypeIdentifier:@"public.url"]) {
+    [item loadItemForTypeIdentifier:@"public.url" options:nil completionHandler:urlHandler];
+  } else {
+    [[[PushNotifier alloc] init] localNotification:@"extension" msg:@"We failed to share your content. Please try from the Keybase app."
```
songgao: Does this happen for types that we can't handle? It seems it'd be better just say we can't share it, as the "upload" feature in app supports less types I think.

shared/react-native/ios/KeybaseShare/FilesViewController.m
```diff
@@ -0,0 +1,222 @@
...
+                              UpOneLevel, @"name",
+                              [NSNumber numberWithInt:1], @"direntType",
+                              nil]];
+    [itemsWithBack addObjectsFromArray:sortedAndFilteredItems];
+    [self setDirectoryEntries:itemsWithBack];
+  }
+}
+
+- (void)didReceiveMemoryWarning {
+  KeybaseExtensionForceGC();
```
songgao: I wonder if it'd make sense to show a popup explaining that we've been warned about memory consumption and are trying to deal with it. This way if we crash eventually, user knows it's because of limited RAM rather than some random bug.

shared/react-native/ios/KeybaseShare/FilesViewController.m
```diff
@@ -0,0 +1,222 @@
...
+  NSArray* pathElems = [self.path componentsSeparatedByString:@"/"];
+  return [pathElems count];
+}
+
+- (void)setPath:(NSString*)path {
+  _path = path;
+  NSArray* pathElems = [self.path componentsSeparatedByString:@"/"];
+  unsigned long pathLength = [pathElems count];
+  if (pathLength <= 2) {
+    [self setTitle:@"Keybase"];
```
songgao: Do we need `self.navigationItem.rightBarButtonItem = nil;` here too?

shared/react-native/ios/KeybaseShare/FilesViewController.m
```diff
@@ -0,0 +1,222 @@
...
+- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath {
+  UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"ConvCell"];
+  if (NULL == cell) {
+    cell = [[UITableViewCell alloc] initWithStyle:UITableViewCellStyleDefault reuseIdentifier:@"ConvCell"];
+  }
+  NSDictionary* item = [self getItemAtIndex:indexPath];
+  [[cell textLabel] setText:item[@"name"]];
+  if ([[item objectForKey:@"direntType"] intValue] == 1) {
+    // Style folders differently from files. Use Keybase blue.
+    [[cell textLabel] setTextColor:[UIColor colorWithRed:76.0/255.0 green:142.0/255.0 blue:1 alpha:1]];
```
songgao: Not suggesting a change as it's ultimately design's call, but I think in the main app it's just font-weight difference. Color is the same.

shared/react-native/ios/KeybaseShare/FilesViewController.m
```diff
@@ -0,0 +1,222 @@
...
+    dispatch_async(dispatch_get_main_queue(), ^{
+      [av stopAnimating];
+      [self.tableView reloadData];
+      [av removeFromSuperview];
+    });
+  });
+}
+
+NSInteger sortEntries(NSDictionary* one, NSDictionary* two, void* context) {
+  int t1 = [[one objectForKey:@"direntType"] intValue];
```
chrisnojima: i believe you can use subscripting aka
```
[one[@"direntType"] intValue]
```

shared/react-native/ios/KeybaseShare/FilesViewController.m
```diff
@@ -0,0 +1,222 @@
...
+                                                               multiplier:1 constant:0]
+                                   ]
+   ];
+  
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    NSError* error = NULL;
+    [self setDirectoryEntries:[NSArray new]];
+    NSString* jsonFiles = KeybaseExtensionListPath(self.path, &error); // returns the path list in JSON format
+    if (jsonFiles == nil) {
+      dispatch_async(dispatch_get_main_queue(), ^{
```
chrisnojima: you maybe just remove this and skip parseFiles if its null and just do the following dispatch_async (the extra reloadData will do nothing right)

shared/react-native/ios/KeybaseShare/FilesViewController.m
chrisnojima: (and apply elsewhere)

shared/react-native/ios/KeybaseShare/FilesViewController.m
```diff
@@ -0,0 +1,222 @@
...
+  NSError* error = nil;
+  NSData* data = [jsonFiles dataUsingEncoding:NSUTF8StringEncoding];
+  NSArray* items = [NSJSONSerialization JSONObjectWithData:data options: NSJSONReadingMutableContainers error: &error];
+  NSArray* sortedAndFilteredItems;
+  if (!items) {
+    if (error) {
+      NSLog(@"parseFiles: error parsing JSON: %@", error);
+    }
+    // At least show an empty folder.
+    sortedAndFilteredItems = [[NSArray alloc] init];
```
chrisnojima: you can do
`@[]` i think

shared/react-native/ios/KeybaseShare/FilesViewController.m
chrisnojima: i don't think we should do this. this can be caused by anything (other apps etc)

shared/react-native/ios/KeybaseShare/FilesViewController.m
```diff
@@ -0,0 +1,222 @@
...
+}
+
+- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath {
+  UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"ConvCell"];
+  if (NULL == cell) {
+    cell = [[UITableViewCell alloc] initWithStyle:UITableViewCellStyleDefault reuseIdentifier:@"ConvCell"];
+  }
+  NSDictionary* item = [self getItemAtIndex:indexPath];
+  [[cell textLabel] setText:item[@"name"]];
+  if ([[item objectForKey:@"direntType"] intValue] == 1) {
```
chrisnojima: `1` and the colors should be constants to improve readability

shared/react-native/ios/KeybaseShare/FilesViewController.m
jzila: I don't think we need to handle "null", but I will check `&error`.

shared/react-native/ios/KeybaseShare/ShareViewController.m
jzila: I'm just going to leave this in the interest of minimizing file size. The quality seems more than good enough.

shared/react-native/ios/KeybaseShare/ShareViewController.m
jzila: This happens for incoming file sources that we don't understand, irrespective of file type.

jzila: I'll change the string.

shared/react-native/ios/KeybaseShare/ShareViewController.m
```diff
@@ -377,33 +452,65 @@ - (void)didSelectPost {
...
+    }
   }
 }
 
 - (NSArray *)configurationItems {
-  SLComposeSheetConfigurationItem *item = [[SLComposeSheetConfigurationItem alloc] init];
-  item.title = @"Share to...";
-  item.valuePending = !self.hasInited; // show a spinner if we haven't inited
+  // Share to chat
+  SLComposeSheetConfigurationItem *chat = [[SLComposeSheetConfigurationItem alloc] init];
```
mmaxim: Would you mind if I played with this a bit? I think this interface can play out a little nicer, I can commit to this branch if that is ok.

shared/react-native/ios/KeybaseShare/FilesViewController.m
jzila: Yep, good call.

shared/react-native/ios/KeybaseShare/ShareViewController.m
mmaxim: I can also do as a followup, whatever you prefer.

shared/react-native/ios/KeybaseShare/FilesViewController.m
jzila: We do this on the chat side of the extension to try to minimize the probability that we'll be OOM killed. I was just following that pattern. @mmaxim WDYT?

shared/react-native/ios/KeybaseShare/ShareViewController.m
jzila: Let's do it as a followup.

shared/react-native/ios/KeybaseShare/FilesViewController.m
mmaxim: Running this is fine here.

jzila:
Ready for another look.

chriswep:
is this PR stuck somewhere @jzila ?

jzila:
Yes, it doesn't work due to memory restrictions on sharing extensions in iOS.

mmaxim:
That LevelDB change that just went in might make this more possible, the amount of memory saved is significant.

jzila:
Ok I'll test again.

chriswep:
@jzila did it work?

PR #14556, opened 2018-11-02
unfurling: add settings module and dev mode conversation storage system CORE-9242

mmaxim:
Patch does the following:

1.) Introduces a new interface `ConversationBackedStorage` with two simple accessor methods with the purpose if being implemented by a chat conversation.
2.) Add `DevConversationBackedStorage` which writes `TEXT` messages into `DEV` conversations, and always uses the contents of the latest message in the return type. Data is JSON encoded into the chat text messages.
3.) Add `unfurl.Settings` which manages the unfurling settings via a `ConversationBackedStorage`.
4.) Add simple methods and new types for tracking settings.


go/chat/devstorage.go
```diff
@@ -0,0 +1,102 @@
...
+			}),
+		}, 0, nil); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *DevConversationBackedStorage) Get(ctx context.Context, name string, dest interface{}) (found bool, err error) {
+	defer s.Trace(ctx, func() error { return err }, "Get(%s)", name)()
+	if s.G().ActiveDevice == nil {
```
joshblum: can you move `assertLoggedInUID` from the server to `utils` so we can use it throughout the chat package? that checks `ActiveDevice.HaveKeys()` and we should keep it consistent (i don't think there is a practical difference but we do this check everywhere so it's confusing if it's different)

go/chat/devstorage.go
```diff
@@ -0,0 +1,102 @@
...
+		return false, libkb.LoggedInError{}
+	}
+	uid := gregor1.UID(s.G().ActiveDevice.UID().ToBytes())
+	username := s.G().ActiveDevice.Username(libkb.NewMetaContext(ctx, s.G().ExternalG())).String()
+
+	convs, err := FindConversations(ctx, s.G(), s.DebugLabeler, true, s.ri, uid, username,
```
joshblum: why not use `NewConversation` here as well (functions as get or create right?) and then both methods can just have a shared `getConversation` method

go/chat/unfurl/settings.go
```diff
@@ -0,0 +1,116 @@
...
+
+	var mr modeRecord
+	found, err := s.storage.Get(ctx, settingsModeName, &mr)
+	if err != nil {
+		return res, err
+	}
+	if !found {
+		s.Debug(ctx, "Get: no mode setting found, using whitelisted")
+		mr.Mode = chat1.UnfurlMode_WHITELISTED
+	}
```
joshblum: abort whitelist fetch if mode isn't whitelisted?

go/chat/unfurl/settings.go
```diff
@@ -0,0 +1,116 @@
...
+	defer s.Trace(ctx, func() error { return err }, "WhitelistAdd(%s)", domain)()
+	var wr whitelistRecord
+	found, err := s.storage.Get(ctx, settingsWhitelistName, &wr)
+	if err != nil {
+		return err
+	}
+	if !found {
+		wr.Whitelist = nil
+	}
+	for _, w := range wr.Whitelist {
```
joshblum: why not make this a map?

go/chat/devstorage.go
mmaxim: Didn't want it to create it if it wasn't already there.

go/chat/unfurl/settings.go
mmaxim: I wanted to always get the whitelist for rendering the settings screen, where I think the whitelist will still show if you are in always or never mode.

go/chat/unfurl/settings.go
mmaxim: Yeah probably better, I'll change it.

go/chat/devstorage.go
mmaxim: Sure

go/chat/unfurl/settings.go
mmaxim: Actually I like it as a list better. The most common operation will be `Get`, and if I change to a map I need to sort it every time (or have some param like should sort). I know the code is a little uglier, but its just contained in this one file.

mmaxim:
@joshblum ok made `assertLoggedInUID` change, let me know if you accept my other replies.

go/chat/unfurl/settings.go
joshblum: ah makes sense

go/chat/unfurl/settings.go
joshblum: i feel it should only be sorted before it's rendered which is probably the least frequent op.  should there be some `IsValid(domain) bool` method to wrap calling `Get` and handing the mode/whitelist?

mmaxim: Changed my mind again, I think map is the way.

joshblum: heheok

mmaxim:
Going to merge since I think I hit all the feedback, let me know if there is anythign else.

PR #14558, opened 2018-11-02
Fix FormatCurrencyWithCodeSuffix for CHF

patrickxb:
swiss francs has a postfix symbol of "CHF" as well as a currency code of "CHF".

Before this fix, this would get rendered as

    1,234 CHF CHF

This fixes it so it is

    1,234 CHF

I believe this is the only currency like this, but if any others are added in the future, this should take care of that as well.

### DIFF
--- a/go/stellar/stellar.go
+++ b/go/stellar/stellar.go
@@ -970,6 +970,20 @@ func FormatCurrencyWithCodeSuffix(ctx context.Context, g *libkb.GlobalContext, a
 	if err != nil {
 		return "", err
 	}
+
+	// some currencies have the same symbol as code (CHF)
+	conf, err := g.GetStellar().GetServerDefinitions(ctx)
+	if err != nil {
+		return "", err
+	}
+	currency, ok := conf.Currencies[code]
+	if !ok {
+		return "", fmt.Errorf("FormatCurrency error: cannot find curency code %q", code)
+	}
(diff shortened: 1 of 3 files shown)


go/stellar/stellar.go
```diff
@@ -970,6 +970,20 @@ func FormatCurrencyWithCodeSuffix(ctx context.Context, g *libkb.GlobalContext, a
...
+	if err != nil {
+		return "", err
+	}
+	currency, ok := conf.Currencies[code]
+	if !ok {
+		return "", fmt.Errorf("FormatCurrency error: cannot find curency code %q", code)
+	}
+	if currency.Symbol.Postfix && currency.Symbol.Symbol == code.String() {
+		return pre, nil
+	}
```
zapu: hm at this point i'll probably modify the function above `func FormatCurrency` to take another parameter `addCodeSuffix bool` and do that logic. what do you think?

patrickxb: I'm not a huge fan of bool parameters as when you see it in the code you have no idea what it is.

    FormatCurrency(ctx, g, amount, code, true)

no clue what true means.

zapu: sure, it's not ideal either. if you are OK with `FormatCurrencyWithCodeSuffix` having to do this amount of boiler plate (number of LOC to get currency definition is unfortunate), then I'm fine as well

PR #14559, opened 2018-11-02
Display ephemeral unboxing error message

joshblum:
displays ephemeral missing keys error instead of `This exploding message is not available to you` (more relevant for pairwise mac failures)

screen shot 2018-11-02 at 11 12 21 am

cc @buoyad


go/chat/errors.go
```diff
@@ -39,7 +39,7 @@ func NewPermanentUnboxingError(inner error) UnboxingError {
 type PermanentUnboxingError struct{ inner error }
 
 func (e PermanentUnboxingError) Error() string {
-	return fmt.Sprintf("error unboxing chat message: %s", e.inner.Error())
```
joshblum: don't think this should make it's way up to the user

mmaxim: Hmm, why? Maybe just change to "decrypting" instead of "unboxing"? How about "error decrypting message: %s"?

joshblum: yeah i can go with that

PR #14560, opened 2018-11-02
Add renderTopmostOnly leaf tag and use on wallet dialogs

buoyad:
Desktop only: if `renderTopmostOnly` and `layerOnTop` are true, then the component won't render unless it's at the top of the route stack.

- Implements `renderTopmostOnly`
- Removes desktop / mobile split in send form root
- Cleans up plumbing callbacks for switching screens
- Cleans checks on `ownProps` in the previously routable or directly shown components. Now all the dialogs are routed

r? @keybase/react-hackers


shared/stories/prop-providers.js
```diff
@@ -49,10 +49,11 @@ export const Usernames = (following: string[] = defaultFollowing, you: string =
   },
 })
 
+// Pass "waitingKey: true" to this to render the waiting state
 export const WaitingButton = () => ({
   WaitingButton: (ownProps: _WaitingButton.OwnProps): _WaitingButton.Props => ({
     ...ownProps,
-    storeWaiting: false,
+    storeWaiting: ownProps.waitingKey === 'true',
```
buoyad: Now we can pass in `waitingKey="true"` to stories to test the `WaitingButton` spinner

shared/wallets/wallet/settings/popups/really-remove-account/index.js
```diff
@@ -22,7 +22,7 @@ class ReallyRemoveAccountPopup extends React.Component<Props, State> {
   state = {
     showingToast: false,
   }
-  _attachmentRef = null
+  _attachmentRef = React.createRef()
```
buoyad: This wasn't attaching correctly

PR #14575, opened 2018-11-05
dont amplify filtered convID list

mmaxim:
cc @maxtaco @malgorithms


mmaxim:
Ok sure, I'll make that change

PR #14532, opened 2018-11-01
iPhone XR support. Better iphone X(s/+/r) support

chrisnojima:
This PR fixes layout for iPhoneXR and cleans up how we deal with the non-safe areas. 
The route tree will inject a safearea above any components that aren't tagged as being underneath the notch

- [x] isIphoneX on iPhone XR
- [x] SafeAreaView in more places
- [x] remove most isIphonex
- [x] compare tabbar to master
- [x] provision
- [x] full screen attachment
- [x] profile
- [x] 5s/X/XR
- [x] storyshots once this pr is approved

PR #14578, opened 2018-11-05
Stellar mobile cleanup: wallet list and platform popups

buoyad:
Refactors `WalletPopup` to make it easier to handle the platform differences. It'll now automatically add a custom header and back button / label with the prop `backButtonType`.

I also hoisted where we were putting `WalletPopup` in to the routable component. So now, `EnterNamePopup` doesn't use a `WalletPopup` and we don't pass through a bunch of props to the buttons. Instead for example, `CreateAccount` and `LinkExisting` wrap with `WalletPopup` and handle setting up the buttons themselves. We were using a bunch of pass through props before that were getting hard to understand, and we would have had to add more to switch up what the header says.

The wallet list cleanup was small, I grabbed `BoxGrow` from #14532 for it.

r? @keybase/react-hackers


shared/wallets/common/wallet-popup.js
```diff
@@ -2,33 +2,50 @@
...
+  containerStyle?: Styles.StylesCrossPlatform,
+  // Header props, only applies on mobile. backButtonType === 'back' renders back button on desktop
+  accountName?: string,
+  backButtonType: 'back' | 'cancel' | 'close', // 'back' -> '<' ; 'cancel' -> 'Cancel' ; 'close' -> 'Close'
+  headerStyle?: Styles.StylesCrossPlatform,
+  headerTitle?: string,
 |}
 
-const WalletPopup = (props: WalletModalProps) => (
+const backButtonTypeToFcnHandle = {
```
buoyad: Translate `onExit` into something `HeaderHOC` understands

shared/wallets/common/wallet-popup.js
```diff
@@ -99,9 +117,13 @@ const styles = Styles.styleSheetCreate({
 })
 
 export default compose(
-  renameProp('onClose', 'onCancel'),
-  withProps({
-    style: Styles.isMobile ? null : {height: 525},
-  }),
+  withProps((props: WalletPopupProps) => ({
```
buoyad: Had some issues with flow in this function (see comments). I think it has to do with recompose typing but I didn't dig too deep.

shared/wallets/common/enter-name-popup/index.js
```diff
@@ -2,54 +2,26 @@
 import * as React from 'react'
```
buoyad: this and `enter-key-popup` are dumber now, pretty much inputs with some decoration

buoyad:
@keybase/react-hackers bump. I added `overflow: 'hidden'` to `WalletPopup` to fix the border radius for things that use it

PR #14585, opened 2018-11-05
[KBFS-3561] Fix downloads on Android

jzila:
This might warrant a hotfix.

### DIFF
--- a/shared/fs/index.stories.js
+++ b/shared/fs/index.stories.js
@@ -81,7 +81,7 @@ const provider = Sb.createPropProviderWithCommon({
   }),
   ViewContainer: () => ({
     url: '/keybase/private/foo/bar.jpg',
-    mimeType: Constants.makeMime({mimeType: 'jpg'}),
+    mimeType: Constants.makeMime({mimeType: 'image/jpeg'}),
     isSymlink: false,
     path: '/keybase/private/foo/bar.jpg',
     onInvalidToken: Sb.action('onInvalidToken'),
--- a/shared/fs/index.stories.js
+++ b/shared/fs/index.stories.js
@@ -81,7 +81,7 @@ const provider = Sb.createPropProviderWithCommon({
   }),
   ViewContainer: () => ({
     url: '/keybase/private/foo/bar.jpg',
-    mimeType: 'jpg',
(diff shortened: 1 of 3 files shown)


taruti:
LGTM, but cannot test at the moment.

songgao:
Would `mime.mimeType` not work?

jzila:
Yes it works, I just like leaving explicit evidence that this is an immutable object in the code.

songgao:
But it's not type-checked anymore right? For example, you could do `mime.get('MimeType')` and it'd pass flow, I think.

jzila:
It appears to me that the type checking here is completely broken anyway, since Flow didn't catch this error in the first place.

jzila:
Does type checking work for Immutable records e.g. with `.mimeType`?

chrisnojima:
i agree, avoid get unless its a dynamic sub thing

songgao:
> It appears to me that the type checking here is completely broken anyway, since Flow didn't catch this error in the first place.

That can't be an argument for giving up type checking there right? I think the reason why flow didn't cache it was probably because of the less typed generator function. But perhaps something like this would work:

```
    const mimeType: Types.MimeType = yield Saga.call(_loadMimeType, path)
```

Also I'm guessing the type checking must be working in `viewTypeFromMimeType`.

shared/fs/index.stories.js
```diff
@@ -81,7 +81,7 @@ const provider = Sb.createPropProviderWithCommon({
   }),
   ViewContainer: () => ({
     url: '/keybase/private/foo/bar.jpg',
-    mimeType: 'jpg',
+    mimeType: Constants.makeMime({mimeType: 'jpg'}),
```
songgao: Did you set this to `jpg` instead of `image/jpeg` because otherwise it'd choose the image viewer?

chrisnojima:
they flow type the same. .get isn't required for records and we shouldn't use it

shared/fs/index.stories.js
jzila: I set it with minimal changes to not be broken any more. Should it be `image/jpeg`?

PR #14587, opened 2018-11-05
Make "To Stellar address" field expand with error message

buoyad:
expand

r? @keybase/react-hackers

### DIFF
--- a/shared/wallets/send-form/participants/to-field.js
+++ b/shared/wallets/send-form/participants/to-field.js
@@ -95,14 +95,13 @@ class ToStellarPublicKey extends React.Component<ToStellarPublicKeyProps, ToStel
       style={styles.toStellarPublicKey}
     >
       <Kb.Box2 direction="vertical" fullWidth={true} style={styles.inputBox}>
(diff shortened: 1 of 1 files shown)


shared/wallets/send-form/participants/to-field.js
```diff
@@ -235,18 +234,16 @@ const styles = Styles.styleSheetCreate({
 
   // ToStellarPublicKey
   toStellarPublicKey: {
-    height: 52,
```
chrisnojima: does this work in mobile

buoyad: No, the send form has a bunch of problems so I didn't work on it since we have another ticket this sprint

chrisnojima: I don't think we can work on these things In isolation. You'll just break the other side

buoyad: I don't think this fix will affect that ticket, bigger layout changes are needed to get it working on mobile

PR #14589, opened 2018-11-05
don't try FindNextMerkleRootAfter with Seqnos < 835903

### LINKED ISSUE #14586: Unable to REKEY two private folders
I have been unable to rekey two of my private folders. I have tried a paper key and opening the two apps listed (my iPhone and my other Mac).  I've looked through other issues here and, while for some people it seems to just fix itself... others have a folder reset. It's persisted for a few days now, not sure what my options are. 

my log id: 642b41217aab88beff370a1c

maxtaco:
- We previous were using `var FirstProdMerkleTreeWithModernShape = keybase1.Seqno(531408)`
as a lefthand guard
- Now let's use `var FirstProdMerkleSeqnoWithSkips = keybase1.Seqno(835903)`
- fix #14586


maxtaco:
Nothing changed, it's just that we didn't set this cutoff high enough when we made this feature, and everything worked as long as someone didn't send up a sequence number between 531408 and 85903, which wasn't very common.

PR #14590, opened 2018-11-05
Markdown crash handler

MarcoPolo:
@keybase/react-hackers

Adds exception handling for exceptional messages in markdown.

Also adds support for nested quotes

### DIFF
--- a/shared/common-adapters/markdown/index.stories.js
+++ b/shared/common-adapters/markdown/index.stories.js
@@ -4,7 +4,7 @@ import * as React from 'react'
 import * as Sb from '../../stories/storybook'
 import * as Kb from '../index'
 import Markdown, {type MarkdownMeta} from '.'
-import {parserFromMeta} from './shared'
+import {simpleMarkdownParser} from './shared'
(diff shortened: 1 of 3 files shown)


MarcoPolo:
This doesn't surface the stack, it just displays the literal unmarkdownd text. I figure it will be a reasonable fallback for most everyone else who happens to run into this, and for us it would make it easier to copy/paste the broken markdown to repro

shared/common-adapters/markdown/index.stories.js
```diff
@@ -246,15 +246,21 @@ class ShowAST extends React.Component<
...
+            inline: false,
+            disableAutoBlockNewlines: true,
+            markdownMeta: this.props.meta,
+          })
+        : OriginalParser.parse(this.props.text, {
+            channelNameToConvID: (channel: string) => null,
+            isValidMention: (mention: string) => false,
+          })
+    } catch (error) {
+      parsed = {error}
```
MarcoPolo: @chrisnojima this is only in the stories when rendering the ast.

PR #14591, opened 2018-11-06
Handle urls without http

MarcoPolo:
@keybase/react-hackers Handles urls that don't have http prefix. 

Just adds `://` if the prefix isn't on there.

### DIFF
--- a/shared/common-adapters/markdown/react.js
+++ b/shared/common-adapters/markdown/react.js
@@ -309,7 +309,7 @@ const reactComponentsForMarkdownType = {
     let url = node.content
 
     if (!url.match(/^https?:\/\//)) {
-      url = `://${node.content}`
+      url = `http://${node.content}`
     }
     return (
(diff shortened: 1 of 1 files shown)


shared/common-adapters/markdown/react.js
```diff
@@ -306,6 +306,11 @@ const reactComponentsForMarkdownType = {
     return <Emoji emojiName={String(node.content).toLowerCase()} size={16} key={state.key} />
   },
   link: (node, output, state) => {
+    let url = node.content
+
+    if (!url.match(/^https?:\/\//)) {
+      url = `://${node.content}`
```
chrisnojima: Maybe this should be https always?

chrisnojima: @malgorithms says use http

MarcoPolo:
```suggestion
      url = `http://${node.content}`
```

PR #14599, opened 2018-11-06
dont let typing in user hud trip the selection accidentally

chrisnojima:
@keybase/react-hackers this fixes the issue where if you put your mouse in the hud area and start typing the filter. before it'd be changing the selection as new nodes popped in and out. now it'll ignore the hover while you're changing the filter for a small period


shared/chat/conversation/input-area/user-mention-hud/index.js
```diff
@@ -99,6 +100,10 @@ class MentionHud extends React.Component<MentionHudProps, State> {
 type ImplProps = MentionHudProps & {|data: Array<Data>, fullList: Array<Data>|}
 class MentionHudImpl extends React.Component<ImplProps> {
   componentDidUpdate(prevProps: ImplProps) {
+    if (this.props.filter !== prevProps.filter) {
```
buoyad: This works for the changing filter case, but when first typing '@' with the mouse nearby the immediate selection still happens.

chrisnojima: sigh, how this component works should change at a higher level. what comes in mutates every time anything changes so i have to do a deeper shallow compare. there's a todo in here to change how this works in general that we should do at some point soon

chrisnojima: ok i fixed this by initing the value (since component did update gets called after the first render). i'm not applying this to channel hud now. we need to clean up how this works first. i'll ticket

buoyad: Sounds good

PR #14604, opened 2018-11-06
Chat flow 85 prep

chrisnojima:
This fixes up ownprops types in chat


chrisnojima:

PR #14606, opened 2018-11-06
Improve generic scraper: twitter card/common meta tags

joshblum:


go/chat/unfurl/scrape_generic.go
```diff
@@ -9,58 +9,136 @@ import (
 	"github.com/keybase/client/go/protocol/chat1"
 )
 
+// Contents are scored based on source. Higher scores win but falsey values
+// always loose.
+const (
+	defaultScore          = 1
+	defaultOpenGraphScore = 10
+	defaultTwitterScore   = 10
```
joshblum: we can tweak these once we start dog fooding

go/chat/unfurl/scrape_generic.go
```diff
@@ -9,58 +9,136 @@ import (
...
+)
+
+func getOpenGraphScore(domain string) int {
+	switch domain {
```
joshblum: and these

go/chat/unfurl/scraper.go
```diff
@@ -20,7 +20,7 @@ func NewScraper(logger logger.Logger) *Scraper {
 }
 
 func (s *Scraper) Scrape(ctx context.Context, uri string) (res chat1.UnfurlRaw, err error) {
-	defer s.Trace(ctx, func() error { return err }, "Scrape(%s)", uri)()
+	defer s.Trace(ctx, func() error { return err }, "Scrape")()
```
joshblum: we shouldn't log this

go/chat/unfurl/utils.go
```diff
@@ -9,10 +9,20 @@ import (
...
 )
 
-func GetHostname(uri string) (res string, err error) {
+func GetDefaultFaviconUrl(uri string) (string, error) {
 	parsed, err := url.Parse(uri)
 	if err != nil {
-		return res, err
+		return "", err
+	}
+	parsed.Path = "favicon.ico"
```
joshblum: this is the default for most sites (keybase doesn't drop a tag on the page but has this path)

protocol/avdl/chat1/unfurl.avdl
```diff
@@ -4,55 +4,67 @@ protocol unfurl {
...
+      string url;
+      string siteName;
+      union { null, string } faviconUrl;
+      union { null, string } imageUrl;
+      union { null, int } publishTime;
+      union { null, string } description;
+
+      // Score each attribute we parse from the webpage. If we encounter
+      // multiple sources we can use the highest rated one.
+      int titleScore;
```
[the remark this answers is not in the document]
joshblum: fixed!

go/chat/unfurl/scrape_generic.go
```diff
@@ -9,58 +9,198 @@ import (
...
+			s.setAndParsePubTime(ctx, content, generic, defaultScore)
+		case "lastmod":
+			s.setAndParsePubTime(ctx, content, generic, defaultScore)
+		}
+	})
+
+	// scrape title
 	c.OnHTML("head title", func(e *colly.HTMLElement) {
-		generic.Title = e.Text
+		generic.setTitle(e.Text, defaultScore)
```
mmaxim: Are there any other common ones for this? OpenGraph must have something which should probably outrank this guy,

go/chat/unfurl/scrape_generic.go
```diff
@@ -9,58 +9,198 @@ import (
...
 	})
+
+	// scrape favicon
 	c.OnHTML("head link[rel][href]", func(e *colly.HTMLElement) {
 		rel := strings.ToLower(e.Attr("rel"))
-		if strings.Contains(rel, "shortcut icon") {
-			generic.FaviconUrl = new(string)
-			*generic.FaviconUrl = fullURL(hostname, e.Attr("href"))
+		if strings.Contains(rel, "shortcut icon") ||
```
mmaxim: Could you add scores for these? Some of these are higher res (like `apple-touch-icon` (I would prefer the higher res ones).

go/chat/unfurl/scrape_generic.go
```diff
@@ -9,58 +9,198 @@ import (
 	"github.com/keybase/client/go/protocol/chat1"
 )
 
+// Contents are scored based on source. Higher scores win but falsey values
+// always loose.
+const (
+	defaultScore          = 1
+	defaultOpenGraphScore = 10
```
mmaxim: Also I think OpenGraph should win over Twitter, they are more likely to be generally applicable.

mmaxim: Sorry one other comment, I think these scores should also be per attribute from the generic unfurl for maximum flexibility. They can "inherit" from this default one some way, but I thought be able to tweak the score for title separately from image.

joshblum: these technically win now since first writer wins for equal scores but i can bump open graph to slightly higher. for per attribute i think that should be determined within the switch statement closest to the processing logic for that tag. if that gets unwieldy we can make a map with comments, but i think it's more likely to be domain not tag specific

go/chat/unfurl/scrape_generic.go
joshblum: yup `og:title`  https://github.com/keybase/client/pull/14606/files#diff-f5a8c5f7c1b29454bfde4a1ff31fc10eR142 or `twitter:title` https://github.com/keybase/client/pull/14606/files#diff-f5a8c5f7c1b29454bfde4a1ff31fc10eR166 but i'll move this up so it flows better

go/chat/unfurl/scrape_generic.go
joshblum: sgtm

go/chat/unfurl/scrape_generic.go
mmaxim: Let’s talk about it when I get in

go/chat/unfurl/scrape_generic.go
mmaxim: Note this is an example why I think per attribute scores make sense

mmaxim: Note this is an example why I think per attribute scores make sense

joshblum:
@mmaxim update ptal!

PR #14608, opened 2018-11-06
Add RPC call to encrypt KBFS favorites cache

jakob223:
Add an RPC call to encrypt the KBFS favorites cache so it can be written to disk.


jakob223:
@jzila simplified a bunch by making a few `encryptedDB` methods public.

PR #14615, opened 2018-11-08
remove dev menu and move user timings

chrisnojima:
@keybase/react-hackers


chrisnojima:
@keybase/react-hackers bump

PR #14628, opened 2018-11-09
owntypes for provision

chrisnojima:
@keybase/react-hackers

### DIFF
--- a/shared/provision/paper-key/index.js
+++ b/shared/provision/paper-key/index.js
@@ -1,75 +1,74 @@
 // @flow
 import * as React from 'react'
 import * as Constants from '../../constants/provision'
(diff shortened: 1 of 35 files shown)


shared/provision/paper-key/index.js
```diff
@@ -1,75 +1,74 @@
...
-          fullWidth={true}
-          onClick={props.onSubmit}
-          enabled={!!props.paperKey}
-          waitingKey={Constants.waitingKey}
-        />
-      </ButtonBar>
-    </Box2>
-  </Box2>
-)
+class PaperKey extends React.Component<Props, {paperKey: string}> {
```
MarcoPolo: Might as well be a PureComponent right?

chrisnojima: I'm not so worried about pure or not on these connected components as the connect hoc is pure

PR #14630, opened 2018-11-09
Remove background color from chat's text input.

adamjspooner:
@keybase/react-hackers, this is ready for review. There's a problem on Windows where the text input on chat overflows its container by 1 pixel and covers the border. I wasn't able to correct the overflow (without specifying exact dimensions, which feels wrong). So, this removes the background color which doesn't seem necessary.

@chrisnojima, will removing the background colors cause any rendering performance degradation?


adamjspooner:
Cool. Thanks!

PR #14634, opened 2018-11-09
Fix hover states and add reactions.

adamjspooner:
@cecileboucheron, I'm not sure about the hover state for usernames. This adds an underline everywhere in the app, but I don't know if that's desired. Do you mind taking a look?


cecileboucheron:
LGTM - maybe fix the "You added you" if possible, otherwise let's fix it in another PR

adamjspooner:

PR #14639, opened 2018-11-10
add timestamps to join messages

mmaxim:
@keybase/react-hackers 

Otherwise you can't see anything about when someone joined (no ... menu for joins)


mmaxim:
Also I think this got worse after the change to drop the divider timestamp, since that was giving some information at least.

mmaxim:
@chrisnojima since you seem to be around, did you have any opinion on this? These join/leave messages got left behind I think with the change for removing the timestamp divider.

PR #14647, opened 2018-11-12
NOMERGE v 2.10.0 hotfix

chrisnojima:
@keybase/react-hackers 

-[x] turns off simplemarkdown on mobile

### DIFF
--- a/shared/common-adapters/markdown/index.native.js
+++ b/shared/common-adapters/markdown/index.native.js
@@ -188,7 +188,11 @@ class OriginalMarkdown extends React.PureComponent<Props> {
         return createComponent('', '', content, {}) || null
       } else {
         return (
-          <Text type="Body" style={this.props.style} allowFontScaling={this.props.allowFontScaling}>
+          <Text
+            type="Body"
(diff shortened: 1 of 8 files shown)


chrisnojima:
- [x] kbfs commit: 8c3f1a4c8189f3e3603f1f741376ea4746518af5

shared/common-adapters/markdown/index.native.js
```diff
@@ -188,7 +188,11 @@ class OriginalMarkdown extends React.PureComponent<Props> {
         return createComponent('', '', content, {}) || null
       } else {
         return (
-          <Text type="Body" style={this.props.style} allowFontScaling={this.props.allowFontScaling}>
+          <Text
+            type="Body"
+            style={collapseStyles([styles.neutral, this.props.style])}
```
mmaxim: I'm a little confused why we have `styles.neutral` in here at all? What is it doing if it is just empty? Does it get set somewhere else?

chrisnojima: it effectively does nothing but there's no point in cleaning this up. next release will have simple-markdown

PR #14650, opened 2018-11-12
pass in xdg cache home environ to systemd unit

heronhaye:


maxtaco:
It looks good to me. Can you check that logsend still works?

heronhaye:
Yes looks good: https://keybase.io/_/admin/logdump_viewer?196091806#5b3868eed7543a0f9e61261c.

heronhaye:
So the GUI was using some XDG variables in `shared/constants/platform.desktop.js` but as noted in the other PR, none of them were being forwarded through systemd so they weren't being used (at least on Linux). With this change I made a new user account and no `~/.cache/keybase` dir was created after run_keybase.

eli-schwartz:
> Well, I know the "righter" way to do this is to use an systemd environment file but since we already have run_keybase, doing it in there works too and we don't have to worry about making that file in every os.

Great -- more instances where Keybase operates in an entirely dirty fashion.

Now people who start the keybase command-line tool via an enabled unit have a *broken* service, since it will be started at the same time the systemd service manager starts up -- which means when your user logs in. Which is generally a long time before the `run_keybase` script is run. Which means it tries to write to the file `/keybase.service.log`. Some indeterminate time later when the user chooses to start the GUI, the `run_keybase` script will force-restart the service and "fix" /usr/bin/keybase itself, assuming the user is both currently on a non-headless system and actually desired to run the electron app in the first place.

Is now a good time to put on my Linux distro developer hat and say I hate that `run_keybase` script and refuse to ship it on my distro?

PR #14651, opened 2018-11-12
Add prefetch status to SimpleFSStat output, modify protocol accordingly

jakob223:
See https://github.com/keybase/kbfs/pull/1904

### DIFF
--- a/go/client/cmd_simplefs_stat.go
+++ b/go/client/cmd_simplefs_stat.go
@@ -97,10 +97,11 @@ func (c *CmdSimpleFSStat) Run() (err error) {
 
 		for _, r := range res.Revisions {
 			e := r.Entry
-			ui.Printf("%d)\t%s\t%s\t%d\t%s\t%s\n",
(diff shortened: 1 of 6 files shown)


go/client/cmd_simplefs_stat.go
```diff
@@ -97,21 +97,23 @@ func (c *CmdSimpleFSStat) Run() (err error) {
 
 		for _, r := range res.Revisions {
 			e := r.Entry
-			ui.Printf("%d)\t%s\t%s\t%d\t%s\t%s\n",
+			ui.Printf("%d)\t%s\t%s\t%d\t%s\t%s\t%s\n",
 				r.Revision, keybase1.FormatTime(e.Time),
 				keybase1.DirentTypeRevMap[e.DirentType],
-				e.Size, e.Name, e.LastWriterUnverified.Username)
+				e.Size, e.Name, e.LastWriterUnverified.Username,
+				keybase1.PrefetchStatusRevMap[e.PrefetchStatus])
```
strib: No need to have the explicit map lookup, since the `%s` will cause the existing `String()` method to be called.  (Here and below.)

PR #14653, opened 2018-11-12
forward xdg_download_dir in linux

heronhaye:
Problem is systemd (I think) doesn't forward env variables to its processes for security, so pass in `XDG_DOWNLOAD_DIR` explicitly. Jack says (in comment) it should be done by default on Arch but it wasn't working for me, and wouldn't work on older Ubuntus anyway.

### DIFF
--- a/packaging/linux/run_keybase
+++ b/packaging/linux/run_keybase
@@ -204,6 +204,7 @@ start_systemd() {
       QT_IM_MODULE
       QT4_IM_MODULE
       XMODIFIERS
+      XDG_DOWNLOAD_DIR
   )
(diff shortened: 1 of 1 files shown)


heronhaye:
No, we do. I actually added in the XDG support a while ago but was confused why it was working in dev but not in packaged version. It's because (I believe) systemd doesn't pass all environment variables to its processes by default, other than, according to Jack, Xsessions on some new OSes. So this forwards the variable into the GUI process.

heronhaye:
ref: https://github.com/keybase/keybase-issues/issues/3296

PR #14659, opened 2018-11-13
disable send button on each recipient change

zanderz:
by resetting `readyToSend` in the built payment every change, outside the debounce


zanderz:
ok thanks. But how do you get it from the field to the footer button? just local state in the send form you mean?

PR #14665, opened 2018-11-13
Call buildPayment when we receive display currency for the send form

buoyad:
Fixes the issue where the "You must send at least ..." banner doesn't show up until you start typing in the send form. We were already calling `buildPayment` on opening the send form. The banner wasn't showing because this case:

`a.type === WalletsGen.displayCurrencyReceived && a.payload.setBuildingCurrency`

was changing `state.wallets.building` without calling `buildPayment`. Then the reducer for `buildPayment` would get a payload with banners, but since `building` has changed we throw it out. So we need to make sure any mutations to `wallets.building` also goes through the `buildPayment` saga.

r? @keybase/react-hackers

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -24,8 +24,12 @@ import {RPCError} from '../util/errors'
 import {isMobile} from '../constants/platform'
 import {actionHasError} from '../util/container'
 
-const buildPayment = (state: TypedState, action: any) =>
-  (state.wallets.building.isRequest
+const buildPayment = (state: TypedState, action: any) => {
+  if (action.type === WalletsGen.displayCurrencyReceived && !action.payload.setBuildingCurrency) {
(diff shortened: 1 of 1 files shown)


shared/actions/wallets.js
```diff
@@ -669,14 +669,16 @@ function* walletsSaga(): Saga.SagaGenerator<any, any> {
   )
   yield Saga.safeTakeEveryPure(WalletsGen.accountsReceived, maybeSelectDefaultAccount)
   yield Saga.actionToPromise(
-    [
```
chrisnojima: maybe easier to read if its just in the handler. thats our typical pattern. we usually listen for N actions but handle cases in the call itself so maybe move this logic to buildPayment and just bail if type is displayCurrencyReceived && !setBuildCurrency in payload

buoyad: Thanks, done

PR #14666, opened 2018-11-13
client: add `keybase fs sync` command and protocol changes

strib:
For enabling/disabling syncing of individual TLFs, and viewing their status as well.

Requires the corresponding KBFS change to pass CircleCI.

Issue: KBFS-3510


strib:
@songgao thanks!  Please review d4708e8, which splits it into three subcommands.

songgao:
New commit looks good to me. Thanks for implementing the change!

PR #14667, opened 2018-11-13
unfurling: add a new RPC to take action on the prompt CORE-9251

mmaxim:
Patch does the following:

1.) Adds the `ResolveUnfurlPrompt` RPC in order to take action on the unfurl prompt for people in whitelist mode.
2.) Integrate into the test.


go/chat/server.go
```diff
@@ -2286,3 +2286,59 @@ func (h *Server) GetStaticConfig(ctx context.Context) (res chat1.StaticConfig, e
...
+	defer h.Trace(ctx, func() error { return err }, "ResolveUnfurlPrompt")()
+	uid, err := utils.AssertLoggedInUID(ctx, h.G())
+	if err != nil {
+		return err
+	}
+	atyp, err := arg.Result.ActionType()
+	if err != nil {
+		return err
+	}
+	fetchAndUnfurl := func() error {
```
joshblum: nit: move this definition up so `atyp` is defined right before we do the `switch`

go/chat/server.go
```diff
@@ -2286,3 +2286,59 @@ func (h *Server) GetStaticConfig(ctx context.Context) (res chat1.StaticConfig, e
...
+			return err
+		}
+		msgs, err := h.G().ConvSource.GetMessages(ctx, conv, uid, []chat1.MessageID{arg.MsgID}, nil)
+		if err != nil {
+			return err
+		}
+		if len(msgs) != 1 {
+			return errors.New("message not found")
+		}
+		h.G().Unfurler.UnfurlAndSend(ctx, uid, arg.ConvID, msgs[0])
```
joshblum: won't this cause problems since it's not idempotent? we'll just keep requeuing the same URLs if there are multiple present in the message? we should add this in the tests but i think the unfurler should have some notion of deduping before making a new task

mmaxim: Hmm you could be right, let me think about it.

mmaxim: Ok @joshblum, PTAL!

go/chat/unfurl/unfurler.go
```diff
@@ -203,13 +209,27 @@ func (u *Unfurler) makeBaseUnfurlMessage(ctx context.Context, fromMsg chat1.Mess
...
+		return
+	}
 	// get URL hits
 	hits := u.extractURLs(ctx, uid, msg)
 	if len(hits) == 0 {
 		return
 	}
+	// get a map for all the URLs we have already unfurled
+	prevUnfurled := make(map[string]bool)
+	for _, u := range msg.Valid().Unfurls {
```
joshblum: isn't there a race for this to show up? maybe also check the urls present in `unfurlMap` in case a dup is pending?

go/chat/unfurl/unfurler_test.go
```diff
@@ -115,6 +115,7 @@ func TestUnfurler(t *testing.T) {
 	}
 	require.NoError(t, settings.WhitelistAdd(context.TODO(), uid, "0.1"))
 
+	unfurler.UnfurlAndSend(context.TODO(), uid, convID, fromMsg)
```
joshblum: throw is in a 5-10x loop? i'm sure CI would have fun with creating interesting execution flows

PR #14669, opened 2018-11-13
Make red and green usernames more consistent around the app

buoyad:
cc @chrisnojima there were a couple problems with updating just that one field in the tracker state, so I made it kick off `Tracker.getProfile`s when we get one of these `updateBrokenState` actions.

This PR makes it so the `state.tracker` and `state.users` are on the same page when it comes to a user's tracker state. We already updated `state.users` when we got trackers with newly broken or fixed users, with this we'll fetch trackers if we find out things changed from elsewhere (which right now is only `ChatIdentifyUpdate`s). 

I guess there's some risk that a client could get bogged down in `identify2`s (not sure how common the `ChatIdentifyUpdate`s are), on my main account this only happens 3-4 times while clicking around conversations.

r? @keybase/react-hackers

### DIFF
--- a/shared/actions/json/users.json
+++ b/shared/actions/json/users.json
@@ -2,6 +2,7 @@
   "prelude": [],
   "actions": {
     "updateBrokenState": {
+      "fromTracker?": "boolean",
       "newlyBroken": "Array<string>",
       "newlyFixed": "Array<string>"
     },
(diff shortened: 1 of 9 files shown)


buoyad:
Talked with @chrisnojima about this and am going to change it to directly populate the tracker state on account of how perf-heavy identify can be.

buoyad:
@chrisnojima I added some more fields to the tracker state, the tracker seems to act normal now. Tested opening the tracker popup after getting one of these and 'accept'ing, it worked as expected. The `TrackerState` has some non-optional fields that I couldn't fill in here, I assume anything that's going to display this will load the rest of it first?

PR #14671, opened 2018-11-13
 ownprops - teams

chrisnojima:


chrisnojima:

shared/teams/role-picker/controlled-container.js
```diff
@@ -23,8 +23,16 @@ export type ControlledRolePickerProps = {
   showNotificationCheckbox?: boolean,
   sendNotificationChecked?: boolean,
   styleCover?: Object,
+  teamname: string,
 }
 
+type OwnProps = RouteProps<
+  {
+    ...ControlledRolePickerProps,
```
MarcoPolo: Why like this instead of just `ControlledRolePickerProps` ?

chrisnojima: i wasn't sure if more stuff would have to be hoisted in there or not while doing it

shared/teams/invite-by-email/container.native.js
```diff
@@ -121,7 +124,7 @@ export default compose(
         _setHasPermission: () => hasPermission => ({hasPermission}),
       }
     ),
-    withPropsOnChange(['onExitSearch'], props => ({
```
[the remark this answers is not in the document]
chrisnojima: i think this gets called once always, but it likely never mutated?

PR #14674, opened 2018-11-14
Add a test for relay payment with reset user

zapu:
Add a test for relay payments claiming and yanking for reset situation.

Claim test simulates situation where someone resets, you add them back to the team, they accept disclaimer and claim the payment.

Yank test simulates yanking the payment back by sender after counterparty resets. Imagine clicking "cancel" on relay payment when the red skull bar is visible.

These tests use mock because I figured we care more about client being able to decrypt relay account keys rather than talking to real testnet. Let me know if this is bad assumption.

### DIFF
--- a/go/stellar/stellarsvc/service_test.go
+++ b/go/stellar/stellarsvc/service_test.go
@@ -489,15 +489,15 @@ func TestRelayTransferInnards(t *testing.T) {
 	require.Equal(t, "hey", relaySecrets.Note)
 }
 
-func TestRelayClaim(t *testing.T) {
-	testRelay(t, false)
+func TestRelaySBSClaim(t *testing.T) {
+	testRelaySBS(t, false)
 }
 
-func TestRelayYank(t *testing.T) {
-	testRelay(t, true)
+func TestRelaySBSYank(t *testing.T) {
+	testRelaySBS(t, true)
 }
 
-func testRelay(t *testing.T, yank bool) {
+func testRelaySBS(t *testing.T, yank bool) {
 	tcs, cleanup := setupTestsWithSettings(t, []usetting{usettingFull, usettingPukless})
 	defer cleanup()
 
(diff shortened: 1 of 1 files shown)


go/stellar/stellarsvc/service_test.go
```diff
@@ -635,6 +635,91 @@ func testRelay(t *testing.T, yank bool) {
...
+		claimant = 1
+	} else {
+		// User0 will try to claim the funds back without readding user1 to the
+		// impteam. Also do not accept disclaimer as user1.
+		claimant = 0
+	}
+
+	history, err := tcs[claimant].Srv.RecentPaymentsCLILocal(context.Background(), nil)
+	require.NoError(t, err)
+	require.Len(t, history, 1)
```
mlsteele: If the claimant is user 0 I'd expect it to see in its history the gift. I guess this is an inaccuracy in the mock. You could only look at the last payment in the list to be future proof against someone fixing the mock. I don't feel strongly.

go/stellar/stellarsvc/service_test.go
```diff
@@ -635,6 +635,91 @@ func testRelay(t *testing.T, yank bool) {
...
+	require.Nil(t, fhistory[0].Err)
+	require.NotNil(t, fhistory[0].Payment)
+	require.NotEmpty(t, fhistory[0].Payment.Id)
+	require.NotZero(t, fhistory[0].Payment.Time)
+	require.Equal(t, stellar1.PaymentStatus_CLAIMABLE, fhistory[0].Payment.StatusSimplified)
+	require.Equal(t, "claimable", fhistory[0].Payment.StatusDescription)
+
+	res, err := tcs[claimant].Srv.ClaimCLILocal(context.Background(), stellar1.ClaimCLILocalArg{TxID: txID.String()})
+	require.NoError(t, err)
+	require.NotEqual(t, "", res.ClaimStellarID)
```
mlsteele: In reality when yank=false, the recipient's autoclaim should do the claiming. If that system doesn't work with these mocks that's fine.

zapu: right, that's why I was surprised after being reminded that we do autoclaims. Because these tests do poll or wait for autoclaim but call the RPC manually. Maybe mocks do not send the autoclaim notification, or test users do not run right services to be able to hear the notification. Should we try to fix? I'd definitely be for that, even if in different ticket.

PR #14676, opened 2018-11-14
Fix perma-spinner on assets/payments for a new account

cjb:
@keybase/react-hackers 

This bug is pretty frustrating!  Here's the deal:

Normally when we get a PaymentStatusNotification for an account, we want to load assets/payments for that account in response, to pick up the new payment info.  But there are some situations where we don't.  There's been a bug when a canceled relay payment gives us the accountID of the relay account, which black bars when we try to list its assets and payments.  Or, since these notifications are not intended to be absolutely synchronous, we could delete an account and then later get a notification for a payment on it, and that would black bar if we tried to update based on it.

So, there are good reasons to avoid following through with listAssets or listPayments for an account that isn't in our accountMap.  Doing so avoids both of the classes of black bar above.

But!  In the case where we just created or linked a new account and want to populate its assets/payments, this causes a problem: the newly-created account isn't going to be in our accountMap immediately.  There's a race between firing off a call to listAssets/listPayments and receiving the new account in listAccounts.  The race causes us to decide not to bother calling listAssets/listPayments for this new account because the new account doesn't exist in our accountMap yet, until listAccounts has returned.

One way to solve this might be to have the reducer for createdNewAccount put the new accountID into the accountMap.  But that's unsatisfying -- it's an orderedMap so there would be jankiness if the service puts the new account in a different order, and we don't know things like the balance of the new account yet.

Instead, I went with a new data structure called `newlyCreatedAccounts`.  The reducer for createdNewAccount just throws the account ID in there, and then in loadAssets/loadPayments we are willing to call the RPC if the account's in our accountMap, or it's in our newlyCreatedAccounts.

The next time we get an accountsReceived, we clear out newlyCreatedAccounts -- at that point, the new account should be in the accountsReceived payload, so there's no need to keep the extra copy of it anymore.  If we didn't do that, we could hit this:

* make new account
* add it to newlyCreatedAccounts
* load its assets
* the user deletes it
* a notification comes in for a received payment on it
* we go ahead and call listPayments since it's in newlyCreatedAccounts, even though it was deleted
* we black bar

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -239,7 +239,8 @@ const loadAssets = (
     | WalletsGen.RefreshPaymentsPayload
 ) =>
   !actionHasError(action) &&
-  Constants.getAccount(state, action.payload.accountID).accountID !== Types.noAccountID &&
+  (Constants.getAccount(state, action.payload.accountID).accountID !== Types.noAccountID ||
+    Constants.isNewlyCreatedAccount(state, action.payload.accountID)) &&
   RPCStellarTypes.localGetAccountAssetsLocalRpcPromise({accountID: action.payload.accountID}).then(res =>
     WalletsGen.createAssetsReceived({
       accountID: action.payload.accountID,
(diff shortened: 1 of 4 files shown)


chrisnojima:
can we discuss this over meet or something i have some questions

cjb:
@chrisnojima Great call! This small change seems to work instead.

PR #14678, opened 2018-11-14
Add lottie support

chrisnojima:
- [x] electron
- [x] ios
- [x] android


chrisnojima:
replaces https://github.com/keybase/client/pull/13686

shared/common-adapters/animation-data.json
```diff
@@ -0,0 +1,277 @@
+{
```
chrisnojima: note, the old pr didn't work with storyshots due to this naming collision. it saw animation.json and thought it shoudl load that. so in general we want to just avoid that so renaming this animation-data.json
cc: @adamjspooner

shared/common-adapters/animation.desktop.js
```diff
@@ -0,0 +1,39 @@
...
+            height: this.props.height || defaultDimension,
+            options: {animationData: animationData[this.props.animationType]},
+            style: this.props.style,
+            width: this.props.width || defaultDimension,
+          })}
+        </Box>
+      )
+    }
+    // jest pukes if the import is on top so just defer till render
+    const Lottie = require('lottie-react-web').default
```
chrisnojima: on import this thing queries canvas so if that doesn't exist (like in storyshots) then it'll die so we just move it down here. it is possible to make a fake canvas for jest but its really not worth it for us

chrisnojima:
to test just go to chat and see the animation dots moving when another user is typing
cc: @adamjspooner @cecileboucheron 
@keybase/react-hackers

chrisnojima:
CI tests will pass once we update yarn on jenkins

chrisnojima:
@jzila this is dep on the jenkins update

chrisnojima:
@jzila actually this isn't blocked on yarn but we should update it, so you can back burner that tnx!

PR #14683, opened 2018-11-14
[KBFS-3596] Fix cannot set on an immutable record error

jzila:
So I don't have a repro, but I went through the code, including Immutable, and it seems that this happens when an Immutable record doesn't have a clear `__ownerID`. It's not clear to me when that happens. But changing the mutability will cause it to be reset, and this error occurred after `_loadMimeType`. So it's likely happening in the reducer in `mimeTypeLoaded`. I have changed that to call `withMutations`, which should reset the mutability.

### DIFF
--- a/shared/reducers/fs.js
+++ b/shared/reducers/fs.js
@@ -116,10 +116,14 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
     }
     case FsGen.downloadProgress: {
       const {key, completePortion, endEstimate} = action.payload
-      return state.updateIn(
-        ['downloads', key, 'state'],
-        original =>
-          original && original.set('completePortion', completePortion).set('endEstimate', endEstimate)
+      return state.withMutations(s => s
+        .updateIn(
+          ['downloads', key, 'state'],
+          original =>
+            original && original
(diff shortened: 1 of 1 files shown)


chrisnojima:
please look at https://github.com/keybase/client/pull/14701

jzila:
@chrisnojima ready for another look.

jzila:
[w=1](https://github.com/keybase/client/pull/14683/files?w=1) helps

shared/reducers/fs.js
```diff
@@ -116,10 +116,17 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
-        ['downloads', key, 'state'],
-        original =>
-          original && original.set('completePortion', completePortion).set('endEstimate', endEstimate)
+      return state.update('downloads', d =>
+        d.update(key, k =>
+          k.update(
+            'state',
+            original =>
+              original &&
+              original.withMutations(o =>
```
chrisnojima: does original.merge work?

jzila: I'll try.

jzila:
ping

PR #14685, opened 2018-11-14
Fix multiline placeholder for mobile "to stellar address"

buoyad:
Adds the placeholder as an overlay over the input field. Also makes it so the error message expands the box correctly on mobile. r? @keybase/react-hackers

### DIFF
--- a/shared/common-adapters/clickable-box.native.js
+++ b/shared/common-adapters/clickable-box.native.js
@@ -22,7 +22,7 @@ class ClickableBox extends React.Component<Props> {
             pointerEvents={props.pointerEvents}
             style={clickStyle}
             underlayColor={props.underlayColor || globalColors.white}
-            activeOpacity={0.7}
+            activeOpacity={this.props.activeOpacity ?? 0.7}
           >
             {props.children}
           </TouchableOpacity>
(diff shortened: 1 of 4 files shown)


shared/common-adapters/clickable-box.native.js
```diff
@@ -22,7 +22,7 @@ class ClickableBox extends React.Component<Props> {
             pointerEvents={props.pointerEvents}
             style={clickStyle}
             underlayColor={props.underlayColor || globalColors.white}
-            activeOpacity={0.7}
+            activeOpacity={this.props.activeOpacity ?? 0.7}
```
buoyad: This prop always existed but was unused. There are a couple places where we set it to .6 or .8, I don't think this will affect things much but I can remove them if someone wants, cc @adamjspooner?

adamjspooner: This is just the opacity of the container when it's being pressed, right?

buoyad: Yep

adamjspooner: LGTM!

PR #14686, opened 2018-11-14
Don't log no such key errors in readLastSentXLM

buoyad:
Uses `string.includes` to figure out whether it's this "no such key" error, since it uses the generic error code. This is mostly to keep it from polluting our logs and bugging us during dev, cc @chrisnojima @keybase/react-hackers

### DIFF
--- a/shared/actions/config/index.js
+++ b/shared/actions/config/index.js
@@ -409,7 +409,12 @@ const readLastSentXLM = () => {
       logger.info(`Successfully read config stellar.lastSentXLM: ${String(value)}`)
       return createSetLastSentXLM({lastSentXLM: value, writeFile: false})
     })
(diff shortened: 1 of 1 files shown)


shared/actions/config/index.js
```diff
@@ -409,7 +409,12 @@ const readLastSentXLM = () => {
       logger.info(`Successfully read config stellar.lastSentXLM: ${String(value)}`)
       return createSetLastSentXLM({lastSentXLM: value, writeFile: false})
     })
-    .catch(err => logger.error(`Error reading config stellar.lastSentXLM: ${err.message}`))
+    .catch(
```
chrisnojima: i think we can just ignore the error. if it doesn't read its ok

buoyad: if it doesn't read and it's set that'd be a bug, though

PR #14690, opened 2018-11-14
protocol: add local disk bytes to sync status

strib:
And print the available space in the `keybase fs sync -s` command.

Issue: KBFS-3518


strib:
> @strib is this communicating the bytes for the sync cache, or overall?

@jzila this one is just for the overall bytes of the local disk that holds the sync cache.  A future PR will more return the total synced bytes on a per-TLF basis.  I don't think we currently have a plan/need to have the total bytes in the sync cache reported over the protocol, but it would be easy to add.

PR #14694, opened 2018-11-15
unfurling: storybook for the unfurl prompt CORE-9474

mmaxim:
@keybase/react-hackers 

Story for the unfurl prompt if the user is in whitelist mode, but has not authorized the domain that is being unfurled. 

image


shared/chat/conversation/messages/unfurl/prompt/index.js
```diff
@@ -0,0 +1,70 @@
...
+const styles = Styles.styleSheetCreate({
+  container: {
+    ...Styles.globalStyles.flexBoxRow,
+    backgroundColor: Styles.globalColors.blue5,
+    borderRadius: Styles.borderRadius,
+    paddingTop: 8,
+    paddingBottom: 8,
+  },
+  choiceContainer: {
+    width: 370,
```
chrisnojima: is the intention of this PR that this works on the phone also? If so this is likely incorrect. We always build against an iphone5s sim to ensure things look ok on small form factor phones. Although this likely passes flow, we don't strictly type a bunch of the css. it just types as string, so things like `marginLeft: 'auto'` below work (as we want to allow `marginLeft: 120.3px` or whatever, but auto isn't a thing in RN

shared/chat/conversation/messages/unfurl/prompt/index.stories.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import * as React from 'react'
+import {storiesOf} from '../../../../../stories/storybook'
+import {Box} from '../../../../../common-adapters/index'
+import UnfurlPrompt from '.'
+
+const props = {
+  domain: 'cnn.com',
+  onAlways: () => {},
```
chrisnojima: You should import * as Sb, then have these be `Sb.action('onAlways')`. This plumbs the buttons to the actions panel in storybook so you can see that clicking is actually working

shared/chat/conversation/messages/unfurl/prompt/index.js
mmaxim: There isn't a design on the phone yet, so I didn't really look at it there. I was just going to have this off for phone in the short term. I could drop this in for platform specific right?

mmaxim: `platformStyles` or whatever

chrisnojima: the preferred solution is to use 
```
Styles.platformStyles({
  common: {},
  isElectron: {},
  isMobile: {},
})
```

mmaxim: I'll change it to that then.

PR #14695, opened 2018-11-15
semicolon's revenge

songgao:


shared/common-adapters/web-view.native.js
```diff
@@ -15,9 +15,9 @@ const combineJavaScriptAndCSS = (injections?: WebViewInjections) =>
   document.body.appendChild(style)
   style.type = 'text/css'
   style.appendChild(document.createTextNode(\`${escape(injections.css)}\`))
-})()
+})();
 
-(function() {\` ${escape(injections.javaScript)} \`})()
+(function() { ${escape(injections.javaScript)} })();
```
songgao: We weren't using injected JS on native side, so this extra quote didn't break anything. But might as well fix it ...

shared/common-adapters/web-view.native.js
```diff
@@ -15,9 +15,9 @@ const combineJavaScriptAndCSS = (injections?: WebViewInjections) =>
   document.body.appendChild(style)
   style.type = 'text/css'
   style.appendChild(document.createTextNode(\`${escape(injections.css)}\`))
-})()
+})();
```
songgao: This is the actual fix. Without this, the parenthesis below become a function call. I can't believe I only realized that now!

PR #14705, opened 2018-11-15
unfurling: storybook for generic unfurl display with bottom+side image CORE-9475

mmaxim:
@keybase/react-hackers 

This shows a "generic" unfurl, which is a scraper mode where we have no special information about how to display the URL.


shared/chat/conversation/messages/unfurl/generic/index.js
```diff
@@ -0,0 +1,118 @@
...
+  showImageOnSide: boolean,
+}
+
+class UnfurlGeneric extends React.PureComponent<Props> {
+  render() {
+    return (
+      <Kb.Box2 style={styles.container} direction="horizontal">
+        <Kb.Box2 style={styles.innerContainer} fullWidth={true} direction="vertical">
+          <Kb.Box2 style={styles.siteNameContainer} fullWidth={true} direction="horizontal">
+            {this.props.faviconURL && <Kb.Image src={this.props.faviconURL} style={styles.favicon} />}
```
chrisnojima: should be !!this.props.faveiconURL else you'll crash out in RN

shared/chat/conversation/messages/unfurl/generic/index.js
```diff
@@ -0,0 +1,118 @@
...
+            <Kb.Text type="BodySmall" style={styles.siteName}>
+              {this.props.siteName}
+            </Kb.Text>
+            {this.props.publishTime && (
```
chrisnojima: same here

shared/chat/conversation/messages/unfurl/generic/index.js
```diff
@@ -0,0 +1,118 @@
...
+            {this.props.onClose && (
+              <Kb.Box2 direction="horizontal" style={styles.closeContainer}>
+                <Kb.Icon type="iconfont-close" onClick={this.props.onClose} fontSize={10} />
+              </Kb.Box2>
+            )}
+          </Kb.Box2>
+          <Kb.Text type="BodyPrimaryLink" style={styles.url} onClickURL={this.props.url}>
+            {this.props.title}
+          </Kb.Text>
+          {this.props.description && <Kb.Text type="Body">{this.props.description}</Kb.Text>}
```
chrisnojima: same here

shared/chat/conversation/messages/unfurl/generic/index.js
```diff
@@ -0,0 +1,118 @@
...
+          {this.props.imageURL &&
```
chrisnojima: same here

shared/chat/conversation/messages/unfurl/generic/index.js
```diff
@@ -0,0 +1,118 @@
...
+            !this.props.showImageOnSide && <Kb.Image src={this.props.imageURL} style={styles.bottomImage} />}
+        </Kb.Box2>
+        {this.props.imageURL &&
```
chrisnojima: same here

shared/chat/conversation/messages/unfurl/generic/index.js
```diff
@@ -0,0 +1,118 @@
...
+    },
+  }),
+  siteName: Styles.platformStyles({
+    isElectron: {
+      marginLeft: 8,
+    },
+  }),
+  bottomImage: Styles.platformStyles({
+    isElectron: {
+      marginTop: 8,
```
chrisnojima: you can likely get rid of these inner margins by using Box2's `gap` property, just do gap='tiny' or whatever (defined in Styles.globalMargins) and it'll inject spaces between the children

shared/chat/conversation/messages/unfurl/generic/index.stories.js
```diff
@@ -0,0 +1,87 @@
...
+import {Box} from '../../../../../common-adapters/index'
+import * as Sb from '../../../../../stories/storybook'
+
+import UnfurlGeneric from '.'
+
+const full = {
+  title: 'U.S. Stocks Jump as Tough Month Sets to Wrap',
+  url: 'https://www.wsj.com/articles/global-stocks-rally-to-end-a-tough-month-1540976261',
+  siteName: 'WSJ',
+  publishTime: Date.now() - 86400 * 1000,
```
chrisnojima: this will cause your shots to change everytime, it needs to be fixed or mocked

shared/chat/conversation/messages/unfurl/generic/index.js
mmaxim: Oh I see

mmaxim: Hmm it's still a little tricky to make this work, since the gaps are different for some of the children in the current setup of boxes (either 4 or 8). Is it ok like this?

PR #14715, opened 2018-11-15
dead code

joshblum:
nukes some dead code in prep for protocol changes


mmaxim:
Are you sure this is dead?

PR #14725, opened 2018-11-16
Deprecate most fork calls, fix tracker disabling the app

chrisnojima:
historically we've avoided safeTakeLatest and its related calls due to saga's complex cancel flows. In reality we rarely want any kind of cancel logic to kill our saga tree.

This PR deprecates all safeTakeLatest* calls and switches all initializing forks to spawns. 
For .join() to work we do need forks so those can still access `Saga._fork` but this should be rare.

The motivation for this PR was from @malgorithms showing me a partially dead UI. Analyzing his logs showed some actions which had been canceled (surprising). 

When a tracker is visible we start a saga loop where we check every minute for tracker updates. If that rpc call happens to error out it'll throw and
1. kill the loop
1. kill the parent sagas all the way up

I protect that loop against this, and the tree of spawns should limit that kind of damage also


chrisnojima:
yeah we should ultimately move to actionTo*

PR #14726, opened 2018-11-16
Red & green usernames using users store

buoyad:
The tracker updates were already being piped through to the users store (I added the fullnames too). This adds `colorBroken` to `NameWithIcon` and makes ConnectedUsernames look into `state.users` instead of the tracker state. Remote code deserves a lot of scrutiny r? @keybase/react-hackers

### DIFF
--- a/shared/common-adapters/usernames/remote-container.js
+++ b/shared/common-adapters/usernames/remote-container.js
@@ -3,19 +3,19 @@ import * as ProfileGen from '../../actions/profile-gen'
 import {Usernames} from '.'
 import {remoteConnect, compose} from '../../util/container'
 import * as Container from './container'
-import * as TrackerTypes from '../../constants/types/tracker'
+import * as UsersTypes from '../../constants/types/users'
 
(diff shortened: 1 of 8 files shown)


buoyad:
@keybase/react-hackers bump

shared/common-adapters/usernames/remote-container.js
```diff
@@ -3,31 +3,41 @@ import * as ProfileGen from '../../actions/profile-gen'
...
 })
 
 const mapDispatchToProps = dispatch => ({
   _onUsernameClicked: (username: string) => dispatch(ProfileGen.createShowUserProfile({username})),
 })
 
+const userDatafromState = (stateProps, usernames) =>
+  usernames.map(username => ({
+    // $FlowIssue optional chain is needed
+    broken: stateProps._userInfo[username]?.broken || false,
```
chrisnojima: short term likely better to use the longer syntax and get rid this cause it'll mask real issues

PR #14729, opened 2018-11-16
make client respect more restrictive stellar account bundle interface from server

xgess:
client tests will otherwise break when https://github.com/keybase/keybase/pull/3130 gets merged

### DIFF
--- a/go/stellar/stellar.go
+++ b/go/stellar/stellar.go
@@ -228,6 +228,20 @@ func ImportSecretKeyAccountBundle(ctx context.Context, g *libkb.GlobalContext, s
 	if err != nil {
 		return err
 	}
+
+	// turn on the feature flag for the v2 stellar account bundles
+	m := libkb.NewMetaContext(ctx, g)
+	_, err = g.API.Post(libkb.APIArg{
+		Endpoint:    "test/feature",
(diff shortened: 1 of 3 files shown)


go/stellar/stellar.go
```diff
@@ -228,6 +228,20 @@ func ImportSecretKeyAccountBundle(ctx context.Context, g *libkb.GlobalContext, s
...
+	_, err = g.API.Post(libkb.APIArg{
+		Endpoint:    "test/feature",
+		SessionType: libkb.APISessionTypeREQUIRED,
+		Args: libkb.HTTPArgs{
+			"feature":   libkb.S{Val: string(libkb.FeatureStellarAcctBundles)},
+			"value":     libkb.I{Val: 1},
+			"cache_sec": libkb.I{Val: 100},
+		},
+		MetaContext: m,
+	})
```
mlsteele: probably don't want to run this in prod mode

xgess: agreed. line 213 (a few up) is this:
```go
 	if g.GetRunMode() == libkb.ProductionRunMode {
 		return errors.New("this doesn't work in production")
 	}
```

mlsteele: oh

PR #14732, opened 2018-11-16
Suspend inbox loading on sends

joshblum:
Suspend inbox loads when sending any messages as a potential optimization of the send path. It seems like the machinery was already in place, were there any other places we should be calling this?


go/chat/utils/utils.go
```diff
@@ -1660,3 +1660,16 @@ func SetUnfurl(mvalid *chat1.MessageUnboxedValid, unfurlMessageID chat1.MessageI
 	}
 	mvalid.Unfurls[unfurlMessageID] = unfurl
 }
+
+// SuspendComponent will suspend the global ConvLoader until the return
```
mmaxim: Suspends any `Suspendable`, not just `ConvLoader`

PR #14739, opened 2018-11-16
add back lib fast static lib

chrisnojima:
@keybase/react-hackers fixes ios crash, this was removed since it was in the recovered folder but i didn't realize its the only copy in the project file


MarcoPolo:
retro thumb 👍

chrisnojima:
Appreciated!

PR #14741, opened 2018-11-17
Add global capture keypress to plain-input

MarcoPolo:
@keybase/react-hackers

Is this the best way to make sure an input will receive all key presses? We do the same thing for chat, and I want this for search. (and likely we want this in other places where the input is the main focus)


akalin-keybase:
If this is the best way, perhaps it should be made into a HOC or an inheritable class or something so it can be reused instead of copied?

MarcoPolo:
That's why it's part of plain input. You only want this on inputs, so it
doesn't make sense to make it a hoc right? And we should be using plain
input moving forward

PR #14746, opened 2018-11-18
better read of libkb.ResolutionError

maxtaco:
- if you want to do an SBS in the case of a resolution failure, you can't assume that SBS allows follows.
- in most error cases, it does not
- in ResolutionErrorNotFound, it does...
- Fix it in the two places I found issues, but there could be more


maxtaco:
cc @malgorithms who found this bug

maxtaco:
it's right, i'll merge.

PR #14749, opened 2018-11-19
Add an explanation for the network reserve.

adamjspooner:
@keybase/react-hackers, this is ready for review. I started to add support for mobile tooltips (the non-centered kind) but reverted those changes and will add support for them later. In the meantime, we're opening a URL on mobile when they tap the question mark icon.


adamjspooner:
@keybase/react-hackers, bump.

shared/wallets/asset/index.js
```diff
@@ -90,58 +84,81 @@ type BalanceSummaryProps = {
...
-        </Text>
-        <Text type="BodySmall" selectable={true}>
-          {props.equivAvailableToSend}
-        </Text>
-      </Box2>
-    </Box2>
-  </Box2>
-)
+const BalanceSummary = (props: BalanceSummaryProps) => {
+  const _openStellarURL = () => {
```
chrisnojima: can we make this a prop thats made in the contianer instead. that can send null or this if its mobile and this 'dumb' component can just pass this.props.openStellarURL to the onClick

adamjspooner: Yup!

adamjspooner:
@chrisnojima, this is ready for another look.

PR #14750, opened 2018-11-19
Hide cancel button on in-flight relay for recipient

mlsteele:
Don't show recipients a cancel button for in-flight relays. Their app ought to claim it soon enough.


shared/wallets/transaction/index.js
```diff
@@ -383,9 +383,7 @@ export const Transaction = (props: Props) => {
             <Box2 direction="horizontal" fullWidth={true} style={styles.marginTopXTiny}>
               {props.onCancelPayment && (
                 <Box2 direction="vertical" gap="tiny">
-                  <Text type="BodySmall">
-                    {props.counterparty} can claim this when they set up their wallet.
-                  </Text>
+                  <Text type="BodySmall">{props.statusDetail}</Text>
```
buoyad: May want to pipe this through to [here](https://github.com/keybase/client/blob/master/shared/chat/conversation/messages/account-payment/container.js#L80) too

go/stellar/transform.go
```diff
@@ -200,16 +222,19 @@ func transformPaymentRelay(mctx libkb.MetaContext, acctID stellar1.AccountID, p
...
-		// If the funding tx is not complete
+		// The funding tx is not complete.
 		loc.StatusSimplified = p.TxStatus.ToPaymentStatus()
 		loc.StatusDetail = p.TxErrMsg
 	} else {
 		loc.StatusSimplified = stellar1.PaymentStatus_CLAIMABLE
-		loc.StatusDetail = "Waiting for the recipient to open the app to claim, or the sender to cancel."
-		loc.ShowCancel = true
+		if isSender {
+			loc.StatusDetail = fmt.Sprintf("%v can claim this when they set up their wallet.", toName)
```
mlsteele: This "can claim" value is StatusDetail is not used. Because it can get overridden by an error message from a failed claim. If `PaymentLocal` had another string field for this then the frontend could lean on the service for this message. But I don't want to make `PaymentLocal` any bigger unless someone thinks it's worth it.

shared/chat/conversation/messages/account-payment/container.js
```diff
@@ -77,8 +79,8 @@ const mapStateToProps = (state, ownProps: OwnProps) => {
           paymentInfo.amountDescription
         )}`,
         balanceChangeColor: WalletConstants.balanceChangeColor(paymentInfo.delta, paymentInfo.status),
-        cancelButtonInfo: youAreSender && cancelable ? makeCancelButtonInfo(theirUsername) : '',
-        cancelButtonLabel: youAreSender && cancelable ? 'Cancel' : '',
+        cancelButtonInfo: paymentInfo.showCancel ? makeCancelButtonInfo(theirUsername) : '',
```
akalin-keybase: there are a couple of other places where we condition on 'cancelable', e.g. shared/chat/conversation/messages/message-popup/payment/container.js . Do we still need 'cancelable' to be a possibility in `statusSimplified`? Can't this new `showCancel` field replace that?

mlsteele: I think I see what you're saying about message-popup. Not sure how to get message-popup to show up in this state.

akalin-keybase: per our chat, I think the case in payment/container.js is fine as-is (or will be once the status is renamed to be `claimable`) since it's a popup for the sender

akalin-keybase:
per our chat, I think the only issue is renaming `cancelable` back to `claimable` so as not to be confused with this new `showCancel` flag :+1:

buoyad:
For context, `claimable` was originally displayed everywhere we showed the status, such as here on the tx details screen:
image
someone commented that `cancelable` would be less confusing for users, hence the switch in the transform

akalin-keybase:
Ah...perhaps that transform should be done in the tx details screen then?

go/stellar/transform.go
```diff
@@ -105,11 +105,22 @@ func TransformRequestDetails(mctx libkb.MetaContext, details stellar1.RequestDet
 
 // transformPaymentStellar converts a stellar1.PaymentSummaryStellar into a stellar1.PaymentLocal.
 func transformPaymentStellar(mctx libkb.MetaContext, acctID stellar1.AccountID, p stellar1.PaymentSummaryStellar, oc OwnAccountLookupCache, exchRate *stellar1.OutsideExchangeRate) (*stellar1.PaymentLocal, error) {
-	loc, err := newPaymentLocal(mctx, p.TxID, p.Ctime, p.Amount, p.Asset, p.From, p.To, acctID, exchRate)
+	loc, err := newPaymentLocal(mctx, p.TxID, p.Ctime, p.Amount, p.Asset, exchRate)
 	if err != nil {
 		return nil, err
 	}
 
+	isSender := p.From.Eq(acctID)
```
akalin-keybase: this stuff seems new...is it supposed to be part of this change?

shared/chat/conversation/messages/account-payment/container.js
```diff
@@ -37,8 +37,10 @@ const makeSendPaymentVerb = (status: WalletTypes.StatusSimplified, youAreSender:
     case 'pending':
       return 'sending'
     case 'canceled': // fallthrough
-    case 'cancelable':
+    case 'claimable':
       return youAreSender ? 'sending' : 'attempting to send'
+    case 'error':
+      return youAreSender ? 'attempted to send' : 'attempted to send'
```
akalin-keybase: both cases are the same, so just `return 'attempted to send'`?

go/stellar/transform.go
patrickxb: Looking

shared/constants/chat2/message.js
```diff
@@ -1080,7 +1082,7 @@ export const shouldShowPopup = (state: TypedState, message: Types.Message) => {
     case 'sendPayment': {
       // Is the payment pending?
       const paymentInfo = getPaymentMessageInfo(state, message)
-      if (!paymentInfo || ['cancelable', 'pending', 'canceled'].includes(paymentInfo.get('status'))) {
+      if (!paymentInfo || ['claimable', 'pending', 'canceled'].includes(paymentInfo.get('status'))) {
```
patrickxb: This `if` statement doesn't seem to match the "Is the payment pending?" comment.

shared/chat/conversation/messages/account-payment/container.js
```diff
@@ -77,8 +79,8 @@ const mapStateToProps = (state, ownProps: OwnProps) => {
           paymentInfo.amountDescription
         )}`,
         balanceChangeColor: WalletConstants.balanceChangeColor(paymentInfo.delta, paymentInfo.status),
-        cancelButtonInfo: youAreSender && cancelable ? makeCancelButtonInfo(theirUsername) : '',
-        cancelButtonLabel: youAreSender && cancelable ? 'Cancel' : '',
+        cancelButtonInfo: paymentInfo.showCancel ? makeCancelButtonInfo(theirUsername) : '',
+        cancelButtonLabel: paymentInfo.showCancel ? 'Cancel' : '',
```
patrickxb: I'm just guessing here, but can't the cancelButtonLabel always be "Cancel" and the `showCancel` determines if it is shown or not?

PR #14754, opened 2018-11-19
Improve ephemeral error logging

joshblum:
once this is in should help pinpoint the root cause of your error @mmaxim

### DIFF
--- a/go/ephemeral/device_ek_storage.go
+++ b/go/ephemeral/device_ek_storage.go
@@ -133,6 +133,7 @@ func (s *DeviceEKStorage) Put(ctx context.Context, generation keybase1.EkGenerat
 }
 
 func (s *DeviceEKStorage) Get(ctx context.Context, generation keybase1.EkGeneration) (deviceEK keybase1.DeviceEk, err error) {
+	defer s.G().CTraceTimed(ctx, fmt.Sprintf("DeviceEKStorage#Get: generation:%v", generation), func() error { return err })()
 	s.Lock()
 	defer s.Unlock()
 


go/ephemeral/user_ek_box_storage.go
```diff
@@ -109,7 +109,8 @@ func (s *UserEKBoxStorage) Get(ctx context.Context, generation keybase1.EkGenera
 
 	defer s.Unlock() // release the lock after we unbox
 	if cacheItem.HasError() {
-		return userEK, cacheItem.Error()
+		err = cacheItem.Error()
```
mmaxim: What is going on here?

go/ephemeral/user_ek_box_storage.go
```diff
@@ -230,7 +231,8 @@ func (s *UserEKBoxStorage) unbox(ctx context.Context, userEKGeneration keybase1.
 		s.G().Log.CDebugf(ctx, "unable to get from deviceEKStorage %v", err)
 		switch err.(type) {
 		case erasablekv.UnboxError:
-			return userEK, newEKUnboxErr(UserEKStr, userEKGeneration, DeviceEKStr, userEKBoxed.DeviceEkGeneration)
+			err = newEKUnboxErr(UserEKStr, userEKGeneration, DeviceEKStr, userEKBoxed.DeviceEkGeneration)
```
mmaxim: And here?

go/ephemeral/user_ek_box_storage.go
joshblum: this will allow `err` to hold the correct value for logs in the `Trace`

mmaxim: If you return it, `err` gets set, so I don't think you need to do this.

PR #14756, opened 2018-11-19
chat1/gregor1 CallCompressed 

joshblum:
compress all `chat1/remote.avdl` calls with gzip, depends on https://github.com/keybase/client/pull/14724, will rebase once that is in


joshblum:
did a pass for any endpoints that might not be suitable for compression and nixed two that sent a `gregor1.SessionToken` i don't think an attacker has a direct influence on the payload but figured it would be good to drop for safety. Are there any other protocols that would benefit from compression we should add?

PR #14758, opened 2018-11-19
fix publish date for unfurls

mmaxim:


go/chat/unfurl/scrape_generic_scoring.go
```diff
@@ -169,6 +169,26 @@ var attrRankMap = map[string]attrRanker{
...
+		setter:  setPublishTime,
+	},
+	"article.published": attrRanker{
+		content: getContentAttr,
+		score:   getOpenGraphScore,
+		setter:  setPublishTime,
+	},
+	"article:published": attrRanker{
+		content: getContentAttr,
+		score:   getOpenGraphScore,
```
joshblum: `getDefaultScore`

go/chat/unfurl/scrape_generic_scoring.go
```diff
@@ -169,6 +169,26 @@ var attrRankMap = map[string]attrRanker{
...
+		setter:  setPublishTime,
+	},
+	"datePublished": attrRanker{
+		content: getContentAttr,
+		score:   getOpenGraphScore,
```
joshblum: `getDefaultScore`

go/chat/unfurl/scrape_generic_scoring.go
mmaxim: I think these should be equal weight as OG (or at least higher than default). Would you prefer having something in between OG and default, or just have these the same?

PR #14760, opened 2018-11-19
Parse ICS tx

heronhaye:


go/protocol/chat1/notify.go
```diff
@@ -106,8 +106,8 @@ type IncomingMessage struct {
 
 func (o IncomingMessage) DeepCopy() IncomingMessage {
 	return IncomingMessage{
-		Message: o.Message.DeepCopy(),
-		ConvID:  o.ConvID.DeepCopy(),
+		Message:                    o.Message.DeepCopy(),
```
heronhaye: not sure why this happened, is it an outdated compiler?

go/chat/wallet/parser.go
```diff
@@ -0,0 +1,112 @@
...
+var newline = []rune("\n")
+
+type newlineBehavior = int
+
+const (
+	sameLine newlineBehavior = iota
+	multiLine
+)
+
+func replaceQuotedSubstrings(xs string, replacer string) string {
```
mmaxim: Seems like you can just do what I did for unfurling here: https://github.com/keybase/client/blob/7a816baeb2b6616a23f7d3e8345807dc5cafd021/go/chat/unfurl/extractor.go#L62

heronhaye: huh, that makes a lot of sense. For some reason I thought this wouldn't work but since there's no arbitrary nesting I guess regex is fine. One comment: this incorrectly parses

```
```hello`world``` 
```
to
```
world`
```
instead of empty string. What do you think about
```
    xs = regexp.MustCompile("(?s)```.*?```").ReplaceAllString(xs, replacer)
    xs = regexp.MustCompile("(?s)`.*?`").ReplaceAllString(xs, replacer)
```?

Comment (mmaxim) [reply go/chat/wallet/parser.go]: Sure sounds good.

File: protocol/avdl/chat1/wallet.avdl [hunk 3]
```diff
@@ -0,0 +1,8 @@
+@namespace("chat.1") protocol wallet {
+
```
mmaxim: I'm not sure this type needs to be here, how come you put it in AVDL?

heronhaye: thought the server might need it, I can take it out though

mmaxim: Yeah don't think so, should just be turned into a normal payment after this step.

mmaxim: In fact it is probably worthwhile to just create whatever type the Stellar send code expects for what would have been generated from the CLI.

mmaxim: Can make another one

PR #14773, opened 2018-11-20
more flow 86

chrisnojima:
@keybase/react-hackers last bit of non-fs related flow stuff
cc: @jzila


shared/profile/post-proof/container.js
```diff
@@ -42,8 +42,7 @@ const mapDispatchToProps = dispatch => ({
   copyToClipboard: text => dispatch(ConfigGen.createCopyToClipboard({text})),
 })
 
-export default compose(
-  // FlowIssue on 86
+export default (compose(
```
buoyad: some of these exports are becoming `any`, do we have a workaround for these cases?

chrisnojima: not now

PR #14786, opened 2018-11-20
store unfurl prompts in a map in the store instead of on messages

mmaxim:
@keybase/react-hackers 

Having the prompts on `Types.Message` causes problems since the prompts are not represented in the service at all. If we replace messages with messages from the service, the prompts get blown away. Putting them into this new map has two nice effects:

1.) Prompts don't get cleared anymore when a message is changed.
2.) Prompts survive switching threads.


shared/reducers/chat2.js
```diff
@@ -429,6 +408,15 @@ const rootReducer = (
 
         s.set('selectedConversation', action.payload.conversationIDKey)
       })
+    case Chat2Gen.unfurlTogglePrompt:
+      return state.updateIn(
+        ['unfurlPromptMap', action.payload.conversationIDKey, action.payload.messageID],
+        prompts => {
```
chrisnojima: i believe this can be
```
const {show, domain} = action.payload
return state.updateIn(['unfurlPromptMap', action.payload.conversationIDKey, action.payload.messageID], I.Set(), prompts => show ? prompts.add(domain) : prompts.delete(domain)
```

mmaxim: Ah yeah, will change.

shared/reducers/chat2.js
```diff
@@ -409,12 +409,12 @@ const rootReducer = (
         s.set('selectedConversation', action.payload.conversationIDKey)
       })
     case Chat2Gen.unfurlTogglePrompt:
+      const {show, domain} = action.payload
       return state.updateIn(
         ['unfurlPromptMap', action.payload.conversationIDKey, action.payload.messageID],
+        I.Set(),
         prompts => {
```
chrisnojima: more concision by doing shorthand `prompts => show ? ...`

PR #14790, opened 2018-11-21
Tweaks on Stellar onboarding

cecileboucheron:
@keybase/react-hackers


shared/wallets/onboarding/disclaimer.js
```diff
@@ -72,25 +73,26 @@ class Disclaimer extends React.Component<DisclaimerProps, DisclaimerState> {
...
+          </Kb.Text>
+        </Kb.Box2>
 
         <Kb.ScrollView style={styles.scrollView} contentContainerStyle={styles.scrollViewContentContainer}>
           <Kb.Text type="Body" style={styles.bodyText}>
             We believe Keybase can help make cryptocurrency usable for 2 reasons:
           </Kb.Text>
 
-          <Kb.Text type="Body" style={styles.bodyText}>
+          <Kb.Text type="Body" style={{...styles.bodyText, ...styles.bodyBullet}}>
```
adamjspooner: We should use `Styles.collapseStyles` instead of spreading.

shared/wallets/wallet-list/index.js
```diff
@@ -93,10 +93,8 @@ class _WalletList extends React.Component<Props> {
           />
         )
       default:
-        /*::
-      declare var ifFlowErrorsHereItsCauseYouDidntHandleAllTypesAbove: (a: empty) => any
-      ifFlowErrorsHereItsCauseYouDidntHandleAllTypesAbove(row.type);
-      */
+        declare var ifFlowErrorsHereItsCauseYouDidntHandleAllTypesAbove: (a: empty) => any
+        ifFlowErrorsHereItsCauseYouDidntHandleAllTypesAbove(row.type)
```
adamjspooner: Did you mean to uncomment this?

cecileboucheron: Ugh. my prettier plugin is messed up

cecileboucheron:
@adamjspooner bump ^

shared/wallets/onboarding/disclaimer.js
```diff
@@ -51,6 +51,7 @@ class Disclaimer extends React.Component<DisclaimerProps, DisclaimerState> {
         type="Secondary"
         onClick={this.props.onAcceptDisclaimer}
         label={label}
+        labelStyle={{color: Styles.globalColors.purple}}
```
adamjspooner: We should put all styles down in the `styleSheetCreate` block.

cecileboucheron: @adamjspooner It looks like this is the only way to change the label color? See same prop used on the Open wallet button: https://github.com/keybase/client/pull/14790/files#diff-e684b085cf794d0fa3273fb86ee9f568R16

adamjspooner: You should be able to create a new object down in the `styleSheetCreate` block and reference it here. Something like:

```
label: {
    color: Styles.globalColors.purple,
},
```

And then `labelStyle={styles.label}` here. Does that make sense?

cecileboucheron: got it. Updated.

PR #14795, opened 2018-11-21
unfurling: RPCs for setting and getting settings intended for UI CORE-9399

mmaxim:
Also fixes a bug in `TestChatSrvUnfurl` intro'd by the favicon changes.


go/chat/server.go
```diff
@@ -2346,3 +2346,41 @@ func (h *Server) ResolveUnfurlPrompt(ctx context.Context, arg chat1.ResolveUnfur
...
+	uid, err := utils.AssertLoggedInUID(ctx, h.G())
+	if err != nil {
+		return res, err
+	}
+	settings, err := h.G().Unfurler.GetSettings(ctx, uid)
+	if err != nil {
+		return res, err
+	}
+	res.Mode = settings.Mode
+	for w := range settings.Whitelist {
```
joshblum: why not `copy`?

mmaxim: Source is a `map`

mmaxim:
Merging over CI

go/chat/server.go
joshblum: ah makes sense

PR #14797, opened 2018-11-21
Use a spinner on team creation, move to a WaitingKey

cjb:
@keybase/react-hackers

### DIFF
--- a/shared/actions/teams.js
+++ b/shared/actions/teams.js
@@ -435,12 +435,11 @@ const _createNewTeamFromConversation = function*(
 
   if (participants) {
     yield Saga.put(TeamsGen.createSetTeamCreationError({error: ''}))
-    yield Saga.put(WaitingGen.createIncrementWaiting({key: Constants.teamCreationWaitingKey}))
     try {
       const createRes = yield Saga.call(RPCTypes.teamsTeamCreateRpcPromise, {
         joinSubteam: false,
         name: teamname,
-      })
+      }, Constants.teamCreationWaitingKey)
       for (const username of participants) {
         if (!createRes.creatorAdded || username !== me) {
           yield Saga.call(RPCTypes.teamsTeamAddMemberRpcPromise, {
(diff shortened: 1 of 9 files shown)


cjb:
@chrisnojima Thanks, done!

@zanderz Sure, sounds good.  (Though, oftentimes we're using createIncrementWaiting because there's more than one RPC to wait on, or some other kind of chain of events that we want to hang out until the end of.)

cjb:
(I suppose given the above, it makes more conceptual sense to tie *sagas* to waiting keys rather than RPCs.  Tying RPCs only works well if there's one of them and there's nothing else to do afterwards before returning to the user.  If we just tie the RPC, then the user's spinner might return before the app is ready for them.)

chrisnojima:
I'd argue the waiting key should be tied to a concept. Something reflected in the ui. I'd multiple rpcs are required to fulfill it they can both wait on the same key (it's a ref count)

cjb:
@chrisnojima bump, made the waitingkey change

shared/actions/teams.js
```diff
@@ -438,7 +435,7 @@ const _createNewTeamFromConversation = function*(
 
   if (participants) {
     yield Saga.put(TeamsGen.createSetTeamCreationError({error: ''}))
-    yield Saga.put(TeamsGen.createSetTeamCreationPending({pending: true}))
+    yield Saga.put(WaitingGen.createIncrementWaiting({key: Constants.teamCreationWaitingKey}))
```
chrisnojima: why not here also

chrisnojima: did you actually try it with the same key to those other calls, we can easily make a grace period for the spinners to not stop immediately

cjb: @chrisnojima I didn't notice a visual glitch, though the render function does get called with waiting=false inbetween teamCreate and teamAdd after your change.  (Maybe the return from render() only makes it to a frame on the screen on slower hardware?)

Adding a grace period doesn't feel great to me -- feels like we're just trying to emulate what the code is already doing, but in a way that doesn't have any visibility into what the user's actually waiting on, we could chat more about what we think is the right granularity to spin on.

chrisnojima: i think if it spins, then stops for a couple of frames (doubtful) and spins some more thats fine. we have plenty of worse situations in the app (no spinners at all) so i think this is a case we can focus on way down the line

cjb: @chrisnojima Okay, done.

PR #14804, opened 2018-11-21
unfurling: settings screen for setting unfurling prompt settings CORE-9486

mmaxim:
@keybase/react-hackers 

Adds a new settings screen for desktop and mobile to control the chat unfurl prompt settings.


shared/reducers/settings.js
```diff
@@ -202,6 +202,31 @@ function reducer(state: Types.State = Constants.initialState, action: SettingsGe
...
+          ...state.chat,
+          unfurl: {
+            unfurlMode: mode,
+            unfurlWhitelist: whitelist,
+            unfurlError: undefined,
+          },
+        },
+      }
+    case SettingsGen.unfurlSettingsError:
+      return {
```
cjb: Ah, I see the settings store isn't using immutable -- if it was, it'd be cool to be able to use state.merge() to avoid all these `...`s.

shared/settings/chat/index.js
```diff
@@ -0,0 +1,198 @@
...
+      width: '100%',
+    },
+  }),
+  whitelistRowContainer: {
+    flexShrink: 0,
+    justifyContent: 'space-between',
+    paddingTop: 3,
+    paddingBottom: 4,
+    paddingLeft: 9,
+    paddingRight: 8,
```
cjb: In general, it feels bad to have magic number pixel offsets like these -- it'd be nicer to use `Styles.globalMargins.` values instead, so that any house style design changes in the future affect this screen too.

shared/constants/settings.js
```diff
@@ -35,6 +35,9 @@ const initialState: Types.State = {
     newPassphraseError: null,
     rememberPassphrase: true,
   },
+  chat: {
```
cjb: Looks like `lockdownModeEnabled` already messed this up, but object keys are supposed to be alphabetical.  (eslint warns about it but doesn't fix it.)

cjb: Looks like `lockdownModeEnabled` already messed this up, but object keys are supposed to be alphabetical.  (eslint warns but doesn't auto-fix)

shared/settings/chat/index.stories.js
```diff
@@ -0,0 +1,48 @@
...
+
+const actions = {
+  onUnfurlSave: (mode: RPCChatTypes.UnfurlMode, whitelist: Array<string>) => {
+    Sb.action('onUnfurlSave')(mode, whitelist)
+  },
+  onRefresh: Sb.action('onRefresh'),
+}
+
+const props = {
+  unfurlMode: RPCChatTypes.unfurlUnfurlMode.whitelisted,
```
cjb: I think we tend to find it unsatisfying when a display component's reaching all the way down into RPC definitions -- we sometimes duplicate the RPC shape inside `constants/settings` directly, or even re-export the RPC value directly from there.

shared/actions/settings.js
```diff
@@ -421,6 +421,36 @@ const setLockdownMode = (state: TypedState, action: SettingsGen.OnChangeLockdown
       return SettingsGen.createLoadLockdownMode()
     })
 
+const unfurlSettingsRefresh = (state: TypedState, action: SettingsGen.UnfurlSettingsRefreshPayload) =>
+  state.config.loggedIn &&
+  ChatTypes.localGetUnfurlSettingsRpcPromise(undefined, Constants.waitingKey)
```
cjb: Totally not your fault, but having a waitingKey per-RPC makes more sense: switching to `Constants.unfurlWaitingKey` for these would be much better.

shared/actions/settings.js
```diff
@@ -421,6 +421,36 @@ const setLockdownMode = (state: TypedState, action: SettingsGen.OnChangeLockdown
...
+    .then((result: ChatTypes.UnfurlSettingsDisplay) => {
+      return SettingsGen.createUnfurlSettingsRefreshed({mode: result.mode, whitelist: result.whitelist || []})
+    })
+    .catch(() => {
```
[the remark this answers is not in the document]
mmaxim: You guys really don’t like return :)

shared/constants/types/settings.js
```diff
@@ -74,6 +75,15 @@ export type EmailState = {
   error: ?Error,
 }
 
+export type ChatUnfurlState = {
+  unfurlMode?: RPCChatTypes.UnfurlMode,
```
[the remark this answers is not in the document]
mmaxim: I had trouble using ! since the mode value can be 0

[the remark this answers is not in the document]
mmaxim: I had trouble using ! since the mode value can be 0

shared/reducers/settings.js
mmaxim: Yeah, I just mimicked what other reducers were doing here. I can see the advantage of using something like Immutable, if the shape of this store was more complicated it seems easy to mess this up.

shared/settings/chat/index.stories.js
mmaxim: Yeah I thought this type was simple enough that it made sense here. I tried to make the unfurling types that come up to the UI as simple and usable as possible so it could just use them directly.

shared/settings/chat/index.js
```diff
@@ -0,0 +1,198 @@
...
+            key="rbnever"
+            label="Never"
+            onSelect={() => this._setUnfurlMode(RPCChatTypes.unfurlUnfurlMode.never)}
+            selected={this._getUnfurlMode() === RPCChatTypes.unfurlUnfurlMode.never}
+            disabled={this.props.unfurlMode === undefined}
+          />
+        </Kb.Box2>
+        <Kb.Divider style={styles.divider} />
+        <Kb.Box2 direction="vertical" gap="tiny">
+          <Kb.WaitingButton
```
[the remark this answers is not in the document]
mmaxim: The design has a save button, and I think the flow makes sense here. I can switch it back to disabled if you go back to the same option though.

shared/settings/chat/index.js
mmaxim: Yeah agreed, sometimes Zeplin just has some random px value in there so I am not sure how to best represent here. Will change these though.

mmaxim:
Thanks @cjb!

PR #14805, opened 2018-11-21
Surya/core 9539/rm log filename

heronhaye:


maxtaco:
@heronhaye are you sure that systemd doesn't go through "auto-fork"?  In the old linux days, when you first ran the CLI, it spawned the daemon in the background if it didn't already exist. Then we passed through the logfile, I believe...

heronhaye:
hey @taruti, we were trying to get rid of the `--log-file-name` in the systemd unit file and found that we needed to call `RotateLog` after setting the default logfile for logging to work. Does this look okay?

go/libkb/env.go
```diff
@@ -1250,6 +1251,21 @@ func (e *Env) GetLogFile() string {
...
 }
 
+func (e *Env) GetUseDefaultLogFile() bool {
+	return e.GetBool(false,
+		e.cmd.GetUseDefaultLogFile,
+		func() (bool, bool) { return e.getEnvBool("KEYBASE_USE_DEFAULT_LOG_FILE") },
+	)
+}
+
+// func (e *Env) GetDisplayRawUntrustedOutput() bool {
```
maxtaco: what's this?

eli-schwartz:
Thanks!

PR #14819, opened 2018-11-25
unfurling: add support for the close button CORE-9514

mmaxim:
**Go**
1.) Change `Storage.updateAllSupersededBy` to return a list of messages affected by unfurls, similar to how reactions work.
2.) Hookup logic to process `DELETE` messages of unfurls.
3.) Change `notifyUnfurls` to operate on the targets of unfurls, and not the unfurl messages themselves.

**JS**
@keybase/react-hackers 
1.) Get the close button displayable in the generic display on desktop and mobile. 
2.) New action to invoke a message delete on the unfurl message when clicking the close button.


go/chat/storage/storage.go
```diff
@@ -1033,6 +1057,28 @@ func (s *Storage) getMessage(ctx context.Context, convID chat1.ConversationID, u
 	return &res[0], nil
 }
 
+func (s *Storage) updateUnfurlTargetOnDelete(ctx context.Context, convID chat1.ConversationID,
+	uid gregor1.UID, unfurlMsg chat1.MessageUnboxed) (res chat1.MessageUnboxed, err error) {
+	defer s.Trace(ctx, func() error { return err }, "updateUnfurlTargetOnDelete(%d)",
```
joshblum:
```suggestion
	defer s.Trace(ctx, func() error { return err }, "updateUnfurlTargetOnDelete unfurlMsg: %v",
```

go/chat/storage/storage.go
```diff
@@ -1033,6 +1057,28 @@ func (s *Storage) getMessage(ctx context.Context, convID chat1.ConversationID, u
...
+		unfurlMsg.GetMessageID())()
```
joshblum:
```suggestion
		unfurlMsg)()
```

shared/chat/conversation/messages/unfurl/generic/index.js
```diff
@@ -16,28 +16,51 @@ export type Props = {
   showImageOnSide: boolean,
 }
 
-class UnfurlGeneric extends React.PureComponent<Props> {
+type State = {
+  showClose: boolean,
+}
+
+class UnfurlGeneric extends React.Component<Props, State> {
+  state = {showClose: false}
```
chrisnojima: would prefer using css for this, esp. since this is effectively desktop only. lets discuss

mmaxim: Done!

PR #14822, opened 2018-11-25
unfurling: add a one-time exemption option for unfurl prompt CORE-9541

mmaxim:
Patch does the following:

1.) Adds a notion of "whitelist exemptions" to `Extractor` in order for domains to get unfurled without being explicitly on the whitelist. Implement `OneTimeWhitelistExemption` to provide the policy of unfurling for a domain a single time.
2.) Hook up the new option on the `Prompt` component. cc @keybase/react-hackers


go/chat/unfurl/exemption.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"sync"
+
+	"github.com/keybase/client/go/chat/types"
+)
+
+type OneTimeWhitelistExemption struct {
+	sync.Mutex
+	used   bool
```
joshblum: should this hold convID/msgID for safety?

go/chat/unfurl/exemption.go
```diff
@@ -0,0 +1,69 @@
...
+	var nextlist []types.WhitelistExemption
+	exempted := false
+	for _, e := range l.exemptions {
+		if exempted || e.Domain() != domain {
+			nextlist = append(nextlist, e)
+			continue
+		}
+		if e.Use() {
+			exempted = true
+			nextlist = append(nextlist, e)
```
joshblum: why append in this case?

go/chat/unfurl/exemption.go
mmaxim: Don't think so, can only get used once.

go/chat/unfurl/exemption.go
mmaxim: It could keep returning true, like say an exception that returns true for the next hour.

go/chat/unfurl/exemption.go
joshblum: right but anyone can grab it, not necessarily the msg that got the exemption

go/chat/unfurl/exemption.go
joshblum: ah gotcha

go/chat/unfurl/exemption.go
mmaxim: Yeah I suppose that is true, let me see what I can do.

PR #14830, opened 2018-11-26
fix for orange line below on mobile. fix for sending icon off the edge

chrisnojima:
@keybase/react-hackers fixes orange line being below on mobile and fixes send icon too far to the right

### DIFF
--- a/shared/chat/conversation/messages/wrapper/wrapper-author/index.js
+++ b/shared/chat/conversation/messages/wrapper/wrapper-author/index.js
@@ -259,7 +259,7 @@ const styles = Styles.styleSheetCreate({
       top: -1,
     },
     isMobile: {
-      right: -14,
+      right: 0,
       top: 2,
     },
   }),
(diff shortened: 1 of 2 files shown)


shared/chat/conversation/messages/wrapper/wrapper-author/index.js
```diff
@@ -259,7 +259,7 @@ const styles = Styles.styleSheetCreate({
       top: -1,
     },
     isMobile: {
-      right: -14,
```
chrisnojima: unclear where this -14 even came from

shared/chat/conversation/messages/wrapper/wrapper-message/index.js
```diff
@@ -141,13 +141,13 @@ class _WrapperMessage extends React.Component<Props & Kb.OverlayParentProps, Sta
 
     return (
       <React.Fragment>
```
chrisnojima: turns out making this a fragment causes the parent container (made by RN) to lay it out (which is normally fine) but it turns out this uses flexDirection: column-reverse for some reason so it flips the relationship!

PR #14831, opened 2018-11-26
dont scroll around as selected changes

chrisnojima:
@keybase/react-hackers


chrisnojima:
unclear. this code is old and hopefully going to be going away soon. the idea that we change selection as you hover over is fine (although it seems like any styling linked to this no longer works on desktop), but having the list scroll around as you mouse around seems like not a good idea. originally i though it'd make sense for keyboard nav but that also seems to not work in search so removing this seems like it changes nothing

PR #14832, opened 2018-11-26
Handle unicode in urls

chrisnojima:
- [x] allow international chars in query strings (https://www.google.com/search?q=ebаy) (the a is actually unicode)
- [x] disallow international chars in the hostname (http://ebаy.com) (the a is actually unicode)
- [x] stricter boundary between urls and other characters (the original implementation made the above link into `y.com` with `ebа` directly in front of this. We likely don't want this behavior so we now look at the lookbehind and throw out the match if its not punctuation


chrisnojima:
@keybase/react-hackers cc: @MarcoPolo

PR #14833, opened 2018-11-26
squash chanmentions from non admin on large teams

joshblum:


go/chat/push.go
```diff
@@ -311,6 +313,45 @@ func (g *PushHandler) TlfResolve(ctx context.Context, m gregor.OutOfBandMessage)
...
+		conv.Info.MembersType != chat1.ConversationMembersType_TEAM ||
+		len(conv.Info.Participants) < chat1.MaxChanMentionConvSize {
+		return false, nil
+	}
+	teamID, err := keybase1.TeamIDFromString(mvalid.ClientHeader.Conv.Tlfid.String())
+	if err != nil {
+		return false, err
+	}
+	extG := g.G().ExternalG()
+	team, err := teams.Load(ctx, extG, keybase1.LoadTeamArg{
```
mmaxim: I wonder if all this code should just be in a helper function in the `teams` package.

joshblum: most of it is in the team package, unless we made a chat specific helper (took a `tlfid` string in) we'd only save the upak load to the get sender's UV, I don't think it's worth it

PR #14835, opened 2018-11-27
Team Building 3. With recs.

MarcoPolo:
@keybase/react-hackers 

Fetches recommendations for the new team building and matches the design a bit more.

### DIFF
--- a/shared/actions/json/team-building.json
+++ b/shared/actions/json/team-building.json
@@ -4,7 +4,7 @@
     "import HiddenString from '../util/hidden-string'"
   ],
   "actions": {
-    "userRecsLoaded": {users: "Array<Types.User>"},
+    "fetchedUserRecs": {users: "Array<Types.User>"},
     "fetchUserRecs": {},
     "finishedTeamBuilding": {},
     "cancelTeamBuilding": {},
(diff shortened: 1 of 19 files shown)


shared/actions/json/team-building.json
```diff
@@ -4,6 +4,8 @@
     "import HiddenString from '../util/hidden-string'"
   ],
   "actions": {
+    "userRecsLoaded": {users: "Array<Types.User>"},
```
chrisnojima: maybe better to be `fetecedUserRecs` so the get/set names are closer 🐑

shared/styles/colors.js
```diff
@@ -80,6 +80,17 @@ const colors = {
   yellowGreen2_75: 'rgba(154, 180, 57, 0.75)',
   yellowGreen2_75_on_white: 'rgb(179,199,107)',
   yellowGreen3: '#d2e697',
+
+  // Service Accent Colors
+  // These are custom per service so they may not be associated with the keybase color scheme
```
chrisnojima: soon we'll be doing the generic proofs so services will come down as an rpc param so this stuff will likely be moved there also. given that we might want to remove it from here cause it won't be a system color

MarcoPolo: The reason I out them here is because it's needed for the icon css color names to work properly. (Esp hover, which won't work without them here).

I can move then out again, and we can revisit later. I don't think we need the css names

PR #14839, opened 2018-11-27
Fix message failing to send if you type "#foo<enter>" or "@foo<enter>"

cjb:
@keybase/react-hackers 

When you type `#foo<enter>` or `@foo<enter>` into a chat on desktop, nothing happens -- we should send the message but we don't.  The reason nothing happens is related to this comment:
```js
    if (this.state.mentionPopupOpen || this.state.channelMentionPopupOpen) {
      // On desktop, this is triggered on Enter, so if a mention
      // popup is open we actually just want to pick whatever's
      // selected.
      this._triggerPickSelectedCounter('Enter')
      return
    }
```
(The `return` prevents us from performing the message send.)

The comment makes sense -- we should suppress the message-send behavior of pressing enter if the mention hud's open, because the mention hud is going to act on the keypress instead and choose the selected hud item.  But the code doesn't work, because the mention popup is "open" even when it's not actually displaying anything.  That's because of this other comment:
```js
// We want to render Hud even if there's no data so we can still have lifecycle methods so we can still do things
// This is important if you type a filter that gives you no results and you press enter for instance
const Hud = ({style, data, loading, rowRenderer, selectedIndex}: HudProps<any>) =>
```

So any time your cursor's inside a word that starts with @ or #, the mention hud is considered open by the existing code.  To fix the logic, we need to consider not just whether the hud's mounted, but whether it's actually displaying results.  The simplest way I could think of to do this is to add a new `this.state.mentionHudIsShowing` test that the hud itself updates when it gets new rows, and that's what this PR does.

### DIFF
--- a/shared/chat/conversation/input-area/channel-mention-hud/index.js
+++ b/shared/chat/conversation/input-area/channel-mention-hud/index.js
@@ -70,6 +70,8 @@ const MentionHud = compose(
   }),
   lifecycle({
     componentDidUpdate(prevProps, prevState) {
+      this.props.setChannelMentionHudIsShowing(this.props.data.length > 0)
+
       if (this.props.data.length === 0) {
         if (prevProps.selectedIndex === 0) {
           // We've already done this, so just get out of here so we don't infinite loop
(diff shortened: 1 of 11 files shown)


cjb:
Agreed, super backwards.. I'll merge this but spend a little time trying to work out what inverting it would look like, that would be a nice improvement even if the HOCs are still mostly there.

chrisnojima:
i think we need to take a step back and look at what this component is. it doesn't need any hocs at all if if there is one its one that manages arrow keys in a list (ala downshift).

PR #14849, opened 2018-11-27
Use Inbox to back TeamChannelSource

joshblum:
Changes `CachingTeamChannelSource` to use the inbox to gather all conversations in a TLF to detect channel mentions in chat. depends on https://github.com/keybase/keybase/pull/3171 for server changes to send down all convs in the inbox. After a few client release cycles we can remove the server support `types.PushTeamChannels`


go/chat/search/indexer.go
```diff
@@ -454,6 +454,10 @@ func (idx *Indexer) allConvs(ctx context.Context, uid gregor1.UID) (map[string]t
 			chat1.ConversationStatus_FAVORITE,
 			chat1.ConversationStatus_MUTED,
 		},
+		MemberStatus: []chat1.ConversationMemberStatus{
```
mmaxim: Why bother with this?

go/chat/teamchannelsource.go
```diff
@@ -4,225 +4,135 @@ import (
...
+	inbox, err := c.getTLFConversations(ctx, uid, teamID, topicType)
 	if err != nil {
-		return res, err
+		return nil, err
 	}
-	convs = tlfRes.Conversations
-
-	// Localize the conversations
-	res, _, err = c.G().InboxSource.Localize(ctx, uid, utils.RemoteConvs(convs),
+	convs, _, err := c.G().InboxSource.Localize(ctx, uid, inbox.ConvsUnverified,
```
mmaxim: Can just call `Read`

go/chat/search/indexer.go
joshblum: `RESET` is included in the default inbox query, might speed things up

go/chat/teamchannelsource.go
joshblum: fixed

joshblum: reverted, using `Read` makes things more complex since we have to convert the `TlfID` to a name and would have to thread the `MembersType` through for that. (just to convert back to a tlfID when we change the local query to a remote one). think it's simpler as is

joshblum:
new test depends on https://github.com/keybase/keybase/pull/3245

joshblum:
merged in @mm

go/chat/storage/inbox.go
```diff
@@ -490,83 +489,78 @@ func (i *Inbox) applyQuery(ctx context.Context, query *chat1.GetInboxQuery, rcs
 			queryConvIDMap[c.String()] = true
 		}
 	}
+
+	queryMemberStatusMap := map[chat1.ConversationMemberStatus]bool{}
```
mmaxim: Can we just make this a static thing so we aren't creating it all the time?

joshblum: but it's built from the query

mmaxim: Right but 99% of the time it will be the same values, I was thinking we could swap in a map with the defaults values in that case instead of creating it.

go/chat/teamchannelsource.go
joshblum: i started threading it through and i think was hitting errors here (if i'm remember correctly) https://github.com/keybase/client/blob/master/go/chat/inboxsource.go#L27

i also prefer it as is since both `GetChannelsTopicName` and `GetChannelsFull` can use `getTLFConversations` instead of duplicating the query between `Read` and `ReadUnverified` (`GetChannelsTopicName` can't call `Localize` or we get into some unboxing cycle)

mmaxim: Sure this is fine.

go/chat/storage/inbox.go
joshblum: sgtm

joshblum:
@mmaxim this is ready for final pass

PR #14850, opened 2018-11-27
People page announcements

chrisnojima:
@keybase/react-hackers this does the front end side of announcements. I made a core ticket to implement this on the server side. It's 80% of the way plumbed through. We can get this in (its behind a feature flag) while we wait


chrisnojima:
It doesn't serve a purpose other than extra code. If the props don't conform flow will complain

PR #14852, opened 2018-11-28
enable moveOrCopy for admins on mobile

songgao:


songgao:
Oh I saw `admin: __DEV__` and thought that was it. I'll just turn it on then, and we can turn it back off before release if we want.

PR #14853, opened 2018-11-28
use proper keys for FS rows

songgao:
When using index as key, if a file gets deleted while a PathItemAction popup is shown, the content shifts to the next file, along with the buttons associated to it. So instead, just use proper keys so mounted rows are moved properly when an item is removed, instead of shifting props for a bunch of them.

cc @chrisnojima


songgao:
Thanks!

PR #14854, opened 2018-11-28
Removed bombs, replaced with stopwatch

cecileboucheron:
We're getting rid of the bomb icon and replacing it with a stopwatch everywhere. 
@keybase/react-hackers 

image

image

image


shared/chat/conversation/messages/wrapper/exploding-meta/index.js
```diff
@@ -115,10 +115,10 @@ class ExplodingMeta extends React.Component<Props, State> {
     let children
     switch (this.state.mode) {
       case 'countdown':
-        let bombIconSize = isMobile ? 22 : 16
+        let timerIconSize = isMobile ? 22 : 16
         if (isAndroid) {
           // icon is 24 high and clips edge of container on android. workaround
```
adamjspooner: Do you know if this is still needed?

cecileboucheron: @chrisnojima Do you know if this is still needed?

adamjspooner: ...taking a look.

chrisnojima: lets touch this stuff very little, i'm in the middle of all of this

PR #14856, opened 2018-11-28
Send payment chat message in a goroutine

patrickxb:
This cuts the sendPayment with quickReturn=true runtime in half.

### DIFF
--- a/go/stellar/stellar.go
+++ b/go/stellar/stellar.go
@@ -589,10 +589,13 @@ func sendPayment(m libkb.MetaContext, remoter remote.Remoter, sendArg SendPaymen
 	}
 
 	if senderEntry.IsPrimary {
-		if err := chatSendPaymentMessage(m, recipient, rres.StellarID); err != nil {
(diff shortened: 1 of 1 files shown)


go/stellar/stellar.go
```diff
@@ -589,9 +589,17 @@ func sendPayment(m libkb.MetaContext, remoter remote.Remoter, sendArg SendPaymen
...
-			m.CDebugf("failed to send chat SendPayment message: %s", err)
+		sendChat := func() {
+			mgo := m.WithCtx(context.Background())
+			if err := chatSendPaymentMessage(mgo, recipient, rres.StellarID); err != nil {
+				// if the chat message fails to send, just log the error
+				mgo.CDebugf("failed to send chat SendPayment message: %s", err)
+			}
+		}
+		if sendArg.QuickReturn {
+			go sendChat()
```
mmaxim: Could just pass `ctx` as a param to `sendChat`

PR #14857, opened 2018-11-28
handle exploded messages better with mark as read CORE-9443

mmaxim:
Patch does the following:

1.) Don't mark boxed expired exploding messages as errors _a priori_. Otherwise, we never see ash for these (common case on mobile to miss them for short durations).
2.) Change JS mark as read to use the large of the max message it knows about, and the max msgID on the conversation.

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -1982,7 +1982,8 @@ const markThreadAsRead = (
     return
   }
 
-  if (!state.chat2.metaMap.get(conversationIDKey)) {
+  const meta = state.chat2.metaMap.get(conversationIDKey)
+  if (!meta) {
     logger.info('marking read bail on not in meta list. preview?')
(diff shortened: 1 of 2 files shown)


shared/actions/chat2/index.js
```diff
@@ -2010,15 +2011,11 @@ const markThreadAsRead = (
     message = mmap.get(ordinal)
   }
 
-  if (!message) {
-    logger.info('marking read bail on no messages')
-    return
-  }
-
-  logger.info(`marking read messages ${conversationIDKey} ${message.id}`)
+  const readMsgID = message ? (message.id > meta.maxMsgID ? message.id : meta.maxMsgID) : meta.maxMsgID
```
joshblum: 👀

mmaxim: 👓

PR #14860, opened 2018-11-28
improve scraper success by making request look more real CORE-9566

mmaxim:


mmaxim:
I don't want to lie, I tried other chat apps, and they come with something similar.

mmaxim:
This apparently works for Telegram:
```
TelegramBot (like TwitterBot)
```

joshblum:
oh wow

PR #14861, opened 2018-11-28
client test for downgrade implicit admin after reset

heronhaye:
Test for https://github.com/keybase/keybase/pull/3176. Fails on master.

### DIFF
--- a/go/teams/rotate_test.go
+++ b/go/teams/rotate_test.go
@@ -620,3 +620,34 @@ func TestRotateAsSubteamWriter(t *testing.T) {
 	require.NoError(t, err)
 	require.EqualValues(t, oldGeneration+1, teamAfter.Generation())
 }
+
+func TestDowngradeImplicitAdminAfterReset(t *testing.T) {
+	tc, owner, otherA, otherB, root, sub := memberSetupSubteam(t)
+	defer tc.Cleanup()
+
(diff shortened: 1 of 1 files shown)


go/teams/rotate_test.go
```diff
@@ -620,3 +620,34 @@ func TestRotateAsSubteamWriter(t *testing.T) {
...
+	_, err = GetForTestByStringName(context.Background(), tc.G, root)
+	require.NoError(t, err)
+
+	_, err = AddMember(context.TODO(), tc.G, root, otherA.Username, keybase1.TeamRole_ADMIN)
+	require.NoError(t, err)
+
+	err = EditMember(context.TODO(), tc.G, root, otherA.Username, keybase1.TeamRole_WRITER)
+	require.NoError(t, err)
+
+	// This fails if a box incorrectly remains live for otherA after the downgrade due
```
zapu: OK I understand what's going on here now. If CLKR fired for the role change, and owner rotated team here, like that:
```
	zzz, err := GetForTestByStringName(context.Background(), tc.G, sub)
	require.NoError(t, err)
	err = RotateKey(context.TODO(), tc.G, zzz.ID)
	require.NoError(t, err)
```

then all is fine - the test passes when rotatekey call is included (without server helper change).

zapu: Now the question is whether we send CLKRs properly when member's role is downgraded.

PR #14862, opened 2018-11-28
WIP: refactor chat message layout

chrisnojima:


shared/app/nav.native.js
```diff
@@ -75,7 +75,8 @@ class CardStackShim extends Component<CardStackShimProps> {
           .map((route, index) => {
             const routeName = route.path.join('/')
             // The bottom/back item of the stack is our top (active) screen
-            const shouldRender = !this.props.hidden && (index === stack.size - 1 || index === stack.size - 2)
+            const shouldRender =
```
chrisnojima: I have another pr with just this. since you can't swipe back on android it seems worth it to just render the topmost always

shared/chat/conversation/messages/index.js
```diff
@@ -1,164 +1,3 @@
...
-    const messageWrapperProps = {
-      isEditing: this.props.isEditing,
-      measure: this.props.measure,
-      message: this.props.message,
-      previous: this.props.previous,
-    }
-
-    switch (this.props.message.type) {
-      case 'text':
-        return <WrapperMessage {...messageWrapperProps} />
```
chrisnojima: since everything is wrapped, this and wrapper were just merged

shared/chat/conversation/messages/reactions-row/index.js
```diff
@@ -98,8 +98,6 @@ const styles = styleSheetCreate({
   container: {
     alignItems: 'flex-start',
     flexWrap: 'wrap',
-    // refer to `WrapperAuthor` styles
-    marginLeft: 32 + globalMargins.tiny + (isMobile ? globalMargins.tiny : globalMargins.small),
```
chrisnojima: things under messages now have a padding applied at a higher level so this kind of hackiness can go away

shared/chat/conversation/messages/set-channelname/index.js
```diff
@@ -1,42 +1,14 @@
...
-  channelname: ?string,
-  onUsernameClicked: () => void,
-  setUsernameBlack: boolean,
-  timestamp: number,
+  message: Types.MessageSetChannelname,
 }
 
-const SetChannelname = (props: Props) => {
-  return (
-    <SmallUserNotice
```
chrisnojima: SmallUserNotice was used to emulate a message w/ author style. Instead we just treat those like messages w/ an author and remove all this code. Plus it'll be 100% the same as normal messages.
Note. This means that these rows now have timestamps on them, like the regular ones do. but design said this is preferred anyways

shared/chat/conversation/messages/text/index.js
```diff
@@ -13,16 +13,26 @@ export type Props = {
...
+      meta={{mentionsAt, mentionsChannel, mentionsChannelName}}
+      styleOverride={Styles.isMobile ? {paragraph: getStyle(type, isEditing)} : undefined}
+      allowFontScaling={true}
+    >
+      {text}
+    </Kb.Markdown>
+  )
+
+  return Styles.isMobile ? (
+    <Kb.Box2 direction="vertical" style={styles.wrapper}>
```
chrisnojima: mobile needs this wrapper as markdown itself will export sibling texts inside a fragment so they get laid out based on whatever the parent is. instead we need these to stack vertically

shared/chat/conversation/messages/wrapper/index.js
```diff
@@ -1,5 +1,530 @@
 // @flow
-import WrapperMessage from './wrapper-message'
-import WrapperAuthor from './wrapper-author'
+import * as Kb from '../../../../common-adapters'
+import * as React from 'react'
+import * as Styles from '../../../../styles'
+import * as Types from '../../../../constants/types/chat2'
+import * as Constants from '../../../../constants/chat2'
+import SystemAddedToTeam from '../system-added-to-team/container'
```
chrisnojima: instead of MessageFactory which is a connect + a big switch statement, which then delegates to the wrapper 100% of the time, we just have wrapper, which knows how to render a message

shared/chat/inbox/row/small-team/top-line.js
```diff
@@ -32,14 +32,9 @@ class _SimpleTopLine extends React.Component<Props> {
   }
 
   render() {
-    const boldOverride = this.props.showBold ? Styles.globalStyles.fontBold : null
+    const boldStyle = this.props.showBold ? styles.bold : null
```
chrisnojima: some small changes to use styleSheets and fastBlank to fix overdraw issues in the inbox

shared/constants/chat2/message.js
```diff
@@ -1100,20 +1100,3 @@ export const messageExplodeDescriptions: Types.MessageExplodeDescription[] = [
   {text: '7 days', seconds: 86400 * 7},
   {text: 'Never explode (turn off)', seconds: 0},
 ].reverse()
-
-// Used to decide whether to show the author wrapper
-export const showAuthorMessageTypes = ['attachment', 'requestPayment', 'sendPayment', 'text']
```
chrisnojima: these are very specific to wrapper so just put it there

shared/styles/colors.js
```diff
@@ -43,7 +43,7 @@ const colors = {
   darkBlue3_75_on_white: 'rgb(82,98,147)',
   darkBlue4: '#1036AC',
   // on iOS overdraw is eliminiated if we use white, on Android it's eliminated if it's transparent /shrug
-  fastBlank: isAndroid ? undefined : '#ffffff',
+  fastBlank: isIOS ? '#ffffff' : undefined,
```
chrisnojima: this lets us use this on desktop files also

shared/util/timestamp.js
```diff
@@ -4,28 +4,28 @@ import moment from 'moment'
...
   const now = moment()
   const today = now.clone().startOf('day')
+  if (m.isSame(today, 'd')) {
+    return hma
+  }
   const yesterday = today
     .clone()
     .subtract(1, 'day')
     .startOf('day')
-  const lastWeek = today.clone().subtract(7, 'day')
```
chrisnojima: defer calling these until we need them. came up in profiling

chrisnojima:
(i'm waiting on snaps until the pr is reviewed)

akalin-keybase:
Tested on desktop. Looks mostly good, although I did notice a regression for the revoked icon. Latest release:

screen shot 2018-12-03 at 11 20 45 pm
screen shot 2018-12-03 at 11 20 51 pm

This PR:

screen shot 2018-12-03 at 11 38 21 pm
screen shot 2018-12-03 at 11 38 15 pm

Minor enough that it can be fixed in a separate PR, though.

akalin-keybase:
However, I did find a more serious problem on iOS. Before:

simulator screen shot - iphone 6 - 2018-12-03 at 23 50 42

After:

simulator screen shot - iphone 6 - 2018-12-03 at 23 53 28

akalin-keybase:
Didn't find any obvious problems on android.

mmaxim:
The lock seems too far to the right on desktop, in fact it seems to temporarily trigger the vertical scroll bar for me when I send messages.

mmaxim:
Orange line is also too close to avatar

image

mmaxim:
Spacing seems to high between messages with thew avatar showing

image

mmaxim:
The exploding message bomb overlaps the lock on message send:
image

mmaxim:
The "Cancel" button on Edit seems to push the edited message to the right:

image

chrisnojima:
waiting on dannys feedback
updated timings: static storybook build (release mode) 3 loads
master: 487, 414, 388 ms
PR: 196, 164, 143 ms

cecileboucheron:
Avatars are indeed too high - already discussed with @chrisnojima

shared/chat/conversation/messages/wrapper/exploding-meta/index.js
```diff
@@ -46,9 +28,7 @@ type State = {
 }
 
 class ExplodingMeta extends React.Component<Props, State> {
```
buoyad: testing on iOS, the left side of this is cut off 
image

shared/chat/conversation/messages/wrapper/send-indicator/index.js
```diff
@@ -102,14 +99,22 @@ class SendIndicator extends React.Component<Props, State> {
       return null
     }
     return (
-      <SendIcon
-        status={this.state.iconStatus}
-        style={{...this.props.style, opacity: this.state.visible ? 1 : 0}}
+      <Kb.Icon
```
buoyad: on iOS this wraps to the next line 
send-wrapping

mmaxim:
image

I think Manage Subscriptions is supposed to be on the next line

mmaxim:
Actually the whole display gets weirdly centered:

image

shared/chat/conversation/messages/wrapper/exploding-meta/index.js
```diff
@@ -111,56 +91,57 @@ class ExplodingMeta extends React.Component<Props, State> {
...
-          </Box2>
+            <Kb.Icon type="iconfont-bomb" fontSize={bombIconSize} color={Styles.globalColors.black_75} />
+          </Kb.Box2>
         )
         break
       case 'boom':
         children = (
-          <Box2 direction="horizontal" style={styles.boomIconContainer}>
-            <Icon
+          <Kb.Box2 direction="horizontal" style={styles.boomIconContainer}>
```
buoyad: This doesn't seem to show on explode anymore (this might also be busted on master)

shared/chat/conversation/messages/react-button/index.js
```diff
@@ -207,7 +207,7 @@ export class NewReactionButton extends React.Component<NewReactionButtonProps, N
                 style={iconCastPlatformStyles(
```
buoyad: probably related to higher level styles and not these changes, but the emoji picker jumps around when trying to choose an emoji on desktop. 
picker-pogo
(only clicked on the icon once in this, hovering on it again brings the picker back)

shared/chat/conversation/messages/wrapper/exploding-meta/index.js
chrisnojima: yeah this happens on master and its something about the timing logic (not styling) in react tools if i force state "boom" then it works fine

PR #14866, opened 2018-11-29
Embolden amount in button.

adamjspooner:


shared/chat/conversation/messages/account-payment/container.js
```diff
@@ -82,8 +82,7 @@ const mapStateToProps = (state, ownProps: OwnProps) => {
         canceled,
         claimButtonLabel:
           !youAreSender && cancelable && !acceptedDisclaimer
-            ? `Claim${paymentInfo.worth ? ' Lumens worth' : ''} ${paymentInfo.worth ||
-                paymentInfo.amountDescription}`
```
zanderz: This changes the content too, was that intended?

adamjspooner: Yup, this takes the amount out of the text.

PR #14867, opened 2018-11-29
Truncate long unfurl descriptions.

adamjspooner:


mmaxim:
Does it put like a `...` if it gets truncated?

adamjspooner:
Yup!

PR #14870, opened 2018-11-29
Update new conversation styles.

adamjspooner:
@keybase/react-hackers, this is ready for review. This is part of the new conversation QA. I'll be filing issues in JIRA for non-styling related changes.


shared/team-building/service-tab-bar.js
```diff
@@ -74,44 +75,33 @@ const ServiceTabBar = (props: Props) => (
 )
 
 const styles = Styles.styleSheetCreate({
-  container: Styles.platformStyles({
```
zanderz: did a tool tell u this was not in use or just eyeballed it?

adamjspooner: I used `rg`. There are no uses of `styles.container` in this file.

PR #14874, opened 2018-11-29
Shorten chat input placeholder with wallets icon

buoyad:
`PlatformInput` needed to know whether the wallets icon was being rendered so it could show the shorter placeholder in that case. I moved the logic from `WalletsIcon`'s container into constants and use it in `PlatformInput` to decide both the placeholder and whether to render the icon. The problem existed on small screens too so I shortened the placeholder further to 'Exploding'.

#### Large screens
Before:
image

After:
image

#### Small screens
Before:
image

After:
image

r? @keybase/react-hackers


buoyad:
@keybase/react-hackers bump

PR #14876, opened 2018-11-30
linkify KBFS paths in chat messages and info-panel

songgao:
* extend MarcoDown to capture KBFS paths if `onOpenInFilesTab` is provided in `meta`
* add a HOC to provide `onOpenInFilesTab`, as it's used more than once
* linkify KBFS paths in chate messages
* linkify KBFS paths in info-panels of big team channels

Works on both desktop and mobile, but back button doesn't work intuitively yet. I've created [KBFS-3654](https://keybase.atlassian.net/browse/KBFS-3654) to track it.

From storybook, to show how KBFS paths are matched: 

screen shot 2018-11-30 at 3 48 35 pm


songgao:
Ugh I'm realizing this doesn't take care of spaces in path. Untagging reviewers!

mmaxim:
Just wanted to pop in and say I'm glad this is going to in the app soon!

songgao:
ready for review again!

shared/fs/common/open-in-files-tab-hoc.js
```diff
@@ -0,0 +1,17 @@
...
+import * as React from 'react'
+import * as Types from '../../constants/types/fs'
+import * as FsGen from '../../actions/fs-gen'
+import {namedConnect} from '../../util/container'
+
+const mapDispatchToProps = dispatch => ({
+  onOpenInFilesTab: (path: Types.Path) => dispatch(FsGen.createOpenPathInFilesTab({path})),
+})
+
+export default (ComposedComponent: React.ComponentType<any>) =>
```
buoyad: Can you parametrize the `ComponentType<T>`? Components this is used on are losing their type.

```
    <BigTeamHeader
     ^ ComponentType<any>
```

buoyad: Then I think you shouldn't need to explicitly add `onOpenInFilesTab` to the props for the components

songgao:
Thanks for looking!

> Do we have a plan to handle the black bars for invalid accesses?

Yeah, there's a [screen](https://app.zeplin.io/project/56b25ffc64fc91773a73aaf8/screen/5a7deef2bb95a5e0086c449b) for that; will do it afterwards!

shared/common-adapters/markdown/shared.js
```diff
@@ -302,6 +319,30 @@ const rules = {
...
+    },
+    parse: (capture, parse, state) => {
+      const content = pathToString(unescapePath(capture[0]))
+      return {
+        type: 'kbfsPath',
+        content,
+        onClick: () => state.markdownMeta.onOpenInFilesTab(stringToPath(content)),
+      }
+    },
+  },
```
songgao: @MarcoPolo mind taking a look at the Marcodown changes here?

MarcoPolo:
Just looking at the screenshot you posted, Shouldn't it match trailing `/` ? also maybe it shouldn't match the `/keybase/` in `/keybase/privatemessedup`. You shouldn't match a partial word. If the whole word doesn't match you don't want to match the first part

shared/common-adapters/markdown/shared.js
```diff
@@ -302,6 +319,30 @@ const rules = {
...
   },
+  kbfsPath: {
+    order: SimpleMarkdown.defaultRules.autolink.order - 0.1, // lower than mention
+    match: (source, state, lookBehind) => {
+      const kbfsPathRegex = getKbfsPathRegex(state.markdownMeta)
+      if (!kbfsPathRegex) {
+        return null
+      }
+
+      const matches = SimpleMarkdown.inlineRegex(kbfsPathRegex)(source, state, lookBehind)
```
MarcoPolo: set `const kbfsPathRegex =` above to be `SimpleMarkdown.inlineRegex(createKbfsathRegex())`

shared/common-adapters/markdown/shared.js
```diff
@@ -78,6 +80,21 @@ function createChannelRegex(meta: ?MarkdownMeta): ?RegExp {
...
+  const username = `([a-zA-Z0-9]+_?)+` // from go/kbun/username.go
+  const usernames = `${username}(,${username})*`
+  const teamName = `${username}(\\.${username})*`
+  const tlfType = `/(private|public|team)`
+  const tlf = `/((private|public)/${usernames}(#${usernames})?|team/${teamName})`
+  const inTlf = `/(\\S|\\\\\\\\|\\\\ )*[^/\\s]+` // don't include any trailing slash
+  return new RegExp(`^/keybase((${tlf}(${inTlf})?)|(${tlfType}))?`)
+}
+
+const kbfsPathRegex = createKbfsPathRegex()
```
MarcoPolo:
```suggestion
const kbfsPathMatcher = SimpleMarkdown.inineRegex(createKbfsPathRegex())
```

MarcoPolo: (See comment below)

shared/common-adapters/markdown/shared.js
```diff
@@ -78,6 +80,21 @@ function createChannelRegex(meta: ?MarkdownMeta): ?RegExp {
...
+
+const getKbfsPathRegex = (meta: ?MarkdownMeta): ?RegExp =>
```
MarcoPolo: Kill this, it shouldn't work this way. It should work like how channel works

shared/common-adapters/markdown/shared.js
```diff
@@ -302,6 +319,30 @@ const rules = {
...
+        return matches
+      }
+      return null
+    },
+    parse: (capture, parse, state) => {
+      const content = pathToString(unescapePath(capture[0]))
+      return {
+        type: 'kbfsPath',
+        content,
+        onClick: () => state.markdownMeta.onOpenInFilesTab(stringToPath(content)),
```
MarcoPolo: 🔥

shared/common-adapters/markdown/shared.js
```diff
@@ -78,6 +80,21 @@ function createChannelRegex(meta: ?MarkdownMeta): ?RegExp {
   return new RegExp(`^#(${meta.mentionsChannelName.keySeq().join('|')})\\b`)
 }
 
+function createKbfsPathRegex(): ?RegExp {
+  const username = `([a-zA-Z0-9]+_?)+` // from go/kbun/username.go
+  const usernames = `${username}(,${username})*`
```
MarcoPolo: 🐑 
```suggestion
  const usernames = `${username}(?:,${username})*`
```

And same thing below so you don't capture this group since you don't have to.

shared/common-adapters/markdown/shared.js
```diff
@@ -78,6 +80,21 @@ function createChannelRegex(meta: ?MarkdownMeta): ?RegExp {
...
+  const teamName = `${username}(\\.${username})*`
+  const tlfType = `/(private|public|team)`
+  const tlf = `/((private|public)/${usernames}(#${usernames})?|team/${teamName})`
+  const inTlf = `/(\\S|\\\\\\\\|\\\\ )*[^/\\s]+` // don't include any trailing slash
```
MarcoPolo: Why do you need the `\\s`?

shared/common-adapters/markdown/shared.js
```diff
@@ -302,6 +319,30 @@ const rules = {
...
+
+      const matches = SimpleMarkdown.inlineRegex(kbfsPathRegex)(source, state, lookBehind)
+      // Also check that the lookBehind is not text
+      if (matches && (!lookBehind || lookBehind.match(/\B$/))) {
+        return matches
+      }
+      return null
+    },
+    parse: (capture, parse, state) => {
+      const content = pathToString(unescapePath(capture[0]))
```
MarcoPolo: This should probably happen in the component for this instead of here. Then you can just rely on the default implementation of parse and kill this code.

shared/chat/conversation/messages/text/index.stories.js
```diff
@@ -18,6 +18,7 @@ const props = {
 const provider = Sb.createPropProviderWithCommon({
   Channel: p => ({name: p.name}),
   Mention: p => ({username: p.username}),
+  OpenInFilesTabHoc: o => ({...o, onOpenInFilesTab: Sb.action('onOpenInFilesTab')}),
```
MarcoPolo: 🔥

shared/common-adapters/markdown/index.js.flow
```diff
@@ -23,9 +24,10 @@ export type MarkdownCreateComponent = (
 ) => ?React.Node
 
 export type MarkdownMeta = {
-  mentionsAt: MentionsAt,
-  mentionsChannelName: MentionsChannelName,
-  mentionsChannel: MentionsChannel,
+  mentionsAt?: MentionsAt,
+  mentionsChannelName?: MentionsChannelName,
+  mentionsChannel?: MentionsChannel,
+  onOpenInFilesTab?: (Path)=>void,
```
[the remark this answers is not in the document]
songgao: @MarcoPolo you mean file path should have its own container in `common-adapters`? I asked in frontend and people seemed to not like the idea of having this inside common-adapters, but it also feels weird to import a container from e.g. `fs/` into `common-adapters`. That's why I went with the HOC approach. I do like having a container here; it'd be much cleaner.

shared/common-adapters/markdown/shared.js
songgao: The channel thing calls `createChannelRegex` in `match`, which creates a new RegExp. Doesn't it result creating new objects all the time?

songgao: Oh I see what you're saying now. It'd still be the same object.

MarcoPolo: That's a good point, we could probably optimize it by doing the regex creation just ones inside the markdown meta (since the channel regex is dynamic)

shared/common-adapters/markdown/index.js.flow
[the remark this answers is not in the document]
MarcoPolo: Not necessarily in common-adapters, but it should have a FilePath component and a container component like how channel does.

MarcoPolo:
Yeah, get rid of the OpenInFilesHoc.

shared/common-adapters/markdown/index.js.flow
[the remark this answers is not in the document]
songgao: Oops I named `KbfsPath`. Let me know if you feel strongly it should be `FilePath` instead and I'll make change.

shared/common-adapters/markdown/shared.js
songgao: Sounds like we should do the same for other stuff (channel mentions, etc.) here too, So we don't create the objects all the time. Is this what you mean?

PR #14884, opened 2018-11-30
simple markdown for all

chrisnojima:
@keybase/react-hackers


mmaxim:
As mentioned in chat, I think we should fix the class of bugs where punctuation at the end of a link gets included in the link itself. It makes it so that clicking the link takes you to the wrong URL (usually a 404), which is a pretty bad experience.

songgao:
Can this resurrect now?

PR #14890, opened 2018-11-30
account bundle migration cleanup

xgess:
make it look like the migration never happened and that we were always only using v2 bundles. 

note: this branch removes the bundle migration code, so a user with a v1 bundle would have bad time on here (just black bars and missing functionality - nothing irreversible). 2.13 isn't coming out for a while and we're not talking about very many people even right now, so i'm not too concerned. regardless, i'm tracking it. 

* rip out all of the code that only supports v1 bundles
* refactoring and renaming so it doesn't look like the v2 stuff is bolted on the way it used to. 
* take all of the encoding/decoding tests from the v1 world and port them into the v2 world
* remove the `stellar_acct_bundles` feature flag
* implement CORE-9574 (the more dynamic Upkeep flow)
* additional cleanup enabled by the Upkeep work: stop returning pukGen in a whole bunch of places that don't care about it since there's a new flow for it.


go/stellar/bundle/boxer_test.go
```diff
@@ -13,314 +15,483 @@ import (
...
+		map[keybase1.PerUserKeyGeneration]libkb.PerUserKeySeed{
+			pukGen: pukSeed,
+		},
+	}
+}
+
+var validAccounts = []accountCan{
+	{"GAWZ7HVPKRGCH2KP6475XV6HA2CAF44MXWWE5RKV4LMMGB6FNNSEPNPE", "hKFlxKUc3fqMOAYv9m6ycpnpQA4CSriSQoPVNbvsXXEgv4WsixkaTThkKGMlYuWvTJdHeqPdYXo/Mw156xq8MaIbzDTriFplFNcLhNYdi8f1ViHqvVIecX2frU/BOtIsAlqknnhl4+Z1u2kUnZYI5pZRvUV5H1loSWC4tBmEoCgK1S6XrLx1POOIiKkH8EFMXVrB6+BjbieW1w8HTXNp0jWbKkq+QoKz8MCjZ2VuA6FuxBibtfMD40wYqnfPsGePn1RphpzcEqv7mZehdgE="},
+	{"GBPLLHOKPRFSCE6FP6D37AUUNYCJ2TSPOHMHAR54TBDLWWPIZCD4RL3G", "hKFlxKX3zqweGMWyl4vOC8ht0jngDnTEpWqGBePh7okF9S003QStjI9Td1d+urqj/k4etwgLiPO8LHaaQ7o2WAwuSXakHJVJ2xIq+T+MYnoobbx5sArk4wsg7AWTX+uw/rXoyN7P9ZAgDSH+4oZ6/0j8dwR6RAoc9c+dog8xnW3eTkeSlj2KYx6XnEOOlOcCBCKsZePi9eoN92CxvB5MRN6tR7w3PH4yynqjZ2VuA6FuxBha4JmmVyysST3avwMxayHfTExp7tnLwHOhdgE="},
+}
```
patrickxb: was this meant to be used in `cans` below for the `accounts` field?

xgess: ahh. yes. you are correct. will remove.

go/stellar/stellar.go
```diff
@@ -1704,7 +1685,7 @@ func DeleteAccount(m libkb.MetaContext, accountID stellar1.AccountID) error {
 	if !found {
 		return fmt.Errorf("account not found: %v", accountID)
 	}
-	return remote.Post(m.Ctx(), m.G(), nextBundle, version)
+	return remote.Post(m, nextBundle)
```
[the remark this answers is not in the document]
xgess: the version being passed into `Post` was so it knew whether to send a v1 bundle to `stellar/bundle` or a v2 bundle to `stellar/acctbundle`. after this branch, there is no more posting to `stellar/bundle` and all bundles are v2 everywhere.

go/stellar/remote/remote.go
```diff
@@ -494,110 +183,103 @@ func FetchV2BundleForAccount(ctx context.Context, g *libkb.GlobalContext, accoun
...
 // of the information. This will error from any device that does not have access
-// to all of the accounts (e.g. a desktop after mobile-only)
-func FetchWholeBundle(ctx context.Context, g *libkb.GlobalContext) (acctBundle *stellar1.BundleRestricted, version stellar1.BundleVersion, pukGen keybase1.PerUserKeyGeneration, err error) {
-	defer g.CTraceTimed(ctx, "Stellar.FetchWholeBundle", func() error { return err })()
-
-	mctx := libkb.NewMetaContext(ctx, g)
-	bundle, version, pukGen, err := FetchSecretlessBundle(mctx)
+// to all of the accounts (e.g. a desktop after mobile-only). This is only used
+// in tests.
+func FetchWholeBundle(mctx libkb.MetaContext) (bundle *stellar1.Bundle, err error) {
```
patrickxb: maybe we could rename it `FetchWholeBundleForTesting`, assuming it won't work to shove it in an _test.go file.

xgess: yeah - this was also used in Upkeep (which i've since replaced) and the CLI dump command (which only works in devel and has also since been replaced). i kinda kept chipping away at it because it seems less safe than fetching by account. but yeah. you're probably right. i think i can get this out of here.

PR #14891, opened 2018-12-01
unfurling: process Giphy URLs custom to display the Gif inline (and no other metadata)

mmaxim:
Patch does the following:

**Go**
1. Add a new `UnfurlType`, `GIPHY`, which is triggered for domains either like `giphy.com` or `gph.is`. Add all the corresponding unfurl types that correspond to this.
2. Generalize `Scraper` a little so that specialized scrapers can still take advantage of things the generic scraper does (like find favicons). Build the Giphy unfurl from just the image in `head` and the favicon.
3. Package up `VIDEO` `Asset`'s that we find on the page as long as they are MIME type `image/gif`. Change the display code to get the height/width from the right place depending on the type.

**JS**
@keybase/react-hackers 
1. Add a new component for displaying Giphy unfurls called `UnfurlGiphy`. This is similar to `UnfurlGeneric` but is much simpler since we aren't displaying as much info.
2. Factor out the image display code into a new component called `UnfurlImage` which gets the dimensions of the displayed image correct. Use this in both `UnfurlGeneric` and `UnfurlGiphy`.
3. Hook everything up to the Giphy type coming up from the service.
4. Tested display on desktop and on iPhone XS real device and iPhone 5s sim and it all looks good. The display of the GIF is turned into a JPG if it is too big just like a chat attachment, but I found that Giphy gives a reasonably sized Gif for all the those I could find anyway.

### DIFF
--- a/go/protocol/chat1/extras.go
+++ b/go/protocol/chat1/extras.go
@@ -1888,13 +1888,15 @@ func (u UnfurlRaw) String() string {
 	return "<unknown>"
 }
 
-func (g UnfurlGenericRaw) String() string {
-	yield := func(s *string) string {
-		if s == nil {
-			return ""
-		}
-		return *s
+func yieldStr(s *string) string {
+	if s == nil {
+		return ""
 	}
+	return *s
+}
+
+func (g UnfurlGenericRaw) String() string {
+
 	publishTime := ""
 	if g.PublishTime != nil {
 		publishTime = fmt.Sprintf("%v", time.Unix(int64(*g.PublishTime), 0))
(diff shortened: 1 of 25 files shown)


go/protocol/chat1/extras.go
```diff
@@ -1882,17 +1882,21 @@ func (u UnfurlRaw) String() string {
...
-		}
-		return *s
+func yieldStr(s *string) string {
+	if s == nil {
+		return ""
 	}
+	return *s
+}
+
+func (g UnfurlGenericRaw) String() string {
```
joshblum: maybe we should call this `UnsafeDebugString` so it's only for testing since it would leak info about the websites visited to the log

mmaxim:
Ok @chrisnojima, this uses the MP4 for the Giphy link if it is available. Seems like a nice win, they are usually much smaller and easier to display.

go/protocol/chat1/extras.go
mmaxim: Done!

shared/chat/conversation/messages/unfurl/giphy/index.js
```diff
@@ -0,0 +1,101 @@
...
+export type Props = {
+  imageHeight: number,
+  imageWidth: number,
+  imageURL: string,
+  isVideo: boolean,
+  faviconURL?: string,
+  onClose?: () => void,
+}
+
+class UnfurlGiphy extends React.Component<Props> {
```
buoyad: looks like this could be a function too

mmaxim:
I tested on a pixel 2 running API 28 and it worked.

PR #14903, opened 2018-12-03
fill in queriedBy fields for results from resolver cache

maxtaco:
- otherwise, callers like ResolveThenIdentify2 will get the wrong answer for WasKBAssertion()


maxtaco:
CI fail is a flake in `TeamRekey` uggg but not related

PR #14905, opened 2018-12-03
Move build/request local stellarsvc->stellar

mlsteele:
Move build/request local handlers from `stellarsvc` down to `stellar`. As build become less simple I suspect having it in the same package as `stellar.Stellar` will be easier. Less threading stuff between packages.


mlsteele:
🙏 I can wait if you'd like

patrickxb:
it's fine, go ahead...

PR #14910, opened 2018-12-03
Surya/core 9183/additional imptofu rpcs

heronhaye:
Fails on local, needs web pr. Otherwise done.

- [x]    expose all email/phone func as rpc calls
- [x]    fix the skipped tests due to visibility changes on server
- [x]    basic happypath tests hitting every endpoint
- [x]    Add API, RPC, client test for a “Edit phone number” endpoint, which exists for emails but not for phone numbers (Rejected in favor of service calling Delete + Add)
- [x]    cmdline for everything (Done except for bulk lookups)


protocol/avdl/keybase1/common.avdl
```diff
@@ -340,4 +344,9 @@ protocol Common {
     string username;
     UnixTime ctime;
   }
+
+  enum IdentityVisibility {
+      PRIVATE_0,
+      PUBLIC_1
+  }
 }
```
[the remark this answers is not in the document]
zapu: same with `RegionCode`

protocol/avdl/keybase1/user.avdl
```diff
@@ -158,4 +160,7 @@ protocol user {
   */
   NextMerkleRootRes findNextMerkleRootAfterReset(UID uid, Seqno resetSeqno, ResetMerkleRoot prev);
 
+  // Phone number in E.164 format.
+  @typedef("string")
+  record EmailAddress {}
```
zapu: wrong comment?

protocol/avdl/keybase1/phone_numbers.avdl
```diff
@@ -10,13 +10,30 @@ protocol phoneNumbers {
...
+      IdentityVisibility visibility;
       @jsonkey("ctime")
       UnixTime ctime;
+
+    }
+    record PhoneNumberLookupResult {
+       @jsonkey("phone_number")
+       PhoneNumber phoneNumber;
+       @jsonkey("coerced_phone_number")
+       PhoneNumber coercedPhoneNumber;
```
zapu: so `phoneNumber` and `coercedPhoneNumber` are the same type? I thought that type will mean "well-formed, E164 phone number with proper region code".

go/client/cmd_email_delete.go
```diff
@@ -0,0 +1,59 @@
...
+		ArgumentHelp: "<email>",
+		Action: func(c *cli.Context) {
+			cl.ChooseCommand(cmd, "delete", c)
+		},
+	}
+}
+
+func (c *CmdDeleteEmail) ParseArgv(ctx *cli.Context) error {
+	if len(ctx.Args()) != 1 {
+		return errors.New("delete requires one argument (email)")
```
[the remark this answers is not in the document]
heronhaye: seems prone to getting out of date tbh...doesn't the usage string display when there's an error?

go/service/emails.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (h *EmailsHandler) GetEmails(ctx context.Context, sessionID int) ([]keybase1.Email, error) {
+	var err error
+	mctx := libkb.NewMetaContext(ctx, h.G())
+	defer mctx.CTraceTimed("EmailsHandler#GetEmails", func() error { return err })()
+	return emails.GetEmails(mctx)
+}
+
+func (h *EmailsHandler) BulkLookupEmails(ctx context.Context, arg keybase1.BulkLookupEmailsArg) ([]keybase1.EmailLookupResult, error) {
+	var err error
```
zapu: `var err error` doesnt seem right, is this an linter mistake?

go/pvl/interp.go
```diff
@@ -1011,9 +1011,9 @@ func runSelectorJSONInner(m metaContext, state scriptState, selectedObject *json
 	logger := func(format string, args ...interface{}) {
 		debugWithState(m, state, format, args)
 	}
-	jsonResults, perr := jsonhelpers.AtSelectorPath(selectedObject, selectors, logger)
-	if perr != nil {
-		return nil, perr
+	jsonResults, perr := jsonhelpers.AtSelectorPath(selectedObject, selectors, logger, libkb.NewInvalidPVLSelectorError)
+	if perrInner, _ := perr.(libkb.ProofError); perrInner != nil {
```
zapu: im worried that if there's a `perr` that does not satisfy the if condition, then it will not branch here and instead continue in bad state

go/client/cmd_email_delete.go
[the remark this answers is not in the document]
heronhaye: what do you think about just saying "invalid args"

protocol/avdl/keybase1/common.avdl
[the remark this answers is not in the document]
heronhaye: what do you suggest? there's no reason they couldn't be used for other things as well. maybe `ServertrustProofIdentityVisibility`? kind of unwieldy

go/client/cmd_email_edit.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func NewCmdEditEmail(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	cmd := &CmdEditEmail{
+		Contextified: libkb.NewContextified(g),
+	}
+	return cli.Command{
+		Name:         "edit",
+		Usage:        "Edit a email",
+		ArgumentHelp: "<old email> <new email> <private|public (visibility for new email)>",
```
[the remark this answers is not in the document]
heronhaye: I was thinking about the edit flow in the designs, since there'll be a switch on that page if design decides to with a switch on every email/phone

protocol/avdl/keybase1/phone_numbers.avdl
heronhaye: yeah maybe weird...but the idea is client doesn't do any input validation and treats these as string wrappers. but since this is the only case where we expected a malformed number I could add a RawPhoneNumber type for this endpoint?

go/client/cmd_email_delete.go
```diff
@@ -34,7 +34,7 @@ func NewCmdDeleteEmail(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.C
 
 func (c *CmdDeleteEmail) ParseArgv(ctx *cli.Context) error {
 	if len(ctx.Args()) != 1 {
-		return errors.New("delete requires one argument (email)")
+		return errors.New("invalid number of arguments.")
```
zapu: thanks for changing this

protocol/avdl/keybase1/common.avdl
```diff
@@ -345,6 +346,7 @@ protocol Common {
     UnixTime ctime;
   }
 
+  // Used for searching by phone number or email
```
zapu: :+1: we can reorganize these types if something else comes along that needs different "Identity Visibility enum"

protocol/avdl/keybase1/phone_numbers.avdl
zapu: thx

PR #14913, opened 2018-12-03
reuse RegExp obj

songgao:


shared/common-adapters/markdown/shared.js
```diff
@@ -106,20 +106,25 @@ function parseMarkdown(
...
+  )
+}
+
+const _linkRegex = _makeLinkRegex()
+
 // TODO, when named groups are supported on mobile, we can use this instead
 // const linkRegex = /^( *)((https?:\/\/)?[\w-]+(?<tld>\.[\w-]+)+\.?(:\d+)?(\/\S*)?)\b/i
 // This copies the functionality of this named group
 // $FlowIssue treat this like a RegExp
 const linkRegex: RegExp = {
```
jzila: Is this typing correct? An executed regexp shouldn't return a regexp, right?

jzila: Oh nvmind I see what you're doing now.

PR #14916, opened 2018-12-04
protocol: add PrefetchProgress to stat/sync status

strib:
And print out progress in the `fs stat` and `fs sync show` commands.

Issue: KBFS-3575

### DIFF
--- a/go/client/cmd_simplefs_stat.go
+++ b/go/client/cmd_simplefs_stat.go
@@ -58,9 +58,12 @@ func NewCmdSimpleFSStat(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.
 }
 
 func prefetchStatusString(e keybase1.Dirent) string {
-	if e.PrefetchStatus != keybase1.PrefetchStatus_IN_PROGRESS ||
-		e.PrefetchProgress.BytesTotal == 0 {
+	if e.PrefetchStatus != keybase1.PrefetchStatus_IN_PROGRESS {
 		return e.PrefetchStatus.String()
+	}
+
+	if e.PrefetchProgress.BytesTotal == 0 {
+		return keybase1.PrefetchStatus_NOT_STARTED.String()
 	}
 
 	return fmt.Sprintf("%.2f%%",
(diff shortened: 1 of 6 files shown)


go/client/cmd_simplefs_stat.go
```diff
@@ -101,7 +113,7 @@ func (c *CmdSimpleFSStat) Run() (err error) {
 				r.Revision, keybase1.FormatTime(e.Time),
 				keybase1.DirentTypeRevMap[e.DirentType],
 				e.Size, e.Name, e.LastWriterUnverified.Username,
-				e.PrefetchStatus)
+				prefetchStatusString(e))
```
jakob223: I feel like this output has just gotten less and less human-intelligible. Do we intend for humans to read the output of this command, or just scripts?

go/client/cmd_simplefs_stat.go
```diff
@@ -56,6 +57,17 @@ func NewCmdSimpleFSStat(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.
 	}
 }
 
+func prefetchStatusString(e keybase1.Dirent) string {
+	if e.PrefetchStatus != keybase1.PrefetchStatus_IN_PROGRESS ||
```
jakob223: I think maybe it's confusing to say "sync in progress" only when there are 0/n bytes synced. Not sure if I have any better alternatives. "Sync waiting to start" or something?

go/client/cmd_simplefs_stat.go
strib: Yeah this output is a mess, I agree.  But we've never had a proper design for it.  I guess we just need headers or something, but that seems like the job for a future PR, if anyone cares enough to do it.

go/client/cmd_simplefs_stat.go
strib: This is actually when 0 is the denominator, that is we haven't even fetched a single block yet to look at it.  I'm changing this to `keybase1.PrefetchStatus_NOT_STARTED.String()` in that case.

PR #14924, opened 2018-12-04
use buildCounter to decide whether to accept built payments / requests

buoyad:
Adds a field `state.wallets.buildCounter` which increments whenever we call `buildPayment` or `buildRequest`. Pass counter to response and check against store to decide whether to accept the response. r? @keybase/react-hackers


shared/actions/wallets.js
```diff
@@ -83,6 +73,23 @@ const buildPayment = (
...
+const spawnBuildPayment = (
+  state: TypedState,
+  action:
+    | WalletsGen.SetBuildingAmountPayload
+    | WalletsGen.SetBuildingCurrencyPayload
+    | WalletsGen.SetBuildingFromPayload
+    | WalletsGen.SetBuildingIsRequestPayload
+    | WalletsGen.SetBuildingToPayload
+    | WalletsGen.DisplayCurrencyReceivedPayload
+) => {
```
buoyad: I funneled these actions into `wallets:buildPayment` so we wouldn't have to keep this list in sync with all the reducer cases that would need to increment the counter

shared/reducers/wallets.js
```diff
@@ -17,14 +17,16 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
       return state.merge({accountMap: accountMap})
     case WalletsGen.assetsReceived:
       return state.setIn(['assetsMap', action.payload.accountID], I.List(action.payload.assets))
+    case WalletsGen.buildPayment:
+      return state.set('buildCounter', state.buildCounter + 1)
```
buoyad: then `wallets:buildPayment` is the only thing that increments the counter

shared/actions/wallets.js
```diff
@@ -24,22 +24,8 @@ import {RPCError} from '../util/errors'
...
-    | WalletsGen.SetBuildingIsRequestPayload
-    | WalletsGen.SetBuildingToPayload
-    | WalletsGen.DisplayCurrencyReceivedPayload
-) => {
-  if (action.type === WalletsGen.displayCurrencyReceived && !action.payload.setBuildingCurrency) {
-    // didn't change state.building; no need to call build
-    return
-  }
-  return (state.wallets.building.isRequest
+const buildPayment = (state: TypedState, action: WalletsGen.BuildPaymentPayload) =>
```
buoyad: and the only thing that triggers the RPC here

PR #14928, opened 2018-12-04
tweak currency list spacing in mobile send

zanderz:
shows the last 2 currency items again.

Couldn't repro the cut-off text in the receive dialog.


shared/wallets/send-form/choose-asset/index.js
```diff
@@ -328,6 +328,9 @@ const styles = styleSheetCreate({
     isElectron: {
       maxHeight: 525 - 48,
     },
+    isMobile: {
+      paddingBottom: 100,
```
chrisnojima: i don't think this is likely the correct fix

zanderz: What do you have in mind?

chrisnojima: this kind of styling fix typically is fixing the symptom and not the actual problem. likely this list isn't setting its height correctly so its layout is wrong.

zanderz:
@keybase/react-hackers Changed to use flow values instead of some absolute percentage. Tried ios and android sims.

PR #14929, opened 2018-12-04
new thread stale type: convupdate

joshblum:
Add new `StaleUpdateType_CONVUPDATE` so we don't have to reload the entire thread cc @chrisnojima @mmaxim


chrisnojima:
@joshblum expected test failure?

joshblum:
likely a flake, restarting

PR #14937, opened 2018-12-05
Chat layout fixes

chrisnojima:
- [x] padding between lines 3
- [x] more space between avatar and hover box bottom
- [x] send indicator is wrapped and can render null, but still is in box2 so it'd make an extra gap when it goes away. its pos absolute anyways so just move it where orange line is
- [x] orange line fixes for mobile


chrisnojima:
@keybase/react-hackers @cecileboucheron updates some chat layout

PR #14939, opened 2018-12-05
make all mark as read calls check inbox cache for fully read

mmaxim:
cc @maxtaco


mmaxim:
Fixes https://github.com/keybase/keybase-issues/issues/1804

mmaxim:
(Note: the fix here is that calling `keybase chat read` on a fully read thread would still generate a remote call to the server)

PR #14943, opened 2018-12-05
React redux 6 / strict mode / small deps

chrisnojima:
This pr updates react redux bindings to 6.0 which allows us to start to use strict mode. This will tell us if we're using deprecated lifecycle events https://reactjs.org/docs/strict-mode.html.
There are some libs we use that still use these but they'll be fixed up after this PR

also updates babel/webpack etc

@keybase/react-hackers


chrisnojima:
@keybase/react-hackers bump

chrisnojima:
@keybase/react-hackers bump

PR #14946, opened 2018-12-05
scraper cache

joshblum:
caches the scraper and packager with a time based cache so we don't pump out stale values. prefetch mode up next so populate the cache during message compose

### DIFF
--- a/go/chat/unfurl/packager.go
+++ b/go/chat/unfurl/packager.go
@@ -23,6 +23,7 @@ import (
 type Packager struct {
 	utils.DebugLabeler
 
+	cache        *unfurlCache
 	ri           func() chat1.RemoteInterface
 	store        attachments.Store
 	s3signer     s3.Signer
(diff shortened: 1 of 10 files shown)


go/chat/unfurl/scraper.go
```diff
@@ -33,16 +35,26 @@ func (s *Scraper) makeCollector() *colly.Collector {
...
+		return res, err
+	}
+
 	var unfurlTyp chat1.UnfurlType
-	var domain string
 	if forceTyp != nil {
 		unfurlTyp = *forceTyp
 	} else {
-		var err error
-		if unfurlTyp, domain, err = ClassifyDomainFromURI(uri); err != nil {
```
joshblum: how would this work with `domain=""` previously?

go/chat/unfurl/packager.go
```diff
@@ -232,9 +234,24 @@ func (p *Packager) packageGiphy(ctx context.Context, uid gregor1.UID, convID cha
 	return chat1.NewUnfurlWithGiphy(g), nil
 }
 
+func (p *Packager) cacheKey(uid gregor1.UID, convID chat1.ConversationID, raw chat1.UnfurlRaw) string {
```
mmaxim: Maybe put type in here just to be safe?

joshblum:
approved via chat

PR #14948, opened 2018-12-05
new callPromise type in chat

chrisnojima:
This fixes major issues w/ losing typing when calling rpcs (and other yields in generators). 
See https://github.com/facebook/flow/issues/2613.

This is part 1 in a series of PRs 

old:
```
const val = yield Saga.call(RPCTypes.whatever, 1234) // val is any :(
```

new:
```
const val = yield * Saga.callPromise(RPCTypes.whatever, 1234) // val is typed :)
```

Saga.call will be deprecated when this is all done, replaced by Saga.callUntyped.
Often we don't care about this typing if we're just yielding to redux saga. This only applies if we want to  know the types of the result of a call.

@keybase/react-hackers everyone should be aware of this change

additionally I put back typed Saga.put since the perf issue seems to be resolved


shared/actions/config/index.js
```diff
@@ -260,7 +260,7 @@ const switchRouteDef = (
   }
 }
 
-const resetGlobalStore = () => Saga.put({payload: undefined, type: ConfigGen.resetStore})
+const resetGlobalStore = () => Saga.put({payload: null, type: 'common:resetStore'})
```
chrisnojima: allows this to type correctly

shared/util/typed-put.js.flow
```diff
@@ -1,25 +1,12 @@
...
-// declare var put: {
-// (action: TypedActions): PutEffect<TypedActions, null>,
-// (channel: Channel, action: TypedActions): PutEffect<TypedActions, Channel>,
-// resolve: {
-// (action: TypedActions): PutEffect<TypedActions, null>,
-// (channel: Channel, action: TypedActions): PutEffect<TypedActions, Channel>,
-// },
-// }
-
-// Uncomment the above to have a typed put. Current flow makes this VERY VERY slow. The merge phase timesout on several files.  This was super useful to find thunks but is likely not that useful now
```
chrisnojima: turns out flow works with this now. in my testing a flow check runs 2 seconds slower w/ this and increases coverage a lot

shared/util/saga.js
```diff
@@ -132,11 +132,18 @@ function safeTakeEveryPure<A, R, FinalEffect, FinalErrorEffect>(
   })
 }
 
+function* callPromise<Args, T>(fn: (...args: Args) => Promise<T>, ...args: Args): Generator<any, T, any> {
+  // $FlowIssue doesn't understand args will be an array
+  return yield Effects.call(fn, ...args)
+}
```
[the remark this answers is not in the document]
songgao: Oh nvm; it's to replace `call`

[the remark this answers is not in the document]
songgao: deleting these to avoid confusion ...

chrisnojima:
Coverage change:

Master: Covered: 94.14% (5593 of 5941 expressions)
PR: Covered: 97.11% (5776 of 5948 expressions)

shared/util/saga.js
```diff
@@ -132,11 +132,18 @@ function safeTakeEveryPure<A, R, FinalEffect, FinalErrorEffect>(
   })
 }
 
+function* callPromise<Args, T>(fn: (...args: Args) => Promise<T>, ...args: Args): Generator<any, T, any> {
```
buoyad: comment about when to use `callPromise` vs `callUntyped` would probably come in handy

chrisnojima: i added one in part 2

PR #14952, opened 2018-12-05
enter key in jump to chat

zanderz:
selects the first thing now, as if the down arrow key were pressed


songgao:
I tried this on Mac and it didn't seem to select the first one in result.

PR #14960, opened 2018-12-06
More saga call changed

chrisnojima:
- [x] Applies new callPromise/callUntyped everywhere and removes Saga.call entirely
- [x] Fixes newly revealed flow issues


shared/actions/fs/platform-specific.desktop.js
```diff
@@ -167,21 +167,24 @@ function* fuseStatusSaga(): Saga.SagaGenerator<any, any> {
...
+  let status = yield* Saga.callPromise(RPCTypes.installFuseStatusRpcPromise, {bundleVersion: ''})
   if (isWindows && status.installStatus !== RPCTypes.installInstallStatus.installed) {
     // Check if another Dokan we didn't install mounted the filesystem
-    const kbfsMount = yield Saga.call(RPCTypes.kbfsMountGetCurrentMountDirRpcPromise)
+    const kbfsMount = yield* Saga.callPromise(RPCTypes.kbfsMountGetCurrentMountDirRpcPromise)
     if (kbfsMount && fs.existsSync(kbfsMount)) {
-      status.installStatus = RPCTypes.installInstallStatus.installed
-      status.installAction = RPCTypes.installInstallAction.none
-      status.kextStarted = true
+      status = {
```
chrisnojima: to be extra careful we mark the return values here as readonly so if you override the same data you need to make your own copy. i didn't validate this logic at all
@jzila

shared/actions/fs/index.js
```diff
@@ -482,7 +482,7 @@ const extractMimeFromContentType = (contentType, disposition: string): Types.Mim
   return Constants.makeMime({displayPreview, mimeType})
 }
 
-const getMimeTypePromise = (localHTTPServerInfo: Types._LocalHTTPServer, path: Types.Path) =>
+const getMimeTypePromise = (localHTTPServerInfo: Types.LocalHTTPServer, path: Types.Path) =>
```
chrisnojima: these inner types shouldn't ever be used (only by the make functions in constants) and not leaked

shared/actions/fs/index.js
```diff
@@ -530,23 +530,28 @@ function* _loadMimeType(path: Types.Path, refreshTag?: Types.RefreshTag) {
     mimeTypeRefreshTags.set(refreshTag, path)
   }
 
-  const state = yield Saga.select()
-  let localHTTPServerInfo: Types._LocalHTTPServer =
-    state.fs.localHTTPServerInfo || Constants.makeLocalHTTPServer()
+  const state: TypedState = yield Saga.select()
+  let localHTTPServerInfo = state.fs.localHTTPServerInfo || Constants.makeLocalHTTPServer()
```
chrisnojima: before this was a mix of the actual record from the store, or the non-record value from the service call which makes the typing  way more complicated. instead lets just use the record

chrisnojima:
cc: @jzila for FS specific changes

chrisnojima:
cc: @jzila for FS specific changes

chrisnojima:
cc: @jzila for FS specific changes

chrisnojima:
cc: @jzila for FS specific changes

chrisnojima:
@keybase/react-hackers bump

PR #14962, opened 2018-12-06
Typed select

chrisnojima:
@keybase/react-hackers this deprecates Saga.select() and replaces it with `yield * Saga.selectState()`


chrisnojima:
@keybase/react-hackers bump

PR #14966, opened 2018-12-06
Action explicit shape generation

chrisnojima:
Unclear why flow is doing this. We were generating the type by using $Call<> but this seems to confuse flow and cause us to lose types. 
When you deal with an action handler with multiple action types and you refine using a switch statement (pretty common) the action.type was devolving into string and not 'actiona' | 'actionb' etc. Changing this to be a static string and not the result of a call seems to fix this for some reason

coverage of actions/chat2/index (just as an example)
old coverage: Covered: 97.19% (5780 of 5947 expressions)
new coverage: Covered: 97.21% (5785 of 5951 expressions)


chrisnojima:
@keybase/react-hackers bump

PR #14969, opened 2018-12-06
vendor triplesec 4

heronhaye:


go/libkb/skb.go
```diff
@@ -237,11 +237,11 @@ func (s *SKB) UnlockSecretKey(m MetaContext, passphrase string, tsec Triplesec,
...
 	var unlocked []byte
 
-	switch s.Priv.Encryption {
-	case 0:
+	switch {
+	case s.Priv.Encryption == 0:
 		m.CDebugf("case: Unlocked")
 		unlocked = s.Priv.Data
-	case int(triplesec.Version):
+	case s.Priv.Encryption > 0 && s.Priv.Encryption < LKSecVersion:
```
heronhaye: Not really sure how to do this. This change reserves 1 through 99 for triplesec versions, and should be OK because newer versions of triplesec can decrypt older versions as well. Not sure what happens if the default ClientTriplesecVersion is 4, but the user provides a version 3 ciphertext. This would cache a version 4 stream. Is that okay as long as the client can decrypt that later?

go/libkb/passphrase_stream.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func NewSecureTriplesec(passphrase []byte, salt []byte) (Triplesec, error) {
-	return triplesec.NewCipher(passphrase, salt)
+	return triplesec.NewCipher(passphrase, salt, ClientTriplesecVersion)
```
heronhaye: Seems to be the only place where we create a new Cipher object

go/libkb/constants.go
```diff
@@ -657,3 +657,5 @@ const MaxTeamMembersForPairwiseMAC = 100
 
 const MaxStellarPaymentNoteLength = 500
 const MaxStellarPaymentBoxedNoteLength = 1000
+
+const ClientTriplesecVersion = 3
```
heronhaye: still on 3 for now

heronhaye: The problem I'm seeing is if we ever update this to 4, even if the keystretching doesn't change, we'll be setting `pwh_version=4` in the passphrase metadata, which seems like a more complicated change. Is the answer to have a `ClientTriplesecKeystretchVersion` that will continue to be 3, and two `TripleSec` instances in `g`, one for key stretching and one for encrypting/decrypting SKBs?

PR #14972, opened 2018-12-06
Prefetch unfurls while typing

joshblum:
depends on https://github.com/keybase/client/pull/14946 for caching changes. will rebase once that's in. patch does the following:
- adds a `Prefetch` method to the `Unfurler` to cache the unfurl results for whitelisted domains
- adds `Text` field to `UpdateTyping` rpc so we can try to prefetch unfurls that are whitelisted while the message is composed

cc @keybase/react-hackers

### DIFF
--- a/go/chat/unfurl/unfurler.go
+++ b/go/chat/unfurl/unfurler.go
@@ -391,24 +391,24 @@ func (u *Unfurler) unfurl(ctx context.Context, outboxID chat1.OutboxID) {
 		task, err := u.getTask(ctx, outboxID)
 		if err != nil {
 			u.Debug(ctx, "unfurl: failed to get task: %s", err)
-			return unfurl, err
+			return nil, err
 		}
 		if err := u.setStatus(ctx, outboxID, types.UnfurlerTaskStatusUnfurling); err != nil {
 			u.Debug(ctx, "unfurl: failed to set status: %s", err)
 		}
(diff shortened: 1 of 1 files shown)


joshblum:
one downside is we might try to process these unfurls multiple times. Once they are in the scraper/package caches it's fine but we might call out a few times if we get multiple typing updates before the unfurl completes and the caches are populated. we could add a cache to prefetch so we don't retry things if we think it's an issue

joshblum:
Added `HiddentString`, what other precautions are there to take to prevent logging of this RPC or hiding the `text` field which contains plaintext now?

mmaxim:
Don't see how it could, it is just using an action that is already being dispatched.

go/chat/unfurl/utils.go
```diff
@@ -58,10 +58,3 @@ func ClassifyDomain(domain string) chat1.UnfurlType {
 	}
 	return chat1.UnfurlType_GENERIC
 }
-
-func ClassifyDomainFromURI(uri string) (typ chat1.UnfurlType, domain string, err error) {
```
mmaxim: Just keep this, it might be useful again

chrisnojima:
Please don't merge until my review. Yes all actions with secrets must dispatch with hidden string. It was a bool before

joshblum:
@chrisnojima no prob we can discuss in the morning. the unfurl work is async on that rpc handler

PR #14974, opened 2018-12-06
put "available" in send confirmation from field

zanderz:


shared/search/user-input/container.js
```diff
@@ -222,10 +222,11 @@ const ConnectedUserInput = compose(
         this.props.onFocusInput()
       }
 
-      if (this.props.searchResultIds !== prevProps.searchResultIds && !this.props.showingSearchSuggestions) {
-        this.props.onUpdateSelectedSearchResult(
-          (this.props.searchResultIds && this.props.searchResultIds[0]) || null
-        )
+      const prevTopResult = prevProps.searchResultIds && prevProps.searchResultIds[0]
```
akalin-keybase: what's this for?

shared/wallets/confirm-form/participants/index.js
```diff
@@ -76,7 +76,7 @@ const Participants = (props: ParticipantsProps) => {
     <Kb.Box2 direction="vertical" fullWidth={true}>
       <ParticipantsRow heading="From">
         <AccountEntry
-          contents={props.fromAccountAssets}
```
akalin-keybase: i guess a previous change already removed the '+ more'?

shared/search/user-input/container.js
zanderz: whoops, from a different pr, sorry

shared/wallets/confirm-form/participants/index.js
zanderz: yeah I only saw it in some storybook text

zanderz:
@keybase/react-hackers ready for 1 more look

shared/wallets/confirm-form/participants/index.js
```diff
@@ -76,7 +76,7 @@ const Participants = (props: ParticipantsProps) => {
...
+          contents={props.fromAccountAssets + ' available'}
```
buoyad: hm, this seems like a string that needs to come from core. since the '+ more' is coming from there too, here's what it looks like for me:
image

buoyad: and because of this line in the ticket
> If I were to send another asset, say BTC/Stronghold.com, it would say "35 BTC/Stronghold.com available".

zanderz: ok thanks

zanderz:
The service was already adding `+ more` if there were other currencies available in the account besides XLM. When exactly should it say `available` instead of `+ more`? I can't find a definitive Zeplin page or anything.

buoyad:
@zanderz I don't think this relates to that field on the account, it's specific to the transaction. The ticket has some more info, I think the rules are if you're sending XLM, say "12.345 XLM available"; if you're sending another asset (not supported yet), say "123.45 OTHERASSET available"

zanderz:
oh whoops, you did say BuildPaymentResLocal and I missed that someow - just going by the current source of that field. I'll change it.

zanderz:
Thanks Danny! I tested a build and saw "available" in the UI

PR #14975, opened 2018-12-06
refresh wallet account details on display currency change

zanderz:


zanderz:
not sure what that might mean except to refresh instead of selecting?

zanderz:
ah ok I think I see

shared/actions/wallets.js
```diff
@@ -778,6 +784,7 @@ function* walletsSaga(): Saga.SagaGenerator<any, any> {
   yield Saga.actionToPromise(WalletsGen.loadDisplayCurrencies, loadDisplayCurrencies)
   yield Saga.actionToPromise(WalletsGen.loadSendAssetChoices, loadSendAssetChoices)
   yield Saga.actionToPromise(WalletsGen.loadDisplayCurrency, loadDisplayCurrency)
+  yield Saga.actionToAction(WalletsGen.loadDisplayCurrency, refreshAssets)
```
chrisnojima: we want 
DisplayCurrenciesReceived and not loadDisplayCurrency

zanderz: ok thanks

PR #14977, opened 2018-12-07
Remove compression from gregor1.outgoing

joshblum:
We need to let clients bake several release cycles with support for compression before enabling this on the server


joshblum:
in person approval, merging over unrelated ci flake

PR #14979, opened 2018-12-07
add prereq to pgp import key for standardized and better error message

xgess:
`keybase pgp gen --multi` from a logged out device and from a deprovisioned device give weird error messages. adding in this prereq (like most of the other pgp engines) standardizes the error message with something a little more helpful: "Login required."


xgess:
@maxtaco maybe peep this again? i had to change the fix because I missed a case with website signups. tricky tricky, but i'm glad there was a test for it.

PR #14980, opened 2018-12-07
on deprovision ensure we ignore the just revoked name

chrisnojima:
@keybase/react-hackers 
to test just deprovision yourself. before your username would be in the dropdown and you'd have to change it, now it should select another name or be blank


chrisnojima:
@keybase/react-hackers bump open 4 days

PR #14982, opened 2018-12-07
Add server-provided Timebounds to Stellar transactions

zapu:


zapu:
@patrickxb ready for first look

go/stellar/stellar.go
```diff
@@ -951,16 +968,11 @@ func claimPaymentWithDetail(ctx context.Context, g *libkb.GlobalContext, walletS
...
-	// Throw a random ID into the transaction memo so that we get a new txID each time.
-	// This makes it easy to resubmit without hitting a txID collision on the server.
-	memoID, err := randMemoID()
-	if err != nil {
-		return res, err
-	}
-	g.Log.CDebugf(ctx, "using randomized memo ID: %v", memoID)
+	mctx := libkb.NewMetaContext(ctx, g)
+	sp := NewSeqnoProvider(mctx, walletState)
+	tb := getTimeboundsForSending(mctx)
```
zapu: looks like we can achieve the same thing using timebounds, that are signed into the transaction and change txid. Unless we plan to retry these more than one time in given second

mlsteele: Yes I like this plan of getting rid of the randomized memo because timebounds also cause a different txID each time. I don't think the 1 second resolution should be a problem in this case.

patrickxb:
Is there a stellarnet PR that I missed?

zapu:
https://github.com/keybase/stellarnet/pull/38 sorry i forgot to tag

patrickxb:
I wonder if this could be a problem if a user's clock is off.  Maybe it should get the timebounds from the server so we can be in control?

zapu:
I was thinking that too. Good idea about asking the server

go/stellar/stellar.go
```diff
@@ -438,6 +438,43 @@ type DisplayBalance struct {
...
+	// not included before the deadline.
+
+	// We ask server for timebounds because local clock might not be accurate,
+	// and typically we will be setting timeout as 30 seconds.
+	start := time.Now()
+	serverTimes, err := walletState.ServerTimeboundsRecommendation(m.Ctx())
+	if err != nil {
+		return nil, err
+	}
+	took := time.Now().Sub(start)
```
patrickxb: took := time.Since(start)

go/stellar/stellar.go
```diff
@@ -438,6 +438,43 @@ type DisplayBalance struct {
...
+		return nil, nil
+	}
+
+	// Offset server time by our latency to the server. We are making two
+	// requests to submit a transaction: one here to get the server time, and
+	// another one to send the signed transaction. Assuming server roundtrip
+	// time will be the same for both requests, we can offset timebounds here
+	// by entire roundtrip time and then we will have MaxTime set as 30 seconds
+	// counting from when the server gets our signed tx.
+	deadline := serverTimes.TimeNow.UnixSeconds() + serverTimes.Timeout + int64(took.Seconds())
```
patrickxb: it probably doesn't matter too much, but you could make this more precise:

    deadline := serverTimes.TimeNow.Add(took).UnixSeconds() + serverTimes.Timeout

PR #14990, opened 2018-12-07
memory lru implicit team lookup

joshblum:


go/teams/implicit.go
```diff
@@ -63,6 +89,20 @@ func LookupImplicitTeamIDUntrusted(ctx context.Context, g *libkb.GlobalContext,
 }
 
 func loadImpteamFromServer(ctx context.Context, g *libkb.GlobalContext, displayName string, public bool) (imp implicitTeam, err error) {
+	cacheKey := implicitTeamCacheKey{displayName: displayName, public: public}
+	cacher := g.GetImplicitTeamCacher()
+	if cacher != nil {
```
joshblum: engine can't call `NewImplicitTeamCacheAndInstall` because of a circular dependency unfortunately but chat tests do set this cache up

mmaxim:
What is the status here? I would like to try this patch in master.

mmaxim:
Why would this get cleared on logout?

mmaxim:
Oh I see, I guess it is not keyed on the user at all.

mlsteele:
Yeah it's private info. I don't know if there's other similar on disk. I would want it gone on logout.

mmaxim:
To make it easier, we could probably get basically all the perf gain by just making this an in-memory cache, and just blow it away on logout.

joshblum:
in memory wouldn't help mobile that much be we can see if there is any noticeable difference. was thrown off since we cache the conflict info on disk unencrypted but it's only a generation/timestamp

joshblum:
updated @mlsteele, ptal!

PR #14991, opened 2018-12-07
Add more ICS parser checks

patrickxb:
Just a few more checks


patrickxb:
No, I was just checking some formats parsed correctly.

PR #14995, opened 2018-12-08
version lock jest-emotion

heronhaye:
@keybase/react-hackers 

Generated with 
```
rm -r shared/node_modules
yarn install
```

Not sure if anything else is needed.


heronhaye:
I did that and got the same diff

PR #15001, opened 2018-12-09
send-link-to-chat 1: components and storyshots

songgao:
I haven't done anything for mobile yet; will do it in a follow-up PR.


songgao:
Hmm I'm realizing mobile is actually a lot different from desktop design. So will do that in a separate ticket, maybe after attachment flow is done too.

shared/fs/common/send-in-app-action.js
```diff
@@ -0,0 +1,38 @@
...
+import * as Styles from '../../styles'
+import {namedConnect} from '../../util/container'
+
+type OwnProps = {
+  path: Types.Path,
+  routePath?: ?I.List<string>,
+  sendIconClassName: string,
+}
+
+const mapDispatchToProps = (dispatch, {path, routePath}: OwnProps) => ({
```
chrisnojima: why is this connected

shared/fs/send-link-to-chat/index.js
```diff
@@ -0,0 +1,151 @@
...
+    case 'person':
+      return props.conversation.name
+    case 'group':
+      return 'your group'
+    case 'small-team':
+    case 'big-team':
+      return 'team members'
+    case 'none':
+      return ''
+    default:
```
chrisnojima: See util/flow

shared/fs/common/send-in-app-action.js
songgao: It needs to dispatch an action. This is populated in the next PR https://github.com/keybase/client/pull/15002

songgao:
@chrisnojima PTAL thanks!

PR #15002, opened 2018-12-09
send-link-to-chat 2: redux and saga

songgao:
Depends on https://github.com/keybase/client/pull/15001

### DIFF
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -2,8 +2,12 @@
 import * as Constants from '../../constants/fs'
 import * as ConfigGen from '../config-gen'
 import * as FsGen from '../fs-gen'
+import * as TeamsGen from '../teams-gen'
+import * as TeamsConstants from '../../constants/teams'
 import * as I from 'immutable'
 import * as RPCTypes from '../../constants/types/rpc-gen'
+import * as RPCChatTypes from '../../constants/types/rpc-chat-gen'
+import * as ChatTypes from '../../constants/types/chat2'
 import * as Saga from '../../util/saga'
 import * as Flow from '../../util/flow'
 import * as SettingsConstants from '../../constants/settings'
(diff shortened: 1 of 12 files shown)


shared/actions/fs/index.js
```diff
@@ -783,6 +787,79 @@ const cancelMoveOrCopy = isMobile
...
+          })
+        )
+      })
+    )
+  } else {
+    // It's a real team, but we don't know if it's a small team or big team. So
+    // fire TeamsGen.getChannels to find out. We'll listen for
+    // TeamsGen.setTeamChannels and see if we should automatically set convID.
+    // We might already have that, but just refresh anyway to get latest info.
+    actions.push(Saga.put(TeamsGen.createGetChannels({teamname: elems[2]})))
```
chrisnojima: i worry about this type of cross store logic. we dispatch some foreign action and then reach into that store to get the value. why not just make the rpcs calls you need and have it be self contained

shared/fs/send-link-to-chat/container.js
```diff
@@ -1 +1,138 @@
...
+  _username: state.config.username,
+  sendLinkToChat: state.fs.sendLinkToChat,
+})
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  _selectChannel: (convID: ChatTypes.ConversationIDKey) =>
+    dispatch(FsGen.createSetSendLinkToChatConvID({convID})),
+  _send: (conversationIDKey: ChatTypes.ConversationIDKey, text: string) => {
+    dispatch(ChatGen.createMessageSend({conversationIDKey, text: new HiddenString(text)}))
+    dispatch(putActionIfOnPath(ownProps.routePath, navigateUp()))
```
chrisnojima: why navigate up when you're just navigating to the thread after

songgao: This way if user clicks the Files tab from the Chat tab, they'd see the folder view. Otherwise the SendLinkToChat modal would still be shown. Not showing it anymore in this case made more sense to me, but please let me know if you think otherwise.

songgao:
@chrisnojima This is ready again. PTAL thanks!

shared/actions/teams.js
```diff
@@ -793,6 +793,8 @@ function _afterGetChannels(fromGetChannels: any[]) {
   })
 
   return Saga.all([
+    // NOTE: FsGen.showSendLinkToChat depends on this action. If we ever change
```
chrisnojima: still true?

PR #15005, opened 2018-12-10
fix unbox logic in inbox

mmaxim:
Previous code would fire unbox every time `componentDidUpdate` ran, but this now only does it on the first load or, if the inbox gets expanded.

### DIFF
--- a/shared/chat/inbox/container/index.js
+++ b/shared/chat/inbox/container/index.js
@@ -7,7 +7,7 @@ import * as Chat2Gen from '../../../actions/chat2-gen'
 import * as RouteTreeGen from '../../../actions/route-tree-gen'
 import * as Inbox from '..'
 import {namedConnect} from '../../../util/container'
-import type {Props as _Props, RowItemSmall, RowItemBig} from '../index.types'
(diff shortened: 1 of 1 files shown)


chrisnojima:
And if we do that we should use the const

mmaxim:
I looked around for one but couldn't find it, although I didn't look very hard since the old stuff wasn't using one. Are you sure there is one, or is it to be added?

chrisnojima:
smallTeamsCollapsedMaxShown Needs to be exported. But honestly likely better for this component not to have logic like this

mmaxim:
K I thought this would just be an improvement in the short term, but we can just skip it if you want.

chrisnojima:
I already have a ticket for this. You can make the fix but definitely look into the no big teams case

mmaxim:
The change isn't in play for that case, it will either be always expanded (> 5 convs), or never expanded (<= 5 convs), in which case `if` is just controlled by `firstTimeLoading`

mmaxim:
K I agree, I was surprised to find this code here and figured it was just old.

chrisnojima:
I say we merge and hopefully toss all this logic and move it to core anyways

PR #15010, opened 2018-12-10
Cleanup make lint target

patrickxb:
Still passes without these filters.


go/Makefile
```diff
@@ -25,9 +25,6 @@ lint:
 		| grep -v stutters \
 		| grep -v "error strings should not be capitalized" \
 		| grep -v prot.go \
-		| grep -v "redundant if ...; err != nil check, just return error instead" \
```
mlsteele: this one is really mean

patrickxb: i don't think that's in golint anymore...

PR #15014, opened 2018-12-10
Inbox scrolling perf

chrisnojima:
This PR addresses some bad perf when scrolling your inbox (aka unboxing -> thrashing). 
There is a larger fundamental problem where the metaMap is used to represent the inbox, and that representation is used in several places where we re-sort and do a bunch of work (even when most of the list hasn't mutated. I'm going to change how this works in the near future. 

Instead what i've done is plumb through the inboxVersion which should act as a proxy for larger updates (let us differentiate between unboxing and getting new messages)

The rest is a mishmosh of random things I saw while profiling:

- Unfurl list mapStateToProps thrashing
- Unfurl items using connect but not really needing the store
- isLoading prop render thrashing the inbox list
- Don't serialize duplicate avatar/following/followers lists
- Don't serialize duplicate broken userinfos 
- Don't generate empty user infos for unbroken users


shared/actions/chat2/index.js
```diff
@@ -549,7 +549,7 @@ const onChatIdentifyUpdate = update => {
     }
   })
 
-  return [UsersGen.createUpdateBrokenState({newlyBroken, newlyFixed})]
+  return Saga.put(UsersGen.createUpdateBrokenState({newlyBroken, newlyFixed}))
```
chrisnojima: don't bother with an array of one

shared/chat/inbox/container/normal.js
```diff
@@ -54,25 +54,28 @@ const sortByTeamChannel = (a, b) =>
...
+        type: 'big',
+      })
 
-    return arr
-  }, [])
-}, shallowEqual)
+      return arr
+    }, [])
+  },
+  (a, b) => shallowEqual(a, b)
```
chrisnojima: next version of memoize-one has params (a,b,idx) so don't just pass this through, extract a,b

shared/chat/inbox/container/remote.js
```diff
@@ -48,11 +48,26 @@ const valuesCached = memoize3(
       .toArray()
 )
 
+// we only care to rerender if the inboxversion increases so
+// given the same inbox version give back the same metaMap
+const inboxVersionToMetaMap = memoize2(
```
chrisnojima: we plumb through inboxVersion now so we can use that to trigger things smartly

shared/chat/inbox/container/remote.js
```diff
@@ -48,11 +48,26 @@ const valuesCached = memoize3(
...
+  (inboxVersion, metaMap) => metaMap,
+  (a, b) => {
+    // custom equality, if inboxVersion is the same, treat it as the same
+    if (typeof a === 'number') {
```
chrisnojima: this is lame but the next version of memoize-one will pass an idx so you can differentiate your params

shared/desktop/remote/sync-avatar-props.desktop.js
```diff
@@ -101,16 +101,23 @@ function SyncAvatarProps(ComposedComponent: any) {
...
+      getRemoteFollowing(state.config.following, ownProps.usernames)
+    )
 
   const getRemoteAvatars = memoize2((avatars, usernames) => avatars.filter((_, name) => usernames.has(name)))
   const getRemoteFollowers = memoize2((followers, usernames) => followers.intersect(usernames))
   const getRemoteFollowing = memoize2((following, usernames) => following.intersect(usernames))
 
+  // use an immutable equals to not rerender if its the same
+  const immutableCached = memoize3(
+    (avatars, followers, following) => ({avatars, followers, following}),
```
chrisnojima: don't thrash if these are immutable equal

shared/desktop/remote/sync-props.desktop.js
```diff
@@ -31,6 +32,7 @@ function SyncPropsFactory(serializer: Serializer) {
       _sendProps = () => {
         if (this.props.remoteWindow) {
           try {
+            measureStart('remoteProps')
```
chrisnojima: add user timings here

shared/menubar/remote-proxy.desktop.js
```diff
@@ -71,7 +67,7 @@ const mergeProps = stateProps => {
     following: stateProps._following,
     loggedIn: stateProps.loggedIn,
     outOfDate: stateProps.outOfDate,
-    userInfo: getBrokenSubset(stateProps.userInfo),
```
chrisnojima: handle the conversion in serialize as it should be

shared/menubar/remote-proxy.desktop.js
```diff
@@ -87,8 +83,9 @@ const RenderExternalWindowBranch = (ComposedComponent: React.ComponentType<any>)
   }
 
 // Actions are handled by remote-container
-export default connect<Props | {}, _, _, _, _>(
+export default namedConnect<Props | {}, _, _, _, _>(
   mapStateToProps,
   () => ({}),
-  mergeProps
+  mergeProps,
+  'MenubarRemoteProxy'
```
chrisnojima: just so its more clear in the react tree what this is, else its just "PureComponent"

shared/menubar/remote-serializer.desktop.js
```diff
@@ -47,7 +47,12 @@ export const serialize: any = {
   loggedIn: v => v,
   outOfDate: v => v,
   totalSyncingBytes: v => v,
-  userInfo: v => v,
+  // Just send broken over, if its the same send null
+  userInfo: (v, o) => {
```
chrisnojima: don't send duplicate stuff over all the time

shared/chat/inbox/container/index.js
```diff
@@ -19,15 +19,27 @@ type OwnProps = {|
...
-  filter: state.chat2.inboxFilter,
-  isLoading: Constants.anyChatWaitingKeys(state),
-  neverLoaded: !state.chat2.inboxHasLoaded,
-})
+const mapStateToProps = state => {
+  const metaMap = state.chat2.metaMap
+  const filter = state.chat2.inboxFilter
+  const username = state.config.username
+  const inboxVersion = state.chat2.inboxVersion
+  const {allowShowFloatingButton, rows, smallTeamsExpanded} = filter
```
chrisnojima: call the memoized function up here so we don't call merge props all the time

shared/chat/inbox/container/index.js
```diff
@@ -66,21 +78,19 @@ const mapDispatchToProps = (dispatch, {navigateAppend}) => ({
 
 // This merge props is not spreading on purpose so we never have any random props that might mutate and force a re-render
 const mergeProps = (stateProps, dispatchProps, ownProps) => {
-  const {allowShowFloatingButton, rows, smallTeamsExpanded} = stateProps.filter
-    ? filteredRowData(stateProps._metaMap, stateProps.filter, stateProps._username)
-    : normalRowData(stateProps._metaMap, stateProps._smallTeamsExpanded)
   return {
-    _isLoading: stateProps.isLoading,
```
chrisnojima: isLoading changes all the time and we don't actually care if its ever false, so encode as canRefreshOnMount doesn't flip back and forth

chrisnojima:
cc: @mmaxim

shared/reducers/users.js
```diff
@@ -32,7 +32,8 @@ const reducer = (
       return state.update('infoMap', map =>
         map.withMutations(m => {
           newlyFixed.forEach(user => {
-            m.update(user, info => (info || blankUserInfo).set('broken', false))
+            // only make it if one exists already
```
buoyad: because we only care if the user is broken?

chrisnojima: else we get these messages about 'fixed' people who we never thought were broken, so we mutate the store a ton while this is essentially a broken only list (if you're not in here we don't consider you broken)

PR #15016, opened 2018-12-10
WIP:  remove race engine saga

chrisnojima:
When mixing incomingCallMap / customIncomingCallMap the response object gets called in two different ways. This PR makes them always be called on the other side of the emitter so ordering is preserved


chrisnojima:
its possible this is possible but i'm goingto ignore this for now

PR #15020, opened 2018-12-11
routing fixes for Files

songgao:
Back button for "open in Files tab":

1. For files, we have a back button on both desktop and mobile. This PR fixes the routes so that the back button goes to the parent folder instead of /keybase as before.
2. For folders, it's a back button on mobile. This PR fixes the route by populating all its parent paths, so user can click back button folder by folder all the way to /keybase, instead of directly to /keybase.

Back button at the root of Files on mobile (previous/abandoned attempt [here](https://github.com/keybase/client/pull/15021)):

This was due to a routing hack that I did earlier where moveOrCopy was a sibling of the folder route rather than children. This PR removes the hack and just make it the children of the folder route. The hack was to solve a specific case where when user taps "Close" in moveOrCopy they'd need to be at the original folder where they initiated move or copy. So in this PR, when user closes, we just grab the path and route from the routeTree, and use `navigateTo`.

### DIFF
--- a/shared/actions/fs/index.js
+++ b/shared/actions/fs/index.js
@@ -606,27 +606,47 @@ const commitEdit = (state: TypedState, action: FsGen.CommitEditPayload) => {
   }
 }
 
-const _getRouteChangeForOpenMobile = (
-  action: FsGen.OpenPathItemPayload | FsGen.OpenPathInFilesTabPayload,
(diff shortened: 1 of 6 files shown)


songgao:
untagging review for now as I'm gonna capture other route fixes in this branch too.

PR #15021, opened 2018-12-11
fix back button for root Files view on mobile

songgao:
The bug was that, on mobile, the `folderRoute` is not a direct children of the SettingsTab anymore. Instead, there's a real `fsRoute` now, whose default selected route is `folderRoute`. So one `navigateUp()` doesn't work here. However, as we want to use `putActionIfOnPath`, it's not easy to put `navigateUp()` twice. So instead just add a `navigateUp2` action.

### DIFF
--- a/shared/fs/header/container.js
+++ b/shared/fs/header/container.js
@@ -3,7 +3,7 @@ import * as I from 'immutable'
 import * as Types from '../../constants/types/fs'
 import * as Chat2Gen from '../../actions/chat2-gen'
 import * as Util from '../../util/kbfs'
-import {putActionIfOnPath, navigateUp} from '../../actions/route-tree'
+import {putActionIfOnPath, navigateUp2} from '../../actions/route-tree'
 import {namedConnect} from '../../util/container'
 import FolderHeader from './header'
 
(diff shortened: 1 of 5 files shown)


chrisnojima:
Let's not do either of these. Either nav directly where you want to go (navigateTo) or change you're routing hierarchy

songgao:
OK; also I just realized this PR has a new bug where it doesn't do it properly for non-/keybase paths. Will push new commits soon to use `navigateTo`

shared/fs/header/container.js
```diff
@@ -22,7 +25,10 @@ const mapDispatchToProps = (dispatch, {path, routePath}: OwnProps) => ({
         ...Util.tlfToParticipantsOrTeamname(Types.pathToString(path)),
       })
     ),
-  onBack: () => dispatch(putActionIfOnPath(routePath, navigateUp())),
+  onBack:
+    isMobile && path === Constants.defaultPath
+      ? () => dispatch(putActionIfOnPath(routePath, navigateTo([TabsConstants.settingsTab])))
```
chrisnojima: why not just fix the route so navigateUp works

songgao: I could figure out a way to make switch to work when user taps “Cancel” in the destination picker (for moveOrCopy), so just went with having folders route and destination picker route as siblings, which means they are both Uber fsRoute and that the fsRoute itself doesn’t render a component. See notes in https://github.com/keybase/client/blob/master/shared/fs/routes.js

Please let me know if there are better route fixes!

songgao: oops s/could/couldn't

songgao:
Talked on Meet with Nojima; we are gonna change the route design so that DestinationPicker and Folder view share the same routes. Will close this now and open a new one when it's ready.

PR #15023, opened 2018-12-11
Render ExplodingMeta in the case where we're exploding the message

cjb:
@keybase/react-hackers, especially @chrisnojima @buoyad 

The 'boom' wasn't rendering during explosion because it's rendererd by the ExplodingMeta component, and that was being unmounted at the moment of explosion because it's inside a `!exploded` check.

I think this is the right fix, adding ExplodingMeta to the path that's rendered during explosion, but please give it a try and check it looks right to you.

### DIFF
--- a/shared/chat/conversation/messages/wrapper/index.js
+++ b/shared/chat/conversation/messages/wrapper/index.js
@@ -254,25 +254,23 @@ class _WrapperMessage extends React.Component<Props & Kb.OverlayParentProps, Sta
   }
 
   _cachedMenuStyles = {}
-  _menuAreaStyle = () => {
-    // $ForceType
-    const exploding = this.props.message.exploding
+  _menuAreaStyle = (exploded, exploding) => {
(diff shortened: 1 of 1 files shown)


shared/chat/conversation/messages/wrapper/index.js
```diff
@@ -419,6 +419,13 @@ class _WrapperMessage extends React.Component<Props & Kb.OverlayParentProps, Sta
       return (
         <Kb.Box2 key="messageAndButtons" direction="horizontal" fullWidth={true}>
           {maybeExplodedChild}
+          <Kb.Box2 direction="horizontal" style={this._menuAreaStyle()}>
```
chrisnojima: i'm fixing up some styling here

cjb:
@chrisnojima Want to thumb this if it's ready? And either fix or ticket the "boom" being too high vertically in the case where there's a timestamp attached to the message?

PR #15024, opened 2018-12-11
dont upgrade kbfs if not logged in

mmaxim:


maxtaco:
hmm does it start up again when the user logs in?

mmaxim:
Will go every minute.

PR #15030, opened 2018-12-11
Mobile only checkbox in GUI

zapu:
Add a checkbox to enable/disable mobile only mode for an account.
screenshot_2018-12-11_20-18-25
screenshot_2018-12-11_20-19-20


shared/constants/wallets.js
```diff
@@ -595,6 +596,11 @@ const rootWalletPath = [...rootWalletTab, ...(isMobile ? [SettingsConstants.wall
 const isLookingAtWallet = (routeState: ?RouteStateNode) =>
   getPath(routeState, rootWalletTab).get(isMobile ? 2 : 1) === 'wallet'
 
+const loadMobileOnlyWaitingKey = (id: Types.AccountID) => `wallets:getMobileOnly:${id}`
```
chrisnojima: we don't need these waiting keys as no UI is even hooked up to this. also if we do want this (likely) there should just be done for loading settings

shared/constants/wallets.js
```diff
@@ -595,6 +596,11 @@ const rootWalletPath = [...rootWalletTab, ...(isMobile ? [SettingsConstants.wall
...
+const changeMobileOnlyWaitingKey = (id: Types.AccountID) => `wallets:setMobileOnly:${id}`
+const getMobileOnly = (state: TypedState, accountID: Types.AccountID) =>
```
chrisnojima: likely don't need this either

go/stellar/remote/remote.go
```diff
@@ -1115,15 +1115,19 @@ func IsAccountMobileOnly(ctx context.Context, g *libkb.GlobalContext, accountID
...
 		return err
 	}
 	if version == stellar1.BundleVersion_V1 {
 		return fmt.Errorf("mobile-only feature requires migration to v2 bundles")
 	}
-	err = acctbundle.MakeMobileOnly(bundle, accountID)
+	if enabled {
+		err = acctbundle.MakeMobileOnly(bundle, accountID)
+	} else {
+		err = acctbundle.MakeAllDevices(bundle, accountID)
```
xgess: a few lines below here (in this file) is a method `MakeAccountAllDevices` which was intended to be the entry point to this behavior. can we get rid of it now? it might also be less work in the tests if you push this branching logic up a level to the caller. and then the caller can either come down to here or to `MakeAccountAllDevices`.

shared/reducers/wallets.js
```diff
@@ -263,6 +263,10 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
       return state.merge({
         acceptingDisclaimerDelay: true,
       })
+    case WalletsGen.loadedMobileOnlyMode:
+      return state.withMutations(stateMutable => {
```
chrisnojima: can just be
`return state.update('mobileOnlyMap', map => map.set(action.payload.accountID, action.payload.enabled))`
or
`return state.updateIn(['mobileOnlyMap', action.payload.accountID], action.payload.enabled)`

shared/wallets/wallet/settings/container.js
```diff
@@ -22,13 +22,15 @@ const mapStateToProps = (state, {routeProps}) => {
     Constants.getDisplayCurrencyWaitingKey(accountID)
   )
   const saveCurrencyWaiting = anyWaiting(state, Constants.changeDisplayCurrencyWaitingKey)
+  const mobileOnlyMode = Constants.getMobileOnly(state, accountID)
```
chrisnojima: you can just reach into state here and not use some helper that is only used here

go/stellar/remote/remote.go
zapu: I missed that, thanks. I'll take a look.

go/stellar/stellarsvc/service_test.go
```diff
@@ -1223,6 +1225,7 @@ func TestMakeAccountMobileOnlyOnRecentMobile(t *testing.T) {
 	// this should not post a new bundle
 	err = tc.Srv.SetAccountMobileOnlyLocal(ctx, stellar1.SetAccountMobileOnlyLocalArg{
 		AccountID: a1,
+		Enabled:   true,
```
xgess: there are other places in this file (just one or two) where we should be calling like this to turn off mobile only, but we're not because it didn't exist yet. currently looks likes this: 
```go
err = remote.MakeAccountAllDevices(ctx, g, a1)
```
but should maybe probably look like this:
```go
err = tc.Srv.SetAccountMobileOnlyLocal(ctx, stellar1.SetAccountMobileOnlyLocalArg{
    AccountID: a1,
    Enabled:   false,
}
```
could you update those too?

zapu:
Review comments addressed.

@patrickxb do you want to take a look at server part as well?

go/stellar/remote/remote.go
patrickxb: So is this function changing?  I'd prefer SetAccountMobileOnly does the MakeMobileOnly.  And another function does MakeAllDevices instead of using a bool flag.

go/stellar/remote/remote_net.go
```diff
@@ -94,8 +94,8 @@ func (r *RemoteNet) IsAccountMobileOnly(ctx context.Context, accountID stellar1.
 	return IsAccountMobileOnly(ctx, r.G(), accountID)
 }
 
-func (r *RemoteNet) SetAccountMobileOnly(ctx context.Context, accountID stellar1.AccountID) error {
-	return SetAccountMobileOnly(ctx, r.G(), accountID)
+func (r *RemoteNet) SetAccountMobileOnly(ctx context.Context, accountID stellar1.AccountID, enabled bool) error {
```
patrickxb: Same for this, two different functions.

shared/actions/wallets.js
```diff
@@ -674,6 +674,28 @@ const rejectDisclaimer = (state: TypedState, action: WalletsGen.AcceptDisclaimer
       : Route.switchTo([state.routeTree.get('previousTab') || Tabs.peopleTab])
   )
 
+const loadMobileOnlyMode = (state: TypedState, action: WalletsGen.LoadMobileOnlyModePayload) => {
+  let accountID = action.payload.accountID
+  return RPCStellarTypes.localIsAccountMobileOnlyLocalRpcPromise({
+    accountID: accountID,
+  }).then(res =>
+    WalletsGen.createLoadedMobileOnlyMode({
+      accountID: accountID,
```
chrisnojima: should just be `accountId,`

shared/actions/wallets.js
```diff
@@ -674,6 +674,28 @@ const rejectDisclaimer = (state: TypedState, action: WalletsGen.AcceptDisclaimer
...
+  )
+}
+
+const changeMobileOnlyMode = (state: TypedState, action: WalletsGen.ChangeMobileOnlyModePayload) => {
+  let accountID = action.payload.accountID
+  let f = action.payload.enabled
+    ? RPCStellarTypes.localSetAccountMobileOnlyLocalRpcPromise
+    : RPCStellarTypes.localSetAccountAllDevicesLocalRpcPromise
+  return f({
+    accountID: accountID,
```
chrisnojima: same here

shared/reducers/wallets.js
```diff
@@ -263,6 +263,8 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
       return state.merge({
         acceptingDisclaimerDelay: true,
       })
+    case WalletsGen.loadedMobileOnlyMode:
+      return state.update('mobileOnlyMap', map => map.set(action.payload.accountID, action.payload.enabled))
```
chrisnojima: should be
```
return state.setIn(['mobileOnlyMap', action.payload.accountID], action.payload.enabled)
```

shared/wallets/wallet/settings/container.js
```diff
@@ -40,6 +42,8 @@ const mapDispatchToProps = (dispatch, {routeProps, navigateUp, navigateAppend})
     dispatch(navigateUp())
     dispatch(WalletsGen.createLoadPayments({accountID}))
   },
+  _onChangeMobileOnlyMode: (accountID: Types.AccountID, enabled: boolean) =>
+    dispatch(WalletsGen.createChangeMobileOnlyMode({accountID: accountID, enabled: enabled})),
```
chrisnojima: should be `{accountId, enabled}`

shared/reducers/wallets.js
zapu: ah ok, i tried `return state.updateIn(['mobileOnlyMap', action.payload.accountID], action.payload.enabled)` before which didn't work and i didn't know why. `setIn` is fine, thanks!

PR #15032, opened 2018-12-11
background poller for home announcements

maxtaco:
- we're eventually going to be showing everyone announcements who is online, so this mechanism
  polls to get those updates from the server
- pretty naive at first, we can optimize it later
- also make AVDL support for announcements that come back from the server


maxtaco:
CC @chrisnojima, I was thinking this morning in the shower that we should handle variants whose types we don't understand (and therefore fall into the `default: void;` case), *but* are badged, since that'll affect badging all the way up and look like an invisible badge.  How about a row that says to upgrade in that case?

go/home/home.go
```diff
@@ -370,3 +386,56 @@ func (h *Home) Create(ctx context.Context, cli gregor1.IncomingInterface, catego
...
+func (h *Home) updateLoopThread(m libkb.MetaContext) {
+	m = m.WithLogTag("HULT")
+	m.CDebugf("Starting Home#updateLoopThread")
+	slp := time.Minute * (time.Duration(5) + time.Duration((rand.Int() % 10)))
+	var err error
+	for {
+		m.CDebugf("Sleeping %v until next poll", slp)
+		m.G().Clock().Sleep(slp)
+		slp, err = h.pollOnce(m)
+		if _, ok := err.(libkb.DeviceRequiredError); ok {
```
heronhaye: this happens if the user is logged out?

go/home/home.go
```diff
@@ -370,3 +386,56 @@ func (h *Home) Create(ctx context.Context, cli gregor1.IncomingInterface, catego
...
+		return time.Duration(0), libkb.DeviceRequiredError{}
+	}
+
+	var raw rawPollHome
+	err = m.G().API.GetDecode(libkb.APIArg{
+		Endpoint:    "home/poll",
+		SessionType: libkb.APISessionTypeREQUIRED,
+		Args:        libkb.HTTPArgs{},
+		MetaContext: m,
+	}, &raw)
```
heronhaye: to clarify: this polls the server, which will send down a gregor message consumed on another path?

go/home/home.go
maxtaco: yes

go/home/home.go
maxtaco: yes, that is the flow. It forces the server to refresh its state, so that new home state, and new badging will be recomputed.

go/home/home.go
```diff
@@ -370,3 +386,56 @@ func (h *Home) Create(ctx context.Context, cli gregor1.IncomingInterface, catego
...
+func (h *Home) RunUpdateLoop(m libkb.MetaContext) {
+	go h.updateLoopThread(m)
+}
+
+func (h *Home) updateLoopThread(m libkb.MetaContext) {
+	m = m.WithLogTag("HULT")
+	m.CDebugf("Starting Home#updateLoopThread")
+	slp := time.Minute * (time.Duration(5) + time.Duration((rand.Int() % 10)))
+	var err error
+	for {
```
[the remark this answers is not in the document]
maxtaco: yeah, that was easier than managing the state machine

PR #15034, opened 2018-12-11
show ash on no ephermeral keys

chrisnojima:
@keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/wrapper/container.js
+++ b/shared/chat/conversation/messages/wrapper/container.js
@@ -99,7 +99,7 @@ const getFailureDescriptionAllowCancel = (message, you) => {
   }
   // $ForceType
   if (message.explodingUnreadable) {
-    failureDescription = 'This exploding message is not available to you.'
+    failureDescription = 'This exploding message is not available to you'
   }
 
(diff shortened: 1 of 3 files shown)


chrisnojima:
then we need some flag

joshblum:
i think the check should just be `isErrMsg && isExploding` if we don't want ash to show it's filtered on the core side already

chrisnojima:
@joshblum ok so we have a flag for `explodingUnreadable` but it was only on for the pairwise error code. now its true if there's any err and its exploding. We force that string but we could plumb through other stuff. we can do that later

chrisnojima:
sure, iw as just fixing a bug but if you want that to work i can change it further

joshblum:
yeah these PRs are from the same ticket we should just fix fully or it will bust accidentally again..

chrisnojima:
it won't bust, it'll just not show your custom message

joshblum:
should show custom msgs

chrisnojima:
done

joshblum:
thx!

chrisnojima:
@keybase/react-hackers bump, tiny change

PR #15042, opened 2018-12-11
account currency request fragility

xgess:
add some retrying to reduce the likelihood of visible errors on a bunch of stellar GET requests

### DIFF
--- a/go/libkb/interfaces.go
+++ b/go/libkb/interfaces.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"golang.org/x/net/context"
+	"stathat.com/c/ramcache"
 
 	"github.com/PuerkitoBio/goquery"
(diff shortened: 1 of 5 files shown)


patrickxb:
A few things:

We should return the default currency in GetWalletAccountsLocal, GetWalletAccountLocal.  Presumably, with that, frontend won't need to call this nearly as often.  They have a ticket for this...

I think we should ditch this global cache of the setting.  I was thinking of putting it into WalletState if we need it, and watching the calls to GetAccountDisplayCurrency, SetAccountDefaultCurrency to keep it up to date.

mmaxim:
I'll let @patrickxb handle this review

xgess:
@patrickxb i'm going to do the rest of the work in a different PR since it's pretty unrelated.

PR #15050, opened 2018-12-12
Add new rotation rules for open teams

zapu:
- Make the client skip all rotations on open teams (e.g. when removing members).
- Allow `PerTeamKey` section in `TeamSettings` link.
- Rotate team key when setting team from open to closed.
    - This happens by a separate `RotateKey` link now (so two new links total), but there is code to make it happen in the same `TeamSettings` link. But teamchain parser change has to bake 🍞.

### DIFF
--- a/go/teams/teams.go
+++ b/go/teams/teams.go
@@ -55,6 +55,11 @@ func (t *Team) CanSkipKeyRotation() bool {
 		return false
 	}
 
+	if t.IsOpen() {
+		// Skip all rotations in open teams.
+		return true
+	}
+
 	// If cannot decide because of an error, return default false.
 	members, err := t.UsersWithRoleOrAbove(keybase1.TeamRole_READER)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


zapu:
Needs server change first to pass CI: `team field per_team_key was unexpected (team.settings) (error 2657)`

go/teams/teams.go
```diff
@@ -1672,12 +1677,36 @@ func (t *Team) PostTeamSettings(ctx context.Context, settings keybase1.TeamSetti
...
+		maybeEKPayload = teamEKPayload // for storeTeamEKPayload, after post succeeds
+	}
+	err = t.postChangeItem(ctx, section, libkb.LinkTypeSettings, nil, payloadArgs)
 	if err != nil {
 		return err
 	}
 
-	t.notify(ctx, keybase1.TeamChangeSet{Misc: true})
+	if rotate {
+		t.notify(ctx, keybase1.TeamChangeSet{KeyRotated: true, Misc: true})
```
maxtaco: this needs to change since i fixed sequence number accounting. That's likely what the merge conflict is...

PR #15057, opened 2018-12-12
Update deps: flow 88. new hot loader

chrisnojima:
- [x] Flow 88
- [x] Better hot loader (handles errors so much better!)
- [x] New memoize* calls. In order to support custom equality functions we deviate from their kinda funky thing and use this now

Old:
```
memoize2((a, b) => 3, 
(newThing, oldThing) => boolean // this function is called 2 times, with oldA, newA, oldB, newB (optional)
```

New:
```
memoize2((a, b) => 3,
(newA, oldA) => boolean, // these are optional
(newB, oldB) => boolean, // these are optional
```

We can now safely type newA/newB etc. you can pass undefined and it'll default to a === b as the check


shared/app/main.desktop.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import {hot} from 'react-hot-loader'
+import {hot} from 'react-hot-loader/root'
```
chrisnojima: new RHL is a lot better! slight api change to opt into the new behavior

shared/chat/inbox/container/normal.js
```diff
@@ -97,12 +97,13 @@ const getRowsAndMetadata = memoize3(
     }
   },
   // ignore changes to metaMap if inboxVersion is the same
-  (a, b) => {
-    if (typeof a === 'number') {
+  (a, b, idx) => {
```
chrisnojima: memoize finally plumbs through the param order here

PR #15061, opened 2018-12-12
Cleanup WalletState refresh

patrickxb:
1. Only call RefreshAll once when the app starts (not 3 times)
2. on reconnect, if we have WalletState primed, wait 4s before refreshing
3. schedule a refresh when a get operation performed that returns wallet state data
4. increase the background loop timer to 5 minutes (I'm just about ready to kill it, but want to keep it for another day or two)
5. all refreshes have a "reason" to help debug why they happened
6. remove extra refresh calls from SubmitPayment

### DIFF
--- a/go/stellar/global.go
+++ b/go/stellar/global.go
@@ -244,11 +244,11 @@ func (s *Stellar) handleReconnect(ctx context.Context) {
 	go func() {
 		ctx := context.Background()
 		if s.walletState.Primed() {
-			s.G().Log.CDebugf(ctx, "stellar received reconnect msg, doing wallet refresh on unprimed wallet")
-		} else {
 			s.G().Log.CDebugf(ctx, "stellar received reconnect msg, doing delayed wallet refresh")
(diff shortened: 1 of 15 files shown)


go/stellar/global.go
```diff
@@ -248,8 +241,19 @@ func (s *Stellar) HandleOobm(ctx context.Context, obm gregor.OutOfBandMessage) (
 
 func (s *Stellar) handleReconnect(ctx context.Context) {
 	defer s.G().CTraceTimed(ctx, "Stellar.handleReconnect", func() error { return nil })()
-	s.G().Log.CDebugf(ctx, "stellar received reconnect msg, refreshing wallet state")
-	s.RefreshWalletState(ctx)
+	go func() {
+		ctx := context.Background()
+		if s.walletState.Primed() {
```
mmaxim: Is this log message right?

patrickxb: nope.

PR #15062, opened 2018-12-12
load send recommendations faster

zanderz:
is this the right way to fix it?


zanderz:
so I tried setting `useStaticSize` to false and a couple of other things, but none of them work. Should I try ReactWindow's FixedSizeList?

chrisnojima:
lets try it. i dislike not knowing why this is happening

zanderz:
so far I get this:
image

zanderz:
I got something showing with FixedSizeList but when you scroll down a few places it starts flashing/thrashing.

PR #15063, opened 2018-12-12
Review payment identifies the recipient

mlsteele:
`reviewPaymentLocal` identifies the recipient. There are 3 outcomes.
1. The identify succeeds, the caller is notified, and the call exits. ReadyToSend
2. If an identify ever succeeds or the recipient is tracked or untracked by this device then the caller is notified just like 1.
3. The identify errors out, the caller gets a banner and then hangs out forever. If 2 ever happens that's great.

### DIFF
--- a/go/engine/identify2_with_uid.go
+++ b/go/engine/identify2_with_uid.go
@@ -453,6 +453,7 @@ func (e *Identify2WithUID) runReturnError(m libkb.MetaContext) (err error) {
 
 	if e.hitFastCache(m) {
 		m.CDebugf("| hit fast cache")
+		e.maybeNotify(m, "hit fast cache")
 		return nil
 	}
 
@@ -482,11 +483,13 @@ func (e *Identify2WithUID) runReturnError(m libkb.MetaContext) (err error) {
 
 		if e.untrackedFastPath(m) {
 			m.CDebugf("| used untracked fast path")
+			e.maybeNotify(m, "untracked fast path")
 			return nil
 		}
 
 		if e.checkSlowCacheHit(m) {
 			m.CDebugf("| hit slow cache, first check")
+			e.maybeNotify(m, "slow cache, first check")
 			return nil
 		}
 	}
@@ -511,6 +514,7 @@ func (e *Identify2WithUID) runReturnError(m libkb.MetaContext) (err error) {
 
 	if e.useRemoteAssertions() && e.allowEarlyOuts() && e.checkSlowCacheHit(m) {
 		m.CDebugf("| hit slow cache, second check")
+		e.maybeNotify(m, "slow cache, second check")
 		return nil
 	}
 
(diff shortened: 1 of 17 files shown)


go/engine/identify2_with_uid.go
```diff
@@ -867,6 +871,10 @@ func (e *Identify2WithUID) runIdentifyUI(m libkb.MetaContext) (err error) {
 		_, err = e.state.Result().GetErrorLax()
 	}
 
+	if outcome.IsOK() {
```
mlsteele: I don't really understand what's behind this. Why `TrackStatus_NEW_FAIL_PROOFS` isn't considered IsOk. But I tried a bunch of stuff and this seems like the behavior that is desired. The ideal is to find a condition that means that the state of their identity satisfies your tracking of them, or that you don't track them.

go/stellar/build.go
```diff
@@ -386,6 +447,67 @@ func ReviewPaymentLocal(mctx libkb.MetaContext, stellarUI stellar1.UiInterface,
...
+	}
+
+	mctx.CDebugf("identifyForReview(%v)", assertion)
+	reason := fmt.Sprintf("Identify transaction recipient: %s", assertion)
+	eng := engine.NewResolveThenIdentify2(mctx.G(), &keybase1.Identify2Arg{
+		UserAssertion:         assertion,
+		CanSuppressUI:         true,
+		NoErrorOnTrackFailure: true, // take heed
+		Reason:                keybase1.IdentifyReason{Reason: reason},
+		IdentifyBehavior:      keybase1.TLFIdentifyBehavior_RESOLVE_AND_CHECK,
```
mlsteele: There are so many to choose from. Is this a good one? I just need track breaks to be part of the result, not error. And no tracker popups should pop up. This seems to unfortunately disable the cache too. @maxtaco

go/stellar/stellar.go
```diff
@@ -515,6 +515,10 @@ func SendPaymentGUI(m libkb.MetaContext, walletState *WalletState, sendArg SendP
 func sendPayment(m libkb.MetaContext, walletState *WalletState, sendArg SendPaymentArg, isCLI bool) (res SendPaymentResult, err error) {
 	defer m.CTraceTimed("Stellar.SendPayment", func() error { return err })()
 
+	if err = m.Ctx().Err(); err != nil {
+		return res, err
+	}
```
patrickxb: what's this block for?

go/stellar/build.go
patrickxb: I think this is good?

go/stellar/stellar.go
mlsteele: It's just a friendly yield. When context's get canceled sometimes it's nice not to have them log a huge tree before realizing they've been canceled at the bottom of the callstack. But I don't remember why this one in particular and it doesn't seem likely for this to be canceled, so I'll remove it.

mlsteele:
This one can wait until after code 🍟

mlsteele:
nah

PR #15068, opened 2018-12-13
Improve ephemeral key error messaging

joshblum:
detects two types of ephemeral errors to have more informative error messages show to users. the following errors are checked:
- the user provisioned the device since the message was sent
- the user cloned a device

relates to https://github.com/keybase/client/pull/15034 and depends on https://github.com/keybase/keybase/pull/3254
screen shot 2018-12-13 at 9 54 53 am
screen shot 2018-12-13 at 9 53 56 am

### DIFF
--- a/go/ephemeral/errors.go
+++ b/go/ephemeral/errors.go
@@ -1,0 +1,57 @@
+package ephemeral
+
+import (
+	"fmt"
+
+	"github.com/keybase/client/go/protocol/gregor1"
+	"github.com/keybase/client/go/protocol/keybase1"
+)
+
+type EKType string
+
+const (
+	DeviceEKStr EKType = "deviceEK"
+	UserEKStr   EKType = "userEK"
+	TeamEKStr   EKType = "teamEK"
+)
+
+type EphemeralKeyError struct {
+	DebugMsg string
+	HumanMsg string
+}
+
+const defaultHumanErr = "This exploding message is not available to you."
+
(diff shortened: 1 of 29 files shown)


go/ephemeral/errors.go
```diff
@@ -0,0 +1,94 @@
...
+	TeamEKStr   EKType = "teamEK"
+)
+
+type EphemeralKeyError struct {
+	DebugMsg string
+	HumanMsg string
+}
+
+const (
+	defaultHumanErr                             = "This exploding message is not available to you"
```
joshblum: cc @malgorithms / @cecileboucheron, any opinion on the copy?

go/ephemeral/team_ek_box_storage.go
```diff
@@ -419,7 +435,7 @@ func (s *TeamEKBoxStorage) MaxGeneration(ctx context.Context, teamID keybase1.Te
 
 // --------------------------------------------------
 
-const MemCacheLRUSize = 200
```
joshblum: seemed unnecessarily low

go/ephemeral/errors.go
malgorithms: which of the unavailable scenarios does this apply to?

joshblum: this pr covers:

 - the user provisioned the device since the message was sent
 - the user cloned a device

that particular line is just the default fallback if we have no specific info

malgorithms: error you have seems fine

PR #15069, opened 2018-12-13
don't force-repoll on implicit team chat send hot path

maxtaco:
- fix the test that this broke, and in general:
- fix the notification system internal to the app that adjusts the locally cached version of the teamchain
 - sometimes bump the chain 0, sometimes 1, sometimes 2, depending on the caller!
 - previously, always bumping +1 definitely caused some bugs, but we likely got away with it

### DIFF
--- a/go/teams/implicit.go
+++ b/go/teams/implicit.go
@@ -35,23 +35,33 @@ func (i *implicitTeam) GetAppStatus() *libkb.AppStatus {
 	return &i.Status
 }
 
+type forceRepollToggle bool
+
 // Lookup an implicit team by name like "alice,bob+bob@twitter (conflicted copy 2017-03-04 #1)"
 // Resolves social assertions.
 func LookupImplicitTeam(ctx context.Context, g *libkb.GlobalContext, displayName string, public bool) (
 	team *Team, teamName keybase1.TeamName, impTeamName keybase1.ImplicitTeamDisplayName, err error) {
-	team, teamName, impTeamName, _, err = LookupImplicitTeamAndConflicts(ctx, g, displayName, public)
+	team, teamName, impTeamName, _, err = LookupImplicitTeamAndConflicts(ctx, g, displayName, public, forceRepollToggle(true))
 	return team, teamName, impTeamName, err
 }
 
 // Lookup an implicit team by name like "alice,bob+bob@twitter (conflicted copy 2017-03-04 #1)"
 // Resolves social assertions.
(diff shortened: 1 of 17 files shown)


maxtaco:
cc @mmaxim  this should fix the chat perf bug

go/teams/teams.go
```diff
@@ -666,10 +669,17 @@ func (t *Team) Leave(ctx context.Context, permanent bool) error {
...
+	if err != nil {
+		return err
+	}
+
+	t.notify(ctx, keybase1.TeamChangeSet{MembershipChanged: true}, latestSeqno)
+	return nil
 }
 
 func (t *Team) deleteRoot(ctx context.Context, ui keybase1.TeamsUiInterface) error {
+	m := t.MetaContext(ctx)
```
mlsteele: mctx

go/teams/transactions.go
```diff
@@ -836,7 +836,7 @@ func (tx *AddMemberTx) Post(mctx libkb.MetaContext) (err error) {
 		return err
 	}
 
-	team.notify(mctx.Ctx(), keybase1.TeamChangeSet{MembershipChanged: true})
+	team.notify(mctx.Ctx(), keybase1.TeamChangeSet{MembershipChanged: true}, nextSeqno)
```
mlsteele: nextSeqno-1 ? next vs last

go/teams/implicit.go
```diff
@@ -176,7 +176,7 @@ func lookupImplicitTeamAndConflicts(ctx context.Context, g *libkb.GlobalContext,
 	team, err = Load(ctx, g, keybase1.LoadTeamArg{
 		ID:          teamID,
 		Public:      impTeamName.IsPublic,
-		ForceRepoll: true,
+		ForceRepoll: false, // it's important to hit the cache here, since we call this once per chat send!
```
mlsteele: This function drives public functions which I'd expect to return the latest answer. Does chat not have its own cache?

mlsteele: Or could plumb forceRepoll through

go/teams/transactions.go
maxtaco: ah good call, thanks. this was the particularly buggy bit.

maxtaco: ...since it could result in stale values being used...

go/teams/implicit.go
maxtaco: no, chat just spams this (cc @mmaxim).  I was thinking though it should be OK to get a slightly stale answer here...

PR #15073, opened 2018-12-13
Use mctx more

patrickxb:
stellar client stuff is a bit of a jumble when it comes to MetaContext...going to try to improve that in little pieces.  Here is piece 1...


patrickxb:
i'm hoping that improves...I started making big changes, but that will be a nightmare with everyone working on the same code, so going to do it in little pieces when time allows.

PR #15082, opened 2018-12-13
Use new stellar notifications

buoyad:
Gets rid of our old handlers to use data from the notifications themselves. Everything on the account page should still update in real time.

Rough stellar RPC deltas doing a single send (counting responses)
before:
out: 28
in: 35

after:
out: 3
in: 10

r? @keybase/react-hackers cc @patrickxb

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -607,23 +607,35 @@ const maybeNavigateToConversation = (state: TypedState, action: WalletsGen.Reque
 
 const setupEngineListeners = () => {
   getEngine().setIncomingCallMap({
-    'stellar.1.notify.paymentNotification': refreshPayments,
-    'stellar.1.notify.paymentStatusNotification': refreshPayments,
+    'stellar.1.notify.accountDetailsUpdate': ({accountID, account}) => [
+      Saga.put(
+        WalletsGen.createAccountsReceived({
+          accounts: [Constants.accountResultToAccount(account)],
+        })
+      ),
+      Saga.put(WalletsGen.createLoadAssets({accountID: Types.stringToAccountID(accountID)})),
+    ],
+    'stellar.1.notify.pendingPaymentsUpdate': ({accountID: _accountID, pending: _pending}) => {
+      if (!_pending) {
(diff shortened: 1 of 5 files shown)


buoyad:
There'll be another part to reduce `getDisplayCurrency` calls after #15056

shared/actions/wallets.js
```diff
@@ -98,6 +98,7 @@ const openSendRequestForm = (state: TypedState, action: WalletsGen.OpenSendReque
           WalletsGen.createSetBuildingCurrency({
             currency:
               action.payload.currency ||
+              (state.wallets.lastSentXLM && 'XLM') ||
```
chrisnojima: === ?

shared/reducers/wallets.js
```diff
@@ -53,6 +53,22 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
...
+      )
+    case WalletsGen.recentPaymentsReceived:
+      const newPayments = I.Map(action.payload.payments.map(p => [p.id, Constants.makePayment().merge(p)]))
+      return state
+        .updateIn(['paymentsMap', action.payload.accountID], (paymentsMap = I.Map()) =>
+          paymentsMap.merge(newPayments)
+        )
+        .updateIn(
+          ['paymentCursorMap', action.payload.accountID],
+          cursor => cursor || action.payload.paymentCursor
```
chrisnojima: why keep an existing cursor?

shared/actions/wallets.js
buoyad: thanks!

buoyad: whoops actually I thought that this was a different file. not a typo! if there isn't an `action.payload.currency` and we know `lastSentXLM`, this sets the currency to `XLM`.

shared/reducers/wallets.js
buoyad: in case we've already loaded payments

shared/actions/wallets.js
chrisnojima: k

PR #15086, opened 2018-12-14
display status on in-chat wallet payments CORE-9656

mmaxim:
**Go**

1. Add "decoration" concept, which allows us to embed JSON into the payload of the chat message we deliver to the UI. Using this, we can encode payment information in the correct spot that it should be displayed to the markdown parser. `UITextDecoration` is the generic type that will be encoded in the message.
2. Add `DecoratePayments` which will replace all valid payment text in a message with a JSON record derived from `chat1.TextPayment`. We have to be careful here to deal with replacing into the text with the markdown in play.
3. Always send a payment notification when running `LoadPayment` in the Stellar loader. This way the UI only needs to care about these notifications, and not the data attached to `UIMessage`, where there can be races.
4. Add a new field to `IncomingMessage` notifications that include the target of an edit message. This allows the UI to just swap this message in, instead of trying to do the edit manually. The edit is now much harder to do with these JSON records floating around, so the service should always be in charge of it.

**JS**
@keybase/react-hackers 

1. Add `PaymentStatus` component which will display the status of a transaction. It is still a TODO to add the hover state to them, but I thought it would be good to get this in to play with before doing that.
2. Add `ServiceDecoration` into the markdown parser in order to parse out the decorations generated by the service. This class can switch on the type from `UITextDecoration` in order to invoke the correct component for the JSON.
3. Add new state to redux to track the status of payment IDs.
4. Use the new `modifiedMessage` from the service to implement edits in `onIncomingMessage`.


shared/common-adapters/markdown/service-decoration.js
```diff
@@ -0,0 +1,40 @@
...
+        case RPCChatTypes.localTextPaymentResultTyp.error:
+          error = parsed.payment.result.error
+          break
+      }
+      return (
+        <PaymentStatus
+          paymentID={paymentID}
+          error={error}
+          text={parsed.payment.paymentText}
+          allowFontScaling={props.allowFontScaling}
```
chrisnojima: why plumb this?

shared/common-adapters/markdown/service-decoration.js
```diff
@@ -0,0 +1,40 @@
...
+
+export type Props = {
+  json: string,
+  onClick?: () => void,
+  allowFontScaling?: ?boolean,
+}
+
+const ServiceDecoration = (props: Props) => {
+  // Parse JSON to get the type of the decoration
+  const parsed = JSON.parse(props.json)
```
chrisnojima: have to wrap this in a try. we don't want to crash out

shared/constants/chat2/message.js
```diff
@@ -170,6 +170,7 @@ export const makeMessageDeleted: I.RecordFactory<MessageTypes._MessageDeleted> =
 export const makeMessageText: I.RecordFactory<MessageTypes._MessageText> = I.Record({
   ...makeMessageCommon,
   ...makeMessageExplodable,
+  decoratedText: null,
```
chrisnojima: why do we need to differentiate between text and decoratedText?

shared/reducers/chat2.js
```diff
@@ -890,7 +890,10 @@ const rootReducer = (
     }
     case Chat2Gen.paymentInfoReceived: {
       const {conversationIDKey, messageID, paymentInfo} = action.payload
-      return state.update('accountsInfoMap', old => old.setIn([conversationIDKey, messageID], paymentInfo))
+      let nextState = state.update('accountsInfoMap', old =>
```
chrisnojima: this is ok but other options 
```
return state
  .update()
  .update()

// or

return state.merge({
  accountsInfoMap: state. accountsInfoMap.update(),
  paymentStatusMap: state. paymentStatusMap.update()
})
```

PR #15098, opened 2018-12-14
Fix jump to chat enter key 

chrisnojima:
Fixes a bug introduced by https://github.com/keybase/client/pull/14952

if you hit enter we just want to ensure something is selected from the filtered list , so plumb a new prop to do this

cc: @zanderz @songgao 
@keybase/react-hackers

### DIFF
--- a/shared/chat/inbox/row/chat-filter-row/index.js
+++ b/shared/chat/inbox/row/chat-filter-row/index.js
@@ -54,7 +54,6 @@ class ChatFilterRow extends React.PureComponent<Props, State> {
       e.preventDefault()
       e.stopPropagation()
       this.props.onSelectUp()
-    } else {
     }
   }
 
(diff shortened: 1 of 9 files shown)


shared/chat/inbox/row/chat-filter-row/index.js
```diff
@@ -1,19 +1,10 @@
 // @flow
 import * as React from 'react'
-import {Icon, Box, ClickableBox, LoadingLine, Input, Text} from '../../../../common-adapters'
```
chrisnojima: 🔥

shared/chat/inbox/row/chat-filter-row/index.js
```diff
@@ -62,17 +54,18 @@ class ChatFilterRow extends React.PureComponent<Props, State> {
       e.preventDefault()
       e.stopPropagation()
       this.props.onSelectUp()
+    } else {
```
buoyad: remove block

PR #15101, opened 2018-12-14
Announcements integration

chrisnojima:
This hooks up announcements in the people tab. There is a new admin panel to turn on announcements (there are 3 there already for admin only). You can play with that to test. Note only one will show up at a time so you should just deactivate any other ones. You also likely have to restart your daemon to get it to show up

There's a bug in the server so the id we get is 0 which causes dismissal to not work but that'll be fixed soon. this can be reviewed i think

- [x] test dismissal (needs server deploy)


chrisnojima:
@keybase/react-hackers

shared/people/announcement/container.js
```diff
@@ -1,35 +1,83 @@
...
 const mapStateToProps = () => ({})
-// TODO
-const mapDispatchToProps = () => ({
-  onConfirm: () => {
-    console.log('announcement onConfirm TODO')
+
+// Really the saga should handle all of this and we shouldn't have any of these ownProps passed in but this
+// is how the other types work in this list. TODO change this to be more modern
+const mapDispatchToProps = dispatch => ({
+  _onConfirm: (id, appLink, url) => {
```
buoyad: you can pull `ownProps` in here and avoid the closures in `mergeProps`

shared/constants/people.js
```diff
@@ -159,16 +158,17 @@ export const reduceRPCItemToPeopleItem = (
...
       )
     }
-  } else if (item.data.t === 3) {
-    if (__DEV__ && flags.peopleAnnouncementsEnabled) {
-      const a: any = item.data.announcement
+  } else if (item.data.t === RPCTypes.homeHomeScreenItemType.announcement) {
+    const a = item.data.announcement
+    if (a) {
       return list.push(
         makeAnnouncement({
```
buoyad: looks like `makeAnnouncement` below is dropping some fields

chrisnojima: fixed

shared/constants/people.js
```diff
@@ -184,6 +184,7 @@ export const makeAnnouncement: I.RecordFactory<Types._Announcement> = I.Record({
   badged: false,
   confirmLabel: null,
   dismissable: false,
```
buoyad: `iconUrl` here too

chrisnojima: k fixed ,txn

PR #15104, opened 2018-12-14
Only accept account updates if we've loaded them already

buoyad:
and load assets as a side effect instead of emitting another action. r? @keybase/react-hackers

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -9,6 +9,7 @@ import * as Chat2Gen from './chat2-gen'
 import * as ConfigGen from './config-gen'
 import * as NotificationsGen from './notifications-gen'
 import * as RouteTreeGen from './route-tree-gen'
+import * as Flow from '../util/flow'
 import HiddenString from '../util/hidden-string'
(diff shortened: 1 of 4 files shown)


shared/actions/wallets.js
```diff
@@ -530,13 +552,18 @@ const exportSecretKey = (state: TypedState, action: WalletsGen.ExportSecretKeyPa
...
-  state.wallets.selectedAccount === Types.noAccountID &&
-  Saga.put(
-    WalletsGen.createSelectAccount({
-      accountID: state.wallets.accountMap.find(account => account.isDefault).accountID,
-    })
-  )
+const maybeSelectDefaultAccount = (action: WalletsGen.AccountsReceivedPayload, state: TypedState) => {
+  if (state.wallets.selectedAccount === Types.noAccountID) {
+    const maybeDefaultAccount = state.wallets.accountMap.find(account => account.isDefault)
+    if (maybeDefaultAccount) {
```
buoyad: Fixes the `property 'accountID' of undefined` black bar

PR #15105, opened 2018-12-14
Stellar: Inflation destination RPC and CLI

zapu:


go/client/cmd_wallet_set_inflation.go
```diff
@@ -0,0 +1,99 @@
...
+		Name:         "set-inflation",
+		Usage:        "Set inflation destination address for an account",
+		ArgumentHelp: "<account id> <destination address or label>",
+		Action: func(c *cli.Context) {
+			cl.ChooseCommand(cmd, "set-inflation", c)
+		},
+		Description: `Read more about Stellar inflation payouts:
+   https://www.stellar.org/developers/guides/concepts/inflation.html
+
+   Destination can be either a user chosen address (any Stellar account ID),
```
patrickxb: user-chosen

go/client/cmd_wallet_set_inflation.go
```diff
@@ -0,0 +1,99 @@
...
+   or one of the following labels:
+
+   * "lumenaut" - Sets account's inflation destination to Lumenaut Pool. Read more
+     on https://pool.lumenaut.net/
+
+   * "self" - Sets inflation destination to the source account ID.
+
+   Inflation destination cannot be removed or cleared from the account right
+   now, but if you want to stop contributing to currently set destination, use
```
patrickxb: to the currently
use the "self" option.

go/stellar/remote/remote.go
```diff
@@ -1260,3 +1260,38 @@ func ServerTimeboundsRecommendation(ctx context.Context, g *libkb.GlobalContext)
...
+	_, err = g.API.Post(apiArg)
+	return err
+}
+
+type getInflationRes struct {
+	libkb.AppStatusEmbed
+	DestinationAccountID *stellar1.AccountID `json:"destination_account_id"`
+}
+
+func GetInflationDestination(ctx context.Context, g *libkb.GlobalContext, accountID stellar1.AccountID) (res *stellar1.AccountID, err error) {
```
patrickxb: will this also be in stellar1.AccountDetails?

zapu: there is no reason not to put it there, right?

patrickxb: If it's there, then you can do something like `walletState.Details(ctx, accountID).InflationDestination`

PR #15111, opened 2018-12-14
feature flag kbfs/chat integration

songgao:
Looks like the loading screen PR won't make it for this release. So we should feature-flag linkifying KBFS paths in chat for now. I'll turn the mobile back on after release.

### DIFF
--- a/shared/common-adapters/markdown/kbfs-path.js
+++ b/shared/common-adapters/markdown/kbfs-path.js
@@ -2,6 +2,7 @@
 import React from 'react'
 import Text from '../text'
 import * as Types from '../../constants/types/fs'
+import features from '../../util/feature-flags'
 
 export type Props = {
   path: Types.Path,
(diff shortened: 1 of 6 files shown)


shared/common-adapters/markdown/kbfs-path.js
```diff
@@ -2,17 +2,22 @@
...
+export default (features.kbfsChatIntegration
+  ? (props: Props) => (
+      <Text type="BodyPrimaryLink" onClick={props.onClick} allowFontScaling={!!props.allowFontScaling}>
+        {Types.pathToString(props.path)}
+      </Text>
+    )
+  : (props: Props) => (
+      <Text type="Body" allowFontScaling={!!props.allowFontScaling}>
+        {Types.pathToString(props.path)}
+      </Text>
```
jzila: This could just be:
```javascript
export default (props: Props) => {
  return (
    <Text type={features.kbfsChatIntegration ? "BodyPrimaryLink" : "Body"} onClick={features.kbfsChatIntegration ? props.onClick : null} ...>
      ...
    </Text>
}
```

songgao: Agree this would look cleaner. Though I was trying to avoid checking against that inside the component, so that the check only happens once at import, instead of every time we need to render. But probably won't be that much different as there are very few KBFS links in chat ...

jzila: Hmm that's a good point. Ok keep it the way you have it now, but please add a comment with this explanation.

songgao: oops pushed too fast! Let me replace that with a comment ...

PR #15118, opened 2018-12-17
show payment details on in-chat payments on interaction CORE-9681

mmaxim:
**Go**

1. Add more information for `UIChatPaymentInfo` to get information relevant to rendering "third party" txns that show up in team chats.
2. Fix a bug where sending a `nil` conversation ID on payment info update would crash the UI.

**JS**
@keybase/react-hackers (@buoyad might want to look at this, since it uses his component).

1. Make `SendPaymentPopup` take an optional payment ID instead of pulling it off the message (and take a text message). Also give it a few more features, like being able to click through to wallet tab on errors, and a more general "top line". 
2. Show a `SendPaymentPopup` on hover/click on `PaymentStatus`.
3. Thread the `Types.Message` from the text message through the markdown parser so we can use its data when rendering `PaymentStatus`, particularly the hover state with details.
4. Add an error hover screen for payment that don't even get submitted.


shared/chat/conversation/messages/text/index.js
```diff
@@ -9,15 +9,24 @@ export type Props = {
...
+  mentionsAt,
+  mentionsChannel,
+  mentionsChannelName,
+  message,
+}: Props) => {
   const markdown = (
     <Kb.Markdown
       style={getStyle(type, isEditing)}
-      meta={{mentionsAt, mentionsChannel, mentionsChannelName}}
+      meta={{mentionsAt, mentionsChannel, mentionsChannelName, message}}
```
chrisnojima: Why pass the entire message down   We extract at already

mmaxim: Why not? The entire message is a prop to `SendPaymentPopup`.

chrisnojima: That thing should be connected and just get the id it needs

chrisnojima: When we discussed on Friday we weee talking about the payment id and that's it

mmaxim: `SendPaymentPopup` is something that already exists, I didn't want to change it too much, since it already seems to work fine. It is connected, but needs the message to render itself when it is being used on a normal Stellar send message. It is also convenient for getting things out, like the author and sending device, which is what this component uses it for.

chrisnojima: If that's the case we should do what we're doing with mentions. Now we're sending the whole message down and yet peeling off several parts

mmaxim: Why? It seems like the right move is to just send the whole thing down and mentions pull off what it needs. Otherwise now I have some dependency in the text component on what some downstream things the markdown parser might generate need. Seems better to send the whole message down and let those components pick out what they want. Is there some performance thing I am missing here?

mmaxim: Did you change should to shouldnt? If so, I didn't want to refactor mentions as part of this, I can pick that up if we change how that works in general (which is to use the method I am using for these Stellar messages).

chrisnojima: Maybe I'm misunderstanding. We're sending a message and mentions. The mentions are in the message also. They come from the message

chrisnojima: I did as that was a typo

mmaxim: Yeah see above, I didn't want a whole bunch of changes to mentions in this PR, seems appropriate to fix that in some other change.

shared/chat/payments/confirm/index.js
```diff
@@ -109,16 +109,9 @@ const PaymentsConfirm = (props: Props) => (
...
         </Kb.ScrollView>
         <Kb.ButtonBar align="center" direction="row" fullWidth={true} style={styles.buttonBar}>
-          <Kb.Button
-            type="Secondary"
-            small={Styles.isMobile}
-            onClick={props.onCancel}
-            style={styles.cancelButton}
-            label="Cancel"
-          />
+          <Kb.Button type="Secondary" onClick={props.onCancel} style={styles.cancelButton} label="Cancel" />
```
chrisnojima: should this be waitingbutton?

shared/chat/payments/status/index.js
```diff
@@ -24,18 +37,78 @@ const getIcon = status => {
...
+        style={styles[this.props.status]}
+        onClick={this._showPopup}
+      >
+        {' '}
+        {this.props.text}{' '}
+        <Kb.Icon
+          type={getIcon(this.props.status)}
+          fontSize={12}
+          boxStyle={styles.iconBoxStyle}
+          style={styles[this.props.status + 'Icon']}
```
chrisnojima: fyi this likely doesn't flow correctly. it can't dynamically understand the type. this is fine short term but a better approach is to do like 
``` 
this.getStatusIconStyle = () => {
 switch(this.props.status) {
case 'whatever':
return 'whateverIcon'
}
}
```

so flow can know all the cases of the map

shared/chat/payments/confirm/index.js
mmaxim: It is immediate, no network or anything in play.

shared/chat/conversation/messages/message-popup/payment/index.js
```diff
@@ -91,6 +92,16 @@ const Header = (props: HeaderProps) =>
           {props.balanceChange}
         </Kb.Text>
       )}
+      {!!props.errorDetails && (
+        <Kb.Box2 direction="horizontal" style={{maxWidth: 200}}>
+          <Kb.Text
+            type="BodyExtrabold"
+            style={Styles.collapseStyles([styles.textAlignCenter, {color: Styles.globalColors.red}])}
```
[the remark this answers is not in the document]
mmaxim: K the code above this does what I did so I thought it was cool, I can just make it a style though.

PR #15123, opened 2018-12-17
default to v2 bundles for new users if they have the feature flag

xgess:
when a user has the `stellar_acct_bundles` feature flag before a wallet is created, go ahead and create the v2 wallet instead of the v1.


patrickxb:
I'm sure there is, but is there a test for a user creating v2 instead of v1?  GH diff is foiling me...

xgess:
@patrickxb yeah. check out `TestAccountBundleFlows`

PR #15125, opened 2018-12-17
flow 89

chrisnojima:
Updates us to flow 89
This greatly improves typing of HOCs finally. 
For an example of a hoc that strips out some props (popup dialog) see: 
https://github.com/keybase/client/pull/15125/files#diff-26a84c8e09c1bd8451c4f34d7c1c9d13R9

For an example of a hoc that adds in some props (timer hoc style), see:
https://github.com/keybase/client/pull/15125/files#diff-3495bbe60d696a1e4df1189e591499e6R4

Note we now have some warnings about $Subtype/Supertype being deprecated in redux_v4/react-redux_v5/prop-types_v15/lodash-es_v4 in flow-typed. Going to have to wait for updates there for the warnings to go away

Please read: https://medium.com/flow-type/supporting-react-forwardref-and-beyond-f8dd88f35544
and https://flow.org/en/docs/react/hoc/


chrisnojima:

shared/common-adapters/hoc-timers.js
```diff
@@ -1,33 +1,36 @@
...
+function getDisplayName(Component): string {
+  return Component.displayName || Component.name || 'Component'
 }
 
-function HOCTimers<Props: TimerProps>(
-  WrappedComponent: React.ComponentType<Props>
-): React.ComponentType<$Diff<Props, TimerProps>> {
-  class TimersComponent extends React.Component<$Diff<Props, TimerProps>> {
-    static displayName = `HOCTimers(${getDisplayName(WrappedComponent)})`
+function hOCTimers<Config: {}, Instance>(
```
buoyad: `HOCTimers`

chrisnojima: this is a function that returns a class so marking it as such. .matches the style in the flow docs: https://flow.org/en/docs/react/hoc/ and react docs: https://reactjs.org/docs/higher-order-components.html

PR #15126, opened 2018-12-17
pagination.Last

joshblum:
cache the `pagination.Last` value and use that for `GetThreadNonblock` requests. if a request has a cached value of `Last=true` return no results

### DIFF
--- a/go/chat/server.go
+++ b/go/chat/server.go
@@ -670,7 +670,7 @@ func (h *Server) GetThreadNonblock(ctx context.Context, arg chat1.GetThreadNonbl
 
 	// Apply any pager mode transformations
 	pagination = h.applyPagerModeIncoming(ctx, arg.ConversationID, pagination, arg.Pgmode)
-	if pagination.Last {
+	if pagination != nil && pagination.Last {
 		return res, nil
 	}
 
--- a/go/chat/server.go
+++ b/go/chat/server.go
@@ -521,24 +521,23 @@ func (h *Server) applyPagerModeIncoming(ctx context.Context, convID chat1.Conver
 			pagination, res)
 	}()
 	switch pgmode {
-	case chat1.GetThreadNonblockPgMode_DEFAULT:
-		return pagination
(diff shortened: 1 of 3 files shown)


go/protocol/chat1/extras.go
```diff
@@ -1019,7 +1019,7 @@ func (p *Pagination) String() string {
 
 // FirstPage returns true if the pagination object is not pointing in any direction
 func (p *Pagination) FirstPage() bool {
-	return p == nil || (len(p.Next) == 0 && len(p.Previous) == 0)
+	return p == nil || (len(p.Next) == 0 && len(p.Previous) == 0 && !p.Last)
```
mmaxim: I'm curious why this is there, seems wrong.

joshblum: might not be necessary but it seemed wrong to say it was the first page if that was set, if we respond to a query with no results `Next` and `Previous` are not set but `Last` is, so i don't think `FirstPage` should return `true` in that case

mmaxim: This is only really used when determining if the `Pagination` is trying to get the first page. It is probably irrelevant what we do with `Last` here, but in the interest of just making the minimal change we could just leave it alone.

go/chat/server.go
```diff
@@ -671,6 +670,9 @@ func (h *Server) GetThreadNonblock(ctx context.Context, arg chat1.GetThreadNonbl
 
 	// Apply any pager mode transformations
 	pagination = h.applyPagerModeIncoming(ctx, arg.ConversationID, pagination, arg.Pgmode)
+	if pagination != nil && pagination.Last {
```
mmaxim: What is this for?

joshblum: if we get a null result from `Pull` our outgoing pagination is set to `&chat1.Pagination{Num:x, Prev:0, Next:0, Last:true}`, if we call the function again Pull will default to give use the `x` results since `Next` is unset so i wanted to short circuit here

PR #15127, opened 2018-12-17
Fetch payment details on paymentstatusnotif

buoyad:
Fixes relay transition between 'pending' -> 'cancelable'. r? @keybase/react-hackers


buoyad:
Fixed by #15130

PR #15136, opened 2018-12-17
Fix unread only

joshblum:
calculates the max visible msg id and compares it to our max read id


go/chat/storage/inbox.go
```diff
@@ -526,10 +526,11 @@ func (i *Inbox) applyQuery(ctx context.Context, query *chat1.GetInboxQuery, rcs
 			*query.TlfVisibility != conv.Metadata.Visibility {
 			ok = false
 		}
-		if query.UnreadOnly && conv.ReaderInfo.ReadMsgid >= conv.ReaderInfo.MaxMsgid {
+		isUnread := conv.IsUnread()
```
mmaxim: We should only call this if the query param is set.

joshblum: done

PR #15137, opened 2018-12-17
Make send/request form routes fullscreen

buoyad:
r? @keybase/react-hackers


shared/common-adapters/safe-area-view.desktop.js
```diff
@@ -1,3 +1,10 @@
...
-export {Box as default, Box as SafeAreaViewTop} from './box'
+import * as React from 'react'
+import type {SafeAreaViewTopBottomProps} from './safe-area-view'
+
+const RenderChildren = (props: SafeAreaViewTopBottomProps) => (
+  <React.Fragment>{props.children}</React.Fragment>
+)
+
+// Do nothing
+export {RenderChildren as default, RenderChildren as SafeAreaViewTop, RenderChildren as SafeAreaViewTopBottom}
```
buoyad: cc @chrisnojima I changed how this works. I think it only affects the provision code page, I checked it out and it looks good.

shared/common-adapters/safe-area-view.native.js
```diff
@@ -1,27 +1,61 @@
...
-}
-export const SafeAreaViewTop = ({style, children}: SafeAreaViewTopProps) =>
+export const SafeAreaViewTop = ({style, children}: Props) =>
   Styles.isAndroid ? (
     <View style={Styles.collapseStyles([styles.androidTopSafeArea, style])}>{children}</View>
   ) : (
     <SafeAreaView style={Styles.collapseStyles([styles.topSafeArea, style])}>{children}</SafeAreaView>
   )
 
+export const SafeAreaViewTopBottom = (props: SafeAreaViewTopBottomProps) => (
```
chrisnojima: spoke irl , don't think this is necesasry. you can just use a SafeAreaView on th ebottom of the screen w/ whatever color you want

buoyad: @chrisnojima I pushed a different approach, ptal

shared/common-adapters/safe-area-view.desktop.js
```diff
@@ -1,3 +1,7 @@
 // @flow
-// Just act like a box
-export {Box as default, Box as SafeAreaViewTop} from './box'
+import type {Props} from './safe-area-view'
+
+const RenderChildren = (props: Props) => props.children
```
MarcoPolo: This works? you don't have to wrap it in a fragment?

chrisnojima: why not? children is React.Node which is what render returns

MarcoPolo: 👍

buoyad:
@MarcoPolo thanks for looking, I still have to add a fix to make the safearea + keyboardavoidingview play nice with eachother

- [ ] fix keyboard covering confirm button on create new account / link existing account

buoyad:
OK, I spent some more time trying to make it work on the create new / link existing dialogs with no luck, so I reverted the changes to their routing and `WalletPopup`. Now this all works _except_ for when you make a new account from the "Send to other account" form, in which case the bottom button is janky on newer iPhones and notched android phones
image

@keybase/react-hackers PTAL

PR #15139, opened 2018-12-17
protocol and implementation stubs for ID3

maxtaco:
- feedback requested


maxtaco:
@chrisnojima  this is pretty close to what you had on confluence

protocol/avdl/keybase1/identify3_ui.avdl
```diff
@@ -0,0 +1,63 @@
...
+    BLACK_3,
+    GREEN_4
+  }
+
+  @typedef("string")
+  record Identify3RowMeta {}
+
+  @lint("ignore")
+  void identify3UpdateRow(
+    Identify3Assertion guiID,      // id used to tie these things together
```
mlsteele: two `guiID`s weird how about assertionID

protocol/avdl/keybase1/identify3_ui.avdl
```diff
@@ -0,0 +1,63 @@
...
+    string key,                    // e.g., "twitter", "web", "github", "btc", "pgp"
+    string value,                  // e.g., "bob", "https://foo.com", "1b3309isofijweorijgj3oij423", "33aabbccddeeff",
```
mlsteele: current profile view shows `foo.com@https` not `https://foo.com@https`

protocol/avdl/keybase1/identify3_ui.avdl
maxtaco: whoops, that just a bug. sorry about that, fixing.

protocol/avdl/keybase1/identify3.avdl
```diff
@@ -0,0 +1,26 @@
+@namespace("keybase.1")
+
+protocol identify3 {
+
+  import idl "identify3_common.avdl";
+
+  // identify3 is called by React to start an ID3 session. It gets an immediate reply on the basis
```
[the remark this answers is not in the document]
maxtaco: frontend it is

go/service/identify3.go
```diff
@@ -0,0 +1,214 @@
...
+
+func newIdentify3State(g *libkb.GlobalContext) *identify3State {
+	cache := ramcache.New()
+	cache.MaxAge = 24 * time.Hour
+	return &identify3State{
+		cache: cache,
+	}
+}
+
+func (s *identify3State) get(key keybase1.Identify3GUIID) (*identify3Session, error) {
```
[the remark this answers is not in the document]
maxtaco: ok, fixing

maxtaco:
Hiding a row would be sketching, the UI would flicker. We can mark it dead mabye?

go/service/main.go
```diff
@@ -69,6 +69,7 @@ type Service struct {
 	teamUpgrader         *teams.Upgrader
 	avatarLoader         avatars.Source
 	walletState          *stellar.WalletState
+	identify3State       *identify3State
```
mlsteele: nuke em on logout

go/service/identify3.go
```diff
@@ -0,0 +1,245 @@
...
+
+// get an identify3Session out of the cache, as keyed by a Identify3GUIID. Return
+// (nil, nil, nil) if not found. Return (nil, nil, Error) if there was an expected error.
+// Return (i, f, nil) if found, where i is the **locked** object, f is the unlocker function
+// that you are required to call if non-nil, and nil is the nil error.
+func (s *identify3State) get(key keybase1.Identify3GUIID) (ret *identify3Session, unlocker func(), err error) {
+	s.Lock()
+	defer s.Unlock()
+	return s.getLocked(key)
+}
```
mlsteele: I think it's nice for `unlocker` to [always be callable](https://github.com/keybase/client/blob/master/go/stellar/global.go#L455). That way [the caller](https://github.com/keybase/client/blob/master/go/stellar/build.go#L37) can just `defer unlocker()` no matter what and not think about it too much.

mlsteele:
> Hiding a row would be sketchy, the UI would flicker. We can mark it dead maybe?

I don't think it would be sketchy. But sure. Could just throw in a meta tag.

How does the frontend decide which of these actions to show?
image
image

chrisnojima:
we show the ignore/accept if there are diffs. you could explicitly tell us whats available

chrisnojima:
aka a new identifyui callback thats needsRefollow or something, only happens if you discover we should show that second panel

chrisnojima:
(or could piggy pack on the row update)

maxtaco:
I think there should be another call in the UI protocol, which is like `identify3Result(someNewEnum)`.  Until then, the buttons ought to be grayed out. Then this call comes in with "All Good", "Broken", or "good but can be upgraded" result. Then the frontend renders the right buttons.

chrisnojima:
sounds great

maxtaco:
@mlsteele ok to merge?

PR #15141, opened 2018-12-18
show bottom image on mobile CORE-9690

mmaxim:
@keybase/react-hackers 

If a generic unfurl has a "bottom image", then we show it on mobile (including videos). Add start/stop ability to videos on mobile as well.


shared/chat/conversation/messages/wrapper/unfurl/image/video.native.js
```diff
@@ -1,27 +1,92 @@
...
+  absoluteContainer: {
+    left: 0,
+    position: 'absolute',
+    top: 0,
+  },
+  container: {
+    alignSelf: 'flex-start',
+    position: 'relative',
+  },
+  playButton: {
```
buoyad: :sheep: you could get rid of this if you make `absoluteContainer` a `Box2 centerChildren={true}`

shared/chat/conversation/messages/wrapper/unfurl/image/video.native.js
```diff
@@ -1,27 +1,92 @@
...
+  playingVideo: boolean,
+}
+
+export class Video extends React.Component<Props, State> {
+  webviewRef: any
+  state = {playingVideo: this.props.autoPlay}
+
+  constructor(props: Props) {
+    super(props)
+    this.webviewRef = React.createRef()
```
buoyad: You can do this at the top level and get rid of this constructor

PR #15143, opened 2018-12-18
Stellar: Hook up builtPayment reviewing

cjb:
@keybase/react-hackers CC @mlsteele 

This PR goes alongside #15142.

It splits up `readyToSend`, adding `readyToReview`.  The button on the send form becomes `readyToReview`.  When it's clicked, we call `reviewPaymentLocal`, and receive a set of review banners (now `state.wallets.builtPayment.reviewBanners`) and a `readyToSend` that can be `spinning | disabled | enabled`.  The send button on the *confirm* screen becomes linked to that new `readyToSend`.

I haven't been able to test review banners for the identify failing.  I added some of Coyne's proofs to my /etc/hosts as 127.0.0.1, db nuked, restarted service, and saw him show up as red in chats, and even on the Stellar send form itself, but then a reviewPayment to him succeeded just fine, so I haven't been able to provoke a non-null reviewBanners.  @mlsteele, any tips?


mlsteele:
Review only does the identify as of [56a7509](https://github.com/keybase/client/pull/15063). Looks like you've got that.

For a review to fail the recipient would need to be in a state that breaks your tracking of them. The recipient can have a failing proof and still pass review. But I think that's what chat does too.

It seems to work for me. Here's what I did
1. Long ago followed cjb
2. Added `127.0.0.1 printf.net` to /etc/hosts/
3. `keybase id -s cjb` to punch through identify cache.  Saw `Proof for printf.net failed`.
4. Open the app and go to send confirmation screen. Saw review banner.
image


Just as an example if you followed me [and your latest follow of me is] before I proved mlsteele@twitter and then mlsteele@twitter breaks or is revoked, the review will _pass_. Because that does not break your tracking.

Did you break one of Coyne's proofs that you are tracking? Not one of the 'New' ones?

mlsteele:
There's a global error here. From `reviewPaymentLocal` having been canceled.
```
Error: ERROR CODE 237 - Canceled in method stellar.1.local.reviewPaymentLocal at new RPCError
```

There are at least two ways to trigger it. In both cases review needs to be failing.
1. Send form with $5 to cjb
2. Click "Send", see confirmation screen with "Please review"
3. Click "Back" and change the amount field. OR. Click outside the form to close the form (sometimes need to click twice).
4. Global error

This happens because either `stopBuildPaymentLocal` or `buildPaymentLocal` will cancel any outstanding `reviewPaymentLocal`s. When the review is failing it will still be open. To fix this the gui should suppress cancelation errors from `reviewPaymentLocal` when it's expected to get canceled. Which I guess starts right before a stop or build call. Maybe [like this](https://github.com/keybase/client/blob/cjb/DESKTOP-8556-reviewpayment/shared/actions/wallets.js#L66)? Basically `buildPaymentLocal` and `reviewPaymentLocal` share an outstanding call slot.

cjb:
@mlsteele Thanks! Got a failing review going, and fixed the cancelation black bar by ignoring the error.

mlsteele:
Cool!
1. Oops it's broken for sending to a *keybase.io federated address. My bad.
2. If you click "back" and then forward again, the final send button should be spinning. As it stands it is enabled. It's hard to notice when the service is acting normally. But I put in a delay and it's clickable. Which means I think a user could click it on a failing review. So anyway the "back" button should disable the final send button.

mlsteele:
I tried sending to an account ID and got it hung here
image
Service said
```
2018-12-18T15:48:23.632237-05:00 ▶ [DEBU keybase build.go:338] 24cb sending UIPaymentReview bid:bb9248cc9b06c71ea5ca1d945bfebf08 sessionID:303 seqno:2 nextButton:enabled banners:0
2018-12-18T15:48:23.634242-05:00 ▶ [DEBU keybase build.go:449] 24cc - ReviewPaymentLocal [time=5.005193361s]
2018-12-18T15:48:23.634340-05:00 ▶ [DEBU keybase local.go:1498] 24cd - LRPC ReviewPaymentLocal -> <nil> [time=5.005505031s] 
2018-12-18T15:48:23.634474-05:00 ▶ [DEBU keybase build.go:354] 24ce error in response to UIPaymentReview: context canceled
```
meaning the UI rejected the notification

shared/reducers/wallets.js
```diff
@@ -232,6 +247,7 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
         accountName: '',
         accountNameError: '',
         accountNameValidationState: 'none',
+        builtPayment: state.get('builtPayment').merge({readyToSend: 'spinning'}),
```
chrisnojima: you can just do `state.builtPayment` and get rid of the gets (here and above)

mlsteele:
I can debug on top of this but don't know enough to merge master in

mlsteele:
@buoyad Pushed a fix to that the out of order reviewedPayment issue. This PR should be ready.

shared/reducers/wallets.js
```diff
@@ -99,14 +99,27 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
           stateMutable.update('building', b => b.merge({currency}))
         }
       })
-    case WalletsGen.reviewedPaymentReceived:
-      const {reviewBanners, readyToSend} = action.payload
+    case WalletsGen.reviewPayment:
+      return state.set('reviewCounter', state.reviewCounter + 1).set('reviewLastSeqno', null)
```
buoyad: can you reset `builtPayment.readyToSend` and `builtPayment.reviewBanners` here too?

PR #15147, opened 2018-12-18
need space for unverified inbox too or this blocks forever

mmaxim:
cc @chrisnojima this fixed the blank state getting stuck.


joshblum:
leave a comment as to why `+1`

PR #15079, opened 2018-12-13
add public box summaries to keyset uploads

maxtaco:
- we might want to disable write access while we let this bake for a bit
- but it posts successfully and populates the relevant tables..
- note there's no good way for the clients to audit this, since they don't get the boxes down from the server, but that's OK

PR #15148, opened 2018-12-18
futher cleanups to as a follow-on to #15079

maxtaco:


maxtaco:
CC @mlsteele as an FYI

maxtaco:
BTW, it works locally with `mm.config.box_public_summary.required = true`, which is good

PR #15151, opened 2018-12-18
Update deps

chrisnojima:
- [x] Update memoize / flow
- [x] react-measure
- [x] react-native-contacts, remove gradle fix
- [x] react-transition-group
- [x] rn-fetch-blob
- [x] electron (upped to kbelectron)
- [x] electron-packager (kick test builds)
- [x] react-hot-loader


chrisnojima:

PR #15152, opened 2018-12-18
Initial drag & drop for Files tab

taruti:
WIP
This adds drag & drop to the current directory in Files tab. Please check the approach, when it is ok then I'll convert chat/conversation/normal/index.desktop.js to use this copy of the code in common-adaptors.


shared/fs/index.js
```diff
@@ -17,10 +17,16 @@ type FolderProps = {
...
   routePath: I.List<string>,
+  onAttach: (path: Types.Path, local: Array<string>) => void,
 }
 
-class Files extends React.PureComponent<FolderProps> {
+class Files extends React.PureComponent<FolderProps, State> {
+  //      this.props.onAttach(this.props.path, paths)
+
   render() {
+    // TODO check whether the path is writable by the user.
```
songgao: [Here](https://github.com/keybase/client/blob/1f621f5e35f0e77c49c0d507667e7b02c2ababbf/shared/fs/header/add-new-container.js#L26)'s how we did it in the "Add New" button.

shared/fs/index.js
```diff
@@ -17,10 +17,16 @@ type FolderProps = {
   path: Types.Path,
   resetParticipants: Array<string>,
   routePath: I.List<string>,
+  onAttach: (path: Types.Path, local: Array<string>) => void,
 }
 
-class Files extends React.PureComponent<FolderProps> {
+class Files extends React.PureComponent<FolderProps, State> {
+  //      this.props.onAttach(this.props.path, paths)
```
songgao: :)

shared/common-adapters/drop-file-box.desktop.js
```diff
@@ -0,0 +1,96 @@
...
+  alignItems: 'center',
+  backgroundColor: globalColors.blue_60,
+  bottom: 0,
+  flex: 1,
+  justifyContent: 'center',
+  left: 0,
+  position: 'absolute',
+  right: 0,
+  top: 0,
+}
```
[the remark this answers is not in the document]
songgao: Also please use `Styles.styleSheetCreate`

shared/common-adapters/drop-file-box.desktop.js
```diff
@@ -0,0 +1,96 @@
+// @flow
+import * as React from 'react'
+import fs from 'fs'
+import logger from '../logger'
+import Icon from './icon'
+import {Box, Box2} from './box'
+import {globalStyles, globalColors} from '../styles'
```
songgao: I think the preferred way to import this became `import * as Styles from '../styles'` at some point because we always ended up importing a bunch of stuff.

shared/common-adapters/drop-file-box.desktop.js
```diff
@@ -0,0 +1,96 @@
...
+    if (paths.length) {
+      for (let path of paths) {
+        // Check if any file is a directory and bail out if not
+        try {
+          // We do this synchronously
+          // in testing, this is instantaneous
+          // even when dragging many files
+          const stat = fs.lstatSync(path)
+          if (stat.isDirectory()) {
+            // TODO show a red error banner on failure: https://zpl.io/2jlkMLm
```
songgao: Is there a reason why we can't drag&drop a directory? Upload supports it.

shared/common-adapters/drop-file-box.desktop.js
```diff
@@ -0,0 +1,96 @@
...
+    const fileList = e.dataTransfer.files
+    const paths = fileList.length ? Array.prototype.map.call(fileList, f => f.path) : []
+    if (paths.length) {
+      for (let path of paths) {
+        // Check if any file is a directory and bail out if not
+        try {
+          // We do this synchronously
+          // in testing, this is instantaneous
+          // even when dragging many files
+          const stat = fs.lstatSync(path)
```
songgao: I wonder if we need to consider the possibility that this could be from a KBFS mount, in which case it'd be slow. I guess in that case we could probably do a link for chat, and copy for Files. Not sure though!

shared/common-adapters/drop-file-box.desktop.js
```diff
@@ -0,0 +1,96 @@
...
+  state = {showDropOverlay: false}
+
+  _onDrop = e => {
+    if (!this._validDrag(e)) {
+      return
+    }
+    const fileList = e.dataTransfer.files
+    const paths = fileList.length ? Array.prototype.map.call(fileList, f => f.path) : []
+    if (paths.length) {
+      for (let path of paths) {
```
songgao: Maybe `paths.forEach`?

shared/common-adapters/drop-file-box.desktop.js
```diff
@@ -0,0 +1,96 @@
...
+
+const containerStyle = {
+  ...globalStyles.flexBoxColumn,
+  flex: 1,
+  position: 'relative',
+}
+
+const dropOverlayStyle = {
+  ...globalStyles.flexBoxColumn,
+  alignItems: 'center',
```
songgao: For `Box2` you could do `direction="vertical" centerChildren={true}`

shared/fs/index.js
songgao: Also since this would be done in the container, it seems we could also just make `onAttach` a `null` if it shouldn't be enabled.

songgao:
@taruti I just merged https://github.com/keybase/client/pull/15088 which moved `fs/index.js` and `fs/container.js` into `fs/folder`, which causes annoying conflicts for this PR. So I've made https://github.com/keybase/client/pull/15178 against this PR to resolve the conflict. Feel free to take it if it looks good, or throw it away if you'd rather do it yourself. Sorry for the inconvenience!

shared/fs/container.js
```diff
@@ -14,7 +15,15 @@ const mapStateToProps = (state, {routeProps}) => {
   }
 }
 
-const mergeProps = (stateProps, dispatchProps, {routeProps, routePath}) => {
+const mapDispatchToProps = dispatch => ({
+  onAttach: (parentPath: Types.Path, paths: Array<string>) => {
+    for (const localPath of paths) {
```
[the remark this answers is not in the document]
taruti: Done.

taruti:
Ended up rewriting the whole thing.

shared/fs/row/drop-target.native.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+import {Box} from '../../common-adapters'
+
+const DropTarget = Box
+
+export default DropTarget
```
songgao: Could merge this and the desktop, but just export Box if `isMobile`. So we don't have to have a `.flow` file here.

shared/fs/row/drop-target.desktop.js
```diff
@@ -0,0 +1,75 @@
...
+  showDropOverlay: boolean,
+|}
+
+const DropOverlay = ({onDragLeave, onDrop}) => (
+  <Box style={styles.dropOverlayStyle} onDragLeave={onDragLeave} onDrop={onDrop}>
+    <Icon type="icon-dropping-file-48" />
+  </Box>
+)
+
+class DropTarget extends React.PureComponent<any, State> {
```
songgao: Mind type the props rather than using `any`?

shared/fs/row/rows-container.js
```diff
@@ -177,14 +178,20 @@ const mapStateToProps = state => ({
...
+})
 
 // $FlowIssue
 const mergeProps = (s, d, o: OwnProps) => ({
   destinationPickerIndex: o.destinationPickerIndex,
   items: getItemsFromStateProps(s, o.path, o.sortSetting),
+  onAttach: d.onAttach,
   path: o.path,
   routePath: o.routePath,
+  writable: s._pathItems.get(o.path, Constants.unknownPathItem).writable,
```
songgao: Instead of adding stuff here, can we instead make `DropTarget` itself a connected component? This component is also used in destination-picker, and it'd get complicated if we start adding these in.

shared/fs/row/drop-target.desktop.js
```diff
@@ -0,0 +1,75 @@
...
+    this.setState({showDropOverlay: false})
+  }
+
+  render() {
+    const {children, onAttach, ...otherProps} = this.props
+    return (
+      <Box style={styles.containerStyle} onDragOver={onAttach ? this._onDragOver : null} {...otherProps}>
+        {children}
+        {this.state.showDropOverlay && <DropOverlay onDragLeave={this._onDragLeave} onDrop={this._onDrop} />}
+      </Box>
```
songgao: I wonder if it's possible to have the `DropOverlay` along side with `Rows`, instead of having them both as children in here.

shared/fs/row/drop-target.native.js
taruti: Done.

shared/fs/row/drop-target.desktop.js
taruti: Done.

shared/fs/row/rows-container.js
taruti: Avoided writable by setting onAttach only if it is writable. Thus no extra props.

shared/fs/row/drop-target.desktop.js
taruti: Not sure that is possible. Now it only has the extra box if necessary.

shared/fs/row/rows-container.js
songgao: It'd still get used by destination-picker though. I think making `DropTarget` connected is cleaner, because 1) The Rows component should only deal with rows, but not stuff like drag and drop; 2) The Rows component is used in the destination picker too; having it in Rows would cause the destination picker to be drag/drop-able as well, which I assume we don't want; 3) if we ever want this to be used somewhere else, we a decoupled `DropTarget` that's connected by itself will be easier to use.

shared/fs/row/drop-target.desktop.js
songgao: Could probably add a `<DropOverlay />` [here](https://github.com/keybase/client/blob/aa76cc7a1c4e06e80210e514a5650cb59b1978ab/shared/fs/folder/index.js#L32) above `<Rows />`? But I don't feel strongly about sibling vs children, as long as the Rows component doesn't need to know about drag/drop.

shared/fs/row/drop-target.js
```diff
@@ -0,0 +1,81 @@
...
+    this.setState({showDropOverlay: false})
+  }
+
+  render() {
+    const {children, onAttach} = this.props
+    return (
+      <Box style={styles.containerStyle} onDragOver={onAttach ? this._onDragOver : null}>
+        {children}
+        {this.state.showDropOverlay && <DropOverlay onDragLeave={this._onDragLeave} onDrop={this._onDrop} />}
+      </Box>
```
songgao: If you decide on using the `children`, I think this could also be like this:

```
<>
  {children}
  {this.state.showDropOverlay && <DropOverlay onDragLeave={this._onDragLeave} onDrop={this._onDrop} />}
</>
```

The reason why I'm suggesting this is this way `Rows` would always be inside the same parent Box, and it seems it's less likely CSS goes wrong in different scenarios.

shared/fs/row/drop-target.desktop.js
taruti: But there still needs to be a Box with onDragOver to trigger the DropOverlay (and Box2 does not work unless extended).

taruti: But the DropTarget can be moved fs/folder/index.js of course.

shared/fs/folder/container.js
```diff
@@ -5,13 +5,21 @@ import Files from '.'
...
+  _pathItems: state.fs.pathItems,
   _tlfs: state.fs.tlfs,
   _username: state.config.username,
   sortSetting: state.fs.pathUserSettings.get(path, Constants.makePathUserSetting()).get('sort'),
 })
 
+const mapDispatchToProps = dispatch => ({
+  onAttach: (parentPath: Types.Path, paths: Array<string>) => {
+    paths.forEach(localPath => dispatch(FsGen.createUpload({localPath, parentPath})))
+  },
```
songgao: You could just make `onAttach` accept a single `paths` param here since `path` is already available from `ownProps` here. Then you don't need to make an `onDrop` [here](https://github.com/keybase/client/pull/15152/files#diff-6df9d6db4b7ccf7cbc83bcd2ba7ee713R26) anymore.

shared/fs/folder/drop-target.js
```diff
@@ -0,0 +1,81 @@
...
+    if (!this._validDrag(e)) {
+      return
+    }
+    const fileList = e.dataTransfer.files
+    const paths = fileList.length ? Array.prototype.map.call(fileList, f => f.path) : []
+    this.props.onAttach(paths)
+    this.setState({showDropOverlay: false})
+  }
+
+  _validDrag = e => Array.prototype.map.call(e.dataTransfer.types, t => t).includes('Files')
```
songgao: Does `e.dataTransfer.types.includes('Files')` not work?

shared/fs/folder/drop-target.js
```diff
@@ -0,0 +1,81 @@
...
+
+class DropTarget extends React.PureComponent<Props, State> {
+  state = {showDropOverlay: false}
+
+  _onDrop = e => {
+    if (!this._validDrag(e)) {
+      return
+    }
+    const fileList = e.dataTransfer.files
+    const paths = fileList.length ? Array.prototype.map.call(fileList, f => f.path) : []
```
songgao: Could do `fileList.map(f=>f.path)`; also, is `e.dataTransfer.files` ever null? If not, could just do `const paths = e.dataTransfer.files.map(f => f.path)` here instead of checking length.

shared/fs/folder/drop-target.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  _onDragLeave = e => {
+    this.setState({showDropOverlay: false})
+  }
+
+  render() {
+    const {children, onAttach} = this.props
+    return (
+      <Box style={styles.containerStyle} onDragOver={onAttach ? this._onDragOver : null}>
```
songgao: Have you tried replacing `<Box/>` with `<>` (`React.Fragment`)? Not a strong opinion though, so if you like the Box better feel free to keep it.

shared/fs/folder/container.js
taruti: Done.

shared/fs/folder/drop-target.js
taruti: Done.

shared/fs/folder/drop-target.js
taruti: Added a comment explaining this. files is a FileList - not an array, thus a .map won't work.

shared/fs/folder/drop-target.js
taruti: Avoided the Box when not using the onDragOver.

PR #15156, opened 2018-12-18
Fix 'claim lumens' button not showing up

buoyad:
#14750 renamed status 'cancelable' to 'claimable', but we still had some checks for cancelable hanging around. In particular, these mean that no 'claim lumens' button would show in chat for recipients of relay payments. r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/account-payment/container.js
+++ b/shared/chat/conversation/messages/account-payment/container.js
@@ -66,7 +66,7 @@ const mapStateToProps = (state, ownProps: OwnProps) => {
       const conv = Constants.getMeta(state, ownProps.message.conversationIDKey)
       const theirUsername = conv.participants.find(p => p !== you) || ''
 
-      const cancelable = paymentInfo.status === 'cancelable'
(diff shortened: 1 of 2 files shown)


patrickxb:
StatusSimplified has an enum in it.  Would that be easier to use?

patrickxb:
there's also a `showCancel` field.

buoyad:
@patrickxb flow restricts us to only using [one of these statuses](https://github.com/keybase/client/blob/master/shared/constants/types/wallets.js#L121), but doesn't care if we compare it to bogus values. `showCancel` is why this didn't break for the sending side too.

patrickxb:
ok, well if there's anything that would make it more robust, let me know.  We could add `showClaim` if you want.

buoyad:
@patrickxb I think the API is alright, `showClaim` might be more robust but I'd hesitate to add a bool to control a UI part this small. Though that might change with another ticket we have this week

buoyad:
@keybase/react-hackers bump

PR #15158, opened 2018-12-18
Surya/core 9529/restart on linux upgrade

heronhaye:
Still a WIP, just wanted to get some feedback on the general strategy in `systemd_safe_restart_services`. Might want to look at diffs by commit as well.

TODO

- [x] Tighten up shell code to be more defensive if possible
- [x] Test on ubuntu/centos/arch with and without systemd
- [x] CLI warning on kbfs client/binary mismatch

### DIFF
--- a/go/install/install_windows.go
+++ b/go/install/install_windows.go
@@ -356,3 +356,8 @@ func StartUpdateIfNeeded(ctx context.Context, log logger.Logger) error {
 	}
 	return nil
 }
+
+func LsofMount(mountDir string, log Log) ([]CommonLsofResult, error) {
+	log.Warning("Cannot use lsof on Windows.")
+	return nil, fmt.Errorf("Cannot use lsof on Windows.")
+}
--- a/packaging/linux/deb/package_binaries.sh
+++ b/packaging/linux/deb/package_binaries.sh
@@ -32,7 +32,9 @@ elif [ "$mode" = "prerelease" ] ; then
   repo_url="http://prerelease.keybase.io/deb"
   # "psmisc" provides "killall", which is used in run_keybase and
   # post_install.sh.
-  dependencies="Depends: libappindicator1, fuse, libgconf-2-4, psmisc"
+  # lsof used in post_install.sh
+  # systemd-container provides machinectl, which is used in post_install.sh
+  dependencies="Depends: libappindicator1, fuse, libgconf-2-4, psmisc, lsof, procps"
 elif [ "$mode" = "staging" ] ; then
   # Note: This doesn't exist yet. But we need to be distinct from the
   # production URL, because we're moving to a model where we build a clean repo
--- a/packaging/linux/post_install.sh
+++ b/packaging/linux/post_install.sh
@@ -55,7 +55,7 @@ if [ -n "$currlink" ] ; then
     fi
 elif [ -d "$rootmount" ] ; then
     # Handle upgrading from old builds that don't have the rootlink.
-    currowner=`stat -c %U "$rootmount"`
+    currowner="$(stat -c %U "$rootmount")"
     if [ "$currowner" != "root" ]; then
         # Remove any existing legacy mount.
         echo Unmounting $rootmount...
@@ -65,25 +65,25 @@ elif [ -d "$rootmount" ] ; then
         rmdir "$rootmount"
         echo You must run run_keybase to restore file system access.
     elif ! redirector_enabled ; then
-        if killall `basename "$krbin"` &> /dev/null ; then
+        if killall "$(basename "$krbin")" &> /dev/null ; then
             echo "Stopping existing root redirector."
         fi
-    elif killall -USR1 `basename "$krbin"` &> /dev/null ; then
+    elif killall -USR1 "$(basename "$krbin")" &> /dev/null ; then
         echo "Restarting existing root redirector."
         # If the redirector is still owned by root, that probably
         # means we're sill running an old version and it needs to be
         # manually killed and restarted.  Instead, run it as the user
         # currently running kbfsfuse.
-        krName=`basename "$krbin"`
-        krUser=`ps -o user= -C "$krName" 2> /dev/null | head -1`
+        krName="$(basename "$krbin")"
+        krUser="$(ps -o user= -C "$krName" 2> /dev/null | head -1)"
         if [ "$krUser" = "root" ]; then
-            newUser=`ps -o user= -C "kbfsfuse" 2> /dev/null | head -1`
+            newUser="$(ps -o user= -C "kbfsfuse" 2> /dev/null | head -1)"
             killall "$krName" &> /dev/null
-            if [ -n "$newUser" -a "$newUser" != "root" ]; then
+            if [ -n "$newUser" ] && [ "$newUser" != "root" ]; then
                 # Try our best to get the user's $XDG_CACHE_HOME,
                 # though depending on how it's set, it might not be
                 # available to su.
-                userCacheHome=`su -c 'echo -n $XDG_CACHE_HOME' - $newUser`
+                userCacheHome="$(su -c "echo -n \$XDG_CACHE_HOME" - "$newUser")"
                 log="${userCacheHome:-~$newUser/.cache}/keybase/keybase.redirector.log"
                 su -c "nohup \"$krbin\" \"$rootmount\" &>> $log &" "$newUser"
                 echo "Root redirector now running as $newUser."
@@ -97,13 +97,13 @@ elif [ -d "$rootmount" ] ; then
                 logdir="${XDG_CACHE_HOME:-$HOME/.cache}/keybase"
                 mkdir -p "$logdir"
                 log="$logdir/keybase.redirector.log"
-                nohup "$krbin" "$rootmount" &>> $log &
+                nohup "$krbin" "$rootmount" &>> "$log" &
             fi
         fi
         t=5
         while ! mountpoint "$rootmount" &> /dev/null; do
             sleep 1
-            t=$[t-1]
+            t=$((t-1))
             if [ $t -eq 0 ]; then
                 echo "Redirector hasn't started yet."
                 break
@@ -116,6 +116,6 @@ fi
 make_mountpoint
 
 # Update the GTK icon cache, if possible.
-if which gtk-update-icon-cache &> /dev/null ; then
+if command -v gtk-update-icon-cache &> /dev/null ; then
   gtk-update-icon-cache -q -t -f /usr/share/icons/hicolor
 fi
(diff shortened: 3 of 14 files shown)


packaging/linux/post_install.sh
```diff
@@ -32,6 +33,68 @@ make_mountpoint() {
...
+    pids="$(pidof /usr/bin/keybase)"
+    while read -r pid; do
+        user="$(ps -o user= -p "$pid")"
+
+        if systemd_unit_active_for "$user" "kbfs.service"; then
+            # KBFS running, need to check mounts before proceeding
+
+            # Don't use `mount(8)` to find kbfs mounts in case there are non-kbfs fuse systems
+            mount="$(exec_as "$user" /usr/bin/keybase config get -b mountdir | tr -d '\r')"
+            uses="$(exec_as "$user" /usr/bin/lsof "$mount")"
```
heronhaye: can't use exit code because lsof returns zero when something is found, and 1 when either nothing was found or there's an error.

heronhaye: What do you think about changing the logic to "if lsof successfully returns in-use files, don't restart. Otherwise, if lsof hits an error or there are no files in use, do restart". A bit more aggressive.

packaging/linux/deb/package_binaries.sh
```diff
@@ -32,7 +32,9 @@ elif [ "$mode" = "prerelease" ] ; then
   repo_url="http://prerelease.keybase.io/deb"
   # "psmisc" provides "killall", which is used in run_keybase and
   # post_install.sh.
-  dependencies="Depends: libappindicator1, fuse, libgconf-2-4, psmisc"
+  # lsof used in post_install.sh
+  # systemd-container provides machinectl, which is used in post_install.sh
+  dependencies="Depends: libappindicator1, fuse, libgconf-2-4, psmisc, lsof, systemd-container"
```
strib: Do you think we'll run into situations where someone wants to turn off systemd support, but this forces them to include the package?  Not sure if we care about that, maybe we're moving towards a systemd-only world anyway?

packaging/linux/post_install.sh
```diff
@@ -37,6 +33,68 @@ make_mountpoint() {
...
+systemd_safe_restart_services() {
+    pids="$(pidof /usr/bin/keybase)"
+    while read -r pid; do
+        user="$(ps -o user= -p "$pid")"
+
+        if systemd_unit_active_for "$user" "kbfs.service"; then
+            # KBFS running, need to check mounts before proceeding
+
+            # Don't use `mount(8)` to find kbfs mounts in case there are non-kbfs fuse systems
+            mount="$(exec_as "$user" /usr/bin/keybase config get -b mountdir | tr -d '\r')"
```
strib: Seems like passing `-d` to `config get` here would be a win, even if we already know the daemon is running.  Though maybe there's some race that can happen and it gets a stale value from the config file?

packaging/linux/post_install.sh
```diff
@@ -37,6 +33,68 @@ make_mountpoint() {
...
+            uses="$(exec_as "$user" /usr/bin/lsof "$mount")"
+            if [ -n "$uses" ]; then
+                using_programs="$(echo "$uses" | tail -n +2 | awk '{print $1}' | tr '\n' ', ')"
+                echo "KBFS mount $mount for user $user currently in use by ($using_programs)."
+                echo "$uses"
```
strib: Doesn't the previous line tell us what we need to know?

packaging/linux/post_install.sh
strib: Is the issue that the error will be present in `$uses`, so we'd still abort as below?

What are some cases where `lsof` returns an error?  Is it basically just the mount not existing?  In that case it seems fine to restart, so I think I agree with your proposal.

packaging/linux/rpm/package_binaries.sh
```diff
@@ -114,12 +114,12 @@ export debian_arch=i386
 # found.  "psmisc" provides "killall", which is used in run_keybase.
 # "initscripts" provides "service", which is used to start atd in the
 # post-install.
-dependencies="Requires: at, fuse, libXss.so.1, /sbin/service, psmisc"
+dependencies="Requires: at, fuse, libXss.so.1, /sbin/service, psmisc, lsof"
```
strib: redhat distros come with `machinectl` for sure?

packaging/linux/post_install.sh
heronhaye: That's one reason but there could be others in which case it might be unsafe, not sure. But this seems like a decent approach. I don't think the error would be in uses, it'd be in stdout. I just meant that that's why I used `-n $uses` instead of checking `$?`. That being said, the exit code isn't even being preserved by machinectl (I have a workaround for that though).

packaging/linux/deb/package_binaries.sh
heronhaye: Yeah I was worried about this exactly. I'd like to do a "systemd-container only if systemd is installed" but I doubt it's possible. In either case I think it should be possible to restart for non-systemd too.

I'm just not sure how to figure out whether they are using systemctl to run keybase without systemd-container. `su` actually does work if you pass in `XDG_RUNTIME_DIR` and `DBUS_SESSION_BUS_ADDRESS` (those are set by systemd I believe), but the only way to know that for a user is to use `machinectl`...

packaging/linux/post_install.sh
heronhaye: I wasn't sure what to do here, I don't think a race would be very common. Since keybase is already running, what's the issue with contacting service? Just for speed?

heronhaye: I suppose the service could be in some weird state of alive but not responding though. There are a lot of edge cases here, I'll add error checking to all these commands.

strib: Yeah, just speed.  I just don't see a real downside to putting `-d` in.

packaging/linux/post_install.sh
heronhaye: Yeah they both work, The first line just says `tail, awk` etc. and this one says the entire lsof output. I can get rid of one or both if you like, or maybe add some more explanation.

packaging/linux/post_install.sh
strib: Ah ok, cool.  This way is fine with me for now.  I thought you were proposing some change that you hadn't coded yet.

packaging/linux/rpm/package_binaries.sh
heronhaye: Will check.

packaging/linux/post_install.sh
strib: I guess I'd have to see sample output, but seems like you can cut this line and not lose anything important, if there wasn't an error.

go/install/install_windows.go
```diff
@@ -356,3 +356,8 @@ func StartUpdateIfNeeded(ctx context.Context, log logger.Logger) error {
 	}
 	return nil
 }
+
+func LsofMount(mountDir string, log Log) []lsof.Process {
```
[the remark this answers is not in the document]
heronhaye: (could also refactor this to return an error)

packaging/linux/post_install.sh
```diff
@@ -37,6 +43,80 @@ make_mountpoint() {
...
+    # Keybase supports running with and without systemd. In order for `su`
+    # to work with systemd commands, we need to pass in XDG_RUNTIME_DIR.
+    # /run/user/{uid} is the default when using systemd. We can't know the
+    # true value without installing systemd itself, which we don't want to
+    # depend on, so we use this default. Users who want to use a custom
+    # XDG_RUNTIME_DIR should specify it in their .profile so it is known
+    # by su -, which doesn't load pam_systemd. Unfortunately, this also
+    # results in an extra su invocation.
+    user_xdg_runtime_dir=""
+    # shellcheck disable=SC2016
```
strib: Remove?

packaging/linux/post_install.sh
```diff
@@ -37,6 +43,80 @@ make_mountpoint() {
...
+    # Intentionally do not expand $XDG_RUNTIME_DIR; we want the user's shell to expand it
+    if ! user_xdg_runtime_dir="$(su --login "$user" -c 'echo $XDG_RUNTIME_DIR')" || [ -z "$user_xdg_runtime_dir" ]; then
+        user_xdg_runtime_dir="/run/user/$(id -u "$user")" || return 1
```
strib: Cool, I didn't know you could `||` a variable setting like this.

packaging/linux/post_install.sh
```diff
@@ -37,6 +43,80 @@ make_mountpoint() {
...
+        restart_instructions="Aborting Keybase autorestart for $user. Restart manually by running 'run_keybase' as $user."
+
+        if ! systemd_unit_active_for "$user" "keybase.service"; then
+            echo "Keybase not running via systemd for $user."
+            echo "$restart_instructions"
+            continue
+        fi
+
+        if systemd_unit_active_for "$user" "kbfs.service"; then
+            if ! mount="$(systemd_exec_as "$user" "/usr/bin/keybase config get --bare mountdir")" || [ -z "$mount" ]; then
```
strib: Maybe a quick comment here about why we're not using `-d`?

packaging/linux/post_install.sh
```diff
@@ -37,6 +43,80 @@ make_mountpoint() {
...
+    while read -r pid; do
+        # Since keybase is running, we can assume run_keybase has been run before
+        # and the mountdir is configured (so, it is not a fresh install).
+        user="$(ps -o user= -p "$pid")"
+
+        restart_instructions="Aborting Keybase autorestart for $user. Restart manually by running 'run_keybase' as $user."
+
+        if ! systemd_unit_active_for "$user" "keybase.service"; then
+            echo "Keybase not running via systemd for $user."
+            echo "$restart_instructions"
```
strib: I think saying "Aborting Keybase autorestart" in the case where systemd isn't running is a bit aggressive, since it feels like an error when really the user configured it that way.  Can we soften it in the non-error cases to something like "Not auto-restarting Keybase..."?

packaging/linux/post_install.sh
heronhaye: I like to leave these in so people in the future don't try to 'fix' something that's intentional

strib: Hmm, I have no idea what this line would do if its uncommented.  Can you add an explanation of what not to do?

heronhaye: It's just a message to the shellcheck linter to ignore the fact that `'echo $XDG_RUNTIME_DIR'` doesn't expand since it's in single quotes on line 61. Without that line it would just show a lint warning if you have shellcheck enabled in your editor.

PR #15159, opened 2018-12-18
Send `kb_ua` for generic proofs

joshblum:
depends on https://github.com/keybase/keybase/pull/3277


joshblum:
will merge over CI after the server PR lands so we don't have a CI failure here

PR #15174, opened 2018-12-19
Fix looking up implicit team with conflicts CORE-9732

zapu:
also more debugging capabilities in `keybase riit`

### DIFF
--- a/go/teams/implicit.go
+++ b/go/teams/implicit.go
@@ -60,34 +60,34 @@ func LookupImplicitTeamAndConflicts(ctx context.Context, g *libkb.GlobalContext,
 
 func LookupImplicitTeamIDUntrusted(ctx context.Context, g *libkb.GlobalContext, displayName string,
 	public bool) (res keybase1.TeamID, err error) {
-	imp, err := loadImpteam(ctx, g, displayName, public, false /* force */)
+	imp, _, err := loadImpteam(ctx, g, displayName, public, false /* force */)
 	if err != nil {
 		return res, err
 	}
 	return imp.TeamID, nil
 }
 
-func loadImpteam(ctx context.Context, g *libkb.GlobalContext, displayName string, public bool, force bool) (imp implicitTeam, err error) {
+func loadImpteam(ctx context.Context, g *libkb.GlobalContext, displayName string, public bool, force bool) (imp implicitTeam, hitCache bool, err error) {
 	cacheKey := impTeamCacheKey(displayName, public)
 	cacher := g.GetImplicitTeamCacher()
 	if !force && cacher != nil {
 		if cv, ok := cacher.Get(cacheKey); ok {
 			if imp, ok := cv.(implicitTeam); ok {
 				g.Log.CDebugf(ctx, "using cached iteam")
-				return imp, nil
+				return imp, true, nil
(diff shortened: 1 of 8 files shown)


zapu:
Max I merged your changes, added some logs/tweaks. Notably, it will only retry loading only when there are conflicts. We wouldn't want to retry on complete failures, because we do not cache bad loads. So if someone wants to lookup 100% nonexisting impteam (not even a non-existing conflict of existing impteam), there's no reason to try twice.

zapu:
That being said, maybe there's another way to do it, because it became a bit messy. I'll spend another hour exploring but probably not today.

go/teams/implicit.go
```diff
@@ -113,35 +113,27 @@ func loadImpteamFromServer(ctx context.Context, g *libkb.GlobalContext, displayN
...
 }
 
-// Lookup an implicit team by name like "alice,bob+bob@twitter (conflicted copy 2017-03-04 #1)"
-// Does not resolve social assertions.
-// preResolveDisplayName is used for logging and errors
-func lookupImplicitTeamAndConflicts(ctx context.Context, g *libkb.GlobalContext,
-	preResolveDisplayName string, impTeamNameInput keybase1.ImplicitTeamDisplayName, opts ImplicitTeamOptions) (
-	team *Team, teamName keybase1.TeamName, impTeamName keybase1.ImplicitTeamDisplayName, conflicts []keybase1.ImplicitTeamConflictInfo, err error) {
+// attemptLoadImpteamAndConflits attempts to lead the implicit team with conflict, but it might hit the cache based on
+// the force flag being true. In that case, the caller ought to recall this function with the force flag set to true.
```
maxtaco: should be `the force flag being false`

maxtaco:
Thanks @zapu. looks good!

maxtaco:
maybe there's an easier fix, let me try it

maxtaco:
See the above commit ^

zapu:
looks perfect. we might even remove ` impTeamName.ConflictInfo.IsConflict()`
from the if condition for retrying now

sob., 22 gru 2018 o 19:52 Maxwell Krohn <notifications@github.com>
napisał(a):

maxtaco:
Good to merge when you are: 👍

go/teams/implicit.go
```diff
@@ -60,34 +60,34 @@ func LookupImplicitTeamAndConflicts(ctx context.Context, g *libkb.GlobalContext,
...
-	imp, err := loadImpteam(ctx, g, displayName, public)
+	imp, _, err := loadImpteam(ctx, g, displayName, public, false /* force */)
 	if err != nil {
 		return res, err
 	}
 	return imp.TeamID, nil
 }
 
-func loadImpteam(ctx context.Context, g *libkb.GlobalContext, displayName string, public bool) (imp implicitTeam, err error) {
+func loadImpteam(ctx context.Context, g *libkb.GlobalContext, displayName string, public bool, force bool) (imp implicitTeam, hitCache bool, err error) {
```
joshblum: maybe `forceCache` or `useCache` instead? i feel like `force` would typically imply forcing a server call not a cache call so a more explicit name might be less confusing in the future

PR #15178, opened 2018-12-19
merge master into taruti/drag-and-drop

songgao:


taruti:
Is this better than just rebasing from master before the next iteration? I'm fine with this if you like this more.

PR #15179, opened 2018-12-19
Fix data race

patrickxb:
I think these are only used for tests, so it doesn't really matter, but I was trying to `go test -race` something else and this kept getting in the way.


patrickxb:
Fixed

patrickxb:
@maxtaco ping

PR #15182, opened 2018-12-19
fix docgen crappiness

chrisnojima:
@keybase/react-hackers

### DIFF
--- a/shared/desktop/yarn-helper/index.js
+++ b/shared/desktop/yarn-helper/index.js
@@ -5,6 +5,8 @@ import electronComands from './electron'
 import fontCommands from './font'
 import prettierCommands from './prettier'
 import {execSync} from 'child_process'
+import path from 'path'
+import fs from 'fs'
 
 const [, , command, ...rest] = process.argv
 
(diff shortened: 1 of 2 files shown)


shared/desktop/yarn-helper/index.js
```diff
@@ -26,11 +28,27 @@ const commands = {
   postinstall: {
     code: () => {
       // Nothing right now
```
MarcoPolo: This comment doesn't make sense anymore

shared/desktop/yarn-helper/index.js
```diff
@@ -26,11 +28,27 @@ const commands = {
...
+function makeShims() {
+  const root = path.resolve(__dirname, '..', '..', 'node_modules', 'babel-plugin-react-docgen')
+
+  try {
+    fs.mkdirSync(root)
+  } catch (_) {}
+
+  try {
+    fs.writeFileSync(path.join(root, 'package.json'), `{"main": "index.js"}`)
+    fs.writeFileSync(path.join(root, 'index.js'), `module.exports = function(){return {};};`)
```
MarcoPolo: Nothing expects this to exist?

chrisnojima: ?

MarcoPolo: I mean that you're effectively removing this file, right? so why doesn't that break anything else?

chrisnojima: only story book needs this so this plugin does nothign now. we don't use it. also it uses lib/index.js and not index.js

PR #15192, opened 2018-12-20
React Native Profiling Docs

MarcoPolo:
After diving into profiling our android app, I found the very act of profiling it confusing. It also seems like not many people play around with these things (i.e. I haven't see anyone else make use of RN's native systrace tools). Instead of going on a deeper dive into some of the possible fixes I found, I thought it would be more helpful to make a write up of profiling so others can also contribute ideas and suggestions as well as be empowered to do their own profiling tests.

The write up is pretty basic. It has a list of useful links, some tips, and a couple case studies.

I'm going to somewhat randomly nominate @buoyad to review this for coherency and see if he can do a simple profile of the app using Systrace to time how long it takes for us to startup to the people tab.

cc @keybase/react-hackers

### DIFF
--- a/shared/docs/react-native/profiling.md
+++ b/shared/docs/react-native/profiling.md
@@ -85,6 +85,8 @@ require.Systrace.endEvent = () => {
 }
 ```
 
+This makes use of User Timings API. If you aren't connected to the remote debugger, you'll need to use a polyfill to run this. See https://github.com/keybase/client/commit/f23f82d941 .
+
 The important thing is that beginEvent and endEvent are defined. Otherwise it
 tries to use some nativeTracingMethod that isn't define.
 
@@ -105,7 +107,7 @@ my pixel with release settings. I wasn't able to bring that number down without
 much more work, but I'll outline the steps I took so we can try this again
 while we do the fixes I'll suggest at the end.
 
-### The starting case
+### The control case
 
 First I used Parashuram setup to mark the tti_complete as the latest
 message in the chat view. Then I built the app with `gradle
@@ -118,8 +120,13 @@ trace-event payload.
 Run this a couple of times to get the starting number that you're going to try
 to improve.
 
+I can use the dev build for debugging/analyzing, but when I've made my fix, I'll switch 
+it back to the release version and time `TTI_COMPLETE` to see how it compares with the control 
+case.
+
 ### Investigating slowdowns
 
+For this, I've switched to a dev build.
 I enabled RN's systrace with the method listed above. as soon as
 I did this I saw the culprits. We load most of the app
 at startup before we can even make a single engine call. Our current routing
@@ -142,7 +149,7 @@ changing all files necessary to start the bootstrapping, we still load most of
 the app. So there isn't a real difference.
 
 
-### Areas to improve (in order or biggest impact and easiest to implement)
+### Areas to improve (in order of biggest impact and easiest to implement)
 
 * Use an alternate routing scheme that will load the screens when required or
   when idle.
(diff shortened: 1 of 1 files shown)


shared/docs/react-native/profiling.md
```diff
@@ -0,0 +1,280 @@
...
+
+## ReactNative.Systrace
+
+Api [here](https://facebook.github.io/react-native/docs/systrace)
+
+This is super useful in analyzing startup time performance cost. You can use
+this to see how long requires are taking. But there's a trick. You have to
+change the implementation of a couple methods for it to work properly.
+
+Add this to your entry file to enable Systrace (you can also enable it via RN
```
buoyad: I can't get this to work, `Can't find variable: performance` (I put it in `index.android`). More specifics for where this should be? Also, button path for enabling via RN devtools would be nice.

shared/docs/react-native/profiling.md
```diff
@@ -0,0 +1,280 @@
...
+with the engine. Part of this is that we need to require so many files before
+we can start the work with the engine.
+
+I tried moving the setup that happens in `app/index.native` to happen before.
+But I didn't see a signifcant startup time difference. My guess is that without
+changing all files necessary to start the bootstrapping, we still load most of
+the app. So there isn't a real difference.
+
+
+### Areas to improve (in order or biggest impact and easiest to implement)
```
buoyad: in order of*

shared/docs/react-native/profiling.md
```diff
@@ -0,0 +1,280 @@
...
+trace-event payload.
+
+[The simple node server](https://gist.github.com/MarcoPolo/0d3186becbce74bc8f21c9deab33dcc0)
+
+Run this a couple of times to get the starting number that you're going to try
+to improve.
+
+### Investigating slowdowns
+
+I enabled RN's systrace with the method listed above. as soon as
```
buoyad: How did you do this if (from above):
> This doesn't work outside of dev mode.

and
> gradle installReleaseUnsigned

shared/docs/react-native/profiling.md
```diff
@@ -0,0 +1,280 @@
...
+* Investigate if we can get by with loading a minimal version of emoij data
+
+
+## Case study: Testing immediate engine calls during bootstrap.
+
+### Abstract
+
+Normally the engine call is setup and called with a `setImmediate`. This is
+usually exactly what we want. Trading a bit of response time from the engine
+for a more response ui. During bootstrapping, our priorities our a little
```
buoyad: our priorities are*

shared/docs/react-native/profiling.md
```diff
@@ -0,0 +1,280 @@
...
+makes sense to avoid that tradeoff.
+
+We'll test if avoiding `setImmediate` during bootstrapping will help reduce
+startup time.
+
+### Test setup
+
+I cherry picked these commits to start the testing process:
+
+* `bd32cdee4c` - Android perf helper files
```
buoyad: Did you push these commits?
```
$ git cat-file -t bd32cdee4c
fatal: Not a valid object name bd32cdee4c
```

buoyad: (same for `507f` and `770c`)

shared/docs/react-native/profiling.md
MarcoPolo: What do you mean button path?

MarcoPolo: I forgot to mention how to add the performance polyfill! which is needed if you aren't connected to a remote debugger.

shared/docs/react-native/profiling.md
MarcoPolo: whoops! You're right, I just pushed it and swapped these with links

shared/docs/react-native/profiling.md
buoyad: Ah, that'd explain it. By button path I mean I don't know how to enable it via RN devtools.

MarcoPolo: Ah yeah, I never used it, but it's there: 



I'll remove that sentence

buoyad: Ah that option isn't on android

MarcoPolo:
Thanks @buoyad!!!

PR #15194, opened 2018-12-20
update dokan to 1.2.1.2000

zanderz:
Verified with https://prerelease.keybase.io/windows/Keybase_2.13.0-20181221100730%2B9022774fa7.amd64.msi
which contains matching KBFS changes, already merged to master.


zanderz:
one issue: `keybase ctl restart` kills the GUI, though it didn't used to. A restart is issued after installing Dokan from the GUI.

go/install/winversion.go
```diff
@@ -0,0 +1,122 @@
...
+		uintptr(unsafe.Pointer(&data[0])),
+	)
+	return ret != 0
+}
+
+// VerQueryValueRoot calls VerQueryValue
+// (https://msdn.microsoft.com/en-us/library/windows/desktop/ms647464(v=vs.85).aspx)
+// with `\` (root) to retieve the VS_FIXEDFILEINFO.
+func VerQueryValueRoot(block []byte) (VS_FIXEDFILEINFO, error) {
+	var offset uintptr
```
taruti: Perhaps a more descriptive name, since this is a pointer and not an offset?

go/install/winversion.go
```diff
@@ -0,0 +1,122 @@
...
+	if ret == 0 {
+		return VS_FIXEDFILEINFO{}, errors.New("VerQueryValueRoot: verQueryValue failed")
+	}
+	start := int(offset) - int(uintptr(blockStart))
+	end := start + int(length)
+	if start < 0 || start >= len(block) || end < start || end > len(block) {
+		return VS_FIXEDFILEINFO{}, errors.New("VerQueryValueRoot: find failed")
+	}
+	data := block[start:end]
+	info := *((*VS_FIXEDFILEINFO)(unsafe.Pointer(&data[0])))
```
taruti: Why not:
```info := *((*VS_FIXEDFILEINFO)(offset)```

File: go/install/winversion.go [hunk 3]
```diff
@@ -0,0 +1,122 @@
...
+	var offset uintptr
+	var length uint
+	blockStart := unsafe.Pointer(&block[0])
+	ret, _, _ := verQueryValue.Call(
+		uintptr(blockStart),
+		uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr(`\`))),
+		uintptr(unsafe.Pointer(&offset)),
+		uintptr(unsafe.Pointer(&length)),
+	)
+	if ret == 0 {
```
taruti: if ret == 0 || length == 0 {

zanderz:
That stuff was all copied from gonutz/w32 so the DLL loading could be safe, I am certain it works as is and would rather not re-build/re-test to change it

taruti:
Ok, then keep it as it is.

PR #15195, opened 2018-12-20
Incoming rpcs as actions (applied to chat2 as an example)

chrisnojima:
The old code was pretty convoluted.
Given an rpc that has sessionid (aka a pure 'incoming' message) was handled like this
1. App starts up
1. Config dispatches ConfiGen.setupEngineListeners telling sagas its time to register themselves
1. Sagas then call engine().setIncomingCallMap with an object of method: callbacks. You can't ever have 2 handlers
1. When engine gets an incoming call w/o a session it checks to see if a callback exists. if so it'll call it, get an array of actions back and then reach into the saga middleware and yield the actions. If those throw it'll send a global error. 

This has the downside of being a mini version of redux saga inside itself. It causes import loops that are hard to break (sagas pull in engine, engine pulls in the saga middleware which pulls in the sagas). It also causes a lot of these handlers to have to return Saga.callUntyped() just so they can get the state.

Instead if we have the engine dispatch real actions when these rpcs come
1. Multiple sagas can register for them
1. You handle it like a regular action (aka you can use actionToAction
1. You have access to state so you can usually just return a put
1. No special handling of things, its all the same type of code
1. You can have multiple rpcs that as a side effect have one action, aka you can do `yield Saga.actionToAction([rpc1, rpc2, rpc3], reloadTheList)`
1. Its now possible for a reducer to directly handle an incoming call vs having a saga handle it

One caveat of how this works is this generates a json file which `yarn build-actions` runs on. This means running make in protocols isn't enough if the UI is doing things w/ an incoming action. Instead we should run `yarn update-protocol`

Next steps:
1. Once approved, apply to all other setupEngineListeners() calls
1. Handle custom reply cases
1. Make engine.actionOn(dis)Connect() just dispatch an action


protocol/Makefile
```diff
@@ -44,6 +44,7 @@ js/rpc.js: build-stamp | config
 	@mkdir -p js/
 	node --harmony ./bin/flow.js
 	cp ./js/rpc* ../shared/constants/types
+	cp ./js/engine-gen.json ../shared/actions/json
```
chrisnojima: we genereate a json file now

protocol/bin/flow.js
```diff
@@ -55,23 +55,28 @@ var projects = {
...
+        map.types = {...map.types, ...next.types}
+        map.messages = {...map.messages, ...next.messages}
+        return map
+      }, {})
+      .then(typeDefs => {
+        writeFlow(typeDefs, project)
+        write(typeDefs, project)
+      })
+  })
+).then(() => {
```
chrisnojima: lets write stuff after everything is done

protocol/bin/flow.js
```diff
@@ -582,7 +646,7 @@ function lintJSON(json) {
 
 function lintError(s, lint) {
   if (lint === 'ignore') {
-    console.log('Ignoring lint error:', colors.yellow(s))
+    // console.log('Ignoring lint error:', colors.yellow(s))
```
chrisnojima: this is noisy and we don't even do anything in reaction to this so lets just silence

shared/actions/chat2/index.js
```diff
@@ -39,13 +39,45 @@ import flags from '../../util/feature-flags'
 // Ask the service to refresh the inbox
 const inboxRefresh = (
   state: TypedState,
-  action: Chat2Gen.InboxRefreshPayload | Chat2Gen.LeaveConversationPayload
+  action:
```
chrisnojima: following the better pattern where n-actions being dispatched cause a refresh as a side effect

shared/actions/chat2/index.js
```diff
@@ -55,11 +87,6 @@ const inboxRefresh = (
       .map(item => Constants.unverifiedInboxUIItemToConversationMeta(item, username))
       .filter(Boolean)
     // Check if some of our existing stored metas might no longer be valid
-    const clearExistingMetas =
```
chrisnojima: made more explicit above

shared/actions/chat2/index.js
```diff
@@ -514,342 +760,114 @@ const onChatThreadStale = updates => {
...
-  logger.info(`Got ${updates.length} reaction updates for convID=${conversationIDKey}`)
-  return [Chat2Gen.createUpdateReactions({conversationIDKey, updates})]
-}
-
-// This is to simplify the changes that setIncomingCallMap created. Could clean this up and remove this
-const arrayOfActionsToSequentially = actions =>
-  Saga.callUntyped(Saga.sequentially, (actions || []).map(a => Saga.put(a)))
-
-// Handle calls that come from the service
-const setupEngineListeners = () => {
```
chrisnojima: 🔥

shared/actions/json/chat2.json
```diff
@@ -357,10 +357,6 @@
       "placeholderID": "RPCChatTypes.MessageID",
       "message": "Types.MessageAttachment"
     },
-    // Update who's typing in a conversation
-    "updateTypers": {
```
chrisnojima: handled directly be the reducer now, see below

shared/actions/json/config.json
```diff
@@ -3,11 +3,11 @@
     "import * as Types from '../constants/types/config'",
     "import * as Tabs from '../constants/tabs'",
     "import * as ChatTypes from '../constants/types/chat2'",
-    "import {RPCError} from '../util/errors'"
+    "import {RPCError} from '../util/errors'",
   ],
   "actions": {
     "setupEngineListeners": {
-      "_description": "when sagas should start creating their incoming handlers / onConnect handlers"
```
chrisnojima: once this is approved i'll apply to other parts of the app

shared/engine/index.js.flow
```diff
@@ -3,36 +3,7 @@ import Session from './session'
...
-  ...SellarIncomingCallMap,
-|}
-
-export type CustomResponseIncomingCallMapType = {|
-  ...KBCustomResponseIncomingCallMap,
-  ...ChatCustomResponseIncomingCallMap,
-  ...GregorCustomResponseIncomingCallMap,
-  ...SellarCustomResponseIncomingCallMap,
-|}
+import type {IncomingCallMapType, CustomResponseIncomingCallMapType} from '../constants/types/rpc-all-gen'
```
chrisnojima: generated now 👊

shared/package.json
```diff
@@ -17,7 +17,7 @@
     "test": "cross-env-shell BABEL_ENV=test BABEL_PLATFORM=Electron jest",
     "update-icon-constants": "yarn run _helper update-icon-constants",
     "update-icon-font": "yarn run _helper update-icon-font",
-    "update-protocol": "cd ../protocol && make clean && make",
+    "update-protocol": "cd ../protocol && make clean && make && cd ../shared && yarn build-actions",
```
chrisnojima: we can make a new engine json, so build actions

shared/reducers/chat2.js
```diff
@@ -722,8 +725,15 @@ const rootReducer = (
         })
       )
     }
-    case Chat2Gen.updateTypers: {
-      return state.set('typingMap', action.payload.conversationToTypers)
+    case EngineGen.chat1NotifyChatChatTypingUpdate: {
```
chrisnojima: woot, this is possible now. before the saga would have to register an incoming handler, then we'd dispatch another action

shared/util/saga.js
```diff
@@ -44,9 +44,13 @@ function* sequentially(effects: Array<any>): Generator<any, Array<any>, any> {
...
 
 // Helper that expects a function which returns a promise that resolves to a put
+// (also can return a raw action. this is going to be renamed soon)
 function actionToPromise<A>(
   pattern: RS.Pattern,
-  f: (state: TypedState, action: A) => null | false | void | Promise<TypedActions | null | false | void>
+  f: (
+    state: TypedState,
+    action: A
+  ) => null | false | void | Promise<TypedActions | null | false | void> | TypedActions
```
chrisnojima: this will change soon anyways, but we often return a Promise.resolve(Action) or worse use a actionToAction when we just want an action to lead to a put

chrisnojima:
@keybase/react-hackers this is ready again

shared/actions/engine-gen-gen.js
```diff
@@ -0,0 +1,790 @@
+// @flow
```
MarcoPolo: lol this filename. We heard you like generated code..

PR #15200, opened 2018-12-21
Score @-mention results

buoyad:
cc @cecileboucheron this adds scoring to suggestions
scoring hierarchy:

Username and fullname starting with filter
 --> Username starting with filter
 ----> Fullname starting with filter
 ------> Username or fullname including filter

r? @keybase/react-hackers


cecileboucheron:
🎉

PR #15201, opened 2018-12-21
identify3

maxtaco:
- identify3 is mainly a shim wrapper around identify3; it runs the existing identify2 engine with an identify3.UIAdpater
  - this adapter speaks the old IdentifyUI to the identify2 engine, and the new keybase1.Identify3UiInterface to electron
- most of the code is in a new directory call identify3/
- but we plumb the underlying state through to all G, and that code is in libkb/identify3
  - this state needs to persist as long as the window is open, so maybe for a while. though we do send a timeout message up to electron after 24 hours
- the code tested so far is only for the flow in which electron starts with identify3. Still to come is showing an ID3 popup card as a result of a CLI activity


maxtaco:
@patrickxb this is a pretty low-risk change, since it's not being called from anywhere yet.

go/identify3/adapter.go
```diff
@@ -0,0 +1,340 @@
...
+
+// return true if we need an upgrade
+func (i *UIAdapter) setRowStatus(arg *keybase1.Identify3UpdateRowArg, lcr keybase1.LinkCheckResult) bool {
+
+	needUpgrade := false
+	i.M().CDebugf("setRowStatus(%+v)", lcr)
+
+	switch {
+
+	// The proof worked, and either we tracked it as working, or we tracked it as working
```
patrickxb: `either` and `or` are the same here.

go/identify3/adapter.go
```diff
@@ -0,0 +1,340 @@
...
+		needUpgrade = true
+
+	// The proof failed, but we didn't track it
+	case lcr.ProofResult.State != keybase1.ProofState_OK && lcr.RemoteDiff == nil:
+		arg.Color = keybase1.Identify3RowColor_ORANGE
+		arg.State = keybase1.Identify3RowState_WARNING
+		arg.Metas = append(arg.Metas, keybase1.Identify3RowMeta("unreachable"))
+
+	// The proof failed, but we did "ignore" it, so it's OK
+	case lcr.ProofResult.State != keybase1.ProofState_OK && (lcr.RemoteDiff == nil || lcr.RemoteDiff.Type == keybase1.TrackDiffType_NONE):
```
patrickxb: do you need `lcr.RemoteDiff == nil`?  won't that get caught by `case` on line 80?

go/identify3/adapter.go
```diff
@@ -0,0 +1,340 @@
...
+		arg.Color = keybase1.Identify3RowColor_ORANGE
+		arg.State = keybase1.Identify3RowState_WARNING
+		arg.Metas = append(arg.Metas, keybase1.Identify3RowMeta("ignored"))
+
+	// The proof failed, but we did "ignore" it via "SNOONZE", so it's OK, for now.
+	case lcr.ProofResult.State != keybase1.ProofState_OK && lcr.RemoteDiff != nil && lcr.RemoteDiff.Type == keybase1.TrackDiffType_NONE_VIA_TEMPORARY:
```
patrickxb: again, i think you know that lcr.RemoteDiff isn't nil already.

go/identify3/identify3_test.go
```diff
@@ -0,0 +1,251 @@
...
+	// use an insecure triplesec in tests
+	tc.G.NewTriplesec = func(passphrase []byte, salt []byte) (libkb.Triplesec, error) {
+		warner := func() { tc.G.Log.Warning("Installing insecure Triplesec with weak stretch parameters") }
+		isProduction := func() bool {
+			return tc.G.Env.GetRunMode() == libkb.ProductionRunMode
+		}
+		return insecureTriplesec.NewCipher(passphrase, salt, warner, isProduction)
+	}
+	return tc
+}
```
patrickxb: I tried this locally and it failed to compile:

    > go test
    # github.com/keybase/client/go/identify3
./identify3_test.go:23:37: not enough arguments in call to triplesec.NewCipher
	have ([]byte, []byte, func(), func() bool)
	want ([]byte, []byte, triplesec.Version, func(), func() bool)
    FAIL	github.com/keybase/client/go/identify3 [build failed]

maxtaco: thx, rebase bug, working on it now!

go/identify3/adapter.go
```diff
@@ -0,0 +1,340 @@
...
+
+func (i *UIAdapter) checkEldest(key keybase1.IdentifyKey) {
+	if key.TrackDiff == nil || key.TrackDiff.Type != keybase1.TrackDiffType_NEW_ELDEST {
+		return
+	}
+	err := i.ui.Identify3UserReset(i.M().Ctx(), i.session.ID())
+	if err != nil {
+		i.M().CDebugf("Error sending user reset message: %s", err)
+	}
+	return
```
patrickxb: you can delete this `return`

go/identify3/adapter.go
patrickxb: sure, i was just trying to read through all the cases and see what was different.

maxtaco: thanks for the careful review!

maxtaco:
cc @chrisnojima  this'll go in once green

PR #15202, opened 2018-12-21
use getsubteams

chrisnojima:
@keybase/react-hackers this uses a new rpc that should be faster (but isn't?)
cc: @xgess


xgess:
should be faster. i suspect fast enough. if it's still slow, there's another chunk of work `core` can do, but i'd rather not do it unless it's necessary.

chrisnojima:
i think we do need to fix this. it takes 10 seconds for me to load that same team. so its not some random server load issue:

chrisnojima:
screen shot 2018-12-28 at 4 01 34 pm

keybase-travis:
Why is this slow?

PR #15203, opened 2018-12-21
add banner for when address not in sigchain

xgess:
add a new banner that says "Your Keybase identity will not be linked to the transaction." inside the payment window when: 
* the sending address is not the primary address
* you aren't just moving money between accounts

### DIFF
--- a/go/stellar/build.go
+++ b/go/stellar/build.go
@@ -148,20 +148,22 @@ func BuildPaymentLocal(mctx libkb.MetaContext, arg stellar1.BuildPaymentLocalArg
 						Message: fmt.Sprintf("Because it's %s first transaction, you must send at least %s XLM.", them, amount),
 					})
 				}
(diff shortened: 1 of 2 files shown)


go/stellar/build.go
```diff
@@ -173,6 +175,12 @@ func BuildPaymentLocal(mctx libkb.MetaContext, arg stellar1.BuildPaymentLocalArg
 						}
 					}
 				}
+				if !sendingToSelf && !fromPrimaryAccount {
+					res.Banners = append(res.Banners, stellar1.SendBannerLocal{
+						Level:   "info",
+						Message: "Your Keybase identity will not be linked to the transaction.",
```
patrickxb: This is the text design wants?

xgess:
```suggestion
						Message: "Your Keybase identity will not be linked to this transaction.",
```

PR #15205, opened 2018-12-21
Humanize err

joshblum:
let me know what you think of the wording! this error should never really appear unless there is a bug but i thought this would make it more clear


cecileboucheron:
The thing is that on desktop (vs. mobile) there is no feedback page - so let's just remove the second part.

joshblum:
fixed!

mmaxim:
I wish this could be changed higher up in the stack.

joshblum:
error directly bubbles up to the UI so i feel this is the right place to change so we don't accidentally return it again in the future

joshblum:
we also have a bunch of logging around these errors in the existing methods so it is still easy to trace in the logs

PR #15208, opened 2018-12-21
update dokan to 1.2.1.2000 (#15194)

zanderz:
* dokan 1.2.1.1000

* put package on s3 after all

* update Dokan again, safer dll loading

* forgot .wxs


taruti:
This should be closed probably?

PR #15209, opened 2018-12-21
New saga utils: Part 1: applied to devices

chrisnojima:
This introduces the 2 new saga utils which will replace safeTake* and actionTo*
as mentioned in the confluence doc we now have 
```
  yield* Saga.chainAction<DevicesGen.RevokePayload>(DevicesGen.revoke, revoke)
```

and 
```
  yield* Saga.chainGenerator<DevicesGen.ShowPaperKeyPagePayload>(DevicesGen.showPaperKeyPage, requestPaperKey)
```

originally I was trying to coerce flow to derive the action types but this proved very hard. Instead you have to explicitly pass the generic. if you don't pass the type then your action will be untyped totally. as we get more flow-strict that should be caught

good news is coverage goes up due to this pr. I think 90% of the time we can just use chainAction and be done

actions/devices.js
master:
`Covered: 96.12% (322 of 335 expressions)`
pr:
`100.00% (319 of 319 expressions)`


chrisnojima:

PR #15215, opened 2018-12-22
WIP: deprecate hand made route-tree actions

chrisnojima:
Just directly use the generated actions


chrisnojima:
@keybase/react-hackers this is ready. this does a big cleanup to deprecate the manual actions in actions/route-tree. we should be using route-tree-gen instead

PR #15224, opened 2018-12-23
New Saga Utils: Part 10 (teams)

chrisnojima:


shared/actions/teams.js
```diff
@@ -133,86 +132,79 @@ const _leaveTeam = function(state: TypedState, action: TeamsGen.LeaveTeamPayload
...
+        retentionPolicy = Constants.serviceRetentionPolicyToRetentionPolicy(policy)
+        if (retentionPolicy.type === 'inherit') {
+          throw new Error(`RPC returned retention policy of type 'inherit' for team policy`)
+        }
+      } catch (err) {
+        logger.error(err.message)
+        throw err
+      }
+    })
+    .finally(() => TeamsGen.createSetTeamRetentionPolicy({retentionPolicy, teamname}))
```
buoyad: I think something is up with this, retention pickers for teams are perma-spinning in team settings + info panel

shared/actions/teams.js
```diff
@@ -307,65 +295,57 @@ const _inviteByEmail = function*(action: TeamsGen.InviteToTeamByEmailPayload) {
...
+      sendChatNotification,
+      username,
+    },
+    [Constants.teamWaitingKey(teamname), Constants.addMemberWaitingKey(teamname, username)]
+  )
+    .then(() => {})
+    .catch(e => {
+      // identify error
+      if (e.code === RPCTypes.constantsStatusCode.scidentifysummaryerror) {
+        if (isMobile) {
```
buoyad: Should I be seeing this?

shared/actions/teams.js
```diff
@@ -624,14 +599,10 @@ const _getDetails = function*(action: TeamsGen.GetDetailsPayload): Saga.SagaGene
   )
 }
 
-function _getDetailsForAllTeams(action: TeamsGen.GetDetailsForAllTeamsPayload, state: TypedState) {
-  const actions = state.teams.teamnames
-    .toArray()
-    .map(teamname => Saga.put(TeamsGen.createGetDetails({teamname})))
-  return Saga.sequentially(actions)
-}
+const getDetailsForAllTeams = (state, action) =>
```
buoyad: 🐑 this action is unused

shared/actions/teams.js
chrisnojima: yeah this is incorrect, you can't return from finally

shared/actions/teams.js
chrisnojima: unclear , nothing logically different should be happening vs master

shared/actions/teams.js
chrisnojima: lets leave this alone for now. how these actions all work is pretty messy and i think a ton of these actions shouldn't exist

shared/actions/teams.js
buoyad: Hmm ok, works the same in master. I thought this was the flow where we pushed the profile

PR #15225, opened 2018-12-23
New Saga Utils: Part 11 (tracker, teambuilding, fs, unlock, users)

chrisnojima:


chrisnojima:
cc: @jzila @songgao for the new saga api (see confluence)

zanderz:
Clicking on "private" in the files tab doesn't do anything in this branch.
Also, made a team from the chat tab and got a red bar on that dialog saying something about signature at position 1 (LMK if u want a log)

chrisnojima:
@zanderz i fixed the routing things. the old code was like
```
Saga.all([
thing that returns], routeactions])
```

these can happen in parallel so i just do routeactions first, then the other stuff. actually makes fs after.

the red dialog is actually happening on master. 'sigchain link cannot be added at seqno: 1' right?

chrisnojima:
This is ready for re review

songgao:
Thanks for the heads up!

PR #15235, opened 2018-12-26
upcall-based ID3 flow

maxtaco:
- the service starts the flow and comes up with the GUI ID
- the UI must register that it can handle ID3, which is then favored over ID.
- add a systest to make sure the flow seems reasonable, though test coverage isn't great
- Very low-risk PR, since it won't do anything until the UI calls into it.


maxtaco:
@chrisnojima this should now work

maxtaco:
10th time was a charm. Merging.

PR #15236, opened 2018-12-26
Support regexes as markers to trigger mentions

buoyad:
Adds user autocomplete to text-based stellar sending. r? @keybase/react-hackers


mmaxim:
Sick

PR #15238, opened 2018-12-26
bad object types in comparison

zanderz:
Turns out the argument was a object with a property named "code". This definitely used to work - is this an Electron change? Also, how is this not caught with flow or anything?


zanderz:
any theories why this would work before then stop working?

Also, the comparison was always false, even though the actual numeric values were the same - that's expected too, because the underlying types are different?

chrisnojima:
I don't know what you mean about the numerical values. Flow doesn't affect anything runtime. A value in an object and the object aren't the same thing.

PR #15239, opened 2018-12-27
New Team Building Improvements

MarcoPolo:
Commits describe the work, but basically:
- [x] First row shouldn't be highlighted until they press the down arrow (from the input)
- [x] Actions (tapping the row, tapping the add button, pressing return) don't work on recommendations
- [x] Tapping on the row for someone already in the conversation should remove them.
- [x] Hovering the row for someone already in the conversation should also change the button
- [x] `username on service` should either be their full name if we know it or nothing
- [x] Switching tabs before searching should either not work or show recommendations from those services
- [x] Result rows for non-Keybase services should have their username on that service above their full name and only the Keybase logo with their Keybase username next to it: https://zpl.io/VYRkDDn
- [x] Result rows for non-Keybase services should show the service's icon instead of the default user image if they don't have an avatar.
- [x] Team box should show service icon of non keybase users
- [x] Scrolling to the bottom of the list should load more results (and show a spinner when loading).

 @keybase/react-hackers @adamjspooner


shared/common-adapters/section-list.desktop.js
```diff
@@ -80,10 +85,13 @@ class SectionList extends React.Component<Props, State> {
   _checkOnEndReached = throttle(target => {
     const diff = target.scrollHeight - (target.scrollTop + target.clientHeight)
     if (diff < 5) {
-      this.props.onEndReached()
+      this._onEndReached()
     }
   }, 100)
 
+  // This matches the way onEndReached works for sectionlist on RN
+  _onEndReached = once(() => this.props.onEndReached && this.props.onEndReached())
```
MarcoPolo: This changes how master works. Not related to this PR.

shared/common-adapters/list.js.flow
```diff
@@ -14,6 +14,7 @@ export type Props<Item> = {
   itemSizeEstimator?: (index: number, cache: {[index: number]: number}) => number, // Desktop only
   keyboardShouldPersistTaps?: 'never' | 'always' | 'handled', // mobile only
   windowSize?: number, // Mobile only, has a non-RN default
+  onEndReached?: () => void,
```
MarcoPolo: This exists now. Previously only worked with section list

shared/stories/index.desktop.js
```diff
@@ -12,6 +12,17 @@ import {initDesktopStyles} from '../styles/index.desktop'
 
 const stories = {...sharedStories, ...desktopStories}
 
+const filter = null
```
MarcoPolo: Useful for when developing to only show the story you're working on. Also makes reloads faster.

shared/common-adapters/icon.js.flow
```diff
@@ -22,6 +22,11 @@ export type Props = {
   inheritColor?: boolean,
   underlayColor?: string,
   className?: string,
+  // External services use custom colors which aren't defined in the stylesheet
+  // So they break the css optimizations. We have this extra prop so we can set
+  // the color without messing with the existing color code.
+  // TODO would be nice to not have to do this.
+  externalServiceColor?: Color,
```
chrisnojima: confused why we need this? why aren't we just setting color?

shared/stories/index.desktop.js
```diff
@@ -12,6 +12,17 @@ import {initDesktopStyles} from '../styles/index.desktop'
...
+const filteredStories = Object.keys(stories).reduce(
```
chrisnojima: can we just load this off the query param in window, then we don't need to do this in code?

chrisnojima: (or env var or something)

shared/common-adapters/icon.js.flow
chrisnojima: likely easier to chat about this but. 
I think this is all temporary. we aren't going to supply these as colors to iconfonts soon. soon it'll all be images. therefore this is all going away. 
short term i'm fine with this approach but. 1 we should comment that its temporary and 2. i'd prefer to make this more generic like `this.props.inlineColorOverride` or something. not something search specific language inside icon

chrisnojima:
when this is in lets ping design so they can test the updates

PR #15247, opened 2018-12-28
generate a notification for outbox messages, and refactor attachment send CORE-9711

mmaxim:
Patch does the following:

**Go**

1. Generate an `IncomingMessage` notification for messages we insert into the outbox. This replaces the flow where the JS would generate the outbox ID and insert into the store before invoking `PostTextLocalNonblock`.
2. Simplify attachment sending code to also take advantage of the new pending message notification. We don't need to generate previews in JS and send to Go anymore.
3. Upgrade `Outbox` to add pending previews to messages pushed and to be the module responsible for generating notifications. We need to do it there to avoid any races where the pending message can lose to the sent message notification.

**JS**
@keybase/react-hackers 

1. Simplify all the message sending flows to no longer insert messages into the store that haven't been received from the service.


chrisnojima:
Please wait for the saga changes to go in before merging

go/chat/server_test.go
```diff
@@ -5513,7 +5531,10 @@ func TestChatSrvStellarUI(t *testing.T) {
 		}
 		if !expectError {
 			select {
-			case <-ui.PostReadyToSend:
+			case msg := <-listener.newMessageLocal:
```
joshblum: use `consumeNewPendingMsg` or `require.True(t, msg.Message.IsOutbox())`

go/chat/server_test.go
```diff
@@ -5582,7 +5603,10 @@ func TestChatSrvStellarUI(t *testing.T) {
 	default:
 	}
 	select {
-	case <-ui.PostReadyToSend:
+	case msg := <-listener.newMessageLocal:
```
joshblum: use `consumeNewPendingMsg` or `require.True(t, msg.Message.IsOutbox())`

PR #15259, opened 2018-12-30
Idea for Helper Process Simplification

maxtaco:


maxtaco:
cc @strib  on 683fa3c, memleak fix

PR #15272, opened 2019-01-02
stop ios from scrolling down as new content comes in if you're scrolled up

chrisnojima:
- [x] use maintainVisibleContentPosition prop to keep scrollview in place as new content comes in

to repro, on master load up a thread and scroll up a little bit. on another client message that thread. each message will push your mobile thread around, making reading back super annoying
this pr stops that from happening

@keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/list-area/normal/index.native.js
+++ b/shared/chat/conversation/list-area/normal/index.native.js
@@ -67,6 +67,11 @@ class ConversationList extends React.PureComponent<Props> {
     }
   }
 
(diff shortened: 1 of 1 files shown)


mmaxim:
Haven't tried it, but seems simple enough to try.

chrisnojima:
if you're curious: https://github.com/facebook/react-native/blob/master/Libraries/Components/ScrollView/ScrollView.js#L150

PR #15282, opened 2019-01-02
use double quote around ${url}

songgao:


shared/fs/filepreview/av-view.desktop.js
```diff
@@ -43,12 +43,14 @@ video {
 }
 `
 
+// Double quote around ${url} is necessary as encodeURIComponent encodes double
+// quote but not single quote.
```
jzila: That's not why though, right? `"${url}"` is equivalent to `url`, whereas `'${url}'` just sets that string.

songgao: Sorry I don’t understand the question. These are inside the injected JavaScript, so they are both strings. `${url}` is populated through the back-tick quote around the injected JavaScript.

jzila: sorry, to clarify:
```javascript
v.setAttribute('src', "${url}")
```
is equivalent to:
```javascript
v.setAttribute('src', url)
```

PR #15286, opened 2019-01-02
Surya/core 9529/systemd replace run keybase

heronhaye:
wip.
Should kill_all in run_keybase kill redirector too? also nohup?


packaging/linux/post_install.sh
```diff
@@ -143,3 +143,8 @@ fi
 if which gtk-update-icon-cache &> /dev/null ; then
   gtk-update-icon-cache -q -t -f /usr/share/icons/hicolor
 fi
+
+# Reload systemd unit files in case they have changed
+if command -v systemctl &> /dev/null; then
+    systemctl --user daemon-reload
```
heronhaye: todo: do it in restart keybase so we can access user manager

PR #15290, opened 2019-01-03
drop LookupIDUntrusted since it messes up SBS sometimes CORE-9786

mmaxim:
We need to get the real team name in `GetInboxQueryLocalToRemote`, otherwise SBS conversations can fail in strange ways. `LookupIDUntrusted` was getting a name which didn't match what we would get back from doing the team load, so let's just use `LookupID` (since it is plenty fast now) and delete `LookupIDUntrusted`.


mmaxim:
No, we can probably move that to using a system like the rate limits, where we collect in the context at some point.

PR #15291, opened 2019-01-03
flip find convos check

mmaxim:


mmaxim:
This guy changed a lot though, might want to double check this is the only thing that needs to get changed.

mmaxim:
I'm just going to merge

PR #15296, opened 2019-01-03
better debugging

maxtaco:


maxtaco:
doesn't really need a thumb, or CI, just I need this to figure out why my FB proof isn't showing up

PR #15298, opened 2019-01-03
allow for unsent text to be set in the store and fed into the input box CORE-9795

mmaxim:
@keybase/react-hackers 

Patch does the following:
1. Add a new map to the store, which maps `convID -> unsentText`. The purpose of this is to be able to override the module level unsent text with stuff from redux.
2. In the input box component, if the unsent text is set in the store, then use it in `getUnsentText()` instead of whatever we have in the module level storage. We also set a flag to let the component know it should re-run `getUnsentText()` to get the new data in there.
3. If the user types anything after that, we blow away the unsent text in the store.
4. If a user cancels a Stellar in-chat send, then we set the unsent text in the store of the message they are composing.

### DIFF
--- a/go/chat/server.go
+++ b/go/chat/server.go
@@ -1216,7 +1216,9 @@ func (h *Server) runStellarSendUI(ctx context.Context, sessionID int, uid gregor
 	if err := ui.ChatStellarShowConfirm(ctx); err != nil {
 		return res, err
 	}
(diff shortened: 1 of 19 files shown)


mmaxim:
Note: I also think this is a nice path forward for when the service is going to run this with drafts.

PR #13486, opened 2018-08-29
New New Conversation in Chat

MarcoPolo:
dLots of things are not implemented, for starters:
* Search counts on other services.
* Optimizations of any kind
  * (Serious render thrashing when using the arrow keys)
* Pending states
* No search results state
* A mobile version
* Tooltips (they are there, but have z-index issues)
* ... A bunch of other little things

What is Implemented
* Ground work to iterate on
* Store, actions, container
* Search in other services (besides keybase)
* Search on keybase
* Building up a team
* Special keys: up,down, enter, backspace
* Creating the conversation.

Caveats aside, if you want to use it, add this feature flag:
`newTeamBuildingForChat=1`

PR #15302, opened 2019-01-04
Reorder Services

MarcoPolo:
Ordering is important here.
@keybase/react-hackers

### DIFF
--- a/shared/constants/team-building.js
+++ b/shared/constants/team-building.js
@@ -3,7 +3,17 @@ import logger from '../logger'
 import * as I from 'immutable'
 import * as Types from './types/team-building'
 
-const allServices: Array<Types.ServiceIdWithContact> = Object.keys(Types._services)
+const allServices: Array<Types.ServiceIdWithContact> = [
+  'keybase',
+  'contact',
+  'twitter',
+  'facebook',
+  'github',
+  'reddit',
+  'hackernews',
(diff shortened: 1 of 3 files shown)


chrisnojima:
Maybe better to just have an array you reduce to make this

MarcoPolo:
I thought it had to be an object so that we could have a type on the keys? `export type ServiceIdWithContact = $Keys<typeof _services>` Let me check if we can do something similar with an array

chrisnojima:
Oh yeah. But maybe it's not derived. Just a regular type. Then the const can be typed as array of that type

MarcoPolo:
Lol, I'm pretty sure that's what it was at some point...
ah here it is: https://github.com/keybase/client/pull/13486/files/58c7b62932af774b69d115579bd4e1e2cf7908fc#r221350024

chrisnojima:
I guess I'm a little worried about it getting broken again

MarcoPolo:
yep, that's reasonable. I'll switch it to an ordered array

chrisnojima:
https://flow.org/en/docs/types/generics/#toc-adding-types-to-generics

MarcoPolo:
yeah, but nothing effectively changed about the code. This is likely the correct behavior, and there was some bug edge case when using $Keys<...>

PR #15304, opened 2019-01-04
show typing status on small team rows

mmaxim:
@keybase/react-hackers 

I think it is a nice touch to have typing status in the inbox view, and most other messengers have this. Here is what it looks like:

image

cc @cecileboucheron @adamjspooner @malgorithms for any comments/concerns.

### DIFF
--- a/shared/chat/inbox/row/small-team/bottom-line.js
+++ b/shared/chat/inbox/row/small-team/bottom-line.js
@@ -17,6 +17,7 @@ type Props = {
   showBold: boolean,
   snippet: ?string,
   snippetDecoration: ?string,
+  snippetStyle?: Object,
   subColor: string,
   youNeedToRekey: boolean,
   youAreReset: boolean,
@@ -61,6 +62,7 @@ class BottomLine extends PureComponent<Props> {
           color: this.props.subColor,
           ...(this.props.showBold ? globalStyles.fontBold : {}),
         },
+        this.props.snippetStyle,
       ])
 
       let snippetDecoration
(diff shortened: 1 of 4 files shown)


mmaxim:
Updated type, not sure why this would cause thrashing of any kind.

mmaxim:
Ok moved the stuff up one function, let me know if that works.

shared/chat/inbox/row/small-team/bottom-line.js
```diff
@@ -97,9 +99,10 @@ class BottomLine extends PureComponent<Props> {
           )
           break
         default:
-          snippetDecoration = this.props.snippetDecoration ? (
-            <Text type="BodySmall">{this.props.snippetDecoration}</Text>
-          ) : null
+          snippetDecoration =
+            this.props.snippetDecoration && !this.props.isTypingSnippet ? (
```
chrisnojima: Needs !!

mmaxim: Ok, that’s been in there for a long time though. Is there anyway to get Flow to catch these?

chrisnojima: we've spent a little time on the flow thing and we haven't found a solution. we'd basically have to override the built in definition of react to disallow this (also this only really affects android which crashes out, ios/electron is fine with strings).
additionally i think this is slightly different than the old code. I think its a good idea to guard all these expressions anyway but in my thinking
old:
```
expr ? 'trueValue' : 'falseValue'
```
will, no matter what eval to 'trueValue' or 'falseValue' and nothing else
whereas the new expression
```
expr1 && expr2 ? 'trueValue' : 'falseValue'
```
is more ambiguous, depending of your knowledge of precedence rules.
aka on reading you might not know if its
```
expr1 && (expr2 ? 'trueValue' : 'falseValue')
// or
(expr1 && expr2) ? 'trueValue' : 'falseValue'
```
so looking at it, you might not be sure.

eitherway, ideally flow would catch this. 
in the meantime lets just be more explicit so its more clear

PR #15305, opened 2019-01-04
Part two: use new header on People tab (and profiles).

adamjspooner:


adamjspooner:
@keybase/react-hackers, bump.

shared/profile/index.native.js
```diff
@@ -418,31 +418,28 @@ class Profile extends Component<Props, State> {
     if (section.title === 'profile') {
       const trackerStateColors = stateColors(this.props.currentlyFollowing, this.props.trackerState)
       return (
-        <Kb.Box
+        <Kb.HeaderHocHeader
```
buoyad: The blue background isn't staying behind the search bar as you scroll
fullscreen_1_14_19__10_03_am

shared/profile/index.native.js
```diff
@@ -418,31 +418,28 @@ class Profile extends Component<Props, State> {
...
+          borderless={true}
+          onLeftAction={this.props.onBack}
           style={{
```
MarcoPolo: Shouldn't it be:
```suggestion
          headerStyle={{
```

I haven't tested that, but that might fix @buoyad issue

shared/people/index.native.js
```diff
@@ -7,48 +7,40 @@ import {globalStyles, styleSheetCreate} from '../styles'
 import {isIOS} from '../constants/platform'
 
 const People = (props: Props) => (
```
MarcoPolo: Can you check your file formatter tool? When I load this and format it with my prettier/eslint settings I get these diffs:

```
diff --git a/shared/people/index.native.js b/shared/people/index.native.js
index b9dfca5484..657aad6319 100644
--- a/shared/people/index.native.js
+++ b/shared/people/index.native.js
@@ -22,14 +22,17 @@ const People = (props: Props) => (
           label: 'Avatar',
         },
       ]}
-      titleComponent={(<PeoplePageSearchBar {...props} />)}
+      titleComponent={<PeoplePageSearchBar {...props} />}
     />
     <Kb.ScrollView
       style={styles.scrollView}
       refreshControl={
         // TODO set refreshing to the actual prop once the bug in RN gets fixed
         // see https://github.com/facebook/react-native/issues/5839
-        <Kb.NativeRefreshControl refreshing={isIOS ? false : props.waiting} onRefresh={() => props.getData()} />
+        <Kb.NativeRefreshControl
+          refreshing={isIOS ? false : props.waiting}
+          onRefresh={() => props.getData()}
+        />
       }
     >
       <PeoplePageList {...props} />
```

shared/profile/index.native.js
adamjspooner: Yup! Thanks for catching that. 🙇

chrisnojima:
just fyi profile is going to change soon

PR #15312, opened 2019-01-04
more right padding on mobile chat

zanderz:
Before:
simulator screen shot - iphone 6 - 2019-01-04 at 10 49 28

After:
simulator screen shot - iphone 6 - 2019-01-04 at 10 48 40

### DIFF
--- a/shared/chat/conversation/messages/wrapper/index.js
+++ b/shared/chat/conversation/messages/wrapper/index.js
@@ -522,7 +522,7 @@ const styles = Styles.styleSheetCreate({
         Styles.globalMargins.tiny + // right margin
         Styles.globalMargins.tiny + // left margin
         Styles.globalMargins.mediumLarge, // avatar
-      paddingRight: Styles.globalMargins.tiny,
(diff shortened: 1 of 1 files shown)


mmaxim:
What does this do to attachments?

mmaxim:
Also why do this? It looks better Before to me.

cecileboucheron:
@mmaxim The text is too close to the right edge now.. This aligns it with the spacing with have in the header and in the nav.

mmaxim:
What does this do to exploding messages, those already get pushed way over due to the clock and timer? The actual text is going to be like 60% across the screen.

zanderz:
It seems that attachments and bomb icons aren't squished over as far as the text. IDK, it is @cecileboucheron 's ticket.
simulator screen shot - iphone 6 - 2019-01-04 at 11 53 47

mmaxim:
Ok if that is the plan, sounds ok to me. Look like a separate bug though with exploding stuff and attachments.

PR #15314, opened 2019-01-04
Fix asset input placeholder color

buoyad:
The style was being overridden by later `style` tags on the screen (encrypted note, asset input). I added the `className` to the `::-webkit-input-placeholder` selector so our selector is the most specific. I also added a warning message when `placeholderColor` is used without `className`. 

r? @keybase/react-hackers

### DIFF
--- a/shared/common-adapters/plain-input.desktop.js
+++ b/shared/common-adapters/plain-input.desktop.js
@@ -262,8 +262,14 @@ class PlainInput extends React.PureComponent<InternalProps> {
 
   render = () => {
     const inputProps = this._getInputProps()
-    const css = `::-webkit-input-placeholder { color: ${this.props.placeholderColor ||
-      globalColors.black_40}; }
+    if (this.props.placeholderColor && !this.props.className) {
+      logger.warn(
+        'PlainInput warning: using prop placeholderColor without setting className. The color might be overridden by a later input in the dom. Set a className unique within the current screen to avoid this.'
+      )
+    }
+    const css = `${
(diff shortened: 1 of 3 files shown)


shared/common-adapters/plain-input.desktop.js
```diff
@@ -262,8 +262,14 @@ class PlainInput extends React.PureComponent<InternalProps> {
 
   render = () => {
     const inputProps = this._getInputProps()
-    const css = `::-webkit-input-placeholder { color: ${this.props.placeholderColor ||
-      globalColors.black_40}; }
+    if (this.props.placeholderColor && !this.props.className) {
```
chrisnojima: lets just use emotion instead

shared/common-adapters/plain-input.desktop.js
```diff
@@ -262,18 +263,26 @@ class PlainInput extends React.PureComponent<InternalProps> {
...
-        <style>{css}</style>
-        {this.props.multiline ? <textarea {...inputProps} /> : <input {...inputProps} />}
+        {this.props.multiline ? <StyledTextArea {...inputProps} /> : <StyledInput {...inputProps} />}
       </React.Fragment>
     )
   }
 }
 
+const StyledTextArea = styled.textarea(props => ({
+  '&::-webkit-inner-spin-button': {'-webkit-appearance': 'none', margin: 0},
```
chrisnojima: why 2x?

buoyad: it's `..inner..` and `..outer..`

PR #15317, opened 2019-01-04
force identify table run in ID3

maxtaco:
- before it was getting confused in the caching mess for tracking popups


maxtaco:
Sorry about this bug, should be the fix...

maxtaco:
@chrisnojima I need to leave early, do you want to push merge when CI is green?

chrisnojima:
sure

maxtaco:
i ran the relevant tests by hand

PR #15318, opened 2019-01-04
Corral user into payments list when payment fails

buoyad:
This makes changes when a payment fails in the actual sendPayment call. For context: when this happens, we might not know if the payment went through or not.

* Add a link to the error banner to `Review payments`
* Make `Review payments` take you to a freshly loaded transaction list for the account (spinner while we're waiting)
* Make other ways to exit the dialog (back/`x` button, clicking outside the dialog) close all dialogs instead of going back to the send form

Easy to test with a service at `9db739b5a0`, add a secret note last and try sending the payment.

cc @mlsteele 

r? @keybase/react-hackers

### DIFF
--- a/shared/wallets/confirm-form/index.js
+++ b/shared/wallets/confirm-form/index.js
@@ -36,11 +36,11 @@ const getBannerAction = (props: ConfirmSendProps, banner: BannerType) => {
 }
 
 const ConfirmSend = (props: ConfirmSendProps) => (
-  <Kb.MaybePopup onClose={props.sendFailed ? props.onExitFailed : props.onClose}>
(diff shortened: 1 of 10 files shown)


shared/wallets/confirm-form/index.js
```diff
@@ -24,6 +25,16 @@ type ConfirmSendProps = {|
   readyToSend: string,
 |}
 
+const getBannerAction = (props: ConfirmSendProps, banner: BannerType) => {
```
chrisnojima: why not have the container do this

buoyad:
Thanks @chrisnojima, done. I also added a label to the loading spinner
failed-payment

PR #15320, opened 2019-01-05
attach-to-chat 1: components and stories

songgao:
Some storybook screenshots:

image

image

image

### DIFF
--- a/shared/chat/conversation-list/choose-conversation.desktop.js
+++ b/shared/chat/conversation-list/choose-conversation.desktop.js
@@ -1,0 +1,55 @@
+// @flow
+import * as React from 'react'
+import * as Kb from '../../common-adapters'
+import * as Styles from '../../styles'
+import ConversationList from './conversation-list-container'
+import type {Props} from './choose-conversation'
+
+type State = {expanded: boolean}
+
+class ChooseConversation extends React.Component<Props & Kb.OverlayParentProps, State> {
+  state = {
+    expanded: false,
+  }
+  _toggleOpen = () => {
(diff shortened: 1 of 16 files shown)


jzila:
@songgao wouldn't it be better if this acted more like the search for conversation, rather than the inbox browser?

mmaxim:
Meta comment: I wonder if it would just be better and simpler to just do like what the mobile share extension does for this. Drop the whole small/big team deal here (its not important for this view), add a search bar, and sort the whole thing on recency (recency gets you what you want 99% of the time with chat). Otherwise, if I want to share into a conversation in the +2 more area (which on my account is +460 more), I have to know to tap there, scroll down to find it and then tap again. In fact, to integrate you could use the same API the mobile share extension uses, and avoid some messy integration with the chat Redux store. 

I know this is a design comment, but this seems in the early stages and maybe some stuff could be tweaked.

mmaxim:
Is there something on Zeplin for this?

jzila:
Yeah @songgao implemented pretty much exactly what's in Zeplin: https://app.zeplin.io/project/56b25ffc64fc91773a73aaf8/screen/5a95d343b034942dab54d67a

jzila:
https://app.zeplin.io/project/56b25ffc64fc91773a73aaf8/screen/5a72122378046f597afa995e

songgao:
@jzila @mmaxim Good points on the need for a search bar. I'm happy to make changes.

What do @keybase/design folks think?

songgao:
Talked to Cécile offline. Will do what you guys suggest. I'm gonna close this for now and work on the new design. Thanks all!

PR #15321, opened 2019-01-05
Surya/core 9529/kbfs systemd replace run keybase

jzila:
Moving from keybase/kbfs#1998.

@heronhaye @songgao @taruti


heronhaye:
@strib, song left some comments on the old PR you might want to look at as well.

heronhaye:
subsumed in https://github.com/keybase/client/pull/15286/files which is still wip.

PR #15322, opened 2019-01-05
libdokan: Refuse to execute files

jzila:
Ported from keybase/kbfs#2001


jzila:
@taruti

strib:
Ok thanks.  Brutal.  We might want to have a toast error explaining that to the user, since I suspect a lot of people will be confused.

PR #15325, opened 2019-01-05
conflict_resolver: track failed resolutions

jzila:
Porting from keybase/kbfs#1990

Create a database to keep track of failed conflict resolutions. In the future this will be used to prevent conflict resolution from being attempted when it has failed repeatedly.

Issue: KBFS-3680

@jakob223 @strib

### DIFF
--- a/go/kbfs/libkbfs/conflict_resolver.go
+++ b/go/kbfs/libkbfs/conflict_resolver.go
@@ -7,6 +7,9 @@ package libkbfs
 import (
 	"encoding/json"
 	"fmt"
+	"os"
+	sysPath "path"
+	"runtime/debug"
 	"sort"
 	"strings"
 	"sync"
@@ -17,7 +20,10 @@ import (
 	"github.com/keybase/client/go/kbfs/kbfsmd"
 	"github.com/keybase/client/go/kbfs/kbfssync"
 	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/go-codec/codec"
 	"github.com/pkg/errors"
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/storage"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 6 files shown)


go/kbfs/libkbfs/conflict_resolver.go
```diff
@@ -3183,6 +3194,121 @@ outer:
...
+		return
+	}
+
+	thisCR := &conflictsSoFar[len(conflictsSoFar)-1]
+	thisCR.ErrorTime = cr.config.Clock().Now()
+	if receivedErr != nil {
+		thisCR.ErrorString = receivedErr.Error()
+	}
+	if panicVar != nil {
+		thisCR.PanicString = fmt.Sprintf("%s", panicVar)
```
strib: It would also be great to get the full stack here.  According to https://groups.google.com/forum/#!topic/golang-nuts/MB8GyW5j2UY, `debug.Stack()` should give you the right thing, so maybe we can append it to the `panicVar` string?

go/kbfs/libkbfs/conflict_resolver.go
```diff
@@ -3183,6 +3194,121 @@ outer:
...
+	}
+
+	conflictsSerialized, err = cr.config.Codec().Encode(conflictsSoFar)
```
strib: This encode+put can also be in a helper.

go/kbfs/libkbfs/conflict_resolver.go
```diff
@@ -3183,6 +3194,121 @@ outer:
...
+//  - in the event of failure, it logs that CR failed and tries to record the
+//    failure to the DB.
+func (cr *ConflictResolver) recordFinishResolve(
+	ctx context.Context, ci conflictInput, receivedErr error) {
+	db := cr.config.KBFSOps().GetConflictResolutionDB()
+	panicVar := recover()
+
+	// If we neither errored nor panicked, this CR succeeded and we can wipe
+	// the DB entry.
+	if receivedErr == nil && panicVar == nil {
```
strib: @jakob223 Oh one more thing: I think this should be something like:

```go
if (receivedErr == nil || receivedErr == context.Canceled) && panicVar == nil {
```

Because legitimate cancellations happen a lot, in fact they happen every time a new write is done to the unmerged branch.  So they should basically be treated like nil errors.  (You can also just modify the caller of this function to pass in a `nil` err in that case, if you think that's cleaner.)

go/kbfs/libkbfs/conflict_resolver.go
```diff
@@ -3183,6 +3194,121 @@ outer:
 	return nil
 }
 
+const conflictRecordVersion = 1
+
+type conflictRecord struct {
+	Version int
+	time.Time
```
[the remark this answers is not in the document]
jakob223: No particular reason. It was causing problems with encoding this record though, so I changed it in one of the PRs that depends on this one. Should I change it here or just wait for that to merge in?

PR #15333, opened 2019-01-05
go/vendor: Update github.com/xanzy/ssh-agent for Go 1.11 on Windows

taruti:
One dev laptop was updated to Go 1.11 and this was needed for things to compile.


jzila:
CI has go1.10.7, so yes.

PR #15335, opened 2019-01-05
fix channel mention navigate

mmaxim:
@keybase/react-hackers 

This fixes clicking a channel name mention.


mmaxim:
No idea why this is failed, since I don't see anything failed on Jenkins, so merging it.

PR #15346, opened 2019-01-07
Add shavendorstellar to Jenkinsfile

patrickxb:
Is this the best place for it?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -365,6 +365,9 @@ def testGo(prefix, packagesToTest) {
     // Make sure we don't accidentally pull in the testing package.
     sh '! go list -f \'{{ join .Deps "\\n" }}\' github.com/keybase/client/go/keybase | grep testing'
 
+    // Make sure vendored stellar external packages haven't changed
+    sh 'make shavendorstellar'
+
     def packageTestList = packagesToTest.keySet()
     println "Go packages to test:\n${packageTestList.join('\n')}"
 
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -365,6 +365,9 @@ def testGo(prefix, packagesToTest) {
     // Make sure we don't accidentally pull in the testing package.
     sh '! go list -f \'{{ join .Deps "\\n" }}\' github.com/keybase/client/go/keybase | grep testing'
 
+    // Make sure vendored stellar external packages haven't changed
+    sh 'make shavendorstellar'
```
jzila: It'd be better in the `test_linux` section so it's only run once.

patrickxb: I moved it, but feel free to give me a line number if still in the wrong place.

Jenkinsfile
```diff
@@ -101,8 +101,12 @@ helpers.rootLinuxNode(env, {
...
               checkDiffs(['./go/', './protocol/'], 'Please run \\"make\\" inside the client/protocol directory.')
             }
+
             parallel (
               test_linux: {
+                // Make sure vendored stellar external packages haven't changed
+                sh 'make shavendorstellar'
+
                 def packagesToTest = [:]
                 if (hasGoChanges) {
```
jzila: I'd put it right here.

patrickxb: how do I ensure cwd is correct?  (it currently isn't)

jzila: Oops sorry about the delay. Looks like you got it.

PR #15348, opened 2019-01-07
readlink in BinPath

songgao:


heronhaye:
This looks good @songgao. Maybe we should have a test for symlink resolution in each OS?

songgao:
@heronhaye thanks for reviewing! Not sure how to add that test since I couldn't repro the issue by just using `install.UpdaterBinPath()`. It might have to do with launchd or something, but I'm not sure.

heronhaye:
My guess (I don't think linux has this behavior) is it's something like an executable called "a.out" which is a symlink to the real file "b.out" will print "a.out" as the path of executable.

go/.go_package_deps_darwin
```diff
@@ -603,6 +603,7 @@
       "github.com/keybase/client/go/kbfs/test" : 1,
       "github.com/keybase/client/go/kbfs/tlf" : 1,
       "github.com/keybase/client/go/kbhttp" : 1,
+      "github.com/keybase/client/go/kbnm" : 1,
```
patrickxb: what's going on here?  what is this .go_package_deps_darwin file?

songgao: Yeah that's what I did, which resulted in these change. I think Patrick was asking what this is for.

jzila: Oh right you're asking what's going on. These files now contain the dependency linkages between packages. This allows CI to perform tests such that only the minimal set of tests is executed in CI, generally improving performance.

jzila: We could exclude them from the code, but then we'd have to add a hard 2 minute step into CI that can't be executed in parallel with other steps. So instead we run that step in parallel and notify the engineer if the dependencies need to be recalculated.

jzila: I will add this to the README.

patrickxb:
i thought it was in libcmdline the first time i looked at the PR?

songgao:
@patrickxb yeah it was libcmdline, but on windwos it has circular dep. I'm moving it into a separate package called utils  now

songgao:
ready for another look!

songgao:
Bump!

patrickxb:
I wish there was a way to test this...

songgao:
@patrickxb with some simple playing around, I couldn't repro the symlink issue in the first place. So not sure how to write a test that fails the problematic commit, but passes this. Would you rather I add a test for `utils.BinPath()` for the happy path?

patrickxb:
does `keybase update check` work on a mac with this patch?

songgao:
Yeah, I tried it using the magic command that builds directly into `/Applications/Keybase.app`

songgao:
@patrickxb mind thumbing this? Thanks!

jzila:
This seems like a risky change the week before a big release. Can we hold off merging this until next week?

songgao:
Sure

songgao:
@patrickxb @jzila now that code is unfrozen, do we want to get this in? It'd be nice to get rid of `osext` in case we accidentally bump it again.

jzila:
Yep, I'm fine with it if Patrick is.

patrickxb:
sure

PR #15354, opened 2019-01-07
Add "(approximately $X.XX ABC)" to direct payments without display currencies

buoyad:
cc @patrickxb for core review (it turns out `WorthCurrency` was unused, so I removed it)

This adds a field `WorthAtSendTime` to `Payment*Local` when there's no display currency set. It's shown it in the tx row, tx details, payment messages, and payment message popup. 

We should show this for requests with no display currencies too, but that was higher hanging fruit; I made a core ticket for it.

- [ ] Storyshots before merge

r? @keybase/react-hackers


shared/wallets/transaction/index.js
```diff
@@ -301,22 +313,25 @@ const TimestampLine = (props: TimestampLineProps) => {
...
-        {status ? ` • ` : null}
-        {!!status && (
-          <Text selectable={props.selectableText} type={status === 'Failed' ? 'BodySmallError' : 'BodySmall'}>
-            {status}
-          </Text>
-        )}
-        {status === 'Failed' && !props.detailView && (
-          <>
-          {' '}(<Text selectable={props.selectableText} type='BodySmallSecondaryLink'>
+    <Text selectable={props.selectableText} title={tooltip} type="BodySmall">
```
buoyad: I didn't do anything here, this is prettier

PR #15300, opened 2019-01-04
make toast work ok

zanderz:
Thanks @MarcoPolo

PR #15358, opened 2019-01-07
[New new conversation] - Ability to Remove users on press

MarcoPolo:
@keybase/react-hackers implements a SwapOnClickHoc that shows a different component when the user presses it. I use this to show the remove bubble when the user presses a user bubble.

Looks like this: 
screen shot 2019-01-07 at 3 14 16 pm
 

cc @adamjspooner 

Also fixes a racey conditon in WithTooltip. A similar bug to what @zanderz fixed a bit ago. Fixes the big warning we got related to passing a ref of a Box instead of the underlying div.

### DIFF
--- a/shared/common-adapters/with-tooltip.desktop.js
+++ b/shared/common-adapters/with-tooltip.desktop.js
@@ -9,21 +9,23 @@ import {type Props} from './with-tooltip'
 type State = {
   mouseIn: boolean,
   visible: boolean,
+  attachmentRef: ?React.Component<any>,
 }
 
 class WithTooltip extends React.Component<Props, State> {
   state = {
+    attachmentRef: null,
     mouseIn: false,
     visible: false,
   }
-  _attachmentRef: ?React.Component<any> = null
   _onMouseEnter = () => {
     this.setState({mouseIn: true})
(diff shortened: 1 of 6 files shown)


shared/common-adapters/with-tooltip.desktop.js
```diff
@@ -53,7 +55,7 @@ class WithTooltip extends React.Component<Props, State> {
               this.props.multiline && styles.containerMultiline,
             ])}
             visible={!!this.props.text && this.state.visible}
-            attachTo={() => this._attachmentRef}
+            attachTo={this.state.attachmentRef && this._getAttachmentRef}
```
buoyad: I don't think `this.state.attachmentRef &&` and the `attachTo` type change is necessary, the behavior will be the same if `this._getAttachmentRef` returns null.

MarcoPolo: It's necessary because if you just use `this._getAttachmentRef` the prop `attachTo` doesn't change whether `this.state.attachmentRef` is null or not. This was the same bug that affected https://github.com/keybase/client/pull/15300.

PR #15360, opened 2019-01-07
Fix merge testing

jzila:


jzila:
@akalin-keybase I tagged you for review since I know you've done a bunch of Jenkinsfile work before, but feel free to add someone else if they'd be better to take a look.

PR #15368, opened 2019-01-08
refactor chat-api listen command

joshblum:
some minor tweaks, main fix is if an ephemeral error comes in we now skip the message based on the `showExploding` flag, previously ephemeral was only checked for valid messages

### DIFF
--- a/go/client/wallet_api_handler.go
+++ b/go/client/wallet_api_handler.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"strings"
 
+	"github.com/keybase/client/go/chat/msgchecker"
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/stellar1"
 	"github.com/keybase/go-framed-msgpack-rpc/rpc"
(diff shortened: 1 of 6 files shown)


go/client/wallet_api_handler.go
```diff
@@ -425,7 +426,7 @@ func (c *sendOptions) Check() error {
 			return ErrInvalidAccountID
 		}
 	}
-	if len(c.Message) > 400 {
+	if len(c.Message) > msgchecker.PaymentTextMaxLength {
```
joshblum: cc @patrickxb for this

joshblum: fixed: ` const MaxStellarPaymentNoteLength = 500`

go/protocol/chat1/extras.go
```diff
@@ -720,28 +720,25 @@ func (m UIMessage) GetOutboxID() *OutboxID {
...
-		}
-		if state == MessageUnboxedState_ERROR {
-			return m.Error().MessageType
-		}
-		if state == MessageUnboxedState_OUTBOX {
-			return m.Outbox().MessageType
-		}
-		if state == MessageUnboxedState_PLACEHOLDER {
-			// All we know about a place holder is the ID, so just
-			// call it type NONE
```
zapu: i'd leave the `PLACEHOLDER` branch / case too since the comment seems important

joshblum: `PLACEHOLDER` behaves as the default of `MessageType_NONE` so i figured i'd just leave it unless that changes in the future and we have to change this function

PR #15370, opened 2019-01-08
Add RequestStatus_DONE

patrickxb:


buoyad:
@keybase/react-hackers I removed all request stuff outside the context of chat, all of it was unused and was causing flow issues. The diff that makes this work is [94105b4](https://github.com/keybase/client/pull/15370/commits/94105b447d162d237761cdfbc6c4c28c95c15c62)

PR #15372, opened 2019-01-08
only mark chat as read if you're actually active

chrisnojima:
On osx when you close a window sibling windows will be foregrounded. There are also times where foregrounding a window can focus/defocus the main window.
This has the bad side effect in chat where this can cause you to read the currently selected thread

Steps to repro:
1. Open chat tab w/ a convo selected
2. Bring another app into focus (not devtools, else that can get focused instead)
3. Click on widget 2x

This should bring the main back on top of the app from step 2 and mark the thread read.

This PR changes the flow so you need the userActive flag (which is driven by InputMonitor) in addition to appFocus

I moved InputMonitor into the render thread so its easier to analyze and it does all window / actions stuff anyways so it doesn't really belong on the node side

- [x] Move input monitor to render thread (platform specific saga)
- [x] Remove unused appState
- [x] Never allow both timers to overlap unless we intend to

### DIFF
--- a/shared/actions/platform-specific/index.desktop.js
+++ b/shared/actions/platform-specific/index.desktop.js
@@ -88,28 +88,25 @@ function* handleWindowFocusEvents(): Generator<any, void, any> {
   }, Saga.buffers.expanding(1))
   while (true) {
     const type = yield Saga.take(channel)
-    switch (type) {
-      case 'focus':
-        if (skipAppFocusActions) {
-          console.log('Skipping app focus actions!')
-        } else {
+    if (skipAppFocusActions) {
+      console.log('Skipping app focus actions!')
+    } else {
+      switch (type) {
+        case 'focus':
           yield Saga.put(ConfigGen.createChangedFocus({appFocused: true}))
-        }
-        break
-      case 'blur':
-        if (skipAppFocusActions) {
-          console.log('Skipping app focus actions!')
-        } else {
+          break
+        case 'blur':
           yield Saga.put(ConfigGen.createChangedFocus({appFocused: false}))
-        }
+          break
+      }
     }
   }
 }
 
 function* initializeInputMonitor(): Generator<any, void, any> {
-  const inputMonitor = new InputMonitor()
   const channel = Saga.eventChannel(emitter => {
-    inputMonitor.notifyActive = isActive => emitter(isActive ? 'active' : 'inactive')
+    // eslint-disable-next-line no-new
+    new InputMonitor(isActive => emitter(isActive ? 'active' : 'inactive'))
     return () => {}
   }, Saga.buffers.expanding(1))
 
--- a/shared/actions/platform-specific/index.desktop.js
+++ b/shared/actions/platform-specific/index.desktop.js
@@ -16,6 +16,8 @@ import {kbfsNotification} from '../../util/kbfs-notifications'
 import {quit} from '../../util/quit-helper'
 import {showDockIcon} from '../../desktop/app/dock-icon.desktop'
 import {writeLogLinesToFile} from '../../util/forward-logs'
+import InputMonitor from './input-monitor.desktop'
+import {skipAppFocusActions} from '../../local-debug.desktop'
 
 export function showShareActionSheetFromURL(options: {url?: ?any, message?: ?any}): void {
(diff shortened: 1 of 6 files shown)


shared/actions/platform-specific/input-monitor.desktop.js
```diff
@@ -1,62 +1,73 @@
 // @flow
 
+const timeToConsiderActiveForAwhile = 300000
```
chrisnojima: you can reduce these values to test how this works

shared/desktop/app/app-state.desktop.js
```diff
@@ -19,7 +19,6 @@ export type State = {
   dockHidden: boolean,
   notifySound: boolean,
   openAtLogin: boolean,
-  isUserActive: ?boolean,
```
chrisnojima: not used at all

shared/desktop/renderer/main.desktop.js
```diff
@@ -91,28 +89,6 @@ function setupApp(store, runSagas) {
   })
   SafeElectron.getIpcRenderer().send('install-check')
 
-  var inputMonitor = new InputMonitor(function(isActive) {
```
chrisnojima: moved to actions

chrisnojima:
cc: @malgorithms

chrisnojima:
@keybase/react-hackers bump

chrisnojima:
@keybase/react-hackers bump

chrisnojima:
@keybase/react-hackers bump

shared/actions/platform-specific/input-monitor.desktop.js
```diff
@@ -1,62 +1,73 @@
...
+    window.addEventListener('blur', this.goInactive)
+    this.goListening()
   }
 
-  startActiveTimer = () => {
-    // wait 5 minutes before adding listeners
-    if (this.activeTimeoutID) window.clearTimeout(this.activeTimeoutID)
-    this.activeTimeoutID = window.setTimeout(this.goListening, 300000)
+  _clearTimers = () => {
+    if (this.activeTimeoutID) {
```
buoyad: :sheep: I don't think these `if`s are necessary, `clearTimeout` is a noop with `null`

shared/actions/platform-specific/index.desktop.js
```diff
@@ -78,6 +80,45 @@ const writeElectronSettingsNotifySound = (_, action) =>
...
+  const channel = Saga.eventChannel(emitter => {
+    window.addEventListener('focus', () => emitter('focus'))
+    window.addEventListener('blur', () => emitter('blur'))
+    return () => {}
+  }, Saga.buffers.expanding(1))
+  while (true) {
+    const type = yield Saga.take(channel)
+    switch (type) {
+      case 'focus':
+        if (skipAppFocusActions) {
```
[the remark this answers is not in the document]
chrisnojima: this is the old behavior, this is so you can see it as you click in and out. if you saw it once before the flood of logs you'd likely not see it

shared/actions/platform-specific/index.desktop.js
```diff
@@ -78,6 +80,45 @@ const writeElectronSettingsNotifySound = (_, action) =>
...
+          console.log('Skipping app focus actions!')
+        } else {
+          yield Saga.put(ConfigGen.createChangedFocus({appFocused: false}))
+        }
+    }
+  }
+}
+
+function* initializeInputMonitor(): Generator<any, void, any> {
+  const inputMonitor = new InputMonitor()
```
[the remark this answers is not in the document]
chrisnojima: i had that originally but its useful to be able to do stuff with it outside the channel. i changed stuff back and forth but i think its fine not being closed over

shared/actions/platform-specific/input-monitor.desktop.js
chrisnojima: i think flow wanted this, lemme see

shared/actions/platform-specific/index.desktop.js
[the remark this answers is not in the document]
buoyad: Moving it before the switch would still do it on click in+out, right?

chrisnojima:
@buoyad updated

PR #15375, opened 2019-01-08
expand rpc canceling mechanism to backgrounding the app, and use on chat CORE-9768

mmaxim:
Patch does the following:

1.) Allow the RPC canceler to work with categories of cancelation so we can use some types with some protocols (like backgrounding, which most things are probably not ready to have cancel everything).
2.) Enable for the chat local protocol for background mode. 
3.) Always send stale messages to UI, not just in foreground.

Goal of this PR is to make it so we don't have RPCs on the service live over the app getting suspended, which can cause bad results in the UI.


go/libkb/rpccanceler_test.go
```diff
@@ -0,0 +1,34 @@
+package libkb
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/net/context"
+)
+
+func TestRPCCanceler(t *testing.T) {
```
patrickxb: nice test...

PR #15378, opened 2019-01-08
Only show coins stacked for completed payment messages

buoyad:
Doesn't affect messages with inline payments, those will still always have the stack. cc @cecileboucheron, when should they show up for inline payments? 

r? @keybase/react-hackers


buoyad:
@cecileboucheron Yeah, will do in a future PR

buoyad:
@keybase/react-hackers bump

PR #15379, opened 2019-01-08
add temp right spacing to header hoc

chrisnojima:
Currently the header hoc changes give a ton of empty space to the left action / right action which eats into the title space. Instead we should maximize space for the title and have actions only take up their own space.
I didn't want to touch this a ton since adam is in here so this is a minimal fix. If you go into a transaction detail you can see on master it shows up as 'Transaction...' instead of 'Transaction Details'

I think this could be done correctly by placing left/right actions on the other side but opacity 0 and have them share a container so the sizes will be the max of both (and the same). anyways we can discuss options

@keybase/react-hackers 
@adamjspooner

### DIFF
--- a/shared/common-adapters/header-hoc/index.native.js
+++ b/shared/common-adapters/header-hoc/index.native.js
@@ -35,6 +35,17 @@ export class HeaderHocHeader extends React.Component<Props, State> {
         ]
       : null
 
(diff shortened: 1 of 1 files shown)


chrisnojima:
@keybase/react-hackers bump

PR #15380, opened 2019-01-08
Fix relay overflow on mobile

buoyad:
Before:
image

After:
image

r? @keybase/react-hackers

### DIFF
--- a/shared/wallets/transaction/index.js
+++ b/shared/wallets/transaction/index.js
@@ -301,22 +301,25 @@ const TimestampLine = (props: TimestampLineProps) => {
       break
   }
   return (
-      <Text selectable={props.selectableText} title={tooltip} type="BodySmall">
-        {human}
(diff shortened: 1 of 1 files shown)


shared/wallets/transaction/index.js
```diff
@@ -451,9 +454,11 @@ const styles = styleSheetCreate({
     padding: globalMargins.tiny,
     paddingRight: globalMargins.small,
   },
+  flexOne: {flex: 1},
   lineThrough: {
     textDecorationLine: 'line-through',
   },
+  marginLeftAuto: {marginLeft: 'auto'},
```
chrisnojima: Surprised this passes flow for mobile

buoyad: Why?

chrisnojima: didn't realize yoga supported it

PR #15383, opened 2019-01-09
add List2

songgao:


shared/common-adapters/list.stories.js
```diff
@@ -0,0 +1,78 @@
...
+import Box, {Box2} from './box'
+import Text from './text'
+import List2 from './list2'
+import * as Sb from '../stories/storybook'
+import * as Styles from '../styles'
+
+const load = () =>
+  Sb.storiesOf('Common/List2', module)
+    .addDecorator(Sb.scrollViewDecorator)
+    .add('fixed - small', () => (
```
chrisnojima: Should not use the scroll view decorator.  This is a scroll view and we can't nest

shared/common-adapters/list2.desktop.js
```diff
@@ -0,0 +1,59 @@
...
+      return String(index)
+    }
+
+    const keyProp = this.props.keyProperty || 'key'
+    return item[keyProp]
+  }
+
+  _fixed = ({height, width}) => (
+    <FixedSizeList
+      style={Styles.collapseStyles([styles.list, this.props.style])}
```
chrisnojima: don't need styles.list or collapse

shared/common-adapters/list2.desktop.js
```diff
@@ -0,0 +1,59 @@
...
+      itemSize={this._variableItemSize}
+      estimatedItemSize={this.props.estimatedItemHeight}
+    >
+      {({index, style}) => <div style={style}>{this.props.renderItem(index, this.props.items[index])}</div>}
+    </VariableSizeList>
+  )
+
+  render() {
+    return typeof this.props.itemHeight === 'number' ? (
+      <AutoSizer>{this._fixed}</AutoSizer>
```
chrisnojima: better to have one autosizer
```
<AutoSizer>
{
typeof this.props.itemHeight === 'number' ? this._fixed : this._variable}
}
```

shared/common-adapters/list2.desktop.js
```diff
@@ -0,0 +1,59 @@
...
+      itemCount={this.props.items.length}
+      itemKey={this._keyExtractor}
+      itemSize={this.props.itemHeight}
+    >
+      {({index, style}) => <div style={style}>{this.props.renderItem(index, this.props.items[index])}</div>}
+    </FixedSizeList>
+  )
+
+  // $FlowIssue doesn't know itemHeight can't be number here -- we only call this when it's a function.
+  _variableItemSize = index => this.props.itemHeight(index, this.props.items[index])
```
chrisnojima: if you made this._variable = ({height, width, itemHeight}) (aka call from render) then flow would know its a valid function

shared/common-adapters/list2.desktop.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+import React, {PureComponent} from 'react'
+import AutoSizer from 'react-virtualized-auto-sizer'
+import * as Styles from '../styles'
+import {FixedSizeList, VariableSizeList} from 'react-window'
+import type {Props} from './list2'
+
+class List2 extends PureComponent<Props<any>, void> {
```
chrisnojima: why use any, can't you just make class List2<T> ... Props<T>?

shared/common-adapters/list2.native.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+  _itemRender = ({item, index}) => {
+    return this.props.renderItem(index, item)
+  }
+
+  _memoizedOffsetGetter = memoize((data, index) => {
+    if (index === 0) {
+      return 0
+    }
+    if (!index) {
```
chrisnojima: why?

shared/common-adapters/list2.native.js
```diff
@@ -0,0 +1,81 @@
...
+      return {
+        index,
+        length: this.props.itemHeight,
+        offset: this.props.itemHeight * index,
+      }
+    }
+    return {
+      index,
+      length: this.props.itemHeight(index, data[index]),
+      offset: this._memoizedOffsetGetter(data, index),
```
chrisnojima: why is this memozied like this? every call it'll get thrown away

shared/common-adapters/list2.native.js
```diff
@@ -0,0 +1,81 @@
...
+      return 0
+    }
+    if (!index) {
+      return 0
+    }
+    // $FlowIssue can't figure out type for data.
+    return this._memoizedOffsetGetter(index - 1) + this.props.itemHeight(index, data[index])
+  })
+
+  _getItemLayout = (data, index) => {
```
chrisnojima: i don't think we likely even want to make a value for this, it can be a prop passed in. its fine for the fixed height version but we likely should skip the variable case

shared/common-adapters/list2.native.js
chrisnojima: additionally i dont think this will be required to be called in order so i don't think you can generalize this

shared/common-adapters/list2.native.js
```diff
@@ -0,0 +1,81 @@
...
+
+    const keyProp = this.props.keyProperty || 'key'
+    return item[keyProp]
+  }
+
+  render() {
+    return (
+      <View style={collapseStyles([styles.outerView, this.props.style])}>
+        {/* need windowSize so iphone 6 doesn't have OOM issues */}
+        <View style={globalStyles.fillAbsolute}>
```
chrisnojima: why is this view needed?

shared/common-adapters/list2.native.js
songgao: I did some simple test and it was going through the array from beginning to end. So memoization like this means we're always calculating the offset based on the previous offset, which saves us from going from the beginning the array everytime. Why is it thrown away?

shared/common-adapters/list2.native.js
songgao: this is just copied over from list1.native.js. I'll see if I can remove some of these containers.

shared/common-adapters/list2.native.js
songgao: Sorry I'm not sure I understand what you mean. There's a `itemHeight: number | (index: number, item: Item) => number` passed in through `props`. This function just converts the format into what the native list wants.

shared/react-native/ios/Keybase/Storybook.m
```diff
@@ -24,7 +24,7 @@ + (BOOL)requiresMainQueueSetup
 - (NSDictionary *)constantsToExport
 {
   // Set this to true to enable storybook mode
-  return @{@"isStorybook": @false};
+  return @{@"isStorybook": @true};
```
songgao: 🤦🏻‍♂️

shared/common-adapters/list2.native.js
chrisnojima: that won't always be true, it can skip things depending on how you scroll etc

shared/common-adapters/list2.native.js
chrisnojima: this function is optional to the underlying flat list. we don't/can't generalize this in the generic case so lets not implement it by default. we can implement it if its a fixed size since we can implement it correctly. we can chat about htis

shared/common-adapters/list2.native.js
chrisnojima: note this is a memoize-one, aka the last value is saved so every item it'll get thrown away

songgao: Even if it skips, it still works. For example, if it memoized index=2, then suddently skipped to 4. We'd try to find 4, which isn't memoized, which would then look for 3, which isn't there either, causing it to look for 2, which exists. All these are recursive, so non of them would override the memoized value until they return. So we'd always benefit from the latest memoized value.

It won't help if it's going backwards though, which I guess can only be solved by having a dictionary memoizing all indices. Would you prefer that?

PR #15384, opened 2019-01-09
split some chat rows for reusability

songgao:


shared/chat/common/avatars.js
```diff
@@ -1,8 +1,8 @@
 // @flow
 import * as React from 'react'
 import shallowEqual from 'shallowequal'
```
chrisnojima: i don't think we need 'common' if its used by multiple things it can just go up a level

shared/chat/common/selectable-big-team-channel.js
```diff
@@ -69,6 +69,8 @@ class FilterBigTeamChannel extends PureComponent<Props, State> {
   }
 }
 
+const smallRowHeight = isMobile ? 64 : 56
```
chrisnojima: we cannot disconnect this code from RowSizes or the virtualized layout of the inbox will get broken. its one shared value

shared/chat/common/selectable-small-team.js
```diff
@@ -72,10 +72,12 @@ class FilterSmallTeam extends React.PureComponent<Props, State> {
   }
 }
 
+const smallRowHeight = isMobile ? 64 : 56
```
chrisnojima: same here

shared/chat/common/avatars.js
songgao: So for this, should it be at `chat/avatars.js`? It's used in both `chat/inbox` and `chat/conversation-list`

shared/chat/selectable-big-team-channel.js
```diff
@@ -69,6 +69,8 @@ class FilterBigTeamChannel extends PureComponent<Props, State> {
   }
 }
 
+export const rowHeight = isMobile ? 64 : 56
```
chrisnojima: why invert this relationship?

songgao: You mean why have the inbox one import this but not the other way around? Because this is where the component is defined, and the inbox filtered row uses this component, so I feel like this is the correct place for the size as well.

chrisnojima: hmm. well the original idea was all the row sizes would be in there so its clear looking at the components that their sizes actually aren't defined in themselves, so you know there's some other entity thats tracking this. else you'll think its just some regular component and its not that clear that it exports these sizes for some reason

songgao: I had to move this up because it'll be used by `chat/conversation-list` as well. What do you think if we move `chat/inbox/row/sizes.js` to `chat/sizes.js`, and still track all size in there?

shared/chat/selectable-big-team-channel-container.js
```diff
@@ -0,0 +1,31 @@
...
+import {namedConnect} from '../util/container'
+
+type OwnProps = {|
+  conversationIDKey: Types.ConversationIDKey,
+  isSelected: boolean,
+  onSelectConversation: () => void,
+|}
+
+const mapStateToProps = (state, {conversationIDKey}) => ({
+  _meta: Constants.getMeta(state, conversationIDKey),
```
chrisnojima: maybe better to extract channelname and teamname here so if other meta changes it won't rerender (i know this is just moved around)

shared/chat/selectable-small-team-container.js
```diff
@@ -0,0 +1,71 @@
+// @flow
+import * as Constants from '../constants/chat2'
+import * as Types from '../constants/types/chat2'
+import SelectableSmallTeam from './selectable-small-team'
+import {namedConnect} from '../util/container'
+
+type OwnProps = {
```
chrisnojima: these should be exact

shared/chat/selectable-small-team-container.js
```diff
@@ -0,0 +1,71 @@
...
+  onSelectConversation: () => void,
+}
+
+const mapStateToProps = (state, ownProps: OwnProps) => {
+  const conversationIDKey = ownProps.conversationIDKey
+
+  return {
+    _hasUnread: Constants.getHasUnread(state, conversationIDKey),
+    _meta: Constants.getMeta(state, conversationIDKey),
+    _username: state.config.username || '',
```
chrisnojima: username is a string so we don't need the || '' anymore

songgao:
bump!

PR #15390, opened 2019-01-09
attach-to-chat 1: components and stories

songgao:
This is the v2 of https://github.com/keybase/client/pull/15320 where we decided to have a new design for the conversation selector.

@cecileboucheron How does this look? Let me know how I should tweak it!

cc @mmaxim @jzila 

image

image

image

depends on https://github.com/keybase/client/pull/15383 and https://github.com/keybase/client/pull/15384


shared/chat/conversation-list/choose-conversation.native.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import * as React from 'react'
+import * as Kb from '../../common-adapters'
+import type {Props} from './choose-conversation'
+
+export default (props: Props) => (
+  <Kb.Box2 direction="horizontal" centerChildren={true} style={{padding: 32}}>
+    <Kb.Text type="BodySmallError">
+      choose-conversation is not supported on mobile. If you see this outside storybook something is wrong.
```
chrisnojima: why even have this? just use choose-conversation.desktop (it works on both) and just never route here

shared/chat/conversation-list/conversation-list-container.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+import {namedConnect} from '../../util/container'
+import ConversationList from './conversation-list'
+
+type OwnProps = {
```
chrisnojima: all these prop types should be exact

shared/chat/conversation-list/conversation-list-container.js
```diff
@@ -0,0 +1,20 @@
...
+  onSelect?: () => void,
+}
+
+const mapStateToProps = state => ({})
+const mapDispatchToProps = dispatch => ({})
+const mergeProps = (stateProps, dispatchProps, ownProps) => ({
+  rows: [],
```
chrisnojima: ???

shared/chat/inbox/row/big-team-channel/container.js
```diff
@@ -5,10 +5,10 @@ import * as Chat2Gen from '../../../../actions/chat2-gen'
 import {BigTeamChannel} from '.'
 import {connect, isMobile} from '../../../../util/container'
 
-type OwnProps = {|
+type OwnProps = {
```
chrisnojima: ???

shared/chat/conversation-list/choose-conversation.native.js
songgao: I was just thinking if someone makes a mistake and uses this on native, this'd make it very clear that it shouldn't happen that way.

shared/chat/conversation-list/conversation-list-container.js
songgao: Just a placeholder; Will be populated in a follow-up PRs

shared/chat/conversation-list/choose-conversation.native.js
chrisnojima: but then we need a .js.flow and a .native that does nothing. lets just not have it

shared/chat/selectable-big-team-channel-container.js
```diff
@@ -4,7 +4,7 @@ import SelectableBigTeamChannel from './selectable-big-team-channel'
 import * as Types from '../constants/types/chat2'
 import {namedConnect} from '../util/container'
 
-type OwnProps = {|
+export type OwnProps = {|
```
chrisnojima: why do we need this

shared/chat/selectable-big-team-channel.js
```diff
@@ -73,29 +83,34 @@ export const rowHeight = isMobile ? 64 : 56
 
 const styles = Styles.styleSheetCreate({
   channelname: Styles.platformStyles({
-    common: {flexBasis: '70%'},
+    common: {
+      flexBasis: '70%',
+      flexShrink: 0.3,
```
chrisnojima: why this change

shared/fs/send-attachment-to-chat/index.desktop.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+import * as React from 'react'
+import * as Types from '../../constants/types/fs'
+import * as Kb from '../../common-adapters'
+import * as Kbfs from '../common'
+import * as Styles from '../../styles'
+import ChooseConversation from '../../chat/conversation-list/choose-conversation'
+import type {Props} from '.'
+
+const SendAttachmentToChat = (props: Props) => (
```
chrisnojima: if we can have these in one file vs 2 i think we should. there's nothing platform specific about the implementation that requires a different file so I think its better to just have index.js and use Styles.isMobile inside

shared/chat/selectable-big-team-channel.js
songgao: flexBasis didn't work well. On desktop, if the team name is too long, then we'd end up having channel name taking 70% space that has a bunch of white space. On mobile if team name is too long, we'd have channel name pushed to far right. Looks like flexShrink doesn't work perfectly either, but I'll play with it and see

songgao: OK I think I've found a way to make it work, by using maxWidth and flexShrink. The result is that channelname only grows if it's overflowing, but only up to 70%. If within 70% channelname can show in full, we just shrink teamname until they fit. If both needs shrinking, then it's 3:7 between teamname and channel name. Here are some screenshots:

image

image

shared/fs/send-attachment-to-chat/index.desktop.js
songgao: Sounds good; I'll merge them in the next PR.

songgao:
@chrisnojima I addressed all feedback except merging the platform files (which I'm gonna do in next PR), and the CSS changes for the big team row (where I made more changes). Ready for another look; thanks!

shared/react-native/ios/Keybase/Storybook.m
```diff
@@ -24,7 +24,7 @@ + (BOOL)requiresMainQueueSetup
 - (NSDictionary *)constantsToExport
 {
   // Set this to true to enable storybook mode
-  return @{@"isStorybook": @false};
+  return @{@"isStorybook": @true};
```
songgao: oops (reminder to myself to revert this before merging)

songgao:
bump!

PR #15391, opened 2019-01-09
fix stories

joshblum:
anything else ?


joshblum:
`yarn test`  passed locally

PR #15392, opened 2019-01-09
Forever spinner sometimes on payment details page

zanderz:
Opening it right after a payment sometimes resulted in a notification from the service about recent payments, which overwrote the whole payment, erasing details and the transaction ID.


shared/wallets/transaction-details/index.js
```diff
@@ -425,6 +425,18 @@ class LoadTransactionDetails extends React.Component<Props> {
   componentDidMount() {
     this.props.onLoadPaymentDetail()
   }
+  componentDidUpdate(prevProps: Props) {
+    // An erased transaction ID likely means the payment was updated,
+    // which means details need to be retrieved again
+    if (
```
chrisnojima: can you explain why this is needed? isn't the saga just updating this itself?

zanderz: if I understand correctly, the store munges payments and payment details into one object, though they are separate across the RPC. Retrieving the details fills out the txID, but a subsequent notification about recentPayments (I think that's the name) can update the store with a payment object that doesn't include a txID. Seems to just replace the object in the store's map, leaving dxID blank. If this happens while the details page is up, it was spinning forever. IOW this is a way for the details page to tell in needs to re-retrieve details. I could repro on a new payment pretty regularly.

Another option is to just let the details page be stale until next time, but this seemed better.

PR #15402, opened 2019-01-09
prefetcher: track wait channels of parent prefetches

strib:
And don't recurse to parent if the instance we were tracked with has already finished.  This can happen in the following scenario.

* A path `a/b/c` gets prefetched.
* The path gets updated via another write to `a'/b'/c`.
* `a` and `b` get canceled.
* `a` gets re-fetched, and `b` gets added to the prefetch list.
* `c` completes and tries to complete its old parent `b`, which prematurely closes the new prefetches for `b` and `c` (which are now only expecting one block, the new `b` prefetch).

This doesn't cause any major problems right now, but once #15328 is in, it really messes up the byte counts.

Issue: KBFS-3575

### DIFF
--- a/go/kbfs/libkbfs/prefetcher.go
+++ b/go/kbfs/libkbfs/prefetcher.go
@@ -171,6 +171,41 @@ func (p *blockPrefetcher) newPrefetch(count int, triggered bool,
 	}
 }
 
+func (p *blockPrefetcher) getParentForApply(
+	pptr BlockPointer, refMap map[BlockPointer]<-chan struct{},
+	ch <-chan struct{}) *prefetch {
+	// Check if the particular prefetch for our parent that we're
+	// tracking has already completed or been canceled, and if so,
+	// don't apply to that parent.  This can happen in the following
+	// scenario:
+	//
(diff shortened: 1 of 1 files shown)


strib:
(By the way @jzila if you were doing reviews and skipped this one because it had a red X, that was just a circle ci connectivity error.  Rerunning now.  No hurry but I thought you might have missed it in your latest round, and it should be a pretty quick review.)

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -184,9 +187,16 @@ func (p *blockPrefetcher) applyToPtrParentsRecursive(
...
+		parentDone := false
+		select {
+		case <-ch:
+			parentDone = true
+		default:
+		}
+
 		parent, ok := p.prefetches[pptr.ID]
-		if !ok {
+		if parentDone || !ok {
```
jzila: Please document this PR's description here as a comment.

strib: Done, and I refactored that logic into a helper so I only had to put the comment in one place.

PR #15408, opened 2019-01-10
Delete staticProofServices

mlsteele:
When adding stuff to `proofServices` having to do it to `staticProofServices` was a bother. This deletes `staticProofServices`. Basically just shuffling around where `G` pointers are stored.

As a side effect the "This uses only the 'static' services which exclude any parameterized proofs " assertion context seems to be deletable. I hope that's a bonus but maybe it's an issue?


mlsteele:
oh kbfs. This is not the way

Issue #15277, opened 2019-01-02
ERROR: Can't create TLF ID for non-team-backed handle


manuel-huez:
Happy New Year!

I've recently reset my account (lost access to all my devices), but I now seem to be unable to encrypt data to my coworkers. Initiating chats also seem to be failing.

I've executed `keybase log send`: 861a8ea2136ff32a6afe1b1c

Thanks!

strib:
@manuel-huez Happy New Year and sorry about this.  I think there's a bug here if you use `keybase encrypt` with a user that you don't yet have a KBFS folder with. Maybe if you try putting a random file into `/keybase/private/you,coworkersname` it'll work?

cc: @maxtaco @heronhaye Saltpack is asking KBFS for keys, but KBFS if the folder doesn't exist yet KBFS is assuming that we won't be asked for keys (i.e., in the implicit team chat case).  So I think Saltpack should do something similar.

manuel-huez:
@strib done! `4275a2782332f2d575d1e21c`

mmaxim:
Probably reset, should be a badge in the UI indicating as such.

khinsen:
I am getting this error message when trying to read the encrypted messages sent to me by the account `spacedrop`. Screenshot attached. There is no team involved, as far as I know.

<img width="802" alt="Screenshot 2019-09-11 at 20 40 59" src="https://user-images.githubusercontent.com/94934/64725388-c54fb000-d4d4-11e9-99cd-017cc29045f5.png">

strib:
@khinsen 	it would be great if you could start a new github issue for this, and include a log ID obtained from `keybase log send`.  Thanks!

PR #15409, opened 2019-01-10
If too many CR failures, don't keep trying to CR.

jakob223:
Whoops, thought I'd put up this PR already..


strib:
@jakob223 same as #15277 15326?

jakob223:
Sigh, I lost it in the migration and more cleanly reimplemented it. It's ~the same.

jakob223:
@strib I prefer this version since the other one had a bug in which the CR DB gets polluted and all the errors get deleted from it even as the CR is ignored.

PR #15410, opened 2019-01-10
Make EK logging less verbose & perf improvements

joshblum:
patch does the following:

- reduces logging around caches, was extremely verbose previously
- removes a `pukring.Sync()` call which would always hit the network but is only needed when generating a new user ek
- reduces calls to `CleanupStaleUserAndDeviceEKs` and `GetOrCreateTeamEK`, could slow things down unnecessarily when unboxing a large number of messages

we should get in before next release

### DIFF
--- a/go/ephemeral/lib.go
+++ b/go/ephemeral/lib.go
@@ -552,14 +552,15 @@ func (e *EKLib) GetTeamEK(ctx context.Context, teamID keybase1.TeamID, generatio
 			e.G().Log.Debug(err.Error())
 			// If we are unable to get the current max generation, try to kick
 			// off creation of a new key.
+			bgctx := libkb.CopyTagsToBackground(ctx)
 			go func() {
-				maxGeneration, err := teamEKBoxStorage.MaxGeneration(ctx, teamID)
+				maxGeneration, err := teamEKBoxStorage.MaxGeneration(bgctx, teamID)
 				if err != nil {
-					e.G().Log.CDebugf(ctx, "Unable to get MaxGeneration: %v", err)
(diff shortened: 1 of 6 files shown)


go/ephemeral/lib.go
```diff
@@ -456,7 +437,10 @@ func (e *EKLib) PurgeCachesForTeamIDAndGeneration(ctx context.Context, teamID ke
 }
 
 func (e *EKLib) GetOrCreateLatestTeamEK(ctx context.Context, teamID keybase1.TeamID) (teamEK keybase1.TeamEk, err error) {
-	defer e.G().CTraceTimed(ctx, "GetOrCreateLatestTeamEK", func() error { return err })()
+	if err = e.checkLogin(ctx); err != nil {
```
joshblum: this was hoisted out of the inner function which has a retry loop

go/ephemeral/lib.go
```diff
@@ -569,14 +544,26 @@ func (e *EKLib) GetTeamEK(ctx context.Context, teamID keybase1.TeamID, generatio
...
+			// If we are unable to get the current max generation, try to kick
+			// off creation of a new key.
+			go func() {
+				maxGeneration, err := teamEKBoxStorage.MaxGeneration(ctx, teamID)
+				if err != nil {
+					e.G().Log.CDebugf(ctx, "Unable to get MaxGeneration: %v", err)
+					return
+				}
+				if generation == maxGeneration {
+					if _, cerr := e.GetOrCreateLatestTeamEK(ctx, teamID); cerr != nil {
```
mmaxim: Use a background context

joshblum: fixed

PR #15414, opened 2019-01-10
test CI with PIE

heronhaye:
(Also not urgent...)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -130,7 +130,7 @@ helpers.rootLinuxNode(env, {
                   ]) {
                     if (hasGoChanges) {
                       dir("go/keybase") {
-                        sh "go build --tags=production"
+                        sh "go build -ldflags \"-s -w\" -buildmode=pie --tags=production"
                       }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -148,17 +148,17 @@ helpers.rootLinuxNode(env, {
                   integrate: {
                     // Build the client docker first so we can immediately kick off KBFS
                     def hasKBFSChanges = packagesToTest.keySet().findIndexOf { key -> key =~ /^github.com\/keybase\/client\/go\/kbfs/ } >= 0
-                    if (hasGoChanges && hasKBFSChanges) {
+                    if (true) {
```
jzila: You can rebase to master and remove all of these `if (true)` segments since now Jenkinsfile changes trigger everything.

heronhaye: nvm, realized the ldflags would trigger it anyway

heronhaye:
ah, sorry about that. this was supposed to be a 'fix package lint errors' pr, but i'll shift it to a different one.

PR #15416, opened 2019-01-10
Wallet main page sectionList tweaks

buoyad:
- No sticky section headers
- Asset row taller on mobile
- Section header more left padding

r? @keybase/react-hackers cc @cecileboucheron


buoyad:
@chrisnojima I added a border to the header for mobile

PR #15419, opened 2019-01-10
Only show coins icon if at least one inline payment succeeded

buoyad:
r? @keybase/react-hackers

### DIFF
--- a/shared/constants/chat2/index.js
+++ b/shared/constants/chat2/index.js
@@ -232,13 +232,9 @@ export const hasSuccessfulInlinePayments = (state: TypedState, message: Types.Me
   if (message.type !== 'text' || !message.inlinePaymentIDs) {
     return false
   }
-  for (let paymentID of message.inlinePaymentIDs) {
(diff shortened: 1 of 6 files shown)


shared/constants/chat2/index.js
```diff
@@ -227,6 +227,20 @@ export const anyToConversationMembersType = (a: any): ?RPCChatTypes.Conversation
   }
 }
 
+const successfulInlinePaymentStatuses = ['completed', 'claimable']
+export const hasSuccessfulInlinePayments = (state: TypedState, message: Types.Message) => {
+  if (message.type !== 'text' || !message.inlinePaymentIDs) {
+    return false
+  }
+  for (let paymentID of message.inlinePaymentIDs) {
```
chrisnojima: i think this could be
`return message.inlinePaymentIDs.some(id => successfulInlinePaymentStatuses.includes(state.chat2.paymentStatusMap.get(id)?.status))
`

PR #15421, opened 2019-01-10
fix react-list mis-measure

chrisnojima:
@keybase/react-hackers so react-list when starting up isn't measuring the items correctly. if you pass uniform it doesn't go through the measurement flow unless something thrashes it (hence us sometimes trying to fix this with forceUpdate).
So there's 3 options
1. don't use react-list (we're already doing this) but thats a larger change
2. use type `variable`  (what this does)


zanderz:
This doesn't fix the load delay in sent recipient suggestions for me - will that come later with ReactWindow?

chrisnojima:
updated, just using variable

PR #15423, opened 2019-01-10
block_retrieval_queue: don't get full block for prefetch status

strib:
When a block is available in the in-memory cache, don't bother getting it from the disk just to get its prefetch status.  Just fetch the prefetch status from disk instead.

A further PR might optimize this more by keeping an LRU cache of prefetch status within the `DiskBlockCache` (and `DiskBlockCacheRemote`) itself, to avoid any disk or RPC calls when possible.

Issue: KBFS-3744


strib:
> This is fine but why not export all of GetMetadata?

Because it didn't seem necessary to serialize all of it through RPC for no reason.

jzila:
Yeah fair enough, and I suppose now protocol changes are cheap.

strib:
6 consecutive unrelated flakes causes the red checkmark.  Merging over the advice of CI.

PR #15424, opened 2019-01-10
disk_block_cache_remote: add an LRU prefetch status cache

strib:
To avoid making RPCs to kbfsfuse just to get the status.  Most of the time, this helper process will be the only one dealing with the git blocks, so it's ok to assume that the status on the disk cache won't change much except through this remote.

[Note: I decided against adding a similar cache to `DiskBlockCache`, since I think leveldb already has one built in.]

Issue: KBFS-3708


jzila:
I bet this will address much of the poor git performance too.

strib:
> I bet this will address much of the poor git performance too.

Yep, hope so!

PR #15425, opened 2019-01-11
a bunch of small fixes for kbfs/chat integration

songgao:
I'm hoping to get this in before the release. Each commit is individually reviewable and revertable.

### DIFF
--- a/shared/actions/fs/shared.js
+++ b/shared/actions/fs/shared.js
@@ -14,16 +14,23 @@ const _getRouteChangeActionForPermissionError = (path: Types.Path) =>
   RouteTreeGen.createNavigateTo({path: [...fsRootRoute, {props: {path}, selected: 'oopsNoAccess'}]})
 
 const makeErrorHandler = (action: FsGen.Actions, retriable: boolean) => (error: any): TypedActions => {
-  if (
-    (typeof error.desc === 'string' && error.desc.includes('does not have read access to')) ||
-    error.desc.includes('You are not a member of this team')
-  ) {
-    // TODO: add a real error code for this
-    return _getRouteChangeActionForPermissionError(
-      // If you get a flow error here after aadding an action that has a 'path'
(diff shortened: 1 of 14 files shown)


shared/fs/common/path-item-action.js
```diff
@@ -15,7 +15,8 @@ import {
   iconCastPlatformStyles,
   type OverlayParentProps,
 } from '../../common-adapters'
-import {PathItemIcon, PathItemInfo} from '../common'
```
jzila: did you want to remove them from `fs/common.js` as well?

jzila:
Oh I tested it too and it looks fine.

shared/actions/fs/shared.js
```diff
@@ -14,16 +14,23 @@ const _getRouteChangeActionForPermissionError = (path: Types.Path) =>
...
+      // public tlf doesn't exist
+      error.desc.includes("Can't create TLF ID for non-team-backed handle")
+    ) {
+      // TODO: add a real error code for this
+      return _getRouteChangeActionForPermissionError(
+        // If you get a flow error here after aadding an action that has a 'path'
+        // field, try rename that field if the field is not of type FsTypes.Path.
+        (action.payload && action.payload.path) || Constants.defaultPath
+      )
+    }
```
jzila: Not sure if this was supposed to display the Oops popup, but it's still hanging on my machine when I go to `/keybase/private/zila.test/abc/nonexistent.txt`

shared/fs/common/path-item-action.js
songgao: They are still supposed to be exported there; this is just because it'd cause circular dependency warning in console.

shared/actions/fs/shared.js
songgao: With this branch? That's expected; https://github.com/keybase/client/pull/15426 is the one that fixes the non-existent thing.

songgao:
@jzila Thanks for the prompt review!

PR #15426, opened 2019-01-11
add proper error model for non-existent paths

songgao:
This should be part of https://github.com/keybase/client/pull/15425 but it's a bit big, so I'm opening a separate PR for review.

@cecileboucheron is working on an icon for this new model; will add it after it's done!


shared/fs/oops/index.js
```diff
@@ -3,12 +3,16 @@ import * as React from 'react'
 import * as Types from '../../constants/types/fs'
 import * as Kb from '../../common-adapters'
 import * as Styles from '../../styles'
+import * as Flow from '../../util/flow'
 import {withProps} from 'recompose'
 import {isMobile} from '../../constants/platform'
 
+export type What = 'no-access' | 'non-existent'
```
jzila: I'd prefer a clearer name here, both for the type and the property below.

jzila:
Oh I tested that same nonexistent file as I commented on in #15425 and it works now.

PR #15434, opened 2019-01-11
Send notification when autoclaim complete

patrickxb:
cc @chrisnojima @buoyad


patrickxb:
How would stellard know that all the autoclaims are complete?

This is for DESKTOP-8573.

mlsteele:
Oh I didn't realize this is a notification for all of the outstanding ones completing.

Still, on the server that's [this case](https://github.com/keybase/keybase/blob/master/go/stellard/server.go#L1475).

mlsteele:
Hopefully the gui is only waiting for this for users who just created a wallet on that device. Cause for users with a wallet ACR might run before the gui connects.

patrickxb:
I believe that's the flow.  The user accepts the disclaimer and then the GUI waits.  Currently they just wait for 6s.  with this notification, i believe they can be more precise.

mlsteele:
If `AcquireAutoClaimLock` errors out then the gui will never show the wallet. I don't know about this strategy. Could just show the wallet and show the payments in the state they are in, which is claiming. Then work on making autoclaim faster.

patrickxb:
GUI could wait for 6s or this notification, but that wouldn't solve the DESKTOP ticket since that is when someone looking at it after 6s, before autoclaim is done.

PR #15437, opened 2019-01-11
block_retrieval_queue_test: fix data race while assembling blocks

strib:


strib:
6 consecutive unrelated flakes causes the red checkmark. Merging over the advice of CI.

PR #15440, opened 2019-01-11
flow 90

chrisnojima:
@keybase/react-hackers small update


MarcoPolo:
for reference: https://github.com/facebook/flow/blob/master/Changelog.md

PR #15443, opened 2019-01-11
Consolidate display currency logic on opening send form

buoyad:
This is the data flow for display currencies given we'll have notifications for when they change.

- Loads accounts on opening send form if we haven't already
- Store display currency on account records
- Sets precedence for display currency 
    - Request currency > XLM if lastSentXLM > "From" acct display currency > Default account display currency.

\-

- [x] Hook up notifications.
- [x] Remove loads on setting etc.

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -90,6 +90,11 @@ const openSendRequestForm = (state, action) => {
   ]
 }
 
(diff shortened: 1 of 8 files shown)


buoyad:
cc @patrickxb let me know when the notifications are ready and I'll hook them up

buoyad:

PR #15444, opened 2019-01-11
Recipient's unfinished relays are PENDING (again)

mlsteele:


mlsteele:
ci failures all look like the other relay thing

PR #15445, opened 2019-01-11
ConversationUpdate gregor notif

joshblum:
listen to new `ConversationsUpdate` and write change in inbox storage. depends on https://github.com/keybase/keybase/pull/3329 for tests. sends a `ConversationStaleUpdate` for tests instead of making a new type although the GUI technically doesn't need this since it already behaves correctly.


joshblum:
i want it to be extendable if we want to add other fields in the future

PR #15447, opened 2019-01-11
Don't let user navigate up while sending

buoyad:
I noticed you could go back to the send form while the `sendPayment` was in progress. This stops that. Also makes the button spin correctly on requesting. r? @keybase/react-hackers

### DIFF
--- a/shared/wallets/confirm-form/container.js
+++ b/shared/wallets/confirm-form/container.js
@@ -4,6 +4,7 @@ import * as Constants from '../../constants/wallets'
 import * as ProfileGen from '../../actions/profile-gen'
 import * as TrackerGen from '../../actions/tracker-gen'
 import * as WalletsGen from '../../actions/wallets-gen'
+import {anyWaiting} from '../../constants/waiting'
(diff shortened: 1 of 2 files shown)


shared/wallets/confirm-form/container.js
```diff
@@ -51,37 +55,42 @@ const mapDispatchToProps = (dispatch, {navigateUp}: OwnProps) => ({
...
-    onBack: stateProps.sendFailed ? dispatchProps.onAbandonPayment : dispatchProps.onBack,
-    onClose: stateProps.sendFailed ? dispatchProps.onAbandonPayment : dispatchProps.onClose,
-    onSendClick: dispatchProps.onSendClick,
-    publicMemo: stateProps.publicMemo,
-    readyToSend: stateProps.readyToSend,
-    sendFailed: stateProps.sendFailed,
-    sendingIntentionXLM: stateProps.sendingIntentionXLM,
-    waitingKey: stateProps.waitingKey,
-  })
+  (stateProps, dispatchProps) => {
```
chrisnojima: lets pull this out into mergeProps

PR #15451, opened 2019-01-11
Add profile proof suggestions RPCs

mlsteele:
RPC for proof suggestions on the profile. Proof suggestions on the profile already exist but the set is hardcoded in js. These should work for parameterized proofs too.

Side goal of this is to keep doing what the client currently does with twitter, pgp, etc even if the server is out to lunch when it comes to parameterized proofs. (The server is currently not serving parameterized proofs on prod).


go/systests/profile_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"testing"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/keybase/client/go/engine"
+	"github.com/keybase/client/go/libkb"
+	keybase1 "github.com/keybase/client/go/protocol/keybase1"
+	"github.com/stretchr/testify/require"
+)
+
+func TestProofSuggestions(t *testing.T) {
```
mlsteele: This test is picky and will make reconfiguring the dev server more annoying. Curious if you think there's another way.

go/service/prove.go
```diff
@@ -83,6 +87,8 @@ func (ph *ProveHandler) StartProof(ctx context.Context, arg keybase1.StartProofA
 
 // Prove handles the `keybase.1.checkProof` RPC.
 func (ph *ProveHandler) CheckProof(ctx context.Context, arg keybase1.CheckProofArg) (res keybase1.CheckProofStatus, err error) {
+	ctx = libkb.WithLogTag(ctx, "PV")
+	defer ph.G().CTraceTimed(ctx, fmt.Sprintf("CheckProof(%v)", arg.SigID), func() error { return err })()
```
joshblum: prefer `CheckProof: SigID: %v` in these fn traces, makes log parsing easier since you know what the arg is without having to go back to this line

go/service/user.go
```diff
@@ -344,6 +345,248 @@ func (h *UserHandler) UploadUserAvatar(ctx context.Context, arg keybase1.UploadU
...
+			return p + maxServerPriority + 1
+		} else {
+			return offlineOrderMap["~other~"] + maxServerPriority + 1
+		}
+	}
+	for i := range suggestions {
+		suggestions[i].Priority = priorityFn(suggestions[i].Key)
+	}
+
+	sort.Slice(suggestions, func(i, j int) bool {
```
joshblum: didn't you want `SliceStable` here?

go/service/user.go
```diff
@@ -344,6 +345,248 @@ func (h *UserHandler) UploadUserAvatar(ctx context.Context, arg keybase1.UploadU
...
+	}
+	if user == nil {
+		return ret, fmt.Errorf("could not load logged-in user")
+	}
+
+	var dummyIcon []keybase1.SizedImage // TODO CORE-9882: Get some icons
+	var suggestions []ProofSuggestion
+	ossifiedSocial := map[string]ProofSuggestion{
+		"github": {
+			Key:           "github",
```
joshblum: why client hardcode?

protocol/avdl/keybase1/user.avdl
```diff
@@ -142,6 +142,36 @@ protocol user {
...
+    string key;
+    string text; // "Prove your Twitter", "Add a PGP key"
+    array<SizedImage> icon; // TODO CORE-9882: this is an empty list for now
+    array<Identify3RowMeta> metas; // for 'new' proof types
+  }
+
+  array<ProofSuggestion> proofSuggestions(int sessionID);
+
+  // A proof the user doesn't have for the "Prove your" picker.
+  record ProofSuggestion {
```
joshblum: why can't this be collapsed with `ProfileProofSuggestion`? it's nice to minimize the number of types so we can share more code for processing them..

protocol/avdl/keybase1/user.avdl
```diff
@@ -142,6 +142,36 @@ protocol user {
 
   void uploadUserAvatar(string filename, union { null, ImageCropRect } crop);
 
+  ProfileProofSuggestionsRes profileProofSuggestions(int sessionID);
+  record ProfileProofSuggestionsRes{
+    array<ProfileProofSuggestion> suggestions;
+    boolean showMore;
```
joshblum: should this just be a `belowFold` flag on each suggestion so click `showMore` doesn't result in another RPC call (that will call the same helper function with a full load/sort etc)

go/service/user.go
```diff
@@ -344,6 +345,248 @@ func (h *UserHandler) UploadUserAvatar(ctx context.Context, arg keybase1.UploadU
...
+			Key:     suggestion.Key,
+			Text:    suggestion.PickerText,
+			Subtext: suggestion.PickerSubtext,
+			Icon:    suggestion.PickerIcon,
+		})
+	}
+	return ret, nil
+}
+
+type ProofSuggestion struct {
```
joshblum: can `keybase1.ProofSuggestion` also have these fields so there is just one type?

go/systests/profile_test.go
joshblum: the externals test is kinda in the same state which was annoying.. it seemed to stabilize though i like the explicitness of these checks

go/service/user.go
```diff
@@ -344,6 +345,248 @@ func (h *UserHandler) UploadUserAvatar(ctx context.Context, arg keybase1.UploadU
...
+		offlineOrderMap[k] = i
+	}
+
+	priorityFn := func(key string) int {
+		if p, ok := serverPriority[key]; ok {
+			return p
+		} else if p, ok := offlineOrderMap[key]; ok {
+			return p + maxServerPriority + 1
+		} else {
+			return offlineOrderMap["~other~"] + maxServerPriority + 1
```
joshblum: why not `len(offlineOrderMap) + maxServerPriority + 1` and ditch `~other~`

go/service/user.go
```diff
@@ -344,6 +345,248 @@ func (h *UserHandler) UploadUserAvatar(ctx context.Context, arg keybase1.UploadU
...
+				Key:           service,
+				ProfileText:   fmt.Sprintf("Prove your %v", serviceType.DisplayName()),
+				ProfileIcon:   dummyIcon,
+				PickerText:    serviceType.DisplayName(),
+				PickerSubtext: serviceType.GetTypeName(), // xxx url or mastodon
+				PickerIcon:    dummyIcon,
+			})
+		}
+	}
+	hasPGP := len(user.GetActivePGPKeys(true)) > 0
```
joshblum: i think these and the web/crypto ones should come from the server as well

protocol/avdl/keybase1/user.avdl
mlsteele: Good point. @chrisnojima do you care if `proofSuggestions` and `proofSuggestionsProfile` are merged?

go/service/user.go
```diff
@@ -344,6 +345,248 @@ func (h *UserHandler) UploadUserAvatar(ctx context.Context, arg keybase1.UploadU
...
+		serviceType := mctx.G().GetProofServices().GetServiceType(service)
+		if serviceType == nil {
+			mctx.CDebugf("missing proof service type: %v", service)
+			continue
+		}
+		if len(user.IDTable().GetActiveProofsFor(serviceType)) > 0 {
+			mctx.CDebugf("user has an active proof: %v", serviceType.Key())
+			continue
+		}
+		if suggestion, ok := ossifiedSocial[service]; ok {
```
joshblum: won't we always use the hardcoded vals above then?

go/service/user.go
mlsteele: this way you can move other, how cool is that. ditching

go/service/user.go
mlsteele: The user should only see a suggestion to prove pgp if they don't have a pgp key. Here that's determined from the loaded user.

go/service/user.go
mlsteele: I'm not sure what you're asking. Yes the key "twitter" will always say "Prove your Twitter" even if the server sends something else.

go/service/user.go
joshblum: we should modify the map based on this value but i think we should drop all of the interspersed hardcoded data

go/service/user.go
joshblum: why don't we want it to be configurable?

mlsteele: 3 weak reasons. Less wiggle room for a malicious server. This is easy. Why would we?

joshblum: - if paramproofs are under server control we should just have a single point of truth, i don't think we gain anything for having some hard coded and some not.
- it's easy but also adds complication to debugging/extending in the future.
- aside from the pvl aspect of old proofs it would be nice if these were finally unified and behaving essentially the same for the other aspects of the system

go/service/user.go
```diff
@@ -344,6 +345,179 @@ func (h *UserHandler) UploadUserAvatar(ctx context.Context, arg keybase1.UploadU
...
+			ProfileIcon:   dummyIcon,
+			PickerText:    serviceType.DisplayName(),
+			PickerSubtext: subtext,
+			PickerIcon:    dummyIcon,
+		}})
+	}
+	hasPGP := len(user.GetActivePGPKeys(true)) > 0
+	if !hasPGP {
+		suggestions = append(suggestions, ProofSuggestion{ProofSuggestion: keybase1.ProofSuggestion{
+			Key:           "pgp",
```
joshblum: can these suggestions be constants defined outside of the function so it reads easier:
`suggestions = append(suggestions, PGPProofSuggestion)`

go/service/user.go
```diff
@@ -344,6 +345,179 @@ func (h *UserHandler) UploadUserAvatar(ctx context.Context, arg keybase1.UploadU
...
+		switch displayConfig.Key {
+		case "zcash.t", "zcash.z", "zcash.s":
+			altKey = "zcash"
+		case "http", "https", "dns":
+			altKey = "web"
+		}
+		serverPriority[displayConfig.Key] = displayConfig.Priority
+		if len(altKey) > 0 {
+			if _, ok := serverPriority[altKey]; !ok {
+				serverPriority[altKey] = displayConfig.Priority
```
joshblum: should this be the min/max value of the `altKey` instead of just the first instance of it?

PR #15459, opened 2019-01-12
Revert "prefetcher: get status from the block retriever instead of request"

strib:
This reverts commit a181f27dbfac333c89d84b31e044f1fba1dd689a, which caused flaky prefetcher behavior.  The underlying issue will need to be fixed in another way.


strib:
Merging despite one flake, to fix a more prominent flake.

PR #15466, opened 2019-01-13
fix banner component and support mobile

songgao:
Following the [banner style guide](https://app.zeplin.io/project/56a15d74e1e098dd3116d10b?seid=5c0f21bafd7d8daf22663ec5), and added actions and close button. I'll have a follow-up PR that uses this for Files errors.

image
image

cc @cecileboucheron I might have mis-remembered some of your in-person feedback. Please let me know if something looks off.


shared/common-adapters/text.native.js
```diff
@@ -130,12 +136,12 @@ function _getStyle(
   type: TextType,
   backgroundMode?: Background = 'Normal',
   lineClampNum?: ?number,
-  clickable?: ?boolean
+  clickable?: ?boolean,
+  forceUnderline: boolean
 ) {
-  if (backgroundMode === 'Normal') return null
```
songgao: @chrisnojima Is there a reason why we disabled underline on mobile? I removed this so the underline below would work.

songgao: Related: I think with this change, on mobile we'll start underlining links, while previously we didn't automatically underline links on mobile.

songgao: I ended up making it not automatically underline links; just when `underline={true}` is provided.

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,112 @@
...
+  actions?: Array<{
+    title: string,
+    onClick: () => void,
+  }>,
+  color: Color,
+  onClose?: () => void,
+  text: string,
+}
+
+const getContainerStyle = memoize((props: Props) =>
```
chrisnojima: we shouldn't memoize common adapters like this. you'll have no control over your cache getting thrashed

chrisnojima: if you want to do this (not that i think this is expensive so we should actually not do this at all) but if you did just make a map of the color to the value and be done

chrisnojima: and in general this is not a good idea to pass props to memoize, esp if you just pulling out one value like color

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,112 @@
...
+  Styles.collapseStyles([styles.container, colorToBackgroundColorStyles[props.color]])
+)
+const getTextStyle = memoize((props: Props) => colorToTextColorStyles[props.color])
+
+const Banner = (props: Props) => (
+  <Box2 direction="horizontal" fullWidth={true} style={getContainerStyle(props)}>
+    <Box2 key="textBox" direction="horizontal" style={styles.textContainer} centerChildren={true}>
+      {props.text.split(' ').map((word, index) => (
```
chrisnojima: why do this

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,112 @@
...
+        <Text key={`word-${index}`} type="BodySmallSemibold" style={getTextStyle(props)}>
+          {word}&nbsp;
+        </Text>
+      ))}
+      {!!props.actions &&
+        props.actions.map(({title, onClick}, index) => (
+          <React.Fragment key={String(index)}>
+            <Text
```
chrisnojima: just use a nested text and get rid of the fragment

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,112 @@
...
+  },
+  iconContainer: {
+    alignSelf: 'flex-start',
+    height: 40,
+  },
+  textContainer: Styles.platformStyles({
+    common: {
+      flex: 1,
+      flexWrap: 'wrap',
+      paddingBottom: 8,
```
chrisnojima: globalMargins

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,112 @@
...
+      paddingTop: 8,
+    },
+    isElectron: {
+      paddingLeft: 64,
```
chrisnojima: globalMargins

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,112 @@
...
+      paddingRight: 64,
+    },
+    isMobile: {
+      paddingLeft: 24,
```
chrisnojima: globalMargins

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,112 @@
...
+          onClick={props.onClose}
+        />
+      </Box2>
+    )}
+  </Box2>
+)
+
+const styles = Styles.styleSheetCreate({
+  container: {
+    minHeight: 40,
```
chrisnojima: globalMargins

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,112 @@
...
+  },
+  icon: {
+    padding: Styles.globalMargins.tiny,
+  },
+  iconContainer: {
+    alignSelf: 'flex-start',
+    height: 40,
```
chrisnojima: globalMargins

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,112 @@
...
+            <Text key="space" type="BodySmallSemibold">
+              &nbsp;
+            </Text>
+          </React.Fragment>
+        ))}
+    </Box2>
+    {!!props.onClose && (
+      <Box2 key="iconBox" direction="vertical" style={styles.iconContainer} centerChildren={true}>
+        <Icon
+          fontSize={Styles.isMobile ? undefined : 12}
```
chrisnojima: globalMargins

cecileboucheron:
@songgao Links should be underlined on mobile.

shared/common-adapters/banner.js
chrisnojima: you can nest text. but only text. so no non-text inside of text

chrisnojima: https://facebook.github.io/react-native/docs/text#nested-text

songgao:
@chrisnojima Thanks fro the review; ready for another look!

songgao:
bump!

songgao:
bump!

shared/common-adapters/banner.js
```diff
@@ -0,0 +1,115 @@
...
+
+const colorToBackgroundColorStyles = {
+  blue: {backgroundColor: Styles.globalColors.blue},
+  green: {backgroundColor: Styles.globalColors.green},
+  grey: {backgroundColor: Styles.globalColors.lightGrey2},
+  red: {backgroundColor: Styles.globalColors.red},
+  yellow: {backgroundColor: Styles.globalColors.yellow},
+}
+
+const colorToTextColorStyles = {
```
buoyad: You could use `styleSheetCreate` for this and `colorToBackground` to reap the native rewards

PR #15467, opened 2019-01-14
use Kb.Banner for fs errors

songgao:
* move the error banners to top, right below the header
* remove "more" button, and just show them all
* add "Let us know" on mobile, directing to feedback screen
* only show tooltip when `admin` is on

screenshots:

screen shot 2019-01-13 at 4 25 55 pm
screen shot 2019-01-13 at 4 25 59 pm
screen shot 2019-01-13 at 4 26 19 pm
screen shot 2019-01-13 at 4 26 27 pm

cc @cecileboucheron


shared/fs/common/errs.js
```diff
@@ -0,0 +1,55 @@
...
+    text={props.msg}
+    color="red"
+    actions={[
+      ...(props.retry ? [{onClick: props.retry, title: 'Retry'}] : []),
+      ...(props.onFeedback ? [{onClick: props.onFeedback, title: 'Let us know'}] : []),
+    ]}
+  />
+)
+
+const Err = features.admin
```
chrisnojima: why this for admin?

cecileboucheron:
image
Can we center-align multiline paths too?

shared/fs/common/errs.js
songgao: It's for displaying error details, so that we don't have to pull up the logs to see which action actually happened. On the other hand user doesn't need to see it.

chrisnojima: you'll see this error in the console etc right? i dislike having us only see some special version of something in the app and not seeing what the user sees

songgao: Yup. OK; I'll just remove the tooltip thing.

songgao:
image

@cecileboucheron Is this better? Or would you rather break that long file name too.

PR #15469, opened 2019-01-14
gzip/base64 installer output

songgao:
This makes it way easier to read the installer logs.


shared/desktop/app/installer.desktop.js
```diff
@@ -99,8 +100,24 @@ const darwinInstall = (callback: CB) => {
     timeout = 90
   }
 
+  const logOutput = (stdout, stderr) =>
```
chrisnojima: what is the purpose of this

songgao: it compress then gzip, so it’s easy to pipe print them out as they were originally. Otherwise they get escaped in the logs and show up as dingle line and are hard to read

PR #15474, opened 2019-01-14
better use of context in GetThreadNonblock

mmaxim:


go/chat/server.go
```diff
@@ -762,17 +766,17 @@ func (h *Server) GetThreadNonblock(ctx context.Context, arg chat1.GetThreadNonbl
...
+		chatUI.ChatThreadCached(ctx, chat1.ChatThreadCachedArg{
 			SessionID: arg.SessionID,
 			Thread:    pthread,
 		})
 		h.Debug(ctx, "GetThreadNonblock: cached response send time: %v", time.Since(start))
-	}()
+	}(localCtx)
 
 	wg.Add(1)
 	go func() {
```
joshblum: why doesn't this take a `localCtx` as a param?

mmaxim: This is for the "remote" portion of the request, only the local portion is supposed to get canceled in this way.

mmaxim:
I put this in because I had an instance where I foregrounded my app and the thread was marked as being fetched offline (and it didn't attempt to retry). Two things went wrong there: one is the context canceled error got lost, and the second was that we dropped the check for the result being marked as offline independent from any error in an earlier patch I put in.

PR #15478, opened 2019-01-14
only use channel name when valid [CORE-9910]

joshblum:
if we can't get the topic name for some reason we wont show a blank channel name in the notification


joshblum:
nothing in the logs and wasn't able to get a repro, from the code it seems like `utils.GetTopicName` could not find the `METADATA` message or the that message is (unlike) `!msg.IsValidFull()`

PR #15483, opened 2019-01-14
disk_block_cache: 4mb block size for blockDb

strib:
Issue: KBFS-3744

### DIFF
--- a/go/kbfs/libkbfs/disk_block_cache.go
+++ b/go/kbfs/libkbfs/disk_block_cache.go
@@ -33,6 +33,7 @@ const (
 	// 10 GB maximum storage by default
 	defaultDiskBlockCacheMaxBytes   uint64 = 10 * (1 << 30)
 	defaultBlockCacheTableSize      int    = 50 * opt.MiB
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/disk_block_cache.go
```diff
@@ -196,6 +197,7 @@ func newDiskBlockCacheLocalFromStorage(
 	}()
 	blockDbOptions := *leveldbOptions
 	blockDbOptions.CompactionTableSize = defaultBlockCacheTableSize
+	blockDbOptions.BlockSize = defaultBlockCacheBlockSize
```
jakob223: Why wasn't there a default before?

strib: There is an existing default, but it's shared amoung all leveldb instances and is a little small for the block db: https://github.com/keybase/client/blob/8e488dd1b95d7626d7ae8d3fc88c56e9f3e054f8/go/kbfs/libkbfs/leveldb.go#L31

PR #15490, opened 2019-01-14
Prefetcher: allow queued prefetches to be cancelled.

jakob223:
Fixes flakes in `TestFolderBlockManagerCleanSyncCache`

### DIFF
--- a/go/kbfs/libkbfs/disk_block_cache.go
+++ b/go/kbfs/libkbfs/disk_block_cache.go
@@ -665,7 +665,8 @@ func (cache *DiskBlockCacheLocal) Put(
 	encodedLen := int64(len(entry))
 	defer func() {
 		cache.log.CDebugf(ctx, "Cache Put id=%s tlf=%s bSize=%d entrySize=%d "+
-			"err=%+v", blockID, tlfID, blockLen, encodedLen, err)
+			"cacheType=%d err=%+v", blockID, tlfID, blockLen, encodedLen,
+			cache.cacheType, err)
 	}()
 	blockKey := blockID.Bytes()
 	hasKey, err := cache.blockDb.Has(blockKey, nil)
@@ -849,6 +850,12 @@ func (cache *DiskBlockCacheLocal) Delete(ctx context.Context,
 	}
 
 	cache.log.CDebugf(ctx, "Cache Delete numBlocks=%d", len(blockIDs))
+	if cache.config.IsTestMode() {
+		for _, bID := range blockIDs {
+			cache.log.CDebugf(ctx, "Cache type=%d delete block ID %s",
+				cache.cacheType, bID)
+		}
+	}
 	return cache.deleteLocked(ctx, blockIDs)
 }
 
--- a/go/kbfs/libkbfs/prefetcher.go
+++ b/go/kbfs/libkbfs/prefetcher.go
@@ -41,6 +41,10 @@ type prefetchRequest struct {
 	prefetchStatus PrefetchStatus
 	action         BlockRequestAction
 	sendCh         chan<- <-chan struct{}
+
+	// obseleted is a channel that can be used to cancel this request while
+	// it is waiting in the queue if the prefetch is no longer necessary.
+	obseleted <-chan struct{}
 }
 
 type ctxPrefetcherTagKey int
@@ -77,6 +81,11 @@ type rescheduledPrefetch struct {
 	timer *time.Timer
 }
 
+type queuedPrefetch struct {
+	waitingPrefetches int
+	channel           chan struct{}
+}
+
 type blockPrefetcher struct {
 	ctx    context.Context
 	config prefetcherConfig
@@ -110,6 +119,10 @@ type blockPrefetcher struct {
 	rescheduled map[kbfsblock.ID]*rescheduledPrefetch
 	// channel that's always closed, to avoid overhead on certain requests
 	closedCh <-chan struct{}
+
+	// map to channels for cancelling queued prefetches
+	queuedPrefetchHandlesLock sync.Mutex
+	queuedPrefetchHandles     map[kbfsblock.ID]queuedPrefetch
 }
 
 var _ Prefetcher = (*blockPrefetcher)(nil)
@@ -123,19 +136,20 @@ func newBlockPrefetcher(retriever BlockRetriever,
 	closedCh := make(chan struct{})
 	close(closedCh)
 	p := &blockPrefetcher{
-		config:               config,
-		makeNewBackOff:       defaultBackOffForPrefetcher,
-		retriever:            retriever,
-		prefetchRequestCh:    NewInfiniteChannelWrapper(),
-		prefetchCancelCh:     NewInfiniteChannelWrapper(),
-		prefetchRescheduleCh: NewInfiniteChannelWrapper(),
-		inFlightFetches:      NewInfiniteChannelWrapper(),
-		shutdownCh:           make(chan struct{}),
-		almostDoneCh:         make(chan struct{}, 1),
-		doneCh:               make(chan struct{}),
-		prefetches:           make(map[kbfsblock.ID]*prefetch),
-		rescheduled:          make(map[kbfsblock.ID]*rescheduledPrefetch),
-		closedCh:             closedCh,
+		config:                config,
+		makeNewBackOff:        defaultBackOffForPrefetcher,
+		retriever:             retriever,
+		prefetchRequestCh:     NewInfiniteChannelWrapper(),
+		prefetchCancelCh:      NewInfiniteChannelWrapper(),
+		prefetchRescheduleCh:  NewInfiniteChannelWrapper(),
+		inFlightFetches:       NewInfiniteChannelWrapper(),
+		shutdownCh:            make(chan struct{}),
+		almostDoneCh:          make(chan struct{}, 1),
+		doneCh:                make(chan struct{}),
+		prefetches:            make(map[kbfsblock.ID]*prefetch),
+		queuedPrefetchHandles: make(map[kbfsblock.ID]queuedPrefetch),
+		rescheduled:           make(map[kbfsblock.ID]*rescheduledPrefetch),
+		closedCh:              closedCh,
 	}
 	if config != nil {
 		p.log = config.MakeLogger("PRE")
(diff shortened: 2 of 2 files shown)


go/kbfs/libkbfs/disk_block_cache.go
```diff
@@ -849,6 +850,10 @@ func (cache *DiskBlockCacheLocal) Delete(ctx context.Context,
 	}
 
 	cache.log.CDebugf(ctx, "Cache Delete numBlocks=%d", len(blockIDs))
+	for _, bID := range blockIDs {
+		cache.log.CDebugf(ctx, "Cache type=%d delete block ID %s",
+			cache.cacheType, bID)
```
jzila: This could be pretty verbose. But I guess we have other very verbose messages (e.g. `Put`)

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -390,10 +396,16 @@ func (p *blockPrefetcher) request(ctx context.Context, priority int,
...
 		// If the block isn't in the tree, we add it with a block count of 1 (a
 		// later TriggerPrefetch will come in and decrement it).
+		obseleted := make(chan struct{})
 		req := &prefetchRequest{ptr, block.NewEmptier(), kmd, priority,
-			lifetime, NoPrefetch, action, nil}
+			lifetime, NoPrefetch, action, nil, obseleted}
 		pre = p.newPrefetch(1, false, req)
 		p.prefetches[ptr.ID] = pre
+		p.queuedPrefetchHandlesLock.Lock()
+		// TODO: figure out what to do in the event that there is already a
```
jzila: This is a pretty big `TODO`; we shouldn't merge without this being figured out.

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -1001,8 +1032,13 @@ func (p *blockPrefetcher) ProcessBlockForPrefetch(ctx context.Context,
 	ptr BlockPointer, block Block, kmd KeyMetadata, priority int,
 	lifetime BlockCacheLifetime, prefetchStatus PrefetchStatus,
 	action BlockRequestAction) {
+	obseleted := make(chan struct{})
 	req := &prefetchRequest{ptr, block.NewEmptier(), kmd, priority, lifetime,
-		prefetchStatus, action, nil}
+		prefetchStatus, action, nil, obseleted}
+	p.queuedPrefetchHandlesLock.Lock()
+	p.queuedPrefetchHandles[req.ptr.ID] = obseleted
```
jzila: This should have the same TODO as above.

go/kbfs/libkbfs/disk_block_cache.go
jakob223: I found this useful when debugging this flake, but I would buy an argument that it's too verbose for normal logging.

strib: Yeah I'd prefer not to have this on by default, though I agree it's useful in tests.  Maybe wrap the `for` loop with `if cache.Mode().IsTestMode()`?

go/kbfs/libkbfs/folder_block_manager.go
```diff
@@ -1376,6 +1376,7 @@ func (fbm *folderBlockManager) doChunkedGetNonLiveBlocks(
 
 func (fbm *folderBlockManager) doCleanDiskCache(cacheType DiskBlockCacheType) (
 	err error) {
+	fbm.log.CWarningf(context.Background(), "Starting to clean %s", cacheType)
```
strib: We should get rid of this too -- I intentionally didn't have a message until after we figured out if cleaning was needed, to avoid log pollution.

go/kbfs/libkbfs/folder_block_manager.go
```diff
@@ -1409,6 +1410,7 @@ func (fbm *folderBlockManager) doCleanDiskCache(cacheType DiskBlockCacheType) (
 				cacheType, recentRev)
 			lastRev = recentRev - 1
 		} else {
+			fbm.log.CDebugf(ctx, "Cannot clean %s", cacheType)
```
strib: Same here.

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -106,6 +107,9 @@ type blockPrefetcher struct {
 	doneCh chan struct{}
 	// map to store prefetch metadata
 	prefetches map[kbfsblock.ID]*prefetch
+	// map to channels for cancelling queued prefetches
+	queuedPrefetchHandles     map[kbfsblock.ID]chan<- struct{}
```
strib: The usual rule we try to follow is that things protected by a specific mutex should be in its own little section, with the mutex listed first.  So these definitions should be down below, surrounded by blank lines, with the mutex first.

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -106,6 +107,9 @@ type blockPrefetcher struct {
...
+	queuedPrefetchHandlesLock *sync.Mutex
```
strib: Should not be a pointer, and thus shouldn't need to be initialized in the constructor below.

go/kbfs/libkbfs/prefetcher.go
strib: Yeah I think we should either keep a slice of channels in `queuedPrefetchHandles`, or keep the full channel in there so we can reuse it in the `prefetchRequest` above if it already exists.

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -699,7 +711,16 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 			ptr := ptrInt.(BlockPointer)
 			pre, ok := p.prefetches[ptr.ID]
 			if !ok {
-				p.log.Debug("nothing to cancel for block %s", ptr)
+				p.queuedPrefetchHandlesLock.Lock()
+				ch, ok := p.queuedPrefetchHandles[ptr.ID]
+				if ok {
+					ch <- struct{}{}
```
strib: This isn't where we should be doing the canceling.  It should be in the actual `CancelPrefetch` function.  Doing it here doesn't do much good, since the the request could have already been processed after the caller decided to cancel it.

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -699,7 +711,16 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 			ptr := ptrInt.(BlockPointer)
 			pre, ok := p.prefetches[ptr.ID]
 			if !ok {
-				p.log.Debug("nothing to cancel for block %s", ptr)
+				p.queuedPrefetchHandlesLock.Lock()
```
strib: Please do longish blocks like this in a helper function, so we can take advantage of `defer` for the unlock.

go/kbfs/libkbfs/prefetcher.go
strib: I think this would make more sense in `triggerPrefetch()` (which is called in one other place).  It can add the channel to the request there, rather than during construction above.

go/kbfs/libkbfs/prefetcher.go
jakob223: I prefer the latter solution, but each of them seems to have a problem w/ deletion:
 - if multiple queued prefetches are using the same channel then we can't know when to delete the channel from `queuedPrefetchHandles` and we get a memory leak
 - if we use a slice, then deleting them gets complicated.

jakob223:
After looking at the prefetcher test code for a while, I still see no clear good way to test this.

go/kbfs/libkbfs/prefetcher.go
strib: Hmm.  I guess you can keep a struct that contains both the single channel and a counter.  Increment the counter on each `triggerPrefetch`, and decrement it when the request is processed.

strib: (And obviously when the counter is 0, you can clean up the map entry.)

jakob223:
@strib  / @jzila other than the test that has been separated into another ticket, I claim that this is ready to merge. Can one of you take another look?

strib:
@jzila the idea for this code was my suggestion.  The problem is that it _can't_ go through the regular flow, because we have to prevent the request from being processed as soon as `CancelPrefetch` is called (which is called from some other goroutine).  Any requests in the queue have to canceled, and there doesn't seem to be a way to do that without adding an extra data structure.

jzila:
Meaning the `prefetcher` queue, right? Those prefetches already exist in `p.prefetches`.

So I have a bunch of questions:
1. If the regular `prefetchCancelCh` isn't enough to handle this (because the cancelations need to preempt handling of the queued fetch instead of competing for them), then we should completely get rid of that channel and always go through this kind of flow for cancelation.
2. If 1 is the case, then why is a specific point in the request handler good enough? Shouldn't we have a context that is canceled such that any point is interrupted?
3. There is no explanation as to the theory of this fix--what is the diagnosed issue? How does this fix address it? That is one reason why I was asking for a test--to make it obvious that the diagnosis is correct, and that the fix addresses that diagnosis.

strib:
@jzila no I don't mean `p.prefetches`.  I mean the requests that are stuck in the channel `p.prefetchRequestCh`.

strib:
> the regular prefetchCancelCh isn't enough to handle this (because the cancelations need to preempt handling of the queued fetch instead of competing for them), then we should completely get rid of that channel and always go through this kind of flow for cancelation.

This is solving a slightly different problem.  The `prefetchCancelCh` is still needed to do proper cleanup of the prefetcher data structures, notify parents, etc.  This new thing is just to prevent requests stuck in `p.prefetchRequestCh` from coming off that queue.

> If 1 is the case, then why is a specific point in the request handler good enough? Shouldn't we have a context that is canceled such that any point is interrupted?

I don't understand the "specific point" reference here.  The request handler isn't involved during this cancellation, except when it takes the request off the channel and sees it's already been canceled and thus should be ignored.

> There is no explanation as to the theory of this fix--what is the diagnosed issue? How does this fix address it? That is one reason why I was asking for a test--to make it obvious that the diagnosis is correct, and that the fix addresses that diagnosis.

The issue is something like this:

* a -> b -> c, with full sync on for the TLF
* Prefetcher queues a request for `b`.
* An update comes in: a' -> b' -> c.
* The update-handling code calls `CancelPrefetch(b)` to cancel the prefetch for the old pointer.
* The sync cache cleanup code runs, deleting any instance of `b` from the sync cache.
* The prefetcher takes the `b` request out of the channel, processes it and reads the `b` block, which pulls the `b` block back into the sync cache, where it is stuck forever since the cleanup already ran.

`b` isn't in `p.prefetches` yet at the time of `CancelPrefetch`. If we try to use the regular cancel flow (and always check for cancellation first), we have to somehow remember that we saw a cancel for a particular request before it comes off the channel, which feels leaky.

There is probably still some slight, very complicated race between the sync cache cleaner and the prefetcher that this new code doesn't solve, but it seems way less likely and would be very hard to coordinate a fix for.

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -750,6 +796,14 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 				continue
 			}
 
+			select {
+			case <-req.obseleted:
+				// This request was cancelled while it was waiting
```
strib: Please end in punctuation.

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -750,6 +796,14 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
...
+				continue
```
strib: I think a log message here would be useful in debugging.  Should be pretty rare.

PR #15492, opened 2019-01-14
Reflect device name when editing message from different device

zanderz:


go/chat/server.go
```diff
@@ -1213,6 +1219,11 @@ func (h *Server) PostEditNonblock(ctx context.Context, arg chat1.PostEditNonbloc
 	parg.IdentifyBehavior = arg.IdentifyBehavior
 	parg.OutboxID = arg.OutboxID
 	parg.Msg.ClientHeader.MessageType = chat1.MessageType_EDIT
+	// Sending device may have changed
+	parg.Msg.ClientHeader.SenderDevice, err = h.getGregorDeviceID()
```
mmaxim: Can just delete all the setting of device ID here, I think this code is old and we set this later on.

go/chat/supersedes.go
```diff
@@ -70,6 +70,12 @@ func (t *basicSupersedesTransform) transformEdit(msg chat1.MessageUnboxed, super
 	mvalid.AtMentionUsernames = superMsg.Valid().AtMentionUsernames
 	mvalid.ChannelMention = superMsg.Valid().ChannelMention
 	mvalid.ChannelNameMentions = superMsg.Valid().ChannelNameMentions
+	if superMsg.Valid().SenderDeviceName != "" {
```
mmaxim: What is the point of this blank check?

zanderz: Because there will be older clients that send superseding messages without a device name?

mmaxim: Don't think so, that is set later on in the pipeline. See my first comment, setting it there was just old code that wasn't needed.

go/chat/server.go
zanderz: for a superseding message? I don't think so but I will try it

zanderz: ok it works without that, thanks

go/chat/supersedes.go
zanderz: ok will remove

zanderz:
@mmaxim how is it now?

go/chat/server.go
```diff
@@ -1126,6 +1126,15 @@ func (h *Server) SetConversationStatusLocal(ctx context.Context, arg chat1.SetCo
 	}, nil
 }
 
+func (h *Server) getGregorDeviceID() (gregor1.DeviceID, error) {
```
zanderz: This is now used in just one place - could revert this part if you want

zanderz:
@mmaxim I think this answers your questions, how does it look now?

mmaxim:
I just pushed up a commit which just remove some of the unneeded code, should also still fix your case.

PR #15496, opened 2019-01-14
More empty account ID checks

patrickxb:
Happened again...i think coming from frontend, but try to handle it better.


patrickxb:
ok, here's an error...i think frontend sees the problem.  i was just trying to squash the black bars.

patrickxb:
passes locally

PR #15503, opened 2019-01-14
fix ghost barePreview layer

songgao:


shared/fs/filepreview/bare-preview.native.js
```diff
@@ -12,7 +12,13 @@ import View from './view-container'
...
 
 const mapDispatchToProps = (dispatch, {routePath}) => ({
-  onBack: () => dispatch(RouteTreeGen.createNavigateUp()),
+  onBack: () =>
+    dispatch(
+      RouteTreeGen.createPutActionIfOnPath({
+        expectedPath: routePath,
+        otherAction: RouteTreeGen.createNavigateUp(),
+      })
+    ),
```
jzila: Will this now navigate back to where you actually came from?:

songgao: Yeah; this particular one isn't actually for barePreview popup, but for the close button on it. Without this change, if user double tap 'Close' quickly they'd end up navigate up twice, going to the parent dir of the dir they're supposed to go to.

songgao:
@jzila Thanks for the review! I'm merging this now. Let me know if you still have concerns.

PR #15506, opened 2019-01-15
simplefs_test: Fix TestRefreshSubscription flakes.

jzila:


go/kbfs/simplefs/simplefs_test.go
```diff
@@ -969,19 +971,38 @@ func (sr *subscriptionReporter) LastPath() string {
...
+	}
+}
+
 func TestRefreshSubscription(t *testing.T) {
 	ctx := context.Background()
 	config := libkbfs.MakeTestConfigOrBust(t, "jdoe")
 	sfs := newSimpleFS(env.EmptyAppStateUpdater{}, config)
 	defer closeSimpleFS(ctx, t, sfs)
-	sr := &subscriptionReporter{Reporter: config.Reporter()}
+	sr := &subscriptionReporter{Reporter: config.Reporter(), lastPathNotify: make(chan struct{}, 1<<30)}
```
jakob223: Why is this channel so enormous?

jzila: I need it to never block on send. `struct{}` is zero-width, so it's safe to do without any additional memory use.
https://dave.cheney.net/2014/03/25/the-empty-struct

PR #15510, opened 2019-01-15
TrackChainLink.GetTrackedUsername never errors

mlsteele:
related https://github.com/keybase/client/pull/15511

This should be equivalent to the current behavior ...


mlsteele:
You think this one over the other one?

maxtaco:
i'm confused and don't understand

PR #15517, opened 2019-01-15
Validate account ID before loading mobile only

buoyad:
cc @patrickxb r? @keybase/react-hackers


patrickxb:
is there any way to check for empty account id one level up?  i.e. do no RPC requests until you have it?

maybe this is the last one, but worried there will be another call happening where the same check is needed.

buoyad:
@patrickxb some RPCs accept an empty account ID so I'm not sure we'd want to do that. I think this and the other case were caused by the same action `selectAccount`, so we can check out why that's happening.

patrickxb:
a check in `selectAccount` sounds good...

PR #15528, opened 2019-01-15
dont identify in chat lookup functions

mmaxim:
Only `EncryptionKey` and `DecryptionKey` (and their exploding equivalents) should identify.


go/chat/teams.go
```diff
@@ -634,14 +631,9 @@ func (t *ImplicitTeamsNameInfoSource) LookupName(ctx context.Context, tlfID chat
...
 	}
 	t.Debug(ctx, "LookupName: got name: %s", impTeamName.String())
-	idFailures, err := t.identify(ctx, tlfID, impTeamName)
-	if err != nil {
-		return res, err
-	}
 	return types.NameInfo{
-		ID:               tlfID,
-		CanonicalName:    impTeamName.String(),
-		IdentifyFailures: idFailures,
```
joshblum: can you remove this field from the struct too?

mmaxim: Yeah I guess nothing reads from it, I'll drop it.

PR #15529, opened 2019-01-15
Check for load more regardless of ignoreScrollToBottom

buoyad:
Moves the check for `loadMoreMessages` outside of the check for `this._ignoreScrollToBottomRefCount > 0`

Examples; `Digging ancient messages...` is the thing to watch.

before (I have to scroll down+up a few times when it gets stuck)
stuck-scrolling

after (I never scroll down)
not-stuck-scrolling

(I changed my client to make the problem easier to repro for those)

r? @keybase/react-hackers


buoyad:
cc @chrisnojima

PR #15532, opened 2019-01-15
prevent error on double join

joshblum:
if there is some legacy reason we need to attempt the server call even if our local store says we are already a member we can plumb an error from the server that we ignore instead


joshblum:
yup! `IsMember` only returns true for ` case chat1.ConversationMemberStatus_ACTIVE, chat1.ConversationMemberStatus_RESET:` and setting the `PREVIEW` status goes through a different handler

PR #15534, opened 2019-01-15
only load Files history when logged in; also clean up load logic

songgao:
I believe this is the cause for those "KBFS client not found" blackbars. The bug is that we add a listener to the `show` event of the menubar window and load the Files edit history in its handle. If the app is just updated, there's a brief period where service/kbfs are not up yet but electron is up. If the widget window is shown during this time, we fire the RPC and get a black bar. Fix it by checking `loggedIn`.

Also some cleanup:

* properly cleanup the handler in `componentWillUnmount` so we don't end up having multiple identical handlers on the event.
* remove `TlfUpdateHoc` as we already load the thing in the handler mentioned above

### DIFF
--- a/shared/__mocks__/electron.js
+++ b/shared/__mocks__/electron.js
@@ -13,6 +13,7 @@ export const remote = {
   Menu: {},
   getCurrentWindow: () => ({
     on: () => {},
+    removeListener: () => {},
   }),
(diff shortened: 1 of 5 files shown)


songgao:
bump!

PR #15535, opened 2019-01-15
fix key warning in widget

songgao:
filename from the same TLF can be same as they can be in different subdirs

### DIFF
--- a/shared/menubar/files.desktop.js
+++ b/shared/menubar/files.desktop.js
@@ -98,9 +98,11 @@ const defaultNumFileOptionsShown = 3
 
 const FileUpdates = (props: FileUpdatesProps & FileUpdatesHocProps) => (
   <Kb.Box2 direction="vertical" fullWidth={true}>
-    {props.updates.slice(0, props.isShowingAll ? props.updates.length : defaultNumFileOptionsShown).map(u => (
-      <FileUpdate key={u.name} {...u} />
-    ))}
+    {props.updates
+      .slice(0, props.isShowingAll ? props.updates.length : defaultNumFileOptionsShown)
+      .map((u, index) => (
+        <FileUpdate key={String(index)} {...u} />
+      ))}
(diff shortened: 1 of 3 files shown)


shared/menubar/files.desktop.js
```diff
@@ -98,9 +98,11 @@ const defaultNumFileOptionsShown = 3
 
 const FileUpdates = (props: FileUpdatesProps & FileUpdatesHocProps) => (
   <Kb.Box2 direction="vertical" fullWidth={true}>
-    {props.updates.slice(0, props.isShowingAll ? props.updates.length : defaultNumFileOptionsShown).map(u => (
-      <FileUpdate key={u.name} {...u} />
-    ))}
+    {props.updates
+      .slice(0, props.isShowingAll ? props.updates.length : defaultNumFileOptionsShown)
+      .map((u, index) => (
+        <FileUpdate key={String(index)} {...u} />
```
jzila: Shouldn't this be globally unique for a given folder? i.e. something like the path?

songgao: That's not too bad; it'll just re-use the components and update everything when things shift around. These rows are pretty simple so I don't think it's gonna be a problem. We actually use this pattern elsewhere, and even have `indexAsKey` in `Kb.List`.

I guess using `path` would be better, but it isn't available here, so we'd need to do some plumbing to make it that way.

chrisnojima: All the key does is help resolve if children move around. Aka if you insert an item at the front and items all moved back an index react could not  rerender them. Else itd rerender them all. These things should be connected so they'd rerender if their own data changed

jzila: Right but that's the problem--a new item at the beginning wouldn't be rerendered if you used the index as a key. Or am I missing something?

songgao:
@jzila please take another look; thanks!

PR #15536, opened 2019-01-16
keybase_service_base: add KSID debug tag for `CurrentSession`

strib:
`CurrentSession` is called from a lot of places that might not have other debug tags, and it would be nice for any RPCs that result from `serviceLoggedIn` to have some kind of tag on them so we can narrow down where they came from.

### DIFF
--- a/go/kbfs/libkbfs/keybase_service_base.go
+++ b/go/kbfs/libkbfs/keybase_service_base.go
@@ -1069,6 +1069,9 @@ func (k *KeybaseServiceBase) getCurrentSession(
 // CurrentSession implements the KeybaseService interface for KeybaseServiceBase.
 func (k *KeybaseServiceBase) CurrentSession(ctx context.Context, sessionID int) (
 	SessionInfo, error) {
+	ctx = CtxWithRandomIDReplayable(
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/keybase_service_base.go
```diff
@@ -1069,6 +1069,9 @@ func (k *KeybaseServiceBase) getCurrentSession(
 // CurrentSession implements the KeybaseService interface for KeybaseServiceBase.
 func (k *KeybaseServiceBase) CurrentSession(ctx context.Context, sessionID int) (
 	SessionInfo, error) {
+	ctx = CtxWithRandomIDReplayable(
```
jakob223: I don't think I understand the discrepancy between your comment about RPCs resulting from `serviceLoggedIn` and this only affecting one specific RPC.

strib: `CurrentSession` isn't really an RPC, it's a local call by other components to get the current session.  It can result it RPCs outside of kbfsfuse, and those are the RPCs that I want tagged.  This is the only caller of `serviceLoggedIn`, which sends is where most of those RPCs get called from.

PR #15539, opened 2019-01-16
change new conversation to not lookup implicit teams multiple times CORE-9884

mmaxim:
The old method would invoke implicit.json 5 times per NewConversationLocal call, now we just do it once.


go/chat/helper.go
```diff
@@ -858,16 +858,43 @@ func newNewConversationHelper(g *globals.Context, uid gregor1.UID, tlfName strin
...
 }
 
-func (n *newConversationHelper) findExisting(ctx context.Context, topicName string, useLocalData bool) (res []chat1.ConversationLocal, err error) {
-	// proceed to findConversations for requested member type
-	return n.findConversations(ctx, n.membersType, topicName, useLocalData)
+func (n *newConversationHelper) getNameInfo(ctx context.Context) (res types.NameInfo, err error) {
+	isPublic := n.vis == keybase1.TLFVisibility_PUBLIC
+	switch n.membersType {
+	case chat1.ConversationMembersType_IMPTEAMUPGRADE:
+		return res, errors.New("cannot create new impteam upgrade conv")
```
joshblum: i think this would cause problems when sending to `zapu@github` for example, this check shouldn't happen unless we don't find an existing conversation

go/chat/helper.go
```diff
@@ -858,16 +858,43 @@ func newNewConversationHelper(g *globals.Context, uid gregor1.UID, tlfName strin
...
+	case chat1.ConversationMembersType_KBFS, chat1.ConversationMembersType_TEAM:
```
joshblum: whey do we allow new creations of `KBFS` chats?

go/chat/helper.go
mmaxim: This is part of the new conversation path, we never invoke that with impteamupgrade, and if we do it is a mistake.

go/chat/helper.go
mmaxim: Tests

mmaxim:
Don't know, this flow is better anyway though.

go/chat/helper.go
joshblum: mm true, i'd just double check that `keybase chat send zapu@github test` works with this patch

PR #15542, opened 2019-01-16
go: Fix more tiny bugs found by staticcheck

taruti:
Fix small bugs (mainly ignored errors) found by staticcheck.

Also delete go/gregor/fastlookup_state.go which was dead code and buggy.

### DIFF
--- a/go/chat/wallet/sender.go
+++ b/go/chat/wallet/sender.go
@@ -113,6 +113,7 @@ func (s *Sender) ParsePayments(ctx context.Context, uid gregor1.UID, convID chat
 	if len(parsed) == 0 {
 		return nil
 	}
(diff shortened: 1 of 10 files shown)


go/chat/wallet/sender.go
```diff
@@ -113,6 +113,7 @@ func (s *Sender) ParsePayments(ctx context.Context, uid gregor1.UID, convID chat
 	if len(parsed) == 0 {
 		return nil
 	}
+	// FIXME error is ignored.
```
strib: cc: @mmaxim on the dropped error here.

mmaxim:
That file gets vendored into the server repo I think, not sure why it isn't just in there if isn't used by the client, but could you just restore it for now?

PR #15552, opened 2019-01-16
Debug logging for empty account ID issue

buoyad:
- Add logging
- Remove unused `refreshPayments`

r? @keybase/react-hackers


buoyad:
I'm changing this so it just removes the unused action

PR #15554, opened 2019-01-16
Don't use the account id from details (could be empty)

patrickxb:


go/stellar/wallet_state.go
```diff
@@ -192,7 +192,7 @@ func (w *WalletState) backgroundRefresh() {
 		rt := a.rtime
 		a.RUnlock()
 
-		mctx := libkb.NewMetaContextBackground(w.G())
+		mctx := libkb.NewMetaContextBackground(w.G()).WithLogTag("WABR")
```
xgess: 👍

PR #15557, opened 2019-01-16
attempt to enable http2 for API calls

mmaxim:
cc @patrickxb @mlsteele


patrickxb:
i thought standard net/http supported http2

mmaxim:
Only if you use `http.DefaultTransport`, otherwise there is just a `nil` pointer on that `http.Transport` we will in with stuff. We also need to change nginx config for this to work. I tried all of this on staging and it looks like it works.

patrickxb:
Starting with Go 1.6, the http package has transparent support for the HTTP/2 protocol when using HTTPS. Programs that must disable HTTP/2 can do so by setting Transport.TLSNextProto (for clients) or Server.TLSNextProto (for servers) to a non-nil, empty map. Alternatively, the following GODEBUG environment variables are currently supported:

https://golang.org/pkg/net/http/

patrickxb:
ok, i was wrong:

```Transport uses HTTP/1.1 for HTTP URLs and either HTTP/1.1 or HTTP/2 for HTTPS URLs, depending on whether the server supports HTTP/2, and how the Transport is configured. The DefaultTransport supports HTTP/2. To explicitly enable HTTP/2 on a transport, use golang.org/x/net/http2 and call ConfigureTransport. See the package docs for more about HTTP/2.```

PR #15562, opened 2019-01-16
Box2: flexShrink 0. alignSelf props. Text center prop

chrisnojima:
@keybase/react-hackers this adds some props that are commonly needed which force us to pass custom styles to Box2 and Text

you can now do
```
<Box2 direction='vertical'>
  <Box2 direction ='horizontal' alignSelf='flex-start'>
```

or

```
<Text type='Header' center={true}>SOMETHING</Text>
```

cc: @cecileboucheron @adamjspooner


chrisnojima:
Maybe text should also be shrink 0

buoyad:
And ellipsize by default? It'd be nice if the adapters made that easier, I think making it the default would break more than it fixes though.

chrisnojima:
k i'm switching this to be opt in easier, i'll add shrink={false}

chrisnojima:
@keybase/react-hackers @buoyad updated. also applied center={true} to a ton of text

PR #15567, opened 2019-01-17
make unpauseBlockPutCh async

songgao:
The deadlocks were mostly on `unpauseBlockPutCh <-`. The receiver side is in the mock `orderedBlockServer` where it receives from the channel for the first put. If the Put doesn't happen (e.g. it errored before actually calling Put), then we'd get stuck here. So just make the channel async.

There might be other bugs i.e. why the Put errored, but they are not exposed in the logs. This would expose those flakes hopefully.

### DIFF
--- a/go/kbfs/libkbfs/tlf_journal_test.go
+++ b/go/kbfs/libkbfs/tlf_journal_test.go
@@ -1439,7 +1439,7 @@ func testTLFJournalPauseBlocksAndConvertBranch(t *testing.T,
 	var lock sync.Mutex
 	var puts []interface{}
 
-	unpauseBlockPutCh := make(chan struct{}, 1)
+	unpauseBlockPutCh := make(chan struct{})
 	bserver := orderedBlockServer{
 		lock:      &lock,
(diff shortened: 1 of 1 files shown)


strib:
I don't think this is the right fix -- the tests depend on knowing that the put has already happened when they successfully send a struct into that channel.  If it's async, that no longer holds and the journal entry checks will likely fail.

If we need to get the logs for some other bug, maybe adding a timeout on that send, with a `t.Fatal`, is the right first step.

songgao:
Hmm it always tries to receive from the `errCh`, which I suspected would error and tell us what the error is.

I doesn't seem like `t.Fatalf` would work as that's equivelant to having a timeout panic. How about having a `select{}` on both channels?

songgao:
actually probably not select; we need both. I'll put up a new commit soon

songgao:
Ran 4 builds and only the forth one failed, due to a different flake. I can't repro the failure locally, either on this branch or master, so I'm gonna just merge this for now. If it happens again we can likely get better logs.

PR #15569, opened 2019-01-17
use the same path from refreshSubscription for notification

songgao:
GUI uses preferred paths, but notification uses canonical path. If they differ, when GUI looks through the list of paths it tracks for different views, it can't find it and would ignore the notification. Fix that by just always using the same path that GUI uses to subscribe to notifications.


songgao:
Thanks for the review! Yeah I went with that approach initially, but realized GUI might still have different paths, so just ended up saving whatever GUI passes in.

PR #15570, opened 2019-01-17
Fix number of searched ppl in new chat

zanderz:


shared/profile/edit-avatar/index.desktop.js
```diff
@@ -347,21 +347,23 @@ class EditAvatar extends React.Component<_Props, State> {
...
-                opacity: this.state.loading ? '0' : '1',
-                position: 'absolute',
-                top: this.state.offsetTop,
-                transition: 'opacity 0.25s ease-in',
-                width: this.state.scaledImageWidth,
-              }}
-              onDragStart={e => e.preventDefault()}
-              onLoad={this._onImageLoad}
-            />
+            {this.state.imageSource && (
```
chrisnojima: This is from that other pr right

PR #15576, opened 2019-01-17
add new conversation command to JSON API

mmaxim:


mmaxim:
cc @malgorithms this allows you to create blank conversations. Might be useful to integrate into the bot frameworks (we could use this to seed conversations with people ahead of the airdrop)

PR #15581, opened 2019-01-17
Improve mobile only test by making it use two fake devices

zapu:


zapu:
*CI SECURED*

@patrickxb can you take a look? this is the test that I initially skipped, then made it only test part of the features, but now I got provisioning working there so it tests the whole mobile only flow.

zapu:
@xgess is out today but if you have any suggestions even after this is merged, let me know!

PR #15594, opened 2019-01-18
disable link in still rows on mobile and cleanup PathItemInfo

songgao:


shared/fs/common/index.js
```diff
@@ -3,4 +3,6 @@ export {default as PathItemAction} from './path-item-action-container'
 export {default as SendInAppAction} from './send-in-app-action'
 export {default as OpenInSystemFileManager} from './open-in-system-file-manager-container'
 export {default as PathItemInfo} from './path-item-info-container'
+export {default as TlfInfo} from './tlf-info-container'
+export {default as AutoChoosingInfo} from './auto-choosing-info'
```
jzila: Can you rename this? It's not obvious what it means. Something like `TlfOrPathInfo`.

PR #15596, opened 2019-01-18
WIP: Perform periodical background merkle root skip audits

pzduniak:
Story CORE-8615

The code is tested, succeeds against the production API. Will require integration with the frontend for sure, including an introduction of a new RPC call in the NotifySession (?).


pzduniak:
Looks like we'll need to integrate with a locally running API anyways, if we want to test the failure code path.

go/service/main.go
```diff
@@ -547,6 +548,28 @@ func (d *Service) runHomePoller(ctx context.Context) {
...
 }
 
+func (d *Service) runMerkleAudit(ctx context.Context) {
+	if !libkb.IsMobilePlatform() {
+		d.G().Log.Debug("MerkleAudit disabled (not desktop, not starting)")
+		return
+	}
+
+	eng := engine.NewMerkleAudit(d.G(), &engine.MerkleAuditArgs{})
+	go func() {
```
maxtaco: Why in the background?

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,157 @@
...
+
+		currentRoot, err := m.G().MerkleClient.LookupRootAtSeqno(m, currentSeqno)
+		if err != nil {
+			return err
+		}
+		currentHash := currentRoot.SkipToSeqno(startSeqno)
+
+		if !startHash.Eq(currentHash) {
+			// Warn the user about the possibility of the server tampering with the roots.
+			return fmt.Errorf("Skip of %d to %d broken", currentSeqno, startSeqno)
```
maxtaco: custom error?

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return keybase1.Seqno(0), err
+	}
+	return keybase1.Seqno(n.Int64()) + lo, nil
+}
+
+func MerkleAuditRound(m libkb.MetaContext) error {
+	if m.G().ConnectivityMonitor.IsConnected(m.Ctx()) == libkb.ConnectivityMonitorNo {
+		m.CDebugf("MerkleAudit giving up offline")
+		panic("offline")
```
maxtaco: why panic?

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,157 @@
...
+		return nil
+	}
+
+	// Acquire the most recent merkle tree root
+	lastRoot := m.G().MerkleClient.LastRoot()
+	if lastRoot == nil {
+		m.CDebugf("MerkleAudit unable to retrieve the last root")
+		panic("last root")
```
maxtaco: why panic?

go/libkb/merkle_client.go
```diff
@@ -63,10 +65,23 @@ func (h1 NodeHashShort) bytes() []byte {
 	return h1[:]
 }
 
+func (h1 NodeHashShort) IsNil() bool {
+	return false
+}
+
+func (h1 NodeHashShort) Eq(h2 NodeHash) bool {
+	return bytes.Equal(h1.bytes(), h2.bytes())
```
maxtaco: use `subtle.Equal` or `hmac.Equal` or something constant-time

go/libkb/merkle_client.go
```diff
@@ -63,10 +65,23 @@ func (h1 NodeHashShort) bytes() []byte {
...
+}
+
 func (h1 NodeHashShort) ExportToHashMeta() keybase1.HashMeta {
 	return keybase1.HashMeta(h1.bytes())
 }
 
+func (h1 NodeHashLong) Check(s string) bool {
+	h2 := sha512.Sum512([]byte(s))
+	return FastByteArrayEq(h1[:], h2[:])
```
maxtaco: also i'd use a constant-time comparison

go/libkb/merkle_client.go
```diff
@@ -127,6 +137,10 @@ func (h NodeHashAny) bytes() []byte {
 	}
 }
 
+func (h NodeHashAny) Eq(h2 NodeHash) bool {
+	return bytes.Equal(h.bytes(), h2.bytes())
```
maxtaco: constant time compare

go/service/main.go
```diff
@@ -547,6 +548,28 @@ func (d *Service) runHomePoller(ctx context.Context) {
...
+		m := libkb.NewMetaContextBackground(d.G())
+		err := engine.RunEngine2(m, eng)
+		if err != nil {
+			m.CWarningf("merkle root background audit error: %v", err)
+		}
+	}()
+
+	d.G().PushShutdownHook(func() error {
+		d.G().Log.Debug("stopping merkle root background audit engine")
```
maxtaco: log with the context

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,157 @@
...
+func randSeqno(lo keybase1.Seqno, hi keybase1.Seqno) (keybase1.Seqno, error) {
+	rangeBig := big.NewInt(int64(hi - lo))
+	n, err := rand.Int(rand.Reader, rangeBig)
+	if err != nil {
+		return keybase1.Seqno(0), err
+	}
+	return keybase1.Seqno(n.Int64()) + lo, nil
+}
+
+func MerkleAuditRound(m libkb.MetaContext) error {
```
maxtaco: log entry and exit into this function

maxtaco: ... unless it's not already logged that is...

go/libkb/env.go
```diff
@@ -187,6 +187,9 @@ type TestParameters struct {
 	// set to true to use production run mode in tests
 	UseProductionRunMode bool
 
+	// Set to override the Keybase API server URI in a test.
+	ServerURI string
```
maxtaco: do you need this or is it sufficient to just set the URI via env?

go/engine/merkle_audit_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/keybase/client/go/libkb"
+	"github.com/keybase/clockwork"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMerkleAuditWork(t *testing.T) {
+	tc := SetupEngineTest(t, "merkleaudit")
+	defer tc.Cleanup()
+
+	tc.G.Env.Test.ServerURI = libkb.ProductionServerURI
```
maxtaco: this isn't great for CI, can it work against dev in CI?

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,157 @@
...
+		m.CDebugf("MerkleAudit unable to retrieve the last root")
+		panic("last root")
+		return nil
+	}
+
+	// We can copy the pointer's value as it can only return nil if root == nil.
+	lastSeqno := *lastRoot.Seqno()
+
+	// Generate a random seqno for the starting root in the audit.
+	startSeqno, err := randSeqno(libkb.FirstProdMerkleSeqnoWithSkips, lastSeqno)
```
maxtaco: there is a function that wraps this value for run mode....

go/engine/merkle_audit.go
pzduniak: Oops, didn't add the files again after finishing debugging. Removed.

go/engine/merkle_audit.go
pzduniak: Same as above.

go/service/main.go
pzduniak: Copied this from PerKeyUserUpgradeBackground, looks like this is a leftover from before a BackgroundTask code reorganization.

go/engine/merkle_audit.go
pzduniak: Alright, looks like this should solve my concerns about running this anywhere but in production.

go/engine/merkle_audit.go
pzduniak: It's logged by (*BackgroundTask).Run.

go/libkb/env.go
pzduniak: Made redundant by the First...WithSkip wrapper that I was not aware of, I will need kbweb access to make this work locally then.

pzduniak: Removed.

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,174 @@
...
+
+		currentRoot, err := m.G().MerkleClient.LookupRootAtSeqno(m, currentSeqno)
+		if err != nil {
+			return err
+		}
+		currentHash := currentRoot.SkipToSeqno(startSeqno)
+
+		if !startHash.Eq(currentHash) {
+			// Warn the user about the possibility of the server tampering with the roots.
+			return merkleAuditError{
```
maxtaco: for now, let's report this error with `CErrorf`.  Everything else will be at level `CDebugf` via `background_task.go`

go/libkb/merkle_client.go
maxtaco: I see.  While you're here, I'd change it to constant time. Seems dangerous to short-circuit when comparing hashes.

jzila:
Yes, this is a PR from a fork, which we don't build.

maxtaco:
ah ok, thanks.. @pzduniak can you open another PR to trigger CI?  thank you

go/engine/merkle_audit.go
maxtaco: Also, can you use `libkb.MerkleClientError` for this?

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,246 @@
...
+
+// NewMerkleAudit creates a new MerkleAudit engine.
+func NewMerkleAudit(g *libkb.GlobalContext, args *MerkleAuditArgs) *MerkleAudit {
+	task := NewBackgroundTask(g, &BackgroundTaskArgs{
+		Name:     "MerkleAudit",
+		F:        MerkleAuditRound,
+		Settings: MerkleAuditSettings,
+
+		testingMetaCh:     args.testingMetaCh,
+		testingRoundResCh: args.testingRoundResCh,
```
zapu: Searching does not find these channels being used anywhere. Did I miss anything?

pzduniak: These channels are used in the tests, I copied this approach from the PUK upgrader.

zapu: ah ok, they are written to in `func (e *BackgroundTask) loop(m libkb.MetaContext) error`. sorry.

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,246 @@
...
+		// Proceed until the last known root
+		if currentSeqno > lastSeqno {
+			break
+		}
+
+		currentRoot, err := m.G().MerkleClient.LookupRootAtSeqno(m, currentSeqno)
+		if err != nil {
+			return err
+		}
+		currentHash := currentRoot.SkipToSeqno(startSeqno)
```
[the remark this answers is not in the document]
zapu: I guess this might happen if the server does not provide skip sequence for the root, but maybe it would bail out somewhere else then. I don't know this code very well

go/libkb/merkle_client.go
```diff
@@ -63,10 +65,23 @@ func (h1 NodeHashShort) bytes() []byte {
 	return h1[:]
 }
 
+func (h1 NodeHashShort) IsNil() bool {
+	return false
+}
+
+func (h1 NodeHashShort) Eq(h2 NodeHash) bool {
+	return subtle.ConstantTimeCompare(h1.bytes(), h2.bytes()) == 1
```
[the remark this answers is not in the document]
pzduniak: I thought about this, I think it'd make more sense to swap this out for an exported HashEq.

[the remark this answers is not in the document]
zapu: ah, right, cause auditor is in another package. I wonder what @maxtaco thinks.

go/engine/merkle_audit.go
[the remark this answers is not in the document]
pzduniak: 165 would error first, just checked it.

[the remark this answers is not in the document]
pzduniak: ... actually I need to recheck this again, considering another usecase.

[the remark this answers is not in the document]
pzduniak: verifySkipSequence gets called in the fetcher, which checks if the skip sequence should be there. Honestly it seems to me that most of the auditor code I wrote is useless, considering that we already verify these things in another place. Maybe just pulling the root should be enough?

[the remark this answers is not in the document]
pzduniak: Checked it anyways.

pzduniak:
Migrated to #15627.

PR #15598, opened 2019-01-18
improve chat JSON API inbox list endpoint CORE-9965

mmaxim:
Patch does the following:

1. Adds support for paginating the inbox in `ListV1`.
2. Adds a new `skip_unbox` parameter that allows the user to skip the unboxing phase of listing the inbox.


mmaxim:
cc @malgorithms you will want to use this `skip_unbox` (probably in combination with pagination) if you ever want to view a bots inbox.

PR #15606, opened 2019-01-19
key_manager: don't error when key gen is too big

strib:
There was an iteam migration race that could have left the `LatestKeyGeneration` of the migrated MD too small.  Instead of erroring in that case, just print a message and let later code take care of the error if the key really doesn't exist.

Issue: KBFS-3774


strib:
Merging unreviewed with @songgao's permission.

songgao:
Could you explain the race a little bit? I think I’m lacking some context to understand the fix.

strib:
@songgao did you read the bulleted explanation in the linked JIRA issue?  https://keybase.atlassian.net/browse/KBFS-3774

songgao:
Oh I opened the ticket on my phone and it had the description collapsed, so I only read the comment. The ticket body looks helpful thanks!

PR #15607, opened 2019-01-19
keybase_service_base: clear the team info cache on migration

strib:
Otherwise, there might be a race where, even though the service has migrated the KBFS keys into the implicit team, KBFS hasn't received the `TeamChangedByID` notification yet and will continue to use its old cache.

Issue: KBFS-3774

### DIFF
--- a/go/kbfs/libkbfs/keybase_service_base.go
+++ b/go/kbfs/libkbfs/keybase_service_base.go
@@ -1336,6 +1336,21 @@ func (k *KeybaseServiceBase) FinalizeMigration(ctx context.Context,
 	if err != nil {
 		return err
 	}
+	if handle.TypeForKeying() == tlf.TeamKeying {
+		// Clear the cache for this implicit team, to ensure we get
+		// all the latest key generations for the team info during the
+		// migration.
+		id := handle.FirstResolvedWriter()
+		if id.IsTeamOrSubteam() {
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/keybase_service_base.go
```diff
@@ -1336,6 +1336,21 @@ func (k *KeybaseServiceBase) FinalizeMigration(ctx context.Context,
 	if err != nil {
 		return err
 	}
+	if handle.TypeForKeying() == tlf.TeamKeying {
```
jzila: Why are we migrating something that is already Team keyed? Or does this handle represent the post-migration state?

strib: The handle itself becomes team-based when the service has already 1) created an implicit team for the parties in the folder, and 2) written the TLF ID into that implicit team's sigchain.  The service does this in between calling `StartMigration` and `FinalizeMigration`. You can see that the actual folder migration in `folder_branch_ops` already relies on the handle being team-based: https://github.com/keybase/client/blob/5cf4f306f3d3574fb99fc7fef82362253749c7ac/go/kbfs/libkbfs/folder_branch_ops.go#L7833-L7836

jzila: Understood, thanks.

PR #15608, opened 2019-01-19
also bust home cache on announcement change

maxtaco:
- fix and test announcement badging


maxtaco:
@mmaxim  / @chrisnojima in testing i found some issues with the current announcement stuff released in client. probably we should wait until this fix goes in before we try it.

maxtaco:
works if you run it on fixed kbweb maste

PR #15616, opened 2019-01-20
Support markdown link syntax

darwin:
Newly Markdown links are properly rendered, e.g. `[My Keybase Link](http://keybase.io "This is a keybase link")` shows as "My Keybase Link" with href http://keybase.io and shows the "This is a keybase link" title on hover. 

## Background story

We have a chat bot which reports various web events into our HipChat rooms (github activity, deployments, tests, etc.). Since HipChat is about to be discontinued I decided to migrate our bot to Keybase instead.

It works well but the problem was that I wasn't able to achieve similar look and feel. HipChat allows me to post a subset of html. Most notably links presented with short descriptions. This is a must when linking to github repos, travis tests and similar long urls - without any ability to present urls with short names I get unwieldy wall of text in keybase chat. True, keybase has link auto-detection but the links appear inline without any way how to shorten them.

Markdown supports links with descriptions which I need here. Also I think it should be supported. So I decided to dive into keybase codebase and implement it.  Hope this could be accepted. Thanks. 

At least it was a nice excuse to play with keybase code a bit 😄


freerainboxbox:
This feature would be useful for read-only channels that need pretty-print links, such as to video tutorials, pages, etc. or even a manifest of file pointers to make downloads from KBFS simpler

PR #15617, opened 2019-01-20
change mention handling to use decoration system instead of pure md parser

mmaxim:
**Go**
1. Add `DecorateWithMentions` to use `DecorateBody` to label @mentions and channel name mentions in text. 
2. Add decorated text to outbox messages.
3. Factor out the quote replacer function into `util` so it can be used in all decoration style functions.

**JS**
@keybase/react-hackers 
1. Remove all custom code in the markdown parser pipeline for mentions of all kinds.
2. Change `ServiceDecoration` to put in `Mention` and `Channel` components when processing the relevant service decorations.

### DIFF
--- a/go/chat/utils/utils.go
+++ b/go/chat/utils/utils.go
@@ -1297,6 +1297,7 @@ func PresentDecoratedTextBody(ctx context.Context, g *globals.Context, msg chat1
 
 	// escape before applying xforms
 	body = EscapeForDecorate(ctx, body)
+	body = EscapeShrugs(ctx, body)
 
 	// Payment decorations
 	body = g.StellarSender.DecorateWithPayments(ctx, body, payments)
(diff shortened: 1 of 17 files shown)


shared/common-adapters/markdown/service-decoration.js
```diff
@@ -3,20 +3,25 @@ import React from 'react'
...
 import PaymentStatus from '../../chat/payments/status/container'
+import Mention from '../mention-container'
+import Channel from '../channel-container'
 
 export type Props = {
   json: string,
   onClick?: () => void,
   allowFontScaling?: ?boolean,
   message: Types.MessageText,
+  styles: Object,
```
chrisnojima: should be `Styles.StylesCrossPlatform`

mmaxim:
@chrisnojima PTAL, type more specific

go/chat/utils/utils.go
```diff
@@ -1781,8 +1784,91 @@ func DecorateBody(ctx context.Context, body string, offset, length int, decorati
 	if err != nil {
 		return res, 0
 	}
-	strDecoration := fmt.Sprintf("%s%s%s", decorateBegin, string(out), decorateEnd)
+	//b64out := string(out)
```
patrickxb: delete this?

mmaxim: It was useful to toggle on/off for tests, so I kind of wanted it to just be easily accessible in the future if I needed it.

patrickxb: sure, just checking.

PR #15621, opened 2019-01-21
Fix header centering.

adamjspooner:

This correctly centers (iOS) the title in the header. The fix I settled on was to absolutely position the text within the header and add appropriate padding based on the left and right actions.

I've also added lots of stories.


adamjspooner:
@zanderz, mind taking another look? I needed to fix a bug with `titleComponent`s.

zanderz:
Tried it again, still looks fine

adamjspooner:
Thanks!

PR #15592, opened 2019-01-17
Frequently used reacji bar component and stories

buoyad:
reacji-bar

Might change a bit depending on what the emoji picker ends up attaching to. r? @keybase/react-hackers 

- [ ] Storyshots before merge

PR #15622, opened 2019-01-21
Frequently used reacji bar integration

buoyad:
r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/messages/react-button/emoji-row/data.js
+++ b/shared/chat/conversation/messages/react-button/emoji-row/data.js
@@ -1,0 +1,11 @@
+// @flow
+import {memoize} from '../../../../../util/memoize'
+import {isMobile} from '../../../../../constants/platform'
+
+const getFrequently = memoize<void, void, void, void, {get: number => Array<string>}>(
+  () => require('emoji-mart').frequently
+)
+
+const getEmojis = () => (isMobile ? [] : getFrequently().get(2))
+
+export default getEmojis
--- a/shared/chat/conversation/messages/react-button/emoji-row/data.js
+++ b/shared/chat/conversation/messages/react-button/emoji-row/data.js
@@ -6,6 +6,11 @@ const getFrequently = memoize<void, void, void, void, {get: number => Array<stri
   () => require('emoji-mart').frequently
 )
 
-const getEmojis = () => (isMobile ? [] : getFrequently().get(2))
+const getEmojis = () =>
+  isMobile
+    ? []
+    : getFrequently()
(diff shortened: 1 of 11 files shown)


shared/chat/conversation/messages/react-button/emoji-row/data.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import {memoize} from '../../../../../util/memoize'
+import {isMobile} from '../../../../../constants/platform'
+
+const getFrequently = memoize<void, void, void, void, {get: number => Array<string>}>(
+  () => require('emoji-mart').frequently
```
zanderz: This means we rely on some facility of emoji-mart to say which ones are frequent>

buoyad:
@zanderz yep that is the designs, it works like that for me, are you up to date? [This change](https://github.com/keybase/client/pull/15622/commits/f22f6ad1db3f5a6bfc25663044757664dc123530) hides it when there are existing reacjis

shared/chat/conversation/messages/react-button/emoji-row/data.js
buoyad: Yeah it does. It uses `LocalStorage` which we can do when we transition away from emoji-mart

zanderz:
I tried `danny/DESKTOP-8657-integrate-2` within the last hour. It is NBD but it just looked like it is supposed to pop up in a different place depending on where you hover but it is not worth worrying about I think. SHIP IT!

buoyad:
Cool thanks @zanderz, I still need review on #15592 though 😄

shared/chat/conversation/messages/react-button/emoji-row/index.js
```diff
@@ -74,6 +96,7 @@ const styles = Styles.styleSheetCreate({
       borderRadius: 4,
       borderStyle: 'solid',
       borderWidth: 1,
+      height: 24,
```
chrisnojima: globalMargins.medium

shared/chat/conversation/messages/react-button/emoji-row/index.js
```diff
@@ -89,6 +112,13 @@ const styles = Styles.styleSheetCreate({
     paddingRight: Styles.globalMargins.xsmall,
     paddingTop: Styles.globalMargins.xtiny,
   },
+  pickerContainer: Styles.platformStyles({
+    isElectron: {
+      borderRadius: 4,
```
chrisnojima: Styles.borderRadius

PR #15623, opened 2019-01-21
removing findDOMNode

zanderz:
This leaves just one instance of findDOMnode, in FloatingBox, which seems like it could be quite complicated to change & test. Making a separate ticket for that.

Also, Danny has a change pending to make the emoji focus work again.


shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
```diff
@@ -24,9 +23,17 @@ type State = {
   height: number,
 }
 class ExplodingHeightRetainer extends React.PureComponent<Props, State> {
+  boxRef: React.ElementRef<any>
```
chrisnojima: these should start w/ underscore, why use any

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
```diff
@@ -24,9 +23,17 @@ type State = {
...
+  setHeight: () => void
   state = {animating: false, children: copyChildren(this.props.children), height: 17}
   timerID: SharedTimerID
 
+  constructor(props: Props) {
+    super(props)
+    // create a ref to store the textInput DOM element
+    this.boxRef = React.createRef()
```
chrisnojima: this can just be initialized above

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
```diff
@@ -23,15 +23,15 @@ type State = {
...
+  _boxRef = React.createRef()
   setHeight: () => void
   state = {animating: false, children: copyChildren(this.props.children), height: 17}
   timerID: SharedTimerID
 
   constructor(props: Props) {
     super(props)
     // create a ref to store the textInput DOM element
-    this.boxRef = React.createRef()
+    this._boxRef
```
chrisnojima: entirely remove constructor

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
zanderz: ok

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
zanderz: ok

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
zanderz: ok

shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.desktop.js
```diff
@@ -24,6 +23,8 @@ type State = {
   height: number,
 }
 class ExplodingHeightRetainer extends React.PureComponent<Props, State> {
+  _boxRef = React.createRef()
+  setHeight: () => void
```
chrisnojima: why is this here?

zanderz: I don't recognize that, no idea. Will nuke.

PR #15625, opened 2019-01-21
folder_block_manager: fix and simplify quota reclamation

strib:
Previously we would give up doing QR if there were more than 100 recent revisions that were less than 2 weeks old.  This is clearly the wrong thing to do.  I think we did this previously due to historical limitations of old data formats (i.e., not having the `LastGCRevision` field, and not being able to tell when an MD revision was made).

These limitations are ancient now, and it's notably preventing QR on some folders.  So fix it by counting MDs forward from the `LastGCRevision`, instead of backward (which could result in an arbitrary amount of work).

Issue: KBFS-3597

### DIFF
--- a/go/kbfs/libkbfs/folder_block_manager.go
+++ b/go/kbfs/libkbfs/folder_block_manager.go
@@ -126,9 +126,9 @@ type folderBlockManager struct {
 	// Remembers what happened last time during quota reclamation.
 	lastQRLock          sync.Mutex
 	lastQRHeadRev       kbfsmd.Revision
-	lastQROldEnoughRev  kbfsmd.Revision
 	wasLastQRComplete   bool
 	lastReclamationTime time.Time
+	lastReclaimedRev    kbfsmd.Revision
 }
 
 func newFolderBlockManager(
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/folder_block_manager.go
```diff
@@ -1175,13 +1112,12 @@ func (fbm *folderBlockManager) doReclamation(timer *time.Timer) (err error) {
...
-	mostRecentOldEnoughRev, lastGCRev, err :=
-		fbm.getMostRecentOldEnoughAndGCRevisions(ctx, head.ReadOnly())
+	lastGCRev, err := fbm.getMostRecentGCRevision(ctx, head.ReadOnly())
 	if err != nil {
 		return err
 	}
-	if mostRecentOldEnoughRev == kbfsmd.RevisionUninitialized ||
-		mostRecentOldEnoughRev <= lastGCRev {
+	if head.Revision() <= lastGCRev {
+		fbm.log.CDebugf(ctx, "QR NOT NEEDED %d", lastGCRev)
```
jzila: This probably doesn't need to be all caps.

go/kbfs/libkbfs/folder_block_manager.go
```diff
@@ -1213,15 +1150,17 @@ func (fbm *folderBlockManager) doReclamation(timer *time.Timer) (err error) {
 
 		// Add a new gcOp to show other clients that they don't need
 		// to explore this range again.
-		return fbm.finalizeReclamation(ctx, nil, nil, latestRev)
+		return fbm.finalizeReclamation(ctx, nil, nil, lastRev)
+	} else if shortened {
+		complete = false
```
jzila: whoa was this a bug?

go/kbfs/libkbfs/folder_block_manager.go
strib: Oops, I meant to take this debugging line out entirely.  Will do, thanks!

go/kbfs/libkbfs/folder_block_manager.go
strib: It was, though it probably didn't have much practical effect besides slowing down QR a bit.

PR #15627, opened 2019-01-21
Perform periodical background merkle root skip audits #2

pzduniak:
Story CORE-8615, follow up to #15596, now with CI!

Remaining notes from the previous pullreq:

1. We're not sure if the NodeHash.Eq method should be an interface method, the auditor should stay in engine or if HashEq should be exported:
https://github.com/keybase/client/pull/15596#discussion_r249565175

2. We were unsure about some code paths, added nil checks just in case.


pzduniak:
I don't like how repetitive the retry test is, but I don't think it's worth to optimize the complexity of this thing when it's just 3 API calls.

maxtaco:
@pzduniak can you tag me and @zapu on the review?

maxtaco:
@pzduniak also can you rebase on top of current master, since there are some CI bugs that I think have been fixed. thanks!

pzduniak:
Rebased on recent master.

maxtaco:
Looking good! @zapu do you have time today for a final review?  Thanks!

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,251 @@
...
+		if !startHash.Eq(currentHash) {
+			// Warn the user about the possibility of the server tampering with the roots.
+			return libkb.NewClientMerkleSkipHashMismatchError(
+				fmt.Sprintf("Invalid skip hash from %d to %d", currentSeqno, startSeqno),
+			)
+		}
+
+		// We're doing this exponentially to make use of the skips.
+		currentSeqno += keybase1.Seqno(step)
+		step *= 2
```
zapu: if this algo is correct to hit the relevant skips, then LGTM.

sorry for leading you to the rabbit hole with the tests.

go/engine/merkle_audit.go
```diff
@@ -0,0 +1,251 @@
...
+var (
+	errAuditOffline    = errors.New("Merkle audit failed to run due to the lack of connectivity.")
+	errAuditNoLastRoot = errors.New("Merkle audit failed to run due to not being able to get the last root.")
+)
+
+var MerkleAuditSettings = BackgroundTaskSettings{
+	Start:        5 * time.Minute,
+	StartStagger: 1 * time.Hour,
+	Interval:     6 * time.Hour,
+	Limit:        1 * time.Minute,
```
joshblum: should set `DisableInBotLite: true` once https://github.com/keybase/client/pull/15672 is merged (or i'll update that PR if this is merged first)

go/service/main.go
```diff
@@ -547,6 +548,25 @@ func (d *Service) runHomePoller(ctx context.Context) {
 	return
 }
 
+func (d *Service) runMerkleAudit(ctx context.Context) {
+	if !libkb.IsMobilePlatform() {
```
zapu: looks like a bug here, it should bail out when `libkb.IsMobilePlatform()`

pzduniak: Oops, fixed.

zapu: kk thanks, trying this again

zapu:
Ok this is what I think we need to do to improve logging on both server and client:
```
diff --git a/go/engine/merkle_audit.go b/go/engine/merkle_audit.go
index 55dd470265..a2880eedef 100644
--- a/go/engine/merkle_audit.go
+++ b/go/engine/merkle_audit.go
@@ -189,7 +189,10 @@ func performMerkleAudit(m libkb.MetaContext, startSeqno keybase1.Seqno) error {
 	return nil
 }
 
-func MerkleAuditRound(m libkb.MetaContext) error {
+func MerkleAuditRound(m libkb.MetaContext) (err error) {
+	m = m.WithLogTag("MAUDT")
+	defer m.CTraceTimed("MerkleAuditRound", func() error { return err })()
+
 	// Look up any previously requested retries
 	startSeqno, err := lookupMerkleAuditRetryFromState(m)
 	if err != nil {

```

pzduniak:
@zapu wouldn't the err in the function declaration get instantly shadowed? I'm not familiar with this style of code so I'm not entirely sure, but IIRC the return declaration in the function acted exactly like a `var` block in the beginning of the function.

maxtaco:
Nope! If you have this:

```go
var e error
i, e := somefunc()
```

it will only create `i` and then reuse `e`

pzduniak:
Then it looks like `go vet` with `-shadow` is overly cautious! Makes sense.

PR #15628, opened 2019-01-22
attach-to-chat 2: connect

songgao:
Not triggering any RPC yet, but components are connected. Navigation, filtering, etc. work. I'll have a follow-up PR for actually making those RPCs and likely another to stream from SimpleFS to chat.


songgao:
bump!

shared/chat/conversation-list/conversation-list-container.js
```diff
@@ -1,15 +1,133 @@
...
+const boolToNumber = (b: boolean): number => (b ? 1 : 0)
+
+const staleToNumber = (convTime: number, staleCutoff: number) => (convTime < staleCutoff ? 1 : 0)
+
+const getAWeekAgo = () => {
+  let t = new Date()
+  return t.setDate(t.getDate() - 7) // works fine for cross-boundary; returns a number
+}
+
+const getSortedConversationIDKeys = memoize(
```
chrisnojima: really all this should move to core, and will soon

shared/reducers/fs.js
```diff
@@ -292,6 +297,13 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
     case FsGen.setSendLinkToChatChannels:
       // $FlowIssue
       return state.setIn(['sendLinkToChat', 'channels'], action.payload.channels)
+    case FsGen.setSendAttachmentToChatConvID:
+      // $FlowIssue
```
chrisnojima: why flow issue

songgao: I think `setIn` with flow is just broken. I'll see if I can get around it with other mutaters.

shared/chat/conversation-list/conversation-list-container.js
songgao: Cool; keep this for now? Also note that this is different from inbox sorting.

shared/actions/fs/index.js
```diff
@@ -899,6 +911,10 @@ function* fsSaga(): Saga.SagaGenerator<any, any> {
   yield* Saga.chainAction<FsGen.ShowMoveOrCopyPayload>(FsGen.showMoveOrCopy, showMoveOrCopy)
   yield* Saga.chainAction<FsGen.CancelMoveOrCopyPayload>(FsGen.cancelMoveOrCopy, cancelMoveOrCopy)
   yield* Saga.chainGenerator<FsGen.ShowSendLinkToChatPayload>(FsGen.showSendLinkToChat, showSendLinkToChat)
+  yield* Saga.chainGenerator<FsGen.ShowSendLinkToChatPayload>(
```
chrisnojima: should just be chainAction

shared/chat/conversation-list/conversation-list.js
```diff
@@ -81,8 +83,19 @@ const ConversationList = (props: Props) => {
...
-      {!!props.filter && <ConversationFilterInput style={styles.filter} {...props.filter} />}
+    <Kb.Box2 direction="vertical" fullWidth={true} fullHeight={true}>
+      {!!props.filter && (
+        <ConversationFilterInput
+          style={styles.filter}
+          isLoading={props.filter.isLoading}
+          filter={props.filter.filter}
+          onSetFilter={props.filter.onSetFilter}
+          filterFocusCount={0}
+          onSelectDown={() => {} /* TODO: make these work */}
```
chrisnojima: where is this implemented

songgao:
I'm gonna squash all these commit into one so conflicts don't kill me

songgao:
ready for another look!

songgao:
⪮

songgao:
Actually making attachment happen was simpler than I thought, and since this hasn't been approved yet I just squeezed it into this branch instead of creating a new one.

songgao:
rebased and bump!

songgao:
bump!

shared/chat/conversation-list/choose-conversation.js
```diff
@@ -22,27 +23,31 @@ class ChooseConversation extends React.Component<Props & Kb.OverlayParentProps,
...
-    this.setState({expanded: false})
-  }
+  _onDone = () => this.setState({expanded: false})
   render() {
     return (
       <>
         <Kb.DropdownButton
-          disabled={this.props.disabled}
-          selected={<Kb.Text type="BodySemibold">{this.props.dropdownButtonDefaultText}</Kb.Text>}
+          disabled={false}
```
chrisnojima: 🔥

shared/chat/conversation-list/choose-conversation.js
```diff
@@ -22,27 +23,31 @@ class ChooseConversation extends React.Component<Props & Kb.OverlayParentProps,
...
           setAttachmentRef={this.props.setAttachmentRef}
           toggleOpen={this._toggleOpen}
           style={Styles.collapseStyles([styles.dropdownButton, this.props.dropdownButtonStyle])}
         />
         <Kb.Overlay
           attachTo={this.props.getAttachmentRef}
           onHidden={this._toggleOpen}
-          position={this.props.position || 'center center'}
-          style={Styles.collapseStyles([styles.overlay, this.props.overlayStyle])}
+          position={'center center'}
```
chrisnojima: dont need curlies

shared/chat/conversation-list/conversation-list-container.js
```diff
@@ -1,17 +1,189 @@
...
+    return metaMap
+      .valueSeq()
+      .toArray()
+      .sort((a, b) => {
+        // leveled order rules:
+        // 1. unmuted before muted
+        // 2. active conversations before inactive (has activity in the past week)
+        // 3. notification type: onAnyActivity before onWhenAtMentioned, before never
+        // 4. activity timestamp being the last tie breaker
+        const mutedBased = boolToNumber(a.isMuted) - boolToNumber(b.isMuted)
```
chrisnojima: if the goal is to see if they're both the same truthy vs falsey you can just be
```
const muteBased = !a.isMuted !== !b.isMuted
```

shared/chat/conversation-list/conversation-list.js
```diff
@@ -81,8 +87,23 @@ const ConversationList = (props: Props) => {
...
+      {!!props.filter && (
+        <ConversationFilterInput
+          onBack={props.onBack}
+          style={styles.filter}
+          isLoading={props.filter.isLoading}
+          filter={props.filter.filter}
+          onSetFilter={props.filter.onSetFilter}
+          filterFocusCount={0}
+          noShortcut={true}
+          onBlur={() => {}}
```
chrisnojima: is it safe to just have these do nothing now? why?

shared/chat/conversation-list/conversation-list-container.js
songgao: I still need a number here to return though, so would still need to use boolToNumber.

shared/chat/conversation-list/choose-conversation.js
songgao: You mean we should keep this prop? This is only used in the attachment popup. I think it's always enabled.

chrisnojima: no i mean you can just delete it, it defaults to false right

shared/chat/conversation-list/conversation-list.js
songgao: I actually just didn't know what they were for, and this isn't used in chat at all, just in the attachment popup I'm building now, so I left them like that. But I guess I'll just figure out and do them right here.

shared/chat/conversation-list/choose-conversation.js
songgao: Ah that's right!

shared/chat/conversation-list/conversation-list.js
songgao: OK I don't think we need these for choose-conversation, but I added focusOnMount for the popup.

songgao:
bump!

songgao:
⪮

songgao:
⪮

songgao:
woohooo thanks Nojima

PR #15629, opened 2019-01-22
Add dice.

adamjspooner:
@keybase/react-hackers, this is ready for review. The dice aren't in use and are awaiting feedback from the rest of the design team. There are only a couple of stories for now.


shared/common-adapters/dice.js
```diff
@@ -0,0 +1,29 @@
...
+  reactDice: any
+  componentDidMount() {
+    this.reactDice.rollAll()
+  }
+  render() {
+    return (
+      <ReactDice
+        disableIndividual={true}
+        dotColor="white"
+        faceColor="linear-gradient(-45deg, #f0a24c, #f6cc60)"
```
adamjspooner: These colors are hardcoded because they're not final.

shared/common-adapters/dice.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+import * as React from 'react'
+import ReactDice from 'react-dice-complete'
+import 'react-dice-complete/dist/react-dice-complete.css'
```
MarcoPolo: Does this work on native too?

shared/common-adapters/dice.js
```diff
@@ -0,0 +1,29 @@
...
+
+type Props = {
+  onRollDone: number => void,
+  values?: Array<number>,
+}
+
+export default class Dice extends React.Component<Props> {
+  reactDice: any
+  componentDidMount() {
+    this.reactDice.rollAll()
```
MarcoPolo: I would just call this `_ref` or `_reactDiceRef`. Since that's more common for us and it'll be understood right away what this is.

shared/common-adapters/dice.js
adamjspooner: Nope. 😞

PR #15631, opened 2019-01-22
Use new header on files.

adamjspooner:

cc @jzila @songgao


shared/fs/header/header.native.js
```diff
@@ -1,68 +1,27 @@
...
-            style={collapseStyles([styles.stylesIcons])}
-            color={globalColors.black_50}
-            fontSize={22}
-            onClick={onChat}
-          />
-        </Box>
-      )}
-    </Box>
-  </Box>
+  <Kb.HeaderHocHeader
```
songgao: Is this called HeaderHocHeader because it's usually used in HeaderHoc, or does it have other implications?

PR #15634, opened 2019-01-22
Tweak when emojirow is visible with existing reactions

buoyad:
Currently it's always there if there are existing reactions, this changes it so `EmojiRow` pops in while hovering the message. Also removes some unused login in `ReactionsRow` related to how this used to work. r? @keybase/react-hackers 
emoji-row-hovering


buoyad:
cc @zanderz I think you mentioned this in your comment on the original PR

PR #15636, opened 2019-01-22
Show error message & 'try again' if accepting the stellar disclaimer fails

buoyad:
You can test this by adding 
```
time.Sleep(3 * time.Second)
return errors.New("no")
``` 
right around [here](https://github.com/keybase/client/blob/master/go/stellar/stellarsvc/frontend.go#L241). r? @keybase/react-hackers

### DIFF
--- a/shared/common-adapters/button-bar.js
+++ b/shared/common-adapters/button-bar.js
@@ -73,11 +73,13 @@ class ButtonBar extends React.PureComponent<Props> {
 // Note explicitly not using globalMargins here. We don't necessarily want this spacing to change ever
 const BigSpacer = () => <Box style={bigSpacerStyle} />
 const bigSpacerStyle = {
+  flexShrink: 0,
   height: 8,
   width: 8,
 }
 const SmallSpacer = () => <Box style={smallSpacerStyle} />
 const smallSpacerStyle = {
+  flexShrink: 0,
   height: isMobile ? 8 : 4,
   width: isMobile ? 8 : 4,
 }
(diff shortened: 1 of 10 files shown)


shared/common-adapters/button-bar.js
```diff
@@ -73,11 +73,13 @@ class ButtonBar extends React.PureComponent<Props> {
 // Note explicitly not using globalMargins here. We don't necessarily want this spacing to change ever
 const BigSpacer = () => <Box style={bigSpacerStyle} />
 const bigSpacerStyle = {
+  flexShrink: 0,
```
buoyad: This causes the storyshot changes, I checked all the screens and they look fine

shared/reducers/wallets.js
```diff
@@ -336,11 +336,14 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
...
     case WalletsGen.walletDisclaimerReceived:
       return state.merge({acceptedDisclaimer: action.payload.accepted})
-    // Saga only actions
     case WalletsGen.acceptDisclaimer:
       return state.merge({
         acceptingDisclaimerDelay: true,
       })
+    case WalletsGen.resetAcceptingDisclaimer:
+      return state.merge({
+        acceptingDisclaimerDelay: false,
```
MarcoPolo: Should you also reset `acceptedDisclaimer`?

buoyad: Hm I don't think so, `acceptedDisclaimer` can only be true if the service says so, and then `wallets-and-details/container` will redirect to the main wallets tab

PR #15639, opened 2019-01-22
NO-EMAIL fixes for frontend

zapu:


shared/settings/landing/container.js
```diff
@@ -12,11 +12,17 @@ const mapStateToProps = state => {
   const {emails} = state.settings.email
   const {rememberPassphrase} = state.settings.passphrase
   let accountProps
-  if (emails.length > 0) {
+  if (emails) {
```
buoyad: `if (emails?.length)` and remove the `if` on L17

shared/settings/email/container.js
```diff
@@ -10,7 +10,7 @@ const mapStateToProps = state => {
   const {emails, error} = state.settings.email
   let email = ''
   let isVerified = false
-  if (emails.length > 0) {
+  if (emails && emails.length > 0) {
```
buoyad: `if (emails?.length)`

shared/settings/landing/index.desktop.js
```diff
@@ -278,6 +278,26 @@ function AccountEmail({
...
+      style={{
+        ...globalStyles.flexBoxRow,
+        alignItems: 'center',
+        justifyContent: 'space-between',
+        minHeight: ROW_HEIGHT,
+      }}
+    >
+      <Text type="Body">
+        No e-mail set. You should{' '}
+        <Text type="Body" style={{color: globalColors.blue}} link={true} onClick={onChangeEmail}>
```
buoyad: `type="BodyPrimaryLink"` and no need for `style` or `link` props. `Text` doesn't use `link` anymore (I don't think it has for 1-2yrs 😮), we should update the rest of this file at some point.

shared/settings/landing/container.js
zapu: I don't think that's good though, because we still want to set 
```
      hasRandomPW: state.settings.passphrase.randomPW,
      onChangeEmail: () => logger.debug('todo'),
      onChangePassphrase: () => logger.debug('todo'),
      rememberPassphrase,
```
even if user has zero emails.

buoyad: Then should it be outside this `if`?

PR #15642, opened 2019-01-23
Reloadable components for teams & single team screens

MarcoPolo:
@keybase/react-hackers

### DIFF
--- a/shared/actions/teams.js
+++ b/shared/actions/teams.js
@@ -5,7 +5,6 @@ import logger from '../logger'
 import {map} from 'lodash-es'
 import * as I from 'immutable'
 import * as SearchGen from './search-gen'
-import * as WaitingGen from './waiting-gen'
 import * as TeamsGen from './teams-gen'
 import * as ProfileGen from './profile-gen'
 import * as Types from '../constants/types/teams'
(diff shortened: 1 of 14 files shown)


shared/actions/teams.js
```diff
@@ -482,124 +483,129 @@ function* getDetails(_, action) {
...
-  }
 
-  // Get requests to join
-  let requests
-  const state = yield* Saga.selectState()
-  if (Constants.getCanPerform(state, teamname).manageMembers) {
-    // TODO (DESKTOP-6478) move this somewhere else
-    requests = yield* Saga.callPromise(
-      RPCTypes.teamsTeamListRequestsRpcPromise,
+  try {
```
MarcoPolo: This is now wrapped in a try block to avoid the black box. The error is still visible in the show more option of the reloadable component.

MarcoPolo:
Easier to parse if you use the `w=1` flag https://github.com/keybase/client/pull/15642/files?w=1

shared/actions/teams.js
```diff
@@ -809,10 +820,17 @@ function* getTeams(state) {
...
+          error: err,
+          key: Constants.teamsLoadedWaitingKey,
+        })
+      )
+
       throw err
     }
   } finally {
-    yield Saga.put(TeamsGen.createSetLoaded({loaded: true}))
+    yield Saga.put(WaitingGen.createDecrementWaiting({key: Constants.teamsLoadedWaitingKey}))
```
buoyad: if this and the `decrementWaiting` in the else block above both happen, do we end up in a bad state?

chrisnojima: we shouldn't be doing this explicitly, the rpc will already decrement right?

chrisnojima: we basically never need manual dispatches of the waiting keys

MarcoPolo: This was to copy the previous behavior of having a loaded value, but now that you mention it. I think we can remove that behavior completely and just rely on the engine

PR #15644, opened 2019-01-23
add slash command interface to chat

mmaxim:
Slash commands for chat, the following are implemented:
1. /hide and /unhide
2. /msg
3. /join and /leave
4. /headline
5. /shrug
6. /me

Works by intercepting message text sends and trying to match against the set of rules for a conversation. This is built with integrating bot advertised commands in mind. 

cc @malgorithms @adamjspooner if you have any comments about this plays out.


adamjspooner:
This is awesome! I tried it out, and it works great. Can we use the channel/username completion component for discoverability (and autocompletion)?

joshblum:
would be cool to hook into @buoyad's emoji HUD

mmaxim:
Coming soon in my next PR

adamjspooner:
🎉

adamjspooner:
Do we have a story for allowing users to add their own? Or is that part of the ongoing bot work?

mmaxim:
I have thought about a system where bots can advertise their own commands into a conversation, but I haven't though much about how users could create commands themselves.

mmaxim:
That being said, it probably wouldn't be hard to do, but we would need some kind of settings design for it.

adamjspooner:
For sure. 👍

go/chat/helper.go
```diff
@@ -766,6 +769,49 @@ func JoinConversation(ctx context.Context, g *globals.Context, debugger utils.De
...
+	// Fetch the TLF ID from specified name
+	nameInfo, err := CreateNameInfoSource(ctx, g, chat1.ConversationMembersType_TEAM).LookupID(ctx,
+		tlfName, vis == keybase1.TLFVisibility_PUBLIC)
+	if err != nil {
+		debugger.Debug(ctx, "JoinConversationByName: failed to get TLFID from name: %s", err.Error())
+		return err
+	}
+
+	// List all the conversations on the team
+	convs, err := g.TeamChannelSource.GetChannelsFull(ctx, uid, nameInfo.ID, topicType)
```
joshblum: why not `TeamChannelSource.GetChannelsTopicName`?

go/chat/inboxsource.go
```diff
@@ -174,6 +174,45 @@ func (b *baseInboxSource) Localize(ctx context.Context, uid gregor1.UID, convs [
...
+	convID chat1.ConversationID, status chat1.ConversationStatus) (err error) {
+	defer b.Trace(ctx, func() error { return err }, "RemoteSetConversationStatus")()
+	if _, err = b.getChatInterface().SetConversationStatus(ctx, chat1.SetConversationStatusArg{
+		ConversationID: convID,
+		Status:         status,
+	}); err != nil {
+		return err
+	}
+	// Send word to API server about the report
+	if status == chat1.ConversationStatus_REPORTED {
```
joshblum: why not short circuit on != so the main code is less indented?

go/chat/server.go
```diff
@@ -1188,8 +1151,16 @@ func (h *Server) PostLocal(ctx context.Context, arg chat1.PostLocalArg) (res cha
 		return res, fmt.Errorf("no TLF name specified")
 	}
 
+	// Check for any slash command hits for an execute
+	if handled, err := h.runSlashCommands(ctx, uid, arg.ConversationID, arg.Msg.ClientHeader.TlfName,
+		arg.Msg.MessageBody); handled {
+		h.Debug(ctx, "PostLocal: handled slash command with error: %s", err)
```
joshblum: only log if error?

go/chat/server.go
```diff
@@ -1471,6 +1442,25 @@ func (h *Server) GenerateOutboxID(ctx context.Context) (res chat1.OutboxID, err
 	return storage.NewOutboxID()
 }
 
+func (h *Server) runSlashCommands(ctx context.Context, uid gregor1.UID, convID chat1.ConversationID,
+	tlfName string, body chat1.MessageBody) (bool, error) {
+	if !body.IsType(chat1.MessageType_TEXT) {
+		return false, nil
+	}
+	text := body.Text().Body
```
joshblum: short circuit if it doesn't start with `/` so we don't have to try to match everything

go/chat/server.go
```diff
@@ -1471,6 +1442,25 @@ func (h *Server) GenerateOutboxID(ctx context.Context) (res chat1.OutboxID, err
 	return storage.NewOutboxID()
 }
 
+func (h *Server) runSlashCommands(ctx context.Context, uid gregor1.UID, convID chat1.ConversationID,
```
joshblum: why not make this as part of the cmdSource? just to keep the logic in a single place

go/chat/commands/base.go
```diff
@@ -0,0 +1,90 @@
...
+		toks := strings.Split(name, "#")
+		return find(chat1.ConversationMembersType_TEAM, toks[0], &toks[1])
+	}
+	if res, err = find(chat1.ConversationMembersType_IMPTEAMNATIVE, name, nil); err != nil {
+		return find(chat1.ConversationMembersType_TEAM, name, nil)
+	}
+	return res, nil
+}
+
+func (b *baseCommand) commandAndMessage(text string) (cmd string, msg string, err error) {
```
joshblum: i think this should consume the first token (unused everywhere) and return the tokenized list for use by the command directly

go/chat/commands/base.go
```diff
@@ -0,0 +1,90 @@
...
+	toks := strings.Split(text, " ")
+	if len(toks) < 2 {
+		return "", "", ErrInvalidArguments
+	}
+	return toks[0], strings.Join(toks[1:], " "), nil
+}
+
+func (b *baseCommand) Match(ctx context.Context, text string) bool {
+	cands := append(b.aliases, b.name)
```
joshblum: short circuit on not starting with `/`

go/chat/commands/dm.go
```diff
@@ -0,0 +1,39 @@
...
+
+func NewDM(g *globals.Context) *DM {
+	return &DM{
+		baseCommand: newBaseCommand(g, "msg", "<conversation> <message>",
+			"Send a message in the specified conversation", "dm"),
+	}
+}
+
+func (d *DM) Execute(ctx context.Context, uid gregor1.UID, convID chat1.ConversationID,
+	tlfName, text string) (err error) {
```
joshblum: why not pass `tokens []string` instead of `text`, cuts the boilerplate for each command to validating the tokens directly

joshblum: if not at least wrap the text splitting to override `commandAndMessage` to keep the `Execute` more streamlined

go/chat/commands/base.go
```diff
@@ -0,0 +1,90 @@
...
+		if len(convs) == 0 {
+			return res, errors.New("conversation not found")
+		}
+		return convs[0], nil
+	}
+	if strings.Contains(name, "#") {
+		toks := strings.Split(name, "#")
+		return find(chat1.ConversationMembersType_TEAM, toks[0], &toks[1])
+	}
+	if res, err = find(chat1.ConversationMembersType_IMPTEAMNATIVE, name, nil); err != nil {
```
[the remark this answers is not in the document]
mmaxim: `FindConversations` falls back to that if it can't find anything.

go/chat/helper.go
mmaxim: ¯\_(ツ)_/¯, I just copy/pasted this.

go/chat/inboxsource.go
mmaxim: Also just copy/pasted this.

go/chat/server.go
mmaxim: I like to see it all the time, it isnt going to be spamming anything.

go/chat/server.go
mmaxim: Sure.

go/chat/helper.go
mmaxim: I think this one is probably fine, I'm not sure we gain much here and would rather not change anything too orthogonal to this PR.

go/chat/commands/dm.go
mmaxim: How important is this to you? I feel like calling `string.Split` in here isn't so bad.

joshblum: just seems messy and wrapping in `commandAndMessage` (maybe called `tokenize`) or passing `tokens` would make it cleaner

go/chat/commands/source.go
```diff
@@ -47,6 +47,27 @@ func (s *Source) makeBuiltin() {
...
 	return []types.ConversationCommandGroup{s.builtin}, nil
 }
+
+func (s *Source) AttemptCommand(ctx context.Context, uid gregor1.UID, convID chat1.ConversationID,
+	tlfName string, body chat1.MessageBody) (handled bool, err error) {
+	defer s.Trace(ctx, func() error { return err }, "AttemptCommand")()
+	if !body.IsType(chat1.MessageType_TEXT) {
+		return false, nil
+	}
+	text := body.Text().Body
```
joshblum: short circuit here?

go/chat/commands/base.go
```diff
@@ -0,0 +1,101 @@
...
+func (b *baseCommand) tokenize(text string, minArgs int) (toks []string, err error) {
+	toks = strings.Split(text, " ")
+	if len(toks) < minArgs {
+		return toks, ErrInvalidArguments
+	}
+	return toks, nil
+}
+
+func (b *baseCommand) commandAndMessage(text string) (cmd string, msg string, err error) {
+	toks := strings.Split(text, " ")
```
joshblum: use `tokenize`

PR #15646, opened 2019-01-23
fix listing non-existent public TLFs

songgao:
As discussed in #kbfs we should show an empty for public TLFs that don't exist, like through FUSE. However, it turns out this was broken through FUSE as well (Input/Output error). After this change, GUI shows an empty folder for non-existent TLFs too.

`libfuse` tests seem to be using traditional keying, so a test would probably need to be in `kbfsdocker`. Let me know if you all think there should be test for this.

### DIFF
--- a/go/kbfs/libkbfs/kbfs_ops.go
+++ b/go/kbfs/libkbfs/kbfs_ops.go
@@ -440,9 +440,15 @@ func (fs *KBFSOpsStandard) getOpsByHandle(ctx context.Context,
 	return ops
 }
 
+type errTryingToResetTlfIDForNonTeamTlf struct{}
+
+func (errTryingToResetTlfIDForNonTeamTlf) Error() string {
+	return "Can't create TLF ID for non-team-backed handle"
+}
+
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/kbfs_ops.go
```diff
@@ -708,7 +714,15 @@ func (fs *KBFSOpsStandard) getMaybeCreateRootNode(
 	}
 
 	err = fs.createAndStoreTlfIDIfNeeded(ctx, h)
-	if err != nil {
+	switch {
+	case err == nil:
+	case !create && err == errTryingToResetTlfIDForNonTeamTlf{}:
+		// We don't have TLF ID, and it's not backed by an impl team. So this
+		// must be a TLF that doesn't exist and we can't or aren't create. Just
```
jzila: we can't or aren't?

songgao:
I think this is fixed in 18a7cea72ee5d6682871e1485ffa67b9cfc649ac

PR #15649, opened 2019-01-23
Fix seqno state on error, plus check amounts

patrickxb:
There was a bug that when SubmitPayment or SubmitRelayPayment
returned an error, the pending tx wasn't removed from WalletState.
This causes seqno issues after that...

Also, this adds some more checks for low payment amounts up front.
Since we are predicting seqnos so much now, it doesn't make sense
to burn one on a low amount payment that we can detect easily.

### DIFF
--- a/go/stellar/stellar.go
+++ b/go/stellar/stellar.go
@@ -590,6 +590,15 @@ func sendPayment(mctx libkb.MetaContext, walletState *WalletState, sendArg SendP
 		post.To = &recipient.User.UV
 	}
 
+	// check if recipient account exists
+	funded, err := isAccountFunded(mctx.Ctx(), walletState, stellar1.AccountID(recipient.AccountID.String()))
+	if err != nil {
+		return res, fmt.Errorf("error checking destination account balance: %v", err)
+	}
+	if !funded && isAmountLessThanMin(sendArg.Amount, minAmountCreateAccountXLM) {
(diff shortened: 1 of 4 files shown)


go/stellar/stellar.go
```diff
@@ -648,6 +650,9 @@ func sendPayment(mctx libkb.MetaContext, walletState *WalletState, sendArg SendP
 	// submit the transaction
 	rres, err := walletState.SubmitPayment(mctx.Ctx(), post)
 	if err != nil {
+		if rerr := walletState.RemovePendingTx(mctx.Ctx(), senderAccountID, stellar1.TransactionID(txID)); rerr != nil {
+			mctx.CDebugf("error calling RemovePendingTx: %s", rerr)
+		}
```
patrickxb: this is the fix 1 of 2.

go/stellar/stellar.go
```diff
@@ -1058,6 +1077,9 @@ func sendRelayPayment(mctx libkb.MetaContext, walletState *WalletState,
 	}
 	rres, err := walletState.SubmitRelayPayment(mctx.Ctx(), post)
 	if err != nil {
+		if rerr := walletState.RemovePendingTx(mctx.Ctx(), accountID, stellar1.TransactionID(relay.FundTx.TxHash)); rerr != nil {
+			mctx.CDebugf("error calling RemovePendingTx: %s", rerr)
+		}
```
patrickxb: this is the fix 2 of 2.

PR #15655, opened 2019-01-23
Retention picker cleanup

buoyad:
- Make `Props` exact
- `import Kb` and `Styles`
- use `OverlayParentHOC`

r? @keybase/react-hackers


buoyad:
@keybase/react-hackers bump

PR #15657, opened 2019-01-23
Update reacji row styles.

adamjspooner:

cc @buoyad


shared/chat/conversation/messages/wrapper/index.js
```diff
@@ -416,10 +416,7 @@ class _WrapperMessage extends React.Component<Props & Kb.OverlayParentProps, Sta
                     conversationIDKey={this.props.conversationIDKey}
                     onShowingEmojiPicker={this._setShowingPicker}
                     ordinal={message.ordinal}
-                    style={Styles.collapseStyles([
-                      styles.emojiRow,
-                      this.props.isLastInThread ? styles.emojiRowLast : null,
```
buoyad: Now this prop `isLastInThread` is unused, can we remove it?

buoyad: Whoops too late

adamjspooner: Sure! I didn't look to see if it was used elsewhere. I'll open another PR removing it.

buoyad: Thanks!

PR #15662, opened 2019-01-23
Test for sbs relay

mlsteele:
This test will be my guiding light for SBS relay payments. It passes when the server is on c30fb128a but stellard is on b8741bf~. And fails when stellard is on c30fb128a.

- [ ] use CI info like `BRANCH_NAME == "master"` to dodge CI


mlsteele:
Since this test doesn't run non non-master CI this PR should pass CI. But should only be merged once https://github.com/keybase/keybase/pull/3378 is merged.

PR #15668, opened 2019-01-24
Add more waiting info to teams page

MarcoPolo:
@keybase/react-hackers 

I basically went through and add a waiting key to every rpc in actions/teams. I played around with the teams tab and fixed a couple bugs I found. Seems like there's a progress indicator now for anything that needs does a network trip.

### DIFF
--- a/shared/actions/teams.js
+++ b/shared/actions/teams.js
@@ -79,9 +79,14 @@ function* joinTeam(_, action) {
     Saga.put(TeamsGen.createSetTeamJoinSuccess({success: false, teamname: ''})),
   ])
   try {
-    const result = yield* Saga.callPromise(RPCTypes.teamsTeamAcceptInviteOrRequestAccessRpcPromise, {
-      tokenOrName: teamname,
-    })
+    // TODO add waiting state here
+    const result = yield* Saga.callPromise(
+      RPCTypes.teamsTeamAcceptInviteOrRequestAccessRpcPromise,
+      {
+        tokenOrName: teamname,
+      },
+      Constants.teamWaitingKey(teamname)
+    )
 
     // Success
     yield Saga.put(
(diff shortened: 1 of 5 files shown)


shared/actions/teams.js
```diff
@@ -680,11 +694,7 @@ function* getTeamPublicity(_, action) {
     let tarsDisabled = false
     // can throw if you're not an admin
     try {
-      tarsDisabled = yield* Saga.callPromise(
-        RPCTypes.teamsGetTarsDisabledRpcPromise,
-        {name: teamname},
-        Constants.teamWaitingKey(teamname)
-      )
+      tarsDisabled = yield* Saga.callPromise(RPCTypes.teamsGetTarsDisabledRpcPromise, {name: teamname})
```
chrisnojima: why remove the waiting key here?

shared/actions/teams.js
```diff
@@ -79,9 +79,13 @@ function* joinTeam(_, action) {
     Saga.put(TeamsGen.createSetTeamJoinSuccess({success: false, teamname: ''})),
   ])
   try {
-    const result = yield* Saga.callPromise(RPCTypes.teamsTeamAcceptInviteOrRequestAccessRpcPromise, {
-      tokenOrName: teamname,
-    })
+    const result = yield* Saga.callPromise(
+      RPCTypes.teamsTeamAcceptInviteOrRequestAccessRpcPromise,
+      {
```
chrisnojima: i've been making objects w/ one key just go on a single line, easier to read i think but 🐑

shared/actions/teams.js
MarcoPolo: This was a little tricky since this throws an error if you aren't an admin of the team. And reloadable will present itself if you have an error in one of your waiting keys. So it made it seem like the page had broken when in reality it was just a forgivable error. 

We can change this to use a different waiting key, but I just removed it since it didn't seem important to the user?

I can add another waiting key and have the progress indicator reference that one as well and have reloadable ignore it.

shared/actions/teams.js
MarcoPolo: o hmm, I thought this was controlled by my formatter. I'll try to change it

chrisnojima: so prettier seems to have different rules given its input so if you do like
```
{a: 1}
``` 
it'll leave it
but if you do like
```
{a:
1}
```
it'll make it 3 lines

shared/actions/teams.js
chrisnojima: my theory of the reload was the waiting keys it uses should be ones that if they fail mean the screen isn't correct and should be covered. so i think the waiting keys should be segregated.
so effectively you'd have a larger set of waiting keys for your waitingbuttons (tied to the actions/sagas) then a subset that are mission critical that are tied to the reload state

MarcoPolo: sounds good

PR #15702, opened 2019-01-25
conflict_resolver: prevent cancellation confusion

jakob223:
This fixes an issue where CR can temporarily look finished if the `cancelExistingLocked` call reduces the wait group counter to 0 before the `Add` call can increase it again.

Thanks @strib for helping figure out what was going on here.

Issue: KBFS-3726

### DIFF
--- a/go/kbfs/libkbfs/conflict_resolver.go
+++ b/go/kbfs/libkbfs/conflict_resolver.go
@@ -254,6 +254,11 @@ func (cr *ConflictResolver) Resolve(ctx context.Context,
 		return
 	}
 
+	// Call Add before cancelling existing CR in order to prevent the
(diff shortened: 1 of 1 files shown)


jakob223:
I was not. Do you have any ideas to induce this particular order of happenings? It seems extremely rare based on my trying to replicate it.

strib:
Bummer.  No, I don't have any good ideas.  As long as this fix doesn't seem to induce other flakes though, it's definitely worth trying.

jakob223:
There's another similar CR flake that I'm also looking into, and

jakob223:
The "close and comment" button is in the place where one might expect a "never mind, don't comment" button to be..

PR #15705, opened 2019-01-25
new size type for icon

chrisnojima:
@keybase/react-hackers adds a new prop for icon

Icon itself is a mess, i made a ticket to redo it


chrisnojima:
@keybase/react-hackers bump

PR #15708, opened 2019-01-25
User is not in team error for getThreadNonblock

buoyad:
Does this look alright @mmaxim? @keybase/react-hackers


mmaxim:
Let me just change this so this error object has `code` on it.

mmaxim:
Ok the error should now have the code on it if you want to change to that.

PR #15710, opened 2019-01-25
Only suggest chat commands at the beginning of a message

buoyad:


buoyad:
There's a problem with this if you have a message like `' hello'` and cursor back to the beginning and enter a slash. The up/down arrows won't work. I'll add a fix soon

PR #15714, opened 2019-01-26
Tweaks on suggestions

cecileboucheron:
@keybase/react-hackers cc @mmaxim


shared/chat/conversation/input-area/normal/index.js
```diff
@@ -289,20 +289,20 @@ class Input extends React.Component<InputProps, InputState> {
       direction="vertical"
       style={Styles.collapseStyles([
         styles.suggestionBase,
-        {height: 50},
+        styles.fixSuggestionHeight,
```
mmaxim: This won't work on mobile

mmaxim: Or at least you should confirm it renders properly there.

PR #15715, opened 2019-01-26
mocks: regenerate mocks

strib:
@jzila Looks like we lost the Jenkinsfile mocks check as part of the monorepo merge.  Intended?


jakob223:
I just submitted a PR that changes one of these functions' signatures. I added the mocks to it, can we just use that commit to avoid merge hell?

Also, I would appreciate re-adding the mock test in Jenkins if it wasn't removed for some higher reason.

jakob223:
also, 😫 this is flaking 4 different KBFSDocker tests once each, is it worth making tickets for each?

strib:
@jakob223 am fine with closing this in favor of your pr.

PR #15717, opened 2019-01-26
dirty_bcache_disk: disk-based dirty block cache, and use it in CR

strib:
Issue: KBFS-3679

### DIFF
--- a/go/kbfs/libkbfs/conflict_resolver.go
+++ b/go/kbfs/libkbfs/conflict_resolver.go
@@ -7,6 +7,7 @@ package libkbfs
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	sysPath "path"
 	"runtime/debug"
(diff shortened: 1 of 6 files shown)


go/kbfs/libkbfs/conflict_resolver.go
```diff
@@ -3324,6 +3325,57 @@ func (cr *ConflictResolver) recordFinishResolve(
...
+		_ = cr.config.(*ConfigLocal).EnableDiskLimiter(os.TempDir())
+
+		dbc, err = newDiskBlockCacheLocalForTest(
+			cr.config, syncCacheLimitTrackerType)
+		if err != nil {
+			return nil, nil, err
+		}
+		cleanupFn = dbc.Shutdown
+	} else {
+		tempDir, err := ioutil.TempDir(cr.config.StorageRoot(), "kbfscr")
```
jakob223: There's already a `kbfs_conflicts` for the CR levelDB - maybe `kbfscr` should get a slightly clearer name? [and maybe the other one should too..]

go/kbfs/libkbfs/conflict_resolver.go
```diff
@@ -3324,6 +3325,57 @@ func (cr *ConflictResolver) recordFinishResolve(
...
+		}
+		cleanupFn = func(ctx context.Context) {
+			dbc.Shutdown(ctx)
+			dirCleanupFn(ctx)
+		}
+	}
+
+	err = dbc.WaitUntilStarted()
+	if err != nil {
+		if cleanupFn != nil {
```
jakob223: Maybe I'm just failing at following control flow, but it doesn't seem to me like `cleanupFn` can be nil.

jakob223: Never mind, I see now I think.

go/kbfs/libkbfs/dirty_bcache_disk.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"context"
+	"sync"
+
+	"github.com/keybase/client/go/kbfs/tlf"
+	"github.com/pkg/errors"
+)
+
+type dirtyBlockCacheDiskConfig interface {
```
jakob223: Why here and not in `interfaces.go`?

go/kbfs/libkbfs/dirty_bcache_disk.go
```diff
@@ -0,0 +1,153 @@
...
+func (d *DirtyBlockCacheDisk) Put(
+	ctx context.Context, tlfID tlf.ID, ptr BlockPointer, branch BranchName,
+	block Block) error {
+	if branch != d.branch {
+		return errors.Errorf(
+			"Branch %s doesn't match branch %s", branch, d.branch)
+	}
+
+	// Need to ready the block, since the disk cache expects encrypted
+	// and a block ID that can be verified against that data.
```
jakob223: I don't understand this sentence.

go/kbfs/libkbfs/conflict_resolver.go
strib: Ok, changing it to "kbfs_conflict_disk_cache".

go/kbfs/libkbfs/conflict_resolver.go
strib: Actually I think it can't be `nil` (it could be in an earlier iteration of the code), but it doesn't seem bad to continue to do the check just in case.

go/kbfs/libkbfs/dirty_bcache_disk.go
strib: This is a pattern we use when specifying a subset of the large `Config` interface, which includes only the exact items that will be used by the service.  We do this in several other places, and it's always defined alongside the struct that uses it.  I guess since it's only used in one place, it's easier to read the code that way, as opposed to putting it in a different file.

go/kbfs/libkbfs/dirty_bcache_disk.go
strib: Oops it's missing a word: "expects encrypted" -> "expects encrypted data".  I'll fix that.

In general though, when we "ready" a block, we take the plaintext block and encrypt it, producing an block ID, a new per-block key, and the encrypted data.  On-disk caches must only ever store encrypted data, so we have to encrypt it first and make a block ID that is truly a hash over the encrypted data.  Otherwise the block cache will reject and/or we won't be able to re-assemble it into the plaintext block later.  Does that help give it context?

strib:
> I don't think I fully understand why this change was needed - was CR failing due to running out of room in memory for the dirty block cache?

@jakob223 thanks for the review!  Yes, if someone has file data that conflicts with the state of the server, then CR makes a copy of all that data (to construct a new file with a ".conflicted" suffix), and it puts those blocks into the dirty block cache until they're written to the journal.  So if there is say a 2 GB file that conflicts, before this PR all that data will go into memory and could cause an out-of-memory panic.  More so if there are multiple big conflicting files at the same time.

PR #15725, opened 2019-01-27
Fix error handling in buildpayment

mlsteele:


mlsteele:
This was the result of my misunderstanding of what the problem was in the ticket. Still might as well merge.

PR #15728, opened 2019-01-28
conflict_resolver: be ok with corrupted DB

jakob223:
Previously, if KBFS shut down uncleanly, it would panic on startup because of a
corrupted levelDb.

This PR switches to using existing KBFS helper methods to interact with levelDB, and thus trying to recover the DB if opening it doesn't work.

### DIFF
--- a/go/kbfs/libkbfs/leveldb.go
+++ b/go/kbfs/libkbfs/leveldb.go
@@ -35,19 +35,22 @@ var leveldbOptions = &opt.Options{
 	OpenFilesCacheCapacity: 10,
 }
 
-type levelDb struct {
+// LevelDb is a libkbfs wrapper for leveldb.DB.
+type LevelDb struct {
 	*leveldb.DB
 	closer io.Closer
 }
 
-func (ldb *levelDb) Close() (err error) {
+// Close closes the DB.
+func (ldb *LevelDb) Close() (err error) {
 	err = ldb.DB.Close()
 	// Hide the closer error.
 	_ = ldb.closer.Close()
 	return err
 }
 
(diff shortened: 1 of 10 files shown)


go/kbfs/libkbfs/leveldb.go
```diff
@@ -35,19 +35,22 @@ var leveldbOptions = &opt.Options{
 	OpenFilesCacheCapacity: 10,
 }
 
-type levelDb struct {
+// LevelDb is a libkbfs wrapper for leveldb.DB.
+type LevelDb struct {
```
jzila: If we're going to export it, then it might as well go into its own package.

jakob223: The linter yelled at me for using it in an exported function and not exporting it. It may be possible to un-export that function instead though.

jakob223: `GetConflictResolutionDB` which is in an interface right now so probably can't be simply unexported. But maybe there's a better way to have a single CR DB that exists always and all CRs can access than putting it on the config.

jzila: OK feel free to leave it then.

jakob223: @jzila do you have package name opinions? `kbldb`?

jzila: `kbfsleveldb` probably, although its patterns could be universally useful.

jakob223:
Moving it to its own package creates an import cycle, so I'm just going to merge this now so that we can make sure it doesn't cause any more problems before the release.

jzila:
Yeah I was afraid of that. It'll take some finessing.

PR #15731, opened 2019-01-28
drop chat dependency on kbfs

mmaxim:


strib:
@mmaxim what happened to the review process here?  Shouldn't we avoid duplicating code?

PR #15740, opened 2019-01-28
accept prefixless passphrase when known during provisioning

heronhaye:


patrickxb:
can you add a test in `login_test.go` that uses a paper key without the first two words?

heronhaye:
Thanks, added, I was having trouble figuring out where to put it

go/libkb/passphrase_helper.go
```diff
@@ -171,6 +179,7 @@ func DefaultPassphrasePromptArg(m MetaContext, username string) keybase1.GUIEntr
 // hint otherwise.
 type PassphraseChecker interface {
 	Check(MetaContext, string) error
+	Automutate(MetaContext, string) (string, error)
```
patrickxb: What about moving this function into libkb/passphrase_helper.go?  Seems weird for the Checker to be doing anything besides Check.

heronhaye: Do you mean something like an `automutate` function being passed to `GetPassphraseUntilCheck`, where it would be nil if it wasn't getting a paperkey?

go/libkb/passphrase_helper.go
```diff
@@ -131,6 +131,14 @@ func GetPassphraseUntilCheck(m MetaContext, arg keybase1.GUIEntryArg, prompter P
 		if checker == nil {
 			return res, nil
 		}
+
+		s := res.Passphrase
+		t, err := checker.Automutate(m, s)
```
patrickxb: I meant that instead of calling `checker.Automutate(...)` here you would call `paperKeyPrefixMutate(m, s, expectedPrefix)` and have that function in this `passphrase_helper.go` file.

heronhaye: But how would this function know if it was prompting for a paperkey? Should that be passed in from the caller?

heronhaye: nvm, I can inspect the gui arg...

heronhaye: but then you still have to pass in the expectedPrefix right

patrickxb: sorry, i thought this was in `GetPaperKeyPassphrase`....GH hid the rest...

patrickxb: ok, we can leave it the way you have it.

PR #15741, opened 2019-01-28
Badge account dropdown if another account has unread txs

buoyad:
Currently we badge the dropdown if the selected account has badgeworthy transactions. That doesn't help direct attention to the dropdown when we want to - when an account besides this one has unread payments. This makes it so we badge if another account has unread. r? @keybase/react-hackers

### DIFF
--- a/shared/wallets/wallet/header/container.js
+++ b/shared/wallets/wallet/header/container.js
@@ -1,9 +1,12 @@
 // @flow
 import {connect, isMobile} from '../../../util/container'
+import {memoize} from '../../../util/memoize'
 import * as Constants from '../../../constants/wallets'
 import * as Types from '../../../constants/types/wallets'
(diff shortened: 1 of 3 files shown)


shared/wallets/wallet/header/container.js
```diff
@@ -1,10 +1,13 @@
 // @flow
 import {connect, isMobile} from '../../../util/container'
+import {memoize} from '../../../util/memoize'
 import * as Constants from '../../../constants/wallets'
 import * as Types from '../../../constants/types/wallets'
 import * as WalletsGen from '../../../actions/wallets-gen'
 import Header from '.'
 
+const otherUnreadPayments = memoize((map, accID) => !!map.delete(accID).some(v => !!v))
```
chrisnojima: some(Boolean)?

PR #15742, opened 2019-01-28
folder_branch_ops: don't modify already-cached MDs

jakob223:
I'm still failing 2 tests, and I'm not sure what the right approach for fixing them is. Putting up this PR to get opinions on that (CC @strib )
Issue: KBFS-3769

### DIFF
--- a/go/kbfs/libkbfs/folder_branch_ops.go
+++ b/go/kbfs/libkbfs/folder_branch_ops.go
@@ -2423,7 +2423,11 @@ func (fbo *folderBranchOps) initMDLocked(
 		return RekeyConflictError{err}
 	}
 
-	md.loadCachedBlockChanges(ctx, bps, fbo.log)
+	md, irmd, err = fbo.loadCachedMDChanges(ctx, bps,
+		session.VerifyingKey, md, irmd)
+	if err != nil {
+		return err
+	}
 
 	fbo.headLock.Lock(lState)
 	defer fbo.headLock.Unlock(lState)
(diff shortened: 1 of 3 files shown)


go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -3558,7 +3559,24 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
 		}
 	}
 
-	md.loadCachedBlockChanges(ctx, bps, fbo.log)
+	mdCopy, err := md.deepCopy(fbo.config.Codec())
```
strib: You can avoid the whole copy if `md.data.Changes.Ops != nil` already.  That's already checked in `loadCachedBlockChanges`, so in fact it might make more sense to change `loadCachedBlockChanges` itself to make and return a copy, if a copy is needed?  Since there are a bunch more callers of this function in this file.

I.e., `md.loadCachedBlockChanges` returns itself if `md.data.Changes.Ops != nil`, otherwise it returns a copy of itself with the changes made.  (It would also need to accept the codec as an argument.)  If it also returned a bool to say whether a copy was made, then all the callers could make a new `ImmutableRootMetadata` in that case.

go/kbfs/libkbfs/kbfs_ops_concur_test.go
```diff
@@ -762,8 +763,11 @@ func TestKBFSOpsConcurBlockSyncTruncate(t *testing.T) {
 	}
 
 	lastKMD := km.getLastKMD()
+	lastRMD, ok := lastKMD.(ReadOnlyRootMetadata)
+	require.True(t, ok)
 
-	if md.ReadOnlyRootMetadata != lastKMD {
+	// TODO: compare these in a different way
+	if md.ReadOnlyRootMetadata != lastRMD {
```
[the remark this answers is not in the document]
jakob223: Well, now they pass, so I guess they were failing because of the `mdID` change?

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -3558,7 +3569,19 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
 		}
 	}
 
-	md.loadCachedBlockChanges(ctx, bps, fbo.log)
+	md, copied := md.loadCachedBlockChanges(ctx, bps, fbo.log,
```
strib: Can we make a `folderBranchOps` helper function for this repeated block of code?  Not sure if `md` or `irmd` has to live on after the function is called, but if so it could return them.

PR #15747, opened 2019-01-29
Identify3 edit own profile

chrisnojima:
- [x] need sigid plumbing from core


buoyad:

shared/actions/json/profile.json
```diff
@@ -4,25 +4,26 @@
     "import * as More from '../constants/types/more'"
   ],
   "actions": {
-    "addProof": { "platform": "More.PlatformsExpandedType" },
+    "addProof": {"platform": "More.PlatformsExpandedType"},
     "backToProfile": {},
     "cancelAddProof": {},
     "cancelPgpGen": {},
     "checkProof": {},
+    "checkProof2": {"sigID": "string"},
```
chrisnojima: this isn't a replacement for checkproof so we should name it differently. maybe it should be like
checkBuildingProof and checkSingleProof or something

shared/util/feature-flags.desktop.js
```diff
@@ -14,7 +14,7 @@ const ff: FeatureFlags = {
   admin: false,
   chatIndexProfilingEnabled: false,
   foldersInProfileTab: false,
-  identify3: false,
+  identify3: true,
```
chrisnojima: lets keep this admin

shared/tracker2/assertion/container.js
```diff
@@ -1,25 +1,38 @@
...
+import * as RouteTreeGen from '../../actions/route-tree-gen'
 import * as WalletsType from '../../constants/types/wallets'
 import * as Constants from '../../constants/tracker2'
 import Assertion from '.'
 import openUrl from '../../util/open-url'
 
 type OwnProps = {|
+  isSuggestion?: boolean,
   username: string,
+  // allowMenu: boolean,
```
chrisnojima: 🔥

go/service/user.go
```diff
@@ -475,6 +475,8 @@ func (h *UserHandler) proofSuggestionsHelper(mctx libkb.MetaContext) (ret []Proo
 		switch displayConfig.Key {
 		case "zcash.t", "zcash.z", "zcash.s":
 			altKey = "zcash"
+		case "bitcoin":
+			altKey = "btc"
```
chrisnojima: this this affect ident2?

shared/util/feature-flags.desktop.js
buoyad: Oops thanks

go/service/user.go
mlsteele: Nope this only affects the ProofSuggestions RPC which was not used prior to id3

shared/actions/profile/proofs.js
```diff
@@ -38,6 +41,18 @@ const checkProof = (state, action) => {
     })
 }
 
+const recheckProof = (state, action) =>
+  flags.identify3 &&
+  RPCTypes.proveCheckProofRpcPromise({sigID: action.payload.sigID}, Constants.waitingKey).then(() =>
+    Tracker2Gen.createLoad({
```
buoyad: @chrisnojima this seems heavy handed, does it seem right to you? I was looking for something that triggers `identify3UpdateRow`

chrisnojima: lets keep it simple, it should be fast as everything should be cached

chrisnojima:
👍

PR #15756, opened 2019-01-29
Stellar settings: be aware if we can send tx, do not show "Set Inflation" button if not

zapu:
Present the distinction between "funded accounts" and not to the frontend, as well as another field that says whether there is enough to make a tx or not. Use this info to hide "Set inflation" button if this account can't set any inflation yet.

### DIFF
--- a/protocol/avdl/stellar1/local.avdl
+++ b/protocol/avdl/stellar1/local.avdl
@@ -15,6 +15,8 @@ protocol local {
     CurrencyLocal currencyLocal;// user configurable, defaults to USD and related details
     AccountMode accountMode;    // mobile only vs all devices
     boolean accountModeEditable;// is this device able to change it
+    boolean isFunded;           // does the account have min amount of XLM (exists on the network)
+    boolean canMakeTx;          // does the account have enough XLM for a transaction fee
   }
   array<WalletAccountLocal> getWalletAccountsLocal(int sessionID);
   WalletAccountLocal getWalletAccountLocal(int sessionID, AccountID accountID);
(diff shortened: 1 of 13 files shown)


zapu:
@patrickxb can you review core side? thanks!

protocol/avdl/stellar1/local.avdl
```diff
@@ -14,6 +14,8 @@ protocol local {
     string seqno;               // the stellar sequence number for this account
     CurrencyLocal currencyLocal;// user configurable, defaults to USD and related details
     AccountMode accountMode;    // mobile only vs all devices
+    boolean isFunded;           // does the account have min amount of XLM (exists on the network)
+    boolean canMakeTx;          // does the account have enough XLM for a transaction fee
```
zapu: I'm fine with renaming this field, I couldn't come up with a good name

patrickxb: I don't have much better....`canSubmitTx`, `isFundedEnoughForTx`, `hasAvailableBalance`

shared/wallets/wallet/settings/index.js
```diff
@@ -177,12 +178,19 @@ class AccountSettings extends React.Component<SettingsProps> {
...
+                <Kb.Button
+                  type="Secondary"
+                  label={props.inflationDestination ? 'Change' : 'Set up'}
+                  onClick={props.onSetupInflation}
+                  style={styles.setupInflation}
+                />
+              )}
+              {!props.canSubmitTx && (
+                <Kb.Text type="BodySmall">
+                  Your account needs a minimum balance to set an inflation destination.
```
buoyad: 'a minimum balance' -> 'more funds'

PR #15757, opened 2019-01-29
use consistent envfile directory

heronhaye:
I had a workaround in the unitfile to use $HOME/.config and not respect $XDG_CONFIG_HOME due to systemd issues, but did not change go code to reflect that.

### DIFF
--- a/go/libkb/env.go
+++ b/go/libkb/env.go
@@ -553,10 +553,16 @@ func (e *Env) GetRootConfigFilename() (string, error) {
 	return filepath.Join(dir, "config.json"), nil
 }
 
+func (e *Env) GetEnvFileDir() string {
+	// Do not respect $XDG_CONFIG_HOME due to debian systemd 229 not supporting %E
+	// see keybase.service systemd unit
+	return filepath.Join(e.GetHome(), ".config", "keybase")
+}
+
 func (e *Env) GetEnvfileName() (string, error) {
 	switch RuntimeGroup() {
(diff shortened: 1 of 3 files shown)


go/libkb/env.go
```diff
@@ -553,10 +553,16 @@ func (e *Env) GetRootConfigFilename() (string, error) {
 	return filepath.Join(dir, "config.json"), nil
 }
 
+func (e *Env) GetEnvFileDir() string {
+	// Do not respect $XDG_CONFIG_HOME due to debian systemd 229 not supporting %E
```
strib: It's confusing to have a comment about XDG here in what appears to be a general-use function.  Can you name it `GetNixEnvFileDir` or something?  Or move it into the platform-specific files and panic on non-nix systems?

packaging/linux/systemd/keybase.service
```diff
@@ -11,6 +11,8 @@ EnvironmentFile=-%t/keybase/keybase.env
 
 # Use %h/.config instead of %E because %E isn't supported in systemd 229
 # though this breaks non-standard $XDG_CONFIG_HOMEs
+# See GetEnvFileDir; change when Debian
```
strib: End the line above with a period.

go/libkb/env.go
heronhaye: good point, i made it return an error if in a unsupported runtimegroup like the other functions.

PR #15761, opened 2019-01-29
add KEYBASE_PROVE_BYPASS

mlsteele:
Using the env var `KEYBASE_PROVE_BYPASS` or flag `--prove-bypass` on the service (not the client) will make your client ignore `creation_disabled` for all proof types. (Except facebook which is still hardcoded in the client)

cc @xgess


xgess:
nice! sorry i missed this

PR #15768, opened 2019-01-29
Joshblum/device ek log core 10016

joshblum:
- adds a separate log file for EK deletion actions to aid in debugging 
- sends this file in log sends
- cleans up extraneous ek cache clearing during deletes

depends on: https://github.com/keybase/keybase/pull/3406
- [ ] test on mobile build


mmaxim:
How big is this file?

joshblum:
capped at 128 mb before rotation, same as other log file: https://github.com/keybase/client/pull/15768/files#diff-653fe7b48864b80d9651af697b63167bR44

mmaxim:
I just mean on average, my worry is that we will xfer a ton of mostly useless data and slow the log send down. Also it should be capped much lower on mobile.

joshblum:
in normal use it will write ~3 lines per day for each deletion, 30 day rotation cap keeps it small still, lemme look into mobile limits

joshblum:
@mmaxim mobile capped to 1mb (but kept 30days for rotate)

PR #15769, opened 2019-01-29
prefetcher: key queued requests by ptr, not by block ID

strib:
If we use block ID, we might end up canceling legitimate requests for deduplicated blocks with the same ID but a different nonce as the one being canceled.

Issue: KBFS-3756


jzila:
This means that we're also re-duplicating prefetches for de-duped blocks. I.e. we'll prefetch it multiple times.

strib:
> This means that we're also re-duplicating prefetches for de-duped blocks. I.e. we'll prefetch it multiple times.

Why does it mean that?  This data structure is just tracking the queued requests for later canceling.  This is not the same as the `prefetches` map, which is still keyed by blockID.

PR #15771, opened 2019-01-30
delegate chat UI

mmaxim:
This makes it easier to get the UI to do stuff from the service without having to use `NotifyRouter`, which KBFS pays attention to now. It also fits in better with existing patterns elsewhere in the codebase.


chrisnojima:
little confused, this changes what? this is a stepping stone to more code?

mmaxim:
Yes, just additive, I broke it out from another PR.

chrisnojima:
k

go/libkb/identify3_test.go
```diff
@@ -23,6 +25,7 @@ func (i *id3FakeUIRouter) GetSecretUI(sessionID int) (SecretUI, error)
 func (i *id3FakeUIRouter) GetRekeyUI() (keybase1.RekeyUIInterface, int, error)       { return nil, 0, nil }
 func (i *id3FakeUIRouter) GetRekeyUINoSessionID() (keybase1.RekeyUIInterface, error) { return nil, nil }
 func (i *id3FakeUIRouter) GetHomeUI() (keybase1.HomeUIInterface, error)              { return nil, nil }
+func (i *id3FakeUIRouter) GetChatUI() (chat1.ChatUiInterface, error)                 { return nil, nil }
```
mmaxim: @maxtaco For this kind of thing I usually put the interface into the fake struct, and implement only the method I need for the test so that people don't need to track these things down when the interface changes. Something like this:

```
type id3FakeUIRouter struct {
   libkb.UIRouter
   ui id3FakeUI
}

func (i *id3FakeUIRouter) GetIdentify3UI(MetaContext) (keybase1.Identify3UiInterface, error) {
	&i.ui, nil
}	
```

PR #15777, opened 2019-01-30
Prove bypass flaggery

mlsteele:


go/libkb/identify_outcome.go
```diff
@@ -65,8 +65,7 @@ func (i *IdentifyOutcome) TrackSet() *TrackSet {
 }
 
 func (i *IdentifyOutcome) ProofChecksSorted() []*LinkCheckResult {
-	// TODO Remove with CORE-9923
-	useDisplayPriority := i.G().Env.GetRunMode() == DevelRunMode || i.G().Env.RunningInCI()
+	useDisplayPriority := i.G().ShouldUseParameterizedProofs()
```
joshblum: any concern that this now runs for admins?

mlsteele: All the better if they notice a bug

mlsteele:
@xgess when released `--prove-bypass` should actually work for non-admins on prod.

PR #15780, opened 2019-01-30
tlf_journal_test: Try to address the deadlock

jzila:
cc @strib 

I'm guessing that the journal just isn't blocking, and this should address that scenario.

### DIFF
--- a/go/kbfs/libkbfs/tlf_journal_test.go
+++ b/go/kbfs/libkbfs/tlf_journal_test.go
@@ -1518,9 +1518,11 @@ func testTLFJournalConvertWhileFlushing(t *testing.T, ver kbfsmd.MetadataVer) {
 		// Now finish the block put, and let the flush finish.  We should
 		// be on a local squash branch now.
 		case unpauseBlockPutCh <- struct{}{}:
+			continue
 		case err := <-errCh:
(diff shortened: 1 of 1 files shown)


jzila:
I think I just changed the flake from one form to another, but at least it no longer deadlocks.

jzila:
Ready for review.

go/kbfs/libkbfs/tlf_journal_test.go
```diff
@@ -1513,15 +1519,11 @@ func testTLFJournalConvertWhileFlushing(t *testing.T, ver kbfsmd.MetadataVer) {
...
-		select {
-		// Now finish the block put, and let the flush finish.  We should
-		// be on a local squash branch now.
-		case unpauseBlockPutCh <- struct{}{}:
-		case err := <-errCh:
-			require.NoError(t, err)
-		}
-	}
+	// Now finish the block put, and let the flush finish.  We
+	// shouldn't be on a branch anymore.
```
strib: We should be on a branch -- the local squash branch.  In fact there's a require just below that checks that.  Same thing below.

strib:
I'm going to fix the comment myself since you're out tomorrow and we should get this in ASAP.

PR #15781, opened 2019-01-30
Don't try to show a loading line on the left side of the inbox sidebar on desktop

cjb:
@keybase/react-hackers CC @songgao 

A little hard to follow the intent of the change here, but I think this gets us back to where we were -- the loading line lives at this point (underneath "Jump to Chat...") on mobile, but here in this desktop codepath we do the loading state on the right hand side of the sidebar instead, so I think this component shouldn't be here.  It was causing "Jump to Chat..." to change size as you move between convos on desktop.


songgao:
Ah! Just realized you did this change here. I had an attempt in another PR. I think think removing this part on desktop makes more sense though. So I'm gonna close mine. Thanks!

PR #15784, opened 2019-01-30
improve handling of slash commands

mmaxim:
Patch does the following:

1. Adds categories of built in commands so we can show only those relevant for the current conversation. For example. /headline is now only available in big team chats, and /leave is only available in big team chats that are not #general.
2. Makes /join pop open the manage channels dialog instead of trying to join you into the channel.

cc @malgorithms


shared/actions/chat2/index.js
```diff
@@ -2920,6 +2928,15 @@ function* chat2Saga(): Saga.SagaGenerator<any, any> {
     EngineGen.chat1NotifyChatNewChatActivity,
     onNewChatActivity
   )
+  yield* Saga.chainAction<EngineGen.Chat1ChatUiChatShowManageChannelsPayload>(
+    EngineGen.chat1ChatUiChatShowManageChannels,
+    onChatShowManageChannels
+  )
+
+  yield* Saga.chainAction<ConfigGen.SetupEngineListenersPayload>(
```
joshblum: why?

mmaxim: Makes the delegated chat UI work in the service.

PR #15789, opened 2019-01-30
Don't try to load Stellar info if we're logged out

cjb:
@keybase/react-hackers 

There's a race where we can log out, and then receive a remote Stellar notification which causes us to queue up a call to LoadAssets, which makes RPCs that black bar because we're logged out.  Throwing a `&& state.config.loggedIn &&` on everything that can be caused by a notification seems like the right idea.


cjb:
@buoyad Thanks! Made those changes, ready for another look.

PR #15797, opened 2019-01-31
dsl_test: call SyncFromServer at the end of every `as` block

strib:
Tests that check the edit history sometimes go in this order:

1) bob writes something.
2) alice checks the edit history for that write.

Unfortunately, the chat notification from bob is sent asynchronously after the write operation, and a `bob.SyncFromServer()` call is needed to make sure that the notification was sent all the way through to all the other users.  If there isn't another `bob` block between steps 1 and 2, there's no guarantee that this will happen, since currently in the code `SyncFromServer` is only called at the _beginning_ of each block.

Instead, let's also call it at the _end_ of each block, unless the test specifically doesn't want to and disables the behavior.

Issue: KBFS-3804

### DIFF
--- a/go/kbfs/test/dsl_test.go
+++ b/go/kbfs/test/dsl_test.go
@@ -432,6 +432,7 @@ type ctx struct {
 	rootNode   Node
 	noSyncInit bool
 	staller    *libkbfs.NaïveStaller
+	noSyncEnd  bool
 }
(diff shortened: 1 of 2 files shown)


strib:
@jakob223 this might also fix the user history test flake you were looking at.

PR #15799, opened 2019-01-31
ID3 fixes: show SigIDs and SiteURLs

maxtaco:
- show sigIDs in almost all cases
- show siteURLs for PGP keys and crypto addresses, whereas previously we weren't


maxtaco:
cc @chrisnojima

maxtaco:
CI all borked, not related though

PR #15803, opened 2019-01-31
don't error out on rekey errors when looking up name, just return no conversations

mmaxim:
Otherwise you get random black bars.


go/chat/helper.go
```diff
@@ -514,8 +514,16 @@ func FindConversations(ctx context.Context, g *globals.Context, debugger utils.D
 		inbox, _, err := g.InboxSource.Read(ctx, uid, types.ConversationLocalizerBlocking, useLocalData, nil,
 			query, nil)
 		if err != nil {
+			acceptableErr := false
+			// if we fail to loas the team for some kind of rekey reason, treat as a complete miss
```
joshblum:
```suggestion
			// if we fail to load the team for some kind of rekey reason, treat as a complete miss
```

PR #15806, opened 2019-01-31
fix random logout bug on desktop

maxtaco:
- the bug was that ClearSecret() was being being called with an empty username, which
  seems to blast everything out of the keychain
- this codepath is triggered on Login, since Login logs out the previous user, which can
  easily be "" for a new install

### DIFF
--- a/go/libkb/globals.go
+++ b/go/libkb/globals.go
@@ -325,7 +325,7 @@ func (g *GlobalContext) Logout(ctx context.Context) (err error) {
 
 	// remove stored secret
 	g.secretStoreMu.Lock()
-	if g.secretStore != nil {
+	if g.secretStore != nil && !username.IsNil() {
 		if err := g.secretStore.ClearSecret(mctx, username); err != nil {
 			mctx.CDebugf("clear stored secret error: %s", err)
 		}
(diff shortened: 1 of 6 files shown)


maxtaco:
cc @mlsteele and @malgorithms who hit this bug

maxtaco:
Really?  I don't want to break anything. Doesn't seem like we need to...

maxtaco:
Unfortunately the password store stuff is maze, I don't trust myself to follow all paths into, out of, and through it.

patrickxb:
This seems fine to me.  In libkb/globals.go#Logout(), we could check for empty username too and not even try, but this is fine too.

maxtaco:
We do check in both places...

go/libkb/globals.go
```diff
@@ -325,7 +325,7 @@ func (g *GlobalContext) Logout(ctx context.Context) (err error) {
 
 	// remove stored secret
 	g.secretStoreMu.Lock()
-	if g.secretStore != nil {
+	if g.secretStore != nil && !username.IsNil() {
```
patrickxb: missed this...looks good.

PR #15807, opened 2019-01-31
Fix list channels

joshblum:
Fixes a bug where we would not process a gregor message for `NewConversation` properly when the conv was in the `NEVER_JOINED` state


go/chat/inboxsource.go
```diff
@@ -584,7 +585,7 @@ func (s *HybridInboxSource) Read(ctx context.Context, uid gregor1.UID,
 	localizerTyp types.ConversationLocalizerTyp, useLocalData bool, maxLocalize *int,
 	query *chat1.GetInboxLocalQuery, p *chat1.Pagination) (inbox types.Inbox, localizeCb chan types.AsyncInboxResult, err error) {
 
-	defer s.Trace(ctx, func() error { return err }, "Read")()
+	defer s.Trace(ctx, func() error { return err }, fmt.Sprintf("Read: query: %v", query))()
```
mmaxim: Topic names could be in here, we cant log this.

mmaxim: Also I'm not a huge fan of spamming out too much logging here, I'd prefer to keep this log line terse.

joshblum: ah gotcha

PR #15808, opened 2019-01-31
disk_block_cache: use PRNG in test

jakob223:
Issue: KBFS-3755

### DIFF
--- a/go/kbfs/kbfsblock/id_test.go
+++ b/go/kbfs/kbfsblock/id_test.go
@@ -7,6 +7,7 @@ package kbfsblock
 import (
 	"encoding/binary"
 	"math"
+	"math/rand"
 	"testing"
 
 	"github.com/keybase/client/go/kbfs/kbfscodec"
@@ -65,6 +66,7 @@ func TestTemporaryIDRandom(t *testing.T) {
 
 // Test that MakeRandomIDInRange returns items in the range specified.
 func TestRandomIDInRange(t *testing.T) {
+	rand.Seed(1)
 	idToInt := func(id ID) uint64 {
 		idBytes := id.Bytes()[1:9]
 		return binary.BigEndian.Uint64(idBytes)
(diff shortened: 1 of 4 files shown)


go/kbfs/libkbfs/disk_block_cache.go
```diff
@@ -897,7 +897,7 @@ func (*DiskBlockCacheLocal) getRandomBlockID(numElements,
 	}
 	// Generate a random block ID to start the range.
 	pivot := (1.0 - (float64(numElements) / float64(totalElements)))
-	return kbfsblock.MakeRandomIDInRange(0, pivot)
+	return kbfsblock.MakeRandomIDInRange(0, pivot, fakeRandForTest)
```
strib: You can avoid passing in the bool above by using `cache.config.IsTestMode()` here instead, and naming the receiver of the function `cache`.

go/kbfs/kbfsblock/id_test.go
```diff
@@ -75,7 +75,7 @@ func TestRandomIDInRange(t *testing.T) {
 		for j := i * 2; j < (math.MaxUint64 / 2); j *= 2 {
 			iAsFloat := float64(i) / maxUintFloat
 			jAsFloat := float64(j) / maxUintFloat
-			id, err := MakeRandomIDInRange(iAsFloat, jAsFloat)
+			id, err := MakeRandomIDInRange(iAsFloat, jAsFloat, false)
```
jzila: Might as well make this `true` so it's deterministic as well (and seed `rand` at the beginning of the test).

go/kbfs/libkbfs/disk_block_cache_test.go
```diff
@@ -804,7 +804,7 @@ func TestDiskBlockCacheHomeDirPriorities(t *testing.T) {
 
 	ctx := context.Background()
 
-	rand.Seed(time.Now().UnixNano())
```
jzila: Interesting. We probably weren't even using `rand` before.

PR #15810, opened 2019-01-31
don't drop trusted data for selected convo

buoyad:
@keybase/react-hackers


shared/constants/chat2/meta.js
```diff
@@ -135,22 +135,34 @@ export const updateMeta = (
     ? old.resetParticipants
     : meta.resetParticipants
 
-  return meta.withMutations(m => {
+  let base = meta
```
buoyad: I'm a little iffy on this part. A newer untrusted meta was wiping out data in an existing trusted meta. This switches the merge direction in that case if the inbox versions are the same.

mmaxim:
I think this is ok, but I wonder if we should hold off until after the release. Bugs in this stuff manifest in subtle ways.

buoyad:
@mmaxim sgtm, for posterity https://github.com/keybase/client/pull/15802 will fix the dropped retention bug in the release

PR #15816, opened 2019-01-31
Change how CanLogout works, make it local and heavily cached

zapu:
Let's do service in this PR and frontend in another one


zapu:
With a fair dice roll, @joshblum was selected. Feel free to pass forward if you are busy, though! Thank you

joshblum:
need to audit the computation ;)

go/service/user.go
```diff
@@ -594,3 +612,27 @@ func (h *UserHandler) LoadHasRandomPw(ctx context.Context, arg keybase1.LoadHasR
...
+		return keybase1.CanLogoutRes{
+			CanLogout: false,
+			Reason:    fmt.Sprintf("Cannot check user state: %s", err.Error()),
+		}, nil
+	}
+
+	if hasRandomPW {
+		return keybase1.CanLogoutRes{
+			CanLogout: false,
+			Reason:    "You signed up without a password. Set a password first.",
```
joshblum:
```suggestion
			Reason:    "You signed up without a password and need to set a password first.",
```

go/service/user.go
```diff
@@ -594,3 +612,27 @@ func (h *UserHandler) LoadHasRandomPw(ctx context.Context, arg keybase1.LoadHasR
...
+func (h *UserHandler) CanLogout(ctx context.Context, sessionID int) (res keybase1.CanLogoutRes, err error) {
+	hasRandomPW, err := h.LoadHasRandomPw(ctx, keybase1.LoadHasRandomPwArg{
+		SessionID:   sessionID,
+		ForceRepoll: false,
+	})
+
+	if err != nil {
+		return keybase1.CanLogoutRes{
+			CanLogout: false,
+			Reason:    fmt.Sprintf("Cannot check user state: %s", err.Error()),
```
joshblum: i think this would be better with a generic error message that is user friendly with internal logging of the raw error

zapu: Thanks, you are right. I will change that in the frontend ticket that I have in TODO to return some kind of special variable here that frontend can react differently to, instead of stuffing everything together in the reason string.

PR #15819, opened 2019-01-31
mount_test: sync fs2 before accessing files via mnt2

strib:
Not syncing used to work fine I think, because the first read would actively sync the folder.  But now with the edit notification stuff, the folder could be initialized at an earlier revision before an explicit read via the mount is done.  Thus, we need explicit syncing to make sure it's up-to-date.

Issue: KBFS-3758


go/kbfs/libdokan/mount_test.go
```diff
@@ -3128,7 +3137,7 @@ func TestKbfsFileInfo(t *testing.T) {
 		t.Fatal(err)
 	}
 	syncFilename(t, myfile1)
-	syncFolderToServer(t, "user1,user2", fs1)
+	syncFolderToServer(t, "user1,user2", fs2)
```
songgao: No need to sync fs1 anymore?

PR #15828, opened 2019-02-01
Client fixes for empty amount bug

patrickxb:


patrickxb:
@mlsteele you might not be happy with this, but these empty amount issues keep coming in despite all server fixes.

it's something when the user is accepting the disclaimer and obviously a crappy experience to get an error right after that, so i'd really like to get this in just to get rid of those errors.  we can investigate the cause later.

go/stellar/stellar.go
```diff
@@ -1606,7 +1606,9 @@ func FormatAmountWithSuffix(amount string, precisionTwo bool, simplify bool, suf
 
 func FormatAmount(amount string, precisionTwo bool, rounding FmtRounding) (string, error) {
 	if amount == "" {
-		return "", errors.New("empty amount")
+		// purists might complain here, but I really just want
+		// to get past this bug
+		amount = "0"
```
mlsteele: as long as this isn't used for anything to do with money it seems fine

mlsteele:
These keep coming in meaning we get logs from users? Let's add stack traces to the logs. Do we get any from admins?

patrickxb:
From users only.

PR #15839, opened 2019-02-01
fix after-delete and after-move routing for filepreview view

songgao:
1. When it's a delete, always navigate to the parent folder of the thing
that's deleted.

2. When it's a move, always navigate to the destination parent folder.


songgao:
beep bop bump

PR #15841, opened 2019-02-01
Fix kbnm findBinary

joshblum:
use `path.Dir` to get the parent directory of the binary to find the one we are looking for.


go/kbnm/findbin.go
```diff
@@ -14,24 +15,24 @@ var errKeybaseNotFound = errors.New("failed to find the keybase binary")
 // findKeybaseBinary returns the path to a Keybase binary, if it finds it.
 func findKeybaseBinary(name string) (string, error) {
 	// Is it near the kbnm binary?
-	dir, err := utils.BinPath()
+	binPath, err := utils.BinPath()
 	if err == nil {
-		path := filepath.Join(dir, name)
-		if _, err := os.Stat(path); !os.IsNotExist(err) {
-			return path, nil
+		binPath := filepath.Join(path.Dir(binPath), name)
```
songgao: Oh also, I think this is supposed to be `filepath.Dir`

joshblum: 😱 oops yes i'll fix

joshblum:
i don't have a great way to test the other branches so i think we should leave it as is without the `EvalSymlinks` call. that work?

songgao:
Yup sounds good! was just thinking out loud.

PR #15842, opened 2019-02-01
Dialog that explains what exploding messages are

zanderz:
Not hooked up yet. Looks ok on both mobile and desktop storybook.


shared/chat/conversation/messages/exploding-explainer-dialog/index.js
```diff
@@ -0,0 +1,103 @@
...
+  },
+  linkText: {
+    color: Styles.globalColors.white,
+  },
+  padding: Styles.platformStyles({
+    common: {
+      ...Styles.globalStyles.flexBoxColumn,
+      alignItems: 'center',
+      backgroundColor: Styles.globalColors.blue,
+      justifyContent: 'center',
```
songgao: Is there a reason not to use `Kb.Box2` here? `<Kb.Box2 direction="vertical" centerChildren={true}>` could replace `flexBoxColumn`+`alignItems`+`justifyContent`.

shared/chat/conversation/messages/exploding-explainer-dialog/index.js
```diff
@@ -0,0 +1,103 @@
...
+      marginRight: 80,
+      marginTop: 40,
+      minHeight: 440,
+    },
+    isMobile: {
+      flexGrow: 1,
+      flexShrink: 1,
+      maxHeight: '100%',
+      paddingTop: Styles.globalMargins.xlarge,
+      width: '100%',
```
songgao: If you used `Kb.Box2`, it seems this would be `fullWidth={isMobile}`

shared/chat/conversation/messages/exploding-explainer-dialog/container.js
```diff
@@ -0,0 +1,27 @@
...
+import {compose, connect} from '../../../../util/container'
+import {isMobile} from '../../../../constants/platform'
+
+type OwnProps = RouteProps<{}, {}>
+
+const mapStateToProps = (state, ownProps: OwnProps) => ({})
+
+const mapDispatchToProps = (dispatch, ownProps: OwnProps) => ({
+  onBack: isMobile ? undefined : () => dispatch(ownProps.navigateUp()),
+  onCancel: () => dispatch(ownProps.navigateUp()),
```
songgao: It looks like this is gonna be hooked up in routeTree directly given the `RouteProps<>` def above for `OwnProps`. So might as well use `putActionIfOnPath` here in case user quickly clicks more than once?

shared/chat/conversation/messages/exploding-explainer-dialog/index.js
zanderz: IDK I was just copying chat/delete-history-warning. Will change.

shared/chat/conversation/messages/exploding-explainer-dialog/index.js
zanderz: IDK I was just copying chat/delete-history-warning. Will change.

shared/chat/conversation/messages/exploding-explainer-dialog/container.js
zanderz: IDK I was just copying chat/delete-history-warning. I assumed this was necessary to go back to the right place? Is there a good example of an alternative?

zanderz:
Thanks Song!

shared/chat/conversation/messages/exploding-explainer-dialog/container.js
songgao: I searched for it, only to realize that only FS uses this now: https://github.com/keybase/client/blob/12d06a25888741544d419c79fe2e2802eb78ca79/shared/fs/oops/container.js#L19

So perhaps there's a better thing now? Not sure; @chrisnojima is `putActionIfOnPath` considered harmful now?

songgao:
NP! Love it that you copied a string multiple times in PR comments to make a point 😂

shared/chat/conversation/messages/exploding-explainer-dialog/container.js
buoyad: @songgao for routable components `ownProps.navigateUp` is actually `putActionIfOnPath` [here](https://github.com/keybase/client/blob/master/shared/route-tree/render-route.js#L85)

zanderz: @buoyad so does this look ok or should it just use navigateUp?

buoyad: @songgao yeah, if you want to do something other than navigate or if you're in a deep child of a routable component and don't want to deal with plumbing the closures through. All this might not be relevant soon because of the imminent nav change.

PR #15845, opened 2019-02-01
Disable auditors in share extension

joshblum:
- [ ] test share ext


mmaxim:
@joshblum  I'd just confirm the main app still does the audit, and the extension does not (no reason to think this is broken, just to be 100% sure)

joshblum:
will do!

joshblum:
confirmed

PR #15847, opened 2019-02-02
fix upload bar

songgao:
* don't be a hobbit in widget
* make broken pattern smooth
* use small pattern asset like the payment one now that `repeat` works
  for Android
* fix bug on mobile where upload bar only shows in mounted screens


songgao:
sorry this one's still having issues. Will tag review again when it's ready.

zanderz:
As per our chat, it's not working right for me

songgao:
OK this is ready again!

songgao:
bump!

songgao:
bumpy bump ⪮ @keybase/react-hackers

songgao:
⪮

jzila:
@songgao I think you need to fix some conflicts. FWIW I've tested and it works well, although my filename is always getting cut off by an ellipsis.

jzila:
image

jzila:
(that's `random.bin`)

songgao:
Ah; good catch!

shared/fs/folder/index.js
```diff
@@ -47,18 +46,9 @@ class Files extends React.PureComponent<FolderProps> {
           )}
           <Footer />
         </Kb.Box2>
-      </Kb.Box2>
+      </Kb.BoxGrow>
```
songgao: @chrisnojima  is this what you meant?

jzila:
Tested and approved

PR #15848, opened 2019-02-02
folder_block_manager: protect against degenerate folders

strib:
Really old TLFs might not have `LastGCRevision` field set in their MDs, and so have to search backwards for the last GC revision.  But trying to read revisions that are too old is too expensive, so just give up and stop QR instead.  Hopefully at some point we can revisit folders like this with a better QR algorithm.  This is just a quick fix for now.

Issue: KBFS-3839


strib:
Closing in preference to #15867 .

PR #15854, opened 2019-02-04
Move random pw signup flag to devel build restricted

zapu:
Batch mode was already disabled on prod builds. So this removes invalid flag combination from `signup` command and fixes the help page.

```
» keybase help signup                                     18:10:31
NAME:
   keybase signup - Signup for a new account

USAGE:
   keybase signup [command options]

OPTIONS:
   -c, --invite-code 	Specify an invite code.
   --email 		Specify an account email.
   --username 		Specify a username.
   --no-passphrase	Sign up without passphrase.
```
`--no-passphrase	Sign up without passphrase.` was useless because it only worked correctly with batch mode. In non-batch, the prompter wouldn't allow empty passphrase, so this command later errors out with
`▶ ERROR Requested random passphrase but passphrase was provided`

### DIFF
--- a/go/client/cmd_signup.go
+++ b/go/client/cmd_signup.go
@@ -36,10 +36,6 @@ func NewCmdSignup(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Comman
 				Name:  "username",
 				Usage: "Specify a username.",
 			},
-			cli.BoolFlag{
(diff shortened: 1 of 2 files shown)


zapu:
@chrisnojima should be fixed now, sorry for not QA-ing this thing properly on prod

PR #15856, opened 2019-02-04
fix setting retention policy on new convos

buoyad:
Confirmed that when the policy is set it sticks for the sender and the convo doesn't appear for the reciever. r? @keybase/react-hackers

### DIFF
--- a/shared/reducers/chat2.js
+++ b/shared/reducers/chat2.js
@@ -127,17 +127,22 @@ const metaMapReducer = (metaMap, action) => {
       })
     case Chat2Gen.updateConvRetentionPolicy:
       const {conv} = action.payload
-      const newMeta = Constants.inboxUIItemToConversationMeta(conv)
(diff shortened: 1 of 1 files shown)


shared/reducers/chat2.js
```diff
@@ -127,17 +127,22 @@ const metaMapReducer = (metaMap, action) => {
...
     case Chat2Gen.updateConvRetentionPolicy:
       const {conv} = action.payload
-      const newMeta = Constants.inboxUIItemToConversationMeta(conv)
+      const newMeta = Constants.inboxUIItemToConversationMeta(conv, true)
       if (!newMeta) {
         logger.warn('Invalid inboxUIItem received in conv retention policy update')
         return metaMap
       }
-      return metaMap.set(newMeta.conversationIDKey, newMeta)
+      // don't insert inbox item if this is the first we've heard about this convo
```
chrisnojima: lets reverse these comments

PR #15858, opened 2019-02-04
Fix ephemeralsuperseder metadata

joshblum:
fixes posting DELETEs when an ephemeral policy is set


mmaxim:
Maybe we should do the check in my new function though? Should it only act on the right message types?

joshblum:
yes i can add it there too

joshblum:
merging over flakes

PR #15859, opened 2019-02-04
Revert "cr: store pointers in the `fileBlockMap`, not full blocks"

strib:
This reverts commit 2101f43334d8b4b836ba16ea828b6266bcf82b49.

It seems like looking the file block up in the dirty bcache during the sync is problematic, because it could be an indirect block containing block pointers from writes happening concurrently with the sync.

Issue: KBFS-3832

### DIFF
--- a/go/kbfs/libkbfs/conflict_resolver.go
+++ b/go/kbfs/libkbfs/conflict_resolver.go
@@ -2242,8 +2242,8 @@ func (cr *ConflictResolver) computeActions(ctx context.Context,
 }
 
 // fileBlockMap maps latest merged block pointer to a map of final
-// merged name -> file block pointer.
-type fileBlockMap map[BlockPointer]map[string]BlockPointer
+// merged name -> file block.
+type fileBlockMap map[BlockPointer]map[string]*FileBlock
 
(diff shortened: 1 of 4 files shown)


jzila:
@strib but why would that be a problem?

strib:
> but why would that be a problem?

@jzila the sequence is something like this:

1. A bunch of writes happen on an indirect file.
2. `SyncAll` is called to start syncing the file.
3. Under a lock, the `SyncAll` saves the pointer P to the file, not the actual `FileBlock`.
4. Lock released, the `SyncAll` continues its thing.
5. A `Write` comes, and starts modifying the block pointed to by P in the dirty bcache.  Because a sync has started, this actually makes a deep copy of that block, and re-puts it in the dirty bcache.
6. The ongoing `SyncAll` now needs to ready and sync the block pointing to P.  In the current master code, it does so by using P to look up the block in the dirty bcache.  This is now the _new_ block containing temporary new pointers, that aren't ready to be synced yet.
7. When the `SyncAll` is done, it deletes the blocks from the dirty bcache that correspond to these new, temporary bcache.
8. Another `Write` comes in, trying to modify those new blocks, and fails because they have been deleted from the dirty bcache.

PR #15863, opened 2019-02-04
Stop systest services

patrickxb:


patrickxb:
i'm not sure i believe it, but i think CI just finished systests in 7 minutes, which would be amazing if true.

PR #15866, opened 2019-02-04
KBFS-3589 KBFS-3778 KBFS-3836: skip flakes

jzila:


go/kbfs/kbfsgit/runner_test.go
```diff
@@ -30,6 +30,7 @@ type testErrput struct {
 }
 
 func (te testErrput) Write(buf []byte) (int, error) {
+	te.t.Helper()
```
jakob223: Why this change now?

jzila: I was trying to comb through the logs and couldn't find the actual failing call because this one kept getting in the way.

PR #15867, opened 2019-02-04
folder_block_manager: don't make a gcop with revision of 0

strib:
If `getUnreferencedBlocks` doesn't find any revisions that are old enough to garbage collect, then stop QR.  Also skip over any bad GCOps made before this bug was fixed.

Issue: KBFS-3839


strib:
Merging with known CI flake (I didn't rebase onto the skip yet).

PR #15870, opened 2019-02-04
Add teams as a top level route on mobile

buoyad:
We have around 60 instances of routing to `tab:teamsTab` which were broken on mobile when we moved teams into settings. I made a ticket to go through and fix those but for the sake of simplicity this adds teams back to the top level routes so the navigation actions will still work. I tested some flows (nav from chat, new team, leave team) and they worked. r? @keybase/react-hackers


mmaxim:
What about Files, does that still work?

PR #15877, opened 2019-02-05
Remove scrollbars on Identify2 popup

zanderz:
@cecileboucheron said we could just take off the scrollbars, like the original Identify.


chrisnojima:
Let's talk to design

zanderz:
@keybase/react-hackers Cecile said we could remove the scrollbars, this is ready

shared/common-adapters/scroll-view.desktop.js
```diff
@@ -1,23 +1,27 @@
...
 }
 
-const styles = styleSheetCreate({
-  overflowAuto: platformStyles({
-    isElectron: {
-      overflow: 'auto',
-    },
-  }),
-})
+// not using a stylesheet because StylesCrossPlatform
```
chrisnojima: We should do the opposite. If it's missing add it to the desktop styles definition

zanderz: ok got it

zanderz:
@chrisnojima how about now?

chrisnojima:
fyi overflowx/y isnt' valid on RN, neither is auto

PR #15885, opened 2019-02-05
fix index for ident3 proof sort

buoyad:
@mlsteele is adding priorities from the service so eventually we'll sort on that. r? @keybase/react-hackers


buoyad:
@chrisnojima I had to fix a few things in the story

PR #15886, opened 2019-02-05
libkbfs: hide .kbfs_deleted_repos from listing

strib:
Autogit was listing `.kbfs_deleted_repos` as a repository name.  This causes issues, especially on the mobile GUI, where recursive lookups try to browse it on parent path lookups, although it is not a proper repo.


strib:
@songgao @jzila I think this will fix the mobile autogit issue.

PR #15889, opened 2019-02-05
libdokan: Re-enable noexec tests

taruti:
Hopefully @strib's recent fixes make the file copying into kbfs in tests less flaky.


taruti:
-count 100 runs out of memory, but testing with ```-test.run=NoEx -test.count=10``` in a loop works without flakes for >10 iterations.

taruti:
After 50x10 runs with no flakes merging this in.

PR #15891, opened 2019-02-05
friendlier ui for mobile only

xgess:
* add AccountModeEditable to GetWalletAccountLocal and GetWalletAccountsLocal rpcs. 
* add a couple little tests for it. this was more of a pain than i expected because during tests, the type of device can't be plucked from global state the way it can in a normal run. i don't love what i did here for this; i'm looping through device keys looking for the one that matches the active device. seems to work ok, but it's a little different from how we do this elsewhere.
* pull AccountModeEditable into wallet settings UI as mobileOnlyEditable. start using it instead of `Styles.isMobile` to determine whether or not the toggle should be disabled.

### DIFF
--- a/go/stellar/stellarsvc/service_test.go
+++ b/go/stellar/stellarsvc/service_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"strings"
 	"sync"
 	"testing"
@@ -1430,6 +1431,7 @@ func setupTestsWithSettings(t *testing.T, settings []usetting) ([]*TestContext, 
 		switch setting {
 		case usettingFull:
 		case usettingMobile:
+			os.Setenv("KEYBASE_APP_TYPE", string(libkb.MobileAppType))
 		case usettingPukless:
 			tc.Tp.DisableUpgradePerUserKey = true
 		}
@@ -1457,6 +1459,7 @@ func setupTestsWithSettings(t *testing.T, settings []usetting) ([]*TestContext, 
 	cleanup := func() {
 		for _, tc := range tcs {
 			tc.Cleanup()
+			os.Unsetenv("KEYBASE_APP_TYPE")
 		}
 	}
 	for i, tc := range tcs {
--- a/go/stellar/stellarsvc/service_test.go
+++ b/go/stellar/stellarsvc/service_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 	"strings"
 	"sync"
 	"testing"
@@ -1431,7 +1430,6 @@ func setupTestsWithSettings(t *testing.T, settings []usetting) ([]*TestContext, 
 		switch setting {
 		case usettingFull:
 		case usettingMobile:
-			os.Setenv("KEYBASE_APP_TYPE", string(libkb.MobileAppType))
 		case usettingPukless:
 			tc.Tp.DisableUpgradePerUserKey = true
 		}
@@ -1459,7 +1457,6 @@ func setupTestsWithSettings(t *testing.T, settings []usetting) ([]*TestContext, 
 	cleanup := func() {
 		for _, tc := range tcs {
 			tc.Cleanup()
-			os.Unsetenv("KEYBASE_APP_TYPE")
 		}
 	}
 	for i, tc := range tcs {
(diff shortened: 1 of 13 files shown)


go/stellar/stellarsvc/service_test.go
```diff
@@ -1430,6 +1431,7 @@ func setupTestsWithSettings(t *testing.T, settings []usetting) ([]*TestContext,
 		switch setting {
 		case usettingFull:
 		case usettingMobile:
+			os.Setenv("KEYBASE_APP_TYPE", string(libkb.MobileAppType))
```
zapu: I think there has to be a way not to do this. Maybe @patrickxb knows?

go/stellar/transform.go
```diff
@@ -448,13 +449,19 @@ func AccountDetailsToWalletAccountLocal(mctx libkb.MetaContext, accountID stella
 		return empty, err
 	}
 
+	isMobile := mctx.G().GetAppType() == libkb.MobileAppType
+	ctime, err := mctx.G().ActiveDevice.Ctime(mctx)
+	ctimeEpochSeconds := int64(ctime) / 1000
+	deviceAge := time.Since(time.Unix(ctimeEpochSeconds, 0)).Hours() / 24.0 // in days
```
zapu: I think you should use `G().Clock()` here instead of `time`

go/stellar/stellarsvc/service_test.go
xgess: please

chrisnojima: won't this break other tests that run at the same time

shared/wallets/wallet/settings/index.stories.js
```diff
@@ -74,13 +75,19 @@ const secondarySettingsProps = {
   name: 'some other account',
 }
 
+const withMobileOnlyEditable = {
```
chrisnojima: we don't need this object, below just do
```
.add('MobileOnlyEditable', () => <Settings {...secondarySettingsProps} mobileOnlyEditable={true} />)
```

shared/constants/types/wallets.js
```diff
@@ -212,12 +212,17 @@ export type Payment = I.RecordOf<_Payment>
...
+export type AccountMode = StellarRPCTypes.AccountMode
+export const accountModeNone = StellarRPCTypes.commonAccountMode.none
+
 export type _Account = {
   accountID: AccountID,
   balanceDescription: string,
   displayCurrency: Currency,
   isDefault: boolean,
   name: string,
+  accountMode: AccountMode,
```
chrisnojima: why is this plumbed through?

shared/constants/types/wallets.js
```diff
@@ -212,12 +212,17 @@ export type Payment = I.RecordOf<_Payment>
 
 export type Currency = I.RecordOf<_LocalCurrency>
 
+export type AccountMode = StellarRPCTypes.AccountMode
+export const accountModeNone = StellarRPCTypes.commonAccountMode.none
```
chrisnojima: we don't this this exported, we can just use StallerRPCTypes

shared/constants/types/wallets.js
xgess: it's not strictly necessary for this effort, but i was thinking about https://keybase.atlassian.net/browse/CORE-9900. exposing accountMode here i think would enable the frontend to drop a whole bunch of machinery for tracking mobileOnly as a separate construct. i'm happy to drop it if you don't think it's helpful, but the data is there in the RPC.

go/stellar/transform.go
```diff
@@ -448,13 +449,22 @@ func AccountDetailsToWalletAccountLocal(mctx libkb.MetaContext, accountID stella
...
-		BalanceDescription: balance,
-		Seqno:              details.Seqno,
-		AccountMode:        accountMode,
+		AccountID:           accountID,
+		IsDefault:           isPrimary,
+		Name:                accountName,
+		BalanceDescription:  balance,
+		Seqno:               details.Seqno,
+		AccountMode:         accountMode,
+		AccountModeEditable: isMobile && deviceAge > 7*24*time.Hour,
```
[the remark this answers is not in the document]
xgess: nope. per designs, you should not be able to set mobile only with a device that cannot unset it.

shared/constants/types/wallets.js
chrisnojima: lets drop it. we prefer to not plumb things through that don't do anything. if we need it later we can add it

go/stellar/stellarsvc/service_test.go
xgess: it's definitely possible, though i don't think it would in this specific case. anyway, i figured out another way to do it. it's a little different from how the rest of the app does it, but i think that's ok in this case.

PR #15905, opened 2019-02-06
FBO: allow moving conflict branches out of the way

jakob223:
This is still a little bit WIP, but I wanted an opinion from @strib before doing further work on it.


go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -722,6 +722,44 @@ func (fbo *folderBranchOps) isUnmergedLocked(lState *lockState) bool {
...
+	}
+
+	err = fbo.cr.clearConflictRecords()
+	if err != nil {
+		return err
+	}
+	lState := makeFBOLockState()
+	fbo.mdWriterLock.Lock(lState)
+	defer fbo.mdWriterLock.Unlock(lState)
+	fbo.setBranchIDLocked(lState, kbfsmd.NullBranchID)
```
strib: This doesn't quite work, because you need to revert the cached MD head to the merged version as well.  But since that changes the view of the data, we also have to tell the kernel to invalidate any caches they have for that old data.

Luckily, there's already a function that does all this for you: `folderBranchOps.unstageLocked()`.  Though it invokes `PruneBranch`, which forces the journal server to delete its current journal state for the TLF, instead of moving it away.  So maybe you can just massage that function to take in a parameter that lets it call `moveAway` instead of `PruneBranch` if we are trying to clear conflicts in this way?

Another point is that you have to unstage all the commits first before you move the journal out of the way, and keep the `mdWriterLock` the whole time; otherwise new reads and writes could slip in and cause races.  But `unstagedLocked` should do all that correctly.

There might be one other option however, that's both a little more foolproof than unstaging, and would also easily allow the user to navigate to the file view of the conflict branch.  We could just rename the TLF to something else, like we do when the TLF name becomes conflicted or finalized.  It would require more work though to expose a name like that, especially after a restart, but it might be the right play in the long run.  If you want to go this route, we can talk more about it tomorrow.  But I'm fine with just using `unstageLocked` for now, and we can do the TLF namespace stuff later.

go/kbfs/libkbfs/tlf_journal.go
```diff
@@ -2472,6 +2473,59 @@ func (j *tlfJournal) doResolveBranch(ctx context.Context,
...
+	defer j.journalLock.Unlock()
+	// j.dir is X/y-z. Current timestamp is T. Move directory X/y-z to X/z-T.bak
+	// and then rebuild it.
+	restDir, lastDir := filepath.Split(j.dir)
+	var ignored, tlfEncoded string
+	_, err := fmt.Sscanf(lastDir, "%s-%s", &ignored, &tlfEncoded)
+	if err != nil {
+		return err
+	}
+	newDirName := fmt.Sprintf("%s-%d.bak", tlfEncoded, time.Now().UnixNano())
```
strib: Should be using `j.config.Clock().Now()`.

jakob223:
TODO:
 - test this manually
 - finish writing that test

jakob223:
looked into the test failure further, it looks like the block server is not being informed that some blocks are superseded for some reason.

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -6570,8 +6622,13 @@ func (fbo *folderBranchOps) undoUnmergedMDUpdatesLocked(
 	return unmergedPtrs, nil
 }
 
+const (
+	doPruneBranches  = true
```
strib: Same thing here with a typed enum instead of plain bools.

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -6583,14 +6640,30 @@ func (fbo *folderBranchOps) unstageLocked(ctx context.Context,
...
+			if err != nil {
+				return err
+			}
+		} else {
+			return errJournalNotAvailable
+		}
 	}
 
 	// now go forward in time, if possible
+	// !isUnmerged is true -> isUnmerged is false -> we are merged but
```
strib: Is this leftover from debugging or a real question?

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -6656,7 +6729,7 @@ func (fbo *folderBranchOps) UnstageForTesting(
 			lState := makeFBOLockState()
 			c <- fbo.doMDWriteWithRetry(ctx, lState,
 				func(lState *kbfssync.LockState) error {
-					return fbo.unstageLocked(freshCtx, lState)
+					return fbo.unstageLocked(freshCtx, lState, true)
```
strib: Should use the named boolean (or enum).

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -7641,7 +7714,7 @@ func (fbo *folderBranchOps) unstageAfterFailedResolution(ctx context.Context,
 	ctx = newLinkedContext(ctx)
 	fbo.log.CWarningf(ctx, "Unstaging branch %s after a resolution failure",
 		fbo.unmergedBID)
-	return fbo.unstageLocked(ctx, lState)
+	return fbo.unstageLocked(ctx, lState, true)
```
strib: Should use the named boolean (or enum).

go/kbfs/libkbfs/interfaces.go
```diff
@@ -577,6 +577,9 @@ type KBFSOps interface {
 	NewNotificationChannel(
 		ctx context.Context, handle *tlfhandle.Handle,
 		convID chat1.ConversationID, channelName string)
+	// ClearConflictView moves the conflict views of the given TLF out of the
```
strib: "views" -> "view", there can be only one.

go/kbfs/libkbfs/kbfs_cr_test.go
```diff
@@ -772,6 +774,157 @@ func TestBasicCRFileConflict(t *testing.T) {
...
+	t.Log("User 2 looks up file a/b.")
+	rootNode2 := GetRootNodeOrBust(ctx, t, config2, name, tlf.Private)
+
+	kbfsOps2 := config2.KBFSOps()
+	dirA2, _, err := kbfsOps2.Lookup(ctx, rootNode2, "a")
+	require.NoError(t, err)
+	fileB2, _, err := kbfsOps2.Lookup(ctx, dirA2, "b")
+	require.NoError(t, err)
+
+	ctrl := gomock.NewController(t)
```
strib: Unused.

go/kbfs/libkbfs/kbfs_cr_test.go
```diff
@@ -772,6 +774,157 @@ func TestBasicCRFileConflict(t *testing.T) {
...
+	err = RestartCRForTesting(
+		libcontext.BackgroundContextWithCancellationDelayer(), config2,
+		rootNode2.GetFolderBranch())
+	require.NoError(t, err)
+
+	t.Log("Try to SyncFromServer on user 2.")
+	ctxWithSoonCancel, _ := context.WithTimeout(ctx, 50*time.Millisecond)
+	err = kbfsOps2.SyncFromServer(ctxWithSoonCancel,
+		rootNode2.GetFolderBranch(), nil)
+	require.Error(t, err, "context deadline exceeded")
```
strib: This doesn't do what you intended, I think.  This is just checking that `err` is not `nil`, and then if it is `nil`, it will print "context deadline exceeded" as a message.  If you look at the test output, it's actually failing because a single CR run isn't taking it out of conflict mode as expected.  So you can either check that `err.Error()` contains the expected string "Conflict resolution didn't take us out of staging", or (better) you can make that a typed error and use `require.IsType()` to check for it.

go/kbfs/libkbfs/kbfs_cr_test.go
```diff
@@ -772,6 +774,157 @@ func TestBasicCRFileConflict(t *testing.T) {
...
+
+	ops, ok := config2.KBFSOps().(*KBFSOpsStandard)
+	require.True(t, ok)
+	fbo := ops.getOpsNoAdd(ctx, rootNode2.GetFolderBranch())
+
+	t.Log("Write a bunch more files as user 2, creating more conflicts.")
+	for i := 0; i < 10; i++ {
```
strib: Is this getting it into the always bail out early mode?  Should you use `maxConflictResolutionAttempts` instead of 10?

go/kbfs/simplefs/simplefs.go
```diff
@@ -2251,6 +2251,17 @@ func (k *SimpleFS) SimpleFSSetFolderSyncConfig(
 	return err
 }
 
+// SimpleFSClearConflictState implements the SimpleFS interface.
+func (k *SimpleFS) SimpleFSClearConflictState(ctx context.Context,
+	path keybase1.Path) error {
+	// TODO: there's got to be a better way to get a TLF ID
```
strib: Yeah you can just copy what's in `getSyncConfig`, or make a helper function for getting the handle:

```go
	t, tlfName, _, _, err := remoteTlfAndPath(path)
	if err != nil {
		return err
	}
	tlfHandle, err := libkbfs.GetHandleFromFolderNameAndType(
		ctx, k.config.KBPKI(), k.config.MDOps(), k.config, tlfName, t)
	if err != nil {
		return err
	}
    tlfID := tlfHandle.TlfID()
```

I guess I was pretty lazy in the function you copied this from.

go/kbfs/libkbfs/folder_block_manager.go
```diff
@@ -383,6 +383,7 @@ func (fbm *folderBlockManager) archiveUnrefBlocks(md ReadOnlyRootMetadata) {
 	if md.MergedStatus() != kbfsmd.Merged {
 		return
 	}
+	// TODO: consider a different action in the event that a conflict is paused.
```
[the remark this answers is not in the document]
jakob223: I... don't remember.

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -714,6 +714,44 @@ func (fbo *folderBranchOps) isUnmergedLocked(lState *kbfssync.LockState) bool {
 	return fbo.unmergedBID != kbfsmd.NullBranchID
 }
 
+var errJournalNotAvailable = errors.New("could not get journal for TLF")
+
+// clearConflictView tells the journal to move any pending writes elsewhere,
+// resets the CR counter, and resets the FBO to have a synced view of the TLF.
+func (fbo *folderBranchOps) clearConflictView(baseCtx context.Context) (
```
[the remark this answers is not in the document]
jakob223: because we have to make another ctx later to call `unstageLocked` on

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -714,6 +714,44 @@ func (fbo *folderBranchOps) isUnmergedLocked(lState *kbfssync.LockState) bool {
...
+		fbo.log.CDebugf(baseCtx, "Done with clearConflictView: %+v", err)
+	}()
+
+	lState := makeFBOLockState()
+	fbo.mdWriterLock.Lock(lState)
+	defer fbo.mdWriterLock.Unlock(lState)
+
+	var ctx context.Context
+	ctx = fbo.ctxWithFBOID(baseCtx)
+	ctx, err = libcontext.NewContextWithCancellationDelayer(ctx)
```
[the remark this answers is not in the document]
jakob223: ISTR there being something in `unstageLocked` that errored if it didn't get a context with a cancellation delayer.

go/kbfs/libkbfs/folder_branch_ops.go
jakob223: Left over from debugging, whoops

go/kbfs/libkbfs/folder_branch_ops.go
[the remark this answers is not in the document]
strib: True, I guess this just turns on the possibility of cancellation, but doesn't actually enable it.

That said, the line right above it, `ctxWithFBOID()`, already returns a ctx that's wrapped in `NewContextWithCancellationDelayer`, so having it here again definitely seems unnecessary.

go/kbfs/libkbfs/folder_branch_ops.go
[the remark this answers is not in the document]
strib: I don't follow.  You never use `baseCtx` again after creating `ctx` below, so you could easily reuse the same name.  But also, I don't remember why you need to make a new context at all -- `unstageLocked` isn't being invoked in a goroutine or anything, so it seems valid to pass in the same ctx.  (Sorry if this is something I told you to do earlier, I legitimately can't remember, so please refresh me if there's a reason.)

go/kbfs/libkbfs/folder_branch_ops.go
strib: Actually I recently merged something that removes this whole function (`unstageAfterFailedResolution`) entirely, so it might be easier if you revert this change here and then rebase on master.

go/kbfs/libkbfs/folder_branch_ops.go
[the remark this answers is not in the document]
jakob223: When I get rid of this line, I get ```			            	Context doesn't have critical awareness or CtxCancellationDelayerKey already exists in ctx but is not of type *cancellationDelayer
```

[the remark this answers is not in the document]
Comment (strib) [reply go/kbfs/libkbfs/folder_branch_ops.go]: Ah I was looking at `newCtxWithFBOID()`, not just `ctxWithFBOID()`.  Sorry.  But still I think you can just use the context passed into the function without making a new one here.

File: go/kbfs/libkbfs/folder_branch_ops.go (hunk 13 above)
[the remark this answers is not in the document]
Comment (jakob223) [reply go/kbfs/libkbfs/folder_branch_ops.go]: Ah, I see. Changed.

File: go/kbfs/libkbfs/folder_branch_ops.go [hunk 15]
```diff
@@ -719,23 +728,22 @@ var errJournalNotAvailable = errors.New("could not get journal for TLF")
...
+		fbo.log.CDebugf(ctx, "Done with clearConflictView: %+v", err)
 	}()
 
 	lState := makeFBOLockState()
 	fbo.mdWriterLock.Lock(lState)
 	defer fbo.mdWriterLock.Unlock(lState)
 
-	var ctx context.Context
-	ctx = fbo.ctxWithFBOID(baseCtx)
+	ctx = fbo.ctxWithFBOID(ctx)
```
strib: I don't think this is needed -- the `ctx` being passed in should already have tags from where it was created (i.e. SFSID from SimpleFS).  Adding an FBOID (usually used for background tasks within `folderBranchOps`) doesn't add much and just pollutes the logs I think.

go/kbfs/libkbfs/kbfs_cr_test.go
jakob223: It's not getting this error, though - it's getting a context deadline exceeded, since the function called will call CR over and over again until it succeeds.

strib: Huh?  Looks like Jenkins passed on the new code that doesn't do this check.

jakob223: But it failed locally for me after I made the change you suggested below.

jakob223: Oh, it's nondeterministic. It sometmes gets one and sometimes the other.

jakob223: I'm gonna extend the timeout by a factor of 2 and just require that there be an error.

go/kbfs/libkbfs/folder_branch_ops.go
jakob223: Hmm, looks like I added this because the context needs to be replayable, but I see that it also has unintended consequences. What's a better way to get a replayable context? What is a replayable context?

go/kbfs/libkbfs/kbfs_cr_test.go
strib: Really?  The version you have in github (commit 4a11e67c9) succeeds 100x in a row for me.  Maybe you can share logs with me?  I'm not sure how you'd be seeing the deadline exceeded error.  `SyncFromServer` only retries if it got out of CR mode somehow, or was never in it (as was the case before when the callbacks were being nil'd out).

go/kbfs/libkbfs/folder_branch_ops.go
strib: Really the context you pass in from the test should already be configured correctly.  Looks like you're using `context.Background()` in the test -- try `ctx` instead.

go/kbfs/libkbfs/kbfs_cr_test.go
jakob223: Huh, maybe I changed some other line - not able to produce that behavior anymore.

PR #15907, opened 2019-02-06
WIP: tweak details page to overcome recentPaymentsReceived race

zanderz:
This is the case again when you send a payment and then click it for details right away. This change makes a bit more sense logically but it really looks like even if the details are requested a second time, they don't come from the service (specifically, a public message). Still looking into that.

### DIFF
--- a/shared/wallets/transaction-details/index.js
+++ b/shared/wallets/transaction-details/index.js
@@ -431,13 +431,10 @@ class LoadTransactionDetails extends React.Component<Props> {
   }
   componentDidUpdate(prevProps: Props) {
     // An erased transaction ID likely means the payment was updated,
-    // which means details need to be retrieved again
-    if (
-      (!this.props.transactionID || !this.props.senderAccountID) &&
-      prevProps.transactionID &&
-      prevProps.senderAccountID &&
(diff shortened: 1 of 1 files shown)


zanderz:
@keybase/react-hackers I'd like to get this in and then concentrate on the core side

shared/wallets/transaction-details/index.js
```diff
@@ -431,12 +431,13 @@ class LoadTransactionDetails extends React.Component<Props> {
...
+    // which means details need to be retrieved again.
+    // Also, recentPaymentsReceived can race with paymentDetailReceived
+    // so that loading finishes before transactionID is ever seen here.
     // which means details need to be retrieved again
     if (
-      (!this.props.transactionID || !this.props.senderAccountID) &&
-      prevProps.transactionID &&
-      prevProps.senderAccountID &&
-      !this.props.loading
+      (!this.props.transactionID && prevProps.transactionID) ||
```
buoyad: Should this one also `&& !this.props.loading`?

zanderz: prettier did that! I added parens though and it seems ok now

PR #15908, opened 2019-02-06
 send-link-to-chat on mobile

songgao:
(unlike what I thought before, we already have FloatingPicker on mobile, so just use that)

### DIFF
--- a/shared/fs/common/path-item-action/layout.js
+++ b/shared/fs/common/path-item-action/layout.js
@@ -52,7 +52,7 @@ const getRawLayout = (path: Types.Path, pathItem: Types.PathItem): Layout => {
         ...empty,
         copyPath: true,
         ignoreTlf: true,
-        sendLinkToChat: false, // TODO enable mobile // desktop uses separate button
+        sendLinkToChat: isMobile, // desktop uses separate button
         showInSystemFileManager: !isMobile,
       }
     default:
@@ -68,7 +68,7 @@ const getRawLayout = (path: Types.Path, pathItem: Types.PathItem): Layout => {
         // share menu items
         // eslint-disable-next-line sort-keys
         sendAttachmentToChat: false, // TODO enable mobile pathItem.type === 'file' && isMobile, // desktop uses separate button
-        sendLinkToChat: false, // TODO enable mobile // desktop uses separate button
+        sendLinkToChat: isMobile, // desktop uses separate button
         sendToOtherApp: pathItem.type === 'file' && isMobile,
       }
   }
--- a/shared/fs/common/path-item-action/layout.js
+++ b/shared/fs/common/path-item-action/layout.js
@@ -2,6 +2,7 @@
 import * as Types from '../../../constants/types/fs'
 import * as Constants from '../../../constants/fs'
 import {isMobile, isIOS} from '../../../constants/platform'
+import * as Flow from '../../../util/flow'
 
 export type Layout = {
   copyPath: boolean,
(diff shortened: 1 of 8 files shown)


shared/fs/send-link-to-chat/index.js
```diff
@@ -58,110 +59,242 @@ const who = (props: Props) => {
...
+      <Kb.CopyText text={props.pathTextToCopy} multiline={Styles.isMobile} />
     </Kb.Box2>
-    <Kb.Box2 direction="horizontal" centerChildren={true} style={styles.footer} gap="tiny">
-      <Kb.Button type="Secondary" label="Cancel" onClick={props.onCancel} />
-      {props.conversation.type !== 'none' && (
-        <Kb.Button type="Primary" label="Send in conversation" disabled={!props.send} onClick={props.send} />
-      )}
+    {props.conversation.type !== 'none' && (
+      <Kb.Text type="BodySmall" style={styles.onlyWhoGetAccess}>
+        Only {who(props)} will get access to the file.
```
buoyad: I was confused by this because "will get access" sounds like this grants them access, when they already have access. Could it change to "Only {...} has/have access to this file."?

shared/fs/common/path-item-action/layout.js
```diff
@@ -52,7 +52,7 @@ const getRawLayout = (path: Types.Path, pathItem: Types.PathItem): Layout => {
         ...empty,
         copyPath: true,
         ignoreTlf: true,
-        sendLinkToChat: false, // TODO enable mobile // desktop uses separate button
+        sendLinkToChat: isMobile, // desktop uses separate button
```
buoyad: This probably shouldn't show when it leads to the 'Copy link' screen, since `copyPath` does that with one fewer step.

shared/fs/send-link-to-chat/container.js
```diff
@@ -115,14 +115,21 @@ const mergeProps = (stateProps, {onCancel, _send, _selectChannel}, ownProps) =>
...
+
   return {
     conversation: {
       channels,
       name: elems[2],
       selectChannel: convID => _selectChannel(convID),
-      selectedChannelname: (
-        channels.find(({convID}) => convID === stateProps._sendLinkToChat.convID) || {channelname: null}
-      ).channelname,
+      selectedChannel:
```
buoyad: So if I have a folder with someone but not a conversation, the 'Send in conversation' button is grayed out without explanation. I think ideally it could create a conv and send the message into it (the "reply privately" case in `createConversation` chat actions is similar).

shared/fs/send-link-to-chat/index.js
```diff
@@ -58,110 +59,242 @@ const who = (props: Props) => {
...
+  state = {visible: false}
+  _show = () => this.setState({visible: true})
+  _hide = () => this.setState({visible: false})
+  _select = convIDStr => {
+    this.props.selectChannel(ChatTypes.stringToConversationIDKey(convIDStr))
+  }
+  render() {
+    return (
+      <>
+        <Kb.ClickableBox onClick={this._show} style={styles.dropdown}>
```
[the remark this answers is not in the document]
songgao: Oh! Didn't know that was exported. Will do thanks!

shared/fs/send-link-to-chat/index.js
songgao: Sounds good with me. @cecileboucheron how does that sound?

shared/fs/send-link-to-chat/container.js
songgao: Sounds good and thanks for the example! I'll do that in a separate PR.

songgao:
@buoyad ready for another look thanks!

PR #15914, opened 2019-02-06
md_ops: wait for MDServer to be initialized

strib:
In the init code, `MDOps` is set before `MDServer`, and so there's a race condition where something (e.g., the disk cache initialization code that triggers when the keybase service connection is established) can try to use the MDServer before it exists.

So just add a simple sleep for the very rare case where this could happen. This doesn't seem like it's worth the code complexity of something that would get notified exactly when the `MDServer` is set.

Issue: KBFS-3587

### DIFF
--- a/go/kbfs/libkbfs/md_ops.go
+++ b/go/kbfs/libkbfs/md_ops.go
@@ -335,6 +335,30 @@ func (md *MDOpsStandard) startOfValidatedChainForLeaf(
 	return min
 }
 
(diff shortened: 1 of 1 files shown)


strib:
> Sweet, this could be responsible for some of the kbfsdocker flakes.

Not sure about that -- the symptom here is a client-side panic, not a communication failure to the server.

jzila:
Hmm, fair enough.

PR #15919, opened 2019-02-06
block_retrieval_worker: don't act on stop-if-full reqs when full

strib:
I noticed that prefetching recent files keeps fetching blocks long after the disk cache is full.  It seemed to be a vicious cycle of:

* The prefetcher queues some blocks requests before the cache is full.
* The cache gets full.
* A worker picks up the queued requests and fetches the blocks.
* The blocks get put in the working-set cache, which causes enough eviction to get the cache down below 99% full.
* The prefetcher requests more blocks, since the cache is no longer full.
* Ad infinitum.

So this PR simply rejects stop-if-full requests in the worker itself, if the cache is already full.

Issue: KBFS-3803

### DIFF
--- a/go/kbfs/libkbfs/block_retrieval_worker_test.go
+++ b/go/kbfs/libkbfs/block_retrieval_worker_test.go
@@ -385,3 +385,54 @@ func TestBlockRetrievalWorkerPrefetchedPriorityElevation(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, testBlock1, block1)
 }
+
+func TestBlockRetrievalWorkerStopIfFull(t *testing.T) {
+	ctx, cancel := context.WithTimeout(
+		context.Background(), individualTestTimeout)
+	defer cancel()
+	dbc, dbcConfig := initDiskBlockCacheTest(t)
+	defer dbc.Shutdown(ctx)
+
+	bg := newFakeBlockGetter(false)
+	q := newBlockRetrievalQueue(
+		1, 1, 0, newTestBlockRetrievalConfig(t, bg, dbc))
+	require.NotNil(t, q)
+	<-q.TogglePrefetcher(false, nil, nil)
+	defer q.Shutdown()
+
+	ptr := makeRandomBlockPointer(t)
+	syncCache := dbc.syncCache
+	workingCache := dbc.workingSetCache
+
+	t.Log("Set the cache maximum bytes to the current total.")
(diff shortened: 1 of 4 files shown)


jzila:
Why though? We still want evictions to happen on the working set cache even if it's full, right?

jzila:
Oh nvmind, this is just for `stop-if-full` requests.

go/kbfs/libkbfs/block_retrieval_worker_test.go
```diff
@@ -385,3 +385,54 @@ func TestBlockRetrievalWorkerPrefetchedPriorityElevation(t *testing.T) {
...
+	t.Log("Request with stop-if-full, when full")
+	testBlock := &FileBlock{}
+	req := q.Request(
+		ctx, 1, makeKMD(), ptr, testBlock, NoCacheEntry,
+		BlockRequestPrefetchUntilFull)
+	select {
+	case err := <-req:
+		require.IsType(t, DiskCacheTooFullForBlockError{}, err)
+	case <-ctx.Done():
+		require.NoError(t, ctx.Err())
```
jzila: Shouldn't this be `t.Fatal()`?

go/kbfs/libkbfs/block_retrieval_worker_test.go
```diff
@@ -385,3 +385,54 @@ func TestBlockRetrievalWorkerPrefetchedPriorityElevation(t *testing.T) {
...
+	req = q.Request(
+		ctx, 1, makeKMD(), ptr, testBlock, NoCacheEntry,
+		BlockRequestWithPrefetch)
+	<-startCh
+	continueCh <- nil
+	select {
+	case err := <-req:
+		require.NoError(t, err)
+	case <-ctx.Done():
+		require.NoError(t, ctx.Err())
```
jzila: Same here.

go/kbfs/libkbfs/block_retrieval_worker_test.go
strib: I've come to prefer `require.NoError`, since it prints out the full stack of the caller and makes it easier to debug.  Do you have a strong preference for `t.Fatal`?

jzila: I guess it's fine, it just seems to imply that we aren't expecting an error where we actually are. Did you consider `require.FailNow(t)`?

strib: Ok, used `FailNow()` instead.  Thanks!

PR #15928, opened 2019-02-06
Rename JournalServer to JournalManager

jakob223:
Expecting `.kbfs_status` change might cause a docker test or 2 to fail.


go/kbfs/libkbfs/tlf_journal.go
```diff
@@ -101,8 +101,8 @@ const (
 	// Maximum number of blocks to delete from the local saved block
 	// journal at a time while holding the lock.
 	maxSavedBlockRemovalsAtATime = uint64(500)
-	// How often to check the server for conflicts while flushing.
-	tlfJournalServerMDCheckInterval = 1 * time.Minute
```
strib: This one should stay as it was before.  `tlfJournal` is the prefix, and `ServerMDCheckInterval` is the real constant name -- it's checking the mdserver for whether there are new merged MDs available.

jakob223: lol, I looked at this briefly and was confused by it but didn't think about it long enough to realize this was the meaning.

PR #15935, opened 2019-02-07
libfs: pretend non-existent TLF is empty on iteam write error

strib:
After the switch to implicit teams, a reader of a folder gets a new kind of error when trying to initialize a TLF that it doesn't have write access for.  Pretend it is empty in that case instead.

### DIFF
--- a/go/kbfs/libfs/tlf.go
+++ b/go/kbfs/libfs/tlf.go
@@ -58,6 +58,13 @@ func FilterTLFEarlyExitError(ctx context.Context, err error, log logger.Logger, 
 			name)
 		return true, nil
 
+	case libkbfs.NonExistentTeamForHandleError:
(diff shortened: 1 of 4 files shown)


go/kbfs/libfs/tlf.go
```diff
@@ -55,7 +58,24 @@ func FilterTLFEarlyExitError(ctx context.Context, err error, log logger.Logger,
 			name)
 		return true, nil
 
+	case libkbfs.NonExistentTeamForHandleError:
```
taruti: Why not?
```case kbfsmd.ServerErrorWriteAccess, libkbfs.NonExistentTeamForHandleError:```

strib: Good point, I always forget you can do that with `(type)` switches as long as you don't actually use the typed variable.  Magic!

PR #15937, opened 2019-02-07
use share menu for share button in filepreview on mobile

songgao:


songgao:
I think so; will do!

songgao:
ready again!

shared/fs/common/index.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-export {default as PathItemAction} from './path-item-action/container'
+export {default as PathItemAction} from './path-item-action/path-item-action-container'
```
jzila: can't this just be `container.js` again?

shared/fs/common/index.stories.js
```diff
@@ -6,7 +6,7 @@ import * as Sb from '../../stories/storybook'
 import * as Styles from '../../styles'
 import * as Types from '../../constants/types/fs'
 import * as Kb from '../../common-adapters'
-import PathItemAction from './path-item-action'
+import PathItemAction from './path-item-action/path-item-action'
```
jzila: can't this be `index` again?

songgao:
@jzila we'd need to have a HOC I think, because we need to pass in `onClick` and `setRef`. It feel cleaner this way, and way easier to flow-type 😅FWIW it's also the pattern used in List.

PR #15938, opened 2019-02-07
Round of fixes for GUI PGP generation

zapu:
- Only invoke tsec passphrase prompt from `PGPKeyGen` engine when user asks to push private key, not always.
- Change screens:
   - `Are you sure you want to drop your Pgp identity?` -> `Are you sure you want to drop your PGP key?`
   - Mention passphrase entry in `Store encrypted private key on Keybase's server` label.


zapu:
On second thought, GUI flow does not export to gpg at all. So there's nothing to encrypt. Let's keep it this way.

zapu:
@mlsteele i clicked on the wrong pr because of bad internet and pressing back button too many times.

please take a look and ill revert if needed

mlsteele:
👍

PR #15939, opened 2019-02-07
Enhance prove interface

mlsteele:


protocol/avdl/keybase1/prove_ui.avdl
```diff
@@ -1,18 +1,58 @@
...
+    union {null, ProveParameters} parameters);
+
+  // Ask the user whether they have posted the proof.
+  // Returns whether to keep checking.
+  // Not used by parameterized proofs
   boolean okToCheck(int sessionID, string name, int attempt);
+
+  // The proof is being checked in the background.
+  // Used by parameterized proofs only
+  void checking(int sessionID, string name);
```
mlsteele: @chrisnojima The only changes here are to add this `checking` alternative to `okToCheck` and `ProveParameters` arguments. My hope is that the this `prove{,_common,_ui}.avdl` protocol is now sufficient for gui work on parameterized proofs.

go/engine/prove.go
```diff
@@ -313,20 +320,49 @@ func (p *Prove) promptPostedLoop(m libkb.MetaContext) (err error) {
 	return err
 }
 
+// Poll forever until the proof succeeds.
```
patrickxb: forever?

mlsteele: The context could get canceled. An hour?

PR #15944, opened 2019-02-07
 attach local metadata to incremental sync convs

mmaxim:
The "local metadata" in the inbox contains the last known value of things derived from "unboxing" the inbox items. If we don't pass up this value during sync, then the frontend will clear out the relevant data while waiting for the unboxed version to make it up.


mmaxim:
I think so.

PR #15945, opened 2019-02-07
move badger notify after gregor resync

mlsteele:
`Resync` should happen after `serverSync` so that `Resync` gets the latest gregor state.

### DIFF
--- a/go/service/gregor.go
+++ b/go/service/gregor.go
@@ -809,11 +809,11 @@ func (g *gregorHandler) OnConnect(ctx context.Context, conn *rpc.Connection,
 		return fmt.Errorf("error running state sync: %s", err)
 	}
 
-	// Update bading from gregor.
+	// Update badging from gregor.
 	if g.badger != nil {
 		state, err := gcli.StateMachineState(ctx, nil, false)
 		if err != nil {
-			g.chatLog.Debug(ctx, "unable to get state: %v", err)
+			g.chatLog.Debug(ctx, "unable to get gregor state for badging: %v", err)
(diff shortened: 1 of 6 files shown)


mmaxim:
Ok so I think we might need to take a different tack here. I recently that down below the sync in order to fix chat badging, but I see now that it breaks the Gregor based badging. Ideally, we should probably run the chat update where it is now, and then do the Gregor update after that (which probably means running something other than `Resync`.

mlsteele:
Split resync in two. Now it's
1. chat badging
2. Syncer.Connected
3. gregor badging

A side effect is that OnConnected will now send not 1 but 2 badge broadcasts. Can silence the first if that's a problem.

go/service/gregor.go
```diff
@@ -804,6 +809,17 @@ func (g *gregorHandler) OnConnect(ctx context.Context, conn *rpc.Connection,
 		return fmt.Errorf("error running state sync: %s", err)
 	}
 
+	// Update bading from gregor.
```
mmaxim: Typo

go/service/gregor.go
```diff
@@ -804,6 +809,17 @@ func (g *gregorHandler) OnConnect(ctx context.Context, conn *rpc.Connection,
...
+	if g.badger != nil {
+		state, err := gcli.StateMachineState(ctx, nil, false)
+		if err != nil {
+			g.chatLog.Debug(ctx, "unable to get state: %v", err)
```
mmaxim: Can this debug line be a little more specific?

PR #15952, opened 2019-02-07
Add extra data to the favorites cache so it can serve queries from the app

jakob223:
TODO:
  - bust cache on new TLF creation
  - write tests
  - implement behavior for when the favorites cache is disabled
  - plumb through `TeamID`s in a few more places (if we actually need them?)


jakob223:
Putting this up for review because I want feedback/help on
 - where to get the `TeamID`s in various places and whether we actually need them
 - testing strategies
 - places I forgot about where we need to bust the cache

go/kbfs/libkbfs/tlf_handle.go
```diff
@@ -412,12 +412,28 @@ func (h *TlfHandle) ToFavorite() Favorite {
...
+// Favorites calls.
+func (h *TlfHandle) FavoriteData() favoriteData {
+	return favoriteData{
+		Name:       string(h.GetCanonicalName()),
+		FolderType: h.Type().FolderType(),
+		// TODO: verify this conversion
+		ID:      keybase1.TLFID(h.tlfID.String()),
+		Private: h.Type() != tlf.Public,
+		// TODO: find this team ID if we care about it
+		TeamID:       keybase1.TeamID(0),
```
strib: For TLFs backed by an implicit or regular team, you should be able to get the TeamID easily with something like:

```go
fd := favoriteData{...}
if h.IsBackedByTeam() {
    fd.TeamID = h.FirstResolvedWriter().AsTeamOrBust()
}
```

If it's not backed by an implicit team, it must be an old folder that we've failed to convert, or an old public folder (haven't converted those yet).  In those cases, I doubt we need a team ID at all.  I'd be surprised if we do.

go/kbfs/libkbfs/favorites.go
```diff
@@ -28,11 +28,11 @@ const (
 	favoritesCacheExpirationTime     = time.Hour * 24 * 7 // one week
 	kbfsFavoritesCacheSubfolder      = "kbfs_favorites"
 	favoritesDiskCacheFilename       = "kbfsFavorites.leveldb"
-	favoritesDiskCacheVersion        = 1
-	favoritesDiskCacheStorageVersion = 1
+	favoritesDiskCacheVersion        = 2
```
strib: I don't _think_ you need a new storage version for these changes.  It seems like you're only adding data/fields, and in that case msgpack does the right thing and will just initialize those fields as empty until you overwrite them.

go/kbfs/libkbfs/favorites.go
```diff
@@ -322,18 +370,41 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
...
 			}
 			session, err := f.config.KBPKI().GetCurrentSession(req.ctx)
 			if err == nil {
 				// Add favorites for the current user, that cannot be deleted.
-				f.cache[Favorite{string(session.Name), tlf.Private}] = true
-				f.cache[Favorite{string(session.Name), tlf.Public}] = true
+				f.favCache[Favorite{string(session.Name), tlf.Private}] = favoriteData{
+					Name:       string(session.Name),
+					FolderType: tlf.Private.FolderType(),
+					// TODO: get the TLF ID
```
strib: Maybe you can have special methods that set these IDs and their team IDs from inside the loggedIn method that initializes the disk caches.

go/kbfs/libkbfs/keybase_daemon_local.go
```diff
@@ -890,14 +890,23 @@ func (k *KeybaseDaemonLocal) FavoriteDelete(
...
+	ctx context.Context, sessionID int) (keybase1.FavoritesResult, error) {
 	if err := checkContext(ctx); err != nil {
-		return nil, err
+		return keybase1.FavoritesResult{}, err
 	}
 
 	k.lock.Lock()
 	defer k.lock.Unlock()
-	return k.favoriteStore.FavoriteList(k.currentUID)
+	// TODO: determine if this is okay
```
strib: This is fine with me.  This is only used for testing really, so unless we write a test that relies on ignored/new folders, we should be ok here.

go/kbfs/libkbfs/tlf_handle.go
```diff
@@ -412,12 +412,28 @@ func (h *TlfHandle) ToFavorite() Favorite {
 	}
 }
 
+// FavoriteData converts a TlfHandle into favoriteData, suitable for
+// Favorites calls.
+func (h *TlfHandle) FavoriteData() favoriteData {
+	return favoriteData{
+		Name:       string(h.GetCanonicalName()),
+		FolderType: h.Type().FolderType(),
+		// TODO: verify this conversion
```
strib: :+1:

go/kbfs/libkbfs/favorites.go
```diff
@@ -273,26 +300,47 @@ func (f *Favorites) closeReq(req *favReq, err error) {
...
+	typeFavorite
+)
+
+func toFolder(fav Favorite, data favoriteData,
+	favType favoriteType) keybase1.Folder {
+	return keybase1.Folder{
+		Name:    fav.Name,
+		Private: data.Private,
+		// TODO: figure out what the deal is here
+		NotificationsOn: favType == typeFavorite,
```
strib: Oof, `NotificationsOn` is super old.  I'm pretty sure it's currently unused in the code.  You can probably just remove it from the protocol.  I don't even remember what it was for.

jakob223:
TODO:

- [ ] write a test
- [ ] rebase
- [ ] cache invalidation
- [ ] figure out what to do about resets
- [ ] get the ID and TeamID into the favorites for home TLFs

jakob223:
I think this is ~complete except for the server bits and one place I don't know where to get IDs from.

go/kbfs/libkbfs/favorites.go
strib: Do you mean to resolve this conversation without changing this back to 1?

go/kbfs/libkbfs/favorites.go
```diff
@@ -79,12 +81,41 @@ type favReq struct {
...
+	Private      bool
+	TeamID       keybase1.TeamID
+	ResetMembers []keybase1.User
+}
+
+func favoriteDataFrom(folder keybase1.Folder) favoriteData {
+	return favoriteData{
+		Name:       folder.Name,
+		FolderType: folder.FolderType,
+		// TODO: get ID??
```
strib: Why is there a TLF ID in `favoriteData` at all?  Doesn't seem like it's used anywhere.

go/kbfs/libkbfs/favorites.go
```diff
@@ -79,12 +81,41 @@ type favReq struct {
...
+}
+
 // Favorites manages a user's favorite list.
 type Favorites struct {
 	config   Config
 	disabled bool
 	log      logger.Logger
 
+	// homeTLFInfo stores the IDs for the logged-in user's home TLFs
+	homeTLFInfo homeTLFInfo
```
strib: Looks like this doesn't get set yet?  I'm guessing you'll need another request type or channel for that, or just use a lock maybe.

go/kbfs/libkbfs/favorites_test.go
```diff
@@ -254,6 +254,39 @@ func TestFavoritesControlUserHistory(t *testing.T) {
...
+	config.mockCodec.EXPECT().Encode(gomock.Any()).Return(nil, nil).Times(2)
+	config.mockKbs.EXPECT().EncryptFavorites(gomock.Any(),
+		gomock.Any()).Return(nil, nil)
+
+	// Require that we correctly return the data inserted into the cache by
+	// the server.
+	res2, err := f.GetAll(ctx)
+	require.NoError(t, err)
+	require.Equal(t, res.IgnoredFolders, res2.IgnoredFolders)
+	require.Len(t, res2.NewFolders, 0)
```
strib: We should get some new and regular favorites in there too, just to test that there are no copy/paste bugs mixing up folders from different lists.

go/kbfs/libkbfs/favorites.go
jakob223: no, I thought I'd changed it back..

jakob223: Oh, this version needs to stay incremented, because the cache has changed from a `map[Favorite]bool` to a `map[Favorite]favoriteData`.

go/kbfs/libkbfs/favorites.go
jakob223: Not sure.. I removed it.

strib:
@jakob223 looks like you accidentally committed an errant copy-pasted line into `folder_branch_ops.go`.  Also, it would be great if you can rebase to resolve the rpc-gen conflict.

jakob223:
@strib whoops... fixed.

go/kbfs/libkbfs/tlf_handle.go
```diff
@@ -412,12 +412,30 @@ func (h *TlfHandle) ToFavorite() Favorite {
...
+func (h *TlfHandle) FavoriteData() FavoriteData {
+	fd := FavoriteData{
+		Name:         string(h.GetCanonicalName()),
+		FolderType:   h.Type().FolderType(),
+		Private:      h.Type() != tlf.Public,
+		ResetMembers: []keybase1.User{},
+	}
+	if h.IsBackedByTeam() {
+		var teamID keybase1.TeamID
+		teamID = h.FirstResolvedWriter().AsTeamOrBust()
```
strib: Why not `teamID :=`?

go/kbfs/libkbfs/favorites.go
```diff
@@ -273,26 +311,36 @@ func (f *Favorites) closeReq(req *favReq, err error) {
...
-	for newFav := range f.cache {
-		if !oldCache[newFav] {
+	for newFav := range f.favCache {
+		if _, present := oldCache[newFav]; !present {
 			f.config.KBFSOps().RefreshEditHistory(newFav)
 		}
 	}
 }
 
+func toFolder(fav Favorite, data FavoriteData) keybase1.Folder {
```
strib: Might want to put this as a method on `Favorite`, but if not, maybe rename it to be more specific since this is in a package with a large namespace.  Maybe `favoriteToFolder`?

go/kbfs/libkbfs/favorites.go
```diff
@@ -607,3 +712,70 @@ func (f *Favorites) Get(ctx context.Context) ([]Favorite, error) {
 	}
 	return <-favChan, nil
 }
+
+// setHomeTLFInfo should be called when a new user logs in so that their home
+// TLFs can be returned as favorites.
+func (f *Favorites) setHomeTLFInfo(ctx context.Context, info homeTLFInfo) {
```
strib: I don't see any callers of this function.

go/kbfs/libkbfs/tlf_handle.go
jakob223: I don't recall. Changed.

jakob223:
The concern about the first access when offline being extra slow seems valid. My first idea is to throw another property on the `Favorites` indicating whether the last refresh failed, and then try to refresh at startup, and whenever that variable is set use a short timeout on the refreshing or maybe do it asynchronously.

I kinda wish we had some system-wide "are we offline" variable that we could set sometimes and then have stuff just work fast without waiting for things to time out repeatedly.

strib:
> The concern about the first access when offline being extra slow seems valid. My first idea is to throw another property on the Favorites indicating whether the last refresh failed, and then try to refresh at startup, and whenever that variable is set use a short timeout on the refreshing or maybe do it asynchronously.

I don't think having that extra variable helps much, since any number of events could have come in while we offline and never got delivered.  We basically have to always attempt to refresh it after a restart or a service disconnect I think.  So I think going with a short timeout is the way to go.  Maybe it can just always use a short timeout, and if that fires return the cache value, but continue going in the background and refreshing it whenever it finally goes through.  Quotas use already use that approach, for example:

https://github.com/keybase/client/blob/fa879a0efe186b6b17c8046c763647d45e4ecf9b/go/kbfs/libkbfs/quota_usage.go#L185-L214

> I kinda wish we had some system-wide "are we offline" variable that we could set sometimes and then have stuff just work fast without waiting for things to time out repeatedly.

This is super tricky given that there are multiple different servers involved, and we could be disconnected from one but not all of them.

jakob223:
@strib I made KBFS-3938 for this improvement, assigned it to myself and put it in next sprint.
Also ready for a final (fingers crossed) review. Will squash this into a reasonable set of commits after, since I don't want to make your life harder w/r/t reviewing only new code.

go/kbfs/libkbfs/interfaces.go
```diff
@@ -348,6 +348,9 @@ type KBFSOps interface {
 	// the local cache.  Idempotent, so it succeeds even if the folder
 	// isn't favorited.
 	DeleteFavorite(ctx context.Context, fav Favorite) error
+	// SetFavoritesHomeTLFInfo sets the hom TLF TeamIDs to initialize the
```
strib: hom -> home

go/kbfs/libkbfs/keybase_service_util.go
```diff
@@ -59,6 +60,13 @@ func setupDiskBlockCache(ctx context.Context, config Config, username string) {
 				"for disk block cache: %+v", err)
 		}
 	}
+
+	// Send to Favorites cache
+	info := homeTLFInfo{
+		PublicTeamID:  *publicHandle.toFavToAdd(false).Data.TeamID,
+		PrivateTeamID: *publicHandle.toFavToAdd(false).Data.TeamID,
```
strib: `publicHandle` -> `privateHandle`?

strib: Also we probably shouldn't blindly make calls on `publicHandle` or `privateHandle` here, since they could be `nil` if there was an error above.  Also also, I'm not sure about blindly dereferencing the TeamIDs either, can they ever legitimately be nil?

PR #15956, opened 2019-02-07
folder_branch_ops: don't implement KBFSOps

jakob223:


jakob223:
I might have missed a place or two but they can be found later I suppose.

PR #15963, opened 2019-02-08
Placeholder common adapter

cecileboucheron:
This creates a <Placeholder> common adapter since we've been using them in 3 different tabs. Also I rounded corned it.
@keybase/react-hackers 

image


songgao:
Looks like you need to update the story shots again.

PR #15967, opened 2019-02-08
Finesse new font styles.

adamjspooner:
@keybase/react-hackers, this is ready for review. (I snuck an update to the reacji row's padding in there too.)


shared/chat/conversation/messages/wrapper/index.js
```diff
@@ -604,7 +606,7 @@ const styles = Styles.styleSheetCreate({
   timestamp: Styles.platformStyles({
     common: {position: 'relative'},
     isElectron: {top: 1},
-    isMobile: {top: 2},
+    isMobile: {top: 1.5},
```
chrisnojima: i doubt we want to do this. why is the half pixel necessary

adamjspooner: I used an on-screen ruler to line it up. It was half a pixel off either way.

chrisnojima: i think this is indicative of a parent container thats not the right size and then doing a centering

adamjspooner: Okay, I'll see what I can do to make this better. 👍

adamjspooner:
@chrisnojima, how's that?

adamjspooner:
Totally! Thanks! 🙌

PR #15972, opened 2019-02-08
Devices improvements 3 - Animation for add device options

buoyad:
Support for animating options in the add device dialog if we want to draw attention to them. Hooks the animation up to the paper key todo item and the nudge from #15964. 
device-animate-desktop
device-animate-mobile

r? @keybase/react-hackers


buoyad:
@keybase/react-hackers bump

chrisnojima:
I don't think we want this pulsing animation

PR #15973, opened 2019-02-08
ident3 perf fixes and remove blackbar

chrisnojima:
@keybase/react-hackers this reduces some render thrashing and fixes a blackbar

Mostly not passing unneeded props to children
Blackbar would happen as we overwrite guiid if we happen to do multiple ids. I think its fine to toss old guiIDs and just ignore incoming calls

### DIFF
--- a/shared/profile/user/index.js
+++ b/shared/profile/user/index.js
@@ -12,10 +12,14 @@ import Friend from './friend/container'
 import Measure from './measure'
 import Teams from './teams/container'
 import Folders from '../folders/container'
+import shallowEqual from 'shallowequal'
+import * as Flow from '../../util/flow'
+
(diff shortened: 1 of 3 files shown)


shared/profile/user/index.js
```diff
@@ -151,7 +183,12 @@ class FriendRow extends React.PureComponent<{|usernames: Array<string>, itemWidt
   }
 }
 
-class BioTeamProofs extends React.PureComponent<Props> {
+type BioTeamProofsProps = {|
```
buoyad: This is missing `assertionKeys`

chrisnojima: updated, also needed suggestionkeys

PR #15976, opened 2019-02-08
add support for collapsing messages with inline previews 

mmaxim:
**Go**

1. Add `utils.Collapses` to manage the local state of collapsed messages in user conversations.
2. Add support or collapsing single messages, or from a message back in time. Also store timestamp on each record to allow for each method to override each other, depending on which one came last.
3. Add /collapse and /expand commands to run the range toggle.
4. Use the `MESSAGES_UPDATED` chat activity to drive the UI changing, except for /collapse and /expand, in which case we just drop a thread stale and reload.

**JS**
@keybase/react-hackers @adamjspooner 

1. Add collapse buttons to image preview messages and unfurls. 
2. For unfurls, we behave like Slack and only collapse bottom images. For images we add some text above with the filename and a caret to collapse. 
3. Hook everything up to the service.


adamjspooner:
This is awesome! +50xlm@mmaxim ❤️ 

Is it possible to keep our position in the thread after the collapses have happened? Right now, my scroll position shifts when I collapse an individual message (or use `/collapse`).

mmaxim:
Hmm it seems to work for me with scroll position, I think if it is shifting it might just be some larger bug that is causing it to reset.

mmaxim:
Well I guess for /collapse it will reset it, but the individual ones should work.

adamjspooner:
Interesting. I used `/collapse` in `#random` when I was at the last message and it jumped me backwards in the thread. I tried it again just now and it was fine. 🤷‍♂️

go/chat/utils/collapses.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+type Collapses struct {
+	globals.Contextified
+	DebugLabeler
+
+	clock clockwork.Clock
+}
+
+func NewCollapses(g *globals.Context) *Collapses {
```
joshblum: should htis have an in mem cache as well?

go/chat/commands/base.go
```diff
@@ -33,6 +34,21 @@ func newBaseCommand(g *globals.Context, name, usage, desc string, aliases ...str
 	}
 }
 
+func (b *baseCommand) getRemoteConvByID(ctx context.Context, uid gregor1.UID,
```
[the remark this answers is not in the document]
mmaxim: Can't use it in this package unfortunately.

go/chat/utils/collapses.go
mmaxim: Didn't want to bother with it, it shouldn't be called too often.

shared/chat/conversation/messages/attachment/image/container.js
```diff
@@ -22,6 +22,15 @@ const mapDispatchToProps = dispatch => ({
       })
     )
   },
+  _onCollapse: (message: Types.MessageAttachment) => {
```
buoyad: no braces here

go/chat/server.go
```diff
@@ -2403,3 +2403,50 @@ func (h *Server) SaveUnfurlSettings(ctx context.Context, arg chat1.SaveUnfurlSet
...
+	}
+	if msg.Valid().MessageBody.IsType(chat1.MessageType_UNFURL) {
+		unfurledMsg, err := GetMessage(ctx, h.G(), uid, arg.ConvID,
+			msg.Valid().MessageBody.Unfurl().MessageID, true, nil)
+		if err != nil {
+			h.Debug(ctx, "ToggleMessageCollapse: failed to get unfurl base message: %s", err)
+			return nil
+		}
+		// give a notification about the unfurled message
+		notif := chat1.MessagesUpdated{
```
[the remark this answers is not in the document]
mmaxim: I think it's not worth it, there is no else to fall into here.

go/chat/server.go
```diff
@@ -2403,3 +2403,50 @@ func (h *Server) SaveUnfurlSettings(ctx context.Context, arg chat1.SaveUnfurlSet
 		Whitelist: wm,
 	})
 }
+
+func (h *Server) ToggleMessageCollapse(ctx context.Context, arg chat1.ToggleMessageCollapseArg) (err error) {
+	ctx = Context(ctx, h.G(), keybase1.TLFIdentifyBehavior_CHAT_GUI, nil, h.identNotifier)
+	defer h.Trace(ctx, func() error { return err }, "ToggleMessageCollapse(%s,%d,%v)", arg.ConvID, arg.MsgID,
```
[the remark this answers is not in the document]
mmaxim: sure

shared/chat/conversation/messages/wrapper/unfurl/generic/index.js
```diff
@@ -55,23 +62,32 @@ class UnfurlGeneric extends React.Component<Props> {
...
           )}
-          {!!this.props.imageURL &&
-            !!this.props.imageHeight &&
-            !!this.props.imageWidth &&
-            !this.props.showImageOnSide && (
+          {showBottomImage && !this.props.isCollapsed && (
+            <Kb.Box2 direction="vertical" fullWidth={true}>
               <UnfurlImage
-                url={this.props.imageURL}
+                url={this.props.imageURL || ''}
```
buoyad: comment around here about `showBottomImage` asserting these are set would be nice

shared/chat/conversation/messages/wrapper/unfurl/unfurl-list/container.js
```diff
@@ -19,6 +19,9 @@ const mapDispatchToProps = (dispatch, {conversationIDKey}: OwnProps) => ({
   onClose: (messageID: Types.MessageID) => {
     dispatch(Chat2Gen.createUnfurlRemove({conversationIDKey, messageID}))
   },
+  onCollapse: (messageID: Types.MessageID, collapse: boolean) => {
```
buoyad: :fire: braces

shared/chat/conversation/messages/wrapper/unfurl/generic/index.js
```diff
@@ -55,23 +62,32 @@ class UnfurlGeneric extends React.Component<Props> {
           {!!this.props.description && (
             <Kb.Text type="Body" lineClamp={5}>
               {this.props.description}
+              {!Styles.isMobile && showBottomImage && (
+                <React.Fragment>
```
buoyad: `<>` and `</>` below

shared/chat/conversation/messages/wrapper/unfurl/generic/index.js
mmaxim: I was annoyed Flow didn't figure this out.

chrisnojima: flow is correct here i believe. just cause you check doesnt' mean something can't come in and just change it aka `this.props.imageUrl = null`. if you really want to just make a const like
`const goodImageURL`, then checks against that should be resolved fine (since its const)

PR #15977, opened 2019-02-08
preliminary implementation of a coinflip secure MPC

maxtaco:
- main code is in dealer.go, which implements the state machine and routing
- should integrate easily with chat
- some testing for happy and sad paths
- needs some more tests for exceptional and error cases
- needs better docs


go/chat/flip/prot.avdl
```diff
@@ -0,0 +1,101 @@
+
+@namespace("flip")
```
mmaxim: Anyway we could make this part of `chat1`? I can also just do that when integrating.

maxtaco: sure, we can.  it was nice to have it all self-contained.  but maybe that's silly

maxtaco: BTW, these prot types are actually never explicit exported. They are all hidden from chat since chat is just passing around opaque messages.

maxtaco: Eh, it's unfortunately a big change...

maxtaco: What if I leave it as is?

mmaxim: Why is there AVDL at all?

mmaxim:
I'm just going to merge this, we can figure it out later.

PR #15978, opened 2019-02-08
simplefs: handle recursive removes

strib:
This is the first step to being able to handle cross-TLF moves properly.

Issue: KBFS-3853

### DIFF
--- a/go/kbfs/simplefs/simplefs.go
+++ b/go/kbfs/simplefs/simplefs.go
@@ -1141,12 +1141,20 @@ func (k *SimpleFS) SimpleFSCopyRecursive(ctx context.Context,
 		})
 }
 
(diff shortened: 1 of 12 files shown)


go/kbfs/simplefs/simplefs.go
```diff
@@ -1164,7 +1172,7 @@ func (k *SimpleFS) SimpleFSMove(ctx context.Context, arg keybase1.SimpleFSMoveAr
 			if err != nil {
 				return err
 			}
-			return k.doRemove(ctx, arg.Src)
+			return k.doRemove(ctx, arg.Src, false)
```
jakob223: Do we plan to support recursive `SimpleFSMove` in the future?

strib: Yep!  Working on it right now, as a followup to this PR.  That was the motivation for this PR, actually.

PR #15979, opened 2019-02-08
chat attachments: use SimpleFS for KBFS paths; also ensure Close()

songgao:
Will use this for sending attachments to chat from Files tab.

### DIFF
--- a/go/chat/attachments/sender.go
+++ b/go/chat/attachments/sender.go
@@ -27,12 +27,7 @@ func NewSender(gc *globals.Context) *Sender {
 
 func (s *Sender) MakePreview(ctx context.Context, filename string, outboxID chat1.OutboxID) (res chat1.MakePreviewRes, err error) {
 	defer s.Trace(ctx, func() error { return err }, "MakePreview")()
-	var src ReadCloseResetter
-	if IsKbfsPath(filename) {
-		src, err = NewKbfsReadResetter(ctx, s.G().GlobalContext, filename)
-	} else {
-		src, err = NewFileReadResetter(filename)
-	}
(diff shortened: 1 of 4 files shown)


go/chat/attachments/utils.go
```diff
@@ -120,6 +129,85 @@ func (f *FileReadResetter) Close() error {
...
+type KbfsReadResetter struct {
+	client *keybase1.SimpleFSClient
+	opid   keybase1.OpID
+	offset int64
+	ctx    context.Context
+}
+
+const (
+	kbfsPrefix        = "/keybase"
+	kbfsPrefixPrivate = "/keybase/private/"
```
strib: Use `kbfsPrefix+` to define the rest of these constants?

go/chat/attachments/sender.go
```diff
@@ -27,7 +27,12 @@ func NewSender(gc *globals.Context) *Sender {
 
 func (s *Sender) MakePreview(ctx context.Context, filename string, outboxID chat1.OutboxID) (res chat1.MakePreviewRes, err error) {
 	defer s.Trace(ctx, func() error { return err }, "MakePreview")()
-	src, err := NewFileReadResetter(filename)
+	var src ReadCloseResetter
+	if IsKbfsPath(filename) {
```
mmaxim: Would be nice if this got moved out to some other function.

songgao:
ready for another look thanks!

songgao:
Thanks for looking! I just tried this in sim with network link conditioner on 3G; tried backgrouding and killing app and uploading still worked. Seems OK. Merging!

PR #15980, opened 2019-02-08
simplefs: better support for moves

strib:
* Moves within a TLF are now just a cheap `Rename`.
* Moves between TLFs now work recursively.

Issue: KBFS-3853


strib:
A better view for the recursive copy refactor: https://github.com/keybase/client/pull/15980/files?w=1

PR #15989, opened 2019-02-11
Don't call getInbox while filtering & simplify initial load

buoyad:
Removes two false positives for triggering `getInboxNonblock` calls

1. Inbox container would mistake the filter clearing for us having expanded the inbox
2. Inbox would mistake the filter clearing for rendering for the first time

Makes it so filter opening / searching / clearing won't trigger `getInbox` calls. Arrowing through convos will still call it. r? @keybase/react-hackers cc @mmaxim 

Edit:
* Adds `state.chat2.trustedInboxHasLoaded`, false until the inbox requests a trusted load
* Tweak list callbacks to unbox on expanding small teams automatically


shared/chat/inbox/container/index.js
```diff
@@ -137,10 +140,12 @@ class InboxWrapper extends React.PureComponent<Props, State> {
...
 
   componentDidUpdate(prevProps) {
-    const loadedForTheFirstTime = prevProps.rows.length === 0 && this.props.rows.length > 0
+    const loadedForTheFirstTime =
+      !prevProps.filter && prevProps.rows.length === 0 && this.props.rows.length > 0
     // See if the first 6 are small, this implies it's expanded
     const smallRowsPlusOne = prevProps.rows.slice(0, 6).filter(r => r.type === 'small')
-    const expandedForTheFirstTime = smallRowsPlusOne.length === 5 && this.props.rows.length > 5
+    // !prevProps.filter because that's a false positive
+    const expandedForTheFirstTime = !prevProps.filter && smallRowsPlusOne.length > 5
```
buoyad: `smallRowsPlusOne === 5` to `smallRowsPlusOne > 5` here, I think this is right because if `smallrowsPlusOne.length <= 5`, we've definitely shown those rows before (before expanding or we have a fresh meta because of incoming activity)

chrisnojima:
Not clear to me why we ever have to call get inbox   Let's discuss tomorrow

buoyad:
@chrisnojima it calls if you filter and arrow to convos we haven’t gotten trusted metas for yet

chrisnojima:
Yeah but why bother. If it's for the thread the thread can do it. Else why bother in this view

buoyad:
@chrisnojima the calls here are for the initial inbox load and loading more due to scroll. The calls when arrowing around are caused by `selectConversation` triggering it if we don't have a trusted meta.

buoyad:
@keybase/react-hackers bump

PR #15991, opened 2019-02-11
Increase timeout of avatar upload to 5 min

zapu:


zapu:
Is there anything else we can do here? Clientside resizing might be a good idea.

PR #15994, opened 2019-02-11
Fix gendeps

joshblum:
cc @jzila not sure what changed caused this but ci is breaking, going to merge


jzila:
@joshblum which CI builds are broken? I don't see any broken builds on master.

joshblum:
@jzila i didn't see anything broken on master either which is why i was confused, initially i saw red here: https://ci.keyba.se/job/client/job/PR-15990/1/execution/node/328/log/

jzila:
There is definitely something wrong with this PR here--I don't think you should have merged this without letting Jenkins finish first. We might need to debug the script.

PR #15997, opened 2019-02-11
Team sigchain stricter checks on owner invites

mlsteele:
Non-owners can't invite owners. If this is wrong or disagrees with the real wild world, then teams will be unloadable. Better block these shenanigans on the server and scan existing teams. I don't want to merge this until those are done.

### DIFF
--- a/go/teams/chain.go
+++ b/go/teams/chain.go
@@ -977,6 +977,7 @@ func (t *teamSigchainPlayer) addInnerLink(
 				signerIsExplicitOwner := true
 				additions, cancelations, err := t.sanityCheckInvites(signer.signer, signerIsExplicitOwner,
 					*team.Invites, sanityCheckInvitesOptions{
+						isRootTeam:   true,
 						implicitTeam: isImplicit,
 					})
 				if err != nil {
@@ -1479,6 +1480,7 @@ func (t *teamSigchainPlayer) addInnerLink(
 
 		additions, cancelations, err := t.sanityCheckInvites(signer.signer, signerIsExplicitOwner,
 			*team.Invites, sanityCheckInvitesOptions{
+				isRootTeam:   !prevState.IsSubteam(),
 				implicitTeam: prevState.IsImplicit(),
 			})
 		if err != nil {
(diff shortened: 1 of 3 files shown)


zapu:
So the current state is that no one can be invited as an owner, no matter by whom, unless it's an implicit team. Is this necessarily a bad thing? We shouldn't have many PUKless cases now, unless we consider in-between resets, in which the admins should really not invite users like that at all.

i'm open for being convinced otherwise though!

zapu:
We are also disallowing owner social invites on RPC level: https://github.com/keybase/client/blob/master/go/teams/teams.go#L897-L899

But I guess with this PR we are positioning ourselves to support admin->owner promotion. But I don't see why we should allow "inviting" owners.

mlsteele:
> So the current state is that no one can be invited as an owner, no matter by whom, unless it's an implicit team.

This isn't what I've observed. This may be made more confusing by the difference in terminology of invite/add in the realms of sigchain and user experience. An owner can sigchain-invite another owner if they have no puk (website user).

mlsteele:
This PR clamps down on admins trying to sigchain-invite owners. Which seems like it ought to be disallowed. I observed this happening locally and it roundtripped.

mlsteele:
Scanned prod teams and they're all ok with this change. Given the exceptions added to chain.go. With @pzduniak's server change no new links should cross the line. 

@maxtaco the sig ID comes from `DecodeOuterLinkV2` so should be good.

@zapu what do you think?

go/teams/chain.go
```diff
@@ -1719,7 +1737,15 @@ func (t *teamSigchainPlayer) sanityCheckInvites(
 	if invites.Owners != nil && len(*invites.Owners) > 0 {
 		additions[keybase1.TeamRole_OWNER] = nil
 		for _, i := range *invites.Owners {
-			if !options.implicitTeam && !assertIsKeybaseInvite(t.G(), i) {
+			if !options.isRootTeam {
+				return nil, nil, fmt.Errorf("encountered invite of owner in non-root team")
+			}
+			if !signerIsExplicitOwner {
+				if !hardcodedInviteRuleExceptionSigIDs[sigID] {
+					return nil, nil, fmt.Errorf("encountered invite of owner by non-owner")
```
zapu: I think a good test would be to try to post a link like that (bypassing service_helper functions and transactions code) and having it caught by precheck. There are some tests like this for implicit teams if you grep for `PrecheckAppendError`.

PR #16004, opened 2019-02-12
conflict_resolver: use a disk-based fileBlockMap

strib:
This stores the top block for each file in a disk cache, to help the case where there are tons of files being CR'd.

Issue: KBFS-3852

### DIFF
--- a/go/kbfs/libkbfs/conflict_resolver.go
+++ b/go/kbfs/libkbfs/conflict_resolver.go
@@ -2252,10 +2252,6 @@ func (cr *ConflictResolver) computeActions(ctx context.Context,
 	return actionMap, append(newUnmergedPaths, moreNewUnmergedPaths...), nil
 }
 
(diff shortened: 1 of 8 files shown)


strib:
@jzila:

> Where do we clean the blocks from the disk?

https://github.com/keybase/client/blob/15e231f2848282b1fcefc1d8acc0c652af7774bb/go/kbfs/libkbfs/conflict_resolver.go#L3358

> And where do we rebuild the in-memory ptrs if we restart with a non-empty fileBlockMapDisk?

We never start with one.  There is a new temp dir for every run of conflict resolution.

jzila:
Got it, so it gets cleaned up on the way out. What if we get a crash that skips the cleanup function?

strib:
Yeah at some point we'll need to clean those up automatically.  We've had the exact same problem with kbfsgit temp dirs since the beginning of time.  If it becomes a priority we'll need to add something that cleans up ones that are older than e.g. a day.

PR #16006, opened 2019-02-12
integrate flip library into chat with UI CORE-10184

mmaxim:
Integrate the coin flip library into chat, highlights:

**Go**
1. Introduce `chat.FlipManager` which serves as the main interface between chat and the new coin flip library.
2. Implement `FlipManager.StartFlip` to do the following:
 a. Create a new text message with /flip text tagged with the game ID of the flip.
 b. Create a new dev conversation with a topic name derived from the game ID.
 c. Handoff new dev conversation to `flip.Dealer` to begin the game.
3. Hook up new message flow to `flip.Dealer.InjectChatMessage` using the topic name of the conversation to pull the game ID (could have also just used the special metadata message at the beginning of the dev conv I guess).
4. Add `LoadFlip` to be called whenever we present a text message with a flip on it to the UI. Will invoke `flip.Replay` and turn the result into a UI update for the relevant game ID.
5. Add UI update apparatus that aggregates updates from the flip library and transmits to the frontend. This is on a 200ms delay to not blow away JS with updates (most relevant in huge flips). The system is designed to transmit data that can essentially be pasted right into the coin flip component props to update the display.

**JS**
@keybase/react-hackers 
1. Add new a component for displaying coin flip information on a message. We key the whole thing on game ID in Redux so that when loading a message, we simply need to take the game ID and look up for the status in the flip status map.
2. Basically just do what we did for wallet in-chat sends, the two systems work basically the same way.

I'm likely just going to merge this once CI passes so that people can play with it, the frontend is just a prototype and is going to change once people experience it.


shared/chat/conversation/messages/coinflip/container.js
```diff
@@ -0,0 +1,38 @@
...
+  RPCChatTypes.chatUiUICoinFlipPhase.reveals,
+  RPCChatTypes.chatUiUICoinFlipPhase.complete,
+]
+
+const mapStateToProps = (state, {flipGameID}: OwnProps) => {
+  const status = state.chat2.getIn(['flipStatusMap', flipGameID])
+  return !status
+    ? {
+        isError: false,
+        participants: [],
```
chrisnojima: Make a const or this render thrashes

mmaxim:
cc @malgorithms

shared/chat/conversation/messages/coinflip/index.js
```diff
@@ -0,0 +1,112 @@
...
+
+type State = {
+  showPopup: boolean,
+}
+
+class CoinFlip extends React.Component<Props, State> {
+  partRef: any
+  state = {showPopup: false}
+  constructor(props: Props) {
+    super(props)
```
chrisnojima: You don't need this constructor

shared/chat/conversation/messages/wrapper/index.js
```diff
@@ -186,6 +187,10 @@ class _WrapperMessage extends React.Component<Props & Kb.OverlayParentProps, Sta
       />
     )
 
+  _coinFlip = () =>
+    // $ForceType
+    this.props.message.flipGameID && <CoinFlip key="CoinFlip" flipGameID={this.props.message.flipGameID} />
```
chrisnojima: !!

shared/constants/types/chat2/index.js
```diff
@@ -86,6 +86,7 @@ export type _State = {
   paymentConfirmInfo: ?PaymentConfirmInfo, // chat payment confirm screen data
   paymentStatusMap: I.Map<Wallet.PaymentID, Message.ChatPaymentInfo>,
   unsentTextMap: I.Map<Common.ConversationIDKey, ?HiddenString>,
+  flipStatusMap: I.Map<string, RPCChatTypes.UICoinFlipStatus>,
 } & TeamBuildingTypes.TeamBuildingSubState
```
chrisnojima: Should be an opaque type

shared/local-debug.native.js
```diff
@@ -53,13 +53,13 @@ let config = {
 }
 
 // Developer settings
+config.isDevApplePushToken = true
```
chrisnojima: Why

shared/chat/conversation/messages/coinflip/container.js
mmaxim: Would be nice if there was like a global blank list

shared/local-debug.native.js
mmaxim: Mistake

chrisnojima:
You can make some util or something

shared/chat/conversation/messages/wrapper/index.js
mmaxim: ?

shared/constants/types/chat2/index.js
mmaxim: This type maps directly onto the component, there isn't any need for some other type here.

shared/chat/conversation/messages/wrapper/index.js
mmaxim: If this is about the `ForceType`, it is also on a bunch other functions like this. The reason is this `flipGameID` doesn't show up on every incarnation of `Types.Message` so it's just easier to just silence Flow here.

go/chat/flipmanager.go
```diff
@@ -0,0 +1,736 @@
...
+
+func (m *FlipManager) addResult(ctx context.Context, status *chat1.UICoinFlipStatus, result flip.Result,
+	convID chat1.ConversationID) {
+	defer func() {
+		if len(status.ResultText) > 0 {
+			status.ProgressText += " (complete)"
+		}
+	}()
+	hmi, err := m.getHostMessageInfo(ctx, convID)
+	if err != nil {
```
maxtaco: might be slightly better as a `switch`, no?

go/chat/flipmanager.go
```diff
@@ -0,0 +1,736 @@
...
+	var status chat1.UICoinFlipStatus
+	rawGame, ok := m.games.Get(gameID.String())
+	if ok {
+		status = rawGame.(chat1.UICoinFlipStatus)
+	} else {
+		status = chat1.UICoinFlipStatus{
+			GameID: gameID.String(),
+		}
+	}
+	if update.Err != nil {
```
maxtaco: also `switch`?

go/chat/flipmanager.go
```diff
@@ -0,0 +1,736 @@
...
+	if err != nil {
+		m.Debug(ctx, "loadGame: failure finding flip conv: %s", err)
+		return err
+	}
+	if len(flipConvs) != 1 {
+		m.Debug(ctx, "loadGame: bad number of convs: num: %d", len(flipConvs))
+		return errors.New("no conv found")
+	}
+	flipConv := flipConvs[0]
+	// TODO: fix me, need to potentiallu loop on this for big flips
```
maxtaco: `potentiallu` -> `potentially`

go/chat/flipmanager.go
```diff
@@ -0,0 +1,736 @@
...
+
+	// Start the game
+	return m.dealer.StartFlipWithGameID(ctx, start, conv.GetConvID(), gameID)
+}
+
+// MaybeInjectFlipMessage implements the types.CoinFlipManager interface
+func (m *FlipManager) MaybeInjectFlipMessage(ctx context.Context, msg chat1.MessageUnboxed,
+	conv chat1.ConversationLocal) {
+	// earliest of outs if this isn't a dev convo, an error, or the outbox ID message
+	if conv.GetTopicType() != chat1.TopicType_DEV || !msg.IsValid() || msg.GetMessageID() == 2 {
```
maxtaco: I wonder if we should comment this `== 2` better or make it slightly less fragile should anything change in the future

go/chat/flipmanager.go
```diff
@@ -0,0 +1,736 @@
...
+	if !body.IsType(chat1.MessageType_TEXT) {
+		return
+	}
+	// check topic name
+	gameID, ok := m.isGameIDTopicName(conv.GetTopicName())
+	if !ok {
+		return
+	}
+	if err := m.dealer.InjectIncomingChat(ctx, sender, conv.GetConvID(), gameID,
+		flip.MakeGameMessageEncoded(body.Text().Body), msg.GetMessageID() == 3); err != nil {
```
maxtaco: same goes here for `== 3`

go/chat/flipmanager.go
mmaxim: Wasn't too sure how to make this a switch.

maxtaco:
```switch {
   case foo != nil:
   case bar != nil:
}```

no?

maxtaco: https://tour.golang.org/flowcontrol/11

mmaxim: TIL

PR #16010, opened 2019-02-12
Plumb full names in search suggestions and new convo suggestions

zapu:


zapu:
cc @heronhaye who also expressed interesting in this fix

zapu:
Tagging Mike for `func (h *UserHandler) InterestingPeople` change:
- uses UIDMapper to get full names, with a network budget.

zapu:
Ok I'm making this use cached data only. Can you remind me why this can never hit network?

maxtaco:
I think the UIDmapper supports that, you can give it a 0-timed network budget.

mmaxim:
It needs to show up instantly on slow network or no network.

zapu:
Max can you confirm? I was looking at it today and it looks like 0-budget means infinite time:
https://github.com/keybase/client/blob/master/go/uidmap/uidmap.go#L270

and it looks like there is no way to parameterize it so it never tries to look up anything:
https://github.com/keybase/client/blob/master/go/uidmap/uidmap.go#L390

zapu:
I wonder if I could expand UIDMapper API so it returns only cached data, but in InterestingPeople RPC fire another UIDMapper query in a goroutine (in background) so it caches data for next time when this is called.

maxtaco:
Set `forceNetworkForFullnames` to be `false`

I'm reasonably certain this already exists, it was in the original requirements

maxtaco:
Maybe one of the comments is lying, IDK.  If it's not already there, this seems like an easy change

zapu:
There might be an inconsistency here:

```
If the `forceNetworkForFullNames` flag is specified, it will block on
// the network too. If the flag is not specified, then stale values (or unknown
// values) are OK, we won't go to network if we lack them.
```

But the way it queues network lookups is the following:

```
// There are 3 important cases when we should go to network:
//
//  1. No username is found (up == nil)
//  2. No FullName found and we've asked to force network lookups (status == notFound && forceNetworkForNullNames)
//  3. The FullName found was stale (status == stale).
//
// Thus, if you provide forceNetworkForFullName=false, and fullNameFreshness=0, you can avoid
// the network trip as long as all of your username lookups hit the cache or are hardcoded.
if u.testNoCachingMode || up == nil ||
	(status == notFound && forceNetworkForFullNames) || (status == stale) {
	apiLookupIndex[len(uidsToLookup)] = i
	uidsToLookup = append(uidsToLookup, uid)
}
```

So if the user package is not found at all it will always go to the network and this cannot be controlled.

zapu:
Fully offline uidmapper API change is done. I added new call instead of adding second bool or enum, looked cleaner this way - I still have the branch with enum if someone wants to see it.

go/uidmap/uidmap.go
```diff
@@ -479,6 +479,29 @@ func (u *UIDMap) ClearUIDAtEldestSeqno(ctx context.Context, g libkb.UIDMapperCon
 	return nil
 }
 
+func (u *UIDMap) MapUIDsToUsernamePackagesOffline(ctx context.Context, g libkb.UIDMapperContext,
```
zapu: @maxtaco let me know if this is good, if so I'll add a test

zapu:
@mmaxim is it good to go now with the cached uidmapper change?

mmaxim:
As long as it fulfills the property that a call to `InterestingPeople` does not hit the network under any circumstance, then it sounds good to me.

PR #16011, opened 2019-02-12
override pointer event disabling so clicks right after scroll aren't lost

chrisnojima:
@keybase/react-hackers adds a workaround per: https://github.com/bvaughn/react-window/issues/128


chrisnojima:
cause it wasn't imported and collapsestyles on desktop will just do what that code does but i changed it to import styles and just modernized more of it

PR #16014, opened 2019-02-12
move disklru to libkb

joshblum:
unfortunately leveldb is not separable from libkb and will be using disk_lru for the dbcleaner state


joshblum:
closing for now, https://github.com/keybase/client/pull/16033 did not end up using this

PR #16015, opened 2019-02-12
Support filtering unread conversations WIP

buoyad:
This makes it so if the filter starts with `@unread` we'll limit the filtering to convos with unread messages. I had to tweak markAsRead handling so we don't mark them as read while arrowing through the list. Hooks up `cmdKey+R` to focusing + prefilling the filter with `@unread `. r? @keybase/react-hackers


buoyad:
Sorry @zanderz I should've commented, this is gonna go a different direction after a talk irl. Instead of `@unread` and a new shortcut it'll default to showing unread conversations first if no filter is set.

buoyad:
Done by inbox search

PR #16019, opened 2019-02-12
Hide deleted team members from Details

mlsteele:
cc @maxtaco

### DIFF
--- a/go/teams/service_helper.go
+++ b/go/teams/service_helper.go
@@ -82,6 +82,8 @@ func Details(ctx context.Context, g *libkb.GlobalContext, name string) (res keyb
 	}
 	res.AnnotatedActiveInvites = annotatedInvites
 
+	membersHideDeletedUsers(ctx, g, &res.Members)
+
 	res.Settings.Open = t.IsOpen()
 	res.Settings.JoinAs = t.chain().inner.OpenTeamJoinAs
 	return res, nil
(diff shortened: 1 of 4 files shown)


zapu:
Is the current situation really bad? They are automatically sweeped in open teams.

In a sigchain world I'd ideally want to present user the _actual state_ instead of smoke and mirrors.

mlsteele:
I'm with you on showing the user the actual state. Max pointed out that even if a cryptomember of the team is deleted, the clients do not box new keys for them. If an owner deleted and is hidden. In order for anyone to exercise that hidden power the deleted owner would have to undelete. So maybe the _actual_ state of the team is the team's sigchain together with the merkle world. It's likely not quite that simple since not everyone loads everyone else.

The problem I'm trying to solve, and maybe there's a better way, is this. An owner cannot remove a deleted owner.
image

mlsteele:
Also a row saying someone has deleted with a button to "Remove" them when they are already deleted arguably isn't valuable.

zapu:
ok I now convinced myself that this is a good idea. let's go!

go/teams/service_helper.go
```diff
@@ -61,6 +61,7 @@ func Details(ctx context.Context, g *libkb.GlobalContext, name string) (res keyb
 	res.AnnotatedActiveInvites = annotatedInvites
 
 	membersHideInactiveDuplicates(ctx, g, &res.Members)
+	membersHideDeletedUsers(ctx, g, &res.Members)
```
zapu: I'm not sure how much this matters but maybe these two above can be collapsed into one, then you'd be iterating over all the lists only once (not counting rest of the code which does this a few times probably)

mlsteele:
Rebased so this one can go first

PR #16020, opened 2019-02-12
prefetcher: send most debug logs to the verbose debug logger

strib:
Expose a `MakeVLogger()` method in the config, and use that in the prefetcher and block retrieval queue to send spammy log messages there.  Only turn it on during certain tests that require it.

This improves testing time a little bit.  With current logging:

```sh
src/github.com/keybase/client/go/kbfs/test$ time ./test.test -test.v > /tmp/x 2>&1

real	0m30.997s
user	0m44.824s
sys	0m4.360s
src/github.com/keybase/client/go/kbfs/test$ ls -l /tmp/x
-rw-rw-r-- 1 strib strib 253229687 Feb 12 13:07 /tmp/x
```

With this PR:

```sh
src/github.com/keybase/client/go/kbfs/test$ time ./test.test -test.v > /tmp/y 2>&1

real	0m27.354s
user	0m38.836s
sys	0m4.372s
src/github.com/keybase/client/go/kbfs/test$ ls -l /tmp/y
-rw-rw-r-- 1 strib strib 121987372 Feb 12 13:08 /tmp/y
```

In production, we can turn this on via the `KEYBASE_VDEBUG` environment variable (e.g., `KEYBASE_VDEBUG=vlog1`) or by setting the `vdebug` key in the Keybase config file (e.g., `keybase config set vdebug vlog1` + a restart).

Issue: KBFS-3875

### DIFF
--- a/go/kbfs/env/context.go
+++ b/go/kbfs/env/context.go
@@ -54,6 +54,7 @@ type Context interface {
 	NewRPCLogFactory() rpc.LogFactory
 	GetKBFSSocket(clearError bool) (net.Conn, rpc.Transporter, bool, error)
 	BindToKBFSSocket() (net.Listener, error)
+	GetVDebugSetting() string
 }
 
 // KBFSContext is an implementation for libkbfs.Context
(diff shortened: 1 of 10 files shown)


strib:
> Can we do the disk cache too?

@jzila which messages specifically?  I don't think `Get`s log anything, and I personally find the `Put` messages useful (and they're usually paired with `bserver_remote` messages anyway, so it's not like they explode the logs on their own).

jzila:
Fair enough. Carry on then!

PR #16021, opened 2019-02-12
add test for loader bug that is fixed server-side

xgess:
this bug appears to be fixable by either this PR or https://github.com/keybase/keybase/pull/3464. neither is a slam dunk in my mind.

### DIFF
--- a/go/teams/loader2.go
+++ b/go/teams/loader2.go
@@ -349,6 +349,17 @@ func (l *TeamLoader) walkUpToAdmin(
 		}
 		load2Res, err := l.load2(ctx, arg)
 		if err != nil {
+			if aerr, ok := err.(libkb.AppStatusError); ok {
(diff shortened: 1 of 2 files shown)


heronhaye:
I think this is a much better solution but Miles and Max should comment

xgess:
i think the right place to fix this is the server. the safer place is the client. we should discuss.

heronhaye:
but you actually aren't in the team though

mlsteele:
You have access to read the team. The value (not the presence) of `read_subteam_id` field is just telling the server something it already knows.

xgess:
@mlsteele it also indicates to the server which links in the chain are actually necessary so the rest can be pruned.

heronhaye:
>value (not the presence)
That makes sense

PR #16024, opened 2019-02-12
quit menu on Windows

zanderz:
confusingly performed the same operation as "close", which appears next to "quit" on the file menu

### DIFF
--- a/shared/util/quit-helper.desktop.js
+++ b/shared/util/quit-helper.desktop.js
@@ -2,6 +2,7 @@
 import * as SafeElectron from '../util/safe-electron.desktop'
 import {quit} from '../desktop/app/ctl.desktop'
 import {hideDockIcon} from '../desktop/app/dock-icon.desktop'
+import {isWindows} from '../constants/platform'
 
 export type Context = 'uiWindow' | 'mainThread' | 'quitButton' | 'beforeQuit'
 export type Action = 'closePopups' | 'quitMainWindow' | 'quitApp'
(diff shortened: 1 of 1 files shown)


shared/util/quit-helper.desktop.js
```diff
@@ -10,7 +11,7 @@ export type Action = 'closePopups' | 'quitMainWindow' | 'quitApp'
 function quitOnContext(context: Context): Array<Action> {
   switch (context) {
     case 'uiWindow':
-      return ['closePopups', 'quitMainWindow']
+      return isWindows ? ['quitApp'] : ['closePopups', 'quitMainWindow']
```
jzila: I support this change, although I've asked for it in the past and experienced pushback.

chrisnojima: what is motivating this change

zanderz: These 2 menu items seem to do the same thing, and people are cranky that "quit' doesn't quit the app. Could also be solved by just removing "quit".
image

chrisnojima: this is how it works on all the other platforms and is a behavior we decided on in design a long time ago. if you want to convince them we should change go for it but lets not change the UX on window just cause

zanderz:
I believe I showed @cecileboucheron and she agreed that quit should quit, I'll double check. There have definitely been user complaints too.

jzila:
😍

zanderz:
OK I couldn't figure out how to separate Ctrl-Q from the menu item for "quit" (note: this isn't a standard Windows shortcut anyway), but hideDockIcon() doesn't seem to be working on any platform, so I changed the Windows one to quit the app and leave background services up.

chrisnojima:
We need to discuss tomorrow. I don't think this is what we intend either

jzila:
What was decided in the discussion?

PR #16029, opened 2019-02-13
cleanup: merge a few unnecessarrily platform-specific files

songgao:
I noticed this while trying to debug that encoding issue. They used to need to be split because we were using `webview` directly. Now that it's in `Kb`, just merge those files in `fs/filepreview`.

Also get rid of `pdf-view` as we are not using it anymore.


shared/fs/filepreview/text-view.js
```diff
@@ -17,9 +20,26 @@ const TextView = (props: TextViewProps) => (
...
+  margin: 0;
+}
+body{
+  background-color: ${Styles.globalColors.white};
+  padding: ${Styles.globalMargins.medium};
+  margin: 0;
+  color: ${Styles.globalColors.black};
+  font-size: 15;
+  line-height: 1.6;
+  font-family: ${Styles.globalStyles.fontTerminal.fontFamily}
```
jzila: I'm realizing this is my fault, but:
1. This needs to be in quotes, and needs to have a semicolon.

songgao: Good catch! While playing with that, I realized this is actually ineffective on `body` since there's an automatic `pre` style inserted that sets it to `monospace` and overrides what's from `body`. So I've also moved `font-family` into an explicit `pre` block.

PR #16033, opened 2019-02-13
leveldb cleaner

joshblum:
Implements a simple db cleaner on leveldb. get/puts are tracked in an in-memory  LRU (dropped the disk backing for simplicity, we can bring it in if we feel this performs poorly), unused keys are cleaned periodically basic on a configuration of desired size. we cycle through the entire level db over the course of our cleaning based on an in-memory pointer to our location in the db so we don't unfairly start purging from the beginning at each run.


mmaxim:
I'm sure you checked this, but LevelDB doesn't have any kind of stats like this builtin?

joshblum:
nothing that i could find in the docs/reading the code

go/libkb/dbcleaner.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (c *levelDbCleaner) clearCache() {
+	c.cache.Purge()
+}
+
+func (c *levelDbCleaner) clean(ctx context.Context) (err error) {
+	c.Lock()
+	defer c.Unlock()
+	if !c.forceClean && c.G().GetClock().Now().Sub(c.lastRun) < c.config.CleanInterval {
```
joshblum: only run i background active mode on mobile

joshblum:
@mmaxim updated ptal
updates:
- monitor the app state to attempt mobile cleaning only when in the BACKGROUNDACTIVE state.
- require a minimum amount of values in the cache before running a clean
- whitelist keys we don't want/need cleaned, introduces a new 'table' to store these values in
- cli/gui trigger options

joshblum:
cc @keybase/react-hackers for admin only trigger

go/libkb/dbcleaner.go
```diff
@@ -0,0 +1,319 @@
...
+	if !force && dbSize < c.config.MaxSize {
+		return nil
+	}
+
+	var totalNumPurged, numPurged int
+	for i := 0; i < 100; i++ {
+		select {
+		case <-c.cancelCh:
+			c.log(ctx, "aborting clean, %d runs, canceled", i)
+			break
```
patrickxb: I don't think these `break` statements are doing what you want.  You would need a label to break out of the `for` loop here.

see the 3rd example here:

https://golang.org/doc/effective_go.html#switch

go/libkb/leveldb_test.go
```diff
@@ -94,35 +97,66 @@ func TestLevelDb(t *testing.T) {
...
+
+				// this key will not be deleted since it is in the permanent
+				// table.
+				require.True(t, IsPermDbKey(DBDiskLRUEntries))
+				permKey := DbKey{Key: "test-key", Typ: DBDiskLRUEntries}
+				err = db.Put(permKey, nil, v)
+				require.NoError(t, err)
+
+				// cleaner will not clean the key since it was recently used
+				err = db.cleaner.clean(context.TODO(), true /* force */)
```
patrickxb: check this error.

go/libkb/leveldb_test.go
```diff
@@ -94,35 +97,66 @@ func TestLevelDb(t *testing.T) {
...
+				_, found, err := db.Get(key)
+				require.NoError(t, err)
+				require.True(t, found)
+				_, found, err = db.Get(permKey)
+				require.NoError(t, err)
+				require.True(t, found)
+
+				db.cleaner.clearCache()
+				err = db.cleaner.clean(context.TODO(), true /* force */)
```
patrickxb: check this error.

go/libkb/dbcleaner.go
joshblum: ah great catch

go/libkb/leveldb_test.go
```diff
@@ -229,20 +249,16 @@ func TestLevelDb(t *testing.T) {
 					defer wg.Done()
 
 					tr, err := db.OpenTransaction()
-					if err != nil {
-						fmt.Println(err)
-						t.Fatal(err)
-					}
+					require.NoError(t, err)
```
patrickxb: the existing test is also broken, but not supposed to call t.Fatal inside a goroutine.  (and `require` package uses `t.FailNow` which is same)...   `if err != nil { t.Error(err) }` would be fine.

joshblum: fixed in https://github.com/keybase/client/pull/16263

PR #16034, opened 2019-02-13
packaging: Resolve tocttou issue by checking that process still exist…

heronhaye:
…s before restarting


heronhaye:
We might eventually want to put a "set -e" on this file.

PR #16036, opened 2019-02-13
include number of files in cache dirs

joshblum:


joshblum:
merging over known flake

PR #16037, opened 2019-02-13
Only scroll to selected convo if the filter is focused

buoyad:
Fixes the inbox scrolling way down if you click `+123 more` and have a big team channel selected. r? @keybase/react-hackers


buoyad:
Jenkins failing on unrelated storyshots (master too)

PR #16041, opened 2019-02-13
leveldb: use bloom filters on all leveldbs

strib:
This gives small performance improvements on read workloads, at a very small cost of more on-disk storage and memory usage.

Issue: KBFS-3882

### DIFF
--- a/go/kbfs/libkbfs/disk_block_cache.go
+++ b/go/kbfs/libkbfs/disk_block_cache.go
@@ -23,6 +23,7 @@ import (
 	"github.com/keybase/client/go/protocol/keybase1"
 	"github.com/pkg/errors"
 	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/filter"
 	"github.com/syndtr/goleveldb/leveldb/opt"
 	"github.com/syndtr/goleveldb/leveldb/storage"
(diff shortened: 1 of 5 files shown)


go/kbfs/libkbfs/disk_block_cache.go
```diff
@@ -199,6 +200,7 @@ func newDiskBlockCacheLocalFromStorage(
 	blockDbOptions := *leveldbOptions
 	blockDbOptions.CompactionTableSize = defaultBlockCacheTableSize
 	blockDbOptions.BlockSize = defaultBlockCacheBlockSize
+	blockDbOptions.Filter = filter.NewBloomFilter(16)
 	blockDb, err := openLevelDBWithOptions(blockStorage, &blockDbOptions)
 	if err != nil {
 		return nil, err
```
jzila: @strib when I was referring to the metadata DB, I meant the one right underneath here. That should give the biggest wins.

strib: @jzila I know.  That one has it too.  It used the default `openLevelDB()` function which populates it.  I only added it explicitly to the DBs that use `openLevelDBWithOptions`.

That said, I couldn't measure a win with that DB.

jzila: Ah I missed that the default options have it now too, thanks for the clarification.

Issue #16043, opened 2019-02-14
Huge amount of animation events cause constant redraw


the-exodus:
After hearing complaints about 30-50% cpu usage from people with integrated GPUs I launched a profiling session and found that 75% of the total runtime is spent animating an invisible bouncing box. Every frame of animation causes a few events, and invalidates the layout forcing a regeneration and redraw.

![image](https://user-images.githubusercontent.com/40674427/52753282-49e95700-2ff6-11e9-91be-97689effe37a.png)

![image](https://user-images.githubusercontent.com/40674427/52753453-f75c6a80-2ff6-11e9-9e3a-f37c37fcd69e.png)

strib:
cc: @chrisnojima -- we've been getting reports of lots of CPU usage from the Keybase process on Linux, and I've seen it too.

chrisnojima:
K it's the typing animation. Easy fix

chrisnojima:
PR up

PR #16044, opened 2019-02-14
don't keep animations mounted all the time

chrisnojima:
@keybase/react-hackers this fixes high cpu usage due to an invisible animation
cc: @adamjspooner 

Fixes https://github.com/keybase/client/issues/16043

### DIFF
--- a/shared/chat/conversation/input-area/normal/typing/index.js
+++ b/shared/chat/conversation/input-area/normal/typing/index.js
@@ -64,18 +64,17 @@ type Props = {|
 |}
 
 export const Typing = (props: Props) => (
-  <Kb.Box
-    style={Styles.collapseStyles([
-      styles.isTypingContainer,
-      props.names.size > 0 && styles.isTypingContainerVisible,
-    ])}
-  >
-    <Kb.Box style={styles.typingIconContainer}>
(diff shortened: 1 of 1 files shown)


shared/chat/conversation/input-area/normal/typing/index.js
```diff
@@ -64,18 +64,17 @@ type Props = {|
...
+    {props.names.size > 0 && (
+      <Kb.Box style={styles.typingIconContainer}>
+        <Kb.Animation animationType="typing" containerStyle={styles.isTypingAnimation} />
+      </Kb.Box>
+    )}
+    {props.names.size > 0 && (
+      <Kb.Text lineClamp={1} type="BodyTiny" style={styles.isTypingText}>
+        <Names names={props.names} />
+      </Kb.Text>
+    )}
```
adamjspooner: Why not just wrap these two with a fragment?

chrisnojima: that changes the hierarchy which i didn't want to do. aka if this was a box2 and i did that, suddenly the gap would be different. there's only 2 things so it was easy enough to do. if you want to change it later you can

PR #16045, opened 2019-02-14
a bunch of audio/video experience improvements

songgao:
Add a `Video` component in `common-adapters` and use it in `av-view` or Files. The new component does following things:

1. uses `<video>` tag on desktop, and `<video>` tag inside a `WebView` on mobile.
2. has a container box that's `fullHeight` and `fullWidth`, so it'd expand to whatever containing block has.
3. detects video dimensions, and resize the container box to exact video dimension, so we don't have a black box around video content.
4. uses a fixed small-ish height for audio.
5. summons the `StatusBar` back when after exiting fullscreen!!!

With the new `Video` component, on desktop video expands to the viewing area, and resizes with window when needed. On mobile, video expands too, but also make sure the height is correct so the blank area is `blue5`. As a bonus point, audio controls has `blue5` background on mobile now.

If design doesn't like how it looks now, we should be able to change it more freely and easily. If this gets merged, I'll move some of chat stuff to the new `Video` component if there's benefit in it.

Here's a short video demo'ing it in the sim: `/keybase/team/keybase/videos/video-view.mov`


shared/common-adapters/video.desktop.js
```diff
@@ -0,0 +1,108 @@
...
+
+  _resizeInFly = false
+  _onResize = () => {
+    if (this._resizeInFly) {
+      return
+    }
+    this._resizeInFly = true
+    setTimeout(() => {
+      this._resizeInFly = false
+      this._updateContainerSize()
```
chrisnojima: This is not safe. Can't call setstate like this

shared/common-adapters/video.desktop.js
```diff
@@ -0,0 +1,108 @@
...
+        />
+      </Box>
+    )
+  }
+}
+
+const styles = Styles.styleSheetCreate({
+  container: {
+    // can't use Box2 as it doesn't have forwardedRef
+    ...Styles.globalStyles.flexBoxColumn,
```
chrisnojima: We should use create ref and forwardref

shared/common-adapters/video.js.flow
```diff
@@ -0,0 +1,19 @@
+// @flow
+import * as React from 'react'
+import * as Styles from '../styles'
+
+export type Props = {
```
chrisnojima: Exact

shared/common-adapters/video.js.flow
```diff
@@ -0,0 +1,19 @@
...
+  hideControls?: ?boolean,
+  style?: ?Styles.StylesCrossPlatform,
+  url: string,
+}
+
+export type State = {
+  containerHeight: number,
```
chrisnojima: Exact

shared/common-adapters/video.native.js
```diff
@@ -0,0 +1,135 @@
...
+    }
+    try {
+      const obj = JSON.parse(nativeEvent.data)
+      return obj
+    } catch {
+      return {}
+    }
+  }
+  _onMessage = ({nativeEvent}) => {
+    const {endFullscreen, size} = this._parseMessage(nativeEvent)
```
chrisnojima: Also not safe

shared/common-adapters/video.desktop.js
songgao: Thanks for the prompt review! You mean we should add forwardedRef for box2?

chrisnojima: no i mean https://reactjs.org/docs/forwarding-refs.html

shared/common-adapters/video.desktop.js
```diff
@@ -0,0 +1,99 @@
...
+    containerHeight: 0,
+    containerWidth: 0,
+    loadedVideoSize: false,
+    videoHeight: 0,
+    videoWidth: 0,
+  }
+
+  _mounted = false
+
+  _onContainerResize = ({bounds}) =>
```
chrisnojima: likely want to check _mounted here too, i don't remember if that can get called in an unsafe way, better safe than sorry

shared/common-adapters/video.desktop.js
```diff
@@ -0,0 +1,99 @@
...
+      return
+    }
+    action(this._videoRef.current)
+  }
+
+  _onVideoLoadedmetadata = ({target}) => {
+    this._mounted &&
+      this.setState({
+        loadedVideoSize: true,
+        // $FlowIssue doesn't know videoHeight
```
chrisnojima: i think you can just make the type and put it into the params vs doing the flow issue

shared/common-adapters/video.desktop.js
```diff
@@ -0,0 +1,99 @@
...
+        // $FlowIssue doesn't know videoWidth
+        videoWidth: target.videoWidth,
+      })
+  }
+
+  _onVideoClick = () => this._videoRefAction(current => (current.paused ? current.play() : current.pause()))
+
+  componentDidMount() {
+    this._mounted = true
+    this._videoRefAction(current => current.addEventListener('loadedmetadata', this._onVideoLoadedmetadata))
```
chrisnojima: why not `onLoadedMetadata` ? https://reactjs.org/docs/events.html#media-events

shared/common-adapters/video.desktop.js
songgao: Ha! The syntheticEvent has it typed nicely

shared/common-adapters/video.desktop.js
```diff
@@ -0,0 +1,88 @@
...
+  _onContainerResize = ({bounds}) =>
+    this._mounted && this.setState({containerHeight: bounds.height, containerWidth: bounds.width})
+
+  _videoRef: {current: HTMLVideoElement | null} = React.createRef()
+  _onVideoClick = () =>
+    this._videoRef.current &&
+    (this._videoRef.current.paused ? this._videoRef.current.play() : this._videoRef.current.pause())
+
+  _onVideoLoadedmetadata = ({target}) => {
+    this._mounted &&
```
songgao: @chrisnojima I couldn't find out if syntheticEvent can be called when unmounted, so kept this here.

songgao:
Ready for another look!

songgao:
bump!

songgao:
buuuuump!

PR #16046, opened 2019-02-14
Use 'View signature' instead of 'View proof' for zcash, as with btc

cjb:
@keybase/react-hackers


shared/profile/index.native.js
```diff
@@ -592,7 +592,7 @@ class UserEntry extends React.PureComponent<UserEntryProps> {
           />
           <Kb.Text
             type="BodySemibold"
-center={true}
+            center={true}
```
cjb: (Prettier made this change, and the one below.)

PR #16048, opened 2019-02-14
Change the text but not the button type, when moving between Following/Unfollow on profile

cjb:
@keybase/react-hackers CC @cecileboucheron

### DIFF
--- a/shared/profile/follow-button/index.js
+++ b/shared/profile/follow-button/index.js
@@ -23,7 +23,7 @@ class FollowButton extends Component<Props, State> {
     if (following) {
       return (
         <WaitingButton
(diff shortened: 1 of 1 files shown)


shared/profile/follow-button/index.js
```diff
@@ -23,7 +23,7 @@ class FollowButton extends Component<Props, State> {
     if (following) {
       return (
         <WaitingButton
-          type={this.state.mouseOver ? 'PrimaryGreen' : 'PrimaryGreenActive'}
+          type={'PrimaryGreenActive'}
```
chrisnojima: remove curlies

PR #16053, opened 2019-02-14
disklru cleaner

joshblum:
fixes some existing bugs with files getting stranded on disk when cache values were cleared, will prevent future stranding when the db cleaner (https://github.com/keybase/client/pull/16033) is running


joshblum:
updated to remove background cleaning and only run on db nuke since the disklru is part of the cleaner whitelist

go/lru/disk_lru.go
```diff
@@ -384,3 +386,67 @@ func (d *DiskLRU) Size(ctx context.Context, lctx libkb.LRUContext) (int, error)
...
+	}
+	for _, entry := range allVals {
+		path, ok := entry.Value.(string)
+		if !ok {
+			continue
+		}
+		cacheRevMap[path] = entry.Key
+	}
+
+	if err := os.MkdirAll(cacheDir, os.ModePerm); err != nil {
```
mmaxim: What is this for?

joshblum: santiy check to make sure the dir exists but i guess not necessary since glob ignores missing dirs https://golang.org/pkg/path/filepath/#Glob

mmaxim: Yeah it was a little confusing to me, since if it isn't there, then presumably there is nothing to clean.

PR #16066, opened 2019-02-14
Normalize Device Names when checking duplicates

joshblum:
improves the client side check so we don't get a server error
- mirrors the server normalization in the service for checking duplicate device names
- includes paper keys in `keybase1.AllDeviceNames()` to check for duplicates
- removes duplicate frontend code and improves error message to avoid confusion on normalization (open to suggestions here)

error when simple collision:
screen shot 2019-02-15 at 9 51 15 am

error when normalization causes collision:
screen shot 2019-02-14 at 5 47 59 pm

### DIFF
--- a/go/engine/login_provision.go
+++ b/go/engine/login_provision.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"sort"
-	"strings"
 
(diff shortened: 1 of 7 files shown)


joshblum:
merging over flake

PR #16069, opened 2019-02-15
folder_branch_ops: fix `keybase fs history` after re-login

strib:
Need to reinitialize the sync.Once object when the edit history is cleared on logout, so chat monitoring can be started up again after login.


go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -7947,6 +7950,7 @@ func (fbo *folderBranchOps) ClearPrivateFolderMD(ctx context.Context) {
 			fbo.cancelEdits = nil
 		}
 		fbo.editHistory = kbfsedits.NewTlfHistory()
+		fbo.launchEditMonitor = sync.Once{}
```
jakob223: If I were looking at this line outside of the context of this PR, I would be confused as to what it did / what the point was. Can you add a comment?

PR #16073, opened 2019-02-15
Scroll inbox to top when filter changes

buoyad:
And use convID instead of index to decide when to auto-scroll. r? @keybase/react-hackers


buoyad:
Done, thanks @chrisnojima

PR #16077, opened 2019-02-15
Make following green again: Avoid stale bootstrap info with one more try

zanderz:
If bootstrap is attempted before the service has established contact with the server, stale info is returned from its cache and never refreshed. This adds an argument to indicate cached info, and a 1 time retry after 10 seconds in that case, which makes the following info fresh again.

To repro the original bug, stop all Keybase processes and erase the leveldb, then start the GUI. It will start the services and ask for bootstrap info before it has established server coms, getting stale/empty cache info, and all the usernames will be blue for those you follow. With this change, they're green again.

### DIFF
--- a/go/engine/bootstrap.go
+++ b/go/engine/bootstrap.go
@@ -87,6 +87,7 @@ func (e *Bootstrap) Run(m libkb.MetaContext) error {
 			m.CWarningf("error running Tracker2Syncer (cached): %s", err)
 			return nil
 		}
+		e.status.Cached = true
 	}
 	e.usums = ts.Result()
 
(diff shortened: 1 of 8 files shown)


zanderz:
so how do we know when it is refreshed? is there a mechanism already?

chrisnojima:
we should get some kind of notification from core, dunno if an appropriate one exists already. maybe core should just tell us when teh follow state changes if its different than its cache cause then we'd need to do nothing

zanderz:
Looks like reachability should do it, I'll try that

zanderz:
@chrisnojima this is a lot simpler! Seems to also work.

patrickxb:
Is there anything you wanted me to look at?  Looks like all JS.

zanderz:
thanks @patrickxb , I had proposed a core change before this simpler one.

PR #16078, opened 2019-02-15
Fix proof row spacing on mobile

cjb:
@keybase/react-hackers 

@cecileboucheron, how's this look?  Extra vertical space before the proofs area starts on mobile and extra left/right padding on each proof.

It's good to test on a user like Coyne, since things other than proofs (like showcased teams) show up in that space too.


cecileboucheron:
LGTM!

PR #16081, opened 2019-02-15
fix a bunch of overflows; more and cleaner stories

songgao:


songgao:
Bump! This is much smaller than it seems; large number is mostly because of storyshots.

songgao:
Bump! I've got another one on top of this; would be nice to get it merged :-)

PR #16084, opened 2019-02-15
Teams as actual list

chrisnojima:
The original implementation rendered the teams in a long scrollview which causes a lot of rendering if you're in a lot of teams. Instead this uses a real list component to virtualize the rows
This also uses ListItem2 adapter so the layout is consistent and gives us hover / dividers / etc for free


shared/local-debug.native.js
```diff
@@ -66,7 +66,7 @@ if (__DEV__) {
   config.printRPC = true
   config.printRPCStats = true
   config.reduxSagaLoggerMasked = false
-  config.userTimings = true
+  config.userTimings = false
```
chrisnojima: this is just annoying while debugging

chrisnojima:
@cecileboucheron @adamjspooner 
@keybase/react-hackers

adamjspooner:
The padding around avatars and above/below rows has increased:

*master:*
screen shot 2019-02-19 at 13 20 17

*this branch:*
screen shot 2019-02-19 at 13 21 49

chrisnojima:
per discussion in chat this matches the design of the common list item

shared/teams/main/index.js
```diff
@@ -5,62 +5,150 @@ import * as Kb from '../../common-adapters'
...
+  const badgeCount = props.newRequests + props.resetUserCount
+
+  return (
+    <Kb.ListItem2
+      type="Large"
+      firstItem={props.firstItem}
+      onClick={props.onViewTeam}
+      icon={
+        <Kb.Box2 direction="vertical" style={styles.avatarContainer}>
+          <Kb.Avatar size={Styles.isMobile ? 48 : 32} teamname={props.name} isTeam={true} />
```
buoyad: These look small with the taller rows. The styleguide has them at 48px on desktop for `type="Large"`

shared/teams/main/index.js
```diff
@@ -5,62 +5,150 @@ import * as Kb from '../../common-adapters'
...
+          {!!badgeCount && <Kb.Badge badgeNumber={badgeCount} badgeStyle={styles.badge} />}
+        </Kb.Box2>
+      }
+      body={
+        <Kb.Box2 direction="vertical" fullWidth={true}>
+          <Kb.Box2 direction="horizontal" gap="tiny" alignSelf="flex-start">
+            <Kb.Text type="BodySemibold">{props.name}</Kb.Text>
```
buoyad: `lineClamp` / more right padding here for long teamnames? 
image

shared/teams/main/index.js
```diff
@@ -5,62 +5,150 @@ import * as Kb from '../../common-adapters'
...
+          </Kb.Box2>
+          <Kb.Box2 direction="horizontal" gap="tiny" alignSelf="flex-start">
+            {props.isNew && <Kb.Meta title="new" backgroundColor={Styles.globalColors.orange} />}
+            <Kb.Text type="BodySmall">
+              {props.membercount + ' member' + (props.membercount !== 1 ? 's' : '')}
+            </Kb.Text>
+          </Kb.Box2>
+        </Kb.Box2>
+      }
+      action={
```
buoyad: `Styles.isMobile ? null : (actions)` and nix the checks below

chrisnojima:
@buoyad good idea. fixed. also updated list items so it constrains the body part

PR #16087, opened 2019-02-16
Since go vet is the first thing failing on invalid Go code, make the output more easily parseable

pzduniak:


Jenkinsfile
```diff
@@ -520,7 +520,7 @@ def testGo(prefix, packagesToTest) {
       }
 
       println "Running go vet for ${pkg}"
-      sh "go vet ${pkg}"
+      sh "go vet ${pkg} || (ERR=\$? && echo \"go vet failed with error code \$ERR\" && exit \$ERR)"
```
jzila: I think this can be `(ERR=\$? echo \"...\")`, although if it works the current way you can leave it.

pzduniak:
For some context:
Today I messed up and accidentally committed broken code to master. Someone else didn't regen some files so master was already broken. I ran jenkwhy on my branch and it didn't print out anything different from master (I didn't know how simple jenkwhy was back then), so I merged it in. A few hours later Josh had to fix my mistake.

The way jenkwhy is currently structured is way too simple and only "hacked onto" the rest of our CI pipeline, resulting in lacking output. Since `go vet` is the first place that parses the AST, all the AST errors will be outputted by it. By wrapping its execution in a bash condition, we should be able to simply look for the error message instead of checking every single error that might pop up.

Jenkinsfile
pzduniak: Not 100% sure about this, it still needs to exit so it has to be shared across 2 commands.

jzila: True, fair enough.

PR #16097, opened 2019-02-18
block_retrieval_queue: LRUify prefetch status cache

jakob223:
Issue: KBFS-3750

### DIFF
--- a/go/kbfs/libkbfs/block_retrieval_queue.go
+++ b/go/kbfs/libkbfs/block_retrieval_queue.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"container/heap"
+	lru "github.com/hashicorp/golang-lru"
 	"io"
 	"reflect"
 	"sync"
@@ -139,8 +140,8 @@ type blockRetrievalQueue struct {
 	// prefetcher for handling prefetching scenarios
 	prefetcher Prefetcher
 
-	prefetchStatusLock    sync.Mutex
-	prefetchStatusForTest map[kbfsblock.ID]PrefetchStatus
+	prefetchStatusLock           sync.Mutex
+	prefetchStatusForNoDiskCache *lru.Cache
 }
 
 var _ BlockRetriever = (*blockRetrievalQueue)(nil)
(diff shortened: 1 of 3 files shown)


jakob223:
untagging review since I forgot to test this..

go/kbfs/libkbfs/block_retrieval_queue.go
```diff
@@ -266,41 +267,55 @@ func (brq *blockRetrievalQueue) notifyWorker(priority int) {
...
 		// permission errors (happens sometimes on iOS for some
 		// reason), we might need to rely on this in-memory map.
-		// TODO(KBFS-3750): make it an LRU cache.
 		brq.log.Warning("No disk block cache is initialized when not testing")
 	}
 	brq.log.CDebugf(nil, "Using a local cache for prefetch status")
-	brq.prefetchStatusForTest = make(map[kbfsblock.ID]PrefetchStatus)
+	// TODO: choose a value for this size
+	var err error
+	brq.prefetchStatusForTest, err = lru.New(1000)
```
songgao: Should we replace the `ForText` suffix with something more descriptive?

go/kbfs/libkbfs/block_retrieval_queue.go
```diff
@@ -266,41 +267,55 @@ func (brq *blockRetrievalQueue) notifyWorker(priority int) {
...
 }
 
 func (brq *blockRetrievalQueue) getPrefetchStatus(
-	id kbfsblock.ID) PrefetchStatus {
+	id kbfsblock.ID) (PrefetchStatus, error) {
 	brq.prefetchStatusLock.Lock()
 	defer brq.prefetchStatusLock.Unlock()
 	if brq.prefetchStatusForTest == nil {
-		brq.initPrefetchStatusCacheLocked()
+		err := brq.initPrefetchStatusCacheLocked()
```
songgao: I know this is not from this PR, but it seems we don't need the `brq.prefetchStatusForTest == nil` check above anymore as we already do that inside the method.

go/kbfs/libkbfs/config_local.go
```diff
@@ -1711,7 +1711,11 @@ func (c *ConfigLocal) PrefetchStatus(ctx context.Context, tlfID tlf.ID,
 		if !ok {
 			return NoPrefetch
 		}
-		return bops.queue.getPrefetchStatus(ptr.ID)
+		status, err := bops.queue.getPrefetchStatus(ptr.ID)
+		if err != nil {
+			return NoPrefetch
```
songgao: Since neither `initPrefetchStatusCacheLocked` or getPrefetchStatus logs the error, perhaps log it here before returning NoPrefetch?

go/kbfs/libkbfs/block_retrieval_queue.go
strib: `ForNoDiskCache`?

strib: Slight preference to assigning the new cache to a temp variable first, and only assigning it to `brq.prefetchStatusForTest` once we check the `err` is `nil`.  Just to guard against `lru.New()` returning some non-`nil` broken cache in the event of an error.

go/kbfs/libkbfs/config_local.go
strib: There's no logger available here, but it's a good idea to log the error so you can just make one with `c.MakeLogger("").CDebugf(ctx, ...)`.

go/kbfs/libkbfs/conflict_resolver.go
```diff
@@ -3687,7 +3687,9 @@ func openCRDBInternal(config Config) (*LevelDb, error) {
 func openCRDB(config Config) (db *LevelDb) {
 	db, err := openCRDBInternal(config)
 	if err != nil {
-		panic(fmt.Sprintf("Could not open conflict resolver DB: %v", err))
+		panic(fmt.Sprintf("Could not open conflict resolver DB. "+
+			"Perhaps multiple KBFS instances are being run concurrently"+
+			"? Error: %v", err))
```
strib: `%v` -> `%+v` (to get the stack trace).

go/kbfs/libkbfs/block_retrieval_queue.go
jakob223: @strib any thoughts on size here?

strib: The in-memory block cache uses 10000, so maybe you can copy that here?  It should probably be the same size.

PR #16099, opened 2019-02-18
fix the flip on InboxSource failure

mmaxim:
cc @malgorithms @joshblum


mmaxim:
Ran tests locally, merging

PR #16100, opened 2019-02-18
set ephemeral policy on flips with ephemeral host msg

joshblum:


go/chat/flipmanager.go
```diff
@@ -532,6 +532,21 @@ func (m *FlipManager) StartFlip(ctx context.Context, uid gregor1.UID, hostConvID
 		return err
 	}
 
+	// Preserve the ephemeral lifetime from the conv/message to the game
+	// conversation.
+	if elf, err := utils.EphemeralLifetimeFromConv(ctx, m.G(), hostConv.Conv); err != nil {
```
mmaxim: I think we need to just get the ephemeral lifetime of the message itself right?

joshblum: we don't have access to the message unless we change the commands interface to thread the message through instead of just the body text, what's the benefit of getting it from the message directly?

go/chat/sender.go
```diff
@@ -817,6 +817,19 @@ func (s *BlockingSender) Send(ctx context.Context, convID chat1.ConversationID,
 					clearedCache = true
 				}
 				continue
+			case libkb.ChatEphemeralRetentionPolicyViolatedError:
+				s.Debug(ctx, "Send: failed because of invalid ephemeral policy, trying the whole thing again")
+				if !clearedCache {
+					s.Debug(ctx, "Send: clearing inbox cache to retry getting ephemeral policy")
+					s.G().InboxSource.Clear(ctx, sender)
```
mmaxim: This is a bold move, why not just remote only the GetUnverifiedConv call?

joshblum: yeah i like that better

PR #16101, opened 2019-02-18
Use stellard's NetworkOptions endpoint to get baseFee

patrickxb:
Instead of using build.DefaultBaseFee, use a suggestion from stellard so that it can change dynamically with network conditions.


go/stellar/options.go
```diff
@@ -0,0 +1,52 @@
...
+func NewOptions() *Options {
+	return &Options{}
+}
+
+// NetworkOptions returns stellar1.NetworkOptions that is less than 1 hour old.
+func (o *Options) NetworkOptions(mctx libkb.MetaContext, r remote.Remoter) stellar1.NetworkOptions {
+	o.Lock()
+	defer o.Unlock()
+
+	if time.Since(o.mtime) < 1*time.Hour {
```
zapu: maybe mctx.G().Clock here instead of always real time

PR #16102, opened 2019-02-18
fix two small bugs in ID3

maxtaco:
- color on ignore orange -> green
- put up a "revoked" meta when a social proof was revoked
- also add more debugging if we hit issues


maxtaco:
UGGG CI fail due to something with gen deps

maxtaco:
@chrisnojima  ok to merge this?

PR #16103, opened 2019-02-18
PullFull

joshblum:


joshblum:
just so we didn't have to duplicate the logic on hybrid/remote sources, seemed like a better fit on helper

mmaxim:
Just put it in base

joshblum:
fixed

PR #16104, opened 2019-02-18
generate dependencies before tests

joshblum:
fixes ci failure like: https://ci.keyba.se/job/client/job/PR-16103/1/execution/node/372/log/ introduced in https://github.com/keybase/client/pull/16088


jzila:
I think you forgot to revert this too, so now we gen deps twice.

PR #16114, opened 2019-02-19
move make gen-deps

joshblum:


joshblum:
@jzila why not just generate per os when we need this file as noted in the TODO? i don't know perl and wanted to get this fix out so builds aren't breaking.. also any idea why master has passed on my recent tweaks to this but then PRs start failing? i thought any change to the jenkins file would trigger a full build

jzila:
@joshblum because master builds don't depend on `make gen-deps`, and neither do Jenkinsfile changes.

jzila:
I'm off this week but I can take a look tonight. I'd roll back to the previous `make gen-deps` situation, and I'll make it all work tonight.

joshblum:
oops i misspoke before i meant my PRs with the changes pass but then subsequent prs fail

jzila:
We can't call per-OS as is right now because only Linux machines have perl installed. Running per-OS loses a lot more parallelism since we'll have a bunch of Linux cores unused during dep calculation, and we'll have the slow macos machines busy working calculating dependencies.

PR #16118, opened 2019-02-19
Id3: use icons and sort order from service

buoyad:
- [Go] move `identify3UpdateRow` params into a struct
- Move some rpc data transforms into `constants/tracker2`
- Store `siteIcon` as array and add `priority` with `Assertion`
- Use priority for sort order and network icons

This breaks the icons in the `tracker2` stories. r? @keybase/react-hackers


buoyad:

buoyad:
@keybase/react-hackers bump

buoyad:
@keybase/react-hackers :bump:

shared/tracker2/assertion/index.js
```diff
@@ -273,13 +273,55 @@ const getMenu = p => {
   }
 }
 
+const siteIconToSrcSet = siteIcon =>
+  `-webkit-image-set(${siteIcon
```
chrisnojima: lets just copy here and not slice in those containers

PR #16119, opened 2019-02-19
Use the full channel list for channel suggestions

cjb:
@keybase/react-hackers 

Previously, we only showed channel suggestions for channels that you're in.  This was because we don't *know* anything about channels you're not in unless you've been to the Manage Channels screen for that team, which is where we load the details of all channels for the first time.

This PR tries to balance the cost of loading them with the payoff, and does it this way:

* four seconds after a selectConversation, if the convo is a team chat and we don't already have any channelInfos for it, try to get them.  So this `getChannels` only happens once per team per launch.

* for channel suggestions, prefer the full details from teamNameToChannelInfos if they exist, else just use the inbox-based suggestions that we use today.

So, if someone hits the "#" in the first four seconds of going to a team channel for the first time for that team that launch, they'll see a smaller list of suggestions that becomes a larger list in a few seconds' time.

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -1955,10 +1955,8 @@ function* loadChannelInfos(state, action) {
     return
   }
   if (!TeamsConstants.hasChannelInfos(state, teamname)) {
-    yield Saga.sequentially([
-      Saga.callUntyped(Saga.delay, 4000),
-      Saga.put(TeamsGen.createGetChannels({teamname})),
-    ])
+    yield Saga.callUntyped(Saga.delay, 4000)
+    yield Saga.put(TeamsGen.createGetChannels({teamname}))
   }
 }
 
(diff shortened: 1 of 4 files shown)


shared/actions/chat2/index.js
```diff
@@ -1941,11 +1941,27 @@ const loadCanUserPerform = (state, action) => {
...
+// Get the full channel names/descs for a team if we don't already have them.
+function* loadChannelInfos(state, action) {
+  const {conversationIDKey} = action.payload
+  const meta = Constants.getMeta(state, conversationIDKey)
+  const teamname = meta.teamname
+  if (!teamname) {
+    return
+  }
+  if (!TeamsConstants.hasChannelInfos(state, teamname)) {
+    yield Saga.sequentially([
```
chrisnojima: we don't need this concept in generators themselves, you can just yield the delay then yield the put

PR #16120, opened 2019-02-19
switch flip protocol to its own message, and other cleanup CORE-10216

mmaxim:
1. improve the parsing of big ints
2. change range delimiter to ..
3. change to use the `FLIP` message type for all flip related activity.
4. remove option to edit/delete flips


go/chat/flipmanager.go
```diff
@@ -544,7 +542,11 @@ func (m *FlipManager) StartFlip(ctx context.Context, uid gregor1.UID, hostConvID
 	if err != nil {
 		return err
 	}
-	if err := m.G().ChatHelper.SendTextByID(ctx, conv.GetConvID(), tlfName, string(infoBody)); err != nil {
+	if err := m.G().ChatHelper.SendMsgByID(ctx, conv.GetConvID(), tlfName,
+		chat1.NewMessageBodyWithFlip(chat1.MessageFlip{
+			Text:   string(infoBody),
+			GameID: gameID,
```
joshblum: drop `gameID` from `hostMessageInfo` since it's part of the message now?

mmaxim: Hmm, yeah good point.

mmaxim: Actually I'm going to keep it there and double check messages against it.

mmaxim:
Ran tests locally

PR #16121, opened 2019-02-19
Allow paperkey generation in oneshot mode

joshblum:
needed to prompt the user for their passphrase to perform lksec operations

### DIFF
--- a/go/engine/login_oneshot_test.go
+++ b/go/engine/login_oneshot_test.go
@@ -43,9 +43,23 @@ func TestLoginOneshot(t *testing.T) {
 	assertNumDevicesAndKeys(tc, fu, 2, 4)
 	err = AssertProvisioned(tc2)
 	require.NoError(t, err)
+
+	// assert paper key generation works
+	uis = libkb.UIs{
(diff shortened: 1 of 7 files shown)


patrickxb:
It would be nice to not change the tests (or change as little as possible) when changing something like this so we could see that the tests that have been running for ages still pass, instead of also wondering if the new tests are the same as the old tests.  In a separate PR you could change the tests...

patrickxb:
Can this ever run with a GUI?  

Because we can't just prompt for a passphrase whenever we want with the GUI.

joshblum:
no problem, i'll split the test refactor to a separate pr. i tested with the gui/cli and both work with the password prompt

patrickxb:
What about mobile?  

I guess there's no way to do one-shot in mobile...mobile definitely doesn't register a secret ui, though...

joshblum:
the only way to get into oneshot is through the cli.. should i add some special casing for mobile?

patrickxb:
I think so just to be safe.

joshblum:
fixed

PR #16128, opened 2019-02-19
Changes to support multi-op txs

patrickxb:
Enforce that accountID is provided for tx details when possible.
Use the includeMulti flag.


protocol/avdl/stellar1/local.avdl
```diff
@@ -178,10 +178,13 @@ protocol local {
 
   // accountID should be the accountID that you want the details for.
   // This is used for formatting the `amountDescription` to know if it
-  // is a payment that increases or decreases the balance. AccountID
-  // is optional, but then `amountDescription` will not be formatted
-  // with "+" or "-", and also `delta` will always be NONE.
-  PaymentDetailsLocal getPaymentDetailsLocal(int sessionID, union { null, AccountID } accountID, PaymentID id);
+  // is a payment that increases or decreases the balance. 
+  PaymentDetailsLocal getPaymentDetailsLocal(int sessionID, AccountID accountID, PaymentID id);
```
mlsteele: check frontend is relying on the existing behavior for anything too important

patrickxb: I checked with @buoyad before making this change.

PR #16129, opened 2019-02-19
allow new conversation membership to be sourced by a convid

joshblum:
server support to follow


joshblum:
depends on https://github.com/keybase/keybase/pull/3486

go/chat/flipmanager_test.go
```diff
@@ -75,7 +82,53 @@ func TestFlipManagerStartFlip(t *testing.T) {
 				consumeNewMsgRemote(t, listener2, chat1.MessageType_SYSTEM)
 			}
 
+			// ensure we only include the channel participants in the flip for team
+			// conversations.
+			expectedDevConvs := 0
```
mmaxim: Can we move this to a different test that doesn't run so many times? It seems independent of members type.

joshblum: yeah good idea

PR #16134, opened 2019-02-20
add reset banner back to desktop

songgao:
This is a quick fix; I'll have a better version when my KBFS-3878 is ready.


songgao:
Thanks @mmaxim !

PR #16135, opened 2019-02-20
fix message test

songgao:


shared/constants/chat2/message.js
```diff
@@ -113,6 +113,7 @@ export const serviceMessageTypeToMessageTypes = (t: RPCChatTypes.MessageType): A
     case RPCChatTypes.commonMessageType.deletehistory:
     case RPCChatTypes.commonMessageType.reaction:
     case RPCChatTypes.commonMessageType.unfurl:
+    case RPCChatTypes.commonMessageType.flip:
```
songgao: I added `flip` here, but when do we need to map it to a different name?

PR #16141, opened 2019-02-20
nuke db on too many open files error, increase compaction size

joshblum:
- try to catch too many open files errors and nuke the db
- increase table compaction size from 2mb to 50mb so we don't end up with so many small tables/files 

cc @heronhaye

### DIFF
--- a/go/libkb/leveldb.go
+++ b/go/libkb/leveldb.go
@@ -116,6 +116,7 @@ func (l *LevelDb) Opts() *opt.Options {
 	return &opt.Options{
 		OpenFilesCacheCapacity: l.G().Env.GetLevelDBNumFiles(),
 		Filter:                 filter.NewBloomFilter(10),
+		CompactionTableSize:    50 * opt.MiB,
 	}
 }
 
@@ -166,7 +167,7 @@ func (l *LevelDb) doWhileOpenAndNukeIfCorrupted(action func() error) (err error)
 
 	// Notably missing here is the error handling for when DB open fails but on
 	// an error other than "db is corrupted". We simply return the error here
-	// without resetting `dbOpenerOcce` (i.e. next call into LevelDb would result
+	// without resetting `dbOpenerOnce` (i.e. next call into LevelDb would result
 	// in a LevelDBOpenClosedError), because if DB open fails, retrying it
 	// wouldn't help. We should find the root cause and deal with it.
 	// MM: 10/12/2017: I am changing the above policy. I am not so sure retrying it won't help,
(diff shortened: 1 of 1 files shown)


go/libkb/leveldb.go
```diff
@@ -241,21 +242,25 @@ func (l *LevelDb) isCorrupt(err error) bool {
...
-	// necesary to block other doWhileOpenAndNukeIfCorrupted() calls.
+	// necessary to block other doWhileOpenAndNukeIfCorrupted() calls.
 	defer l.Unlock()
 	defer l.G().Trace("LevelDb::Nuke", func() error { return err })()
 
+	// even if we can't close the db try to nuke the files directly
 	err = l.closeLocked()
 	if err != nil {
-		return "", err
+		l.G().Log.Debug("Error closing leveldb %v, attempting nuke anyway", err)
```
mlsteele: is this safe?

go/libkb/leveldb.go
```diff
@@ -241,21 +242,25 @@ func (l *LevelDb) isCorrupt(err error) bool {
 	if strings.Contains(err.Error(), "corrupt") {
 		return true
 	}
-
+	// if our db is in a bad state with too many open files also nuke
+	if strings.Contains(strings.ToLower(err.Error()), "too many open files") {
+		return true
+	}
```
mlsteele: I don't think running out of fds is all that uncommon. It would be nice to only do this if leveldb is actually the problem. Knowing how many fds leveldb tried to open would be ideal but that's probably annoying. Could we try to open some random other file, and only if that single open succeeds, return true?

go/libkb/leveldb.go
joshblum: even the close couldn't completely successful i don't think it matters since we blow away all of the files right after this. our local state is valid since we set the `db=nil` even if the close failed: https://github.com/keybase/client/blob/6e033eaa414f7d9fa45b5713cd781076c9fdf287/go/libkb/leveldb.go#L221

heronhaye: does this log people out? or only if they restart keybase?

go/libkb/leveldb.go
heronhaye: good idea (since this leveldb fd spike seems to be temporary, I guess you can't lsof it or anything)

go/libkb/leveldb.go
joshblum: doesn't affect their session just the sigchain/team/chat caches

go/libkb/leveldb.go
joshblum: for simplicity i'd prefer to leave as is. loosing a db isn't the end of the world, if a user is really out of fds nuking the db might free it up and at least allow a log send which wasn't possible before

PR #16142, opened 2019-02-20
new docs for new cert

maxtaco:


maxtaco:
@mmaxim I cut a new version of the installer with the new cert.

PR #16152, opened 2019-02-20
show replay errors properly

mmaxim:
I think this hopefully makes it such that we never accidentally plumb up a replay error for an active game.


mmaxim:
Ran tests locally

PR #16156, opened 2019-02-20
folder_branch_status: print MD timestamp in KBFS status

strib:
This will print the string of the server's timestamp for the MD object, but adjusted for the time skew against the local clock.  Example:

```
$ cat /keybase/private/strib/.kbfs_status
{
  "Staged": false,
 ...
  "LocalTimestamp": "2019-02-19T10:01:21.263666333-08:00",
 ...
}
```

### DIFF
--- a/go/kbfs/libdokan/mount_test.go
+++ b/go/kbfs/libdokan/mount_test.go
@@ -2454,6 +2454,15 @@ func TestStatusFile(t *testing.T) {
 	var bufStatus libkbfs.FolderBranchStatus
 	json.Unmarshal(buf, &bufStatus)
 
+	// Use a fuzzy check on the timestamps, since it could include
+	// monotonic clock stuff.
+	if !timeEqualFuzzy(
+		status.LocalTimestamp, bufStatus.LocalTimestamp, time.Millisecond) {
+		t.Fatalf("Local timestamp (%s) didn't match expected timestamp %v",
+			bufStatus.LocalTimestamp, status.LocalTimestamp)
+	}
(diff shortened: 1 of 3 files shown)


kg4zow:
Would it also be possible to add the same value, without any "localtime" translation? i.e. a straight unix timestamp?

kg4zow:
Also, are the timestamps for each rev going to be available separately through `.kbfs_archived_rev=nnn/.kbfs_status`? (I don't know what an "MD" is.)

strib:
> Would it also be possible to add the same value, without any "localtime" translation? i.e. a straight unix timestamp?

Depends what you're asking for exactly.  When I say "local timestamp" here, I mean that it's adjusted based on the measured clock skew between the local device and the server.  That way, the timestamp makes sense when you compare it to the local clock.  (i.e., if you want to check how long ago the revision happened.)  It's kind of hard to undo that adjustment after the fact, since the clock skew might have changed.

But if all you want is the unix timestamp of that local-adjust time, I could easily add that.  But it seems redundant, when you could easily convert it to a unix timestamp like this:

```sh
date --date=`jq -r .LocalTimestamp .kbfs_status` +%s
```

> Also, are the timestamps for each rev going to be available separately through `.kbfs_archived_rev=nnn/.kbfs_status`? (I don't know what an "MD" is.)

Yes. There is one MD object associated with each revision, and the data we're printing in the status is mostly from the MD for that revision.  Each archived rev has its own MD, and that's what will be used to make it's `.kbfs_status`.

PR #16162, opened 2019-02-21
temporary fix for fs tab on mobile

songgao:
This is just a quick fix that brings the Files tab back by disabling the reset banner on mobile (thus removing the scrollview). The real and right fix is covered by https://github.com/keybase/client/pull/16161 which makes the banners part of the list. If that one doesn't get merged before the release, we should merge this to unbrick Files on mobile.


songgao:
Thanks Mike!

PR #16164, opened 2019-02-21
Offer signup link if someone types a non-existent username into username field

cjb:
@keybase/react-hackers CC @keybase/design 

This PR fixes two issues:

* giving an incorrect login username just errors out the provision completely, doesn't let you correct your mistake on the same screen

* you might accidentally type your desired *signup* username into this field, and in that case it'd be nice to offer a link to the signup flow instead

By:

* adding a new `inlineError` field.  Provisioning is controlled by the daemon and we usually get a `finalError` if we're going to bail out and show the error screen.  The action now detects that it was the user-not-found error and creates an `inlineError` instead.  In the future we could decide to push more error types through this new inline flow.

* We had almost identical duplicated components for desktop and mobile -- unify them and use `withStateHandlers` instead of explicit state funcs.

* We want to show a link to the signup flow in the inline error, but `<Input errorText={} />` doesn't allow links, so I added an `errorTextComponent` prop to `Input` to allow putting in a link, and manually styled it.  (An alternative was just having this inline error be outside of the `<Input />` completely.)


shared/provision/username-or-email/index.js
```diff
@@ -0,0 +1,114 @@
...
+import * as Kb from '../../common-adapters'
+import * as Styles from '../../styles'
+import Container from '../../login/forms/container'
+import * as Constants from '../../constants/provision'
+import {RPCError} from '../../util/errors'
+import {constantsStatusCode} from '../../constants/types/rpc-gen'
+
+type Props = {
+  error: string,
+  inlineError: ?RPCError,
```
chrisnojima: lets not expose a raw error here. the container can expose inlineErrorText or badUsernameError boolean or something

shared/provision/username-or-email/index.js
```diff
@@ -0,0 +1,114 @@
...
+// TODO remove Container
+import * as React from 'react'
+import * as Kb from '../../common-adapters'
+import * as Styles from '../../styles'
+import Container from '../../login/forms/container'
+import * as Constants from '../../constants/provision'
+import {RPCError} from '../../util/errors'
+import {constantsStatusCode} from '../../constants/types/rpc-gen'
+
+type Props = {
```
chrisnojima: props should be exact

shared/provision/username-or-email/container.js
```diff
@@ -24,5 +27,11 @@ export default compose(
     dispatchToProps,
     (s, d, o) => ({...o, ...s, ...d})
   ),
-  safeSubmit(['onBack', 'onSubmit'], ['error'])
+  withStateHandlers(
```
chrisnojima: should likely not use recompose and just move this state into the component. its not used at all by this container

shared/provision/username-or-email/index.js
```diff
@@ -0,0 +1,114 @@
...
+      alignItems: 'center',
+      justifyContent: 'center',
+    },
+  }),
+  inlineErrorText: Styles.platformStyles({
+    common: {
+      color: Styles.globalColors.red,
+    },
+    isElectron: {
+      textDecoration: 'underline',
```
chrisnojima: can just use textDecorationLine and move to common

shared/provision/username-or-email/container.js
```diff
@@ -22,7 +27,14 @@ export default compose(
   connect<OwnProps, _, _, _, _>(
     mapStateToProps,
     dispatchToProps,
-    (s, d, o) => ({...o, ...s, ...d})
+    (stateProps, dispatchProps, _) => ({
```
chrisnojima: if this does anything it should get pulled out into a mergeProps function

shared/provision/username-or-email/index.js
```diff
@@ -0,0 +1,116 @@
...
+    </Kb.Text>
+  </Kb.Box2>
+)
+
+class UsernameOrEmail extends React.Component<Props, State> {
+  state: State
+
+  constructor(props: Props) {
+    super(props)
+    this.state = {usernameOrEmail: ''}
```
chrisnojima: don't need this constructor

cjb: Huh -- how would `this.state` be initialized without it?

chrisnojima: `state = {usernameOrEmail: ''}` above

PR #16166, opened 2019-02-21
Unskip re-add tests

mlsteele:
This should fix the race between `require.Equal(t, expectedRole, role, "expected role")` and the invite->cryptomember background task.

### DIFF
--- a/go/systests/team_reset_test.go
+++ b/go/systests/team_reset_test.go
@@ -2,6 +2,7 @@ package systests
 
 import (
 	"testing"
+	"time"
 
 	"golang.org/x/net/context"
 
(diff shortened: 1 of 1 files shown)


go/systests/team_reset_test.go
```diff
@@ -660,11 +660,30 @@ func testTeamReAddAfterReset(t *testing.T, pukful, adminOwner bool) {
...
+		if role == expectedRole {
+			return true
+		}
+		require.FailNowf(t, "unexpected role", "got %v on the hunt for %v", role, expectedRole)
+		return false
+	}
+
+	if pukful {
+		// Bob should have been synchronously made a cyrptomember by re-add.
+		require.Equal(t, true, pollFn(0))
```
zapu: nice

PR #16168, opened 2019-02-21
Fix available to send

patrickxb:
cc @buoyad


zapu:
Can we also fix
`AccountDetailsToWalletAccountLocal`

```
	availableInt, err := stellarnet.ParseStellarAmount(details.Available)
	if err != nil {
		return empty, err
	}
	canSubmitTx := availableInt > 100 // base fee is 100
```

zapu:
but it's used for inflation only right now. so maybe should be fixed when inflation endpoint is changed to use new fee

patrickxb:
ok, i'll make a ticket for inflation setting.  thanks.

PR #16171, opened 2019-02-21
speculative iOS keychain deletion fix

maxtaco:
- it's a conjecture that the old keychain code didn't actually delete keychain items if their encodedSecret value differed
- this prevented some people, possibly, from reinstalling the app and being able to use the secret store, since their old value was never deleted
- we'll know for sure if this helps out affected users


patrickxb:
We should get this in iOS, macOS admin builds ASAP to try it out.

maxtaco:
ok, builds running now

PR #16175, opened 2019-02-21
fast path flip messages in dev conversations to bypass any state updates CORE-10266

mmaxim:
The goal is to make it so we can process these messages faster, to reduce chances of failed flips.

### DIFF
--- a/go/chat/flipmanager.go
+++ b/go/chat/flipmanager.go
@@ -639,6 +639,10 @@ func (m *FlipManager) MaybeInjectFlipMessage(ctx context.Context, boxedMsg chat1
 	if err != nil {
 		m.Debug(ctx, "MaybeInjectFlipMessage: failed to unbox: %s", err)
 	}
+	if err := storage.New(m.G(), nil).SetMaxMsgID(ctx, convID, uid, msg.GetMessageID()); err != nil {
+		m.Debug(ctx, "MaybeInjectFlipMessage: failed to write max msgid: %s", err)
+		// charge forward from this error
+	}
 	// Ignore anything from the current device
 	sender := flip.UserDevice{
 		U: msg.Valid().ClientHeader.Sender,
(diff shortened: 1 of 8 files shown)


mmaxim:
cc @maxtaco

go/chat/flipmanager.go
```diff
@@ -616,29 +616,47 @@ func (m *FlipManager) StartFlip(ctx context.Context, uid gregor1.UID, hostConvID
...
+		m.Debug(ctx, "MaybeInjectFlipMessage: failed to update inbox version: %s", err)
+		// charge forward here, we will figure it out
+	}
+	// Unbox the message
+	conv, err := utils.GetUnverifiedConv(ctx, m.G(), uid, convID, types.InboxSourceDataSourceAll)
+	if err != nil {
+		m.Debug(ctx, "MaybeInjectFlipMessage: failed to get conversation for unbox: %s", err)
+		return true
+	}
+	msg, err := NewBoxer(m.G()).UnboxMessage(ctx, boxedMsg, conv.Conv, nil)
```
joshblum: won't these won't get exploded since we don't merge into storage?

go/chat/push.go
```diff
@@ -501,6 +501,13 @@ func (g *PushHandler) Activity(ctx context.Context, m gregor.OutOfBandMessage) (
 				g.Debug(ctx, "chat activity: newMessage: outboxID is empty")
 			}
 
+			// Coin flip manager can completely handle the incoming message
+			if g.G().CoinFlipManager.MaybeInjectFlipMessage(ctx, nm.Message, nm.InboxVers, uid, nm.ConvID,
+				nm.TopicType) {
+				g.Debug(ctx, "chat activity: flip message handled, early out")
+				return
```
joshblum: why not allow the rest of the processing so we don't have to worry about weird state with these functions not being called. could return a `*chat1.MessageUnboxed` and have a `ConvSource.PushUnboxed`?

joshblum: (or just unbox twice if we're not worried about the perf there)

go/chat/flipmanager.go
mmaxim: Why do we care then?

go/chat/push.go
mmaxim: The point is to skip this processing, I don't think it is necessary for these. The only thing we absolutely need to do is update the inbox version, which this guy does.

mmaxim:
Ran locally

PR #16177, opened 2019-02-21
folder_block_manager: cancel all prefetches before cleaning cache

strib:
When `folderBranchOps` applies the changes from a new revision, it goes through each updated pointer, and cancels prefetches for the old version of the pointer.  But the pointers aren't ordered in anyway, so this scenario is possible, when block A points to block B:

1) Cancel prefetch for B.
2) Prefetch for A starts, which schedules a prefetch for B.
3) Cancel prefetch for A.
4) Cache-cleaning completes.
5) Prefetch for B starts, and block B is put into the cache after it was cleaned.

To fix this, let's do a second round of prefetch-cancellations before cleaning cache, and wait for each one individually to complete.  These two cancellation loops together should be enough to ensure that the block won't end up back in the cache.

Also, cancel prefetches for unreferenced (not just updated) blocks, which was an oversight in the existing code.

Issue: KBFS-3784

### DIFF
--- a/go/kbfs/libkbfs/folder_block_manager.go
+++ b/go/kbfs/libkbfs/folder_block_manager.go
@@ -1379,6 +1379,30 @@ func (fbm *folderBlockManager) doCleanDiskCache(cacheType DiskBlockCacheType) (
 		// archive races (see comment below).
 		ptrs := getUnrefPointersFromMD(rmd, true)
 
+		// Cancel any prefetches for these blocks that might be in
+		// flight, to make sure they don't get put into the cache
+		// after we're done cleaning it.  Ideally we would cancel them
+		// in a particular order (the lowest level ones first, up to
+		// the root), but since we already do one round of
+		// prefetch-canceling as part of applying the MD and updating
+		// the pointers, doing a second round here should be good
+		// enough to catch any weird relationships between the
+		// pointers where one non-yet-canceled prefetch can revive the
+		// prefetch of an already-canceled child block.
+		for _, ptr := range ptrs {
+			fbm.config.BlockOps().Prefetcher().CancelPrefetch(ptr)
+			c, err := fbm.config.BlockOps().Prefetcher().
+				WaitChannelForBlockPrefetch(ctx, ptr)
+			if err != nil {
+				return err
+			}
+			select {
+			case <-c:
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
+
 		var ids []kbfsblock.ID
 		if cacheType == DiskBlockSyncCache {
 			// Wait for our own archives to complete, to make sure the
--- a/go/kbfs/libkbfs/folder_block_manager_test.go
+++ b/go/kbfs/libkbfs/folder_block_manager_test.go
@@ -685,6 +685,7 @@ func TestFolderBlockManagerCleanSyncCache(t *testing.T) {
 	var userName kbname.NormalizedUsername = "test_user"
 	config, _, ctx, cancel := kbfsOpsInitNoMocks(t, userName)
 	defer kbfsTestShutdownNoMocks(t, config, ctx, cancel)
+	config.vdebugSetting = "vlog2"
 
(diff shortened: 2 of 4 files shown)


go/kbfs/libkbfs/folder_block_manager.go
```diff
@@ -1379,6 +1379,30 @@ func (fbm *folderBlockManager) doCleanDiskCache(cacheType DiskBlockCacheType) (
...
+				WaitChannelForBlockPrefetch(ctx, ptr)
+			if err != nil {
+				return err
+			}
+			select {
+			case <-c:
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
```
jakob223: Why wait one at a time instead of waiting for all of them? 

I don't see why this should improve the probability of something triggering another and getting through.

go/kbfs/libkbfs/folder_block_manager_test.go
```diff
@@ -685,6 +685,7 @@ func TestFolderBlockManagerCleanSyncCache(t *testing.T) {
 	var userName kbname.NormalizedUsername = "test_user"
 	config, _, ctx, cancel := kbfsOpsInitNoMocks(t, userName)
 	defer kbfsTestShutdownNoMocks(t, config, ctx, cancel)
+	config.vdebugSetting = "vlog2"
```
jakob223: Should this be a constant?

strib: Probably but that ship may have sailed.  It's over in libkb and probably not appropriate to change in this PR.

go/kbfs/libkbfs/folder_block_manager.go
strib: It's harder to code a select across a variable number of channels, and I don't think it would improve performance much, since the prefetcher is single-goroutined anyway.  This way seems both easier to understand and easier to code.

strib: Also, performance is not really a concern here anyway, since it's a background goroutine that shouldn't ever block the user.

strib:
> OOC how did you catch this?

There was a test CI flake, and I ran it 1000 times locally to repro, adding logging when needed, and reasoned through what was happening.

go/kbfs/libkbfs/folder_block_manager.go
jakob223: Didn't think about the prefetcher being single threaded, seems reasonable.

go/kbfs/libkbfs/folder_block_manager.go
```diff
@@ -1379,6 +1379,30 @@ func (fbm *folderBlockManager) doCleanDiskCache(cacheType DiskBlockCacheType) (
...
+		// the root), but since we already do one round of
+		// prefetch-canceling as part of applying the MD and updating
+		// the pointers, doing a second round here should be good
+		// enough to catch any weird relationships between the
+		// pointers where one non-yet-canceled prefetch can revive the
+		// prefetch of an already-canceled child block.
+		for _, ptr := range ptrs {
+			fbm.config.BlockOps().Prefetcher().CancelPrefetch(ptr)
+			c, err := fbm.config.BlockOps().Prefetcher().
+				WaitChannelForBlockPrefetch(ctx, ptr)
```
jzila: What's stopping something else from adding one of these pointers back to the prefetcher before we're done cleaning them?

strib: What would do so?  At this point, the pointer isn't accessible from the main folder anymore, so nothing should be prefetching it.

jzila: Browsing the history, for example.

strib: Archived data isn't cached because of: https://github.com/keybase/kbfs/pull/1919

jzila: Good call, thanks.

PR #16179, opened 2019-02-21
don't leak exploded message in snippet

buoyad:
When a message explodes via timeout (not explode now) the snippet updates but the inbox version doesn't. The recently new `updateMeta` function doesn't use the new meta in this case, so we started to leak the exploded message in the snippet. This changes `updateMeta` so if:
```
old.inboxVersion === new.inboxVersion && new.trustedState === 'trusted'
```
we take the new meta. This will take the new meta more often than the old condition which was
```
old.inboxVersion === new.inboxVersion && old.trustedState === 'untrusted' && new.trustedState === 'trusted'
```

Other options if we don't want to make this change:
* Increment the inboxVersion when this happens 🤞
* Signal that this meta update is due to an explode and take the new meta indiscriminately
* Keep the old logic, but accept the new snippet/decoration no matter what

Thoughts? @chrisnojima @mmaxim @joshblum

### DIFF
--- a/shared/constants/chat2/meta.js
+++ b/shared/constants/chat2/meta.js
@@ -123,17 +123,17 @@ export const updateMeta = (
     // new is older, keep old
     return oldMeta
   } else if (oldMeta.inboxVersion === newMeta.inboxVersion) {
-    // same version, only take data if untrusted -> trusted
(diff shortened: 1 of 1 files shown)


buoyad:
Changed to make the most minimal change until the local inbox version exists. r? @keybase/react-hackers

mmaxim:
Thanks @buoyad !

PR #16185, opened 2019-02-22
generic framework for offline RPCs

maxtaco:
- plug in resolve3
- some basic tests show it works

### DIFF
--- a/go/offline/rpc_cache.go
+++ b/go/offline/rpc_cache.go
@@ -57,8 +57,15 @@ func dbKey(rpcName string, arg interface{}) (libkb.DbKey, error) {
 
 }
 
-func (c *RPCCache) get(m libkb.MetaContext, rpcName string, encrypted bool, arg interface{}, res interface{}) (found bool, err error) {
-	defer m.CTraceString(fmt.Sprintf("RPCCache#get(%s, %v, %+v)", rpcName, encrypted, arg), func() string { return fmt.Sprintf("(%v,%v)", found, err) })()
+type Version int
+
+type Value struct {
+	Version Version
+	Data    []byte
+}
+
+func (c *RPCCache) get(mctx libkb.MetaContext, version Version, rpcName string, encrypted bool, arg interface{}, res interface{}) (found bool, err error) {
+	defer mctx.CTraceString(fmt.Sprintf("RPCCache#get(%d, %s, %v, %+v)", version, rpcName, encrypted, arg), func() string { return fmt.Sprintf("(%v,%v)", found, err) })()
 	c.Lock()
 	defer c.Unlock()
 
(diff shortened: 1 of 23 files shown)


maxtaco:
cc @strib this has support for offline-ready `resolve3`, does it fit what's needed?

go/offline/rpc_cache.go
```diff
@@ -0,0 +1,128 @@
...
+
+	dbk, err := dbKey(rpcName, arg)
+	if err != nil {
+		return err
+	}
+
+	if encrypted {
+		err = c.edb.Put(m.Ctx(), dbk, res)
+	} else {
+		err = m.G().LocalDb.PutObj(dbk, nil, res)
```
joshblum: can you use the `GetRaw`/`PutRaw` instead so we don't json encode these? i want to cleanup the interface at some point to make it more clean how the data is stored and only use json where we need that encoding

maxtaco: Thanks for looking at the PR! Is there an interface that msgpacks and then writes binary (and the opposite on the way out?)

joshblum: your change looks great. i'll cut a ticket to see if the speed/storage benefits of msgpack are worthwhile to migrate existing calls since we'll have to do some kind of migration or have all clients start with blank caches

go/offline/rpc_cache.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func dbKey(rpcName string, arg interface{}) (libkb.DbKey, error) {
+	raw, err := hash(rpcName, arg)
+	if err != nil {
+		return libkb.DbKey{}, err
+	}
+	return libkb.DbKey{
+		Typ: libkb.DBOfflineRPC,
+		Key: hex.EncodeToString(raw[0:16]),
```
mlsteele: Basically storing sha256 of rpc args in plain text on the hdd. That gonna be ok?

go/offline/rpc_cache.go
```diff
@@ -0,0 +1,127 @@
...
+	}, nil
+
+}
+
+func (c *RPCCache) get(m libkb.MetaContext, rpcName string, encrypted bool, arg interface{}, res interface{}) (found bool, err error) {
+	defer m.CTraceString(fmt.Sprintf("RPCCache#get(%s, %v, %+v)", rpcName, encrypted, arg), func() string { return fmt.Sprintf("(%v,%v)", found, err) })()
```
mlsteele: Logging whole `arg` in something so general seems apt to accidentally log something we'd rather not have in log sends

go/offline/rpc_cache.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if encrypted {
+		err = c.edb.Put(m.Ctx(), dbk, res)
+	} else {
+		err = m.G().LocalDb.PutObjMsgpack(dbk, nil, res)
+	}
+	return err
+}
+
+func (c *RPCCache) Serve(m libkb.MetaContext, oa keybase1.OfflineAvailability, rpcName string, encrypted bool, arg interface{}, res interface{},
```
mlsteele: m->mctx

go/service/identify.go
```diff
@@ -145,10 +147,18 @@ func (h *IdentifyHandler) identifyLiteUser(netCtx context.Context, arg keybase1.
...
-	defer h.G().CTrace(ctx, fmt.Sprintf("IdentifyHandler#Resolve3(%s)", arg), func() error { return err })()
-	return h.resolveUserOrTeam(ctx, arg)
+func (h *IdentifyHandler) Resolve3(ctx context.Context, arg keybase1.Resolve3Arg) (ret keybase1.UserOrTeamLite, err error) {
+	mctx := libkb.NewMetaContext(ctx, h.G()).WithLogTag("RSLV")
+	defer mctx.CTrace(fmt.Sprintf("IdentifyHandler#Resolve3(%+v)", arg), func() error { return err })()
+	retp := &ret
+	err = h.service.offlineRPCCache.Serve(mctx, arg.Oa, "identify.resolve3", false, arg, retp, func(mctx libkb.MetaContext) (interface{}, error) {
+		tmp, err := h.resolveUserOrTeam(mctx.Ctx(), arg.Assertion)
+		if err == nil {
+			*retp = tmp
```
mlsteele: can Serve do this?

go/offline/rpc_cache.go
maxtaco: Good point

go/service/identify.go
maxtaco: not with type safety

maxtaco: ... or type reflection .....

go/offline/rpc_cache.go
maxtaco: yeah, i was hoping it was ok, and that the marshaling of msgpack is deterministic!

maxtaco:
Great point about versioning. I thought about that at one point and then forgot about it

go/service/identify.go
mlsteele: oh that's funny. So the offline pathway doesn't actually use this statement but yet this statement kinda ties the types all together.

mlsteele: we should just switch to rust

maxtaco:
I think we're ok with the inexactness of our offline/online check, and also races.  I think the "best effort" notion, at first, is good enough, and we'll iterate if not.

PR #16190, opened 2019-02-22
Revert Revert virtualization fix

chrisnojima:
This undoes the revert i had earlier. @mmaxim can't repro nor can i so i say we get this in after the release


chrisnojima:
@keybase/react-hackers

chrisnojima:
- [ ] after release

PR #16191, opened 2019-02-22
allow loadTeamPlusApplicationKeys to hit the offline RPC cache

maxtaco:
- don't add testing, but it should be covered mainly by existing tests and independent tests to the offline RPC system

### DIFF
--- a/go/service/teams.go
+++ b/go/service/teams.go
@@ -20,18 +20,18 @@ import (
 type TeamsHandler struct {
 	*BaseHandler
 	globals.Contextified
-	gregor *gregorHandler
-	connID libkb.ConnectionID
+	connID  libkb.ConnectionID
+	service *Service
 }
 
 var _ keybase1.TeamsInterface = (*TeamsHandler)(nil)
 
-func NewTeamsHandler(xp rpc.Transporter, id libkb.ConnectionID, g *globals.Context, gregor *gregorHandler) *TeamsHandler {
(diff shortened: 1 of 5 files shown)


maxtaco:
cc @strib

go/service/teams.go
```diff
@@ -469,8 +469,28 @@ func (h *TeamsHandler) LoadTeamPlusApplicationKeys(ctx context.Context, arg keyb
...
+		tmp, err := teams.LoadTeamPlusApplicationKeys(ctx, h.G().ExternalG(), arg.Id, arg.Application, arg.Refreshers,
+			arg.IncludeKBFSKeys)
+		if err == nil {
+			*resp = tmp
+		}
+		return tmp, err
+	}
+
+	// argKey is a copy of arg that's going to be used for a cache key, so clear out
+	// refreshers and sessionID, since they don't affect the cache key value.
```
mlsteele: This doesn't mesh with what refreshers currently mean.
```
    // Load at least up to the keygen. Returns an error if the keygen is not loaded.
    PerTeamKeyGeneration needKeyGeneration;
```

protocol/avdl/keybase1/teams.avdl
```diff
@@ -1033,7 +1033,7 @@ protocol teams {
    * If refreshers are non-empty, then force a refresh of the cache if the requirements
    * of the refreshers aren't met.
```
mlsteele: "force a refresh of the cache if the requirements"
unless doing the offline thing, then might return old stuff

PR #16196, opened 2019-02-22
Show placeholder text for empty followers/following on mobile profile view

cjb:
@keybase/react-hackers

This was more involved than I expected.

~It looks like mobile's using identifyv3 (`profile/user/`), but desktop's still using the old mechanisms and store (via `friendships.desktop.js`).  Sounds like we should file a ticket for migrating the Followers/Following on desktop to identifyv3?~

I had to find a way to distinguish between a tracker2 profile being loaded or not loaded yet, and went with `followersCount === null` as the difference.

And there's now a sentinel list entry for the `profile/user` list renderer to use to show the placeholder text.

### DIFF
--- a/shared/profile/user/index.js
+++ b/shared/profile/user/index.js
@@ -22,7 +22,9 @@ export type Props = {|
   backgroundColorType: BackgroundColorType,
   followThem: boolean,
   followers: Array<string>,
+  followersCount: ?number,
   following: Array<string>,
+  followingCount: ?number,
   onBack: () => void,
   onReload: () => void,
   onSearch: () => void,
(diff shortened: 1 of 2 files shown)


shared/profile/user/index.js
```diff
@@ -307,7 +319,13 @@ class User extends React.Component<Props, State> {
     const friends = this.state.selectedFollowing ? this.props.following : this.props.followers
     const {itemsInARow, itemWidth} = widthToDimentions(this.state.width)
     // TODO memoize?
-    const chunks = this.state.width ? chunk(friends, itemsInARow) : []
+    let chunks = this.state.width ? chunk(friends, itemsInARow) : []
+    if (chunks.length === 0 && this.props.followingCount !== null && this.props.followingCount !== null) {
+      chunks.push({
+        text: this.state.selectedFollowing ? 'You are not following anyone.' : 'You have no followers.',
+        type: 'dummy',
```
chrisnojima: lets name this `noFriends` or something

PR #16201, opened 2019-02-22
Replace text prop backgroundMode with negative

buoyad:
I checked some things in the app and all the changed stories (only 5!), things seem ok. Cc @cecileboucheron if you want to check anything out. r? @keybase/react-hackers 

edit: This also makes `Text`'s props exact, which uncovered a bunch of unused props we had.


chrisnojima:
- [ ] after release

shared/common-adapters/text.js.flow
```diff
@@ -14,15 +14,15 @@ type Background =
 
 type TextType = $Keys<typeof allTextTypes>
 
-type Props = {
+type Props = {|
```
chrisnojima: awww yisss

shared/wallets/airdrop/qualify/index.js
```diff
@@ -32,30 +31,29 @@ const Accepted = p =>
...
+            <Kb.Text negative={true} center={true} type="Body" style={styles.friendText}>
               Share this link:{' '}
             </Kb.Text>
             <Kb.Text
-              backgroundMode="Terminal"
+              negative={true}
               center={true}
               type="BodyPrimaryLink"
-              onClick={() => openURL('https://keybase.io/airdrop')}
-              selectable={true}
```
buoyad: you can't select this without also triggering the click handler

PR #16212, opened 2019-02-22
home incoming call map + connect actions

chrisnojima:
This applies the new incoming actions to home and additionally adds a new connect action.
We build the engine, register sagas, then tell the engine we're done building sagas. If its connected it'll emit a connect event. This lets us remove another one of these custom callback schemes in engine. i back applied this to some of the actions i already cleaned up


chrisnojima:
@keybase/react-hackers

chrisnojima:
@keybase/react-hackers bump

PR #16217, opened 2019-02-23
implement grid flip progress by generating PNGs

mmaxim:
Visualize commitments and secrets using grids of colors derived from the commitments and secrets.

image

### DIFF
--- a/go/chat/flipmanager.go
+++ b/go/chat/flipmanager.go
@@ -957,6 +957,28 @@ func (v *FlipVisualizer) fillRow(img *image.NRGBA, startY, cellHeight, cellWidth
 	}
 }
 
+func (v *FlipVisualizer) getHeights(numParts int) (res []int) {
+	res = make([]int, numParts)
+	rawHeight := float64(v.height) / float64(numParts)
+	floor := int(math.Floor(rawHeight))
+	ceil := int(math.Ceil(rawHeight))
+	if ceil == 1 {
+		for i := range res {
+			res[i] = 1
+		}
+		return res
+	}
+	for i := range res {
+		res[i] = floor
+	}
+	floorTotal := numParts * floor
+	shortFall := v.height - floorTotal
+	for i := 0; i < shortFall; i++ {
+		res[i] = ceil
+	}
+	return res
+}
+
 func (v *FlipVisualizer) Visualize(status *chat1.UICoinFlipStatus) {
 	numParts := len(status.Participants)
 	if numParts == 0 {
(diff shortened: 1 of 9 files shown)


mmaxim:
cc @malgorithms @adamjspooner

go/chat/flipmanager.go
```diff
@@ -129,6 +137,7 @@ func NewFlipManager(g *globals.Context, ri func() chat1.RemoteInterface) *FlipMa
 		dirtyGames:   make(map[string]chat1.FlipGameID),
 		forceCh:      make(chan struct{}, 10),
 		loadGameCh:   make(chan loadGameJob, 100),
+		visualizer:   NewFlipVisualizer(128, 50),
```
malgorithms: this should be 256,100. Typically you want to generate 2x the pixels so it looks good in retina. the bottom of the 2 images I sent you in chat was 256,100 and the top was 128,50.

malgorithms:
@mmaxim not sure the appropriate way to make a multi-line suggestion in a draft PR like this. but I noticed that your getHeights function isn't necessary. Just a minor change to `Visualize`:

```go
func (v *FlipVisualizer) Visualize(status *chat1.UICoinFlipStatus) {
	numParts := len(status.Participants)
	if numParts == 0 {
		return
	}
	commitmentImg := image.NewNRGBA(image.Rect(0, 0, v.width, v.height))
	secretImg := image.NewNRGBA(image.Rect(0, 0, v.width, v.height))
	cellWidth := v.width / 32
	startY := 0
         // just add these next 2 things
	heightAccum := float64(0) // how far into the image we should be
	rawRowHeight := float64(v.height) / float64(numParts)
	for _, p := range status.Participants {
		heightAccum += rawRowHeight 
		rowHeight := int(math.Round(heightAccum - float64(startY)))
		if (rowHeight > 0) {
			if p.Reveal != nil {
				v.fillRow(commitmentImg, startY, rowHeight, cellWidth, p.Commitment, v.commitmentMatchColors)
				v.fillRow(secretImg, startY, rowHeight, cellWidth, *p.Reveal, v.secretColors)
			} else {
				v.fillRow(commitmentImg, startY, rowHeight, cellWidth, p.Commitment, v.commitmentColors)
			}
			startY += rowHeight
		}
	}
	var commitmentBuf, secretBuf bytes.Buffer
	png.Encode(&commitmentBuf, commitmentImg)
	png.Encode(&secretBuf, secretImg)
	status.CommitmentVisualization = base64.StdEncoding.EncodeToString(commitmentBuf.Bytes())
	status.RevealVisualization = base64.StdEncoding.EncodeToString(secretBuf.Bytes())
}
```

this will achieve the same desired effects (image size correct, handling more than row participants),  and I'm guessing it should run faster without building an array as a pre-pass. 

I tried it with some extreme cases (drawing a rectangle of only height 3 but with more than 3 participants) and it seems to work.

malgorithms:
also the `rowHeight > 0` check just prevents you from needlessly writing over the same row multiple times if you have more participants than rows, and just picks the first participant to try to fill it.

mmaxim:
Great call on the suggestion for get rid of `getHeights`. That implementation was just my literal thinking process about how the solution to the main problem should work, but yours is definitely better.

mmaxim:
Maybe we could just merge this? It seems strictly better than what is there now.

PR #16222, opened 2019-02-24
Moved team-related type definitions to the correct protocol definition

the-exodus:
keybase1/common.avdl had teams-related type definitions, including a reference to the record `TeamName` which is defined in keybase1/teams.avdl.

To get rid of the undefined reference I simply moved the stuff over to the correct definition file.


the-exodus:
There are a few more places that leak. I'll make a new PR after fixing them.

PR #16224, opened 2019-02-25
rate limit playing in flips on a per conversartion basis CORE-10268

mmaxim:
track number of participations per conversation, and will opt out of a flip if we are over the max in the time limit window. 

Needs a test, but I wanted to get this here to keep things moving (I'll get a test added before merge).


maxtaco:
Whoops, I just went ahead and merged.  How about a test in another branch

mmaxim:
No prob, it also merged into another of my branches, not master

PR #16227, opened 2019-02-25
Id3: assertion popup icons & timestamp

buoyad:
- Full color icon from service
- Show timestamp
- Add proof status icon
image

r? @keybase/react-hackers


shared/util/timestamp.js
```diff
@@ -120,7 +120,12 @@ export function formatTimeForStellarTooltip(timestamp: Date) {
 
 export function formatTimeForRevoked(time: number): string {
   const m = moment(time)
-  return m.format('ddd MMM DD') // Wed Jan 5 2016
+  return m.format('ddd MMM DD') // Wed Jan 05
+}
+
+export function formatTimeForAssertionPopup(time: number): string {
```
chrisnojima: should add to __mocks_/timestamp.js

PR #16229, opened 2019-02-25
GetClientStatus RPC

joshblum:
turns out `GetExtendedStatus` was called twice on `keybase status` which can be slow (and once on other commands https://github.com/keybase/client/blob/master/go/keybase/main.go#L356) this should speed things up


go/libkb/status.go
```diff
@@ -48,20 +48,23 @@ func getPlatformInfo() keybase1.PlatformInfo {
...
+		for i, client := range res {
+			res[i].NotificationChannels = m.G().NotifyRouter.GetChannels(ConnectionID(client.ConnectionID))
+		}
+	}
+	return res
+}
+
 func GetExtendedStatus(m MetaContext) (res keybase1.ExtendedStatus, err error) {
-	defer m.CTrace("GetExtendedStatus", func() error { return err })()
+	defer m.CTraceTimed("GetExtendedStatus", func() error { return err })()
```
maxtaco: Also:  `m = m.WithLogTag("EXTSTATUS")` ?

PR #16230, opened 2019-02-25
Fix unfurl test race

joshblum:
remove clock advance since we don't serve back cached errors and advancing the clock and race the deliverer loop force attempt with the normal send interval https://github.com/keybase/client/blob/master/go/chat/sender.go#L1378.

most of the diff is just shuffle so the test code reads easier, the essential change is
`-  		clock.Advance(2 * 10 * time.Minute) // unfurl.DefaultCacheTime`


joshblum:
gonna try this out

PR #16231, opened 2019-02-25
Warn in add to team dialog when role 'owner' is selected

buoyad:
(removed screenshots of unused)
Also fixes some styling in this footer area and of course swaps out `owner` for `admin` in the RPC call in this case. r? @keybase/react-hackers

### DIFF
--- a/shared/actions/teams.js
+++ b/shared/actions/teams.js
@@ -656,12 +656,17 @@ function* addUserToTeams(state, action) {
   const errorAddingTo = []
   for (const team of teams) {
     try {
+      let rpcRole = RPCTypes.teamsTeamRole[role]
+      if (rpcRole === RPCTypes.teamsTeamRole.owner && Constants.getRole(state, team) !== 'owner') {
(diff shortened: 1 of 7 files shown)


chrisnojima:
spoke IRL about changing the ux of this a bit

buoyad:
@chrisnojima updated to the new UX. There is one behavior that I find a little sketchy:
1. open add-to-team dialog -> select 'owner'
2. select a bunch of teams including a subteam or one you're an admin of
3. 'owner' sneakily flips to 'admin', you might not notice

I don't think it's _too_ bad because it doesn't result in a role higher than what you expected and the dropdown is right next to the submit button, increasing your chance of noticing. But it's there.

buoyad:
image
image

chrisnojima:
it does't have to flip for you, it can keep it and disable the 'ok' button

buoyad:
How would the user know they have to go into the role picker to fix it?

PR #16233, opened 2019-02-25
fix wallet pagination on desktop

buoyad:
`react-native` exports are `any` because of this: https://github.com/keybase/client/blob/master/shared/libs/flow-interface.js.flow#L21
I started on redefining type `Props` here but the full type is huge (inherits from `VirtualizedSectionList` which inherits from `ScrollView`). Left a TODO to import the type from react-native. r? @keybase/react-hackers


chrisnojima:
there's a ticket to use the rn types and remove the any actually

PR #16236, opened 2019-02-25
wrap saga in flag. swallow errors

chrisnojima:
@keybase/react-hackers


chrisnojima:
@keybase/react-hackers bump

PR #16237, opened 2019-02-25
more incoming cleanup: notifications + settings

chrisnojima:
@keybase/react-hackers


chrisnojima:
@keybase/react-hackers bump

zanderz:
Sanity checked on windows and mac. Mac one never got past loading screen if I start it without prod keybase already running, but windows one seemed to ok.

chrisnojima:
its not really safe to run mulitple copies at the same time. ui calls get all messed up

zanderz:
whoops, my mistake, that case doesn't work on master either for Mac. Does on windows though.
But I've always run them side by side, usually it's fine? how do you run it otherwise?

chrisnojima:
some stuff works side by side but some calls (like delegation of ui stuff) the daemon only allows one thing to be registered

PR #16243, opened 2019-02-25
Section divider common adapter

buoyad:
Use on wallets and devices lists. Also remove a `marginLeft` on the `NEW` tag for devices.
r? @keybase/react-hackers


shared/common-adapters/section-divider.js
```diff
@@ -0,0 +1,52 @@
...
+import ClickableBox from './clickable-box'
+
+type Props = {|
+  collapsed?: boolean, // if set, render the appropriate caret
+  label: string,
+  onToggleCollapsed?: () => void,
+|}
+
+const SectionDivider = (props: Props) => {
+  let collapsible = 'collapsed' in props
```
chrisnojima: const
also `hasOwnProperty` likely better than `in`

buoyad:
@chrisnojima thanks, done PTAL

PR #16244, opened 2019-02-25
Deprecate a bunch of engine things

chrisnojima:
This PR does some long standing cleanup to the engine layer. 
The PRs before this switch from the registerIncomingCall flow to just emitting actions so we can handle things w/ our sagas in a similar way and reduce a bunch of boiler plate
This PR handles the rarer case where we want to handle an incoming call with an action, yet we still want to handle the response ourselves.
The simplest way to do this was to register the method which tells the engine just to not auto respond
This does touch some super old rarely used flow so lets merge after the release

- [ ]  after release


protocol/bin/flow.js
```diff
@@ -394,8 +394,14 @@ function writeActions() {
               .split('.')
               .map((p, idx) => (idx ? capitalize(p) : p))
               .join('')
+
+            let response = ''
```
chrisnojima: plumb through response object

shared/actions/config/index.js
```diff
@@ -118,15 +118,6 @@ function* loadDaemonBootstrapStatus(state, action) {
   }
 }
 
-let dispatchSetupEngineListenersOnce = false
```
chrisnojima: we don't dispatch this anymore

shared/actions/fs/platform-specific.desktop.js
```diff
@@ -356,8 +356,8 @@ function* platformSpecificSaga(): Saga.SagaGenerator<any, any> {
     FsGen.openPathInSystemFileManager,
     openPathInSystemFileManager
   )
-  yield* Saga.chainGenerator<ConfigGen.SetupEngineListenersPayload | FsGen.FuseStatusPayload>(
-    [ConfigGen.setupEngineListeners, FsGen.fuseStatus],
+  yield* Saga.chainGenerator<ConfigGen.InstallerRanPayload | FsGen.FuseStatusPayload>(
```
chrisnojima: this is a close proxy for the behavior before

shared/actions/pinentry.js
```diff
@@ -1,21 +1,16 @@
...
 import * as PinentryGen from '../actions/pinentry-gen'
 import * as Saga from '../util/saga'
 import * as I from 'immutable'
 import * as RPCTypes from '../constants/types/rpc-gen'
-import engine from '../engine'
+import {getEngine} from '../engine/require'
 
-// We keep track of sessionID to response objects since this is initiated by the daemon
-type IncomingErrorCallback = (?{code?: number, desc?: string}) => void
-const sessionIDToResponse: {
```
chrisnojima: spoke to @maxtaco about this. this likely never worked. so lets simplify and just keep one response. in the future if we want this to be smarter its likely easier for the GUI to make a call to the daemon registering itself with an engineSaga and just getting pinenetry requests with a 'real id' and not this session id

shared/actions/pinentry.js
```diff
@@ -60,50 +49,38 @@ const onNewPinentry = (_, action) =>
   })
 
 const onSubmit = (_, action) => {
-  const {sessionID, passphrase} = action.payload
-  _respond(sessionID, {passphrase})
+  const {passphrase} = action.payload
+  if (_response) {
+    // $FlowIssue flow is correct in that we need store secret but we weren't sending it before and i don't want to change any existing behavior
+    _response.result({passphrase})
```
chrisnojima: flow will infer _response correctly now and it correctly points out we're missing a param. i didn't want to break anything so i left it as-is

shared/actions/platform-specific/index.desktop.js
```diff
@@ -332,18 +329,6 @@ export function* platformConfigSaga(): Saga.SagaGenerator<any, any> {
   )
   yield* Saga.chainAction<ConfigGen.ShowMainPayload>(ConfigGen.showMain, showMainWindow)
   yield* Saga.chainAction<ConfigGen.DumpLogsPayload>(ConfigGen.dumpLogs, dumpLogs)
-  yield* Saga.chainGenerator<ConfigGen.SetupEngineListenersPayload>(
```
chrisnojima: there's no value in waiting for this action so just spawn it

shared/actions/platform-specific/index.desktop.js
```diff
@@ -361,6 +346,11 @@ export function* platformConfigSaga(): Saga.SagaGenerator<any, any> {
     EngineGen.keybase1NotifySessionClientOutOfDate,
     onOutOfDate
   )
+  getEngine().registerCustomResponse('keybase.1.logsend.prepareLogsend')
```
chrisnojima: this is how we tell the engine that we intend on responding to these actions ourselves

shared/actions/tracker.js
```diff
@@ -286,7 +285,8 @@ function _serverCallMap(
 
         // cleanup bookkeeping
         delete sessionIDToUsername[sessionID]
-        engine().cancelSession(sessionID)
+        // $FlowIssue exist but this is the only thing that uses it and this is going away
```
chrisnojima: this entire file is going away after the next release

chrisnojima:
@keybase/react-hackers @jzila @songgao

shared/actions/fs/platform-specific.desktop.js
songgao: Does this always happen regardless of user settings (fuse driver, autostart, etc.)?

chrisnojima: this matches the logic before. before config saga would listen for InstallerRan and dispatch setupEngineListeners

songgao: Got it; sounds good!

chrisnojima:
@keybase/react-hackers bump

chrisnojima:
@keybase/react-hackers bump

PR #16246, opened 2019-02-25
cr: disk-based local dir block map

strib:
This turns the existing data structure `localBcache` into an interface called `dirBlockMap`.  This structure holds all the directory blocks while prepping updates, either for normal sync operations, or for conflict resolution.  In the latter case, the number of directories being updated could in theory be very large, so we want to store those in a disk-based cache instead of in memory.

To get this to work, this PR also fixes all the place in the code that assumed a directory block could be modified without re-putting it into the cache, or subsequently re-getting it from the cache.  This is necessary since, with the disk-based cache, the `*DirBlock` is re-assembled from disk each time it's accessed.

Issue: KBFS-3818

### DIFF
--- a/go/kbfs/libkbfs/dir_block_map_disk.go
+++ b/go/kbfs/libkbfs/dir_block_map_disk.go
@@ -1,0 +1,76 @@
+// Copyright 2019 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libkbfs
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+)
+
+// dirBlockMapDisk tracks dir block info while making a revision, by
+// using a disk-based block cache.
+type dirBlockMapDisk struct {
+	dirtyBcache *DirtyBlockCacheDisk
+	kmd         KeyMetadata
+	ptrs        map[BlockPointer]bool
+}
+
+var _ dirBlockMap = (*dirBlockMapDisk)(nil)
+
+func newDirBlockMapDisk(
+	dirtyBcache *DirtyBlockCacheDisk, kmd KeyMetadata) *dirBlockMapDisk {
+	return &dirBlockMapDisk{
+		dirtyBcache: dirtyBcache,
+		kmd:         kmd,
+		ptrs:        make(map[BlockPointer]bool),
+	}
+}
+
+func (dbmd *dirBlockMapDisk) putBlock(
+	ctx context.Context, ptr BlockPointer, block *DirBlock) error {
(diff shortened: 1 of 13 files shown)


strib:
(Might be easiest to review commit-by-commit.)

go/kbfs/libkbfs/folder_update_prepper.go
```diff
@@ -376,6 +376,13 @@ func (fup *folderUpdatePrepper) prepUpdateForPath(
 			for _, unref := range unrefs {
 				md.AddUnrefBlock(unref)
 			}
+			// Fetch the current block again, since `setEntry` might
+			// not modify the original `currBlock`, but some
+			// re-assembled version if the disk cache is in use.
+			currBlock, err = dbm.getBlock(ctx, currDD.rootBlockPointer())
```
jakob223: why does getting it again prevent us from modifying a re-assembled version? Because if it's been queried recently it will live in the memory cache instead?

go/kbfs/libkbfs/dir_block_map_disk.go
```diff
@@ -0,0 +1,76 @@
...
+	return &dirBlockMapDisk{
+		dirtyBcache: dirtyBcache,
+		kmd:         kmd,
+		ptrs:        make(map[BlockPointer]bool),
+	}
+}
+
+func (dbmd *dirBlockMapDisk) putBlock(
+	ctx context.Context, ptr BlockPointer, block *DirBlock) error {
+	err := dbmd.dirtyBcache.Put(
```
jakob223: Why does the `dirBlockMapDisk` use the `dirtyBcache` instead of having a second `levelDb`?

go/kbfs/libkbfs/folder_update_prepper.go
strib: The comment is about the earlier call to `setEntry()`.  That code adds a new entry to the directory, but it doesn't use `currBlock`.  Instead, it `Get`s the block from the `dirBlockMap`, which if it's disk-based, ends up constructing a new instance of the block, and that instance is the one that gets modified and written back to disk.  So this `getBlock` call here is to read the newly-modified block from disk and replace `currBlock` with it.

go/kbfs/libkbfs/dir_block_map_disk.go
strib: Entirely to save on code duplication.  `DirtyBlockCacheDisk` already has code to turn in-memory blocks into disk blocks (by "readying" them on puts, and re-assembling them from the encrypted components when reading them back).  Seems like there would be no harm in putting everything together in a single leveldb, so I just went with that approach.

PR #16247, opened 2019-02-25
file_data: remove "file too big for CR" error

strib:
Now that we store blocks on disk during CR, there's no need for a hard limit on file size.  If the disk fills up during CR, the code will `RemoveAll` the CR directory and treat it like a normal error, rather than just panic'ing.

### DIFF
--- a/go/kbfs/libkbfs/errors.go
+++ b/go/kbfs/libkbfs/errors.go
@@ -1041,17 +1041,6 @@ func (e cachePutCacheFullError) Error() string {
 		e.blockID)
 }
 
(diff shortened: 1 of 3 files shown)


jakob223:
I assume the code that actually does the removing was part of the PR that made that a disk-based store?

strib:
> I assume the code that actually does the removing was part of the PR that made that a disk-based store?

Yeah this is made possible by the previous PRs that started storing the `fileBlockMap` and the `blockPutState` on disk, instead of in memory.

PR #16249, opened 2019-02-26
in-progress for enabling/disabling banner and cleanup

songgao:
Also restructure driver related state a little bit so what's in redux store is now more "processed".

The kext popup thing only works on a fresh macOS install. Here's a screencast in vm: /keybase/team/keybase/videos/macOS-finder-integration.mov


songgao:
bump!

shared/fs/common/try-enable-driver-on-focus.js
```diff
@@ -0,0 +1,61 @@
...
+})
+
+type Props = {
+  appFocusedCount: number,
+  driverStatus: Types.DriverStatus,
+  onEnabled: () => void,
+  enable: () => void,
+}
+
+class Component extends React.PureComponent<Props> {
```
chrisnojima: lets talk about this, i think this is likely a dangerous pattern

songgao:
ready again!

shared/fs/banner/system-file-manager-integration-banner/kext-permission-popup.js
```diff
@@ -36,9 +36,21 @@ const Illustration = () => (
   </Kb.Box>
 )
 
+class CancelWhenEnabled extends React.PureComponent<Props> {
+  componentDidMount() {
```
chrisnojima: should make a helper like _cancelOnUpdate or something

PR #16252, opened 2019-02-26
Refactor Context to globals.ChatCtx and StorageLockTab to utils pkg [CORE-10300]

joshblum:
did some basic profiling of the app running and it seemed like there was a lot of lock contention for reading storage. pr breaks `ConversationLockTab` into utils to have a per conversation lock when reading messages out of storage. involved some unfortunate reshuffling of contents from the main `chat` package to avoid circular dependencies, @mmaxim lmk if you have better ideas for avoiding this.


mmaxim:
I wonder if it makes sense to use `ConversationLockTab` here, maybe easier to just have a per conv ID lock within storage? How come you didn't do that?

joshblum:
doesn't the `ConversationLockTab` essentially do that already it just also allows sharing per trace?

mmaxim:
Yeah but that locktab is potentially blocking on all kinds of things, I worry this change actually increases contention. Could I convince you to try using `libkb.LockTable` instead?

joshblum:
`libkb.LockTable` is probably fine. i make a new instance of the locktab so there would be no contention from convsource if that's what you meant. what do you think about keeping some of the reorg of Context so the locktab is accessible in the future if we want since we both hit this problem before and punted on refactoring

mmaxim:
Maybe we could make that refactor a separate PR since it will end up being unrelated?

joshblum:
yeah i can repurpose this to be the refactor and then do a simple one for the `libkb.LockTable` if you're on board

joshblum:
@mmaxim this is ready for another pass

joshblum:
patch now does the following:

- refactors `ConversationLockTab` to `utils`
- moves `KeyFinder` and `UPAKFinder` to the global context instead of request context
- refactors `chat/context.go` into `globals/request_context.go`

@mmaxim deviated a bit from our discussion, let me know what you think

mmaxim:
Why is `KeyFinder` on `G` here?

joshblum:
@mmaxim cleaned up and ready for a pass

PR #16253, opened 2019-02-26
folder_block_manager: wrap defer in func to set next QR period

strib:
`nextPeriod` can be changed in the code below, so it shouldn't be interpreted until the `defer` actually executes.


go/kbfs/libkbfs/folder_block_manager.go
```diff
@@ -1022,7 +1022,9 @@ func (fbm *folderBlockManager) doReclamation(timer *time.Timer) (err error) {
 	fbm.setReclamationCancel(cancel)
 	defer fbm.cancelReclamation()
 	nextPeriod := fbm.config.Mode().QuotaReclamationPeriod()
-	defer timer.Reset(nextPeriod)
+	defer func() {
+		timer.Reset(nextPeriod)
```
jzila: Add a comment here please since the `nextPeriod` is updated inside another `defer`, which gets unrolled before this one. It's all correct, but a comment indicating that awareness would be helpful.

PR #16254, opened 2019-02-26
Bring flip help text up faster

buoyad:
Adds a bool `hasHelpText` to `ConversationCommand` and sets it to true for `flip`. On the js side, skip debouncing our call into the service if the text exactly matches a command with `hasHelpText: true`.

flip-help-text

cc @malgorithms this fixes hitting enter right after entering /flip. r? @keybase/react-hackers


mmaxim:
Just make sure to try with /giphy to make sure we aren' searching too much.

PR #16255, opened 2019-02-26
nav2 signup flow

chrisnojima:
this adds some niceties to the signup flow for nav v2
- [x] Progress indicator on invitation page so if it flashes you just see a spinner
- [x] nav up and append moving past the invite page so you don't click back and see it
- [x] fix layout of main app in login route, don't let the header touch the upper left (where on osx you have the system close buttons)
- [x] tested mobile
- [x] tested nav v2
- [x] tested nav v1
- [x] tested auto invite not being supplied


chrisnojima:
@keybase/react-hackers

chrisnojima:
@keybase/react-hackers bump

PR #16256, opened 2019-02-26
Fix a typo in the device name checker regex

pzduniak:
For some reason I totally ignored the fact that `['+_-]` only didn't resolve to a range because there was nothing after the dash and simply added new chars after it, resulting in a massive ASCII-unicode range that somehow didn't fail any tests. I fixed the typo and added tests checking `@` and `(`, which should be triggered by such mistakes.

### DIFF
--- a/go/libkb/checkers.go
+++ b/go/libkb/checkers.go
@@ -14,7 +14,7 @@ import (
 
 var emailRE = regexp.MustCompile(`^\S+@\S+\.\S+$`)
 
-var deviceRE = regexp.MustCompile(`^[a-zA-Z0-9][ _'a-zA-Z0-9+-—–‘’]*$`)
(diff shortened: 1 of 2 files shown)


go/libkb/checkers.go
```diff
@@ -14,7 +14,7 @@ import (
 
 var emailRE = regexp.MustCompile(`^\S+@\S+\.\S+$`)
 
-var deviceRE = regexp.MustCompile(`^[a-zA-Z0-9][ _'a-zA-Z0-9+-—–‘’]*$`)
+var deviceRE = regexp.MustCompile(`^[a-zA-Z0-9][ _'a-zA-Z0-9+\-—–‘’]*$`)
```
maxtaco: usually i workaround this issue by just putting `-` last.  That might actually be easier to read too, given the`-` and the em-dash smush together .....

pzduniak: On the other hand the escaping helps differentiate it from the en dash, although I agree that we should probably reorder them for consistency.

PR #16259, opened 2019-02-26
Fix emoji being cut off by making them smaller.

adamjspooner:

### DIFF
--- a/shared/chat/conversation/input-area/normal/platform-input.native.js
+++ b/shared/chat/conversation/input-area/normal/platform-input.native.js
@@ -16,7 +16,6 @@ import FilePickerPopup from '../filepicker-popup'
 import WalletsIcon from './wallets-icon/container'
 import type {PlatformInputPropsInternal} from './platform-input'
 import AddSuggestors, {standardTransformer} from '../suggestors'
-import {metaData} from '../../../../common-adapters/text.meta.native'
 
 type menuType = 'exploding' | 'filepickerpopup'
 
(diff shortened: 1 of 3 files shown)


shared/chat/conversation/input-area/normal/platform-input.native.js
```diff
@@ -324,7 +324,8 @@ const styles = Styles.styleSheetCreate({
   },
   input: {
     flex: 1,
-    fontSize: metaData['BodyBig'].fontSize,
+    // Override BodyBig's default weight.
+    fontWeight: Styles.globalStyles.fontRegular.fontWeight,
```
adamjspooner: We need the line height that `PlainInput` takes from `textType` (changed above). So, we'll override the weight instead of the height.

shared/common-adapters/plain-input.native.js
```diff
@@ -175,7 +175,7 @@ class PlainInput extends Component<InternalProps, State> {
 
   _getCommonStyle = () => {
     const textStyle = getTextStyle(this.props.textType)
-    return collapseStyles([{lineHeight: this._lineHeight()}, styles.common, textStyle])
+    return collapseStyles([styles.common, textStyle])
```
adamjspooner: We already get this from `textStyle`.

mmaxim:
This might be the "cause" of the shifting messages problem in chat threads.

adamjspooner:
I thought that predated this merge.

PR #16264, opened 2019-02-26
kbfsedits: track deleted file history and add option to show from command line

strib:
This PR is a first-pass at including deleted files in the the edit history for each TLF.  The basic idea is:

* Search backward in the chat history for at most 10 deleted files per writer per TLF.
* If a file has been deleted, but subsequently overwritten, don't include in the deleted files history.
* Don't order history clusters by the delete times, unless there are no other edits in the cluster.
* Add a `-d` option to `keybase fs history`, which will show the deleted history of each cluster (but ordered by the usual overall modification time of the cluster).
* Deletes will always show up in a separate "deletes" section in `.kbfs_edit_history` for each cluster.

This isn't perfect; consider this a first dev-designed pass.  In particular, there are some weird/bad things about this:
* We'll search backward in each writer's history until we see 10 deletes.  So if the user did a ton of writes after their most recent deletes (or never did any deletes at all), we'll have to load the whole history.  In practice we can probably limit this by figuring out the last GC revision and stop checking for deleted files earlier than that, since we won't be able to restore them anyway.  But not sure if that's worth doing for now, it seems pretty fast as it is.
* There are probably some deletes we should filter out, like `*~` or `*.swp` files, but I haven't attempted to enumerate that list yet.
* It's weird that `keybase fs history -d` sorts clusters by the most recent edit notification time, rather than the most recent delete time, but I'm not sure if this matters enough to care about.  The GUI could always decide to do it differently.

Open to suggestions!

Issue: KBFS-3901

### DIFF
--- a/go/kbfs/kbfsedits/tlf_history.go
+++ b/go/kbfs/kbfsedits/tlf_history.go
@@ -18,13 +18,16 @@ import (
 
 const (
 	// The max number of edits needed for each writer.
-	maxEditsPerWriter    = 10
+	maxEditsPerWriter = 10
+	// The max number of deletes needed for each writer.
+	maxDeletesPerWriter  = 10
 	maxWritersPerHistory = 10
 )
 
 type writerNotifications struct {
 	writerName    string
 	notifications notificationsByRevision
+	deletes       notificationsByRevision
 }
 
 // writersByRevision sorts sets of per-writer notifications in reverse
@@ -40,7 +43,21 @@ func (wbr writersByRevision) Less(i, j int) bool {
 	iHasZero := len(wbr[i].notifications) == 0
 	jHasZero := len(wbr[j].notifications) == 0
 	if jHasZero {
-		return iHasZero
+		if iHasZero {
+			// If neither has any notifications, sort by the latest
+			// delete.
+			iHasZeroDeletes := len(wbr[i].deletes) == 0
+			jHasZeroDeletes := len(wbr[j].deletes) == 0
+			if jHasZeroDeletes {
+				return iHasZeroDeletes
+			} else if iHasZeroDeletes {
+				return false
+			}
+
+			// Reverse sort, so latest deletes come first.
+			return wbr[i].deletes[0].Revision > wbr[j].deletes[0].Revision
+		}
+		return false
 	} else if iHasZero {
 		return false
 	}
@@ -135,7 +152,7 @@ func (th *TlfHistory) AddNotifications(
 	defer th.lock.Unlock()
 	wn, existed := th.byWriter[writerName]
 	if !existed {
-		wn = &writerNotifications{writerName, nil}
+		wn = &writerNotifications{writerName, nil, nil}
 	}
 	oldLen := len(wn.notifications)
 	newEdits = append(newEdits, wn.notifications...)
@@ -167,7 +184,7 @@ func (th *TlfHistory) AddUnflushedNotifications(
 	th.lock.Lock()
 	defer th.lock.Unlock()
 	if th.unflushed == nil {
-		th.unflushed = &writerNotifications{loggedInUser, nil}
+		th.unflushed = &writerNotifications{loggedInUser, nil, nil}
 	}
 	if th.unflushed.writerName != loggedInUser {
 		panic(fmt.Sprintf("Logged-in user %s doesn't match unflushed user %s",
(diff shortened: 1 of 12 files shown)


go/kbfs/kbfsedits/tlf_history.go
```diff
@@ -386,6 +406,56 @@ func (r *recomputer) processNotification(
...
+		}
+
+		if len(wn.deletes) == maxDeletesPerWriter {
+			// We don't need any more deletes, but we
+			// should continue looking for more edit notifications.
+			return false
+		}
+
+		if hasEvent && event.delete {
+			// It's already been deleted, no need to track it further.
```
jakob223: why would a file have multiple delete events?

go/kbfs/kbfsedits/tlf_history.go
```diff
@@ -386,6 +406,56 @@ func (r *recomputer) processNotification(
...
+
+		wn.deletes = append(wn.deletes, notification)
+
+		if len(wn.notifications) == maxEditsPerWriter &&
+			len(wn.deletes) == maxDeletesPerWriter {
+			// We have enough edits and deletes for this user.
+			return true
+		}
+
+		// TODO: limit the number (or time span) of notifications we
```
jakob223: Do we care enough about this to store the most recent n deletes somewhere instead of computing them from the LARGE_NUM most recent operations each time?

How recent is the most recent GC'd revision typically?

go/kbfs/kbfsedits/tlf_history_test.go
```diff
@@ -427,9 +429,131 @@ func TestTlfHistoryRenameParentSimple(t *testing.T) {
...
+	expected[1].notifications = allAliceExpected[:maxEditsPerWriter]
+	rev, err = th.AddNotifications(aliceName, aliceMessages)
+	require.NoError(t, err)
+	require.Equal(t, allAliceExpected[0].Revision, rev)
+	rev, err = th.AddNotifications(bobName, bobMessages)
+	require.NoError(t, err)
+	require.Equal(t, allBobExpected[0].Revision, rev)
+	checkTlfHistory(t, th, expected, aliceName)
+
+	// Re-creating a deleted file should remove the delete.
```
jakob223: If I were looking in my edit history and I had changed something, reverted it, and then changed it again, I would expect there to be 3 entries. Why do we want to act differently here?

go/kbfs/kbfsedits/user_history.go
```diff
@@ -161,10 +185,14 @@ func (uh *UserHistory) Get(loggedInUser string) (
...
-			if len(wh.Edits) > 0 {
+			if len(wh.Edits) > 0 || len(wh.Deletes) > 0 {
+				var serverTime keybase1.Time
+				if len(wh.Edits) > 0 {
+					serverTime = wh.Edits[0].ServerTime
+				}
 				clusters = append(clusters, keybase1.FSFolderEditHistory{
 					Folder:     history.Folder,
-					ServerTime: wh.Edits[0].ServerTime,
+					ServerTime: serverTime,
```
jakob223: displaying time 0 if there are only deletes seems a little icky... not that I have a better suggestion of what time to display, but it. might be confusing.

jakob223: Do we show it in the UI as something other than Jan 1970 or Jan 1 year 0000?

go/kbfs/kbfsedits/tlf_history.go
strib: It could be created and deleted multiple times.  The edit history tracks files simply by their names, not by block ID or anything like the CR does.

go/kbfs/kbfsedits/tlf_history.go
strib: The computed history is already cached until new data comes in.  We could probably come up with some complicated story to update the deletes separately from the rest of the edit history only when a new delete comes in, but I think this is pretty fast.  Good suggestion, but I probably don't care enough about it to do anything, unless I see some evidence that this is super slow in some cases.

go/kbfs/kbfsedits/tlf_history_test.go
strib: I was extending the design from the regular edit history that we should only show the most recent event for each file.  I'm not 100% sure it's the right thing for deletes until we have a proper design for it, but it seemed intuitive to me that we'd mainly use this list as a way to restore completely deleted files -- if a file currently exists, the recovery design would probably list the available versions of that file along with the file itself, rather than on some "deleted files" screen, and listing pre-deleted versions of that file would require more extensive changes here, I think.

Either way, I think our different instincts here shows that we should probably do the simplest thing until design weighs in.

go/kbfs/kbfsedits/user_history.go
strib: Hmm, I don't know if this time is shown anywhere in the GUI, if there are no edits in the folder.  If you only have deletes, I don't think the GUI will currently show anything.

Note that this is just the overall time of this cluster of edits -- it's mainly used for sorting.  Each edit/delete still has its own timestamp that can be displayed by the UI.

PR #16268, opened 2019-02-27
always include dokan uninstall string in fuse status on windows

zanderz:


go/install/fuse_status_windows.go
```diff
@@ -38,31 +37,31 @@ func detectDokanDll(dokanPath string, log Log) bool {
...
 	//   https://msdn.microsoft.com/en-us/library/windows/desktop/aa384129(v=vs.110).aspx
 	if wow64 {
-		access = access | registry.WOW64_64KEY
+		access = access | registry.WOW64_32KEY
 	}
 
 	k, err := registry.OpenKey(registry.LOCAL_MACHINE, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall", access)
 	if err != nil {
-		fmt.Printf("Error %s opening uninstall subkeys\n", err.Error())
+		log.Info("Error %s opening uninstall subkeys\n", err.Error())
```
songgao: Should these be `log.Error` or `log.Debug` instead? I think Info and above get printed into CLI when you do Keybase commands.

go/install/fuse_status_windows.go
```diff
@@ -38,31 +37,31 @@ func detectDokanDll(dokanPath string, log Log) bool {
...
-func findDokanUninstall(wow64 bool) (result string) {
+func findDokanUninstall(log Log, wow64 bool) (result string) {
 	dokanRegexp := regexp.MustCompile("^Dokan Library.*Bundle")
 	var access uint32 = registry.ENUMERATE_SUB_KEYS | registry.QUERY_VALUE
-	// Assume this is build 32 bit, so we need this flag to see 64 bit registry
+	// Assume this is build 64 bit, so we need this flag to see 32 bit WOW registry
 	//   https://msdn.microsoft.com/en-us/library/windows/desktop/aa384129(v=vs.110).aspx
 	if wow64 {
-		access = access | registry.WOW64_64KEY
+		access = access | registry.WOW64_32KEY
```
songgao: Not that it'd matter for result since we call with both wow64=true and wow64=false, but it looks like the `if` condition here should be `if !wow64` now?

PR #16270, opened 2019-02-27
[KBFS-3531] Android Sharing

jzila:


jzila:
@songgao ready for a re-review based on your feedback in chat.

shared/actions/fs/index.js
```diff
@@ -710,15 +710,31 @@ const moveOrCopy = (state, action) => {
       kbfs: Constants.fsPathToRpcPathString(
         Types.pathConcat(
           action.payload.destinationParentPath,
-          Types.getPathName(state.fs.moveOrCopy.sourceItemPath)
+          state.fs.destinationPicker.source.type === 'move-or-copy'
+            ? Types.getPathName(state.fs.destinationPicker.source.path)
+            : state.fs.destinationPicker.source.type === 'incoming-share'
+              ? Types.getLocalPathName(state.fs.destinationPicker.source.localPath)
+              : ''
```
songgao: Perhaps there can be a guard at beginning

```
if (state.fs.destinationPicker.source.type === 'none') {
return
}
```

so we don't need to use an empty string here.

shared/fs/destination-picker/index.js
```diff
@@ -116,9 +116,14 @@ export default (isMobile
...
-                {props.targetName}
-              </Kb.Text>
+              <Kb.Box2 direction="horizontal">
+                <Kb.Text type="BodySmallSemibold" lineClamp={1}>
+                  {props.targetNameWithoutExtension}
+                </Kb.Text>
+                <Kb.Text type="BodySmallSemibold">
+                  {props.targetExtension}
+                </Kb.Text>
+              </Kb.Box2>
```
songgao: Is the extra box necessary?

shared/actions/platform-specific/push.native.js
```diff
@@ -70,6 +74,18 @@ const listenForNativeAndroidIntentNotifications = emitter => {
...
+  // TODO: move this out of this file.
+  RNEmitter.addListener('onShareData', (evt) => {
+    logger.info('[ShareDataIntent]', evt)
+    emitter(RouteTreeGen.createNavigateTo({path: FsShared.fsRootRoute}))
+    emitter(FsGen.createSetIncomingShareLocalPath({localPath: FsTypes.stringToLocalPath(evt.path)}))
+    emitter(FsGen.createShowIncomingShare({initialDestinationParentPath: FsTypes.stringToPath('/keybase')}))
+  })
+  RNEmitter.addListener('onShareText', (evt) => {
+    logger.info('[ShareTextIntent]', evt)
+    emitter(RouteTreeGen.createNavigateTo({path: FsShared.fsRootRoute}))
```
songgao: If we're not handling text, can we disable it on java side?

shared/fs/destination-picker/container.js
```diff
@@ -57,16 +67,21 @@ const canWrite = memoize(
 
 const canCopy = memoize(
   (stateProps, ownProps: OwnProps) =>
-    canWrite(stateProps, ownProps) &&
-    getDestinationParentPath(stateProps, ownProps) !==
-      Types.getPathParent(stateProps._moveOrCopy.sourceItemPath)
+    canWrite(stateProps, ownProps) && (
+      stateProps._destinationPicker.source.type === 'incoming-share' || (
+        stateProps._destinationPicker.source.type === 'move-or-copy' &&
+        // $FlowIssue can't figure out that we're in a MoveOrCopySource here, for whatever reason.
```
songgao: I think this is because `getDestinationParentPath` returns either a `Type.Path` or a `string`. There's no way flow can figure out which one it is.

Why not just do `Types.getPathParent` and `Types.getLocalPathParent` respectively instead of `getDestinationParentPath`?

shared/react-native/android/app/src/main/java/io/keybase/ossifrage/MainActivity.java
```diff
@@ -168,4 +275,5 @@ protected void onDestroy() {
     protected String getMainComponentName() {
         return "Keybase";
     }
+
```
songgao: Intentional white space?

shared/react-native/android/app/src/main/java/io/keybase/ossifrage/MainActivity.java
```diff
@@ -59,6 +74,102 @@ private void createDummyFile() {
...
+                byte[] buf = new byte[64 * 1024];
+                int len;
+                while ((len = istream.read(buf)) != -1) {
+                    ostream.write(buf, 0, len);
+                }
+                filePath = file.getPath();
+            } catch (IOException ex) {
+                Log.w(TAG, "error writing shared file " + uri.toString());
+            }
```
songgao: What do you think if we use this routine to handle onShareText so we always get a file that's actionable for us?

shared/constants/types/fs.js
```diff
@@ -291,7 +292,13 @@ export type _MoveOrCopy = {
...
   destinationParentPath: I.List<Path>,
 }
-export type MoveOrCopy = I.RecordOf<_MoveOrCopy>
+
+export type _IncomingShare = {
+  type: 'incoming-share',
+  sourceItemLocalPath: LocalPath,
+  destinationParentPath: I.List<Path>,
+}
+export type DestinationPicker = I.RecordOf<_IncomingShare> | I.RecordOf<_MoveOrCopy>
```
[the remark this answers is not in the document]
songgao: oops this is old comment I forgot to delete; I'm gonna hit resolve.

shared/constants/types/fs.js
```diff
@@ -450,7 +471,11 @@ export const stringToPathType = (s: string): PathType => {
 }
 export const pathTypeToString = (p: PathType): string => p
 export const pathConcat = (p: Path, s: string): Path =>
-  p === '/' ? stringToPath('/' + s) : stringToPath(pathToString(p) + '/' + s)
+  s === ''
```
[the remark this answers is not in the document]
jzila: I'd prefer to have it here, since otherwise we're asking callers to add even more boilerplate in many places.

shared/actions/fs/index.js
```diff
@@ -734,7 +750,20 @@ const moveOrCopy = (state, action) => {
...
+  RouteTreeGen.createPutActionIfOnPath({
+    expectedPath: action.payload.routePath,
+    otherAction: RouteTreeGen.createNavigateAppend({
+      path: [{props: {index: action.payload.currentIndex + 1}, selected: 'destinationPicker'}],
+    }),
+  }),
+]
+
+const incomingShareOpen = (state, action) => [
+  FsGen.createSetDestinationPickerParentPath({
```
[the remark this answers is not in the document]
jzila: Sure, I'll try that out.

shared/actions/fs/index.js
jzila: I prefer this, since we need to return something.

shared/actions/platform-specific/push.native.js
jzila: We will handle text, this is mostly a placeholder. I'll add a comment to that effect.

shared/react-native/android/app/src/main/java/io/keybase/ossifrage/MainActivity.java
jzila: I was thinking of doing that, but it won't apply when we want to share to chat. Still figuring out the best approach for that.

shared/actions/fs/index.js
songgao: It's a Saga handler. We can return `null` meaning we'd do nothing for that action, which makes more sense than calling a RPC as we don't have a source path.

shared/fs/destination-picker/container.js
jzila: It actually always returns a path.

jzila: We don't ever want the local path here. This is only for the destination, which we only supply as the source by default in a `move-or-copy` dialog. In any other situation we just want it to be `/keybase`.

songgao: You're right; sorry I mis-read.

shared/fs/destination-picker/index.js
jzila: Maybe, maybe not. I believe I tried it without and it didn't work visually. But my build is currently not working due to "new.target can't come after a prefix operator."...

jzila: OK yeah it's necessary, the cells don't collapse together without the `Box2`

jzila:
@songgao ready for a final look!

shared/fs/destination-picker/index.js
songgao: Oh I see; I guess it's because of the gap attr in outter box. Sounds good!

PR #16271, opened 2019-02-27
thread MetaContext through the IdentifyUI

maxtaco:
- this should fix the problem of the context getting canceled in upcall-initiated identify3 from KBFS
- also, improvements with debug-tag logging throughout

### DIFF
--- a/go/chat/identify_ui.go
+++ b/go/chat/identify_ui.go
@@ -1,30 +1,35 @@
 package chat
 
-import "github.com/keybase/client/go/protocol/keybase1"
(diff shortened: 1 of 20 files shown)


maxtaco:
cc @chrisnojima this is the fix the issue @malgorithms saw

maxtaco:
We are, but if we have to conform to a rpc-generated server interface, we can't pass through `MetaContext`, we need to pass through `Context` and that necessitates getting the `G` from somewhere else.

PR #16274, opened 2019-02-27
giphy full grid preview display

mmaxim:
Looks like this:

image


shared/chat/conversation/giphy/width.js
```diff
@@ -1,10 +1,221 @@
...
+  }
+  // Pick which row is the closest, biasing for compression if both succeed.
+  const compWidth = groupWidth(compressed)
+  const expandWidth = groupWidth(expanded)
+  const compDistance = totalWidth - compWidth
+  let expandDistance = totalWidth - expandWidth
+  if (expandDistance < 0) {
+    expandDistance = 0
+  }
+  if (compDistance < expandDistance) {
```
joshblum: why not `<=` and then we don't need the `else if` or the `else`?

shared/chat/conversation/giphy/width.js
```diff
@@ -1,10 +1,221 @@
...
+  widths.forEach(w => {
+    images.push(new Image(scaledWidth(w), 0))
+  })
+  // set to the desired size, some GIFs are very wide and take up too much space initially. We will allow
+  // the expansion phase to go over this clamp however.
+  clampAtMaxWidths(images)
+  let res = []
+  let longRow = []
+  for (let index = 0; index < images.length; index++) {
+    const im = images[index]
```
joshblum: since `index` is reassigned from `pickRow` probably worth a sanity check the index is within bounds here

shared/chat/conversation/giphy/width.js
mmaxim: 👍

mmaxim:
I"m going to figure out the absolute positioning stuff when I try on other platforms, worst case this just reverts to what it does in master.

PR #16276, opened 2019-02-27
Show errors when adding to teams from profile view

cjb:
@keybase/react-hackers 

When adding someone to teams through their profile view, we're no longer displaying any errors if one of the teams fails.  This broke when identifyv3 landed: we were displaying the errors overlaid on the main profile view via `state.teams.addUserToTeamsResults`,  and the new identifyv3 components no longer try to do that.

While talking about it, we decided that they shouldn't be being displayed this way anyway, even though it's in the design.  It makes more sense to keep the modal showing if there were errors and display them, else just dismiss the modal if the teams adds were all successful.

So this PR moves the error display into the `profile/add-to-team` component, updates the styles on that component, and adds logic to dismiss the modal it if everything was successful.  This logic is controlled by redux and componentDidUpdate, not the saga.

Along the way I found a nasty bug: because of using `Object.keys(selectedTeams)` without checking for truth in each key's value, it was possible to use this modal to check and then uncheck a team, then check a different team, submit, and have the saga add the user to *both* teams even though one was now unselected.


shared/profile/add-to-team/container.js
```diff
@@ -54,11 +55,13 @@ const mapDispatchToProps = (dispatch, {navigateUp, routeProps, navigateAppend})
       ])
     )
   },
+  clearAddUserToTeamsResults: () => dispatch(TeamsGen.createClearAddUserToTeamsResults()),
```
chrisnojima: why both?

shared/profile/add-to-team/container.js
```diff
@@ -109,6 +112,7 @@ export default compose(
   ),
   lifecycle({
     componentDidMount() {
+      this.props.clearAddUserToTeamsResults()
       this.props.loadTeamList()
```
chrisnojima: could this just be a side effect of load

shared/profile/add-to-team/container.js
cjb: Thanks, removed one.

shared/profile/add-to-team/container.js
cjb: Can't do that: we need to reload the team list immediately after getting an error, so that the text next to each checkbox stays correct for the case where some adds succeeded and some failed, so that would dismiss the error immediately.

chrisnojima: could be a flag to load

cjb: There's another reason not to do this: when you toggle any checkbox while seeing an error, we want to clear errors (since they no longer apply to your selection) but not reload teams.  So I think it's correct that they're separate actions.

PR #16280, opened 2019-02-27
show more generic proof types behind a feature flag

xgess:
### before
```
$ keybase prove -l
Supported services are: dns, github, hackernews, http, https, reddit, rooter, twitter, web.
```
### after
```
$ keybase prove -l
Supported services are: dns, facebook, github, hackernews, http, https, mastodon.social, mastokeyhole.herokuapp.com, reddit, rooter, twitter, web.
```

see https://github.com/keybase/keybase/pull/3522

### DIFF
--- a/go/libkb/features.go
+++ b/go/libkb/features.go
@@ -64,8 +64,9 @@ type FeatureFlagSet struct {
 }
 
 const (
-	FeatureFTL     = Feature("ftl")
-	FeatureIMPTOFU = Feature("imptofu")
+	FeatureFTL                = Feature("ftl")
+	FeatureIMPTOFU            = Feature("imptofu")
+	ExperimentalGenericProofs = Feature("experimental_generic_proofs")
 )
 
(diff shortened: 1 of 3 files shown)


mlsteele:
There's already bypass. Does this solve a problem for frontend? Is this flag set on mobile?

xgess:
yeah - @joshblum said it would be helpful for testing on mobile, because we can't set environment variables.

mlsteele:
The sock puppet accounts won't be admins though

xgess:
@mlsteele but that's an easy thing to toggle on and off without a server deployment

xgess:
@joshblum now with feature flag

mlsteele:
Add the flag to `ShouldUseParameterizedProofs`
https://github.com/keybase/client/blob/2219b50f7644e58dc8a0a4d7f90db6c06d372745/go/libkb/globals.go#L1341

`ShouldUseParameterizedProofs` doesn't just return true since users shouldn't see a suggestion until creation works in the gui.

PR #16282, opened 2019-02-27
Admins can now remove pukless post-reset ex-owners

mlsteele:
1. alice%1 is an admin of team T.
2. bob%1 is an owner of T.
3. bob resets but does not get a PUK.
4. alice re-admits bob as an admin. Now bob%1 is a cryptomember, but shouldn't get keyed for since that's pre-reset. Alice just added bob%0 as an admin sigchain-invite.
5. alice removes bob from the team using `TeamRemoveMember`.

Step 5 would fail because alice's client would try to remove the old bob%1 cryptomember. But that's not allowed since an admin can't remove an owner.

This fixes that flow by making alice's client instead remove the bob%0 invite.

### DIFF
--- a/go/teams/chain.go
+++ b/go/teams/chain.go
@@ -552,7 +552,7 @@ func (t *TeamSigChainState) IsInviteObsolete(id keybase1.TeamInviteID) bool {
 	return ok
 }
 
-// FindActiveKeybaseInvite finds and returns *first* Keybase-type
+// FindActiveKeybaseInvite finds and returns a Keybase-type
 // invite for given UID. Ordering here is not guaranteed, caller
 // shouldn't assume that returned invite will be the oldest/newest one
 // for the UID.
--- a/go/teams/service_helper.go
+++ b/go/teams/service_helper.go
@@ -672,7 +672,6 @@ func MemberRole(ctx context.Context, g *libkb.GlobalContext, teamname, username 
 }
 
 func RemoveMember(ctx context.Context, g *libkb.GlobalContext, teamname, username string) error {
-
 	var inviteRequired bool
 	uv, err := loadUserVersionByUsername(ctx, g, username, false /* useTracking */)
 	if err != nil {
@@ -687,6 +686,15 @@ func RemoveMember(ctx context.Context, g *libkb.GlobalContext, teamname, usernam
 		err = nil
 	}
 
+	me, err := loadMeForSignatures(ctx, g)
+	if err != nil {
+		return err
+	}
+
+	if me.GetNormalizedName().Eq(libkb.NewNormalizedUsername(username)) {
+		return Leave(ctx, g, teamname, false)
+	}
+
 	return RetryOnSigOldSeqnoError(ctx, g, func(ctx context.Context, _ int) error {
 		t, err := GetForTeamManagementByStringName(ctx, g, teamname, true)
 		if err != nil {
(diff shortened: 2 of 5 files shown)


go/teams/service_helper.go
```diff
@@ -687,38 +686,37 @@ func RemoveMember(ctx context.Context, g *libkb.GlobalContext, teamname, usernam
...
 		}
 
-		me, err := loadMeForSignatures(ctx, g)
+		existingUV, err := t.UserVersionByUID(ctx, uv.Uid)
 		if err != nil {
-			return err
-		}
-
-		if me.GetNormalizedName().Eq(libkb.NewNormalizedUsername(username)) {
-			return Leave(ctx, g, teamname, false)
```
zapu: oh nice catch doing that nested `RetryOnSigOldSeqnoError`

go/teams/service_helper.go
```diff
@@ -687,38 +686,37 @@ func RemoveMember(ctx context.Context, g *libkb.GlobalContext, teamname, usernam
...
-		if err != nil {
-			// Try to remove as an keybase-invite
-			if ierr := removeKeybaseTypeInviteForUID(ctx, g, t, uv.Uid); ierr == nil {
-				return nil
-			}
-			return libkb.NotFoundError{Msg: fmt.Sprintf("user %q is not a member of team %q", username,
-				teamname)}
+		if _, _, found := t.FindActiveKeybaseInvite(uv.Uid); found {
+			// Remove keybase invites.
+			return removeKeybaseTypeInviteForUID(ctx, g, t, uv.Uid)
```
zapu: The semantics here seem a bit weird - `FindActiveKeybaseInvite` only looks for active invites, but `removeKeybaseTypeInviteForUID` will remove either active or obsolete invite from the chain. This seems to make sense in this context, but was that something that you considered?

I wonder if calling `removeKeybaseTypeInviteForUID` alone and "falling through" on `libkb.NotFoundError` could also work - but this would change the logic slightly wrt obsoleted invites.

go/teams/teams.go
```diff
@@ -569,7 +569,7 @@ type ChangeMembershipOptions struct {
 }
 
 func (t *Team) ChangeMembershipWithOptions(ctx context.Context, req keybase1.TeamChangeReq, opts ChangeMembershipOptions) (err error) {
-	defer t.G().CTrace(ctx, "Team.ChangeMembershipPermanent", func() error { return err })()
```
zapu: :+1:

zapu:
I wonder if the scenario that you described in pull request description could warrant a separate test - even if it's covered by the parametrized testing - just to make it clear what the issue was.

go/teams/chain.go
```diff
@@ -552,7 +552,7 @@ func (t *TeamSigChainState) IsInviteObsolete(id keybase1.TeamInviteID) bool {
 	return ok
 }
 
-// FindActiveKeybaseInvite finds and returns *first* Keybase-type
+// FindActiveKeybaseInvite finds and returns a Keybase-type
```
zapu: thanks. the intention was good but I might have made it more confusing rather than less

go/teams/service_helper.go
zapu: in grand scheme of things it probably doesn't matter since we try hard not to leave any obsolete invites anymore, but there might be a number of them from older, simpler, times.

I'm :+1:  in general but I want to know what you think. Thanks!

mlsteele:
`TestTeamResetReAddRemoveAdminOwnerPukless` is exactly that

go/teams/service_helper.go
mlsteele: I didn't consider that. At this point it seems like it's hard to prove invariants so paths have to be able to function on large input spaces.

zapu: no worries, I spent some time trying to find a test case where it would break but couldn't get a simple one.

PR #16295, opened 2019-02-27
simplefs: add RPC and CLI flag for getting team quota usage

strib:
Issue: KBFS-3905


jakob223:
Not as a part of this ticket necessarily, but it would be nice to include remaining too so users don't need to subtract, especially when remaining is measured in KB and the others in GB.

strib:
> Not as a part of this ticket necessarily, but it would be nice to include remaining too so users don't need to subtract, especially when remaining is measured in KB and the others in GB.

Good suggestion, I made a ticket.  Thanks!

PR #16296, opened 2019-02-27
show reload on profile error and plumb through more errors

chrisnojima:
@keybase/react-hackers


chrisnojima:
you can either edit the service to give and error you can kill the service while you're in the middle of iding

PR #16298, opened 2019-02-28
add identities button

zanderz:


zanderz:
So this should be done over again after your big identv3 PR it sounds like.

chrisnojima:
Nothing is blocked here just put it in the right place

zanderz:
ok @chrisnojima how about now?

shared/__mocks__/feature-flags.js
```diff
@@ -7,6 +7,7 @@ if (!__STORYBOOK__) {
 }
 
 const ff: FeatureFlags = {
+  addIdentity: true,
```
chrisnojima: this feature flag should be used for all the generic proofs so it should likely be about that, not just this one button

shared/styles/colors.js
```diff
@@ -16,6 +16,7 @@ const colors = {
   black_50_on_white: 'rgb(127,127,127)',
   black_60: 'rgba(0, 0, 0, 0.60)',
   black_63: 'rgba(0, 0, 0, 0.63)',
+  black_75: 'rgba(0, 0, 0, 0.75)',
```
chrisnojima: @cecileboucheron is this correct

adamjspooner: `black` (at 85%) replaced `black_75`.

shared/profile/user/index.js
```diff
@@ -484,6 +505,9 @@ const styles = Styles.styleSheetCreate({
   }),
   innerContainer: {...Styles.globalStyles.fillAbsolute},
   invisible: {opacity: 0},
+  label: {
+    color: Styles.globalColors.black_75,
```
adamjspooner: This should be `black`.

shared/styles/colors.js
zanderz: so the zeplin file is not the source of truth? https://app.zeplin.io/project/56b25ffc64fc91773a73aaf8/screen/5bf33ccc2790467ebf9a9bfc

cecileboucheron: No let's remove that line and use `globalColors.black`. We darkened our blackest black and decided to just call it `black`.

shared/__mocks__/feature-flags.js
zanderz: :white_check_mark:

shared/styles/colors.js
zanderz: :white_check_mark:

shared/profile/user/index.js
zanderz: :white_check_mark:

shared/profile/user/index.js
```diff
@@ -250,6 +252,17 @@ class BioTeamProofs extends React.PureComponent<BioTeamProofsProps> {
           </Kb.Text>
           <Teams username={this.props.username} />
           <Proofs {...this.props} />
+          {flags.proofProviders && (
+            <Kb.Box2 direction="horizontal" style={styles.addIdentityContainer}>
```
chrisnojima: this block should only show up given some other kind of prop. as 1. it should only show if there are additional identities and 2. if should only show on your own profile, so we at least need `props.showOtherIdentities` or something

zanderz: :white_check_mark:

PR #16301, opened 2019-02-28
ICS parser tweak

mlsteele:
The requirement that an ICS starts and ends with a space (ish) was messing it up due to overlapping matches of the regex.
```
+25eur +25eur +25eur
was matching into
(+25eur )+25eur( +25eur)

whereas ideally we want
+25eur +25eur +25eur
(+25eur) (+25eu) (+25eur)
which is to say
(+25eur[ )+25eur( ]+25eur)
```

Basically I think `FindAll` can't handle finding all the `+25eur`'s surrounded by spaces (ish). This patch instead of using `FindAll` uses a loop that instead of advancing each time to the end of the regex match, advances to a capture group before the last space.


go/chat/wallet/parser.go
```diff
@@ -10,21 +10,24 @@ import (
...
 		// Currency code
 		`([A-Za-z]{2,6})` +
 		// At sign and username, optional for direct messages
-		// If not used, must be followed by a non-tx-character or end of string
-		`(?:(?:@((?:[a-zA-Z0-9]+_?)+))|(?:[^A-Za-z@]|\z)))`,
+		`(?:@((?:[a-zA-Z0-9]+_?)+))?` +
+		// Sentinel group for advancing the scan
+		`()` +
+		// Must be followed by a nice character or the end of the string.
+		`(?:[\s)\]}:;.,!?"']|\z)`,
```
heronhaye: maybe add a comment about \z not working inside the []s

heronhaye: do we want to support `+5xlm(for lunch)`? (I'd say no, just wanted to make sure though)

mlsteele: nah it's not a function

mlsteele:
Made a test pass by just adding a space. One of those with a lot of backticks that I don't know whether they're real. You basically can't have a payment adjacent to real matching ticks right now.

heronhaye:
by "real" do you mean ticks that aren't part of a codeblock or codeline? That sounds fine to me

mlsteele:
By "real" I meant whether they're testing something that matters. Whether I papered over a regression or not

heronhaye:
Maybe you could get the diff of testcases from before your changes and see what Patrick says? They all seem reasonable to me though

PR #16303, opened 2019-02-28
idv3 cleanup

chrisnojima:
@keybase/react-hackers this cleans up the old tracker/idv2 code now that its not needed anymore
- [x]  remove old stuff
- [x] i noticed a bunch of imports of isMobile from constants/platform even though its available in styles, so i removed a bunch of those
- [x] added a new tracker2 action to showUser to make the card showing code easier. With the new nav2 this'll all go away as we'll stop showing tracker for this anyways

- [x] after release


zanderz:
one thing though: this leaves no profile in storybook?

chrisnojima:
it can be added. tracker2 exists

PR #16304, opened 2019-02-28
ICS skips known non-currency codes

mlsteele:
If you send +10eur or +10xlm it will ICS. But if you send +10ish or +10banana it will no longer ICS, the message goes straight through. The caveat being that the currencies must be loaded for some reason. I think if you don't have a wallet that doesn't happen. So people with no wallets still can't send +10ish without a bad experience.

### DIFF
--- a/go/chat/wallet/sender.go
+++ b/go/chat/wallet/sender.go
@@ -119,6 +119,9 @@ func (s *Sender) ParsePayments(ctx context.Context, uid gregor1.UID, convID chat
 	parts, membersType, err := s.getConvParseInfo(ctx, uid, convID)
 	for _, p := range parsed {
 		var username string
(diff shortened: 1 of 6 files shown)


go/chat/wallet/sender.go
```diff
@@ -119,6 +119,9 @@ func (s *Sender) ParsePayments(ctx context.Context, uid gregor1.UID, convID chat
 	parts, membersType, err := s.getConvParseInfo(ctx, uid, convID)
 	for _, p := range parsed {
 		var username string
+		if known, ok := s.G().GetStellar().KnownCurrencyCodeInstant(ctx, p.CurrencyCode); ok && !known {
```
patrickxb: can you add a comment on this line explaining ok && !known

PR #16306, opened 2019-02-28
explicit set fuse options

jinyangli:
Explicitly set fuse.MaxReadahead to make FUSE issue large READ (128KB)
Explicitly set fuse.AsyncRead to make FUSE issue concurrent READs

i have problem testing OSX kbfs.... so i did not do the equivalent change in mount_osx.go


jzila:
Tested on OSX and read perf is already great. 16MB read-ahead!

PR #16308, opened 2019-02-28
very simple RPC frontend can spam to see if app is out-of-date

maxtaco:


maxtaco:
cc @chinmaykunkikar

maxtaco:
cc @chrisnojima

PR #16226, opened 2019-02-25
ICS no-wallet error presentation

mlsteele:
image

PR #16313, opened 2019-02-28
Fix stuck ics error modal

mlsteele:


mlsteele:
Actually now it doesn't get stuck but if you rapid fire it randomly closes. This is definitely the same bug that's plaguing https://github.com/keybase/client/pull/16226

mlsteele:
rolled into https://github.com/keybase/client/pull/16226

PR #16316, opened 2019-02-28
error handling improvements for Files

songgao:


shared/actions/fs/platform-specific.desktop.js
```diff
@@ -315,38 +314,37 @@ const openAndUpload = (state, action) =>
...
-        })
-      ),
-      FsGen.createUserFileEditsLoaded({tlfUpdates}),
-    ]
-  })
+  RPCTypes.SimpleFSSimpleFSUserEditHistoryRpcPromise().then(writerEdits =>
+    FsGen.createUserFileEditsLoaded({
+      tlfUpdates: Constants.userTlfHistoryRPCToState(writerEdits || []),
+    })
+  )
```
songgao: I thought we needed to do `SimpleFSStat` for each in edit history in order to properly display icons in the widget, but looks like not. My guess is we needed to have the thing in `state.fs.pathItems` before try to display. That has changed since we had KBFS links in chat, and now we just load `PathItem` on-demand. So just get rid of this.

shared/actions/json/fs.json
```diff
@@ -26,11 +26,12 @@
       "public": "I.Map<string, Types.Tlf>",
       "team": "I.Map<string, Types.Tlf>"
     },
-    "filePreviewLoad": {
+    "pathItemLoad": {
       "path": "Types.Path",
-      "identifyBehavior?": "?RPCTypes.TLFIdentifyBehavior"
+      "identifyBehavior?": "?RPCTypes.TLFIdentifyBehavior",
+      "silenceErrors?": "?boolean"
```
songgao: This ends up not needed in this PR as we actually don't need this in widget, but since I already did it, I'm leaving it here in case we need it later.

shared/actions/fs/platform-specific.desktop.js
```diff
@@ -316,38 +315,37 @@ const openAndUpload = (state, action) =>
...
 
+let pingKbfsDaemonLoopRunning = false
+function* pingKbfsDaemonLoop() {
+  if (pingKbfsDaemonLoopRunning) {
+    return
+  }
+  while (true) {
+    pingKbfsDaemonLoopRunning = true
+    const state = yield* Saga.selectState()
+    yield Saga.delay(state.fs.kbfsDaemonConnected ? 10 * 1e3 : 1e3)
```
jzila: `1e3` can just be `1000`--it's easier to read.

shared/actions/fs/platform-specific.desktop.js
```diff
@@ -316,38 +315,37 @@ const openAndUpload = (state, action) =>
...
   ConfigGen.createShowMain(),
   ...(path ? [FsGen.createOpenPathInFilesTab({path})] : [RouteTreeGen.createSwitchTo({path: [Tabs.fsTab]})]),
 ]
 
+let pingKbfsDaemonLoopRunning = false
+function* pingKbfsDaemonLoop() {
+  if (pingKbfsDaemonLoopRunning) {
+    return
+  }
+  while (true) {
```
jzila: Why don't we ever return from this loop?

songgao: What I have in mind is we'd continue monitoring KBFS connections here even after KBFS is connected. It's possible to restart KBFS without restarting the whole app (e.g. enabling fuse ext).

songgao: @jzila does it sound reasonable?

jzila: I'm a little worried about such a loop. @chrisnojima is this a reasonable approach or should we figure out something different?

chrisnojima: we don't do this kind of polling a lot. is it really necessary

jzila: Should we have a universal service->GUI RPC concerning connection status notifications?

songgao: That'd be useful. I guess until then we'll just poll at start, and stop polling when KBFS is connected?

jzila: Sounds good to me. But what happens if KBFS never connects?

songgao: that'll just keep polling. We do expect KBFS to connect eventually right? Or is KBFS-less client something we want to support?

jzila: Not to actively support it, I'm just trying to think of bugs when someone's KBFS has crashed or whatever.

songgao: If KBFS crashes at some point after connected, we'd think we're still connected because the polling has stopped (as of last change). If KBFS crashes at start, that'd be a rare scenario where we likely have other problems too. This RPC is relatively cheap, so I think it's not a big deal. In most cases we'll detect we're connected and be done with it.

PR #16317, opened 2019-02-28
Critical out of date banner

chrisnojima:
This calls the new critical out of date rpc. We call it once every hour. Once it returns critical you get a perma banner on top telling you what to do.
On desktop the message component must contain some kind of description / action to get help (link to install or something)
On mobile there is an upgrade button which sends you to the correct app store

- [x] tested android nav1
- [x] tested android nav2
- [x] tested ios nav1
- [x] tested ios nav2
- [x] tested desktop nav1
- [x] tested desktop nav2


shared/constants/config.js
```diff
@@ -80,6 +80,8 @@ export const urlToUsername = (url: {
 export const makeState: I.RecordFactory<Types._State> = I.Record({
   appFocused: true,
   appFocusedCount: 0,
+  appOutOfDateMessage: 'aaa',
+  appOutOfDateStatus: 'critical',
```
songgao: Any reason not to use https://github.com/keybase/client/blob/59a25dce1e8758a890e50b2767fbd6625f838915/shared/constants/types/config.js#L42 ? It'd be nice to unify this with the one in widget.

chrisnojima: we should talk to max

chrisnojima:
cc: @songgao @maxtaco

songgao:
Oh I see this is using configGetUpdateInfo2 not configGetUpdateInfo. What's the difference? If 2 is better it seems we should just deprecate the original altogether, and merge the result into `config.outOfDate` so widget is consistent.

maxtaco:
I rewrote the system because the existing one was pretty bad and didn't handle all platforms properly (it mashed up darwin and iOS, for instance).  This one is fine to use elsewhere. It is very simple and is plumbed all the way through to the API server.

songgao:
Sweet! @chrisnojima sounds like we can remove all outOfDate stuff in this file: https://github.com/keybase/client/blob/master/shared/actions/platform-specific/index.desktop.js and have your code mutate config.outOfDate instead?

songgao:
Actually I jumped gun. @maxtaco does the new API handle normal out-of-date as well, or just critical ones? Currently the widget actually runs the updater binary to check if there's new update available at all, but only use the API call to figure out if it's critical.

maxtaco:
@songgao  I saw you were working on something similar but didn't want to get in the way, since we were hoping to put this together very quickly.  But if what you were working on is based on the old way of getting out-of-dateness (as a weirdly non-deterministic side effect of other RPCs), I recommend this system because it's simpler and less likely to flake.  The old one that power the toasts was pretty unreliable

songgao:
@maxtaco I worked on that months ago; just forgot to turn it on for admins. What I described above is how it works in the current widget, and I'd be happy to have it migrated over. Just need to know if the new API handles the part where we just pushed an update, but it's not critical or security related.

maxtaco:
@songgao it doesn't handle that, it's basically just plumbed through to this:

```iced
  _handle : (cb) ->
    esc = make_esc cb, "_handle"
    await @_parse_input {}, esc defer()

    d = mm.config.clients.pkg.check
    platkey= @platform.toLowerCase()
    plat = if d.hasOwnProperty(platkey) then d[platkey] else d.default
    psc = constants.constants.pkg.status

    # For now, just say that all updates are critical. In the future, we might also have
    # suggested updates... best to launch clients today with that capability.
    res = {}
    if semver.gte(@version, plat.min)
      res.status = psc.OK
    else
      res.status = psc.CRITICAL
      res.critical = plat
    @pub { res }
    cb null
```

songgao:
Understood; Thanks for the context! I'm OK with this split for now then. I can come back to it later to merge the two so that 1) GetUpdateInfo2 checks for updates from the updater too; 2) widget uses GetUpdateInfo2.

maxtaco:
Or we can just have 2 that work slightly differently.  The real crucial thing is that we need to support `GetUpdateInfo2` as is indefinitely, it can never break.

shared/actions/config/index.js
```diff
@@ -430,6 +430,35 @@ const newNavigation = (_, action) => {
...
+      switch (s.status) {
+        case RPCTypes.configUpdateInfoStatus2.ok:
+          break
+        case RPCTypes.configUpdateInfoStatus2.suggested:
+          status = 'suggested'
+          message = s.suggested?.message
+          break
+        case RPCTypes.configUpdateInfoStatus2.critical:
+          status = 'critical'
+          message = s.critical?.message
```
songgao: Does this make message `null` if `critical` is `null`? What is it called / what can I google to learn more about it?

shared/actions/config/index.js
```diff
@@ -534,6 +563,7 @@ function* configSaga(): Saga.SagaGenerator<any, any> {
   // Kick off platform specific stuff
   yield Saga.spawn(PlatformSpecific.platformConfigSaga)
   yield Saga.spawn(avatarSaga)
+  yield Saga.spawn(criticalOutOfDateCheck)
```
songgao: No need to wait for `installerRan`? Is this guaranteed to run after service is up?

shared/actions/platform-specific/index.native.js
```diff
@@ -364,6 +364,13 @@ const editAvatar = () =>
     })
   })
 
+const openAppStore = () =>
+  Linking.openURL(
+    isAndroid
+      ? 'http://play.google.com/store/apps/details?id=io.keybase.ossifrage'
```
songgao: https?

maxtaco:
@songgao  also larger context here is that this will likely ever run if: (1) user is on mobile or linux, with no installer; or (2) is on macos and windows and the updater is broken.  It's like a last-ditch effort, after everything else has repeatedly failed over 6 months, to get the user to update.

songgao:
Sounds good! I think I'll change `GetUpdateInfo` into `HasUpdate`, and only check with the updater.

zanderz:
I tried it and hardcoded the results to critical and saw the perma-banner, LGTM
For desktop, shouldn't it be enough to link to https://keybase.io/download ?

shared/actions/config/index.js
chrisnojima: It'll be undefined

shared/actions/config/index.js
chrisnojima: Don't think it matters. Checks every hour

shared/actions/platform-specific/index.native.js
chrisnojima: They say http. https://developer.android.com/distribute/marketing-tools/linking-to-google-play

chrisnojima:
We can link to whatever we want in the message

shared/actions/config/index.js
songgao: Is this part of es6? How do I google it?

shared/actions/config/index.js
songgao: It'd probably black bar the first time if it races. Also it's nice to not miss the first one.

chrisnojima: No black bar there's a try

shared/actions/config/index.js
chrisnojima: https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining

shared/actions/platform-specific/index.native.js
songgao: I wonder if that's intentional It gets redirected to https in browser. I'm not sure how it works on Android; I guess if its skips browser and goes to the app store directly it's fine.

shared/actions/config/index.js
songgao: Ah you're right. OK

PR #16318, opened 2019-02-28
Add designs for each kind of flip.

adamjspooner:
@keybase/react-hackers, this is a draft, but I'd appreciate any feedback now. There are a couple of small design tweaks left to go (and one bug Mike found).

I'm not going to update the storyshots until it's ready for a full review.


shared/chat/conversation/messages/coinflip/results.js
```diff
@@ -0,0 +1,346 @@
...
+    </Kb.Box2>
+    <Kb.Box2 direction="vertical" centerChildren={true}>
+      <Kb.Text selectable={true} type="Header">
+        {props.coin ? 'Heads!' : 'Tails!'}
+      </Kb.Text>
+    </Kb.Box2>
+  </Kb.Box2>
+)
+
+type HandType = {|
```
chrisnojima: just fyi you don't need any of these types inside this file. they'll all get inferred

shared/chat/conversation/messages/coinflip/results.js
```diff
@@ -0,0 +1,346 @@
...
+)
+
+type ShuffleType = {|
+  shuffle: ?Array<string>,
+|}
+const CoinFlipResultShuffle = (props: ShuffleType) => (
+  <Kb.Box2 direction="vertical" alignSelf="flex-start" gap="xtiny">
+    {props.shuffle &&
+      props.shuffle.slice(0, 5).map((item, i) => (
+        <Kb.Box2 key={i} direction="horizontal" alignSelf="flex-start" centerChildren={true}>
```
chrisnojima: should just pull this out into its own component

shared/util/string.js
```diff
@@ -1,8 +1,8 @@
 // @flow
 
 // Add pluralization rules as necessary
-function pluralize(str: string): string {
-  return str.endsWith('s') ? str : `${str}s`
+function pluralize(str: string, count?: number): string {
+  return count && count === 1 ? str : str.endsWith('s') ? str : `${str}s`
```
chrisnojima: i think this can just be count === 1

shared/chat/conversation/messages/coinflip/container.js
```diff
@@ -27,12 +27,16 @@ const mapStateToProps = (state, {flipGameID}: OwnProps) => {
     : {
         commitmentVis: status.commitmentVisualization,
         errorInfo: status.errorInfo,
+        hasCommitments: status.phase > RPCChatTypes.chatUiUICoinFlipPhase.commitment,
```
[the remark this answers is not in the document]
adamjspooner: And do the same for `hasSecrets` and `isComplete`?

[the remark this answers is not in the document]
chrisnojima: i think so. this way we can assume the phases are subsets of each other and thats it. if these numbers change then this logic won't be correct but that always will be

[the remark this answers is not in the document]
adamjspooner: 👌

adamjspooner:
@keybase/react-hackers, this is ready for a review. The coin image will change, but I'd like to get this in as soon as possible.

shared/chat/conversation/messages/coinflip/container.js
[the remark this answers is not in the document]
mmaxim: Yeah agreed here, this is one of the reasons @chrisnojima doesn't like when I put the RPC types this far down into the components. It's best to treat these values as something you really know nothing about (in terms of their exact nature), and to make checks like this explicit.

shared/chat/conversation/messages/coinflip/index.js
```diff
@@ -39,63 +45,91 @@ class CoinFlip extends React.Component<Props, State> {
...
-  render() {
-    const popup = (
-      <CoinFlipParticipants
-        attachTo={this._getAttachmentRef}
-        onHidden={this._hidePopup}
-        participants={this.props.participants}
-        visible={this.state.showPopup}
-      />
+  _renderStatusText = () => {
+    const participants = (
```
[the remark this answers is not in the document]
chrisnojima: It may be better to encode this as one value since they aren't orthogonal to each other. Like maybe just use phase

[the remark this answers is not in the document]
mmaxim: You basically just don't want to do this if there is an error, so you could check for that.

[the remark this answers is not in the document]
adamjspooner: How's this? https://github.com/keybase/client/pull/16318/commits/4b71278605dfa7e8a3be4dc99d3c01eea4cbed77

adamjspooner:

shared/chat/conversation/messages/coinflip/container.js
```diff
@@ -27,11 +28,14 @@ const mapStateToProps = (state, {flipGameID}: OwnProps) => {
...
         errorInfo: status.phase === RPCChatTypes.chatUiUICoinFlipPhase.error ? status.errorInfo : null,
+        isError: status.phase === RPCChatTypes.chatUiUICoinFlipPhase.error,
         participants: status.participants || [],
+        phase: Constants.flipPhaseToString(status.phase),
         progressText: status.progressText,
+        resultInfo: status.resultInfo,
         resultText: status.resultText,
         revealVis: status.revealVisualization,
-        showParticipants: participantStatuses.indexOf(status.phase) >= 0,
+        showParticipants: participantStatuses.indexOf(status.phase) === 2,
```
chrisnojima: can't this be phase === 'complete'

shared/chat/conversation/messages/coinflip/index.js
```diff
@@ -3,16 +3,20 @@ import * as React from 'react'
...
 import CoinFlipError from './errors'
+import CoinFlipResult from './results'
 
 export type Props = {|
   commitmentVis: string,
   revealVis: string,
   resultText: string,
   errorInfo?: ?RPCChatTypes.UICoinFlipError,
   participants: Array<RPCChatTypes.UICoinFlipParticipant>,
+  phase?: 'commitments' | 'secrets' | 'complete' | 'loading',
```
chrisnojima: doesn't have to be optional now right

shared/chat/conversation/messages/coinflip/index.js
```diff
@@ -39,63 +43,99 @@ class CoinFlip extends React.Component<Props, State> {
     }, 0)
     return `${revealed} / ${total}`
   }
+  _renderStatusText = () => {
+    if (this.props.showParticipants) {
+      const participants = (
```
chrisnojima: 🐑 can put the ismobile check here and then its just `{participants}` below

shared/chat/conversation/messages/coinflip/results.js
```diff
@@ -0,0 +1,332 @@
...
+
+type CardType = {|
+  card: number,
+  hand?: boolean,
+|}
+const Card = (props: CardType) => (
+  <Kb.Box2
+    direction="vertical"
+    centerChildren={true}
+    // gap={Styles.isMobile ? undefined : 'xtiny'}
```
chrisnojima: 🔥

shared/chat/conversation/messages/coinflip/results.js
```diff
@@ -0,0 +1,332 @@
...
+        style={Styles.collapseStyles([styles.listOrder, props.index === 0 && styles.listOrderFirst])}
+      >
+        {props.index + 1}
+      </Kb.Text>
+    </Kb.Box2>
+    <Kb.Markdown
+      allowFontScaling={true}
+      styleOverride={
+        props.index === 0
+          ? {
```
chrisnojima: can pull this into a constant below

shared/chat/conversation/messages/coinflip/index.js
adamjspooner: I'm not sure how to do that. The `isMobile` check is to either next it in a `Text` block or not.

adamjspooner:
@chrisnojima, ready for one more look? 🤞

PR #16336, opened 2019-03-01
New new chat button and filter style

buoyad:
- Don't use `HeaderHocHeader` on native, unify desktop + mobile impl
- New chat button
- (x) icon on the right when filtering

Small screen fit (5S) - nav2 off:
image
nav2 on:
image

r? @keybase/react-hackers


buoyad:
@keybase/react-hackers bump

PR #16337, opened 2019-03-01
Limit the size of the loader's internal storage

patrickxb:


mmaxim:
Could you just use an LRU?

patrickxb:
This seemed a little simpler

PR #16341, opened 2019-03-01
fix bug in leaving as the last owner

xgess:
a team is not valid without an owner, so the last owner may not leave it. in JS land, we don't bother letting a user make this attempt and instead show a helpful explanation. 

previously, we were showing this message only if you were the very last remaining user in the team. this is a sufficient but not necessary condition for being the final owner, for example if there is another non-owner user in the team. in this case we should also be preventing the request instead of letting the it go through and black-bar-ing.

### DIFF
--- a/shared/constants/teams.js
+++ b/shared/constants/teams.js
@@ -265,6 +265,12 @@ const hasChannelInfos = (state: TypedState, teamname: Types.Teamname): boolean =
 const getTeamMemberCount = (state: TypedState, teamname: Types.Teamname): number =>
   state.teams.getIn(['teammembercounts', teamname], 0)
 
+const isLastOwner = (state: TypedState, teamname: Types.Teamname): boolean => {
+  const allTeamMembers = state.teams.teamNameToMembers.getIn([teamname], I.Map())
+  const countOfOwners = allTeamMembers.count(m => m.type === 'owner')
+  return countOfOwners === 1 && getRole(state, teamname) === 'owner'
+}
+
 const getTeamID = (state: TypedState, teamname: Types.Teamname): string =>
   state.teams.getIn(['teamNameToID', teamname], '')
 
@@ -463,6 +469,7 @@ export {
   hasChannelInfos,
   getEmailInviteError,
   getTeamMemberCount,
+  isLastOwner,
   userIsActiveInTeamHelper,
   getTeamChannelInfos,
   getChannelInfoFromConvID,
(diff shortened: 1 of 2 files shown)


xgess:
@keybase/react-hackers  ^

shared/constants/teams.js
```diff
@@ -265,6 +265,12 @@ const hasChannelInfos = (state: TypedState, teamname: Types.Teamname): boolean =
 const getTeamMemberCount = (state: TypedState, teamname: Types.Teamname): number =>
   state.teams.getIn(['teammembercounts', teamname], 0)
 
+const isLastOwner = (state: TypedState, teamname: Types.Teamname): boolean => {
+  const allTeamMembers = state.teams.teamNameToMembers.getIn([teamname], I.Map())
```
chrisnojima: can be
`state.teams.teamNameToMembers.get(teamname, I.Map())`

shared/constants/teams.js
```diff
@@ -265,6 +265,12 @@ const hasChannelInfos = (state: TypedState, teamname: Types.Teamname): boolean =
...
+  const countOfOwners = allTeamMembers.count(m => m.type === 'owner')
```
chrisnojima: not a big deal but if you check your own role you can opt out of looking at this entire list of members

chrisnojima: also not a big deal but you could bail earlier than going through the whole list

shared/constants/teams.js
xgess: done

shared/constants/teams.js
xgess: done

xgess: done

shared/constants/teams.js
```diff
@@ -265,6 +265,21 @@ const hasChannelInfos = (state: TypedState, teamname: Types.Teamname): boolean =
 const getTeamMemberCount = (state: TypedState, teamname: Types.Teamname): number =>
   state.teams.getIn(['teammembercounts', teamname], 0)
 
+const isLastOwner = (state: TypedState, teamname: Types.Teamname): boolean =>
+  isOwner(getRole(state, teamname)) && !isMultiOwnerTeam(state, teamname)
+
+const isMultiOwnerTeam = (state: TypedState, teamname: Types.Teamname): boolean => {
+  let countOfOwners = 0
+  const allTeamMembers = state.teams.teamNameToMembers.get(teamname, I.Map())
+  const moreThanOneOwner = allTeamMembers.valueSeq().some(tm => {
```
chrisnojima: don't need valueseq here, Map has a some() function

xgess: done

PR #16343, opened 2019-03-01
clean up local http server info logic

songgao:


songgao:
TFW your own PR doesn't look familiar ...

PR #16351, opened 2019-03-02
make flip dealer robust to reordering

maxtaco:
- for non-leaders, it doesn't really matter which order commitments and commitmentCompletes come in.
- if chat is behaving, they should be strictly ordered, but if there's a bug, the dealer can be robust to it
- handle out-of-order delivery and add a test to show that it works


maxtaco:
cc @malgorithms 

@mmaxim we don't need to merge this, but it seems like a reasonably clean change, i'm somewhat confident in it

PR #16352, opened 2019-03-02
just messing around with b&w spheres

malgorithms:
@adamjspooner @mmaxim if either of you cares to see the spheres I was playing with in action...

it was pretty hacky so I'd clean the math up a bit. also it's generating slightly bigger than it needs. although for anti-aliasing reasons it should be more than 2X the display for retina unlike the rectangles which are pixel perfect.

also just dropped the idea of a palette that's waiting for commitments. instead it simulates an alpha channel by taking the colors 25% closer to white

### DIFF
--- a/go/chat/flipmanager.go
+++ b/go/chat/flipmanager.go
@@ -13,7 +13,6 @@ import (
 	"image/png"
 	"math"
 	"math/big"
-	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


mmaxim:
I think one thing I miss in this version is the notion of the commitments filling in, how come you decided to drop it here? I also feel like the spheres tend to get large sections of them that are totally black, just seems a little less interesting than the colored versions.

The actual implementation is pretty cool though, and I think spheres could work, but maybe with different colors?

mmaxim:
Also I think in massive flips the current implementation is a lot better, you really get a sense of the scale of the flip, whereas with this one all flip sizes seem to play out about the same.

malgorithms:
> I think one thing I miss in this version is the notion of the commitments filling in, how come you decided to drop it here?

Hmm, not sure what you mean.  They still fill in, coming up from the bottom... and they go from 75% opaque to 100% opaque when the secrets come in. is that not what you're seeing?

PR #16356, opened 2019-03-03
Info Pane Stutter

MarcoPolo:
@keybase/react-hackers fixes the scroll jank in the info-panel. The issue is that react-list doesn't take into account the margins of our `Divider`s (this is on purpose for them). I added some dev checks to warn you when you're using margins that will fail and warnings when your estimates are off. The estimates don't really matter a whole lot in the items at the beginning of the list since they will be rendered right away.

| Before        | After           |
| ------------- |-------------| 
| Before | After|


chrisnojima:
I'd pull out those go changes

shared/chat/conversation/info-panel/index.js
```diff
@@ -274,7 +274,7 @@ const typeSizeEstimator = (row: Row): number => {
       return 47
 
     case 'leave channel':
-      return 17
+      return 34
```
chrisnojima: Do we even need these estimates at all?

MarcoPolo: The useful ones are the participant estimates since it will affect how big the scroll container should be. If these are too off then the scroll bar jumps as the size of the list updates.

The early items don't really matter much because you haven't started scrolling yet. So we could safely remove these if we wanted to.

chrisnojima: Seems like the scroll bar jumping isn't a big deal vs having to keep these values in sync

shared/common-adapters/list.desktop.js
```diff
@@ -16,7 +17,47 @@ class List extends PureComponent<Props<any>, void> {
       return null
     }
     const item = this.props.items[index]
-    const children = this.props.renderItem(index, item)
+    let children
+
+    // If we're in dev, let's check our estimates and warn if we're using margins
+    if (__DEV__ && !!this.props.itemSizeEstimator) {
```
MarcoPolo: I realize this adds an extra div that might affect the layout of things, although I'm not sure exactly when it would affect it. It is pretty nice to have this warning instead of having to remember to check these estimates manually when you update code.

MarcoPolo:
If they're wrong the only bad thing is the scroll bad jumps. So this just
makes it jump less

shared/common-adapters/list.desktop.js
chrisnojima: ideally we could do this w/o this change in hierarchy. (ideally we'd move away from react-list to react-window anyways)
maybe we can just get a ref and iterate over the elements and get the bounding rects and compare to the list vs adding measure

PR #16360, opened 2019-03-04
fix line height for BodySmall

mmaxim:
@keybase/react-hackers 


mmaxim:
This didn't seem to change any story shots, so I think it is likely safe to release.

PR #16362, opened 2019-03-04
Fix list-following failing to print out due to server's aggressive stubbing

pzduniak:


pzduniak:
still somewhat wip, found out how tests work

maxtaco:
This is looking good so far. Just be sure not to disable stubbing for the common case, which looks like you haven't....

xgess:
this looks like what we talked about, and i can't think of anything wrong with it. but i'd feel better if someone who has more context on this code also gives you a thumbs up. the good news is that this should only effect reading of chains and only in this pretty thin slice of functionality. so that inspires some confidence.

pzduniak:
Sure, tests are there now, I didn't expect the stubbing to be so aggressive! Definitely helped with preparing the tests.

pzduniak:
looks like its that flake, josh fixed it a few hours later

PR #16363, opened 2019-03-04
fix settings precedence for linux home dir computations

maxtaco:
- Highest precendence is now `--home`, which overrides all XDG_* variables
- Then fallback to XDG_* variables
- Then fallback to home dir specification via config file (rarely if ever used)
- Then fallback to HOME directory

### DIFF
--- a/go/libkb/home.go
+++ b/go/libkb/home.go
@@ -80,14 +80,14 @@ func (x XdgPosix) MobileSharedHome(emptyOk bool) string {
 func (x XdgPosix) dirHelper(env string, prefixDirs ...string) string {
 	var prfx string
 
-	// If the use explicitly provided a `--home` directory, it overrides any XDG_*
+	// If the user explicitly provided a `--home` directory, it overrides any XDG_*
 	// variables. All XDR dirs are taken relative to that.
 	if x.getHomeFromCmd != nil {
 		prfx = x.getHomeFromCmd()
 	}
 
 	// If the command line didn't specify anything, then we're going to go with the XDG_
-	// environment variablre specification.
+	// environment variable specification.
 	if prfx == "" {
 		prfx = os.Getenv(env)
 	}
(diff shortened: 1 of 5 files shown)


maxtaco:
CC @mmaxim  / @patrickxb  / @strib on the potentially disruptive change

heronhaye:
what is the reason for this? bot library?

maxtaco:
Yes, we had a request from a bot writer

go/libkb/home_nix_test.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func TestPosix(t *testing.T) {
-	hf := NewHomeFinder("tester", nil, nil, "posix", func() RunMode { return ProductionRunMode },
+	hf := NewHomeFinder("tester", nil, nil, nil, "posix", func() RunMode { return ProductionRunMode },
```
patrickxb: maybe we could add tests for the new behavior?

patrickxb:
This seems fine in theory, and I saw the request for this change.

I just remember we struggled a lot with XDG stuff, which makes me a little nervous...

go/libkb/home.go
```diff
@@ -65,12 +79,28 @@ func (x XdgPosix) MobileSharedHome(emptyOk bool) string {
 
 func (x XdgPosix) dirHelper(env string, prefixDirs ...string) string {
 	var prfx string
-	prfx = os.Getenv(env)
-	if len(prfx) == 0 {
+
+	// If the use explicitly provided a `--home` directory, it overrides any XDG_*
+	// variables. All XDR dirs are taken relative to that.
+	if x.getHomeFromCmd != nil {
+		prfx = x.getHomeFromCmd()
```
heronhaye: So if we pass in a home directory ~/a/b/c, won't `prefixDirs` not be appended to it since the check is if `prfx == ""` on line 98?

maxtaco: Good point, that's a bug.  Will fix

PR #16364, opened 2019-03-04
Filter arguments for `chat api-listen`

zapu:
Experimental filtering for `api-listen`, for chat messages. It's at the command level, not at the service level, so the messages still go through RPC layer, but they are not printed to consumer.

Example usage:
```
keybase chat api-listen --filter-channels '[{"name":"tuserac88e1@rooter,tuser5ca897,tuserac88e0" }, {"name":"ttca897", "topic_name": "core", "members_type": "team"}]'
keybase chat api-listen --filter-channel '{"name":"tuserac88e1@rooter,tuser5ca897,tuserac88e0" }'
```

### DIFF
--- a/go/client/chat_api_listen_display.go
+++ b/go/client/chat_api_listen_display.go
@@ -56,7 +56,7 @@ func (d *chatNotificationDisplay) setupFilters(ctx context.Context, channelFilte
 			// to normalize display name without using resolve RPC.
 			conv, _, err := d.svc.findConversation(ctx, v)
 			if err != nil {
-				return fmt.Errorf("Unable to find chat channel for: %s: %s", v.Name, err.Error())
+				return fmt.Errorf("Unable to find chat channel for %q: %s", v.Name, err.Error())
 			}
 			v.Name = conv.Info.TlfName
 		} else {
--- a/go/client/chat_api_listen_display.go
+++ b/go/client/chat_api_listen_display.go
@@ -37,6 +37,13 @@ type msgNotification struct {
 	Pagination *chat1.UIPagination `json:"pagination,omitempty"`
 }
 
+type notificationDisplayFilter struct {
+	Team    string `json:"team,omitempty"`
+	Channel string `json:"channel,omitempty"`
+	// or: display name for user conversation, cannot be combined with team/channel.
+	Users string `json:"users,omitempty"`
+}
+
 func newMsgNotification(source string) *msgNotification {
 	return &msgNotification{
 		Type:   notifTypeChat,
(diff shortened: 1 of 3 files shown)


go/client/chat_api_listen_display.go
```diff
@@ -98,6 +121,42 @@ func (d *chatNotificationDisplay) formatMessage(inMsg chat1.IncomingMessage) *Me
...
+func (d *chatNotificationDisplay) matchFilters(conv *chat1.InboxUIItem) bool {
+	if len(d.filtersNormalized) == 0 {
+		// No fliters - every message is relayed.
+		return true
+	}
+	for _, v := range d.filtersNormalized {
+		if matchMembersTypeToFilter(v, conv.MembersType) && v.Name == strings.ToLower(conv.Name) {
+			// If any of the following were specified by user and differ from
+			// what was received, filter the msg out.
+			if v.TopicType != "" && conv.TopicType.String() != v.TopicType {
```
joshblum: why not combine these checks into the conditional as well?

go/client/chat_api_listen_display.go
```diff
@@ -98,6 +121,42 @@ func (d *chatNotificationDisplay) formatMessage(inMsg chat1.IncomingMessage) *Me
...
+		default:
+			return false
+		}
+	}
+	return filter.MembersType == typ.String()
+}
+
+func (d *chatNotificationDisplay) matchFilters(conv *chat1.InboxUIItem) bool {
+	if len(d.filtersNormalized) == 0 {
+		// No fliters - every message is relayed.
```
joshblum:
```suggestion
		// No filters - every message is relayed.
```

go/client/cmd_chat_api_listen.go
```diff
@@ -79,13 +118,29 @@ func sendPing(cli keybase1.SessionClient) error {
...
 	}
 
 	chatDisplay := newChatNotificationDisplay(c.G(), c.showLocal, c.hideExploding)
+
+	if err := chatDisplay.setupFilters(context.TODO(), c.channelFilters); err != nil {
+		return err
+	}
+
+	if filterLen := len(chatDisplay.filtersNormalized); filterLen > 0 {
+		c.ErrWriteLn("Message filtering is active: %d filters", filterLen)
```
joshblum:
```suggestion
		c.ErrWriteLn("Message filtering is active with %d filters", filterLen)
```

go/client/chat_api_listen_display.go
zapu: so fold all checks into one ifs? or extract to separate function or to `matchMembersTypeToFilter` (probably renamed then)? thanks

joshblum: whichever you think makes it most readable, probably putting them all into one `if` and putting each condition on a new line is easiest

PR #16371, opened 2019-03-04
Update /flip styles.

adamjspooner:


shared/chat/conversation/messages/coinflip/results.js
```diff
@@ -159,27 +160,41 @@ const CoinFlipResultCoin = (props: CoinType) => (
...
-                {hand.target}
-              </Kb.Text>
-            </Kb.Box2>
-            <CoinFlipResultDeck deck={hand.hand} hand={true} />
-          </Kb.Box2>
-        ))}
+const CoinFlipResultHands = (props: HandType) => {
+  if (!props.hands) return null
+  const handsWithCards = props.hands.filter(hand => hand.hand)
+  const handsWithoutCards = props.hands ? props.hands.slice(handsWithCards.length) : []
```
chrisnojima: why not just filter the opposite (or use partition from lodash)

adamjspooner: Good point.

PR #16374, opened 2019-03-04
Surya/core 9581/box auditor

heronhaye:


go/teams/loader.go
```diff
@@ -308,6 +308,11 @@ func (l *TeamLoader) load1(ctx context.Context, me keybase1.UserVersion, lArg ke
 		}
 	}
 
+	newMctx, shouldReload := VerifyBoxAudit(libkb.NewMetaContext(ctx, l.G()), teamID)
```
heronhaye: @mlsteele this is going to happen every team load so wanted to get your opinion on this (this is in FTL too). In the normal case it hits a LRU cache that says the team is not in jail and returns normally. On startup it needs to get the value from disk, and then fill the LRU cache with that info, so subsequent calls will be faster.

When the team is in jail (failed 6 audits spread out over 42 hours), however, which might happen if you're offline for several days in a row, or you were deleted from a team and readded later, or the server is malicious, it first tries to load the team, and if the team is not open and user is a writer or greater, rotate the team and then do a reaudit. If the reaudit fails, a blackbar appears. shouldReload is true because the team has now changed, so the load happens again, but with a context that will no-op the VerifyBoxAudit. If reaudit fails the team load still works, you just get a blackbar everytime you try to load the team (until it passes reaudit).

Totally open to making changes or changing how all of this works, let me know.

mlsteele: Ok. Using the context value to deal with reentrancy seems pretty nice compared to passing a parameter all the way down.

Seems like a failed audit should be able to cause the team to fail to load. I'm not seeing how that would happen here.

heronhaye: It doesn't right now because this code might be buggy and not play well with users who don't have puks or who have reset (though I have tested both cases, might be others). It manifests as constant blackbars in the app whenever you do anything with the team.
It might even be a good idea to put this behind a server driven feature flag for a few weeks. But eventually we could make it so it just returns an error instead.

mlsteele: Guess I was misunderstanding. I took black bars to mean team loads failed which caused operations to fail which caused black bars. But if you mean a special message is routed to the ui dedicated to showing black bars, but also everything just works.. Then ok makes sense for starters.

But watch out for the issue where when the ui gets a barrage of black bars the black bar just turns black and stops rendering text.

zapu:
Hey I added a test that shows an issue with implicit admin boxing that we've discovered. We are not going to do anything about this right now though, waiting for you to get back. Thanks!

maxtaco:
Needed:  a feature flag system that's controlled by the server as we roll this out

go/teams/box_audit.go
```diff
@@ -0,0 +1,1098 @@
...
+	if team == nil {
+		return nil, fmt.Errorf("got nil team from loader")
+	}
+
+	// If the team sigchain state was constructed with support for box summary
+	// hashes, the map will be non-nil but empty. It will only be nil if the
+	// state is cached from a team load before box summary hash support.
+	if team.GetBoxSummaryHashes() == nil {
+		if force {
+			return nil, fmt.Errorf("failed to get a non-nil box summary map after full reload")
```
zapu: there may be teams that still don't have box summary hashes even after full reload, no?

heronhaye: yes, these will fail the audit and will be rotated and retried (at which point they will pass since they now have hashes)

heronhaye:
@maxtaco featureflag added

go/teams/box_audit.go
zapu: ah, nice! that's a feature then, perfect. thanks

go/client/cmd_audit.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	return nil
+}
+
+type AuditResult struct {
+	teamID keybase1.TeamID
+	err    error
+}
+
+func (c *CmdAuditBox) Run() error {
```
maxtaco: cleaner to make one `context.Context` for this whole function, rather than lots of `context.Background`s

go/engine/box_audit_retry_background.go
```diff
@@ -0,0 +1,75 @@
...
+	"time"
+
+	"github.com/keybase/client/go/libkb"
+)
+
+var BoxAuditRetryBackgroundSettings = BackgroundTaskSettings{
+	Start:        2 * time.Minute,
+	StartStagger: 1 * time.Minute,
+	WakeUp:       1 * time.Minute,
+	Interval:     7 * time.Hour,
```
maxtaco: This might be too aggressive.  It would be nice if we can server control this for a bit, is that possible?

go/engine/box_audit_retry_background.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2018 Keybase, Inc. All rights reserved. Use of
```
maxtaco: 2019

go/engine/box_audit_scheduler_background.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2018 Keybase, Inc. All rights reserved. Use of
```
maxtaco: 2019

why do you need 2 engines here?

go/libkb/leveldb.go
```diff
@@ -367,6 +372,33 @@ func (l *LevelDb) OpenTransaction() (LocalDbTransaction, error) {
 	return ltr, nil
 }
 
+func (l *LevelDb) KeysWithPrefixes(prefixes ...[]byte) (DBKeySet, error) {
```
maxtaco: I'm not sure why this is needed....  I wouldn't have guessed it was needed... Is this going to be fast enough?

go/libkb/notify_router.go
```diff
@@ -2106,3 +2108,26 @@ func (n *NotifyRouter) HandleRootAuditError(msg string) {
 
 	n.G().Log.Debug("- merkle tree audit notification sent")
 }
+
+func (n *NotifyRouter) HandleBoxAuditError(msg string) {
```
maxtaco: Can you pass in `context.Context` and contextify any debugging?

go/service/audit.go
```diff
@@ -0,0 +1,49 @@
...
+	mctx := libkb.NewMetaContext(ctx, h.G())
+	defer mctx.TraceTimed("AuditHandler#AttemptBoxAudit", func() error { return err })()
+	return h.G().GetTeamBoxAuditor().Attempt(mctx, arg.TeamID, arg.RotateBeforeAudit), nil
+}
+
+func (h *AuditHandler) KnownTeamIDs(ctx context.Context, sessionID int) (res []keybase1.TeamID, err error) {
+	mctx := libkb.NewMetaContext(ctx, h.G())
+	defer mctx.TraceTimed("AuditHandler#KnownTeamIDs", func() error { return err })()
+	return teams.KnownTeamIDs(mctx)
+}
```
maxtaco: ----- left off here-----

go/teams/loader.go
```diff
@@ -309,6 +309,15 @@ func (l *TeamLoader) load1(ctx context.Context, me keybase1.UserVersion, lArg ke
 		}
 	}
 
+	if l.G().Env.GetFeatureFlags().HasFeature(libkb.FeatureBoxAuditor) {
+		newMctx, shouldReload := VerifyBoxAudit(libkb.NewMetaContext(ctx, l.G()), teamID)
```
zapu: `VerifyBoxAudit` will return `shouldReload` after reauditing, is this necessary if auditing did not trigger rotating? Or am I missing something?

heronhaye: Not necessary. AssertUnjailedOrReaudit will return `didReaudit=true` if it rotated, which happens iff it is reaudited.

go/teams/box_audit.go
```diff
@@ -0,0 +1,1104 @@
...
+	SenderKID keybase1.KID `json:"sender_kid"`
+	Summary   string       `json:"summary"`
+}
+
+type summaryAuditResponse struct {
+	Batches []summaryAuditBatch `json:"batches"`
+	Status  libkb.AppStatus     `json:"status"`
+}
+
+func (r *summaryAuditResponse) GetAppStatus() *libkb.AppStatus {
```
zapu: I think you can use `libkb.AppStatusEmbed` in `summaryAuditResponse`

go/teams/box_audit.go
```diff
@@ -0,0 +1,1104 @@
...
+		return nil, err
+	}
+	memberUVs := members.AllUserVersions()
+
+	uvSet := make(map[keybase1.UserVersion]bool, len(memberUVs)) // prealloc; will be at least this size
+	for _, memberUV := range memberUVs {
+		uvSet[memberUV] = true
+	}
+
+	if team.IsSubteam() {
```
zapu: should we remove this part? it probably doesn't hurt to keep it, maybe with some additional comment / note here.

go/engine/box_audit_retry_background.go
heronhaye: This is for the failed audit retry queue, so it's usually a noop. Server controls feature flag so we can make them all noops as well.

go/engine/box_audit_scheduler_background.go
heronhaye: One is for box auditing a random team every day or so. The other is for reauditing teams that failed an audit

go/libkb/leveldb.go
heronhaye: I needed some way to select a random team to audit, thought it would be easiest to just use the teamload/ftl cache. I believe finding all keys with a certain prefix is fast because of leveldb's architecture where they are sorted

go/teams/box_audit.go
heronhaye: nah, i'd rather keep this in without a comment since its correct (since I feel like we would just forget to delete the comment later)

heronhaye: test

heronhaye: test

heronhaye: test

heronhaye: test

heronhaye:
@maxtaco / @zapu all comments either thumbed-up and code changed or responded to

go/teams/box_audit.go
```diff
@@ -0,0 +1,1100 @@
...
+func loadTeamForBoxAuditInner(mctx libkb.MetaContext, teamID keybase1.TeamID, force bool) (team *Team, err error) {
+	defer mctx.TraceTimed("loadTeamForBoxAuditInner", func() error { return err })()
+	arg := keybase1.LoadTeamArg{
+		ID:              teamID,
+		ForceRepoll:     true,
+		Public:          teamID.IsPublic(),
+		ForceFullReload: force,
+	}
+
+	team, err = Load(mctx.Ctx(), mctx.G(), arg)
```
zapu: Something I've been thinking about - maybe MetaContext in this function should have a different type to signify that it's passing `SkipBoxAuditCheckContextKey` flag. In general I like how `initMctx` works in this file, but it's a bit of work to trace back where `mctx` is created to ensure it has the right flags.

go/teams/box_audit.go
```diff
@@ -0,0 +1,1100 @@
...
+	log, err := a.maybeGetLog(mctx, teamID)
+	if err != nil {
+		return ClientBoxAuditError{err}
+	}
+	if log == nil {
+		log = NewBoxAuditLog(a.Version)
+	}
+
+	isRetry := log.InProgress
+	rotateBeforeAudit := isRetry && !mctx.G().TestOptions.NoAutorotateOnBoxAuditRetry
```
zapu: Maybe I'm misunderstanding what's going on here, but I'd maybe we should save "last seen key generation" somewhere in the `BoxAuditLog` struct and only generate if the generation hasn't changed? Otherwise someone might have rotated between our retry attempts and we might be doing unnecessary rotations.

heronhaye: The audit still might fail if some reset since that rotate so it would have to be something like "audit, and if failed rotate and audit again". That could help, but I'm imagining that audits should never fail and it's unlikely that people will be auditing the same teams at the same time given how long the intervals are.

go/teams/box_audit.go
heronhaye: interesting, like a wrapper struct around a mctx?

go/teams/box_audit.go
zapu: Right, we don't expect the retry path to be usually taken so it makes sense to keep it simple here. Thanks!

heronhaye: That being said I think there is something for keeping more data for auditing purposes, like the diff of the box summaries. Max wanted to keep it simple for now though, can easily add it in later

go/teams/box_audit.go
zapu: hm I'm not sure. Maybe @maxtaco or @mlsteele have opinions on this. For this PR I think it's good as it is, but maybe as a further cleanup. I'm afraid that in future some other code path will call into this without the "right" context and things will break subtly

heronhaye: ok, I generally think things like this are a good idea (preventing errors with types). I'm not sure exactly how it would work here though, let me know if you have an idea and I'll try to implement it

go/teams/box_audit.go
```diff
@@ -0,0 +1,1099 @@
...
+	// but have not been let back into the team by an admin.
+	for uv, logPoints := range teamchain.inner.UserLog {
+		logPoint := logPoints[len(logPoints)-1]
+		if logPoint.Role == keybase1.TeamRole_NONE {
+			continue
+		}
+		merkleSeqno := logPoint.SigMeta.PrevMerkleRootSigned.Seqno
+		latest := merkleSeqno
+		if latest < merkleSeqnoCheckpoint {
+			latest = merkleSeqnoCheckpoint
```
zapu: this one took me a moment to grasp as well, can we have a comment here that:
> we gather merkle root seqnos for the points in time in which each user has been keyed in the team, which is either the merkle seqno of key rotation, or merkle seqno of change_membership if user has been added afterwards

if that is correct, of course

zapu: oh but that's only true for the case when we give it merkleSeqnoCheckpoint of key generation. maybe it's fine then

zapu: if we are making a "current summary" for recently fetched root, we should only see logPoints with root seqno < latest, right? if not, we entered another race

go/teams/box_audit.go
```diff
@@ -739,10 +723,10 @@ func NewBoxAuditJail(version Version) *BoxAuditJail {
...
 
 func (a *BoxAuditor) shouldAudit(mctx libkb.MetaContext, team Team) (bool, *keybase1.BoxAuditAttemptResult, error) {
-	if team.IsSubteam() {
-		res := keybase1.BoxAuditAttemptResult_OK_NOT_ATTEMPTED_SUBTEAM
-		return false, &res, nil
-	}
+	// if team.IsSubteam() {
+	// 	res := keybase1.BoxAuditAttemptResult_OK_NOT_ATTEMPTED_SUBTEAM
+	// 	return false, &res, nil
+	// }
```
zapu: do we want to remove this completely?

go/teams/box_audit.go
```diff
@@ -0,0 +1,1099 @@
...
+func calculateSummaryAtMerkleSeqno(mctx libkb.MetaContext, team *Team, merkleSeqno *keybase1.Seqno) (summary *boxPublicSummary, err error) {
+	defer mctx.TraceTimed(fmt.Sprintf("calculateSummaryAtMerkleSeqno(%s, %v)", team.ID, merkleSeqno), func() error { return err })()
+
+	if merkleSeqno == nil {
+		tmp, err := merkleSeqnoAtGenerationInception(mctx, team.chain())
+		if err != nil {
+			return nil, err
+		}
+		merkleSeqno = &tmp
+	}
```
zapu: so instead of doing this here, change `merkleSeqno` to a value type, and do this calculation in `calculateChainSummary`

if this is the only call site, of course, maybe i missed something

PR #16378, opened 2019-03-04
fix no convos inbox on mobile

buoyad:
check if we're loaded instead of `neverLoaded` because that's only set on `metasReceived`, which doesn't happen here. r? @keybase/react-hackers


buoyad:
nvm there's actually a bug in the reducer

PR #16379, opened 2019-03-04
Profile add-to-team no longer unselects checkboxes

mlsteele:
Before this patch clicking "Add to 2 teams" to add a user to teams from their profile deselected all the checkboxes for the duration of the spinner. With this patch the boxes stay selected. It doesn't seem to exhibit the opposite bug of keeping teams selected across sessions.


mlsteele:
cjb points out this was in to fix a bug where one team errors out the checkboxes didn't clear

PR #16381, opened 2019-03-04
fix setting loaded flags in metasReceived reducer

buoyad:
Fixes the no-conversations inbox view on mobile and desktop. @keybase/react-hackers

### DIFF
--- a/shared/reducers/chat2.js
+++ b/shared/reducers/chat2.js
@@ -955,12 +955,12 @@ const rootReducer = (
     case Chat2Gen.staticConfigLoaded:
       return state.set('staticConfig', action.payload.staticConfig)
     case Chat2Gen.metasReceived: {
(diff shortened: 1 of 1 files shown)


shared/reducers/chat2.js
```diff
@@ -956,7 +956,7 @@ const rootReducer = (
       return state.set('staticConfig', action.payload.staticConfig)
     case Chat2Gen.metasReceived: {
       let nextState = action.payload.fromInboxRefresh ? state.set('inboxHasLoaded', true) : state
-      nextState = action.payload.initialTrustedLoad ? state.set('trustedInboxHasLoaded', true) : state
```
chrisnojima: maybe we should just move that into the withMutations (or just make it a merge) instead of this chaining

buoyad:
@chrisnojima changed to use merge

PR #16391, opened 2019-03-04
keybase_service_base: pass offline availability to `Resolve3`

strib:
This pass `BEST_EFFORT` offline availability mode to `Resolve3`, whenever the TLF path or ID in question is being synced.  This tells the service that it is ok to serve a stale answer from the cache, if the API server can't be reached in a timely fashion over the internet.

Issue: KBFS-3918

### DIFF
--- a/go/kbfs/libkbfs/tlf_handle_test.go
+++ b/go/kbfs/libkbfs/tlf_handle_test.go
@@ -5,6 +5,7 @@
 package libkbfs
 
 import (
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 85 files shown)


strib:
The amount of plumbing here is ridiculous.  Here are the key files to look at, roughly in order:

* `interfaces.go`
* `config_local.go`
* `keybase_service_base.go`
* `tlf_handle_resolve.go`
* `tlf_handle_test.go`

go/kbfs/libkbfs/tlf_handle_test.go
```diff
@@ -1269,3 +1277,168 @@ func TestParseTlfHandleImplicitTeams(t *testing.T) {
...
+		bestEffortOfflineCounts: make(map[string]int),
+	}
+
+	osg := &testOfflineStatusPathsGetter{make(map[string]bool)}
+	osg.bestEffortPaths["/keybase/private/u2"] = true
+
+	t.Log("Check unsynced private TLF")
+	_, err := ParseTlfHandle(ctx, kbpki, nil, osg, "u1", tlf.Private)
+	require.NoError(t, err)
+	require.Equal(t, kbpki.bestEffortOfflineCounts["u1"], 0)
```
jzila: I didn't quite dig into the call stack here, but is it possible that these counts would be updated racily, such that they wouldn't be ready yet at this point?

strib: No that shouldn't happen.  `ParseTlfHandle` waits for all goroutines doing the resolves to complete.  But let me know if you had a specific scenario in mind that I'm missing.

PR #16403, opened 2019-03-05
resolveIdentifyImplicitTeam via offline cache

pzduniak:


go/service/identify.go
```diff
@@ -175,16 +175,25 @@ func (h *IdentifyHandler) resolveUserOrTeam(ctx context.Context, arg string) (u
...
 	writerAssertions, readerAssertions, err := externals.ParseAssertionsWithReaders(h.G(), arg.Assertions)
 	if err != nil {
 		return res, err
 	}
-	return h.resolveIdentifyImplicitTeamHelper(ctx, arg, writerAssertions, readerAssertions)
+
+	resp := &res
+	err = h.service.offlineRPCCache.Serve(mctx, arg.Oa, offline.Version(1), "identify.resolveIdentifyImplicitTeam", false, arg, resp, func(mctx libkb.MetaContext) (interface{}, error) {
+		tmp, err := h.resolveIdentifyImplicitTeamHelper(mctx.Ctx(), arg, writerAssertions, readerAssertions)
+		if tmp.DisplayName != "" {
```
maxtaco: Are there cases where we return non-nil error, and a valid result?  Usually we'd only do this if `err != nil`

go/service/identify.go
```diff
@@ -175,16 +175,25 @@ func (h *IdentifyHandler) resolveUserOrTeam(ctx context.Context, arg string) (u
...
+	mctx.Debug("ResolveIdentifyImplicitTeam assertions:'%v'", arg.Assertions)
 
 	writerAssertions, readerAssertions, err := externals.ParseAssertionsWithReaders(h.G(), arg.Assertions)
 	if err != nil {
 		return res, err
 	}
-	return h.resolveIdentifyImplicitTeamHelper(ctx, arg, writerAssertions, readerAssertions)
+
+	resp := &res
+	err = h.service.offlineRPCCache.Serve(mctx, arg.Oa, offline.Version(1), "identify.resolveIdentifyImplicitTeam", false, arg, resp, func(mctx libkb.MetaContext) (interface{}, error) {
```
maxtaco: for the `arg` cache key, zero out all fields except for `assertions`, `suffix` and `isPublic`

go/service/identify.go
pzduniak: `TestResolveIdentifyImplicitTeamWithIdentifyFailures` [fails without this](https://ci.keyba.se/job/client/job/PR-16403/2/). IIRC it was a resolve success with an assertion fail.

PR #16405, opened 2019-03-05
Fix C c accepting keybase prove warnings

pzduniak:
eof-like stuff results in `(!ok && err != nil) == true`


pzduniak:
not waiting for a review since it's _very_ simple

PR #16407, opened 2019-03-05
Warn user if logdir is unwritable

heronhaye:
I'm not sure about how to word this error message. There's been a few reports about `sudo run_keybase` creating unwritable directories but I can't figure out how it would've happened, When I run `sudo run_keybase`, it makes `/root/.cache/keybase` (and then fails when the binary says it needs `KEYBASE_ALLOW_ROOT=1` and not `/home/me/.cache/keybase` owned by my root. Regardless, if it does happen, they should get this message.

### DIFF
--- a/packaging/linux/run_keybase
+++ b/packaging/linux/run_keybase
@@ -141,12 +141,20 @@ show_cryptosquirrel() {
 }
 
 init() {
-  # Cannot do in go due to background processes being piped to log in bash
   logdir="${XDG_CACHE_HOME:-$HOME/.cache}/keybase"
+  runtime_dir="${XDG_RUNTIME_DIR:-$HOME/.config}/keybase"
+
+  if ! [ -w "$logdir" ]; then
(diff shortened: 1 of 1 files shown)


packaging/linux/run_keybase
```diff
@@ -140,13 +140,27 @@ show_cryptosquirrel() {
   [ "${KEYBASE_NO_SQUIRREL:-}" != "1" ] && cat /opt/keybase/crypto_squirrel.txt
 }
 
+warn_if_unwritable() {
+  if [ -w "$1" ]; then
+      return
+  fi
+  echo "WARNING: Cannot write to $1. Did you run 'run_keybase' with sudo?"
```
strib: "Did you run `run_keybase`" -> "Did you previously run 'run_keybase'"

heronhaye:
```suggestion
  echo "WARNING: Cannot write to $1. Did you previously run 'run_keybase' with sudo?"
```

PR #16412, opened 2019-03-05
bookkeep top reactions locally

joshblum:
keep a count of reactions used locally in an encrypted level db and expose a simple rpc to serve back the top 5 reactions (can make the size configurable) for the gui. 

cc @chrisnojima


go/chat/globals/globals.go
```diff
@@ -13,6 +13,7 @@ type ChatContext struct {
 	ServerCacheVersions types.ServerCacheVersions // server side versions for chat caches
 	RegexpSearcher      types.RegexpSearcher      // For searching chat messages in a conversation via regexp
 	Indexer             types.Indexer             // For searching chat messages in the entire inbox
+	ReactjiStore        types.ReactjiStore        // Stores the user's favorite reactions on this device
```
mmaxim: *Reacji

joshblum: sigh the ticket title is misspelled

mmaxim: Why is this so high up anyway, can it just be a thing in `storage`?

joshblum: frontend is going to be calling it often, i wanted the state in memory

mmaxim: How often? I hope it isn't any time that reacji bar is rendered cc @chrisnojima

joshblum: he wanted to call it every time a reaction is posted in case the ordering changed. was going to potentially refactor this to put the update on the notification we already plumb up to the frontend on a reaction post but wanted to discuss

mmaxim: What is the plan for initially getting these into the JS state when clicking into a conversation? Is it going to be part of the `InboxUIItem` that gets sent up, or are they going to call that RPC? Calling the RPC is a problem, since it will be async with loading the thread.

joshblum: i assumed it would be called once on startup/login since it's global per user and then updated when posting a  reaction

mmaxim: In that case, you can make it part of the chat "static config" like the slash commands are. There is already an RPC for that.

mmaxim: Actually maybe not, since that RPC is called before login I think.

joshblum:
@mmaxim this is ready for another pass, addressed the issues we talked about

@chrisnojima the `TopReacjis` are now plumbed through on `REACTION_UPDATE` notifications and part of the `GetBootstrapStatus` init call

joshblum:
merging over test flake

PR #16415, opened 2019-03-05
Add a link to the blog post.

adamjspooner:

### DIFF
--- a/shared/chat/conversation/messages/coinflip/container.js
+++ b/shared/chat/conversation/messages/coinflip/container.js
@@ -36,9 +36,6 @@ const mapStateToProps = (state, {flipGameID}: OwnProps) => {
 export default namedConnect<OwnProps, _, _, _, _>(
   mapStateToProps,
   (d, o) => ({}),
-  (s, d, o) => ({
-    ...s,
-    howThisWorksURL: 'https://keybase.io/coin-flip',
-  }),
+  (s, d, o) => ({...s}),
   'CoinFlip'
 )(CoinFlip)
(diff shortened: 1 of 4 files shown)


shared/chat/conversation/messages/coinflip/container.js
```diff
@@ -36,6 +36,9 @@ const mapStateToProps = (state, {flipGameID}: OwnProps) => {
 export default namedConnect<OwnProps, _, _, _, _>(
   mapStateToProps,
   (d, o) => ({}),
-  (s, d, o) => ({...s}),
+  (s, d, o) => ({
+    ...s,
+    howThisWorksURL: 'https://keybase.io/coin-flip',
+  }),
```
adamjspooner: Is there any reason to do it this way (instead of hard-coding it on the `Text` element?

shared/chat/conversation/messages/coinflip/container.js
```diff
@@ -36,6 +36,9 @@ const mapStateToProps = (state, {flipGameID}: OwnProps) => {
 export default namedConnect<OwnProps, _, _, _, _>(
   mapStateToProps,
   (d, o) => ({}),
-  (s, d, o) => ({...s}),
+  (s, d, o) => ({
+    ...s,
+    howThisWorksURL: 'https://keybase.io/coin-flip',
```
chrisnojima: i don't think this needs to be in the container, just move it into the components and itd remove a bunch of code

adamjspooner: 👌🏻

PR #16416, opened 2019-03-05
make inject robust to games with gaps in current played state

mmaxim:
cc @malgorithms @chrisnojima


mmaxim:
tested locally so ignoring CI

PR #16417, opened 2019-03-05
Center username and timestamp when there's no bottom line on small teams in the inbox.

adamjspooner:


shared/chat/inbox/container/remote.js
```diff
@@ -90,6 +90,7 @@ export const serialize = ({
     channelname: conversation.channelname,
     conversationIDKey: conversation.conversationIDKey,
     hasBadge,
+    hasBottomLine: true,
```
chrisnojima: you checked that this worked in teh widget right

adamjspooner: Yup!

PR #16419, opened 2019-03-05
delay error report from replay contingent on a status showing up during delay

mmaxim:
If we get an error from replay, then wait 5 seconds before reporting the error, and abort any report if any other status has been reported in the meantime.

cc @malgorithms @chrisnojima


mmaxim:
testing locally so ignoring CI

PR #16422, opened 2019-03-05
more custom displays for flip errors

mmaxim:
cc @adamjspooner @malgorithms

### DIFF
--- a/shared/chat/conversation/messages/coinflip/errors.js
+++ b/shared/chat/conversation/messages/coinflip/errors.js
@@ -103,7 +103,7 @@ const CoinFlipDupError = (props: DupProps) => {
   return (
     <Kb.Box2 direction="vertical" fullWidth={true} gap="tiny">
       <Kb.Text selectable={true} type="Body">
-        Duplicate {`${props.desc}`} received from the following participant:
+        Duplicate {props.desc} received from the following participant:
       </Kb.Text>
       <Kb.Text selectable={true} type="BodySemibold">
         {props.offender.user} (device: {props.offender.device}
--- a/shared/chat/conversation/messages/coinflip/errors.js
+++ b/shared/chat/conversation/messages/coinflip/errors.js
@@ -101,7 +101,7 @@ type DupProps = {|
 
 const CoinFlipDupError = (props: DupProps) => {
   return (
-    <Kb.Box2 direction="vertical" gap="tiny">
+    <Kb.Box2 direction="vertical" fullWidth={true} gap="tiny">
       <Kb.Text selectable={true} type="Body">
         Duplicate {`${props.desc}`} received from the following participant:
       </Kb.Text>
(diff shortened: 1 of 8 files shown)


shared/chat/conversation/messages/coinflip/errors.js
```diff
@@ -13,7 +12,26 @@ const CoinFlipError = (props: Props) => {
     return <CoinFlipGenericError error={props.error.generic} />
   } else if (props.error.typ === RPCChatTypes.chatUiUICoinFlipErrorTyp.absentee && props.error.absentee) {
     return <CoinFlipAbsenteeError error={props.error.absentee} />
+  } else if (props.error.typ === RPCChatTypes.chatUiUICoinFlipErrorTyp.timeout) {
+    return <CoinFlipTimeoutError />
+  } else if (props.error.typ === RPCChatTypes.chatUiUICoinFlipErrorTyp.aborted) {
+    return <CoinFlipAbortedError />
+  } else if (props.error.typ === RPCChatTypes.chatUiUICoinFlipErrorTyp.dupreg && props.error.dupreg) {
+    return <CoinFlipDupError offender={props.error.dupreg} desc={'registration'} />
```
chrisnojima: by convention we pass strings directly to jsx elements aka
```
desc="registration"
```

shared/chat/conversation/messages/coinflip/errors.js
```diff
@@ -58,10 +78,55 @@ const CoinFlipAbsenteeError = (props: AbsenteeProps) => {
   )
 }
 
-const styles = Styles.styleSheetCreate({
-  error: {
-    color: Styles.globalColors.red,
-  },
-})
+const CoinFlipTimeoutError = () => {
```
chrisnojima: when just returning we prefer
```
const CoinFlipTimeoutError = () => <Kb.Text ...
```

shared/chat/conversation/messages/coinflip/errors.js
```diff
@@ -58,10 +78,55 @@ const CoinFlipAbsenteeError = (props: AbsenteeProps) => {
...
+    <Kb.Text selectable={true} type="Body">
+      Flip timed out before a result was obtained.
+    </Kb.Text>
+  )
+}
+
+const CoinFlipAbortedError = () => {
+  return (
+    <Kb.Text selectable={true} type="Body">
+      Flip aborted before a result was obtained.
```
adamjspooner: Is this caused by the flip's initiator?

shared/chat/conversation/messages/coinflip/errors.js
mmaxim: Why? It's just more annoying if you need to change it later, and doesn't really make the code easier to read or anything else.

shared/chat/conversation/messages/coinflip/errors.js
mmaxim: It just means the game ended before we got a commitment complete, so I guess you could pin it on him.

shared/chat/conversation/messages/coinflip/errors.js
chrisnojima: its a style choice thats applied everywhere. it does reduce how much noise there is but ultimately its a stylistic choice. what I'd like to avoid is having code that is against the style, which i think is worse than one style or the other

PR #16425, opened 2019-03-05
keybase_service_base: plumb offline status for LoadTeamPlusKeys

strib:
This affects all the paths for getting keys, checking team membership, and finding a team's root team ID.

Issue: KBFS-3918


strib:
This one is basically pure plumbing.  There's not really any particular file that needs a deep review.

PR #16429, opened 2019-03-05
kbfs: enable v2 encryption for blocks by default

strib:
And use in all tests.

Note that older clients won't be able to evict new v2 blocks from the disk block cache, or delete them via mark-and-sweep, because the old max block hash was still based on v1-hashes.

Issue: KBFS-3461

### DIFF
--- a/go/kbfs/kbfshash/hash.go
+++ b/go/kbfs/kbfshash/hash.go
@@ -55,6 +55,9 @@ const (
 	SHA256Hash HashType = 1
 	// SHA256HashV2 is the type of a SHA256 hash over V2-encrypted data.
 	SHA256HashV2 HashType = 2
+
+	// MaxHashType is the numerically-largest hash type.
+	MaxHashType HashType = SHA256HashV2
 )
 
 // MaxDefaultHash is the maximum value of RawDefaultHash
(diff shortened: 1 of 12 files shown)


go/kbfs/kbfshash/hash.go
```diff
@@ -55,6 +55,9 @@ const (
 	SHA256Hash HashType = 1
 	// SHA256HashV2 is the type of a SHA256 hash over V2-encrypted data.
 	SHA256HashV2 HashType = 2
+
+	// MaxHashType is the numerically-largest hash type.
+	MaxHashType HashType = SHA256HashV2
```
songgao: Is it the the largest as in all bytes in the hash are 0xff like `MaxDefaultHash`, or is it the largest supported hash version? The comment sounds like former, but making it same value as V2 feels like latter.

songgao: Oh maybe it's the hash type of MaxDefaultHash?

strib: It's the maximum defined numeric value of `HashType`.  I tried to be clear about that in the comment -- how can I improve it?

`MaxDefaultHash` is the actual maximum hash value itself, excluding the type (which needs to be prepended).

songgao: Oh I see; how about "highest supported hash version"?

PR #16431, opened 2019-03-05
logger: use bufio.NewWriter to buffer logs

jakob223:
On master:
```
rm -rf new_rmrf3/deep  0.07s user 1.09s system 2% cpu 44.075 total
```
On this branch:
```
rm -rf new_rmrf1/deep  0.05s user 0.86s system 3% cpu 23.531 total
```

### DIFF
--- a/go/logger/buffer.go
+++ b/go/logger/buffer.go
@@ -1,0 +1,115 @@
+// Copyright 2019 Keybase, Inc. All rights reserved. Use of
+// this source code is governed by the included BSD license.
+
+package logger
+
+import (
+	"bufio"
+	"io"
+	"sync"
+	"time"
+)
+
+type triggerableTimer struct {
+	C          chan struct{}
+	timer      *time.Timer
+	sentinelCh chan struct{}
+	shutdownCh chan struct{}
+}
+
+func newTriggerableTimer(d time.Duration) *triggerableTimer {
+	t := &triggerableTimer{
+		C:          make(chan struct{}, 1),
+		sentinelCh: make(chan struct{}, 1),
+		timer:      time.NewTimer(0),
+		shutdownCh: make(chan struct{}),
+	}
+	go func() {
+		for {
+			select {
+			case <-t.timer.C:
+				<-t.sentinelCh
+				t.C <- struct{}{}
+			case <-t.shutdownCh:
+				t.timer.Stop()
+				return
+			}
+		}
+	}()
+	return t
+}
+
+func (t *triggerableTimer) ResetIfStopped(d time.Duration) {
+	if d == 0 {
+		return
+	}
+	select {
+	case t.sentinelCh <- struct{}{}:
+		t.timer.Reset(d)
+	default:
+	}
+}
+
+func (t *triggerableTimer) Close() {
+	close(t.shutdownCh)
+}
+
+type autoFlushingBufferedWriter struct {
+	lock           sync.RWMutex
+	bufferedWriter *bufio.Writer
+	backupWriter   *bufio.Writer
+
+	frequency time.Duration
+	timer     *triggerableTimer
+	shutdown  chan struct{}
+}
+
+var _ io.Writer = &autoFlushingBufferedWriter{}
+
+func (writer *autoFlushingBufferedWriter) backgroundFlush() {
+	for {
+		select {
+		case <-writer.timer.C:
+			// Swap out active and backup writers
+			writer.lock.Lock()
+			writer.bufferedWriter, writer.backupWriter = writer.
+				backupWriter, writer.bufferedWriter
+			writer.lock.Unlock()
+
+			writer.backupWriter.Flush()
+		case <-writer.shutdown:
+			writer.timer.shutdownCh <- struct{}{}
+			writer.bufferedWriter.Flush()
+			return
+		}
+	}
+}
+
+// NewAutoFlushingBufferedWriter returns an io.Writer that buffers its output
+// and flushes automatically after `flushFrequency`.
+func NewAutoFlushingBufferedWriter(baseWriter io.Writer,
+	flushFrequency time.Duration) (w io.Writer, shutdown chan struct{}) {
+	result := &autoFlushingBufferedWriter{
+		bufferedWriter: bufio.NewWriter(baseWriter),
+		backupWriter:   bufio.NewWriter(baseWriter),
+		frequency:      flushFrequency,
+		timer:          newTriggerableTimer(flushFrequency),
+	}
+	go result.backgroundFlush()
+	return result, result.shutdown
+}
+
+func (writer *autoFlushingBufferedWriter) Write(p []byte) (int, error) {
+	// The locked resource here, the pointer bufferedWriter, is only being read
+	// even though this function is Write.
+	writer.lock.RLock()
+	defer writer.lock.RUnlock()
+
+	n, err := writer.bufferedWriter.Write(p)
+	if err != nil {
+		return n, err
+	}
+	writer.timer.ResetIfStopped(writer.frequency)
+
+	return n, nil
+}
--- a/go/logger/global.go
+++ b/go/logger/global.go
@@ -3,16 +3,23 @@ package logger
 import (
 	"os"
 	"sync"
+	"time"
 
-	logging "github.com/keybase/go-logging"
-	isatty "github.com/mattn/go-isatty"
+	"github.com/keybase/go-logging"
+	"github.com/mattn/go-isatty"
 )
+
+const logMS = 10
+const loggingFrequency = logMS * time.Millisecond
 
 var globalLock sync.Mutex
 var stderrIsTerminal = isatty.IsTerminal(os.Stderr.Fd())
 var currentLogFileWriter *LogFileWriter
+var stdErrLoggingShutdown chan<- struct{}
 
 func init() {
-	logBackend := logging.NewLogBackend(ErrorWriter(), "", 0)
+	writer, shutdown := NewAutoFlushingBufferedWriter(ErrorWriter(), loggingFrequency)
+	stdErrLoggingShutdown = shutdown
+	logBackend := logging.NewLogBackend(writer, "", 0)
 	logging.SetBackend(logBackend)
 }
--- a/go/logger/file.go
+++ b/go/logger/file.go
@@ -47,6 +47,12 @@ func SetLogFileConfig(lfc *LogFileConfig) error {
 		w.config = *lfc
 	} else {
 		w = NewLogFileWriter(*lfc)
+
+		// Clean up the default logger, if it is in use
+		select {
+		case stdErrLoggingShutdown <- struct{}{}:
+		default:
+		}
 	}
 
 	if err := w.Open(time.Now()); err != nil {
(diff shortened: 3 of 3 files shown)


go/logger/global.go
```diff
@@ -12,7 +15,21 @@ var globalLock sync.Mutex
...
 
+func getBufferedErrorWriter() io.Writer {
+	writer := ErrorWriter()
+	buf := bufio.NewWriter(writer)
+	go func() {
+		for range time.Tick(300 * time.Millisecond) {
+			// TODO: do we care if this errors
+			buf.Flush()
+		}
+	}()
```
jzila: This goroutine will leak. You'll need to store a ticker and `Stop()` it when this writer needs to be shutdown.

patrickxb: I agree that using `time.NewTicker()` is a better idea than `time.Tick`.

patrickxb: Also a good idea to call `Flush()` when shutdown happens.

patrickxb:
Don't you also need to change `logger/file.go` where it calls NewLogBackend?

patrickxb:
(whoops, pressed wrong button...sorry about that)

jakob223:
One notable aspect of this change is that in the case that another thing panics, we don't have the last 300ms of logs. I could change it down to 50ms probably without reducing the performance, but I don't have any good ideas on what to do to try to make it flush in the event that something else panics.

One thing we might do is flush immediately on warning. But that's definitely more complicated than what's going on here now.

jzila:
We could add a `recover` to flush the logger and then re-throw.

jakob223:
Er, this is a problem if *another goroutine* panics. Very likely any panics that happen will not be in the logging goroutine but in another.

jzila:
Oh in that case you can `defer Flush` inside the logging goroutine.

go/logger/global.go
```diff
@@ -11,8 +14,30 @@ import (
 var globalLock sync.Mutex
 var stderrIsTerminal = isatty.IsTerminal(os.Stderr.Fd())
 var currentLogFileWriter *LogFileWriter
+var stdErrLoggingShutdown chan<- struct{}
+
+func bufferLogs(writer io.Writer) (io.Writer, chan<- struct{}) {
+	buf := bufio.NewWriter(writer)
```
jzila: Add a `defer buf.Flush()` here.

go/logger/global.go
```diff
@@ -11,8 +14,30 @@ import (
...
+	shutdown := make(chan struct{})
+	go func() {
```
jzila: Add a `defer buf.Flush()` here.

jakob223: if we put the `defer Flush()` in the goroutine here, it will never execute. See https://github.com/golang/go/issues/14542

jakob223: or https://play.golang.org/p/o31M7ElNK0Q which is how I convinced myself I was reading that issue correctly.

jzila: TIL. I think we've been working under the opposite assumption to date.

jakob223:
This version works in theory but not in practice. There is a bug in my concurrency reasoning somewhere, or else in my understanding of `time.Timer`. After a short while, it reaches a race condition and stops logging.

go/logger/buffer.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+var _ io.Writer = &autoFlushingBufferedWriter{}
+
+func (writer *autoFlushingBufferedWriter) backgroundFlush() {
+	for {
+		select {
+		case <-writer.timer.C:
+			// TODO: I claim this is concurrency safe, but how do I tell the race
+			//  checker that?
```
jzila: It's not concurrency safe because writes can come in while you're flushing.

jzila: Oh nvmind, you're talking about the timer itself.

jzila: https://golang.org/pkg/time/#Timer.Reset

> This should not be done concurrent to other receives from the Timer's channel.

> Note that it is not possible to use Reset's return value correctly, as there is a race condition between draining the channel and the new timer expiring. Reset should always be invoked on stopped or expired channels, as described above. The return value exists to preserve compatibility with existing programs.

go/logger/buffer.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (writer *autoFlushingBufferedWriter) backgroundFlush() {
+	for {
+		select {
+		case <-writer.timer.C:
+			// Swap out active and backup writers
+			writer.lock.Lock()
+			writer.currentlyWaiting = false
+			previouslyActive := writer.bufferedWriter
+			writer.bufferedWriter = writer.backupWriter
```
jzila: You can do `writer.bufferedWriter, writer.backupWriter = writer.backupWriter, writer.bufferedWriter`.

go/logger/buffer.go
```diff
@@ -0,0 +1,88 @@
...
+			writer.lock.Unlock()
+
+			// Flush the previously active writer and complete the swap. We do
+			// not need to complete the swap under lock because this is the
+			// only goroutine accessing backupWriter.
```
jzila: Technically true, but IMO following my suggestion above results in better code clarity.

go/logger/buffer.go
```diff
@@ -0,0 +1,88 @@
...
+		backupWriter:   bufio.NewWriter(baseWriter),
+		frequency:      flushFrequency,
+		timer:          time.NewTimer(flushFrequency),
+	}
+	go result.backgroundFlush()
+	return result, result.shutdown
+}
+
+func (writer *autoFlushingBufferedWriter) Write(p []byte) (int, error) {
+	writer.lock.RLock()
```
jzila: Let's add a comment explaining that we're using an `RLock` on a "write" because the locked resource is only experiencing a read here.

jakob223:
I've added some locking and re-measured performance with `logMS` at `10`, `20`, and `200`. Each of them has logging consuming `5%` or so of the CPU time consumed by `cp` and `rm` on large directory structures; the overall KBFS process takes up almost 2 CPUs, so this is around `9%` of a single CPU.

I think this is close to the best we can get using a strategy similar to this one / without investing significantly more time. Of the logging, time is spent roughly in a 3:3:2:2 ratio between string formatting, file I/O, goroutine and mutex overhead, and everything else.

go/logger/buffer.go
```diff
@@ -0,0 +1,88 @@
...
+	// that the timer is stopped when it has not yet fired, or it fires and
+	// then is reset. In the latter case, we get slightly more frequent
+	// flushing than we need; in the former, we get a slight delay. This race
+	// is unlikely, so it's exceedingly unlikely that the delay will accumulate
+	// over more than one successive race.
+	writer.timer.Reset(writer.frequency)
+	// Any race over `currentlyWaiting` will have minimal impact. The only race
+	// possible given the locking is between two copies of the below line
+	// both trying to set the value to `true` simultaneously.
+	writer.currentlyWaiting = true
```
jzila: I think this whole variable can be replaced with calls to `writer.Buffered()`.

go/logger/buffer.go
```diff
@@ -0,0 +1,88 @@
...
+		return n, nil
+	}
+	// Docs say this should not be called concurrently with channel receive.
+	// I've read the internal code and the only 2 ways it could resolve are
+	// that the timer is stopped when it has not yet fired, or it fires and
+	// then is reset. In the latter case, we get slightly more frequent
+	// flushing than we need; in the former, we get a slight delay. This race
+	// is unlikely, so it's exceedingly unlikely that the delay will accumulate
+	// over more than one successive race.
+	writer.timer.Reset(writer.frequency)
```
jzila: I'd rather not leave races lying around by design. Especially when we can avoid them.

You can make a timer abstraction that is goroutine-safe, and keep all the timing state in it. It would be able to do only the things we need here: `ResetIfStopped`. Then you don't need to track `currentlyWaiting` at all.

Here's what I'm thinking: https://play.golang.org/p/hGRbggkPDBE
```go
type TriggerableTimer struct {
	C          chan struct{}
	timer      *time.Timer
	sentinelCh chan struct{}
	shutdownCh chan struct{}
}

func NewTriggerableTimer() *TriggerableTimer {
	t := &TriggerableTimer{
		C:          make(chan struct{}, 1),
		sentinelCh: make(chan struct{}, 1),
		timer:      time.NewTimer(0),
		shutdownCh: make(chan struct{}),
	}
	go func() {
		for {
			select {
			case <-t.timer.C:
				<-t.sentinelCh
				t.C <- struct{}{}
			case <-t.shutdownCh:
				t.timer.Stop()
				return
			}
		}
	}()
	return t
}

func (t *TriggerableTimer) ResetIfStopped(d time.Duration) {
	if d == 0 {
		return
	}
	select {
	case t.sentinelCh <- struct{}{}:
		t.timer.Reset(d)
	default:
	}
}

func (t *TriggerableTimer) Close() {
	close(t.shutdownCh)
}
```

go/logger/buffer.go
jzila: Actually with my feedback below, this is no longer relevant.

jakob223:
I ran some benchmarks on this code - overhead in the new buffering/timer code and the rest of logging now each take about 2-3% on large copy operations and 4-5% on large remove operations.

go/logger/buffer.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+type triggerableTimer struct {
+	C          chan struct{}
+	timer      *time.Timer
+	sentinelCh chan struct{}
+	shutdownCh chan struct{}
+}
+
+func newTriggerableTimer(d time.Duration) *triggerableTimer {
```
jzila: I think we should trigger this with no duration, otherwise we'll block in the `select` below, on `<-t.sentinelCh`, until a `ResetIfStopped` comes in. Then it'll fire right away.

maxtaco:
@jakob223 / @jzila  this PR has really wide-ranging implications and has the ability to deadlock the service or cause it to log-amplify, can we get a chance to review it too before it goes in?

patrickxb:
Can you please revert this?

When this project started, I said it should get approval from a few people on CORE.  

Logging changes are dangerous...this looks more complicated than the last time I looked at it.

maxtaco:
Also, was it resolved above that we'll still get panics in the logs? I hope so, those are extremely valuable

maxtaco:
I don't think we need to revert it necessarily, unless we see there are changes needed

jzila:
Panics will still appear in the logs. We might just be missing up to ~10ms before the panic.

patrickxb:
Could we make this buffered logging optional?

To give us flexibility.  Desktop KBFS could use it, service could use the existing system.  Maybe mobile uses existing too since KBFS shouldn't be logging heavily on mobile.

jzila:
We could, although our measurement of the impact assumed all logging was buffered. I think it's worth testing this for all logging, but I'll defer to the core folks on when to run such a test. @jakob223 please make a ticket to make buffered logging optional, and switch the non-kbfs loggers to unbuffered.

maxtaco:
I mentioned in chat that I think this is a good idea that we should try on the service too.  But we would like a chance to review it. And we should think really hard about if we should release it in the release.

patrickxb:
copying from chat:

i would like to again vote for a revert on the logging PR.  it looks fine, but i think making it optional is important.  and i'd love more CORE eyes on it.  i saw a little weirdness in the output when running a test just now.

seems a lot safer to have it as an option rather than replacing all logging with this.
also would allow admin builds to have it on as it burns in.
the code looks fine, btw, i just would prefer a more gradual introduction of it.

jakob223:
https://github.com/keybase/client/pull/16664 reverts these changes

PR #16437, opened 2019-03-06
WIP: preload some system dlls

zanderz:
Found one more place that needed safer loading


taruti:
Having tests for this would be kind of nice, but that can be a separate PR. E.g. have a DLL that crashes on load and copy it to the same dir as the exe, does the exe run or crash? The DLL loading bit looks good, but for the install dir I'm somewhat conflicted.

zanderz:
so is this approved for merging then @taruti ?

PR #16438, opened 2019-03-06
more riit debug

maxtaco:


maxtaco:
cc @pzduniak there was a superfluous debug in there

PR #16439, opened 2019-03-06
remove listener on errors during startflip

mmaxim:
If it stays, it can deadlock `NotifyRouter`.

cc @maxtaco


mmaxim:
tested locally

PR #16445, opened 2019-03-06
Do not fail entire signup if gpg fails

zapu:
We should not fail entire signup engine if GPG importing fails, because it's too late when this happens. Failing there leaves GUI in bad state where user is logged in but GUI doesn't know, and also some other stuff like `CreateWalletSoft` does not get a chance to run (but that's ok - it'd run again in future).

This bug was unfortunate because it affected both GUI and CLI, where GUI does not care about GPG importing at all - it attaches an UI which will always respond "no" to `WantToAddGPGKey` prompt. But this error happens before this prompt is even reached, failing signup process that has already succeeded.

### DIFF
--- a/go/engine/signup.go
+++ b/go/engine/signup.go
@@ -116,8 +116,13 @@ func (s *SignupEngine) Run(m libkb.MetaContext) (err error) {
 		}
 	}
 
-	if err = s.doGPG(m); err != nil {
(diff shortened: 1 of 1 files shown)


zapu:
I've rebased this morning and this chat test is still failing, so I'm merging over it.

PR #16449, opened 2019-03-06
Use locktable in chat storage

joshblum:
use `libkb.LockTable` for message storage to reduce lock contention

### DIFF
--- a/go/chat/storage/version.go
+++ b/go/chat/storage/version.go
@@ -18,7 +18,6 @@ type ServerVersions struct {
 }
 
 func NewServerVersions(g *globals.Context) *ServerVersions {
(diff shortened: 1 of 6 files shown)


go/chat/storage/version.go
```diff
@@ -18,6 +18,7 @@ type ServerVersions struct {
 }
 
 func NewServerVersions(g *globals.Context) *ServerVersions {
+	initLocksRepoOnce(g)
```
mmaxim: Can you do this instead?

https://golang.org/doc/effective_go.html#init

joshblum: fixed

PR #16456, opened 2019-03-06
Add title and collapse action to mobile attachments.

adamjspooner:


chrisnojima:
we spoke irl this is changing a bunch

shared/common-adapters/icon.desktop.js
```diff
@@ -72,7 +72,7 @@ class Icon extends Component<Props, void> {
         }
       : null
 
-    const hasContainer = (this.props.onClick && this.props.style) || isFontIcon
+    const hasContainer = (onClick && this.props.style) || isFontIcon
```
chrisnojima: this and below can likely be the same as in master
this should also respect the noClickableBox prop

adamjspooner:
@chrisnojima, how's this?

PR #16457, opened 2019-03-06
Fix the team banner not dismissing

cjb:
@keybase/react-hackers 

1: Move the banner outside of the memoized list, so that it re-renders itself on prop changes.  This was the bug causing the banner to not dismiss until you leave and re-enter the view.

2: When it's dismissed, as well as firing the Gregor action to update Gregor state async, also just update the component's local state so that we know to dismiss it immediately -- using `this.state.sawChatBanner || this.props.sawChatBanner` to decide to not show it.


shared/teams/main/index.js
```diff
@@ -135,6 +136,9 @@ class Teams extends React.PureComponent<Props> {
           onCreateTeam={this.props.onCreateTeam}
           onJoinTeam={this.props.onJoinTeam}
         />
+        {this.state.sawChatBanner || this.props.sawChatBanner ? null : (
```
chrisnojima: i think this should just use this.state. we can add a componentDidUpdate and if the prop value changed it can update the state

shared/teams/main/index.js
```diff
@@ -83,25 +83,26 @@ export const TeamRow = React.memo<RowProps>((props: RowProps) => {
...
 })
 
-class Teams extends React.PureComponent<Props> {
+type State = {
+  sawChatBanner: boolean,
+}
+class Teams extends React.PureComponent<Props, State> {
+  state = {sawChatBanner: false}
   _teamsAndExtras = memoize(teamnames => {
-    return [
```
chrisnojima: this can still be in here it just should be passed to memoize so it updates

cjb:
@chrisnojima Made these changes, ready for another look.

PR #16462, opened 2019-03-06
another list2 bug

songgao:
If `itemData` is not passed in, items don't get re-rendered when their props change.


chrisnojima:
Let's discuss. We might not want this

PR #16478, opened 2019-03-07
Add IncludeMulti to Details protocol, set it true

patrickxb:
This has the protocol change and sets include_multi to true in the api call for new clients.


patrickxb:
i give up on ci...

PR #16479, opened 2019-03-07
Add an RPC to check passphrase

zapu:


go/service/account.go
```diff
@@ -164,3 +164,19 @@ func (h *AccountHandler) SetLockdownMode(ctx context.Context, arg keybase1.SetLo
...
+
+func (h *AccountHandler) TestPassphrase(ctx context.Context, sessionID int) (err error) {
+	mctx := libkb.NewMetaContext(ctx, h.G())
+	defer mctx.Trace("TestPassphrase", func() error { return err })()
+
+	username := h.G().GetEnv().GetUsername().String()
+
+	arg := libkb.DefaultPassphrasePromptArg(mctx, username)
+	secretUI := h.getSecretUI(sessionID, h.G())
+	res, err := secretUI.GetPassphrase(arg, nil)
```
zapu: @patrickxb can you take a look? for simplicity it would be preferable to pass the passphrase over as argument to this RPC, but are we fine with that?

patrickxb: Wouldn't we want to use pinentry?  Which secret ui will handle.

protocol/avdl/keybase1/account.avdl
```diff
@@ -49,4 +49,6 @@ protocol account {
   
   GetLockdownResponse getLockdownMode(int sessionID);
   void setLockdownMode(int sessionID, boolean enabled);
+
+  void testPassphrase(int sessionID);
```
patrickxb: Can we change this name so `test` isn't the prefix?

zapu:
After couple more iterations:
- The RPC is now called `PassphraseCheck`.
- It accepts `passphrase` as an argument, and when it's empty, it will run the `secretUI` prompt. CLI calls the RPC with an empty passphrase.
   - The way account RPC handlers call secretUI is a bit weird, if not outdated, but I'm not sure there. I couldn't find a way to attach "fake" secretUI to test it in an unit test.
    - Re yesterday's discussion, I think it's fine to pass passphrases through arguments. That's what other RPCs do like `passphraseChange` or `resetAccount`. I don't know why I was under impression we should always use UI.

keybase-travis:
We have fake secret uis in many tests.

keybase-travis:
It is fine not to use a UI

go/service/account.go
```diff
@@ -165,3 +165,32 @@ func (h *AccountHandler) SetLockdownMode(ctx context.Context, arg keybase1.SetLo
 	_, err = mctx.G().API.Post(mctx, apiArg)
 	return err
 }
+
+func (h *AccountHandler) PassphraseCheck(ctx context.Context, arg keybase1.PassphraseCheckArg) (ret bool, err error) {
```
patrickxb: I think this should be in `engine` package.  Might make your testing easier too.

zapu: done, can you take a look? hopefully it's not too rough, half way in the code i realised this is essentially wrapping `libkb.VerifyPassphraseForLoggedInUser` call in an RPC, and maybe engine is a bit of an overkill... but still

PR #16484, opened 2019-03-07
rpc_cache: for best effort, only wait a short time for handler

strib:
But the handler should still run to completion, in order to populate the cache for next time.

Because the handler could keep running after the function returns, there's a new requirement that the handler cannot modify variables directly in the caller's stack.  Instead, if the handler's value is returned directly from `Serve()` if it is used, and the caller must check whether it should be using the returned value, or the value in `resPtr` filled in by the cache.

Ah the joys of generics in Go.

It looks like all current users of the RPC cache are RPC handlers in service that reply directly to the user, so I don't think this introduces any new problems of out-of-date values sitting in other in-memory service caches that never get invalidated.

With this change, a kbfsdocker test that disables kbweb connectivity while reading a synced TLF passes.

Issue: KBFS-3948

### DIFF
--- a/go/offline/rpc_cache.go
+++ b/go/offline/rpc_cache.go
@@ -10,12 +10,17 @@ import (
 	"github.com/keybase/client/go/protocol/keybase1"
 	"golang.org/x/net/context"
 	"sync"
+	"time"
 )
 
 type RPCCache struct {
 	sync.Mutex
 	edb *encrypteddb.EncryptedDB
 }
+
+const (
+	bestEffortHandlerTimeout = 500 * time.Millisecond
+)
 
 func newEncryptedDB(g *libkb.GlobalContext) *encrypteddb.EncryptedDB {
 	keyFn := func(ctx context.Context) ([32]byte, error) {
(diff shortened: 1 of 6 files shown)


go/offline/rpc_cache.go
```diff
@@ -121,40 +126,96 @@ func (c *RPCCache) put(mctx libkb.MetaContext, version Version, rpcName string,
...
+// a background goroutine after this function returns, to populate the
+// cache. `handler` also returns the return value for the RPC as an
+// interface, so it can be inserted into the offline cache in the
+// success case. We also pass this function a `version`, which will
+// tell the cache-access machinery to fail if the wrong version of the
+// data is cached. Next, we pass the `rpcName`, the argument, and the
+// pointer to which the result is stored if we hit the cache. This
+// flow is unfortunately complicated, but the issue is that we're
+// trying to maintain runtime type checking, and it's hard to do
+// without generics.
```
maxtaco: This comment is now out-of-date ^

go/offline/rpc_cache.go
```diff
@@ -121,40 +126,96 @@ func (c *RPCCache) put(mctx libkb.MetaContext, version Version, rpcName string,
...
+		err error
 	}
-	tmp := c.put(mctx, version, rpcName, encrypted, arg, res)
-	if tmp != nil {
-		mctx.Warning("Error putting RPC to offline storage: %s", tmp.Error())
+	resCh := make(chan handlerRes, 1)
+
+	// New goroutine needs a new metacontext, in case the original
+	// gets canceled after the timeout below.
+	newMctx := libkb.NewMetaContextBackground(mctx.G())
```
maxtaco: This'll kill the log tags.  Maybe try `mctx.BackgroundWithLogTags()`?

go/offline/rpc_cache.go
```diff
@@ -121,40 +126,96 @@ func (c *RPCCache) put(mctx libkb.MetaContext, version Version, rpcName string,
...
+		timerCh = make(chan time.Time)
+	}
+	select {
+	case hr := <-resCh:
+		if hr.err != nil {
+			return nil, hr.err
+		}
+		return hr.res, nil
+	case <-timerCh:
+		mctx.Debug("Timeout waiting for handler; using cached value instead")
```
maxtaco: Should we cancel the context passed to `handler` in this case (`newMctx`)?

go/offline/rpc_cache.go
maxtaco: or `mctx.BackgroundWithLogTags().WithCancel()` (see below)

go/offline/rpc_cache.go
strib: We definitely don't want to cancel the handler's ctx, since the whole point of this goroutine/context shenanigans is to have the handler finish up and populate the cache in the background, even if we serve the request from the cache after the timeout.

go/offline/rpc_cache.go
strib: Oo, I was looking for something like `BackgroundWithLogTags` but somehow didn't find it.  Perfect, thanks!

go/offline/rpc_cache.go
maxtaco: OK, makes sense.

go/offline/rpc_cache.go
strib: @maxtaco  I did my best to update the comment -- what about it is out of date now?  The last sentence still seems valid to me, and I updated all the stuff about what the handler should be doing.

maxtaco: I was thinking the last sentence is now no longer needed since you're casting back out of the handle (rather than currying in the reply pointer).

strib: Hmm ok thanks.  I'll get rid of it.

go/offline/rpc_cache.go
maxtaco: Not a big deal either way.

go/offline/rpc_cache.go
maxtaco: Not a big deal either way

PR #16487, opened 2019-03-08
folder_branch_ops: re-resolve and identify handle on sync enable

strib:
We need to make sure we re-trigger those RPCs with BEST_EFFORT availability passed in.

Issue: KBFS-3918


go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -8296,6 +8296,69 @@ func (fbo *folderBranchOps) triggerMarkAndSweepLocked() {
...
 
+func (fbo *folderBranchOps) reResolveAndIdentify(
+	ctx context.Context, oldHandle *TlfHandle, rev kbfsmd.Revision) {
+	fbo.log.CDebugf(ctx, "Re-resolving handle")
+	defer func() { fbo.log.CDebugf(ctx, "Done") }()
+
+	fbo.config.KeybaseService().ClearCaches(ctx)
+
+	// Once with the iteam-resolver, and once without it, to cache the
+	// individual users of the folder, and once with it.
```
jzila: This comment isn't clear. Once with, once without, then once with? That doesn't seem to be what happens below.

PR #16492, opened 2019-03-08
API#PostDecode requires a MetaContext

maxtaco:


go/teams/create.go
```diff
@@ -330,14 +331,14 @@ func CreateSubteam(ctx context.Context, g *libkb.GlobalContext, subteamBasename
 	public := false
 	subteamID := NewSubteamID(public)
 
-	perUserKeyUpgradeSoft(ctx, g, "create-subteam")
+	perUserKeyUpgradeSoft(ctx, mctx.G(), "create-subteam")
 
-	me, err := loadMeForSignatures(ctx, g)
+	me, err := loadMeForSignatures(ctx, mctx.G())
```
zapu: i wonder if we could have kept access through `g` argument here. what do you think should be a preference?

PR #16493, opened 2019-03-08
Finesse file rows in the widget.

adamjspooner:


jzila:
old: 
Screen Shot 2019-03-11 at 10 26 31 AM

new:
Screen Shot 2019-03-11 at 10 26 25 AM


I think we should do what we do here:
https://github.com/keybase/client/blob/cefa197f9be5b8c385a7c41366a1ec9cd9dd7856/shared/fs/destination-picker/index.js#L118-L125

adamjspooner:
Works for me.

@cecileboucheron, what do you think? (The suggestion is to use an ellipsis on long filenames before the file extension instead of wrapping: `213fo09jq4v0a9jv0jaw...mp4`).

cecileboucheron:
Agreed!

adamjspooner:
@jzila, any idea when this will make its way into master? https://github.com/keybase/client/blob/953cba21d7a7e101b2d44038f6af86a206090b30/shared/constants/fs.js#L876

jzila:
@adamjspooner you mean https://github.com/keybase/client/blob/master/shared/constants/fs.js#L876 ?

adamjspooner:
Maaaaaybe. 🤦‍♂️

adamjspooner:
@jzila, I've tried a bunch of different ways and can't get the truncated filenames to work. This PR places the file icon on the first row and wraps the filename (as seen in your screenshot). Want to have a go on making the truncation work?

jzila:
K I'll tinker with it a bit today. I feel like there's a grand unified theory of flexbox somewhere that we are missing.

PR #16501, opened 2019-03-08
output debug to see what's what

maxtaco:


maxtaco:
privacy issue, let's not do this.

PR #16505, opened 2019-03-08
Profile show unbroken proofs as 'new'

mlsteele:
Back when we tracked the proof it was failing. Now it's succeeding.


zapu:
is this a new design or just IDv3 got that incorrectly compared to what we had before?

mlsteele:
I think it's an improvement. I saw jakob's reddit proof had "upgraded". Didn't make sense to me. So I did some digging and found that I had tracked him back when his reddit was broken. "upgraded" reads like http -> https or something. Upgrading reddit doesn't make sense.

mlsteele:
cc @maxtaco

maxtaco:
👍

PR #16509, opened 2019-03-08
fix update now from widget

songgao:
Tested on my mac. I think it's working properly now.

### DIFF
--- a/go/install/install_darwin.go
+++ b/go/install/install_darwin.go
@@ -1398,6 +1398,9 @@ func StartUpdateIfNeeded(ctx context.Context, log logger.Logger) error {
 		return err
 	}
 	cmd := exec.Command(updaterPath, "check")
+	// Run it in a new process group so when we are killed eventually by the
+	// updater, we don't bring down the updater too.
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 	if err = cmd.Start(); err != nil {
 		return err
 	}


maxtaco:
@songgao  --- I think @chrisnojima , @mmaxim  and I all need more context and a fuller understanding of what's happening here.  I don't understand the basic architecture, and this is already a part of the code that is unfortunately complicated and brittle.  The last thing we should be doing is making it more so.

maxtaco:
Also cc @cjb

chrisnojima:
Lets just have a meeting next week

maxtaco:
Yes, good idea.

songgao:
Sure; happy to Meet™️. Before that here's some background:

This is called by RPC handler `keybase1.config.StartUpdateIfNeeded`. This RPC is called when user clicks "Update now" in widget. The bug is after "update now" is clicked, new version app never comes back up. What happens before this PR is:

1. User clicks "Update now", and we fire `keybase1.config.StartUpdateIfNeeded` RPC which calls `install.StartUpdateIfNeeded`.
2. `install.StartUpdateIfNeeded` spawns a process running the `updater` binary.
3. The `updater` process pulls the update, and stops the app before starting new one.
4. When `updater` process stops the app, it uses a mix of launchd, `keybase ctl stop`, and if all failed, `kill -9`. Launchd (I think) sends a signal to the service process to stop.
5. Since the spawned `updater` process is in the same **process group** as the service daemon, the `updater` process gets terminated too, and never gets a chance to start the new app.

The key here is step 5, where the `updater` is in the same process group as `service`, and get terminated together. So this PR just makes the spawned `updater` process a new process group so it doesn't die when app restarts.

Note that this isn't a problem for `keybase update check` CLI because in that case the CLI process rather than the service daemon spawns the `updater` process directly, so it never gets terminated by launchd.

PR #16515, opened 2019-03-10
track sent flips so we don't replay them during inject

mmaxim:
There currently is a problem if the following happens:

1. User sends a commitment into a flip.
2. Client is processing a bunch of other messages, but gets a message ID out of order for some unknown bizarre reason.
3. Client fetches from the flip thread all past message to inject any missed messages.
4. Picks up its own sent commitment (which hasn't yet shown up in the inject flow) and injects it.
5. Gets a duplicate registration error.

To fix, we maintain a map of outbox IDs that have been queued up to be sent, and make sure not to inject them if we pull them from the flip thread.

### DIFF
--- a/go/chat/flipmanager.go
+++ b/go/chat/flipmanager.go
@@ -139,11 +139,13 @@ type FlipManager struct {
 	cardMap        map[string]int
 	cardReverseMap map[int]string
 
-	gamesMu    sync.Mutex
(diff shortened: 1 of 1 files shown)


mmaxim:
@chrisnojima @malgorithms this might be worth releasing before post, we can discuss

PR #16525, opened 2019-03-11
Files header tooltip

mlsteele:
Wasn't obvious what this share button would do. I think a tooltip may help.
image


shared/fs/header/header.desktop.js
```diff
@@ -39,7 +39,9 @@ const FolderHeader = ({path, onChat, routePath}: FolderHeaderProps) => (
                 style={styles.headerIcon}
               />
             )}
-            <SendInAppAction path={path} sendIconClassName="" />
+            <Kb.WithTooltip text="Send link">
```
songgao: Maybe change this to something like “Send to chat”? This will be used for both links and attachments.

PR #6317, opened 2017-03-21
Mobile iOS doesn't support MatchLimitAll on delete

gabriel:
We also specify access group (which is empty on simulator).

This fixes the secret store getting cleared on mobile.

Care was made to not affect desktop at all (we special case iOS here on delete).

PR #16526, opened 2019-03-11
Skip matchlimit on ios

joshblum:
Patch does the following:

- refactors to use `ClearSecret` when trying to delete the keychain item before adding it. this helps with code readability to limit the ways we interact with these keychain items. (reverts https://github.com/keybase/client/pull/16171)
- skips `item.SetMatchLimit` on ios when trying to update accessibility, this param does not work on iOS and returns a `keychain.ErrorParam` (see also https://github.com/keybase/client/pull/6317).

cc @patrickxb


patrickxb:
Wasn't match limit already skipped on iOS?

patrickxb:
Oh I see, not in `updateAccessibility`

PR #16527, opened 2019-03-11
Add IsInflation, InflationSource to PaymentLocal, PaymentDetailsLocal

patrickxb:
We have had these fields in the remote interface for a while, this copies them to the local interface so frontend can use them.


patrickxb:
CI unrelated

PR #16533, opened 2019-03-11
autogit: support a file showing the contents of a commit

strib:
Special files of the form:

```
/keybase/<type>/<tlfname>/.kbfs_autogit/<reponame>/.kbfs_autogit_commit_<commithash>
```

now contain a git-show-like representation of the full commit metadata and patch.

Known drawbacks:

* The commit hash must be the full hash, not a short one.  Hopefully we have the full one in the place where we render the git commit messages in chat.
* go-git can't currently handle showing a "diff" of the initial commit, so for now the file for the first commit just shows the commit metadata.

Issue: KBFS-3949


go/kbfs/libgit/commit_file.go
```diff
@@ -0,0 +1,113 @@
...
+const (
+	commitFileDefaultMaxBufSize = 4 * 1024 * 1024 // 4 MB
+)
+
+type commitFile struct {
+	hash       plumbing.Hash
+	r          *strings.Reader
+	maxBufSize int64
+	mtime      time.Time
+}
```
jzila: Could you generalize this into a `diffFile`, rather than being focused on outputting the patch for a single commit?

strib: Ok done, hopefully what I did is what you were thinking.

PR #16535, opened 2019-03-11
Update file upload styles and create a common component.

adamjspooner:


shared/common-adapters/drag-and-drop.desktop.js
```diff
@@ -0,0 +1,130 @@
...
+      <Box style={styles.containerStyle} onDragOver={this._onDragOver}>
+        {children}
+        {this.state.showDropOverlay && <DropOverlay onDragLeave={this._onDragLeave} onDrop={this._onDrop} />}
+      </Box>
+    )
+  }
+}
+
+const styles = Styles.styleSheetCreate({
+  containerStyle: {
```
chrisnojima: we can clean up a bunch of these styles by using box2

shared/fs/folder/index.js
```diff
@@ -8,7 +8,7 @@ import FolderHeader from '../header/container'
 import Footer from '../footer/footer'
 import {isMobile} from '../../constants/platform'
 import Rows from '../row/rows-container'
-import DropTarget from './drop-target'
+import DragAndDrop from '../../common-adapters/drag-and-drop.desktop'
```
chrisnojima: you should make common-adapters/electron.desktop.js (like mobile.native.js) so we can put desktop specific stuff in there

adamjspooner: KBFS is using this in a place shared between desktop and mobile. So, I copied the `SafeAreaView` pattern instead. Let me know if you'd prefer me to do something else, though.

adamjspooner:
@chrisnojima, this is ready for another look.

shared/fs/folder/index.js
chrisnojima: if its native or shared it can't pull in .desktop files, is that what you mean?

adamjspooner: Right. So, I'm only rendering the children on native: https://github.com/keybase/client/pull/16535/files#diff-dd41b322f4d13e08cb35827f54986e11R4

adamjspooner:
@keybase/react-hackers, bump.

shared/common-adapters/drag-and-drop.desktop.js
```diff
@@ -0,0 +1,122 @@
+// @flow
+import * as React from 'react'
+import fs from 'fs'
+import * as Styles from '../styles'
+import {Box2} from './box'
+import Icon from './icon'
+import Text from './text'
+import type {Props} from './drag-and-drop'
+import logger from '../logger'
```
chrisnojima: just FYI a thing i've been doing lately is to put this on top
```
const Kb = {
Icon, Text, Box2
}
```

etc then you can do <Kb. whatever below
(you don't have to make this change in this pr but for future reference)

PR #16537, opened 2019-03-11
Upgrade to Dokan 1.2.2.1000

zanderz:
This version has been shown to fix some user problems and seems to work ok for me.
Also fix a slight UI regression.


taruti:
LGTM for the Dokan upgrade, it should fix all the "Filesystem goes away shortly after starting" issues.

PR #16539, opened 2019-03-11
use waitForClient to detect kbfs daemon connection status

songgao:
also handle not only KBFS daemon not connected, but also when KBFS disconnects/crashes while user is in Files tab.

cc @chrisnojima @patrickxb


shared/actions/fs/index.js
```diff
@@ -870,24 +877,30 @@ const clearRefreshTag = (state, action) => {
   mimeTypeRefreshTags.delete(action.payload.refreshTag)
 }
 
-let pingKbfsDaemonLoopRunning = false
-function* pingKbfsDaemonUntilConnected() {
-  if (pingKbfsDaemonLoopRunning) {
+// Can't rely on kbfsDaemonStatus === 'waiting' as that's set by reducre and
```
chrisnojima: reducer

shared/fs/container.js
```diff
@@ -78,9 +81,14 @@ class ChooseComponent extends React.PureComponent<ChooseComponentProps> {
     if (this.props.path !== prevProps.path) {
       this.props.loadPathMetadata()
     }
+    if (this.props.kbfsDaemonStatus !== 'connected') {
```
chrisnojima: we should discuss. i dislike components driving something like this. also i dislike a thing that could get called a million times by accident like this can

songgao: This doesn’t always results in a RPC call; saga part deduplicates. I just wanted a way to trigger the waitForClient call again after timeout first time. It seems this would result in less calls since it never calls as long as user doesn’t go into the Files tab.

Would hooking this up with routing be better here?

songgao: @chrisnojima bump!

PR #16540, opened 2019-03-11
Nav2 Provisioning

MarcoPolo:
@keybase/react-hackers Upgrades login flows to support the new nav model.

Adds support for `customOnBack` in navigation param which will be used when supplied instead of the default onBack in headers.

I needed this customOnBack because just going back one level is broken in the login flow, it's only really valid to go all the way back to the start.

todo (in another PR maybe?)
- [ ] Cancel provision rpc when the user goes back (it works fine without this, but might be nice to clean things up)


chrisnojima:
I didn't look at the code but I think for floes like this we want a replace vs a push

MarcoPolo:
O interesting. That makes sense, I'll change that now

MarcoPolo:
@chrisnojima updated

chrisnojima:
We should make a routetree actiknfor replace. Or add a flag to append

MarcoPolo:
I thought the goal was to remove routetree stuff eventually?

MarcoPolo:
And remove the idea of us dispatching to affect navigation

chrisnojima:
Eventually but short term we'll have actions that will affect nav. Longer term maybe all the actions go away. More likely the api will shrink. While we support both it makes sense to support the old actions but when we deprecate that system we can type it better and drop old concepts (like parent path and arrays in path)

chrisnojima:
Either way we should dispatch some action instead of using _getnavigator in sagas

chrisnojima:
i made a getRouteProps helper in util/container that we can use

MarcoPolo:
Killing this in favor of the single PR

PR #16544, opened 2019-03-12
Better NameWithIcon stories

cecileboucheron:
@keybase/react-hackers


shared/common-adapters/name-with-icon/index.stories.js
```diff
@@ -23,34 +22,63 @@ const provider = Sb.createPropProviderWithCommon(
...
+          {...commonProps}
+          teamname="keybasefriends"
+          title="keybasefriends"
+          metaOne="786 members"
+          size="small"
         />
-        <NameWithIcon {...commonProps} username="cecileb" metaOne="Cécile Boucheron" size="small" />
         <NameWithIcon
           {...commonProps}
-          icon="iconfont-crown-admin"
```
akalin-keybase: no need for an example with the admin anymore?

cecileboucheron: Nah, we never really use the component with an icon from the icon font.

PR #16556, opened 2019-03-12
slottedAccounts for KeychainSecretStore

joshblum:
```
// NOTE There have been bug reports where we are unable to store a secret in
// the keychain since there is an existing corrupted entry that cannot be
// deleted (returns a keychain.ErrorItemNotFound) but can also not be written
// (return keychain.ErrorDuplicateItem). As a workaround we add a slot number
// to the accountName field to write the secret multiple times, using a new
// slot if an old one is corrupted. When reading the store we return the last
// secret we have written down.
```

- [x] test on mobile

### DIFF
--- a/go/libkb/secret_store_darwin.go
+++ b/go/libkb/secret_store_darwin.go
@@ -143,22 +143,21 @@ func (k KeychainSecretStore) RetrieveSecret(mctx MetaContext, accountName Normal
 
 	// find the last valid item we have stored in the keychain
 	var previousSecret LKSecFullSecret
-Loop:
 	for i := 0; i < maxKeychainItemSlots; i++ {
 		account := newKeychainSlottedAccount(accountName, i)
 		secret, err = k.retrieveSecret(mctx, account)
-		switch err {
-		case nil:
+		if err == nil {
 			previousSecret = secret
-		case keychain.ErrorItemNotFound, NewErrSecretForUserNotFound(accountName):
+			mctx.Debug("successfully retrieved secret on attempt: %d, checking if there is another filled slot", i)
+		} else if _, ok := err.(SecretStoreError); ok || err == keychain.ErrorItemNotFound {
 			// We've reached the end of the keychain entries so let's return
 			// the previous secret we found.
 			secret = previousSecret
 			k.updateAccessibility(mctx, account)
 			err = nil
-			mctx.Debug("successfully retrieved secret on attempt: %d", i-1)
-			break Loop
-		default:
+			mctx.Debug("found last slot: %d, finished read", i)
+			break
+		} else {
 			mctx.Debug("unable to retrieve secret: %v, attempt: %d", err, i)
 		}
 	}
--- a/go/libkb/secret_store_darwin.go
+++ b/go/libkb/secret_store_darwin.go
@@ -7,138 +7,226 @@ package libkb
 
 import (
 	"encoding/base64"
+	"fmt"
 	"os"
+	"strings"
 
 	keychain "github.com/keybase/go-keychain"
 )
 
-type KeychainSecretStore struct {
+const maxKeychainItemSlots = 50
+const slotSep = "/"
+
+type keychainSlottedAccount struct {
+	name NormalizedUsername
+	slot int
 }
+
+func newKeychainSlottedAccount(name NormalizedUsername, slot int) keychainSlottedAccount {
+	return keychainSlottedAccount{
+		name: name,
+		slot: slot,
+	}
+}
+
+// keychainSlottedAccount is used in case we can not longer delete/update an entry
+// due to keychain corruption. For backwards compatibility the initial slot
+// just returns the accountName field.
+func (a keychainSlottedAccount) String() string {
+	if a.slot == 0 {
+		return a.name.String()
+	}
+	return fmt.Sprintf("%s%s%d", a.name, slotSep, a.slot)
+}
+
+func parseSlottedAccount(account string) string {
+	parts := strings.Split(account, slotSep)
+	if len(parts) == 0 {
+		return account
+	}
+	return parts[0]
+}
+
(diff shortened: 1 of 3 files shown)


joshblum:
@patrickxb tested the login/logout/provision flow on desktop and mobile? anything else worth trying?

go/libkb/secret_store_darwin.go
```diff
@@ -7,138 +7,226 @@ package libkb
...
+	// find the last valid item we have stored in the keychain
+	var previousSecret LKSecFullSecret
+Loop:
+	for i := 0; i < maxKeychainItemSlots; i++ {
+		account := newKeychainSlottedAccount(accountName, i)
+		secret, err = k.retrieveSecret(mctx, account)
+		switch err {
+		case nil:
+			previousSecret = secret
+		case keychain.ErrorItemNotFound, NewErrSecretForUserNotFound(accountName):
```
patrickxb: `err` will never equal the object returned by NewErrSecretForUserNotFound, right?  you could check the types for equality.

go/libkb/secret_store_darwin.go
```diff
@@ -7,138 +7,226 @@ package libkb
...
 	if accountName.IsNil() {
-		m.Debug("NOOPing KeychainSecretStore#ClearSecret for empty username")
+		mctx.Debug("NOOPing KeychainSecretStore#ClearSecret for empty username")
 		return nil
 	}
-	query := keychain.NewGenericPassword(k.serviceName(m), string(accountName), "", nil, k.accessGroup(m))
+
+	// Try all slots to fully clear any secrets for this user
+	epick := FirstErrorPicker{}
+	for i := 0; i < maxKeychainItemSlots; i++ {
```
maxtaco: Is this slow? Iterating over 50 slots for every user? Even the 99.99% who don't have this issue?

maxtaco: Another possibility is just to stop iterating once we hit a slot that was never set?

go/libkb/secret_store_darwin.go
```diff
@@ -7,138 +7,226 @@ package libkb
...
+
+	// find the last valid item we have stored in the keychain
+	var previousSecret LKSecFullSecret
+Loop:
+	for i := 0; i < maxKeychainItemSlots; i++ {
+		account := newKeychainSlottedAccount(accountName, i)
+		secret, err = k.retrieveSecret(mctx, account)
+		switch err {
+		case nil:
+			previousSecret = secret
```
maxtaco: Is logging useful here?

go/libkb/secret_store_darwin.go
joshblum: yeah i was worried about it being an issue but in tests it runs very quickly: `KeychainSecretStore#ClearSecret: accountName: username -> ok [time=12.313227ms]` so i thought it was better to be safe and attempt the delete for any possible values since actual `DeleteItem` get false positives of `keychain.ErrorItemNotFound` (we could try to read ahead like `StoreSecret` does but thought it wasn't worth the added complexity)

go/libkb/secret_store_darwin.go
joshblum: i'll add something

go/libkb/secret_store_darwin.go
maxtaco: ah ok.  that's fast enough.  Is that on mobile or desktop?

go/libkb/secret_store_darwin.go
joshblum: we get equality since the tests depend this error being in the switch but think the type check will be cleaner

go/libkb/secret_store_darwin.go
joshblum: desktop, i can double check on mobile but i think it was similar during my manual testing

joshblum: mobile: `KeychainSecretStore#ClearSecret: accountName: test5 -> ok [time=71.138ms] `

joshblum:
@maxtaco @patrickxb this is ready for another pass

go/libkb/secret_store_darwin.go
```diff
@@ -7,138 +7,225 @@ package libkb
...
+		account := newKeychainSlottedAccount(accountName, i)
+		secret, err = k.retrieveSecret(mctx, account)
+		if err == nil {
+			previousSecret = secret
+			mctx.Debug("successfully retrieved secret on attempt: %d, checking if there is another filled slot", i)
+		} else if _, ok := err.(SecretStoreError); ok || err == keychain.ErrorItemNotFound {
+			// We've reached the end of the keychain entries so let's return
+			// the previous secret we found.
+			secret = previousSecret
+			k.updateAccessibility(mctx, account)
```
zapu: @joshblum do you remember why do we try to update an item that we know doesn't exist? or is that we are correcting for the case where item exists but is somehow inaccessible and we are hoping for it to work during next try?

zapu: oh but that not what `updateAccessibility` is doing... I don't think this is right what's in here

joshblum: i don't remember why, it might have been keeping the previous functionality for this case. is it causing a specific issue?

zapu: while I'm looking for issues in `secret_store_darwin.go`, i don't think this causes any of the issues I'm trying to solve.

I think `updateAccessibility` was used to change some parameter of existing keychain entries. I think `k.updateAccessibility(mctx, account)` should happen e.g. on line 152 - if an entry is found, use the opportunity to update it. At least that's what I gather is how it used to work becuase the change to multiple slots for username.

There may be a possibility that this code is not needed anymore, too.

I think @mmaxim wrote the original update code, maybe he will remember.

PR #16557, opened 2019-03-12
libkbfs: two fixes to prevent double archiving of block pointers

strib:
The server should be able to handle a request to archive the same block twice, but there's no point in saving/sending redundant data.  So two fixes: 

1. When creating a `resolutionOp`, don't add an unref'd pointer that also exists in an already-included `rmOp` in the resolved MD.
2. Uniquify the pointers we send in our batch archive/delete calls to the bserver.

Issue: KBFS-3966

### DIFF
--- a/go/kbfs/libkbfs/folder_block_manager.go
+++ b/go/kbfs/libkbfs/folder_block_manager.go
@@ -856,7 +856,6 @@ func getUnrefPointersFromMD(
 			// See syncInfo.replaceRemovedBlock for an example
 			// of how this can happen.
 			if ptr != zeroPtr && !ptrMap[ptr] {
-				ptrs = append(ptrs, ptr)
 				ptrMap[ptr] = true
 			}
 		}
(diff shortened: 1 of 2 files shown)


go/kbfs/libkbfs/folder_block_manager.go
```diff
@@ -854,17 +855,19 @@ func getUnrefPointersFromMD(
...
 		}
 		for _, update := range op.allUpdates() {
 			// It's legal for there to be an "update" between
 			// two identical pointers (usually because of
 			// conflict resolution), so ignore that for quota
 			// reclamation purposes.
-			if update.Ref != update.Unref {
+			if update.Ref != update.Unref && !ptrMap[update.Unref] {
 				ptrs = append(ptrs, update.Unref)
+				ptrMap[update.Unref] = true
```
jzila: I think it'd be better to use the map alone for accounting within this function, then producing the slice one time at the end. That would help prevent bugs where we forget to do both in each spot.

PR #16559, opened 2019-03-13
don't let writers change team avatar

zanderz:
The service was letting team writers change the avatar and description - that was the root cause, right?

For a generic error, the description will be shown, too:
image


go/teams/service_helper.go
```diff
@@ -1613,7 +1613,7 @@ func CanUserPerform(ctx context.Context, g *libkb.GlobalContext, teamname string
 
 	ret.DeleteChannel = admin
 	ret.RenameChannel = writer
-	ret.EditChannelDescription = writer
+	ret.EditChannelDescription = admin
```
chrisnojima: Do we want this change?

chrisnojima: Description and avatar are the same?

chrisnojima:
Cc: @mmaxim

go/teams/service_helper.go
mmaxim: Definitely not.

mmaxim:
If the avatar UI is using that field to control who can do it, we should just make a new field, not change the channel description one.

chrisnojima:
That's what I thought thanks

go/teams/service_helper.go
zanderz: so someone with write permission is supposed to be able to edit the team description? It didn't work that way yesterday, so I assumed they were equivalent.

mmaxim: Assuming this is used for how it is named, this is referring to channel description.

go/systests/teams_test.go
```diff
@@ -1279,6 +1279,7 @@ func TestTeamCanUserPerform(t *testing.T) {
 	require.True(t, annPerms.DeleteChannel)
 	require.True(t, annPerms.RenameChannel)
 	require.True(t, annPerms.EditChannelDescription)
+	require.True(t, annPerms.ChangeChannelAvatar)
```
mmaxim: It's not a channel avatar, that is not a thing that exists. The avatar is for the team. I think the root cause here is that `EditChannelDescription` is used in the teams code for teams related stuff, beyond this avatar problem (which is I guess why you were involved with it in the first place). You should make a new field here called `EditTeamDescription`, and use it for both the avatars and the actual team description.

You get the same error trying to edit the team description as a writer, but no black bar for some reason. cc @buoyad

go/teams/service_helper.go
zanderz: :white_check_mark:

go/systests/teams_test.go
zanderz: :white_check_mark:

PR #16563, opened 2019-03-13
Frontend: No password flow

cjb:
@keybase/react-hackers CC @cecileboucheron @zapu 

This PR:

* changes signup to skip the screen that asks for a password, and signup passwordless by default

* changes the Sign out flow to launch one of two modals:
  * if you have a password set: you can optionally test it before you log out
  * if you don't have a password set: you must set one before you can log out

Setting a password re-uses the component from the normal Change Passphrase flow.

I'm sure there'll be design tweaks, but here's a starting point to make them from.


shared/actions/settings.js
```diff
@@ -403,6 +403,14 @@ function* processorProfile(_, action) {
 const rememberPassphrase = (_, action) =>
   RPCTypes.configSetRememberPassphraseRpcPromise({remember: action.payload.remember})
 
+const testPassphrase = (_, action) =>
```
zapu: what do you think about this becoming `checkPassphrase` as well? there was some concern at CORE that it might be mistaken for a test code of some kind - I don't know if you share these. but feel free to leave it as is, I don't feel strongly about it

shared/reducers/settings.js
```diff
@@ -257,6 +263,7 @@ function reducer(state: Types.State = Constants.initialState, action: SettingsGe
     case SettingsGen.processorProfile:
     case SettingsGen.unfurlSettingsRefresh:
     case SettingsGen.loadHasRandomPw:
+    case SettingsGen.testPassphrase:
```
chrisnojima: this should actually set testPasspraseIsCorrect to null to indicate we're in the act of determining it

shared/constants/settings.js
```diff
@@ -39,6 +39,7 @@ const initialState: Types.State = {
     randomPW: null,
     rememberPassphrase: true,
   },
+  testPassphraseIsCorrect: null,
```
chrisnojima: one of the last stragglers that not immutable 😭

shared/settings/logout/container.js
```diff
@@ -0,0 +1,43 @@
...
+    dispatch(SettingsGen.createOnSubmitNewPassphrase())
+  },
+  onTestPassphrase: passphrase =>
+    dispatch(SettingsGen.createTestPassphrase({passphrase: new HiddenString(passphrase)})),
+})
+
+const mergeProps = (stateProps, dispatchProps, ownProps) => ({
+  ...stateProps,
+  ...dispatchProps,
+  heading: stateProps.hasRandomPW
```
chrisnojima: should likely just go into the component

cecileboucheron:
- Both the "Set a passphrase" and the "Change passphrase" screens from the Account tab should be modals. So that modal would have 4 states: 
    - Just setting a passphrase (Account tab)
    - Changing my existing passphrase (Account tab)
    - Signing out and setting a passphrase (Sign out tab)
    - Signing out and checking user knows their passphrase (Sign out tab)

- With my account without a password, I went to sign out, it popped up the Create passphrase screen. I created a passphrase, and it took my to my Account page, instead of signing me out.

- Let's not pop up the "Your password needs to be at least 8 characters" error when you actually just started typing it. We don't even need this error right? The button gets enabled once you have enough letters anyway.
image

- Account page: when there's no passphrase set yet, let's hide the "Remember passphrase" checkbox instead of disabling it:
Screen Shot 2019-03-13 at 4 46 16 PM

- Modal should be 560px wide and let's use the full-width ButtonBar for the buttons, with a 16px padding all around (on mobile too). Here is a design proposal for the 5 states: https://zpl.io/bABMvwB

- In terms of copy, we should say "Sign out" everywhere and we're switching to "Password" instead of "Passphrase".

(cc @malgorithms)

adamjspooner:
> Let's not pop up the "Your password needs to be at least 8 characters" error when you actually just started typing it. We don't even need this error right? The button gets enabled once you have enough letters anyway.

We need something, though. Maybe a note below the input? It would be really confusing to enter a 7-character password and not be able to do anything.

adamjspooner:
> Not set! You should set a passphrase.

Should we explain why?

cecileboucheron:
> Maybe a note below the input?
Sounds good to me. @cjb Maybe under the first input, in BodySmall: "Your password should be at least 8-character long.".

> "Not set! You should set a passphrase." Should we explain why?
This is coming with my new designs so let's not bother for now.

malgorithms:
> This is coming with my new designs so let's not bother for now.

as you think about alternate wording, we don't actually want people to set keybase passwords. (so "you should set a passphrase" is the opposite..."you only need a keybase password if you want to log in to the keybase.io website or temporarily log out of one of your installs." That kinda thing, only as concise as possible.)

cjb:
@zapu @chrisnojima @cecileboucheron @adamjspooner Made all of these changes! Ready for another look.

I didn't change passphrase->password in this PR, since some of our uses of passphrase come from core (and even from the API server).

cecileboucheron:
* On mobile clicking "Test password" when the field is empty gets me stuck 
image
* On mobile we should be using the Modal type of page for that page (with no main nav and sliding from the bottom)
* Clicking "Safely sign out" shows the previous button bar again for 1 or 2 sec - could we avoid that?

image

cjb:
@cecileboucheron Fixed the 1st and 3rd bullet points, converting to a wallet-like modal (with the button bar pinned against the keyboard) is taking longer 'cause we don't have a common component for that kind of modal, might have to ticket adding it afterwards.

cjb:
@chrisnojima bump, think this is all ready to go in

shared/settings/logout/container.js
```diff
@@ -0,0 +1,49 @@
...
+
+type OwnProps = RouteProps<{}, {}>
+
+const mapStateToProps = (state, {routeProps}) => ({
+  checkPassphraseIsCorrect: state.settings.checkPassphraseIsCorrect,
+  hasRandomPW: state.settings.passphrase.randomPW,
+  waitingForResponse: state.settings.waitingForResponse,
+})
+
+const mapDispatchToProps = (dispatch, {navigateUp, routeProps}) => ({
```
chrisnojima: just a note this isn't compat w/ nav2

shared/settings/logout/index.js
```diff
@@ -0,0 +1,149 @@
...
+  onLogout: () => void,
+|}
+
+type State = {|
+  passphrase: string,
+  showTyping: boolean,
+|}
+
+class OfferToCheckPassphrase extends React.Component<TestProps, State> {
+  state: State
```
chrisnojima: don't need this extra state line

shared/settings/logout/index.js
```diff
@@ -0,0 +1,149 @@
...
+                Your password is correct.
+              </Kb.Text>
+            </Kb.Box2>
+          )}
+        </Kb.Box2>
+        <Kb.ButtonBar align="center" direction="row" fullWidth={true} style={styles.buttonbar}>
+          <Kb.Button
+            fullWidth={true}
+            label={this.props.checkPassphraseIsCorrect ? 'Safely sign out' : 'Just sign out'}
+            onClick={() => this.props.onLogout()}
```
chrisnojima: can just be this.props.onLogout

cjb:
@chrisnojima Thanks! All fixed and tested in nav2.

PR #16566, opened 2019-03-13
Update shh and boom! icons.

adamjspooner:
@keybase/react-hackers, this is ready for review. Aimee and I both worked on it.


shared/chat/avatars.js
```diff
@@ -31,9 +31,9 @@ const MutedIcon = ({isHovered, isMuted, isSelected, isLocked}) => {
   let type
   if (isMuted) {
     if (isMobile) {
-      type = isSelected ? 'icon-shh-active-24' : 'icon-shh-24'
+      type = isSelected ? 'icon-shh-active-26-21' : 'icon-shh-26-21'
```
chrisnojima: why do these have this naming convention

adamjspooner: It's their dimensions. They're not square.

PR #16573, opened 2019-03-13
Nav2 Chat part 2

buoyad:
Adds a `popToScreen` route action that will pop until it reaches a certain screen if found. Uses it to go back to the inbox after leaving a conversation. r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/attachment-fullscreen/container.js
+++ b/shared/chat/conversation/attachment-fullscreen/container.js
@@ -2,10 +2,12 @@
 import * as Types from '../../../constants/types/chat2'
 import * as Constants from '../../../constants/chat2'
 import * as Chat2Gen from '../../../actions/chat2-gen'
+import * as RouteTreeGen from '../../../actions/route-tree-gen'
 import * as FsGen from '../../../actions/fs-gen'
 import Fullscreen from './'
 import {compose, withStateHandlers, connect, withProps} from '../../../util/container'
 import {type RouteProps} from '../../../route-tree/render-route'
+import flags from '../../../util/feature-flags'
 
 const blankMessage = Constants.makeMessageAttachment({})
 
@@ -18,7 +20,7 @@ const mapStateToProps = state => {
   }
 }
 
-const mapDispatchToProps = (dispatch, {navigateUp, navigateAppend}: OwnProps) => ({
+const mapDispatchToProps = (dispatch, {navigateUp}: OwnProps) => ({
   _onDownloadAttachment: (message: Types.MessageAttachment) => {
     dispatch(
       Chat2Gen.createAttachmentDownload({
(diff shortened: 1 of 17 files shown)


shared/router-v2/router.shared.js
```diff
@@ -86,19 +99,27 @@ export const oldActionToNewActions = (action: any, navigation: any) => {
...
+    case RouteTreeGen.clearModals: {
+      const numModals = getNumModals(navigation)
       return numModals ? [StackActions.pop({n: numModals})] : []
+    }
     case RouteTreeGen.navigateUp:
       return [StackActions.pop()]
+    case RouteTreeGen.popToScreen: {
+      const modals = getNumModals(navigation)
+      const appPath = Constants._getAppStackForNavigator(navigation.state)
+      if (!appPath.length) {
```
chrisnojima: unclear if this is what we want. seems like it should just go back through everything. if it finds it it pops to there.aka if my 'target' is the root modal this'll do nothing (etc)

shared/teams/routes.js
```diff
@@ -137,21 +137,21 @@ export default routeTree
...
-  },
   member: {getScreen: () => require('./team/member/container').default},
-  showJoinTeamDialog: {getScreen: () => require('./join-team/container').default},
-  showNewTeamDialog: {getScreen: () => require('./new-team/container').default},
   'tabs.teamsTab': {getScreen: () => require('./container').default},
   team: {getScreen: () => require('./team/container').default},
 }
 
 export const newModalRoutes = {
   addPeople: {getScreen: () => require('./add-people/container').default},
```
chrisnojima: can these be upgraded: true

shared/router-v2/router.shared.js
buoyad: Good point, yeah this assumes the target is in the app path.

buoyad: @chrisnojima so there was another problem with this. If we're on the infopanel modal, we can't `pop` past the conversation, because that is in a different navigator. I changed it to use `NavigationActions.navigate`.

buoyad:
@chrisnojima ok, upgraded a bunch of routes + left a TODO for the rest. PTAL

shared/chat/conversation/attachment-fullscreen/container.js
```diff
@@ -45,7 +47,7 @@ const mapDispatchToProps = (dispatch, {navigateUp, navigateAppend}: OwnProps) =>
       dispatch(FsGen.createOpenLocalPathInSystemFileManager({localPath: message.downloadPath}))
   },
   onClose: () => {
-    dispatch(navigateUp())
+    dispatch(flags.useNewRouter ? RouteTreeGen.createNavigateUp() : navigateUp())
```
chrisnojima: just fyi i think its safe to just dispatch the global one and ignore the local one. nav1 will go away so im' not worried about losing the actionifonpath version

buoyad:
@chrisnojima changed to do multiple `pop`s instead, works like a charm. also switched the containers I changed to use the global navigateUp.

buoyad:
@keybase/react-hackers bump

PR #16574, opened 2019-03-13
libkbfs: move favorite data types into new `favorites` package

strib:
Eventually I'd like to also move all the favorites.go logic, but that's tied up in a lot of libkbfs stuff. But the data types are pretty self-contained and used by another type I'd like to move next, so I thought I'd get this done first.

### DIFF
--- a/go/kbfs/favorites/data_types.go
+++ b/go/kbfs/favorites/data_types.go
@@ -1,0 +1,110 @@
+// Copyright 2019 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package favorites
+
+import (
+	"strings"
+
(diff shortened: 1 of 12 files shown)


strib:
> This is fine, although it might be a good opportunity to think about the "favorites" nomenclature. Since it doesn't really represent something that is necessarily favorited. It could be ignored, for example.

I'm cool with that, though I don't have any good suggestions.  In this PR the word "Favorite" is out of all the data structures and is only in the package name.  I guess we could call the package "tlfutils" or something generic if you prefer.  But in terms of what's left in libkbfs, I'm worried it might get a little confusing if the keybase1 and api.keybase.io APIs are still using "favorite" everywhere.  Changing those is a bit more than I wanted to bite off in this PR...

jzila:
Agreed, hence why I was suggesting doing it in a subsequent PR. Anyway this is fine.

PR #16576, opened 2019-03-13
Add fallback profile proof order

mlsteele:
Add a fallback ordering for clients that aren't fetching proof configs (almost everyone).

Also turns out admins on macos probably aren't fetching proof config because setting `KEYBASE_FEATURES` for most shells doesn't set it for launchd.


mlsteele:
I think it's less complicated with less fake stuff. This preserves that display configs come only from the server

PR #16578, opened 2019-03-13
Mobile builds get 'prelease production' tags

mlsteele:
ios and android builds were'nt using the build tag `production` when compiling go. @strib this affects `defaultMountType`

ios looks good in the simulator. I didn't try android locally.


mmaxim:
Did you try it?

mlsteele:
I tried ios in the simulator. It definitely had an effect judging by the presence of Rooter on my profile. Sending chats and making a folder in the Files tab both worked.

PR #16189, opened 2019-02-22
select users design

zanderz:

PR #16579, opened 2019-03-13
Don't allow font generation without fontforge

buoyad:
Before, it would still write the font, but the metrics would be wrong. r? @keybase/react-hackers

### DIFF
--- a/shared/desktop/yarn-helper/font.js
+++ b/shared/desktop/yarn-helper/font.js
@@ -83,7 +83,9 @@ function updateIconFont(web) {
       execSync('fontforge')
     } catch (e) {
       if (e.message.includes('not found')) {
-        throw new Error('FontForge is required to generate the icon font. Install and try again.')
+        throw new Error(
+          'FontForge is required to generate the icon font. Run `yarn`, install FontForge CLI globally, and try again.'
+        )
       }
       throw e
     }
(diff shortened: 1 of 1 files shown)


adamjspooner:
Does this fix the problem we were seeing with @zanderz's PR?

shared/desktop/yarn-helper/font.js
```diff
@@ -77,6 +77,18 @@ const getSvgPaths = skipUnmatchedFile =>
  * For config options: https://github.com/sunflowerdeath/webfonts-generator
  */
 function updateIconFont(web) {
+  if (!web) {
+    // Check if fontforge is installed, required to generate the font
+    try {
+      execSync('fontforge')
+    } catch (e) {
+      if (e.message.includes('not found')) {
+        throw new Error('FontForge is required to generate the icon font. Install and try again.')
```
chrisnojima: you should say they need to run `yarn` and not `yarn modules` and install fontforge globally

buoyad: fontforge isn't installed by yarn though

chrisnojima: yeah additionally they need to install fontforge globally, they need both and its not the usual flow

buoyad:
@adamjspooner #16189 doesn't generate the font. Unless the font in master was generated without fontforge I don't think this fixes it.

PR #16580, opened 2019-03-13
Fix background ek generation

joshblum:
Fixes a bug where the background creation of a team ek would not happen even if we were unable to access the latest team ek. the issue was that we incorrectly calculated the max generation, which previously skipped errored items. also exposed a bug where we did not correctly cache errors which were allowing tests to pass incorrectly


zapu:
i forgot about this, looking now, sorry!

PR #16581, opened 2019-03-13
Files: use ListItem2 for rows

songgao:


shared/common-adapters/index.js
```diff
@@ -51,7 +51,11 @@ export {default as List} from './list'
 export {default as List2} from './list2'
 export {default as LoadingLine} from './loading-line'
 export {default as ListItem} from './list-item'
-export {default as ListItem2} from './list-item2'
+export {
```
chrisnojima: why do this

songgao: We need the height here: https://github.com/keybase/client/pull/16581/files#diff-162da979450f7a607e53a90a91add4aaR84

Would you prefer to not have these exported and import the list-item2 directly rather from Kb?

PR #16584, opened 2019-03-14
only localize new CHAT type conversations

joshblum:
skip localization for non chat type conversations, should help client performance on flips

### DIFF
--- a/go/chat/push.go
+++ b/go/chat/push.go
@@ -655,17 +655,31 @@ func (g *PushHandler) Activity(ctx context.Context, m gregor.OutOfBandMessage) (
 
 			uid := m.UID().Bytes()
 
-			// We need to get this conversation and then localize it
+			// If the topic type is not CHAT, skip the localization step
 			var inbox types.Inbox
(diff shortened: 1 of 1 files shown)


mmaxim:
Glancing at KBFS, they might need these notifications to work for file edits, so we probably want to do it for those as well.

go/chat/push.go
```diff
@@ -655,34 +655,53 @@ func (g *PushHandler) Activity(ctx context.Context, m gregor.OutOfBandMessage) (
...
+						ConvIDs:      []chat1.ConversationID{nm.ConvID},
+						MemberStatus: chat1.AllConversationMemberStatuses(),
+					}, nil); err != nil {
+					g.Debug(ctx, "chat activity: unable to read unverified conversation: %v", err)
+					return
+				}
 			}
-			if len(inbox.Convs) != 1 {
+
+			if len(inbox.ConvsUnverified) != 1 {
```
mmaxim: Isn't this always going to fail for chat convs?

joshblum: i don't think so or this line would have already broken: https://github.com/keybase/client/pull/16584/files#diff-50b204c7a958bc27261ba02d3f66bd7eR689

PR #16590, opened 2019-03-14
Show chat download file errors

mlsteele:
If the service returns an error from `DownloadFileAttachmentLocal`, show it to the user
out


shared/chat/conversation/messages/attachment/file/index.js
```diff
@@ -69,6 +77,7 @@ const styles = Styles.styleSheetCreate({
     position: 'absolute',
     right: 0,
   },
+  fail: {color: Styles.globalColors.red},
```
mmaxim: Why not error?

shared/actions/json/chat2.json
```diff
@@ -325,7 +325,8 @@
     // We saved an attachment to the local disk
     "attachmentDownloaded": {
       "message": "Types.Message",
-      "path?": "string"
+      "path?": "string",
+      "canError": {"error": "string", "message": "Types.Message"}
```
mmaxim: What is this?

shared/chat/conversation/messages/attachment/file/index.js
mlsteele: you mean why not name it "error" instead of "fail"? I just copied it from somewhere else. Can do

shared/actions/json/chat2.json
mmaxim: Forgot about this thing.

shared/chat/conversation/messages/attachment/file/index.js
```diff
@@ -40,6 +41,13 @@ class FileAttachment extends React.PureComponent<Props> {
               {this.props.hasProgress && <Kb.ProgressBar ratio={this.props.progress} />}
             </Kb.Box>
           )}
+          {!!this.props.errorMsg && (
+            <Kb.Box style={styles.progressContainerStyle}>
+              <Kb.Text type="BodySmall" style={styles.fail}>
+                {this.props.errorMsg}
```
mmaxim: Why not just have some static text in here saying "Failed to download attachment, please retry" instead of whatever the service is going to produce.

mlsteele: Depends what happened whether retrying will help. Won't help if you're offline or your network is dead or the directory doesn't exist.

mmaxim: It is not likely that whatever error the server plumbs up is going to give users any kind of useful information anyway, certainly will be tricky to parse it to figure out if retrying is worth their time. Worst case is they try again and it doesn't work again.

mlsteele: ok

PR #16591, opened 2019-03-14
add notification plumbing for a verified email address

xgess:
grab the new `email.verified` notification (see https://github.com/keybase/keybase/pull/3572) and plumb it up to the front end. 

this depends on that server-side PR (i.e. one test failure is expected)


zapu:
also we will be able to send a notification like "email.changed" if someone adds/removes an email, so frontend can refresh as well. this is cool!

PR #16594, opened 2019-03-15
libkbfs: extract all identify and tlfhandle code into new packages

strib:
This creates two new kbfs subpackages, consisting of existing code moved around:

* idutil: Basic data structures, interfaces, and helper code for dealing with identity data for users and teams.
* tlfhandle: The data structure for "Handles" to top-level folders (TLFs), which represent an identifier for each TLF, containing all the user or team IDs associated with the it.

Not only are these decent cleanups on their own, they will enable further refactoring of code that uses TlfHandles into other new packages (which is a lot of them).

Also I fixed up the README to point to all the new packages we've added since it was written.

### DIFF
--- a/go/kbfs/libkbfs/kbfs_ops_test.go
+++ b/go/kbfs/libkbfs/kbfs_ops_test.go
@@ -296,7 +296,7 @@ func checkBlockCache(
 }
 
 // parseTlfHandleOrBust parses the given TLF name, which must be
-// canonical, into a TLF handle, and failing if there's an error.
+// canonical, into a TLF handle, failing if there's an error.
 func parseTlfHandleOrBust(t logger.TestLogBackend, config Config,
 	name string, ty tlf.Type, id tlf.ID) *tlfhandle.Handle {
 	ctx := context.Background()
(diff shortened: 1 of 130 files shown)


go/kbfs/libkbfs/kbfs_ops_test.go
```diff
@@ -292,6 +295,20 @@ func checkBlockCache(
 	}
 }
 
+// parseTlfHandleOrBust parses the given TLF name, which must be
+// canonical, into a TLF handle, and failing if there's an error.
```
songgao: shouldn't it be "and fail" or just "failing"

go/kbfs/tlfhandle/handle.go
```diff
@@ -48,15 +48,32 @@ type TlfHandle struct {
...
+	name tlf.CanonicalName, tlfID tlf.ID) *Handle {
+	return &Handle{
+		tlfType:           ty,
+		resolvedWriters:   resolvedWriters,
+		unresolvedWriters: unresolvedWriters,
+		unresolvedReaders: unresolvedReaders,
+		name:              name,
+		tlfID:             tlfID,
+	}
+}
```
songgao: No `resolvedReaders` here. Is it because it's irrelevant now that we are all iteam based?

strib: Frankly it's because, in the old code, we didn't have any users of the constructor in tests that populated that field, so I didn't bother making a field for it.  If it bums you out, I'm happy to add it.

songgao: No worries! I was mostly just curious. This looks good!

PR #16595, opened 2019-03-15
ellipsize full user names in follow tabs

zanderz:
picked 80 pixels for cutting off full name - note keybase name can still be longer
image


songgao:
Could adding a padding in the container box achieve the same result? Also do we want this on mobile too?

zanderz:
without `whiteSpace: 'nowrap'`? I don't think so

songgao:
No, I meant without `width: 80`. This way the width limit would be connected to the actual width of the container.

zanderz:
Padding did not work. Note the avatar is hardcoded to 64 anyway.

PR #16597, opened 2019-03-15
add padding to mobile message input if it's height has expanded

buoyad:
Seems like the simplest way to do this without having to rework the styling of this component. r? @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/input-area/normal/platform-input.native.js
+++ b/shared/chat/conversation/input-area/normal/platform-input.native.js
@@ -20,8 +20,12 @@ import AddSuggestors, {standardTransformer} from '../suggestors'
 type menuType = 'exploding' | 'filepickerpopup'
 
 type State = {
+  addBottomPadding: boolean,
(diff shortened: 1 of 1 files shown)


shared/chat/conversation/input-area/normal/platform-input.native.js
```diff
@@ -124,7 +129,15 @@ class _PlatformInput extends PureComponent<PlatformInputPropsInternal, State> {
     nativeEvent: {
       layout: {x, y, width, height},
     },
-  }) => this.props.setHeight(height)
+  }) => {
+    if (minimumInputHeight < 0 || height < minimumInputHeight) {
+      minimumInputHeight = height
+    }
+    if (this.state.addBottomPadding ? height <= minimumInputHeight : height > minimumInputHeight) {
+      this.setState({addBottomPadding: !this.state.addBottomPadding})
```
chrisnojima: should do the check to bail if its already set

PR #16599, opened 2019-03-15
downloadAttachment not offlineable

mlsteele:
Skip offlineable handling in `DownloadAttachmentLocal` and `DownloadFileAttachmentLocal`. These are called as a result of a user clicking an attachment and there's no gui support for `.offline=true`. Offlineable errors followed by a successful 'reconnect' were getting effectively ignored.

Now if you're actually offline it shows a zero progress bar for ~35 seconds and then this. Could do more work to make it nicer. Better than "Show in Finder" to 0 byte file
image


mmaxim:
This one should probably get merged into the PR with the JS changes.

mlsteele:
Yep that's its merge base

mmaxim:
Ok just merge it in there then.

PR #16601, opened 2019-03-15
add the ability to load a thread centered at a give message ID (desktop)

mmaxim:
Patch does the following:

This patch is desktop specific, it shouldn't change anything about mobile (I'll do that later, and it should be easier anyway)

1. Adds new options for loading from a thread in `GetThreadNonblock`.
2. Adds new actions for loading newer messages, and loading messages around a search hit.
3. Change how the message thread is kept locked to bottom, using the new `onRequestScrollToBottom` plumbing.
4. Add the ability to center a message thread on a particular ordinal.
5. Add new state to the store indicating a thread has been loaded to the most recent message
 - Toss incoming message events for conversations not loaded to max.
6. Add option to scroll into the future of threads if the scroll position is on the bottom, and we don't have the latest message.


mmaxim:
I used this to test it:

DEBUGStore.dispatch({type: 'chat2:loadMessagesFromSearchHit', payload: {conversationIDKey: "00005a00c5758b515e9e572cc5dba194d01e4e42a4d79900b0b9b19c85fd8cbb", messageID: 100}})

mmaxim:
I just merged in support for the "Jump to Recent" button into this PR, since I also fixed a few things while doing it.

mmaxim:
Bump on this guy, the other PR with the storybook is actually on top of this branch (although I guess it doesn't really need to be)

shared/chat/conversation/list-area/normal/index.desktop.js
```diff
@@ -30,16 +31,18 @@ const ordinalsInAWaypoint = 10
...
+
 type State = {
-  isLockedToBottom: boolean,
+  lockedToBottom: boolean,
 }
 
 type Snapshot = ?number
 
-const debug = __STORYBOOK__
+const debug = __STORYBOOK__ || true
```
chrisnojima: should probably remove

shared/chat/conversation/list-area/normal/jumptorecent.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+
+import * as React from 'react'
```
chrisnojima: by convention this should be named jump-to-recent.js

shared/constants/chat2/message.js
```diff
@@ -17,6 +17,7 @@ import {isMobile} from '../platform'
 import type {TypedState} from '../reducer'
 import {noConversationIDKey} from '../types/chat2/common'
 import logger from '../../logger'
+import {getMeta} from './meta'
```
chrisnojima: unused?

shared/reducers/chat2.js
```diff
@@ -735,15 +750,46 @@ const rootReducer = (
...
+          const conversationIDKey = Types.stringToConversationIDKey(cid)
+          if (map.get(conversationIDKey, false)) {
+            return
+          }
+          const meta = state.metaMap.get(conversationIDKey, null)
+          const ordinals = messageOrdinals.get(conversationIDKey, I.OrderedSet()).toArray()
+          let maxMsgID = 0
+          for (let i = ordinals.length - 1; i >= 0; i--) {
+            const ordinal = ordinals[i]
+            const message = messageMap.getIn([conversationIDKey, ordinal])
```
chrisnojima: might want to pull out messageMap.get(conversationIDKey) outside the loop so we don't have to do this map lookup each time in here. likely doesn't matter but doesn't really add anything to not

shared/reducers/chat2.js
```diff
@@ -735,15 +750,46 @@ const rootReducer = (
...
+        s.set('containsLatestMessageMap', containsLatestMessageMap)
         // only if different
         if (!state.messageOrdinals.equals(messageOrdinals)) {
           s.set('messageOrdinals', messageOrdinals)
         }
         s.set('pendingOutboxToOrdinal', pendingOutboxToOrdinal)
       })
     }
+    case Chat2Gen.jumpToRecent:
+      return state.setIn(['containsLatestMessageMap', action.payload.conversationIDKey], true)
```
chrisnojima: not sure this is true

shared/chat/conversation/list-area/normal/index.desktop.js
mmaxim: Yeah forgot about it

shared/reducers/chat2.js
mmaxim: All this action does it load the most recent messages, so it has to be true.

shared/chat/conversation/list-area/normal/index.desktop.js
```diff
@@ -173,27 +219,60 @@ class Thread extends React.PureComponent<Props, State> {
...
+      // maintain scroll to bottom?
+      this._scrollToBottom('componentDidUpdate-maintain-scroll')
+    }
+
+    // Check if we just added new messages from the future. In this case, we don't want to adjust scroll
+    // position at all, so just bail out if we detect this case.
+    if (
+      this.props.messageOrdinals.size !== prevProps.messageOrdinals.size &&
+      this.props.messageOrdinals.first() === prevProps.messageOrdinals.first() &&
+      !this.props.containsLatestMessage
```
buoyad: I don't think `&& !this.props.containsLatestMessage` should be included. As it is scrolling down to the most recent page can jump you to the bottom

shared/actions/chat2/index.js
```diff
@@ -933,6 +958,26 @@ function* loadMoreMessages(state, action) {
...
+      key = action.payload.conversationIDKey
+      reason = 'scroll forward'
+      break
+    case Chat2Gen.loadMessagesFromSearchHit:
+      key = action.payload.conversationIDKey
+      reason = 'search hit'
+      messageIDControl = {
+        mode: RPCChatTypes.localMessageIDControlMode.centered,
+        num: Constants.numMessagesOnInitialLoad,
+        pivot: action.payload.messageID,
```
buoyad: Would be nice not to have to reload everything if the ID is already in the messageMap. Guess it's not easy to find that out here, though.

shared/chat/conversation/list-area/normal/container.js
```diff
@@ -32,29 +37,52 @@ const mapStateToProps = (state, {conversationIDKey}: OwnProps) => {
...
 })
 
 // We load the first thread automatically so in order to mark it read
 // we send an action on the first mount once
 let markedInitiallyLoaded = false
 
+const loadMoreMessages = (state, props, loadFn) => ordinal => {
+  if (state._conversationIDKey === props.conversationIDKey) {
+    if (state._lastLoadMoreOrdinalTime + 1000 > Date.now()) {
+      // ignore a load if its too recent for the same ordinal
```
buoyad: so should there be `&& ordinal === state.lastLoadMoreOrdinal`? also can you combine these `if`s?

shared/chat/conversation/list-area/normal/index.desktop.js
```diff
@@ -173,27 +219,60 @@ class Thread extends React.PureComponent<Props, State> {
...
+      // do nothing do scroll position if this is true
+      this._scrollHeight = 0 // setting this causes us to skip next resize
+      return
+    }
+
+    // Check to see if we jumped to some other set of ordinals, or that our centered ordinal is different
+    // than what we scroll to previously
+    if (
+      !!this.props.centeredOrdinal &&
+      (this.props.centeredOrdinal !== this._lastCenteredOrdinal ||
```
buoyad: any reason not to check against `prevProps.centeredOrdinal`? then you don't need `this._lastCenteredOrdinal`

shared/chat/conversation/list-area/normal/jump-to-recent.js
```diff
@@ -0,0 +1,59 @@
...
+      display: 'inline',
+    },
+  }),
+  arrowText: {
+    color: Styles.globalColors.white,
+  },
+  container: {
+    backgroundColor: Styles.globalColors.blue,
+    borderRadius: 28,
+    paddingBottom: Styles.globalMargins.tiny,
```
buoyad: 🐑 you can use `...Styles.padding(Styles.globalMargins.tiny, Styles.globalMargins.small)` instead of these `padding*`

shared/actions/chat2/index.js
mmaxim: I just did it like this to make it simpler, thought I could revisit later if it is a problem. In practice, it doesn't seem to be too bad and isn't really noticable.

shared/chat/conversation/list-area/normal/container.js
mmaxim: I just moved this around, I didn't really change anything here.

shared/chat/conversation/list-area/normal/index.desktop.js
mmaxim: Yeah I can't remember why I put that in, and agree it works better without.

shared/chat/conversation/list-area/normal/index.desktop.js
mmaxim: I ran into a problem where the centeredOrdinal would change, but the ordinals list was the same and nothing would happen. In other words, there were two updates, one for the centered ordinal, and another for the new list of messages. I wanted both to be in place before doing the scroll.

mmaxim:
@buoyad PTAL

shared/chat/conversation/list-area/normal/jump-to-recent.js
mmaxim: I made this change, but I wonder if it is even an improvement? It is the same number of line, but now I need to know what each parameter is.

mmaxim: Oh wait I see how to use it now.

PR #16608, opened 2019-03-15
"Jump to unread messages" shortcut in inbox

buoyad:
If there are unread messages **below** your current position in the inbox, this renders an orange banner at the bottom that you can tap to scroll down to them. Desktop + mobile. r? @keybase/react-hackers


buoyad:
This animates the scroll on mobile but not on desktop. I can get a PR up to animate the desktop side after this.

buoyad:
rebased to pick up #16613 and tweaked the scrollTo params

chrisnojima:
code lgtm but it doesn't handle the case of a new message coming in and triggering it. you have to scroll for it to show up. aka i get a notification for a message yet no banner. scroll a pixel and it shows

PR #16561, opened 2019-03-13
RN 0.59.8

chrisnojima:
- [x] waiting on app bundle changes from @MarcoPolo 
- [ ] also evaluate text input android fixes
- [x] diff purge
- [x] ios dev
- [x] ios profile
- [x] webpack dev
- [x] webpack prod
- [x] android dev
- [x] android maven updates
- [x] storybook
- [x] ios storybook
- [x] android storybook
- [x] bot deskto build
- [x] split android apk
- [x] android prod
- [x] ios prod
- [x] pull some code changes back into master (video, require loops)

PR #16610, opened 2019-03-15
Random fixes from RN 0.59 branch

chrisnojima:
while making rn 59 i came across these fixes

https://github.com/keybase/client/pull/16561


.circleci/config.yml
```diff
@@ -41,7 +41,6 @@ jobs:
         working_directory: shared
         environment:
           REACT_NATIVE_MAX_WORKERS: 1
-          BABEL_PLATFORM: ReactNative
```
chrisnojima: found a way to not need this!

shared/constants/chat2/message.js
```diff
@@ -356,7 +356,7 @@ const makeMessageSystemChangeRetention: I.RecordFactory<MessageTypes._MessageSys
     isInherit: false,
     isTeam: false,
     membersType: 0,
-    policy: RPCChatTypes.RetentionPolicy,
```
chrisnojima: this isn't a value so it was null!

chrisnojima:
@keybase/react-hackers

shared/constants/chat2/message.js
```diff
@@ -356,7 +356,7 @@ const makeMessageSystemChangeRetention: I.RecordFactory<MessageTypes._MessageSys
...
+    policy: RPCChatTypes.commonRetentionPolicyType.none,
```
buoyad: This is typed as
```
export type RetentionPolicy = {typ: 1, retain: ?RpRetain} | {typ: 2, expire: ?RpExpire} | {typ: 3, inherit: ?RpInherit} | {typ: 4, ephemeral: ?RpEphemeral}
```
it's probably better to make this `{typ: RPCChatTypes.commonRetentionPolicyType.none}`. I think this should be changed to the UI type but that doesn't need to be done here.

buoyad: `canChange` in this object actually isn't in `_MessageSystemChangeRetention` either. TIL flow doesn't check these.

.circleci/config.yml
Ikpeogu: 359

Ikpeogu: How it works

Ikpeogu: Okay nice

Ikpeogu (commented):
Wow

Ikpeogu (commented):
It
Duplicate of #

shared/constants/chat2/message.js
Ikpeogu: https://user-images.githubusercontent.com/86695676/123932968-da3b5300-d989-11eb-8b55-eef21c1c964e.mp4

Ikpeogu: Is that good

PR #16611, opened 2019-03-15
Fs: auto-refresh text preview when content changes

songgao:


go/kbfs/simplefs/simplefs.go
```diff
@@ -587,42 +586,49 @@ func (k *SimpleFS) refreshSubscription(
 		return nil
 	}
 
+	tlfType, tlfName, _, _, err := remoteTlfAndPath(path)
+	if err != nil {
+		return err
+	}
+	tlfPathFromGUI := stdpath.Join("/keybase", tlfType.PathString(), tlfName)
```
songgao: This is necessary now because links in chat means user can go to a path inside a TLF without going through its root. So the first path SimpleFS sees about a TLF isn't necessarily the TLF root.

songgao: This is necessary now because links in chat means user can go to a path inside a TLF without going through its root. So the first path SimpleFS sees about a TLF isn't necessarily the TLF root.

strib: You could use `libkbfs.BuildCanonicalPath` for this (or, soon, `tlfhandle.BuildCanonicalPathForTlfName`).

shared/fs/filepreview/view.js
```diff
@@ -0,0 +1,128 @@
...
+  type: Types.PathType,
+  routePath: I.List<string>,
+  url: string,
+|}
+
+type State = {|
+  loadedLastModifiedTimestamp: number,
+|}
+
+export default class extends React.PureComponent<Props, State> {
```
chrisnojima: should name this class

shared/fs/filepreview/view.js
```diff
@@ -0,0 +1,128 @@
...
+  constructor(props: Props) {
+    super(props)
+    this.state = {
```
chrisnojima: this is likely not what you want. if the other props change you also likely want to reset this etc (like if the file changes). you probably want getDerivedStateFromProps or better componentDidUpdate and reset this if any of the actual data changes (at least path)

go/kbfs/simplefs/simplefs.go
songgao: This has to be the path that GUI tracks for notifications to work properly, so it might not necessarily be canonical path. I guess it'd still work with the function, just feels weird (confusing?) to pass in a non-canonical name as `tlf.CanonicalName`.

songgao: Would you rather I use `tlfHandle.BUildCanonicalPath` and pass in the non-canonical TLF name? This one accepts strings rather than tlf.CanonicalName.

strib: `tlfhandle.BuildCanonicalPath` is fine with me if you prefer it. The "canonical" here refers to the formatting of `/keybase/type/something`, not that the tlf name itself is canonical, so it seems fine to me.  But feel free to rename it or add to the comments as needed.

songgao:
@strib @chrisnojima ready for another look!

go/kbfs/tlfhandle/path.go
```diff
@@ -47,7 +48,9 @@ func BuildCanonicalPath(pathType PathType, paths ...string) string {
 	return strings.Join(pathElements, "/")
 }
 
-func buildCanonicalPathForTlfType(t tlf.Type, paths ...string) string {
+// BuildCanonicalPathForTlfType is like BuildCanonicalPath, but accepts a
+// tlf.Type instead of TlfType.
```
strib: "TlfType" -> "PathType".

songgao: 🤦🏻‍♂️will fix!

go/kbfs/tlfhandle/path.go
```diff
@@ -47,7 +48,9 @@ func BuildCanonicalPath(pathType PathType, paths ...string) string {
 	return strings.Join(pathElements, "/")
 }
 
-func buildCanonicalPathForTlfType(t tlf.Type, paths ...string) string {
+// BuildCanonicalPathForTlfType is like BuildCanonicalPath, but accepts a
+// tlf.Type instead of PathhType.
```
strib: "PathhType" -> "PathType"!

PR #16612, opened 2019-03-16
libkbfs: extract key server and key metadata logic into new pkg

strib:
Depends on #16594.


go/kbfs/libkey/key_server_local.go
```diff
@@ -174,10 +200,11 @@ func (ks *KeyServerLocal) DeleteTLFCryptKeyServerHalf(ctx context.Context,
 	return ks.db.Delete(serverHalfID.ID.Bytes(), nil)
 }
 
-// Copies a key server but swaps the config.
-func (ks *KeyServerLocal) copy(config Config) *KeyServerLocal {
-	return &KeyServerLocal{config, ks.db, config.MakeLogger(""),
-		ks.shutdownLock, ks.shutdown, ks.shutdownFunc}
+// Copy copies a key server but swaps the config.
+func (ks *KeyServerLocal) Copy(
+	config KeyServerConfig, log logger.Logger) *KeyServerLocal {
```
songgao: Looks like it swaps the logger too now. If it's intended, perhaps should fix the comment as well? Also, might be more clear to rename the method as `WithConfigAndLogger`.

strib: The old code made an entirely new logger, so it still swapped it, but sure I'll make it explicit and change the name.

PR #16615, opened 2019-03-16
highlight first search result

zanderz:
unless moused over or the current one is in the list

### DIFF
--- a/shared/chat/inbox/index.desktop.js
+++ b/shared/chat/inbox/index.desktop.js
@@ -19,13 +19,11 @@ import {virtualListMarks} from '../../local-debug'
 import {inboxWidth, getRowHeight} from './row/sizes'
 
 type State = {
-  isHovered: boolean,
   showFloating: boolean,
 }
 
 class Inbox extends React.PureComponent<Props, State> {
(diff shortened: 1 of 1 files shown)


shared/chat/inbox/index.desktop.js
```diff
@@ -18,17 +19,23 @@ import {virtualListMarks} from '../../local-debug'
 import {inboxWidth, getRowHeight} from './row/sizes'
 
 type State = {
+  isHovered: boolean,
   showFloating: boolean,
 }
 
 class Inbox extends React.PureComponent<Props, State> {
   state = {
+    isHovered: false,
```
buoyad: This being in `state` makes it kinda hard to use the gear menu:
team-menu
so either that should be fixed or maybe this can move out of `state` and be on the class.

zanderz: thanks for the suggestion, seemed to help.
:white_check_mark:

zanderz:
I never thought of it that way. Maybe there should be 2 different colors? (kidding!) Maybe I'll take it out and see if someone complains.

PR #16624, opened 2019-03-18
dont keep log archives for nonadmins on mobile

heronhaye:
I thought "RotateLogFile" was confusing because it just set the configuration and didn't actually rotate anything so I removed it


go/libkb/env.go
```diff
@@ -1774,3 +1793,24 @@ func (e *Env) AllowPTrace() bool {
...
 }
+
+func (e *Env) GetLogFileConfig(filename string) *logger.LogFileConfig {
+	megabyte := int64(1024 * 1024)
+	var maxKeepFiles int
+	var maxSize int64
+
+	if e.GetAppType() == MobileAppType && !e.GetFeatureFlags().Admin() {
+		maxKeepFiles = 1
+		maxSize = 32 * megabyte
```
heronhaye: too small?

heronhaye:
(Failed because of kbfs flake)

go/libkb/env.go
joshblum: I think this is fine since we only send 10mb in log sends right now. Can you add a comment to both places in case we bump the log send or drop this size? https://github.com/keybase/client/blob/master/go/bind/keybase.go#L278

go/libkb/env.go
```diff
@@ -1774,3 +1793,24 @@ func (e *Env) AllowPTrace() bool {
 		func() (bool, bool) { return e.getEnvBool("KEYBASE_ALLOW_PTRACE") },
 	)
 }
+
+func (e *Env) GetLogFileConfig(filename string) *logger.LogFileConfig {
+	megabyte := int64(1024 * 1024)
```
joshblum: could use `opt.MiB`

heronhaye:
Didn't test it, would be great if you could.  Thanks

joshblum:
addresses https://github.com/keybase/keybase-issues/issues/3359

joshblum:
tested on ios with admin/nonadmin account, config changes look good

go/bind/keybase.go
```diff
@@ -214,6 +214,8 @@ func Init(homeDir, mobileSharedHome, logFile, runModeStr string,
 		EK:      config.GetEKLogFile(),
 	}
 
+	fmt.Printf("Go: Using config: %+v\n", kbCtx.Env.GetLogFileConfig(config.GetLogFile()))
```
heronhaye: rm?

joshblum: it only shows up in the xcode console as far as i know, figured it was worth leaving in case someone was wondering what happened to other log files logged in as a test user

joshblum: (we log the logfile paths above that)

PR #16625, opened 2019-03-18
Remove blocked convs on incremental sync

mlsteele:
Blocked conversations were showing up unexpectedly. In the following scenario: Devices X and Y have the inbox loaded. Device X goes offline. Device Y blocks a conversation. Device X comes online. An incremental sync would occur that pushed the blocked conv to the gui.

This patch makes it so that in an incremental sync, blocked convs are omitted from `items` and get put into a `removals` list. The gui deletes those convs from the metamap.


chrisnojima:
can't we just do a full sync? blocking barely happens so i'm not worried about the perf. diff should be a lot smaller

mlsteele:
That would lead to full refreshes if some conv is stuck the removal list on every sync. Which could be considered a bug, but pretty hard to rule out, and not necessarily even a bug since service and gui sync different sets of convs. Does a full sync cause the screen to wipe? Anyway this should be robust against all that

zanderz:
I tried to test it, and it seemed to make a difference, but not sure I duplicated the intended case.

mlsteele:
good?

shared/actions/chat2/index.js
```diff
@@ -594,9 +594,10 @@ const onChatInboxSynced = (state, action) => {
         }
         return arr
       }, [])
+      const removals = syncRes.incremental && syncRes.incremental.removals || []
```
buoyad: `syncRes.incremental?.removals || []`

shared/actions/json/chat2.json
```diff
@@ -115,6 +115,7 @@
     // Got some new inbox rows
     "metasReceived": {
       "metas": "Array<Types.ConversationMeta>",
+      "removals?": "Array<string>", // convIDs to remove
```
buoyad: This should be `Types.ConversationIDKey` and converted in the saga instead

PR #16627, opened 2019-03-18
libgit: register repo nodes for updates on branches

strib:
If the user _only_ looked at a non-master branch, we were not registering with the `AutogitManager` to get updates for the repository's root node.  The `AutogitManager` already makes sure to only register with KBFS once for each TLF, so it's safe to register with the manager for each different branch.


strib:
> So this was causing autogit not to receive updates on non-master branches when there wasn't also a master branch autogit that had been loaded?

Yep.

PR #16629, opened 2019-03-18
Revert "add padding to mobile message input if it's height has expanded"

buoyad:
This doesn't work on android. Reverts keybase/client#16597 @keybase/react-hackers


buoyad:
@cjb good question. I thought it was the `setState`/rerender that caused it but you're right, changing the style is what breaks it.

PR #16631, opened 2019-03-18
folder_branch_ops: return real error when stating a fake file

strib:
Also make sure FUSE gives the right error code when a file doesn't exist.


go/kbfs/libfuse/errors.go
```diff
@@ -75,5 +76,10 @@ func filterError(err error) error {
 	case libkbfs.RevGarbageCollectedError:
 		return errorWithErrno{err, syscall.ENOENT}
 	}
+
+	if os.IsNotExist(errors.Cause(err)) {
```
songgao: Do we wanna add `libkbfs. NoSuchNameError` above too just in case?

strib: libfuse/libdokan/libfs all handle that specifically right now, but sure I guess it doesn't hurt to add this here too.

PR #16639, opened 2019-03-19
fix attaching from Files to chat

songgao:
This stat call is causing attaching from Files to chat to fail if KBFS mount is not present -- which is always the case with mobile. Not sure how I missed that in the first place. I tested attaching on mobile and it worked. Perhaps it has to do with the file size and didn't go through this routine if file is smaller or larger. Anyway, this PR fixes that by using SimpleFSStat for KBFS paths.

I grep'ed and didn't find any other `os.` calls in chat/attachments that could encounter KBFS paths, so I think this is the only one. But let me know if you think of anything else!

### DIFF
--- a/go/chat/attachments/utils.go
+++ b/go/chat/attachments/utils.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 	"sync/atomic"
+	"time"
 
 	"github.com/keybase/client/go/chat/globals"
 	"github.com/keybase/client/go/libkb"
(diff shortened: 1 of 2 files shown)


go/chat/attachments/utils.go
```diff
@@ -224,12 +235,65 @@ func (f *KbfsReadCloseResetter) Close() error {
 // or SimpleFS depending on if p is a KBFS path.
 func NewReadCloseResetter(ctx context.Context, g *libkb.GlobalContext,
 	p string) (ReadCloseResetter, error) {
+	fmt.Printf("SONGGAO-NewReadCloseResetter %q %v\n", p, isKbfsPath(p))
```
joshblum: rm?

PR #16644, opened 2019-03-19
mark viewed on people tab switch

chrisnojima:
support nav2 in people tab
@keybase/react-hackers


chrisnojima:
@keybase/react-hackers bump

shared/router-v2/tab-bar/container.js
```diff
@@ -16,9 +17,12 @@ const mapStateToProps = state => ({
   username: state.config.username,
 })
 
-const mapDispatchToProps = dispatch => ({
+const mapDispatchToProps = (dispatch, ownProps) => ({
   _onProfileClick: username => dispatch(ProfileGen.createShowUserProfile({username})),
   _onTabClick: tab => {
```
buoyad: does this mean that if you leave the people tab via, say, the chat button on someone's profile, the badge won't be cleared?

buoyad: I think this is the case but don't want to change the UX since I assume this was discussed. @keybase/react-hackers this is ready

PR #16645, opened 2019-03-19
Picker icons for all types

mlsteele:


mlsteele:
cc @buoyad picker icons merged

PR #16646, opened 2019-03-19
Desktop-only: only commit to coinflips when awake+unlocked

mlsteele:
There have been reports of non-mobile devices causing flips to fail. Probably due to sending commitments while asleep. I wasn't able to repro using my laptop. It could require a long sleep or bad karma.

This patch disables sending commitments to coinflips when locked or asleep according to electron. Hopefully that fixes the problem. If it does not, the logs should tell us how the powermonitor relates to that.

Rename AppState to MobileAppState all over the place. So that DesktopAppState can sit alongside it. Fusing MobileAppState and DesktopAppState is not desirable since the states and meanings of everything to do with power management are so different on desktop and mobile.

Existing usages of MobileAppState (was AppState) looked like they didn't check the device type. But on desktop the AppState would just stay at `FOREGROUND` the whole time. That's how I left it. Both states are available but using the wrong one will just result in what acts like perpetual awakeness.


go/chat/flip/dealer.go
```diff
@@ -730,7 +730,9 @@ func (d *Dealer) handleMessageStart(ctx context.Context, msg *GameMessageWrapped
 	// with our commitment. We are now in the inner loop of the Dealer, so we
 	// have to do this send in a Go-routine, so as not to deadlock the Dealer.
 	if !isLeader {
-		go d.sendCommitment(ctx, md, me)
+		if d.dh.ShouldCommit(ctx) {
```
maxtaco: @mlsteele will we get warnings in the log about the fact that we weren't included in the flip but we should have been?

maxtaco: See

```go
if !g.iWasIncluded {
		g.clogf(ctx, "The leader didn't include me (%s) so not sending a reveal (%s)", g.me.me, g.md)
		return nil
	}
```

in `maybeReveal`

mlsteele: Yep we do get those lines. One for each `++Chat: | FlipManager: ShouldCommit -> false`. I'll look at filtering them out

PR #16647, opened 2019-03-19
Prime secret store before provisioning

zapu:
This PR adds a `PrimeSecretStores` function that check if the secret store is functioning - that is, it tries to store, retrieve, and then cleanup a random secret for randomly generated username.

Calls to this new functions were inserted in `loginProvision`, `SelfProvisionEngine`, and `SignupEngine`. The engines are stopped early if they discover that the secret store does not work. Normal login (`LoginProvisionedDevice`) is not protected, so it will still work, just not remember user's passphrase in case we can't use secret store.

This is important for NOPW sign up, in which we would sign someone up, (silently) fail to store their passphrase stream, and then they would be out of luck after returning to the app, unable to login again.

I tested this manually using GUI, and what happens is that signup fail without creating the user. So they try to fix their device or try another one, without "burning" that username.

### DIFF
--- a/go/engine/login_provision.go
+++ b/go/engine/login_provision.go
@@ -90,6 +90,10 @@ func (e *loginProvision) Run(m libkb.MetaContext) error {
 		return err
 	}
 
+	if err := m.G().SecretStore().PrimeSecretStores(m); err != nil {
+		return fmt.Errorf("secret store is not functional: %s", err)
+	}
+
 	var err error
 	e.perUserKeyring, err = libkb.NewPerUserKeyring(m.G(), e.arg.User.GetUID())
 	if err != nil {
--- a/go/engine/login_provision.go
+++ b/go/engine/login_provision.go
@@ -91,7 +91,7 @@ func (e *loginProvision) Run(m libkb.MetaContext) error {
 	}
 
 	if err := m.G().SecretStore().PrimeSecretStores(m); err != nil {
-		return fmt.Errorf("secret store is not functional: %s", err)
+		return SecretStoreNotFunctionalError{err}
 	}
 
 	var err error
--- a/go/engine/signup_test.go
+++ b/go/engine/signup_test.go
@@ -5,6 +5,7 @@ package engine
 
 import (
 	"fmt"
+	"runtime"
 	"testing"
 
 	"github.com/keybase/client/go/libkb"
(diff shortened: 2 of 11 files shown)


go/libkb/secret_store.go
```diff
@@ -246,3 +248,74 @@ func (s *SecretStoreLocked) GetUsersWithStoredSecrets(m MetaContext) ([]string,
...
+	} else if err != nil {
+		mctx.Debug("PrimeSecretStore: error when retrieving secret that wasn't stored yet: %q, as expected", err)
+	}
+
+	// Put secret in secret store through `SecretStore` interface.
+	err = ss.StoreSecret(mctx, testNormUsername, testSecret)
+	if err != nil {
+		return err
+	}
+	defer func() {
```
joshblum: I think you should move this up in the function, it's possible the `StoreSecret` returns an error but there is data written that we want to clear

go/engine/login_provision.go
```diff
@@ -90,6 +90,10 @@ func (e *loginProvision) Run(m libkb.MetaContext) error {
 		return err
 	}
 
+	if err := m.G().SecretStore().PrimeSecretStores(m); err != nil {
+		return fmt.Errorf("secret store is not functional: %s", err)
```
joshblum: can you wrap this in a `SecretStoreNotFunctionalErr` type?

go/libkb/secret_store.go
```diff
@@ -246,3 +249,80 @@ func (s *SecretStoreLocked) GetUsersWithStoredSecrets(m MetaContext) ([]string,
...
+
+// PrimeSecretStore runs a test with current platform's secret store, trying to
+// store, retrieve, and then delete a secret with an arbitrary name. This should
+// be done before provisioning or logging in
+func PrimeSecretStore(mctx MetaContext, ss SecretStoreAll) (err error) {
+	defer mctx.TraceTimed("PrimeSecretStore", func() error { return err })()
+
+	// Generate test username and test secret
+	testUsername, err := RandString("test_ss_", 5)
+	testUsername = strings.Replace(testUsername, "=", "_", -1)
```
joshblum: Curious why this is necessary? Did something fail without it?

go/libkb/secret_store.go
```diff
@@ -246,3 +249,80 @@ func (s *SecretStoreLocked) GetUsersWithStoredSecrets(m MetaContext) ([]string,
...
+	if err != nil {
+		return fmt.Errorf("error while retrieving secret: %s", err)
+	}
+	mctx.Debug("PrimeSecretStore: Retrieved secret: %v", retrSecret.f)
+	if !retrSecret.Equal(testSecret) {
+		return errors.New("managed to retrieve test secret but it didn't match the stored one")
+	}
+
+	mctx.Debug("PrimeSecretStore: success - trying to clear test entry")
+	shouldClear = false
```
joshblum: why not have `shouldClear` determine if `err = err2` in the defer, then `ClearSecret` only has to get called once

go/libkb/secret_store_file_test.go
```diff
@@ -218,3 +219,42 @@ func TestSecretStoreFileNoise(t *testing.T) {
...
+	tc := SetupTest(t, "secret_store_file", 1)
+	defer tc.Cleanup()
+
+	mctx := NewMetaContextForTest(tc)
+	secretStore := NewSecretStoreFile(td)
+	err := PrimeSecretStore(mctx, secretStore)
+	require.NoError(t, err)
+}
+
+func TestPrimeSecretStoreFileFail(t *testing.T) {
```
joshblum: why not have this test instead as part of a signup/provisioning test to verify the user creation short circuits

go/libkb/secret_store.go
zapu: let me know if this is better now, thank you!

go/libkb/secret_store_file_test.go
patrickxb: sounds like a good idea...

zapu: working on this right now

zapu: thanks!

zapu:
I've addressed the comments and I've done some additional manual testing - this is ready for another look. Thank you!

go/engine/signup_test.go
```diff
@@ -321,3 +322,58 @@ func TestSignupNonAsciiDeviceName(t *testing.T) {
...
 	}
 }
+
+func TestSignupNOPWBadParams(t *testing.T) {
+	tc := SetupEngineTest(t, "signup_nopw")
+	defer tc.Cleanup()
+
+	fu, _ := NewFakeUser("sup")
+	arg := MakeTestSignupEngineRunArg(fu)
+	arg.StoreSecret = false
```
patrickxb: For the record, I believe the `StoreSecret` field is from ancient times where we allowed it to be off.  It should always be true.

PR #16649, opened 2019-03-19
Fix nav2 on Linux

cjb:
@keybase/react-hackers 

Huh, the question here is actually why nav2 *was* working on macOS.  We set the nav area to draggable, which means we need to counter-set `WebkitAppRegion: 'no-drag' (via `desktopStyles.windowDraggingClickable`) in clickable areas, but that wasn't happening at all.  So the code intent was for nothing to be clickable in the nav area.

### DIFF
--- a/shared/profile/search/bar.js
+++ b/shared/profile/search/bar.js
@@ -60,6 +60,7 @@ const styles = Styles.styleSheetCreate({
     isElectron: {
       ...Styles.desktopStyles.boxShadow,
       ...Styles.globalStyles.flexBoxColumn,
+      ...Styles.desktopStyles.windowDraggingClickable,
       ...(flags.useNewRouter ? {marginRight: 12} : {}),
       alignSelf: 'center',
       borderRadius: 5,
@@ -78,6 +79,7 @@ const styles = Styles.styleSheetCreate({
     },
     isElectron: {
       ...Styles.desktopStyles.clickable,
+      ...Styles.desktopStyles.windowDraggingClickable,
       height: 24,
       marginLeft: flags.useNewRouter ? 'auto' : Styles.globalMargins.small,
       marginRight: flags.useNewRouter ? Styles.globalMargins.xsmall : Styles.globalMargins.small,
--- a/shared/profile/search/bar.js
+++ b/shared/profile/search/bar.js
@@ -60,7 +60,6 @@ const styles = Styles.styleSheetCreate({
     isElectron: {
       ...Styles.desktopStyles.boxShadow,
       ...Styles.globalStyles.flexBoxColumn,
-      ...Styles.desktopStyles.windowDraggingClickable,
       ...(flags.useNewRouter ? {marginRight: 12} : {}),
       alignSelf: 'center',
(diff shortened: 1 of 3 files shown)


shared/router-v2/router.desktop.js
```diff
@@ -262,11 +262,6 @@ const createElectronApp = Component => {
 const ElectronApp = createElectronApp(RootStackNavigator)
 
 const styles = Styles.styleSheetCreate({
```
cjb: This style fixes the problem and looks like it was intended for the back button, but this `styles.back` is unused anywhere.

shared/profile/search/bar.js
```diff
@@ -78,6 +79,7 @@ const styles = Styles.styleSheetCreate({
     },
     isElectron: {
       ...Styles.desktopStyles.clickable,
+      ...Styles.desktopStyles.windowDraggingClickable,
```
cjb: Should we consider having `desktopStyles.clickable` just imply `desktopStyles.windowDraggingClickable` directly?

shared/profile/search/bar.js
```diff
@@ -60,6 +60,7 @@ const styles = Styles.styleSheetCreate({
     isElectron: {
       ...Styles.desktopStyles.boxShadow,
       ...Styles.globalStyles.flexBoxColumn,
+      ...Styles.desktopStyles.windowDraggingClickable,
```
cjb: Added it for the search bar inside the profile nav too.

shared/profile/search/bar.js
chrisnojima: hmmmm, i think there's side effects to doing it, so lets keep the changes minimal

shared/router-v2/header/index.desktop.js
```diff
@@ -83,9 +83,12 @@ const styles = Styles.styleSheetCreate({
       alignItems: 'center',
     },
   }),
-  icon: {
-    marginRight: 6,
-  },
+  icon: Styles.platformStyles({
+    isElectron: {
+      ...Styles.desktopStyles.windowDraggingClickable,
```
chrisnojima: i don't think we want the entire header to be draggable (and therefore we need to add this style to all clickable things in the header). i think we just want the top part (where the chrome used to be) to be draggable

cjb:
@chrisnojima Cool, this commit sounds like what you'd like -- the draggable area was the entire nav2 header, but now it's just goes down to the bottom of the back button.  CC @cecileboucheron

shared/router-v2/router.desktop.js
zanderz: rats, I tried to apply that to the back button but did it wrong, I didn't know about `WebkitAppRegion`, just `-webkit-app-region`

cjb:
@chrisnojima (ah, sounds like @cecileboucheron prefers having the whole area draggable, so perhaps I should revert these last two commits and go back to the original PR)

cjb:
@chrisnojima After discussing with @cecileboucheron I think this is ready to go in, we're going to have the whole header area draggable.

PR #16652, opened 2019-03-19
process gregor notifications for autoreset state 

joshblum:
depends on server changes in https://github.com/keybase/keybase/pull/3616

patch does the following:
- stores reset state from gregor in badger
- warns on the cli when a reset is in progress, the gui show a similar wanring in-app
- adds devel-only cli commands to start/cancel the reset process. (note there is a subsequent ticket to handle entering with a password fully to keep the scope of this ticket down)

in action:
```
$ kb account reset-start
$ kb chat list
▶ WARNING Uh-oh. Your account is scheduled to be reset in 3 days. Open the app to cancel. The clock is ticking!
▶ WARNING To cancel the process run `keybase account reset-cancel`
$ kb account reset-cancel
Account reset cancelled.
$ kb chat list
```


joshblum:
@mlsteele this is ready for a pass

go/engine/account_reset.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// SubConsumers returns the other UI consumers for this engine.
+func (e *AccountReset) SubConsumers() []libkb.UIConsumer {
+	return nil
+}
+
+// Run starts the engine.
+func (e *AccountReset) Run(mctx libkb.MetaContext) error {
+	return libkb.EnterResetPipeline(mctx, e.username, e.email)
```
mlsteele: why does this engine exist?

go/libkb/reset_delete.go
```diff
@@ -4,30 +4,70 @@ import (
...
-	return resetOrDeleteAccount(m, username, passphrase, "nuke")
+func EnterResetPipeline(mctx MetaContext, username, email string) (err error) {
+	defer mctx.TraceTimed("EnterResetPipeline", func() error { return err })()
+	// TODO CORE-10466, fully support the `uid` and `self` fields, we should
+	// not attempt to enter the pipeline if the user has an active device. We
+	// should establish a session with a passphrase if at is available.
+	self := mctx.HasAnySession()
+	if self { // clear out other args if we have a session.
+		username = ""
+		email = ""
```
mlsteele: This is scary to ignore non-empty arguments

go/libkb/apiarg.go
```diff
@@ -11,6 +11,7 @@ const (
 	APISessionTypeNONE     APISessionType = 0
 	APISessionTypeOPTIONAL APISessionType = 1
 	APISessionTypeREQUIRED APISessionType = 2
+	APISessionTypeNONIST   APISessionType = 3
```
mlsteele: what does NONIST mean and why is it necessary?

go/libkb/reset_delete.go
mlsteele: What do you think about 3 separate functions here and in avdl `EnterResetPipeline{Self,Username,Email}`?

go/engine/account_reset.go
joshblum: when we handle the user's password we'll need a UI registered which is why i made it an engine, let's leave for now and see if it can be removed in the next PR when i handle password stuff

go/libkb/reset_delete.go
joshblum: let's talk over the flow in person

go/libkb/apiarg.go
joshblum: this might not be necessary after the password flow is in place but I needed a away to have a NISTless session i tests since the pipeline checks the csrf token and the api lib use's a nist when available

PR #16657, opened 2019-03-19
fix TLF permission at TLF root in destination-picker

songgao:
If user has never navigated into a TLF and tries to move or copy into it, we'd think the TLF is not writeable even if user is a writer. This is because loadPathMetadata was never called. So extract the flow for that out of `fs/container.js`, into `fs/common/load-path-metadata-when-needed.js` and use it in both `fs/container.js` and `fs/destination-picker`.


shared/fs/common/load-path-metadata-when-needed.js
```diff
@@ -0,0 +1,42 @@
...
+  componentDidMount() {
+    this.props.loadPathMetadata()
+  }
+  componentDidUpdate(prevProps) {
+    if (this.props.path !== prevProps.path) {
+      this.props.loadPathMetadata()
+    }
+  }
+  render() {
+    return null
```
jzila: Wouldn't it make more sense for this to be a HOC?

songgao: I think this is cleaner and easier to type. HOCs are necessary when we need to pass props to the wrapped component. In this case, it doesn't actually provide any props to a component; all we need is to dispatch actions at the right time. We do this with `fs/row/load-files-when-needed.js` too, and use that in `fs/row/rows.js`

jzila: I prefer the HOC way since it better represents what is happening: it's triggering an event that should cause upstream props to be loaded. Whereas the current approach makes it look like it's a visible component that displays something.

But it's not a big deal.

songgao: The way I look at it is it just triggers an action that causes side effect. This action or its side effect is not owned by any components. When we are ready to move to hooks, things like this could probably just use `useEffect`.

PR #16661, opened 2019-03-20
warn on unwritable leveldb

heronhaye:
A "leveldb (but not ~/.local/share/keybase itself) owned by root" case, happened before as well. Don't know how it happened, user says they did `sudo run_keybase` and `sudo keybase ...` but those commands should exit early before writing anything to db. And even if they didn't, it would write to root's home, not the user's. /shrug

### DIFF
--- a/packaging/linux/run_keybase
+++ b/packaging/linux/run_keybase
@@ -140,8 +140,8 @@ show_cryptosquirrel() {
   [ "${KEYBASE_NO_SQUIRREL:-}" != "1" ] && cat /opt/keybase/crypto_squirrel.txt
 }
 
-warn_if_unwritable() {
-  if [ -w "$1" ]; then
+warn_if_exists_andunwritable() {
+  if ! [ -a "$1" ] || [ -w "$1" ]; then
       return
   fi
   echo "WARNING: Cannot write to $1. Did you previously run 'run_keybase' with sudo?"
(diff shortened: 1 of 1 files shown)


packaging/linux/run_keybase
```diff
@@ -140,8 +140,8 @@ show_cryptosquirrel() {
   [ "${KEYBASE_NO_SQUIRREL:-}" != "1" ] && cat /opt/keybase/crypto_squirrel.txt
 }
 
-warn_if_unwritable() {
-  if [ -w "$1" ]; then
+warn_if_exists_and_unwritable() {
+  if ! [ -a "$1" ] || [ -w "$1" ]; then
```
strib: I'm probably just dumb, but I can't find `-a` in the `man test` page, except for as the and operator.  Does it mean something else too?  From the name of the function, I would have expected `-e`.

packaging/linux/run_keybase
```diff
@@ -154,8 +154,9 @@ init() {
   logdir="${XDG_CACHE_HOME:-$HOME/.cache}/keybase"
   runtime_dir="${XDG_RUNTIME_DIR:-$HOME/.config}/keybase"
 
-  warn_if_unwritable "$logdir"
-  warn_if_unwritable "$runtime_dir"
+  warn_if_exists_and_unwritable "$logdir"
```
strib: If these things don't exist yet, should we be making sure that their parent directory is writable too, so that Keybase will be able to successfully create them when the time comes?

packaging/linux/run_keybase
heronhaye: oh, apparently it's deprecated now (https://www.tldp.org/LDP/abs/html/fto.html, though the reference they cite is not very helpful haha). I guess because it's confusing in addition to the infix -a. Switched to -e

packaging/linux/run_keybase
heronhaye: I don't think that has ever happened (~/.config or ~/.cache being unwritable). There's some bug I can't figure out that's causing Keybase to make these directories in the user's home directory when run with root which is causing the problem, that's what I wanted to address here. But I can add it too if you think it'll be helpful

strib: Meh I guess if it's never happened, it's not worth checking.  Just thought maybe in some kind of fresh environment, if `sudo run_keybase` was the first thing they did, they might have created those parent directories with bad ownership.  Anyway, I'm fine without it.

PR #16667, opened 2019-03-20
MerkleRoot don't store sigs after verify

mlsteele:
Sigs make up 2/3 of the leveldb storage of merkle roots. This patch omits sigs from storage if they have been verified. It's probably an invariant that sigs are checked before storage, but rather than proving that there's a book-keeping bool.

You mentioned a just-store-the-hash way of doing this. There's also `ProofServicesHash` and `PvlHash` accessed on `MerkleRoot` loaded from disk. I think the way to omit the final 1/3 would be to split the MerkleRoot type into two. One with all the info from the server. And one with a minimal set that can come from disk or from the first type.

### DIFF
--- a/go/libkb/interfaces.go
+++ b/go/libkb/interfaces.go
@@ -122,9 +122,6 @@ type CommandLine interface {
 	GetBool(string, bool) (bool, bool)
 }
 
-type Server interface {
-}
-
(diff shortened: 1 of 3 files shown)


mlsteele:
Don't bother looking. There's likely a simpler and more effective solution.

PR #16668, opened 2019-03-20
Un-revert buffered logging, make buffered logging optional

jakob223:
This PR enables buffered logging in KBFS without using it elsewhere.

CC @patrickxb , @zapu , @maxtaco

### DIFF
--- a/go/kbfs/kbfsfuse/main.go
+++ b/go/kbfs/kbfsfuse/main.go
@@ -97,6 +97,9 @@ func start() *libfs.Error {
 			fuseLog, false /* superVerbose */)
 	}
 
+	logger.EnableBufferedLogging()
+	defer logger.Shutdown()
+
 	options := libfuse.StartOptions{
 		KbfsParams:        *kbfsParams,
 		PlatformParams:    *platformParams,
--- a/go/logger/buffer.go
+++ b/go/logger/buffer.go
@@ -59,9 +59,10 @@ type autoFlushingBufferedWriter struct {
 	bufferedWriter *bufio.Writer
 	backupWriter   *bufio.Writer
 
-	frequency time.Duration
-	timer     *triggerableTimer
-	shutdown  chan struct{}
+	frequency    time.Duration
+	timer        *triggerableTimer
+	shutdown     chan struct{}
+	doneShutdown chan struct{}
 }
 
 var _ io.Writer = &autoFlushingBufferedWriter{}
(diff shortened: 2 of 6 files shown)


go/kbfs/kbfsfuse/main.go
```diff
@@ -97,6 +97,9 @@ func start() *libfs.Error {
 			fuseLog, false /* superVerbose */)
 	}
 
+	logger.EnableBufferedLogging(loggingFrequency)
```
patrickxb: this causes an error during compiling.

go/logger/buffer.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// EnableBufferedLogging turns on buffered logging - this is a performance
+// boost at the expense of not getting all the logs in a crash.
+func EnableBufferedLogging() {
+	writer, shutdown, done := NewAutoFlushingBufferedWriter(ErrorWriter(), loggingFrequency)
+	stdErrLoggingShutdown = shutdown
+	stdErrLoggingShutdownDone = done
+	logBackend := logging.NewLogBackend(writer, "", 0)
+	logging.SetBackend(logBackend)
```
jzila: Are you sure this works as you expect it to? It seems to me that this would change all logging backends, not just the one for KBFS.

zapu: not if this is only called from KBFS binaries, right?

jzila: But everything is one binary on mobile.

zapu: oh you're right!

go/kbfs/kbfsfuse/main.go
jzila: `EnableBufferedLogging` does not take a `loggingFrequency`.

maxtaco:
BTW, @mmaxim also wants a look.

go/logger/buffer.go
jakob223: I forgot that everything is one binary on mobile. Should I go for the slightly bigger refactor to cut the global variables out entirely?

zapu: that's why I mentioned that maybe it would be easier to special case logger on core side

because from what I observed, we always access logger through this variable:
https://github.com/keybase/client/blob/master/go/libkb/globals.go#L49

so we have our own global, in a way

patrickxb: I don't want to cause any more work.  As @zapu says we can turn it off in the service if we want.  I just wanted the flexibility to do that, previous PR I believe had it always on...thanks!

jakob223: This PR has it on in the single process on mobile by default. Are you okay with this going in like that, or should we wait until you modify the service / should I modify the service to use a specific log backend?

jzila:
@mmaxim we still need your review here.

@patrickxb, @zapu did you decide whether you want us to try to factor out the global logger or not?

jakob223:
Got approval from patrick over chat, merging.

PR #16669, opened 2019-03-20
fix for TestMerkleAuditRetry flake

pzduniak:
I would've never thought we'd run into it but it happened! The auditor randomly selected the same block twice in a row in CI.

### DIFF
--- a/go/engine/merkle_audit.go
+++ b/go/engine/merkle_audit.go
@@ -48,6 +48,7 @@ type MerkleAuditArgs struct {
 
 type merkleAuditState struct {
 	RetrySeqno *keybase1.Seqno `json:"retrySeqno"`
+	LastSeqno  *keybase1.Seqno `json:"lastSeqno"`
 }
 
 // NewMerkleAudit creates a new MerkleAudit engine.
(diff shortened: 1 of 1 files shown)


go/engine/merkle_audit.go
```diff
@@ -222,12 +223,19 @@ func MerkleAuditRound(m libkb.MetaContext) (err error) {
 		}
 
 		// 3. Generate a random seqno for the starting root in the audit.
-		randomSeqno, err := randSeqno(*firstSeqno, lastSeqno)
-		if err != nil {
-			return err
+		for {
```
maxtaco: can you split this into a little function?

pzduniak: moved this logic into randSeqno

go/engine/merkle_audit.go
```diff
@@ -100,34 +101,39 @@ func (e *MerkleAudit) Shutdown() {
...
+// randSeqno picks a random number between [low, high) that's different from prev.
+func randSeqno(lo keybase1.Seqno, hi keybase1.Seqno, prev *keybase1.Seqno) (keybase1.Seqno, error) {
+	for {
+		rangeBig := big.NewInt(int64(hi - lo))
+		n, err := rand.Int(rand.Reader, rangeBig)
+		if err != nil {
+			return keybase1.Seqno(0), err
+		}
+		newSeqno := keybase1.Seqno(n.Int64()) + lo
+		if prev == nil || *prev != newSeqno {
```
maxtaco: Actually is it possible for this to loop forever?  What if the params are [0,1), with prev=0?

pzduniak: Sure, but isn't that even less likely than the flake itself? Even locally / in dev mode we preload the database with quite a few roots.

maxtaco: Still it makes me nervous, without any comment or anything.  Just make it error out in that case, I don't want to spin a goroutine up to 100% without knowing it sometime down the road.

pzduniak: fixed

PR #16674, opened 2019-03-20
Add "delete" for folders in the UI

jakob223:
Behind a feature flag - except I can't seem to get it to show for me now that I've put it behind the feature flag, even though the feature flag is set for me :(

### DIFF
--- a/shared/__mocks__/feature-flags.js
+++ b/shared/__mocks__/feature-flags.js
@@ -11,6 +11,7 @@ const ff: FeatureFlags = {
   airdrop: true,
   chatIndexProfilingEnabled: false,
   dbCleanEnabled: false,
+  enableDeleteFolder: false,
   foldersInProfileTab: true,
   moveOrCopy: true,
   newTeamBuildingForChatAllowMakeTeam: false,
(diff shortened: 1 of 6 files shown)


jakob223:
oops this is polluted with other work, will rebase

jakob223:
Why is there a `.js.flow` file as well as the `.native.js` and the `.desktop.js` files? Is the flow compiled to JS by something in the build toolchain but the JS it's compiled into doesn't show up anywhere?

songgao:
The flow file is because flow doesn’t understand the electron/native split. Flow file is not compiled to js. When you run flow if there’s a flow file flow just looks at that and not the actual files.

songgao:
I think you need to add it to `__mocks__/feature-flags` too

jakob223:
@songgao I thought you said __mocks__ only needed the flag for storyshots, and it looks like there isn't a story shot for this flow.

songgao:
@jakob223 it still needs to satisfy the feature flag’s type, so even if no story touches it you’d still need it in there. (I think there’s a story for path-item-action; not sure if it currently exposes the Delete button though.)

PR #16676, opened 2019-03-20
Only allow clicks on 'manage chat channels' icon on teams list if you're a member of the team

cjb:
@keybase/react-hackers 

Trying to load Manage Chat Channels for a team you're an implicit admin of doesn't work, so remove the onClick handler from the icon on the main team list in that case.

While we have yourOperations entries for a team once you've loaded it, the main team list is rendered without any teams loaded.  Instead we get an untrusted team list from the server.  That includes a field for our role in each team, so this PR uses "role !== 'none'" as the canManageChannels equivalent, rather than yourOperations.


buoyad:
Why show the icon at all? Seems like that was the intent of the original jsx, but they never passed in null.

cjb:
@buoyad Yeah, I tried that first but didn't like the way it looked:

Screen Shot 2019-03-21 at 8 57 12 AM

So I guess our choices are:

1) this screenshot

2) chat icon isn't present if you can't chat and there's an empty space where it would otherwise be, pushing the files icon to its usual place

3) chat icon is always present but doesn't have a click handler (so no color change on hover) if you can't chat

4) chat icon is always present but has some kind of opacity applied if you can't chat

@keybase/design, what'd you think?  This is for the team list when you're an implicit admin of a team so you can't go into its chat channels screen yet.

adamjspooner:
Any reason to not swap the order (chat then folder)?

cecileboucheron:
Rather than hiding the Chat icon, why not disabling it and have a tooltip that says "You need to be a member of the team to be able to chat."?

cjb:
Cool, this should do it.  (Does this seem like the best way to do a conditional tooltip?)

PR #16678, opened 2019-03-21
fix fs routing for nav2

songgao:
I haven't done anything for the headers yet -- I'll do that in follow-up PRs. This is just to make navigations work with nav2.

Also, moved a bunch of navigation stuff from Saga to containers.


songgao:
Thanks @cjb for the quick review!

PR #16679, opened 2019-03-21
test: benchmark for writing and flushing lots of small files

strib:
Writes 1000 5-byte files to the journal in one phase, and flushes them in a second phase, and each phase can be timed independently.

Note that we can't use `b` directly in the benchmarks, because they are actually sub-benchmarks that each get their own `*testing.B` object (due to our framework code that runs a copy of each test for different MD versions).  So this required some new DSL test helper functions that access the correct `*testing.B` for the sub-benchmark. All our benchmarks have been broken since that change.

Issue: KBFS-3979

### DIFF
--- a/go/kbfs/test/benchmark_test.go
+++ b/go/kbfs/test/benchmark_test.go
@@ -282,6 +282,8 @@ func benchmarkMultiFileSync(
 				cb(getBenchN(&n))
 				buf := make([]byte, numFiles*fileSize+fileSize)
 				for i := 0; i < numFiles*fileSize; i++ {
+					// Make sure we mix up the byte values a bit, so
+					// we don't accidentally trigger any deduplication.
 					buf[i] = byte(i)
 				}
 				cb(resetTimer())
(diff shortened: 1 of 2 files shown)


go/kbfs/test/benchmark_test.go
```diff
@@ -231,72 +239,106 @@ func BenchmarkWriteMixedFilesNormalBandwidth(b *testing.B) {
...
-					}
-					err := cb(truncate(files[i], 0))
-					if err != nil {
-						return err
-					}
+				var n int
+				cb(getBenchN(&n))
+				buf := make([]byte, numFiles*fileSize+fileSize)
+				for i := 0; i < numFiles*fileSize; i++ {
+					buf[i] = byte(i)
```
jzila: This value doesn't really matter, right? Because it'll get truncated. Maybe a comment to indicate that it's just to populate the buffer with non-zero values.

strib: The value sort of matters because it gets "truncated" to the full size of the file, leaving all the bytes there, just as a way to force a sync. So I wanted the bytes to all be different, to make sure we didn't trigger any weird dedup stuff (which should be off for journal tests anyway, but better safe than sorry).

I'll add a comment though to be extra clear.

jzila: Oh I mean that `i` gets truncated to `0-255`, and just loops there over those numbers. But yeah I get your point.

PR #16680, opened 2019-03-21
Surya/nil/ctlstop

heronhaye:
Addresses part of https://github.com/keybase/keybase-issues/issues/3242


heronhaye:
@keybase/react-hackers for frontend changes

heronhaye:
Windows code is in go/client/cmd_ctl_stop_windows.go, go/client/cmd_ctl_stop.go has  `// +build !darwin,!windows` at the top

shared/desktop/app/ctl.desktop.js
```diff
@@ -12,6 +12,10 @@ export function ctlStop(callback: any) {
     args = ['ctl', 'stop']
     plat = 'win32'
   }
+  if (isLinux) {
```
maxtaco: why isn't there an RPC to do this? Seems like calling out to the CLI is pretty gross. Just wondering about why this happened.

go/service/ctl.go
```diff
@@ -32,6 +32,12 @@ func (c *CtlHandler) Stop(_ context.Context, args keybase1.StopArg) error {
 	return nil
 }
 
+func (c *CtlHandler) StopAll(_ context.Context, args keybase1.StopAllArg) error {
+	c.G().Log.Debug("Received stopAll(%d) RPC; shutting down", args.ExitCode)
+	go c.service.StopAll(args.ExitCode)
```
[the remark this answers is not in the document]
heronhaye: I'm not sure, it doesn't seem necessary since killall/pkill won't hang. Should I change it for the existing `Stop` rpc too?

shared/desktop/app/ctl.desktop.js
heronhaye: No idea, just followed pattern. @zanderz do you know?

maxtaco: Feel really brittle to shell out here

maxtaco: Can you move the "Stopall" code into the service?  And the last thing it does is harakiri?

heronhaye: what do you mean by move into the service? the `-all` flag? the last thing the js does is stop the process itself, but ctl stop should've killed it before - I think this is only for the case where ctl stop did not work, it still closes the gui anyway. 

Also, I wonder if the intent was so that it still works if somehow the rpc connection between the service and the gui were disconnected but the service still listens to cmdline rpcs. Don't know if that's likely.

zanderz: I don't think the RPC existed yet when this was put in, I'll take a look and refresh my memory. Note the updater does a blunt kill as well rather than an RPC I believe.

zanderz: wait, sorry, that doesn't make sense - of course `ctl.avdl` and so forth were there, maybe it should call that directly, like you say.

maxtaco:
@heronhaye i don't see how that case could happen. Yes, I was asking to move the `keybase ctl stop --all` machinery out of `client` and into `service`, and for the service to perform these actions. Ideally, electron would just make an RPC to cause all of these processes to shut down.

maxtaco:
The idea of electron shelling is out isn't great, and plus, we're going to stop allowing these types of behaviors as we harden the electron process (cc @chrisnojima)

chrisnojima:
Security concerns aside I don't want the GUI to even understand this kind of thing. Def prefer an rpc call

shared/desktop/app/ctl.desktop.js
zanderz: maybe trying to call the RPC was just too scary? I just made a PR to do it this way but I don't think it quite works. Still don't have the hang of it after 2 years.  https://github.com/keybase/client/pull/16835

shared/actions/platform-specific/index.desktop.js
```diff
@@ -245,7 +245,7 @@ const onConnected = () => {
       argv: process.argv,
       clientType: RPCTypes.commonClientType.guiMain,
       desc: 'Main Renderer',
-      pid: process.pid,
+      pid: SafeElectron.getRemote().process.pid,
```
heronhaye: @keybase/react-hackers is this change ok? before the pid was the renderer's pid, now it is the main process's pid so killing it kills everything

heronhaye:
Closing this because scope changed due to frontend pr. Made https://github.com/keybase/client/pull/16949

PR #16683, opened 2019-03-21
Improve performance of loading git repos

joshblum:
depends on https://github.com/keybase/keybase/pull/3622
- uses server given role flag to determine team role instead of team load
- unboxes metadata in parallel

### DIFF
--- a/go/git/delete.go
+++ b/go/git/delete.go
@@ -17,9 +17,6 @@ func DeleteMetadata(ctx context.Context, g *libkb.GlobalContext, folder keybase1
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


go/git/delete.go
```diff
@@ -17,9 +17,6 @@ func DeleteMetadata(ctx context.Context, g *libkb.GlobalContext, folder keybase1
 		return err
 	}
 
-	// The GUI doesn't give us the repo_id back, so we need to figure it out.
```
joshblum: i started implementing this but the CLI only has the repo name so I left as is

PR #16685, opened 2019-03-21
Generic proofs creation in GUI

buoyad:
This hooks up the 'Additional identities' button -> available proof list -> generic proof creation. It works like this:

1. We call `startProof`
2. Service sends up some `ProveParameters` that we use to render the enter-username screen
3. GUI submit the username to the service which either gives us an error message to show or says it's A-OK
4. If it's OK, service sends up a URL (`outputInstructions`) that we direct the user to and starts polling for the verification; in the meantime we show a red "proof unreachable" screen with a button that goes to the same link
4. When verified, `startProof` returns and we show the success screen.

To test this, you need `KEYBASE_FEATURES=admin` or `keybase config set features admin` and a whitelisted test user. 

- [x] Storyshots before merge

r? @keybase/react-hackers


zanderz:
not "provenly"? (ha, ha)
image

buoyad:
@zanderz I think that is intended, cc @xgess ?

xgess:
huh. i think `provenly` does a better job of implying that this is a thing that has happened as opposed to one that could potentially happen. but i'm not super passionate about it.

mmaxim:
I think the real question was about the "View Proof" link.

xgess:
oops. so, there isn't really a human-friendly link for the proof we can send someone to anymore, because the proof itself is more like a setting on the external account. i think there are three choices here: (1) the external profile page, e.g. `mastodon.whatever/@user`, but i think we already have a link for that, (2) our internal representation of the proof (whatever it is now or similar), (3) the json api endpoint we hit for that user, which looks like this `https://mastodon.social/api/proofs?provider=keybase&username=xgess` but is far from human-optimized.

xgess:
@buoyad ^

mlsteele:
@xgess Assuming you're talking about this link on the profile screen.
image
How about [this target](https://keybase.io/testkeyhole1/sigs/48dfaa5c1476dfe5b073319c0912969ede049439dd5c0d0046f7aedb6480df890f)? The services determines that link in adapter.go

buoyad:
I like that target @mlsteele, as an aside on that banner I think it's hard to find where the proof is on mastodon's side. Maybe it could say "See the proof on [keybase]() or [mastokeyhole.herokuapp.com]()" and the second link can lead to an anchor here?
image
(or just the mastodon profile page)

PR #16692, opened 2019-03-21
mocks: use reflect mode with mockgen, and renable in Jenkins

strib:
We used to use "source" mode, which apparently now doesn't like packages named "context" and "errors" being imported two different ways.

Unfortunately this means we have to explicitly list every symbol we want mocked.


Jenkinsfile
```diff
@@ -381,6 +381,26 @@ def testGo(prefix, packagesToTest) {
...
+
+    if (isUnix()) {
+      // Windows `gofmt` pukes on CRLF, so only run on *nix.
+      println "Running mockgen"
+      retry(5) {
+        sh 'go get -u github.com/golang/mock/mockgen'
+      }
+      retry(5) {
+        timeout(activity: true, time: 90, unit: 'SECONDS') {
+          sh '''
```
jzila: You can wrap this into a `dir('kbfs/libkbfs')` instead of the `cd` surround.

PR #16695, opened 2019-03-21
cr: no longer unstage automatically on a "block not found" error

strib:
This happened to someone and they lost real data from their conflict branch.  Now that we abandon CR attempts after a number of failed attempts, we're no longer at risk of infinite CR loops.  And soon we'll have tools for letting the user move conflict branches out of the way if needed (and hopefully, we'll eventually surface issues to the GUI with a badge somehow).  So it doesn't seem like we need to take the risk of automatically nuking data anymore.


jzila:
Oops, approved too fast. Looks like you might need to update a test.

PR #16698, opened 2019-03-22
fix accidental triggerring click from floating menu

songgao:
In Files, we don't use auto hide. So click event was never stopped from propagating. Not sure how this worked before but only broke after moving to ListItem2. Anyway, this fixes that.


buoyad:
This might break stuff. Might be better to figure out why it broke in this case?

songgao:
The only things affected by this is when you have `closeOnSelect` set to `false`, and the only place outside Files that uses it is profile/user/teams/index.js which seems fine when I tested it.

I think the reason why it only broken after moving to ListItem2 has to do the way click events are handled. Before that we were just using plain ClickablBox and a bunch of Box2's.

buoyad:
Cool, thanks for checking SGTM

PR #16702, opened 2019-03-22
Inline errors when bad username during login

zapu:


zapu:
@cjb I tried to clean up what I had so far but I'm still not exactly happy in how the error is propagated and later how it's decided what component to display. Would you mind taking a look? Thanks!

cjb:
@zapu Looks good! But let's do the decoding of error number into a string in the username container.js rather than the display component, and add a bool for whether to display the sign up link which is set by the container too? It can be a function in that file that's called by mapStateToProps.

Does that make sense? The display components should usually only receive things like bools and strings.

zapu:
Thanks! I moved logic to `shared/provision/username-or-email/container.js`, but I'm not sure if I didn't try to be "too clever" there. Can you take another look?

cjb:
@zapu Pretty good! I'm surprised that flow types successfully -- inlineError is a required prop of the Username component in index, but it's only returned via the spread if there was an error.  Any idea what I'm missing?

It does seem better to have mapStateToProps show what it's returning explicitly, with something like this (untested), what do you think?

```js
const decodeInlineError = inlineRPCError => {
  let inlineError = ''
  let inlineSignUpLink = false
  if (inlineRPCError) {
    switch (inlineRPCError.code) {
      case RPCTypes.constantsStatusCode.scnotfound:
        inlineError = "This username doesn't exist."
        inlineSignUpLink = true
        break
      case RPCTypes.constantsStatusCode.scbadusername:
        inlineError = 'This username is not valid.'
        inlineSignUpLink = false
        break
    }
  }
  return {inlineError, inlineSignUpLink}
}

  const {inlineError, inlineSignUpLink} = decodeInlineError(state.provision.inlineError)
  return {
    error: state.provision.error.stringValue(),
    inlineError,
    inlineSignUpLink,
    // So we can clear the error if the name is changed
    submittedUsername: state.provision.username,
  }
}
```

zapu:
I think its because its nullable. Ill take another look in a bit, thanks!

cjb:
I tried making it `inlineError: string` instead of `inlineError: ?string` and it still passed flow :)  Seems like a Flow bug, though usually when I feel that way it turns out to be a cjb bug..

zapu:
thanks this is perfect now!

zapu:
@patrickxb can you take a look at more cleanups in engines regarding logging in with e-mail? thanks

PR #16703, opened 2019-03-22
Option to hide small team convs

mlsteele:
Option to hide and un-hide small team convs. Once hidden a conv will no longer appear in the inbox. It can still be found by searching the inbox. When anybody sends into the conv, it will be unhidden for all users.

The option is not available for big teams. After this patch it's not available for ad-hoc convs but that is a mistake. The gear that leads to `info-panel/menu` doesn't show up for ad-hoc convs at the moment. Future patch will make it available for ad-hoc too.

image


mmaxim:
Isn't making it available for those convs an easy change? Why not put it in this PR? That's the main goal of this feature is to hide those convs.

mlsteele:
It's easy but not that easy. Trying to keep the PR small and avoid conflicts in the non-view code

shared/constants/chat2/index.js
```diff
@@ -216,7 +216,7 @@ export const makeInboxQuery = (
     convIDs: convIDKeys.map(Types.keyToConversationID),
     readOnly: false,
     status: Object.keys(RPCChatTypes.commonConversationStatus)
-      .filter(k => !['ignored', 'blocked', 'reported'].includes(k))
+      .filter(k => !['blocked', 'reported'].includes(k))
```
mmaxim: I don't think we should do this, is this just so these show up in "jump to conv"?

mlsteele: But then you couldn't get the conv back from the gui

mmaxim: JUst go to new conversation.

mlsteele:
Now the gui doesn't request status=ignored conversations. Whenever a conv is 'previewed' that conv is set to status=unfiled. Resurfacing a hidden conv is still quite buggy. Hopefully a bit better.

mlsteele:
Using `metasReceived` might be better than unhiding. But `findConversationsLocal` returns `ConversationLocal` which ain't a meta as far as I can tell.

PR #16705, opened 2019-03-22
enter reset pipeline with passphrase

joshblum:
depends on https://github.com/keybase/keybase/pull/3629

- prompts the user for a passphrase, if this errors out we start the pipeline with the email/username instead.

### DIFF
--- a/go/client/cmd_account_reset_start.go
+++ b/go/client/cmd_account_reset_start.go
@@ -39,9 +39,6 @@ func (c *CmdAccountResetStart) ParseArgv(ctx *cli.Context) error {
 	}
 
 	c.usernameOrEmail = ctx.Args().Get(0)
-	if len(c.usernameOrEmail) == 0 {
-		return fmt.Errorf("username or email is required")
-	}
 	return nil
 }
 
--- a/go/client/cmd_account_reset_start.go
+++ b/go/client/cmd_account_reset_start.go
@@ -7,17 +7,16 @@ import (
 	"github.com/keybase/client/go/libcmdline"
 	"github.com/keybase/client/go/libkb"
 	keybase1 "github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/go-framed-msgpack-rpc/rpc"
 	"golang.org/x/net/context"
 )
 
 // Start the reset pipeline
 type CmdAccountResetStart struct {
 	libkb.Contextified
-	username string
-	email    string
+	usernameOrEmail string
 }
 
-// TODO CORE-10466, add support for optional password prompt.
 func NewCmdAccountResetStart(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
 	return cli.Command{
 		Name:  "reset-start",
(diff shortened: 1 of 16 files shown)


go/client/cmd_account_reset_start.go
```diff
@@ -44,24 +39,35 @@ func NewCmdAccountResetStartRunner(g *libkb.GlobalContext) *CmdAccountResetStart
...
 
 func (c *CmdAccountResetStart) ParseArgv(ctx *cli.Context) error {
-	c.email = ctx.String("email")
-	c.username = ctx.String("username")
-
-	if len(c.email) > 0 && len(c.username) > 0 {
-		return fmt.Errorf("only email or username can be specified")
+	c.usernameOrEmail = ctx.String("account-name")
+	if len(c.usernameOrEmail) == 0 {
+		return fmt.Errorf("Account name is required")
```
mlsteele: should print usage here. Maybe this does. Positional arg might be nicer for the user, seeing as this command has no other options

go/engine/account_reset.go
```diff
@@ -42,6 +42,61 @@ func (e *AccountReset) SubConsumers() []libkb.UIConsumer {
...
 // Run starts the engine.
-func (e *AccountReset) Run(mctx libkb.MetaContext) error {
-	return libkb.EnterResetPipeline(mctx, e.username, e.email)
+func (e *AccountReset) Run(mctx libkb.MetaContext) (err error) {
+	mctx = mctx.WithLogTag("RST")
+	defer mctx.TraceTimed("Account#Run", func() error { return err })()
+
+	// User's with active devices cannot reset at all
+	if mctx.ActiveDevice().Valid() {
+		return libkb.ResetWithActiveDeviceError{}
```
mlsteele: this seems like libkb.EnterResetPipeline's job

mlsteele: nevermind i see

go/engine/account_reset.go
```diff
@@ -42,6 +42,61 @@ func (e *AccountReset) SubConsumers() []libkb.UIConsumer {
...
+	}
+
+	// We can enter the reset pipeline with exactly one of these parameters
+	// set. We first attempt to establish a session for the user. Otherwise we
+	// send up an email or username.
+	var self bool
+	var username, email string
+
+	mctx = mctx.WithNewProvisionalLoginContext()
+	err = libkb.PassphraseLoginPrompt(mctx, e.usernameOrEmail, 3)
```
mlsteele: how does this feel for the user? Do they get a prompt when they just told you they've lost the account? Some text on the prompt that says "if you don't know it, that's fine you can use email instead"

Maybe asking them whether they'd like to enter their passphrase or use an email would be better

joshblum: I'll change the prompt to be specific to this call so we can customize it further. it currently has cancel/close buttons that allow you to skip the prompt, but I agree that it could be clarified as optional. When entering a passphrase you have to provide a username or email (or you will be prompted for one) which is why I made the cli arg required

go/client/cmd_account_reset_start.go
joshblum: agreed positional is better

joshblum:
custom prompt:
Screen Shot 2019-03-26 at 2 36 19 PM
will refactor to whatever design decides upon for the gui flow. what do you think?

PR #16707, opened 2019-03-22
fix Tooltip flashing by giving Kb.Icon proper padding

songgao:


shared/common-adapters/icon.shared.js
```diff
@@ -140,3 +140,19 @@ export function getMultsMap(imgMap: {[size: string]: any}, targetSize: number):
 export function castPlatformStyles(styles: any) {
   return styles
 }
+
+const makePaddingStyles = () =>
+  Object.entries(Styles.globalMargins).reduce(
+    (styles, [paddingName, padding]) => ({
+      ...styles,
+      [paddingName]: {
+        display: 'block',
```
jzila: Why `block` and not `inline-block`?

songgao: I tried inline-block and it seemed to work, but then I saw this: https://developer.mozilla.org/en-US/docs/Web/CSS/display#Formal_syntax which seems to suggest it's legacy. I _think_ `inline block` might work but didn't try it as `block` seems to be working fine. Is there a reason to not use `block`?

jzila: I'm not sure why it's flagged as legacy, but AFAIK `block` won't allow other elements beside it.
https://stackoverflow.com/a/9189873

songgao: That doesn't seem correct. We definitely have `block` items besides each other. Perhaps with a `flow` parent this is different? Also note that [inside `Kb.Icon`](https://github.com/keybase/client/blob/db1cb82d9d95aba7c735313bec4c48e645e06112/shared/common-adapters/icon.desktop.js#L132) we still wrap it inside a `<div/>` (which I assume is `block` by default).

songgao: Seems like there's outer vs inner display type: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flow_Layout/Block_and_Inline_Layout_in_Normal_Flow#The_display_property_and_flow_layout

jzila: I think in situations where we have `block` elements beside each other, their display is overridden by a parent `display: flex; flex-direction: row`.

I guess if `display: block` works here, then that's fine.

In any case, someone from frontend should also review this.

songgao:
@keybase/react-hackers bump!

songgao:
Bump! My Nav2 PR is based on top of this and approved, so it'd be nice to merge this.

shared/common-adapters/icon.shared.js
MarcoPolo: `display: inline` is more semantically correct. See https://stackoverflow.com/a/3099045. LMK if that doesn't work as expected though

PR #16708, opened 2019-03-23
MetaContext for ephemeral, erasablekvstore packages

joshblum:
refactor `ephemeral` and `erasablekvstore` to use `libkb.MetaContext` throughout to pave the way for user switching


maxtaco:
Nice! I don't think this is required for the first pass on user switching, but it's good to have.

Is this a real bug? --- FAIL: TestChatSrvGetThreadNonblockServerPage (27.11s)

joshblum:
Yeah I figured it would be good to have a gradual transition to mctx I’ll
check out the test case

joshblum:
```
$ go test -run TestChatSrvGetThreadNonblockServerPage -count 5
PASS
ok      github.com/keybase/client/go/chat       12.350s
```

flake!

PR #16712, opened 2019-03-25
first stab at a switch user operation, works pretty well!

maxtaco:
- access this feature via `keybase login -s user2`
- works pretty well, but there are some issues:
  - Files tab totally not working (cc @jzila / @songgao)
  - sometimes the chat UI has a hiccup
  - ~~i think the UI doesn't figure out that the service is disconnected should it crash. i'm not sure if this is on this branch or not~~ (this was nav2 that breaks it)
- i think we should get this in as a pretty good start and then iterate on it

### DIFF
--- a/go/client/cmd_login.go
+++ b/go/client/cmd_login.go
@@ -19,6 +19,12 @@ import (
 )
 
 func NewCmdLogin(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	flags := []cli.Flag{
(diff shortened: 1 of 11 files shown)


maxtaco:
Another thing is that `logout` will likely not work very well with this change, but I'll fix that in an upcoming PR.

maxtaco:
Merging over a kbfs-master RED, which I think was a CI failure since it worked previously

PR #16713, opened 2019-03-25
Files: whitelist supported image types for preview

songgao:


shared/constants/fs.js
```diff
@@ -503,13 +503,14 @@ export const userTlfHistoryRPCToState = (
   return I.List(updates)
 }
 
+const supportedImgMimeTypes = ['image/png', 'image/jpeg', 'image/gif', 'image/webp']
```
jzila: `new Set(['image/png, ...])`
...
`supportedImgMimeTypes.has(mimeType)`

songgao: OK. Although I actually thought about that, but there are four items here, and I'm not sure having that would actually be faster.

jzila: There are 4 now, but there will likely be more over time.

PR #16714, opened 2019-03-25
Fix desktop SectionList blank rendering with tall items sometimes

buoyad:
Two problems: 

1. `ReactList` doesn't support returning `null` from itemRenderer, at least in the case on the profile screen.
2. The code assumed `itemRenderer` only has one param in `ReactList`, but it actually has two.

I also tweaked the calculation of when to render the sticky header. This is a tradeoff, on the plus side it gets rid of the jitter in the old implementation. The con is that when scrolling real fast the header can sometimes end up in the wrong state, bit it's fixed by any subsequent scrolling so this is pretty minor IMO. r? @keybase/react-hackers

### DIFF
--- a/shared/common-adapters/box.desktop.js
+++ b/shared/common-adapters/box.desktop.js
@@ -25,7 +25,7 @@ const injectGaps = (component, _children, gap, gapStart, gapEnd) => {
   return children
 }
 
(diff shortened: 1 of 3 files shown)


buoyad:
@keybase/react-hackers bump

PR #16719, opened 2019-03-26
when logging out, switch to other active user

maxtaco:
- a new Logout engine
- in Login, don't return the global NIST if we're logging in as a different user
- we might want to revisit the whole strategy to logging in as a separate user, since we wind up mutating global state for a failed login, if we're already
logged in as A and we're switching over to B. this is a bigger change, so maybe let's wait on it.

### DIFF
--- a/go/engine/login_test.go
+++ b/go/engine/login_test.go
@@ -196,6 +196,87 @@ func TestUserEmails(t *testing.T) {
 	}
 }
 
+func TestProvisionAfterSwitchWithWrongUser(t *testing.T) {
+	// device X (provisioner) context:
+	t.Logf("setup X")
+	tcX := SetupEngineTest(t, "kex2provision")
+	defer tcX.Cleanup()
+
+	// device Y (provisionee) context:
+	t.Logf("setup Y")
+	tcY := SetupEngineTest(t, "template")
+	defer tcY.Cleanup()
+
+	// provisioner needs to be logged in
+	t.Logf("provisioner login")
+	userX := CreateAndSignupFakeUserPaper(tcX, "login")
+	var secretX kex2.Secret
+	_, err := rand.Read(secretX[:])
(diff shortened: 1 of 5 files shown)


patrickxb:
How does this affect other calls of G().Logout()?

For example, 

    func (e *loginProvision) cleanup(m libkb.MetaContext) {                              
        if !e.cleanupOnErr {                                     
                return                                                                               
        }                                                  
                                                                                                
        // the best way to cleanup is to logout...
        m.Debug("an error occurred during provisioning, logging out")
        m.G().Logout(m.Ctx())                                                                                             
    }

maxtaco:
That's a good point. It might make sense just to logout in those cases though since I'm not sure login should be calling logout, which then calls login?

patrickxb:
We used Logout in the past to (help) leave the user in a clean state if there was an error.

patrickxb:
Are there any new tests for provisioning when logged in with multiple users?

go/engine/login_test.go
```diff
@@ -177,6 +196,184 @@ func TestUserEmails(t *testing.T) {
...
+
+	// after provisioning, the device keys should be cached
+	assertDeviceKeysCached(tcY)
+
+	// after provisioning, the secret should be stored
+	assertSecretStored(tcY, userX.Username)
+
+}
+
+func TestProvisionAfterSwitchWithWrongUser(t *testing.T) {
```
patrickxb: 👍

go/engine/login_test.go
```diff
@@ -177,6 +196,184 @@ func TestUserEmails(t *testing.T) {
 	}
 }
 
+func TestProvisionDesktopAfterSwitch(t *testing.T) {
```
patrickxb: 👍

patrickxb:
LGTM...thanks for the tests.

PR #16721, opened 2019-03-26
ShouldUseParameterizedProofs -> true

mlsteele:
Always use parameterized proofs apparatus. This means fetching the list of available proofs from the merkle tree. Affects the order and allowedness of non-parameterized proof types too.

Users will see any existing parameterized proofs. Whether they can create new ones of each type is gated by the server.

This should be safe now that the gui is in. The app should now be forwards compatible. Any reason it wouldn't be?

### DIFF
--- a/go/libkb/constants.go
+++ b/go/libkb/constants.go
@@ -461,7 +461,7 @@ var RemoteServiceTypes = map[string]keybase1.ProofType{
 	"generic_social": keybase1.ProofType_GENERIC_SOCIAL,
 }
 
-// TODO Remove with CORE-9923
+// remove when ShouldUseParameterizedProofs is removed
 var RemoteServiceOrder = []keybase1.ProofType{
 	keybase1.ProofType_KEYBASE,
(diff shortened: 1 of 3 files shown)


xgess:
will this now show the uncreatable proof types in the list when you do `keybase prove -l`?

mlsteele:
@xgess only if you have some [weird flag](https://github.com/keybase/keybase/pull/3634) on your account

mlsteele:
`keybase prove -l` and the gui will show the same list. That list is all proof types for which `!creation_disabled` on the server.

Unless you have `experimental_generic_proofs` set on your account from the server or have run the service with `--prove-bypass` locally. In which case you will see all proof known types suggested  regardless of `creation_disabled`.

Whether you can actually make it through the server to post a proof that has `creation_disabled:true` depends on whether `experimental_generic_proofs` is set on the account. `--prove-bypass` alone is not enough for the server to accept those proofs.

PR #16723, opened 2019-03-26
Animate scroll for unread messages shortcut on desktop

buoyad:
Calculate the item heights between the last visible item and the item we're trying to scroll to, and call `scrollBy` on the div directly. r? @keybase/react-hackers

### DIFF
--- a/shared/chat/inbox/index.desktop.js
+++ b/shared/chat/inbox/index.desktop.js
@@ -183,11 +183,10 @@ class Inbox extends React.PureComponent<Props, State> {
     if (this._firstOffscreenIdx <= 0 || !this._scrollDiv.current) {
       return
     }
-    let top = 0
+    let top = 100 // // give it some space below
     for (let i = this._lastVisibleIdx; i <= this._firstOffscreenIdx; i++) {
       top += this._itemSizeGetter(i)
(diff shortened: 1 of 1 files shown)


shared/chat/inbox/index.desktop.js
```diff
@@ -179,10 +180,14 @@ class Inbox extends React.PureComponent<Props, State> {
   }, 200)
 
   _scrollToUnread = () => {
-    if (this._firstOffscreenIdx <= 0 || !this._list) {
+    if (this._firstOffscreenIdx <= 0 || !this._scrollDiv.current) {
       return
     }
-    this._list.scrollToItem(this._firstOffscreenIdx, 'center')
+    let top = 100 // // give it some space below
```
cjb: :sheep: double //.  And, I tried out top = 25 and I prefer it myself, but it's totally your call!

buoyad: Ah, thanks. 
How far it scrolls depends a bit on how big the item you started at is, so with smaller numbers occasionally it wouldn't scroll enough to fully show the unread row. I think 100 makes the variation less noticeable because it's ≫ our row heights.

Issue #16564, opened 2019-03-13
How does the keybase app "remember my passphrase" on linux?


presencia:
I just installed the keybase app on my linux pc. I noticed that when the "remember  my passphrase" option is checked in the gui, I am not prompted for my passphrase on startup of the app, even after rebooting my computer.

I understand from the [docs about local key security](https://keybase.io/docs/crypto/local-key-security) keybase uses os keyrings to store the passphrase when possible, but this feature is said to be only implemented on OSX. I am on a linux machine using gnome but keybase seems to login even with the gnome-keyring still locked.

I am confused. How/where does the keybase app store my login passphrase to remember it? Is it somewhere in plan text? Or some protected keyring I am unaware of? Or does the "remember my passphrase" feature work without storing the passphrase?

I have searched the keybase docs and github issues for an answer without success. I would greatly appreciate any help in understanding this feature. In case this is not the right place to ask this kind of question, please point me to the right direction. Thanks.

junderw:
1. session token saved to disc in your home folder
2. device private key saved to disk in your home folder

maxtaco:
Here's the overall protocol (cc @patrickxb):

* Your private keys are encrypted with a symmetric key *p*, which is the XOR of some server-stored data, and a portion of the scrypt stream derived from your passphrase.
* on macOS, iOS  and Android we store *p* to the OS's keychain, since those OSes have nice hardware integrations and reasonably standard keychain systems (compare to Linux which has one for each windows manager -- groan!)
* on Windows and Linux, we don't have a nice keychain system, so instead we:
  * Generate a 2MB file of random data, call it *f*
  * Hash *f* to a 32-byte key *h*
  * Encrypt *p* with *h*, and write that to your home directory.
* This means that all the elements to decrypt your private keys are in your home directory, but when you "logout" or unclick the "remember passphrase" box, we zero out the file *f*. The thought here is that even if *f* is on an SSD (which makes it hard to actually delete file blocks), it's still very challenging for an attacker who has access to your unencrypted SSD to construct *f*, decrypt *p*, and then decrypt your secret keys.

presencia:
Thank you for the quick and insightful reply, @maxtaco. This is just what I was searching for.
Although I am not a computer security expert and therefore forced to trust in the expertise of you developers in evaluating the security implications of any implementation approach, I can at least now understand the general idea. I feel much more comfortable in deciding whether or not I want to check the "remember my passphrase" option.

[...]

PR #16732, opened 2019-03-26
Surya/core 10499/secretservice impl

heronhaye:
Addresses https://github.com/keybase/client/issues/16564.

Max wanted me to disable this for nopw since it would be bad if we lost a user's randompw. Not sure if I did it in the cleanest way, open to restructuring the SecretStoreOptions thing. I also disabled automatic upgrades for now, so only new devices or people who log out/log in will have it stored in the keyring. So hopefully we get some useful feedback.

There are generic tests for the secretstore so I didn't add any specifically for secretservice, since it won't work in CI anyway.

Outstanding question: should we add gnome keyring or kwallet to our dependencies? Users of the one we don't pick might be irked though

### DIFF
--- a/go/engine/signup.go
+++ b/go/engine/signup.go
@@ -122,7 +122,7 @@ func (s *SignupEngine) Run(m libkb.MetaContext) (err error) {
 		return err
 	}
 
-	if err = s.registerDevice(m, s.arg.DeviceName); err != nil {
+	if err = s.registerDevice(m, s.arg.DeviceName, s.arg.GenerateRandomPassphrase); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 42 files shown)


eli-schwartz:
> Outstanding question: should we add gnome keyring or kwallet to our dependencies? Users of the one we don't pick might be irked though

At most, you should add an optdepends (pacman) or Suggests (dpkg). Many Desktop Environments already pull in a secretservice frontend for the user anyway.

There is precedent in Arch Linux for packages which optionally depend on e.g. gnome-keyring. See for example https://www.archlinux.org/packages/extra/x86_64/chromium/

go/engine/signup.go
```diff
@@ -122,7 +122,7 @@ func (s *SignupEngine) Run(m libkb.MetaContext) (err error) {
 		return err
 	}
 
-	if err = s.registerDevice(m, s.arg.DeviceName); err != nil {
+	if err = s.registerDevice(m, s.arg.DeviceName, s.arg.GenerateRandomPassphrase); err != nil {
```
patrickxb: What's this random passphrase stuff in signup?

heronhaye: "random passphrase" is the same as "no passphrase". I needed to thread it through to the point where we store it so we can decide to not use the keyring in this case.

patrickxb: But isn't "no passphrase" a separate project?

PR #16733, opened 2019-03-26
favorites: improve behavior when offline / on startup

jakob223:
Don't wait the whole normal timeout when trying to refresh favorites on startup. Instead, wait 500ms and then serve the request from the disk cache while refreshing in the background.

Issue: KBFS-3938

### DIFF
--- a/go/kbfs/libkbfs/favorites.go
+++ b/go/kbfs/libkbfs/favorites.go
@@ -370,7 +370,7 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 			if err == context.DeadlineExceeded {
 				newCtx, _ := context.WithTimeout(context.Background(),
 					favoritesBackgroundRefreshTimeout)
-				go f.RefreshCache(newCtx, FavoritesRefreshModeBlocking)
+				go f.RefreshCache(newCtx, FavoritesRefreshModeBackground)
 			}
 			f.log.CDebugf(req.ctx,
 				"Serving possibly stale favorites; new data could not be"+
@@ -652,20 +652,28 @@ func (f *Favorites) Delete(ctx context.Context, fav favorites.Folder) error {
 	})
 }
 
+// FavoritesRefreshMode controls how a favorites refresh happens.
 type FavoritesRefreshMode int
 
 const (
+	// FavoritesRefreshModeInMainFavoritesLoop means to refresh the favorites
+	// in the main loop, blocking any favorites requests after until the refresh
+	// is done.
 	FavoritesRefreshModeInMainFavoritesLoop = iota
-	FavoritesRefreshModeBlocking
+	// FavoritesRefreshModeBackground means to refresh the favorites outside
+	// of the main loop.
+	FavoritesRefreshModeBackground
 )
 
-// RefreshCache refreshes the cached list of favorites. If async is true, then
-// this request is processed outside of the normal Favorites queue order.
+// RefreshCache refreshes the cached list of favorites.
 //
-// In FavoritesRefreshModeBlocking, request the favorites in this thread,
-// then send them off to the main thread to be processed.
+// In FavoritesRefreshModeBackground, request the favorites in this function,
+// then send them to the main goroutine to be processed. This should be called
+// in a separate goroutine from anything mission-critical, because it might wait
+// on network for up to 15 seconds.
+//
 // In FavoritesRefreshModeInMainFavoritesLoop, this just sets up a request and
-// sends it to the main thread to process it - this is useful if e.g.
+// sends it to the main goroutine to process it - this is useful if e.g.
 // the favorites cache has not been initialized at all and cannot serve any
 // requests until this refresh is completed.
 func (f *Favorites) RefreshCache(ctx context.Context, mode FavoritesRefreshMode) {
@@ -692,7 +700,7 @@ func (f *Favorites) RefreshCache(ctx context.Context, mode FavoritesRefreshMode)
 	}
 	f.wg.Add(1)
 
-	if mode == FavoritesRefreshModeBlocking {
+	if mode == FavoritesRefreshModeBackground {
 		favResult, err := f.config.KBPKI().FavoriteList(ctx)
 		if err != nil {
 			f.log.CDebugf(ctx, "Failed to refresh cached Favorites: %+v", err)
--- a/go/kbfs/libkbfs/favorites.go
+++ b/go/kbfs/libkbfs/favorites.go
@@ -34,6 +34,7 @@ const (
 	// How long to block on favorites refresh when cache is expired (e.g.,
 	// on startup). Reasonably low in case we're offline.
 	favoritesServerTimeoutWhenCacheExpired = 500 * time.Millisecond
+	favoritesBackgroundRefreshTimeout      = 15 * time.Second
 )
 
 var errNoFavoritesCache = errors.New("disk favorites cache not present")
@@ -64,6 +65,9 @@ type favReq struct {
 	favsAll     chan<- keybase1.FavoritesResult
 	homeTLFInfo *homeTLFInfo
 
+	// For asynchronous refreshes, pass in the Favorites from the server here
+	favResult *keybase1.FavoritesResult
+
 	// Closed when the request is done.
 	done chan struct{}
 	// Set before done is closed
@@ -87,8 +91,10 @@ type Favorites struct {
 	// homeTLFInfo stores the IDs for the logged-in user's home TLFs
 	homeTLFInfo homeTLFInfo
 
-	// Channels for interacting with the favorites cache
+	// Channel for interacting with the favorites cache
 	reqChan chan *favReq
+	// Channel that is full when there is already a refresh queued
+	refreshWaiting chan struct{}
 
 	wg kbfssync.RepeatedWaitGroup
 
@@ -128,10 +134,11 @@ func newFavoritesWithChan(config Config, reqChan chan *favReq) *Favorites {
 	}
 
 	f := &Favorites{
-		config:       config,
-		reqChan:      reqChan,
-		inFlightAdds: make(map[favorites.Folder]*favReq),
-		log:          log,
+		config:         config,
+		reqChan:        reqChan,
+		refreshWaiting: make(chan struct{}, 1),
+		inFlightAdds:   make(map[favorites.Folder]*favReq),
+		log:            log,
 	}
 
 	return f
@@ -318,6 +325,10 @@ func favoriteToFolder(fav favorites.Folder, data favorites.Data) keybase1.Folder
 func (f *Favorites) handleReq(req *favReq) (err error) {
 	defer func() { f.closeReq(req, err) }()
 
+	if req.refresh {
+		<-f.refreshWaiting
+	}
+
 	kbpki := f.config.KBPKI()
 	// Fetch a new list if:
 	//  (1) The user asked us to refresh
@@ -331,12 +342,19 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 	if needFetch || wantFetch {
 		getCtx := req.ctx
 		if !needFetch {
-			timeoutCtx, cancel := context.WithTimeout(req.ctx, favoritesServerTimeoutWhenCacheExpired)
+			var cancel context.CancelFunc
+			getCtx, cancel = context.WithTimeout(req.ctx,
+				favoritesServerTimeoutWhenCacheExpired)
 			defer cancel()
-			getCtx = timeoutCtx
 		}
-		// load cache from server
-		favResult, err := kbpki.FavoriteList(getCtx)
+		// Load the cache from the server. This possibly already happened
+		// asynchronously and was included in the request.
+		var favResult keybase1.FavoritesResult
+		if req.favResult == nil {
+			favResult, err = kbpki.FavoriteList(getCtx)
+		} else {
+			favResult = *req.favResult
+		}
 		if err != nil {
 			if needFetch {
 				// if we're supposed to refresh the cache and it's not
@@ -350,14 +368,15 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 			// If we weren't explicitly asked to refresh, we can return possibly
 			// stale favorites rather than return nothing.
 			if err == context.DeadlineExceeded {
-				// TODO: make sure this doesn't get queued like 10 times
-				go f.RefreshCache(context.Background())
+				newCtx, _ := context.WithTimeout(context.Background(),
+					favoritesBackgroundRefreshTimeout)
+				go f.RefreshCache(newCtx, true)
 			}
 			f.log.CDebugf(req.ctx,
 				"Serving possibly stale favorites; new data could not be"+
 					" fetched: %v", err)
 		} else { // Successfully got new favorites from server.
-			session, sessionErr := f.config.KBPKI().GetCurrentSession(req.ctx)
+			session, sessionErr := kbpki.GetCurrentSession(req.ctx)
 			oldCache := f.favCache
 			f.newCache = make(map[favorites.Folder]favorites.Data)
 			f.favCache = make(map[favorites.Folder]favorites.Data)
@@ -473,7 +492,6 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 			IgnoredFolders:  ignoredFolders,
 			FavoriteFolders: favFolders,
 		}
-		// TODO: send info on reset users
 	}
 
 	if req.homeTLFInfo != nil {
@@ -634,9 +652,19 @@ func (f *Favorites) Delete(ctx context.Context, fav favorites.Folder) error {
 	})
 }
 
-// RefreshCache refreshes the cached list of favorites.
-func (f *Favorites) RefreshCache(ctx context.Context) {
+// RefreshCache refreshes the cached list of favorites. If async is true, then
+// this request is processed outside of the normal Favorites queue order.
+func (f *Favorites) RefreshCache(ctx context.Context, async bool) {
 	if f.disabled || f.hasShutdown() {
+		return
+	}
+
+	// Insert something into the refreshWaiting channel to guarantee that we
+	// are the only current refresh.
+	select {
+	case f.refreshWaiting <- struct{}{}:
+	default:
+		// There is already a refresh in the queue
 		return
 	}
 	// This request is non-blocking, so use a throw-away done channel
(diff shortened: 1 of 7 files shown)


go/kbfs/libkbfs/favorites.go
```diff
@@ -336,6 +348,10 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 			}
 			// If we weren't explicitly asked to refresh, we can return possibly
 			// stale favorites rather than return nothing.
+			if err == context.DeadlineExceeded {
+				// TODO: make sure this doesn't get queued like 10 times
+				go f.RefreshCache(context.Background())
```
jzila: This should probably use a real context with a real timeout, derived from `req.ctx` preferably so log tags are preserved.

go/kbfs/libkbfs/favorites.go
```diff
@@ -316,16 +319,25 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
...
+	//  (3) It's stale
+	//
+	// If just (3), use a short timeout so we can return the correct result
+	// quickly when offline.
+	needFetch := (req.refresh || f.favCache == nil) && !req.clear
+	wantFetch := f.config.Clock().Now().After(f.cacheExpireTime) && !req.clear
+	if needFetch || wantFetch {
+		getCtx := req.ctx
+		if !needFetch {
+			timeoutCtx, cancel := context.WithTimeout(req.ctx, favoritesServerTimeoutWhenCacheExpired)
```
jzila: IMO this would be clearer as:
```go
if !needFetch {
	var cancel context.CancelFunc
	getCtx, cancel = context.WithTimeout(getCtx, ...)
	defer cancel()
}

File: go/kbfs/libkbfs/favorites.go [hunk 3]
```diff
@@ -366,7 +382,7 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 				f.newCache[*favorites.NewFolderFromProtocol(
 					folder)] = favorites.DataFrom(folder)
 			}
-			if err == nil {
+			if sessionErr == nil {
```
jzila: whoa nice catch

go/kbfs/libkbfs/favorites.go
jakob223: If I use one derived from `req.ctx`, then it will get cancelled if that gets cancelled, which would be unfortunate if it were to happen soon after.

jakob223: Oh, oops, apparently internally we mostly ignore this context anyway:
```

// RefreshCache refreshes the cached list of favorites.
func (f *Favorites) RefreshCache(ctx context.Context) {
	if f.disabled || f.hasShutdown() {
		return
	}
	// This request is non-blocking, so use a throw-away done channel
	// and context.
	req := &favReq{
		refresh: true,
		done:    make(chan struct{}),
		ctx:     context.Background(),
	}
	f.wg.Add(1)
	select {
	case f.reqChan <- req:
	case <-ctx.Done():
		f.wg.Done()
		return
	}
}
```

jakob223:
@jzila can you take a look at this added channel for preventing simultaneously queued refresh requests?

If that looks good and the comment above is resolved, I will merge.

go/kbfs/libkbfs/favorites.go
```diff
@@ -317,6 +321,10 @@ func favoriteToFolder(fav favorites.Folder, data favorites.Data) keybase1.Folder
 func (f *Favorites) handleReq(req *favReq) (err error) {
 	defer func() { f.closeReq(req, err) }()
 
+	if req.refresh {
+		<-f.refreshWaiting
+	}
```
jzila: Do you still need a `req.refresh`? Seems like you  can just do:
```go
select {
case <-f.refreshWaiting:
	...
default:
}
```

jakob223:
In looking at this I've realized that this "refresh in the background" doesn't really make sense as currently implemented - it still blocks the `Favorites` from serving requests from the cache when it eventually gets queued. I think the right answer might be to let this refresh happen in parallel to other requests, which breaks the queue request-processing format somewhat.

@jzila do you think that's worth it? Or better to just leave the requests as blocking, clean up the synchronization, and call it a day?

jzila:
> it still blocks the Favorites from serving requests from the cache when it eventually gets queued

I don't follow what you mean by this.

> I think the right answer might be to let this refresh happen in parallel to other requests

I thought this was the idea.

go/kbfs/libkbfs/favorites.go
```diff
@@ -336,6 +357,11 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 			}
 			// If we weren't explicitly asked to refresh, we can return possibly
 			// stale favorites rather than return nothing.
+			if err == context.DeadlineExceeded {
+				newCtx, _ := context.WithTimeout(context.Background(),
+					favoritesBackgroundRefreshTimeout)
+				go f.RefreshCache(newCtx)
```
jzila: I'm worried that this could trigger a bunch of goroutines.

go/kbfs/libkbfs/favorites.go
jakob223: `req.refresh` indicates that this particular request is a refresh request. In order to prevent multiple refresh requests from working at the same time, I added this channel in addition to the wait group; however, I claim the boolean is not a sync construct at all.

go/kbfs/libkbfs/favorites.go
jakob223: RefreshCache checks whether there are any other refreshes currently running before it does anything. 

	select {
	case f.refreshWaiting <- struct{}{}:
	default:
		// There is already a refresh in the queue
		return
	}

jakob223:
> I don't follow what you mean by this.

Er, the implementation as of the previous commit didn't really have asynchronous fetching since everything happened in the one goroutine processing the queue. This is fixed now.

go/kbfs/libkbfs/favorites.go
```diff
@@ -630,10 +672,22 @@ func (f *Favorites) RefreshCache(ctx context.Context) {
 		ctx:     context.Background(),
 	}
 	f.wg.Add(1)
+
+	if async {
+		favResult, err := f.config.KBPKI().FavoriteList(ctx)
+		if err != nil {
+			f.log.CDebugf(ctx, "Failed to refresh cached Favorites: %+v", err)
+			f.wg.Done()
+			<-f.refreshWaiting
```
jzila: Won't this deadlock with the `<-f.refreshWaiting` here?
https://github.com/keybase/client/pull/16733/files#diff-5919763ea12fd1848e70a4ca1d6ce411R328

go/kbfs/libkbfs/favorites.go
```diff
@@ -630,10 +672,22 @@ func (f *Favorites) RefreshCache(ctx context.Context) {
...
+			return
+		}
+		req.favResult = &favResult
```
jzila: Shouldn't we `f.wg.Done()` when we're successful?

go/kbfs/libkbfs/favorites.go
jakob223: The only place a request with `refresh: true` is ever created is on line 670, right above here. That request would only get to the processing logic if the results had already been retrieved in this function, or if `async` were false.  In either case, we won't get to the channel drain on line 681, but we will get to either the drain on line 690 or the one in the processing logic.

go/kbfs/libkbfs/favorites.go
jakob223: No, because we've sent the request into the queue when we're successful, and the main thread will process the request and then call that. See lines 503-508:

```func (f *Favorites) loop() {
	for req := range f.reqChan {
		f.handleReq(req)
		f.wg.Done()
	}
}```

Should I add a comment to this effect?

Comment (jzila) [reply go/kbfs/libkbfs/favorites.go]: Yes please.

File: go/kbfs/libkbfs/favorites.go (hunk 6 above)
Comment (jzila) [reply go/kbfs/libkbfs/favorites.go]: OK please comment with a quick explanation.

File: go/kbfs/libkbfs/favorites.go (hunk 7 above)
Comment (jakob223) [reply go/kbfs/libkbfs/favorites.go]: done.

File: go/kbfs/libkbfs/favorites.go (hunk 6 above)
Comment (jakob223) [reply go/kbfs/libkbfs/favorites.go]: done.

**Comment by jakob223** [comment]
Rebased this since it was failing KBFS-server CI.

**Comment by jakob223** [comment]
@jzila do you have more comments here? Or should I merge?

**Comment by jzila** [comment]
@jakob223 it's a LGTM from me, but I'd prefer you wait for Strib's review before merging.

**Comment by jakob223** [comment]
Oh, sorry John, I think I commented on an outdated page and didn't see your newest review before posting my question. I'm fine waiting for strib.

File: go/kbfs/libkbfs/favorites.go [hunk 8]
```diff
@@ -631,10 +675,26 @@ func (f *Favorites) RefreshCache(ctx context.Context) {
 		ctx:     context.Background(),
 	}
 	f.wg.Add(1)
+
+	if async {
+		favResult, err := f.config.KBPKI().FavoriteList(ctx)
```
strib: This seems like the opposite of async -- it's synchronously blocking on a call to the service, right? Did you mean `!async` above?

go/kbfs/libkbfs/interfaces.go
```diff
@@ -270,7 +270,7 @@ type KBFSOps interface {
 	// favorites list and fetch a new list from the server.  The
 	// effects are asychronous; if there's an error refreshing the
 	// favorites, the cached favorites will become empty.
-	RefreshCachedFavorites(ctx context.Context)
+	RefreshCachedFavorites(ctx context.Context, async bool)
```
strib: Please explain what `async` is in the comment above.

go/kbfs/libkbfs/keybase_daemon_rpc.go
```diff
@@ -465,6 +465,6 @@ func newNotifyServiceHandler(config Config, log logger.Logger) keybase1.NotifySe
 func (k *KeybaseDaemonRPC) FavoritesChanged(ctx context.Context,
 	uid keybase1.UID) error {
 	k.log.Debug("Received FavoritesChanged RPC.")
-	k.config.KBFSOps().RefreshCachedFavorites(ctx)
+	go k.config.KBFSOps().RefreshCachedFavorites(ctx, true)
```
strib: Why does this both need to be in a goroutine, and pass `async = true`?  Seems like one is canceling out the other.  I would think the only reason to block on fetching the favorites list is so that we're guaranteed it's done by the time we return, but the `go` construct means we're not even waiting for it to return.  (I agree that we shouldn't be blocking notifications while refreshing the favorites -- I guess I just don't understand the need for `async = true` here, as it seems useless).

strib: And if you do use a `go` here, you have to make a new context, since `ctx` will probably be canceled by the RPC layer as soon as the function returns.

jakob223:
@strib I changed the `bool` to a new type, and I added a bunch of comments in the code to address your questions.

go/kbfs/libkbfs/favorites.go
```diff
@@ -618,23 +652,65 @@ func (f *Favorites) Delete(ctx context.Context, fav favorites.Folder) error {
 	})
 }
 
-// RefreshCache refreshes the cached list of favorites.
-func (f *Favorites) RefreshCache(ctx context.Context) {
+type FavoritesRefreshMode int
```
strib: These will all need comments if you plan to keep them exported.

go/kbfs/libkbfs/favorites.go
```diff
@@ -618,23 +652,65 @@ func (f *Favorites) Delete(ctx context.Context, fav favorites.Folder) error {
...
+
+const (
+	FavoritesRefreshModeInMainFavoritesLoop = iota
+	FavoritesRefreshModeBlocking
+)
+
+// RefreshCache refreshes the cached list of favorites. If async is true, then
```
strib: `async` isn't a thing anymore.

go/kbfs/libkbfs/favorites.go
```diff
@@ -618,23 +652,65 @@ func (f *Favorites) Delete(ctx context.Context, fav favorites.Folder) error {
...
+// this request is processed outside of the normal Favorites queue order.
+//
+// In FavoritesRefreshModeBlocking, request the favorites in this thread,
+// then send them off to the main thread to be processed.
+// In FavoritesRefreshModeInMainFavoritesLoop, this just sets up a request and
+// sends it to the main thread to process it - this is useful if e.g.
+// the favorites cache has not been initialized at all and cannot serve any
```
strib: Sorry I still don't understand why it needs to be initialized in this goroutine -- why wouldn't doing it in the favorites loop be good enough?  Seems like in either case it won't be able to serve favorites request until the background loop gets the request and populates the cache.

Does this have anything to do with the timeout, or are they orthogonal issues that you just happen to be addressing in the same PR?

go/kbfs/libkbfs/favorites.go
```diff
@@ -618,23 +652,65 @@ func (f *Favorites) Delete(ctx context.Context, fav favorites.Folder) error {
...
+
+const (
+	FavoritesRefreshModeInMainFavoritesLoop = iota
+	FavoritesRefreshModeBlocking
+)
+
+// RefreshCache refreshes the cached list of favorites. If async is true, then
+// this request is processed outside of the normal Favorites queue order.
+//
+// In FavoritesRefreshModeBlocking, request the favorites in this thread,
```
strib: s/thread/goroutine/a.

go/kbfs/libkbfs/keybase_daemon_rpc.go
```diff
@@ -465,6 +465,7 @@ func newNotifyServiceHandler(config Config, log logger.Logger) keybase1.NotifySe
 func (k *KeybaseDaemonRPC) FavoritesChanged(ctx context.Context,
 	uid keybase1.UID) error {
 	k.log.Debug("Received FavoritesChanged RPC.")
-	k.config.KBFSOps().RefreshCachedFavorites(ctx)
+	go k.config.KBFSOps().RefreshCachedFavorites(context.Background(),
+		FavoritesRefreshModeBlocking)
```
strib: I still don't understand why this requires "blocking".

go/kbfs/libkbfs/favorites.go
jakob223: The reason to avoid doing it in the main loop is that we want to be able to serve requests from the cache while waiting for the refresh to come from the network.

go/kbfs/libkbfs/keybase_daemon_rpc.go
jakob223: I think "blocking" was probably a bad name for this. I've renamed it to `Background` and clarified in a comment.

go/kbfs/libkbfs/favorites.go
strib: Ohh I see.  It would be great if you could explain that in the comments.

But in that case, there seems to be a problem. You could have multiple sequential `FavoritesChanged` RPCs come in, corresponding to different changes, and if the first one takes longer to complete than the second one, it would overwrite the second one's results in the cache and we'd lose the latest favorites changes.

Seems to me you still have to have some serialization mechanism to make sure there's only one outstanding fetch at a time, and that new refresh requests trigger new API server requests in serial.  For example, there could be a new goroutine managed by `Favorites` that processes requests one at a time and then sends the results over to the main goroutine when each request is complete.

That's assuming we want the property where requests don't block after one of these notifications come in; it's not clear to me that we do.  If we've received a refresh notification, seems like we shouldn't knowingly serve stale data.  A delay in that case seems justified, even in the case where we go offline immediately and it blocks for the full timeout.

go/kbfs/libkbfs/keybase_daemon_rpc.go
strib: I think `Background` is a much worse name, since the call isn't actually made in the background.

I wasn't concerned with the name "blocking", which seems descriptive of what's going to happen.  I was confused about why blocking was needed at all, which you explained in the other comment thread.

go/kbfs/libkbfs/favorites.go
jakob223: > Seems to me you still have to have some serialization mechanism to make sure there's only one outstanding fetch at a time

That's what the `refreshWaiting` channel is for - it only lets one refresh happen at a time. There *is* a possible race here:

(1) Alice does something to invalidate Bob's cache
(2) Bob initiates a refresh
(3a) Bob's request is processed by the server. 
(4a) Carly does something to invalidate Bob's cache
(4b) Bob gets Carly's invalidation and ignores it since there's a request being processed
(3b) Bob's cache finishes handling Alice's change, and believes it is up to date

However, it requires very specific timing between steps 3 and 4, so it doesn't seem much worth worrying about to me.

>  If we've received a refresh notification, seems like we shouldn't knowingly serve stale data.

As soon as we receive such a notification, we'll both invalidate the cache and send off a network request to get new data. However, if that request fails (with the full 15s timeout), I don't think we should make people wait the full 15 seconds the next time they want to look at their TLF list if they went offline right after receiving the notification.

strib: Ah I forgot about `refreshWaiting`, thanks.  I'm not sure I agree about that race you mention not being a big deal, because when it does inevitably happen it will be super confusing.  But I'm willing to table that for now.

> As soon as we receive such a notification, we'll both invalidate the cache and send off a network request to get new data

I don't see where the immediate invalidation happens.  It looks to me like it blocks on `FavoriteList` before sending anything to the main goroutine, right?  Seems like we'll keep knowingly serving stale data until `FavoriteList` returns.

> I don't think we should make people wait the full 15 seconds the next time they want to look at their TLF list if they went offline right after receiving the notification.

You could un-invalidate the cache in the case where the request failed.  That should be pretty easy.

strib: I see you requested another review after renaming the mode stuff, but I've been waiting for a response about my proposal about just doing the fetch in the main goroutine (blocking new favorites requests while we do the fetch), and if it fails just continue to use the previously-cached value for new requests after that, so there's only the one 15-sec blocking period.  Seems like it would be so rare that you'd go offline immediately after getting a invalidation notification that this would be the best plan.

jakob223: I've still got it on my list as TODO respond to this comment and fix stuff - not sure why I requested said review - maybe I requested it before I saw this comment. I've been acting under the assumption the ball is in my court.

jakob223: Coming back to this, I agree that it's better to wait the 15 seconds, and the chances that we're offline are very slim. I changed one call to be `FavoritesRefreshModeInMainFavoritesLoop` - I think that addresses this concern entirely. Also fixed mocks and unswallowed an error. I think this is now actually re-ready for review.

go/kbfs/libkbfs/keybase_daemon_rpc.go
```diff
@@ -465,6 +465,7 @@ func newNotifyServiceHandler(config Config, log logger.Logger) keybase1.NotifySe
 func (k *KeybaseDaemonRPC) FavoritesChanged(ctx context.Context,
 	uid keybase1.UID) error {
 	k.log.Debug("Received FavoritesChanged RPC.")
-	k.config.KBFSOps().RefreshCachedFavorites(ctx)
+	go k.config.KBFSOps().RefreshCachedFavorites(context.Background(),
```
strib: This no longer needs to be a goroutine with a new context, right?

jakob223:
- [ ] squash
 - [ ] pass CI

jakob223:
merging since this passed the previous run and the most recent change is really tiny and CI is borked.

PR #16734, opened 2019-03-27
Make the Settings store immutable

cjb:
@keybase/react-hackers 

This uncovered a few bugs in the process of moving to immutable and fixing typing, I'll mention them in the diff.


shared/constants/types/settings.js
```diff
@@ -15,49 +16,48 @@ export type PendingInvite = {
...
 
-export type NotificationsState = {
+export type _NotificationsState = {
   allowEdit: boolean,
-  groups: {
-    email?: NotificationsGroupState,
-    app_push?: NotificationsGroupState,
-    sms?: NotificationsGroupState,
-  },
+  groups: I.Map<string, NotificationsGroupState>,
```
cjb: These keys (email, app_push, sms) are coming down from the server and rendered dynamically, so it's better not to hardcode them in the type.

shared/reducers/settings.js
```diff
@@ -1,22 +1,21 @@
...
+
+function reducer(state: Types.State = initialState, action: SettingsGen.Actions): Types.State {
   switch (action.type) {
     case SettingsGen.resetStore:
-      return {...Constants.initialState}
+      return initialState
     case SettingsGen.setAllowDeleteAccount:
-      const {allow} = action.payload
-      return {
-        ...Constants.initialState,
```
cjb: This was a bug -- going to delete me (but not yet accepting it) wiped out your settings store because the spread started with initialState instead of current state.

shared/settings/invites/index.desktop.js
```diff
@@ -201,7 +201,6 @@ function AcceptedInviteItem({
       <Kb.Avatar username={invite.username} size={32} />
       <Kb.Box style={{...Styles.globalStyles.flexBoxColumn, marginLeft: Styles.globalMargins.small}}>
         <Kb.ConnectedUsernames type="BodySemibold" usernames={[invite.username]} />
-        <Kb.Text type="BodySmall">{invite.fullname}</Kb.Text>
```
cjb: This was a bug masked by a $FlowIssue -- fullname would never be shown and isn't in the RPC payload.  I thought about starting to fix it and then realized that we don't even use account invites anymore, so I think it can stay as-is.

shared/constants/settings.js
```diff
@@ -2,54 +2,75 @@
...
-
-const traceInProgressKey = 'traceInProgress'
-
-const traceInProgress = (state: TypedState) => WaitingConstants.anyWaiting(state, traceInProgressKey)
-
-const processorProfileInProgressKey = 'processorProfileInProgress'
+})
 
-const processorProfileInProgress = (state: TypedState) =>
+export const traceInProgressKey = 'traceInProgress'
```
songgao: should this be something like `settings:traceInProgress`?

shared/constants/settings.js
```diff
@@ -2,54 +2,75 @@
...
+export const traceInProgress = (state: TypedState) => WaitingConstants.anyWaiting(state, traceInProgressKey)
+export const processorProfileInProgressKey = 'processorProfileInProgress'
```
songgao: Same here

shared/constants/types/settings.js
```diff
@@ -15,49 +16,48 @@ export type PendingInvite = {
...
   name: string,
   subscribed: boolean,
   description: string,
 }
+export type NotificationsSettingsState = I.RecordOf<_NotificationsSettingsState>
 
-export type NotificationsGroupState = {
-  settings: ?Array<NotificationsSettingsState>,
+export type _NotificationsGroupState = {
+  settings: Array<NotificationsSettingsState>,
```
songgao: Did you mean to keep this as an `Array`?

shared/reducers/settings.js
```diff
@@ -57,197 +54,77 @@ function reducer(state: Types.State = Constants.initialState, action: SettingsGe
...
           allowEdit: false,
-          groups: {
-            ...state.notifications.groups,
-            ...changed,
-          },
-        },
-      }
+          groups: state.notifications.groups.merge(I.Map(changed)),
+        }),
+      })
```
songgao: It seems this could also `state.update('notifications', notifications => notifications.merge...)`; same below. Is there a reason to prefer `merge` over `update`?

shared/settings/landing/container.js
```diff
@@ -12,12 +12,13 @@ const mapStateToProps = state => {
   const {emails} = state.settings.email
   const {rememberPassphrase} = state.settings.passphrase
   let accountProps
-  if (emails) {
+  if (emails && emails.count() > 0) {
     let emailProps = {}
-    if (emails.length) {
+    const email = emails.get(0)
```
songgao: FWIW, I think just do `emails && emails.first()` would also work. It just gives you `undefined` if the thing is empty.

cjb:
@songgao Thanks for the review! Made all of these changes.

PR #16739, opened 2019-03-27
leveldb: tune options

jakob223:
Change a few leveldb parameters to increase performance of operations with lots of small files in big directory structures

TODO:
 - figure out how to deal with compaction

### DIFF
--- a/go/kbfs/libkbfs/disk_block_cache.go
+++ b/go/kbfs/libkbfs/disk_block_cache.go
@@ -35,6 +35,7 @@ const (
 	defaultDiskBlockCacheMaxBytes   uint64 = 10 * (1 << 30)
 	defaultBlockCacheTableSize      int    = 50 * opt.MiB
 	defaultBlockCacheBlockSize      int    = 4 * opt.MiB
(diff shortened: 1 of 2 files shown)


jakob223:
Compaction changes are more difficult so I'm dealing with that in a separate PR.

go/kbfs/libkbfs/disk_block_cache.go
```diff
@@ -35,6 +35,7 @@ const (
 	defaultDiskBlockCacheMaxBytes   uint64 = 10 * (1 << 30)
 	defaultBlockCacheTableSize      int    = 50 * opt.MiB
 	defaultBlockCacheBlockSize      int    = 4 * opt.MiB
+	defaultBlockCacheCapacity       int    = 8 * opt.MiB
```
jzila: Is this best to only be double the block size?

jakob223: It was previously 4MB - I tried larger values and they empirically didn't give better results (equal performance [within measurement error] and with more resource usage if I understand that parameter correctly).

PR #16748, opened 2019-03-28
sharing from fs to chat: auto create conv if not exist; add spinner

songgao:
It turns out we don't need at much of waiting state as I thought. Basically only two cases:

1) when user opens `sendLinkToChat` and we need to locate the conversation (small team or impl team) or load a list of conversations (big team);
2) when sending link to chat, and in a rare case where we have the folder but not a chat conversation, and we wait until creating the conv is done before posting the message in a non blocking way.

So this PR adds waiting state on the Send button for both these cases. Also, move navigation for both `sendLinkToChat` and `sendAttachmenttToChat` from Saga into their containers.


songgao:
bump!

shared/actions/fs/index.js
```diff
@@ -690,96 +691,108 @@ const closeDestinationPicker = (state, action) => {
...
+      // noConversationIDKey, causes a transition into 'read-to-send'
+      // state, which is what we want here. If we don't have a
+      // conversation we should create it when user tries to send.
+      FsGen.createSetSendLinkToChatConvID({
+        convID:
+          result.conversations && result.conversations.length
+            ? ChatTypes.conversationIDToKey(result.conversations[0].info.id)
+            : ChatConstants.noConversationIDKey,
       })
     )
```
jzila: Should we also catch an error so that we can process and display it?

shared/actions/fs/index.js
```diff
@@ -690,96 +691,108 @@ const closeDestinationPicker = (state, action) => {
...
+
+          ...(result.convs && result.convs.length === 1
+            ? [
+                // Auto-select channel if it's the only one.
+                FsGen.createSetSendLinkToChatConvID({
+                  convID: ChatTypes.stringToConversationIDKey(result.convs[0].convID),
+                }),
+              ]
+            : []),
+        ]
```
jzila: Shouldn't this return a promise?

songgao: This is used in `chainAction`, so the resolved result is treated as an action or an array of actions to be put into the store. It _can_ return a promise, which would just continue resolving until it's not a promise anymore. In this case, these actions are the final result.

shared/actions/fs/index.js
songgao: Good question; note that chat doesn't catch the error here: https://github.com/keybase/client/blob/d0a8481657b695448c68125250c4173a3bdf0654/shared/actions/chat2/index.js#L1586

So I'm guessing either this RPC is not expect to fail, i.e. if it fails we'd want a black bar, or somewhere else we're filtering on this error to not show a blackbar.

jzila: I know but it seems like an opportunity to reduce the number of black bars.

songgao: You mean if it's not expected to fail? Catching it would just show a red bar instead of black bar. I feel like in that case we should just blackbar and fix it if needed.

PR #16750, opened 2019-03-28
Core changes for passphrase => password in UI strings

cjb:
There's a request from @keybase/design to change "passphrase" to "password" everywhere in the UI.  This PR does that for some visible strings from Core.  There are probably others -- I think it's okay if we don't change every last one as long as we get the common cases.  Let me know if you can think of others to change here.

There's a separate PR that does this all across the frontend (the target branch of this PR).

r? @patrickxb @zapu

### DIFF
--- a/shared/constants/login.js
+++ b/shared/constants/login.js
@@ -3,6 +3,9 @@ import * as I from 'immutable'
 import * as Types from './types/login'
 import HiddenString from '../util/hidden-string'
 
+// An ugly error message from the service that we'd like to rewrite ourselves.
+export const invalidPasswordErrorString = 'Bad password: Invalid password. Server rejected login attempt..'
+
 export const waitingKey = 'login:waiting'
 
 export const makeState: I.RecordFactory<Types._State> = I.Record({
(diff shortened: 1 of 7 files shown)


pzduniak:
Are we going to [eventually] rename `keybase passphrase recover`?

cjb:
@pzduniak Up to you! Having the help string point to "password recover" and accepting "passphrase recover" instead silently makes sense to me.

shared/constants/login.js
```diff
@@ -3,6 +3,9 @@ import * as I from 'immutable'
 import * as Types from './types/login'
 import HiddenString from '../util/hidden-string'
 
+// An ugly error message from the service that we'd like to rewrite ourselves.
+export const invalidPasswordErrorString = 'Bad password: Invalid password. Server rejected login attempt..'
```
patrickxb: One period instead of two?

patrickxb:
What about CLI?  I'm sure we say passphrase all over the place.

shared/constants/login.js
zapu: it's a bug that's been there for a while now, there is some formatting which takes one message that ends with period and adds another one like `fmt.Errorf("Bad password: %s.")`

PR #16769, opened 2019-03-29
Make sure all bundle changes update wallet state

patrickxb:
There's a bug where if you change your bundle (say change an account name) the wallet state can be stale for some time and send old data with its notifications to the frontend.

This PR fixes that so that any bundle changes update wallet state.

cc @buoyad

### DIFF
--- a/go/stellar/wallet_state.go
+++ b/go/stellar/wallet_state.go
@@ -120,6 +120,33 @@ func (w *WalletState) AccountName(accountID stellar1.AccountID) (string, error) 
 	return a.name, nil
 }
 
+// IsPrimary returns true if an account is the primary account for the user.
+func (w *WalletState) IsPrimary(accountID stellar1.AccountID) (bool, error) {
+	a, ok := w.accountState(accountID)
+	if !ok {
+		return false, ErrAccountNotFound
+	}
+
+	a.RLock()
+	defer a.RUnlock()
+
+	return a.isPrimary, nil
+}
+
+// AccountMode returns the mode of the account (USER or MOBILE).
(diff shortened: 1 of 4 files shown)


go/stellar/wallet_state.go
```diff
@@ -166,6 +206,36 @@ func (w *WalletState) Primed() bool {
...
+	defer mctx.TraceTimed(fmt.Sprintf("WalletState.UpdateAccountEntriesWithBundle [%s]", reason), func() error { return err })()
+
+	if bundle == nil {
+		return errors.New("nil bundle")
+	}
+
+	for _, account := range bundle.Accounts {
+		a, _ := w.accountStateBuild(account.AccountID)
+		a.updateEntry(account)
+	}
```
zapu: should it remove accounts that don't exist anymore from the state?

go/stellar/wallet_state.go
```diff
@@ -107,6 +107,46 @@ func (w *WalletState) BaseFee(mctx libkb.MetaContext) uint64 {
 	return w.options.BaseFee(mctx, w)
 }
 
+// AccountName returns the name for an account.
+func (w *WalletState) AccountName(accountID stellar1.AccountID) (string, error) {
```
zapu: so these are only used in tests right now? will they be used in other code as well?

patrickxb: Just in tests, but seems fine if they are needed elsewhere.  Currently they aren't...

go/stellar/wallet_state.go
patrickxb: ok

patrickxb: done.

PR #16776, opened 2019-03-29
[KBFS-4010] Stopgap for full disk caches

jzila:
This doesn't completely solve the problem, since under the hood the disk cache still won't evict. But at least we will not surface errors in caching to the user.

I also added logging to help debug what's happening. And I created a ticket to be able to obtain KBFS status on mobile.

### DIFF
--- a/go/kbfs/libkbfs/block_retrieval_queue.go
+++ b/go/kbfs/libkbfs/block_retrieval_queue.go
@@ -6,11 +6,12 @@ package libkbfs
 
 import (
 	"container/heap"
-	lru "github.com/hashicorp/golang-lru"
 	"io"
 	"reflect"
 	"sync"
 	"time"
+
+	lru "github.com/hashicorp/golang-lru"
 
 	"github.com/eapache/channels"
 	"github.com/keybase/client/go/kbfs/kbfsblock"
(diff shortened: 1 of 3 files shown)


go/kbfs/libkbfs/block_retrieval_queue.go
```diff
@@ -610,6 +611,8 @@ func (brq *blockRetrievalQueue) FinalizeRequest(
 		if err != nil {
 			brq.log.CDebugf(
 				retrieval.ctx, "Couldn't put block in cache: %+v", err)
+			// swallow the error if we were unable to put the block into caches.
```
jakob223: Do we want to do something else here eventually? If so, should we TODO and ticket?

Or are we planning to evict so this won't happen?

go/kbfs/libkbfs/disk_limiter.go
```diff
@@ -9,6 +9,23 @@ import (
 
 type diskLimitTrackerType int
 
+func (d diskLimitTrackerType) String() string {
```
jakob223: I love more understandable debug messages 😍

go/kbfs/libkbfs/block_retrieval_queue.go
jzila: I believe that it was never the intent to surface caching errors.

jzila:
Good to merge? I'd like to kick off an iOS build so we can get some testing over the weekend.

jakob223:
If you're asking me, 👌

PR #16781, opened 2019-03-29
a bunch of tweaks and fixes

songgao:
Each commit is individually reviewable. The first two commits are mostly refactoring:

* "don't show time-based sorting options in TLF list": cleans up the SortBar related types by just using a string for different sort options
* "split sortbar into its own component and have a top-bar around": now we have a TopBar component that includes the loading state and sorting, sort of like the group of actions in nav2 header. We are adding the sync toggle in the upcoming KBFS offline mode, so clean this up a bit to make it easier to work with.

The other two are basically styling fixes.


songgao:
Thanks John! I did test on both mobile and desktop. Will merge after CI passes.

PR #16782, opened 2019-03-29
Make a common-adapter confirmation modal

zanderz:
image


shared/common-adapters/index.js
```diff
@@ -21,6 +21,7 @@ export {default as BoxGrow} from './box-grow'
 export {default as Checkbox} from './checkbox'
 export {default as ChoiceList} from './choice-list'
 export {default as ClickableBox} from './clickable-box'
+// TODO: doesn't work: export {default as ConfirmModal} from './confirm-modal/index'
```
zanderz: Note: I spent hours stuck on this and never got it working - any suggestions?

songgao: Hmm what exactly doesn't work? `yarn flow` works for me here, although I'm getting `Cannot resolve module ./really-leave-team/index.stories.` in `teams/index.stories.js`. That file doesn't exist it seems.

shared/common-adapters/confirm-modal/index.js
```diff
@@ -0,0 +1,129 @@
...
+      backgroundColor: Styles.globalColors.white,
+      padding: Styles.globalMargins.xsmall,
+    },
+    isElectron: {
+      borderRadius: 4,
+    },
+  }),
+  container: Styles.platformStyles({
+    common: {
+      backgroundColor: Styles.globalColors.yellow3,
```
songgao: Is this really supposed to be yellow or is it just guideline color in the design mock up?

zanderz: are you suggesting the yellow isn't supposed to be yellow? 
https://app.zeplin.io/project/56a15d74e1e098dd3116d10b/screen/5c75bc9426c62e6244ba583e

adamjspooner: It's not meant to be yellow. The yellow is used to highlight the margins/padding.

shared/common-adapters/index.js
zanderz: idk, storybook just stopped working with no explanation when that was in

zanderz: oh and also, I had forgotten that file, sorry, it's there now

songgao: Oh I think I know why now. This is because in confirm-modal/index you are importing from `..`, which resolves to `common-adapters/index` and there's an import loop there. I tried importing those components locally in `confirm-modal` and it started working again.

shared/common-adapters/confirm-modal/index.js
```diff
@@ -0,0 +1,126 @@
...
+              <WaitingButton
+                fullWidth={true}
+                type="Secondary"
+                label="Cancel"
+                onClick={this.props.onCancel}
+                waitingKey={this.props.waitingKey}
+              />
+            )}
+            <WaitingButton
+              fullWidth={true}
```
songgao: This and above: `fullWidth` here means the full width available in its containing Box. Since there are two, I think this should just be `flex: 1`. Otherwise when the content changes, the buttons' width change too. (You can tell by switching to different confirm-modal stories.)

shared/teams/really-leave-team/index.js
```diff
@@ -1,19 +1,9 @@
...
-  MaybePopup,
-  ProgressIndicator,
-  ScrollView,
-  Text,
-  WaitingButton,
-} from '../../common-adapters'
-import {globalStyles, globalMargins, isMobile} from '../../styles'
+import {Avatar, Box, HeaderOnMobile, Icon, MaybePopup, ProgressIndicator} from '../../common-adapters'
+import ConfirmModal from '../../common-adapters/confirm-modal'
+import {globalStyles, globalMargins} from '../../styles'
```
songgao: Maybe switch these to `Kb` and `Styles` while you are here?

shared/common-adapters/confirm-modal/index.js
```diff
@@ -0,0 +1,126 @@
...
+  description: string,
+  header?: React.Node,
+  icon?: IconType,
+  onCancel: () => void,
+  onConfirm: () => void,
+  prompt: string,
+  waitingKey?: string,
+|}
+
+class _ConfirmModal extends React.Component<Props> {
```
songgao: Not a big deal, but is there a reason to use `React.Component` over `React.PureComponent`? This seems to be pure. Could also be a function component I suppose.

shared/common-adapters/confirm-modal/index.js
```diff
@@ -0,0 +1,126 @@
...
+            direction="vertical"
+            fullWidth={true}
+            fullHeight={true}
+            style={styles.inner}
+          >
+            {this.props.icon && (
+              <Icon
+                boxStyle={styles.icon}
+                color={Styles.globalColors.black_50}
+                fontSize={Styles.isMobile ? 64 : 48}
```
songgao: There's a `sizeType` now: https://github.com/keybase/client/blob/9de8767f5dec74c2171490fa4301b6f36469eadc/shared/common-adapters/icon.shared.js#L76-L87

But that doesn't cover this hugeness. Perhaps there should be a `"Huge"` type there? cc @adamjspooner

shared/teams/really-leave-team/index.js
```diff
@@ -33,45 +23,26 @@ const _Spinner = (props: Props) => (
...
-      </ButtonBar>
-    </ScrollView>
-  </MaybePopup>
+  <ConfirmModal
+    confirmText={`Yes, leave ${props.name}`}
+    description={`You will lose access to all the ${
+      props.name
+    } chats and folders, and you won't be able to get back
+    unless an admin invites you.`}
+    header={Header(props)}
```
songgao: Could also be `header={<Header {...props} />}`. That's a personal preference though I guess.

shared/common-adapters/confirm-modal/index.js
adamjspooner: What would that size be used for? cc @cecileboucheron

shared/common-adapters/index.js
zanderz: thanks Song, I think I got it

shared/common-adapters/confirm-modal/index.js
zanderz: ok now it's not yellow anymore

shared/common-adapters/confirm-modal/index.js
zanderz: I was wondering about that, thanks, should be ok now

shared/common-adapters/confirm-modal/index.js
zanderz: not at all, just copy and paste, will try changing it

shared/common-adapters/confirm-modal/index.js
zanderz: It didn't look like the design without that, IDK

shared/teams/really-leave-team/index.js
zanderz: ok

shared/common-adapters/confirm-modal/index.js
zanderz: I changed it to 'Big':
image

songgao: @zanderz  Sorry, I meant to suggest adding `Huge` to Kb.Icon's `sizeType` prop, and use that to specify the size you were using originally, then this fontSize can just be `sizeType="Huge"` instead. But I guess I'm also OK with having the fontSize for now. We can always change it afterwards if we end up adding `Huge` to `Kb.Icon`.

shared/teams/really-leave-team/index.js
zanderz: :white_check_mark:

shared/common-adapters/confirm-modal/index.js
zanderz: :white_check_mark:

shared/common-adapters/index.js
zanderz: :white_check_mark:

shared/common-adapters/confirm-modal/index.js
zanderz: :white_check_mark:

shared/teams/really-leave-team/index.js
zanderz: I thought I couldn't just import * from common-adapters because of the loop?

shared/common-adapters/confirm-modal/index.js
zanderz: no

shared/common-adapters/confirm-modal/index.js
```diff
@@ -38,9 +40,9 @@ class _ConfirmModal extends React.Component<Props> {
               <Icon
                 boxStyle={styles.icon}
                 color={Styles.globalColors.black_50}
-                fontSize={Styles.isMobile ? 64 : 48}
+                fontSize={Shared.typeToFontSize('Big')}
```
songgao: Sorry, I meant this would be `sizeType="Huge"`, and in case of `Big`, `sizeType="Big"`. Although it seems we don't have a conclusion if `Huge` should be added at this time, so to make it huge this would still needs to be what you had before `fontSize={Styles.isMobile ? 64 : 48}`.

zanderz: :white_check_mark:

zanderz:
Thanks Song!

PR #16783, opened 2019-03-30
add filter for folder view

songgao:
Also, both CMD+F and ESC work!

Screenshot in nav2:

image

image

### DIFF
--- a/shared/fs/common/folder-view-filter.js
+++ b/shared/fs/common/folder-view-filter.js
@@ -1,0 +1,127 @@
+// @flow
+import {namedConnect} from '../../util/container'
+import * as Platforms from '../../constants/platform'
+import * as Types from '../../constants/types/fs'
+import * as Kb from '../../common-adapters'
+import * as FsGen from '../../actions/fs-gen'
+import * as Styles from '../../styles'
+import * as React from 'react'
+
+type Props = {|
+  filter: string,
+  gap?: ?$Keys<typeof Styles.globalMargins>,
+  onUpdate: string => void,
+  path: Types.Path,
+|}
+
+type State = {|
+  editing: boolean,
+|}
+
+const KeyHandler = (Platforms.isMobile ? c => c : require('../../util/key-handler.desktop').default)(
+  () => null
+)
+
+class FolderViewFilter extends React.PureComponent<Props, State> {
+  state = {editing: false}
+
+  _input = React.createRef()
+
+  _onBlur = () => this.setState({editing: false})
+  _onFocus = () => this.setState({editing: true})
+  _onHotkey = (cmd: string) => {
+    cmd.endsWith('+f') && this._input.current && this._input.current.focus()
+  }
+  _onKeyDown = (e: SyntheticKeyboardEvent<>, isComposingIME: boolean) => {
+    if (e.key === 'Escape' && !isComposingIME) {
+      this.props.onUpdate('')
+      this._input.current && this._input.current.blur()
+    }
+  }
+
+  // Clear the filter if path changes, or if we get unmounted.
+  componentDidUpdate(prevProps) {
+    prevProps.path !== this.props.path && this.props.onUpdate('')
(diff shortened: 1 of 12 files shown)


songgao:
bump!

shared/fs/common/folder-view-filter.js
```diff
@@ -0,0 +1,127 @@
...
+            },
+          ])}
+          centerChildren={true}
+          gap="tiny"
+          gapStart={true}
+          gapEnd={true}
+        >
+          <Kb.Icon type="iconfont-search" color={Styles.globalColors.black_35} sizeType="Default" />
+          <KeyHandler onHotkey={this._onHotkey} hotkeys={Platforms.isDarwin ? ['command+f'] : ['ctrl+f']} />
+          <Kb.Input
```
buoyad: This should be `PlainInput` or `NewInput`.

shared/fs/common/index.stories.js
```diff
@@ -64,6 +65,14 @@ export const commonProvider = {
     enableDriver: Sb.action('enableDriver'),
     openInSystemFileManager: Sb.action('openInSystemFileManager'),
   }),
+  FolderViewFilter: ({path}: {path: Types.Path}) => ({
+    filter: folderViewFilter,
+    onUpdate: (filter: string) => {
+      Sb.action('onUpdate')(filter)
+      folderViewFilter = filter
```
buoyad: ooh, this is cool

shared/fs/common/upload-button.js
```diff
@@ -5,18 +5,33 @@ import * as Constants from '../../constants/fs'
...
   if (isDarwin) {
-    return <Kb.Button small={true} type="Primary" onClick={props.openAndUpload('both')} label="Upload" />
+    return (
+      <Kb.Button
+        small={true}
+        type="Primary"
+        onClick={props.openAndUpload('both')}
+        label="Upload"
+        style={
+          props.desktopButtonGap && {
```
buoyad: could this / `FolderViewFilter` just take a `style` prop?

shared/fs/common/index.stories.js
songgao: Doesn't work unfortunately; I forgot to tested after adding it. Will remove to avoid confusion ...

shared/fs/common/folder-view-filter.js
```diff
@@ -0,0 +1,127 @@
...
+
+class FolderViewFilter extends React.PureComponent<Props, State> {
+  state = {editing: false}
+
+  _input = React.createRef()
+
+  _onBlur = () => this.setState({editing: false})
+  _onFocus = () => this.setState({editing: true})
+  _onHotkey = (cmd: string) => {
+    cmd.endsWith('+f') && this._input.current && this._input.current.focus()
```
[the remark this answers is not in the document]
songgao: I tried but couldn't get to trigger anything when the Input is focused. No idea why it works for chat. Perhaps being used as a HOC makes a difference there?

songgao:
@buoyad ready for another look; thanks!

shared/fs/common/folder-view-filter.js
```diff
@@ -50,41 +56,40 @@ class FolderViewFilter extends React.PureComponent<Props, State> {
...
+              icon="iconfont-search"
+              hideBorder={true}
+              value={this.props.filter}
+              placeholder="Filter ..."
+              onChangeText={this.props.onUpdate}
+              onBlur={this._onBlur}
+              onFocus={this._onFocus}
+              onKeyDown={this._onKeyDown}
+              ref={this._input}
+              style={styles.input}
```
buoyad: you should set `containerStyle` to this too

shared/fs/common/index.stories.js
buoyad: I guess this makes sense. This "connector" doesn't get re-called since the store never changes

PR #16784, opened 2019-03-30
[KBFS-4010] disk_block_cache: Another log line for eviction

jzila:


go/kbfs/libkbfs/disk_block_cache.go
```diff
@@ -1039,6 +1052,8 @@ func (cache *DiskBlockCacheLocal) evictLocked(ctx context.Context,
 		// blockIDs is a slice of blocks from which evictions will be selected.
 		blockIDs := make(blockIDsByTime, 0, numElements)
 
+		cache.log.CDebugf(ctx, "There are %d TLFs in priority \"%s\" for eviction", len(shuffledSlice), priorityToEvict)
```
jakob223: If the cache is full and we're evicting each time we insert something, this line will be printed a lot of times. Are we okay with that? Is that going to pollute the logs too much?

jzila:
Closing because we know the bug. And this isn't it.

PR #16787, opened 2019-04-01
RevokeEngine,RevokeSigsEngine check merkle after posting sig

mlsteele:


go/libkb/merkle_tools.go
```diff
@@ -273,3 +273,24 @@ func VerifyMerkleRootAndKBFS(m MetaContext, arg keybase1.VerifyMerkleRootAndKBFS
 
 	return nil
 }
+
+// Verify that the given link has been posted to the merkle tree.
+// Used to detect a malicious server silently dropping sigchain link posts.
+func MerkleCheckPostedUserSig(mctx MetaContext, uid keybase1.UID,
+	seqno keybase1.Seqno, linkID LinkID) (err error) {
```
mlsteele: SigID may be better than LinkID. LinkID is what's convenient here since that's what UPAK's have.

mlsteele: max explained link IDs are better since with stubbed links in play link IDs are available when sig IDs may not be

PR #16788, opened 2019-04-01
Fix bad access in previewConversationFindExisting

mlsteele:


mlsteele:
fallout from `localFindConversationsLocalRpcPromise` using `callUntyped`

mlsteele:
Actually that's not right. I removed callUntyped and it works fine, so not sure why that was there. And flow still doesn't catch the problem. The typing looks right `uiConversations?: ?Array<InboxUIItem>`

mlsteele:
`yield` throws out the type information. `RPCChatTypes.localFindConversationsLocalRpcPromise(...)` has solid typing. But `results` in `const results = yield ...` is `any`

PR #16793, opened 2019-04-01
Signup stories 2: devicename and email

buoyad:
- Factor out screen structure into common `SignupScreen`
- Add banner for username taken case
- Enter device name + enter email screens

Next and last PR will be phone number screens
r? @keybase/react-hackers


buoyad:
Thanks @cjb! This is ready for another look.

PR #16795, opened 2019-04-01
previewConversationAfterFindExisting more flow types

mlsteele:
callPromise is the typed version. Now flow is in gear and caught some potential bugs.

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -1485,10 +1485,10 @@ function* previewConversationAfterFindExisting(state, action, results, users) {
     action.type === Chat2Gen.previewConversation && (action.payload.teamname || action.payload.channelname)
   if (action.type === Chat2Gen.previewConversation && action.payload.conversationIDKey) {
     existingConversationIDKey = action.payload.conversationIDKey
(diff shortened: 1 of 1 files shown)


shared/actions/chat2/index.js
```diff
@@ -1593,14 +1593,13 @@ function* previewConversationFindExisting(state, action) {
...
-    )
-    yield* previewConversationAfterFindExisting(state, action, results, users)
+    const resultMetas = (results.uiConversations || [])
+      .map(row => Constants.inboxUIItemToConversationMeta(row))
+      .filter(Boolean)
+    yield Saga.put(Chat2Gen.createMetasReceived({metas: resultMetas}))
+    yield* previewConversationAfterFindExisting(state, action, resultMetas, users)
   } else {
-    yield* previewConversationAfterFindExisting(state, action, undefined, [])
+    yield* previewConversationAfterFindExisting(state, action, [], [])
```
mmaxim: You sure this is ok?

mlsteele: The only usages of that `[]` value are in this diff. As `results` in `previewConversationAfterFindExisting`. And I tried a couple things. Should be ok

PR #16796, opened 2019-04-01
Add some basic template nav2 headers, we can improve them later

cjb:
@keybase/react-hackers

### DIFF
--- a/shared/wallets/routes.js
+++ b/shared/wallets/routes.js
@@ -167,27 +167,34 @@ class WalletsSubNav extends React.PureComponent<any> {
   }
 }
 
-const WalletsSubNavigator = createNavigator(
-  WalletsSubNav,
-  StackRouter(Shim.shim(walletsSubRoutes), {initialRouteName: 'wallet'}),
-  {}
-)
-
-WalletsSubNavigator.navigationOptions = {
-  header: undefined,
-  headerTitle: 'Wallet',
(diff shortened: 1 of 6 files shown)


shared/wallets/routes.js
```diff
@@ -167,19 +167,21 @@ class WalletsSubNav extends React.PureComponent<any> {
   }
 }
 
+const WalletsSubNavigator = createNavigator(
```
buoyad: Mobile gets mad about this
image

cjb:
@buoyad Oops, thanks! Maybe this, then?

shared/wallets/routes.js
```diff
@@ -167,21 +167,28 @@ class WalletsSubNav extends React.PureComponent<any> {
...
-}
-
 export const newRoutes = {
   'tabs.walletsTab': {
-    getScreen: () => isMobile ? require('./wallet/container').default : WalletsSubNavigator, upgraded: true,
+    getScreen: () => {
+      if (isMobile) {
+        return require('./wallet/container').default
+      } else {
+        const WalletsSubNavigator = createNavigator(
```
buoyad: Creating a new navigator in here strikes me as fishy - how often does react navigation call this?

cjb: Ah, looks like only once per launch: added a console.warn() and it only fired once.

PR #16797, opened 2019-04-01
PGP private key double prompt bug

zapu:
The issue here is that when you import PGP secret key via `keybase import --push-secret`, then when trying to decrypt, you end up with two SKBs - one local `LKSec` and one remote `Triplesec`. The order in which they are parsed during `UnlockKeyById` seems to be deterministic and the Triplesec one comes first.

So this is what happens when PGP decrypt is invoked through keybase after service is first ran (so passphrases are not cached yet):
1) run `keybase decrypt`, get to `GetKeysById` flow
2) find two SKBs, one from the server and one from local keyring
3) start working on first SKB which contains the key, it happens to be Triplesec
4) **get prompted for passphrase** because tsec/pps is not cached in passphrase stream yet
5) get to `SKB#UnlockSecretKey`, store cached tsec in passphrase cache, **but only tsec**
6) go to the second SKB, it's LKSec, try to get cached pps
7) oops, we don't have cached pps (despite asking for passphrase in 4), **prompt for passphrase again**.
8) Go to `SKB#UnlockSecretKey`, cache both tsec and pps after calling `UnverifiedPassphraseStream`

A proof of concept in this PR is to make SKB.UnlockSecretKey always use `UnverifiedPassphraseStream` (even for triplesec) and store both pps and tsec in cache, even when SKB is Triplesec and PPS isn't needed. I don't know whether this has any side effects or not though. And `UnlockSecretKey` is probably not the right place to do this.

### DIFF
--- a/go/libkb/skb.go
+++ b/go/libkb/skb.go
@@ -237,6 +237,24 @@ func (s *SKB) UnlockSecretKey(m MetaContext, passphrase string, tsec Triplesec, 
 	}
 	var unlocked []byte
 
+	fmt.Printf("SKB#UnlockSecretKey %d\n", s.Priv.Encryption)
+
+	if passphrase != "" && (tsec == nil || pps == nil) {
+		uid := s.uid
+		if uid.IsNil() {
+			uid = m.CurrentUID()
+		}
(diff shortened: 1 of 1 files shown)


zapu:
non-invasive fix would be to first process LKSec keys and later Triplesec

PR #16798, opened 2019-04-01
strip out pre-emptive offline checks, and sub in new IsOffline check 

mmaxim:
Patch does the following:

1.) Removes all `IsOffline` pre-emptive checks in `HybridConversationSource`, they cause more trouble than they are worth.
2.) Remove `identifyTLF`, since it doesn't do anything.
3.) Add a new `IsOffline` check that runs at the end of `GetThreadNonblock` and sets the offline bit on the result. Since we don't need to worry about blocking anything other than the current RPC, we can wait longer, and drop the "delayed" mechanic.


mmaxim:
Also the purpose of this is to avoid spurious gray offline bars when foregrounding the app.

PR #16800, opened 2019-04-01
Add SagaLogger to automatically log context in sagas

buoyad:
@keybase/react-hackers opening this as a draft to discuss. This adds a third parameter passed in to sagas by `chainAction` and `chainGenerator`, a `SagaLogger`. `SagaLogger` is a lil convenience wrapper for the four log fcns that prefixes them with the action type and the name of the saga function. 

Right now we add the context ourselves which 1. makes logging harder which makes us less likely to do it and 2. isn't in any consistent format which makes logs harder to read. This PR shortens some of our log messages with this in mind (e.g. log messages like `Error writing config stellar.lastSentXLM: ${err.message}` become `Error: ${err.message}`).

One caveat: originally this used `function.name` so all this context came to the saga writer for free. In minified code function names get mangled, so we can't do this. I added an optional last parameter to `chainAction` and `chainGenerator` for the name of the function. Most of the +diff comes from prettier expanding these all out into separate lines 😞


MarcoPolo:
This seems like it's useful. I think it's different than the saga logging middleware because it attaches context to the log line.

Can you give an example of what the log line would be outputted?

buoyad:
Here are some lines @MarcoPolo
```
3633 ["Info","2019-04-01T19:38:29.617Z","markThreadAsRead [route-tree:navigateTo]: marking read messages 0000d15d6c50fdd130a3ed3e2306afab8da162ffbd0f25fd03c7f2d3df8af591 40391"]
3634 ["Info","2019-04-01T19:38:29.625Z","loadMoreMessages [chat2:selectConversation]: calling rpc convo: 0000d15d6c50fdd130a3ed3e2306afab8da162ffbd0f25fd03c7f2d3df8af591 num: 20 reason: inboxBig"]
3635 ["Info","2019-04-01T19:38:29.718Z","markThreadAsRead [chat2:updateUnreadline]: marking read messages 0000d15d6c50fdd130a3ed3e2306afab8da162ffbd0f25fd03c7f2d3df8af591 40391"]
3636 ["Info","2019-04-01T19:38:29.851Z","markThreadAsRead [chat2:messagesAdd]: marking read messages 0000d15d6c50fdd130a3ed3e2306afab8da162ffbd0f25fd03c7f2d3df8af591 40391"]
3637 ["Info","2019-04-01T19:38:29.890Z","onNewChatActivity [engine-gen:chat1NotifyChatNewChatActivity]: Got new chat activity of type: 2"]
3638 ["Info","2019-04-01T19:38:30.251Z","queueMetaToRequest [chat2:metaNeedsUpdating]: skipping meta queue run, queue unchanged"]
3640 ["Info","2019-04-01T19:38:32.105Z","loadMoreMessages [chat2:selectConversation]: bail: no conversationIDKey"]
3641 ["Info","2019-04-01T19:38:32.105Z","getUnreadline [chat2:selectConversation]: Load unreadline bail: no conversationIDKey"]
3642 ["Info","2019-04-01T19:38:32.108Z","markThreadAsRead [route-tree:navigateUp]: marking read bail on not in meta list. preview?"]
3694 ["Info","2019-04-01T19:40:33.931Z","writeLastSentXLM [wallets:setLastSentXLM]: Writing config stellar.lastSentXLM: true"]
3695 ["Info","2019-04-01T19:40:33.986Z","markThreadAsRead [route-tree:navigateTo]: marking read bail on not in meta list. preview?"]
3696 ["Info","2019-04-01T19:40:34.109Z","onNewChatActivity [engine-gen:chat1NotifyChatNewChatActivity]: Got new chat activity of type: 1"]
```
Many of these log messages could be shorter now. (from the bottom of [this](https://keybase.io/_/admin/logdump_viewer?704619185#31c6cbae5e9c4e0e9451501c)) 

@chrisnojima yeah, I did see the existing middleware - what Marco said. The goal of this is to make `logger` more useful inside the sagas.

MarcoPolo:
Also not sure if you are aware of this: https://github.com/keybase/client/blob/master/shared/logger/action-transformer.js#L57, but we could use that to transform the action to a log-safe version so we can output more useful context

buoyad:
@MarcoPolo yeah I have seen that - do you think there's a way here to encourage people to write those more often?

buoyad:
@keybase/react-hackers this is ready for review!

- Add `SagaLogger` as third parameter to sagas
- Replace global logger uses with `SagaLogger` in chat, teams, and wallets

shared/util/saga.js
```diff
@@ -23,20 +45,23 @@ function* sequentially(effects: Array<any>): Generator<any, Array<any>, any> {
 // TODO I couldn't get flow to figure out how to infer this, or even force you to explicitly do it
 // maybe flow-strict fixes this
 type MaybeAction = void | boolean | TypedActions | null
-function* chainAction<Actions>(
+function* chainAction<Actions: {+type: string}>(
```
MarcoPolo: Does this really need to be covariant?

shared/util/saga.js
```diff
@@ -23,20 +45,23 @@ function* sequentially(effects: Array<any>): Generator<any, Array<any>, any> {
...
+    action: Actions,
+    logger: SagaLogger
+  ) => MaybeAction | $ReadOnlyArray<MaybeAction> | Promise<MaybeAction | $ReadOnlyArray<MaybeAction>>,
+  fcnTag?: string // tag for logger
 ): Generator<any, void, any> {
   type Fn = Actions => RS.Saga<void>
   return yield Effects.takeEvery<Actions, void, Fn>(pattern, function* chainActionHelper(
     action: Actions
   ): RS.Saga<void> {
+    const sl = new SagaLogger(action.type, fcnTag || 'unknown')
```
MarcoPolo: At first I thought maybe we could use f.name, but then I realized that would be useless when we minified it.

shared/util/saga.js
```diff
@@ -11,6 +12,27 @@ import {isArray} from 'lodash-es'
...
+    const prefix = `${fcnTag} [${actionType}]:`
+    this.debug = (...args) => logger.debug(prefix, ...args)
+    this.error = (...args) => logger.error(prefix, ...args)
+    this.info = (...args) => logger.info(prefix, ...args)
+    this.warn = (...args) => logger.warn(prefix, ...args)
+  }
+  // call this first in your saga if you want chainAction / chainGenerator to log
+  // before and after you run
+  tag = () => {
+    this.debug('->')
```
MarcoPolo: This part is `debug` but the other part is `info` on purpose?

shared/util/saga.js
```diff
@@ -54,24 +83,30 @@ function* chainAction<Actions>(
...
   return yield Effects.takeEvery<Actions, void, Fn>(pattern, function* chainActionHelper(
     action: Actions
   ): RS.Saga<void> {
+    const sl = new SagaLogger(action.type, fcnTag || 'unknown')
     try {
       const state = yield* selectState()
-      yield* f(state, action)
+      yield* f(state, action, sl)
+      if (sl.isTagged) {
+        sl.info('calling -> ok')
```
MarcoPolo: Why is this `calling -> ok` and the other one is ` -> ok` ?

shared/util/saga.js
```diff
@@ -11,6 +12,27 @@ import {isArray} from 'lodash-es'
...
+  constructor(actionType, fcnTag) {
+    const prefix = `${fcnTag} [${actionType}]:`
+    this.debug = (...args) => logger.debug(prefix, ...args)
+    this.error = (...args) => logger.error(prefix, ...args)
+    this.info = (...args) => logger.info(prefix, ...args)
+    this.warn = (...args) => logger.warn(prefix, ...args)
+  }
+  // call this first in your saga if you want chainAction / chainGenerator to log
+  // before and after you run
+  tag = () => {
```
MarcoPolo: Is this used at all right now?

MarcoPolo:
Did you think about the performance implication of adding this?

I think we'll pay for more space to store the new SagaLogger objects, but we pay for that at the beginning when we startup the sagas. After that I think it should be the same in terms of performance.

And the startup cost is probably pretty low.

does that sound right?

buoyad:
@MarcoPolo there's a runtime implication too, the `SagaLoggers` are created every time a saga runs, not just at the beginning. Should be easy to make them only once though, I'll make that change

shared/util/saga.js
buoyad: Yeah it was originally that :

shared/util/saga.js
buoyad: It's because `type` is covariant in the generated actions

shared/util/saga.js
buoyad: No - it was inspired by the service's log helper which makes lines like these
```
▶ [DEBU keybase conflicts.go:48] 8fc7 + GetConflictInfo(...) 
▶ [DEBU keybase teams.go:395] 8f6f - GetConflictInfo(...) -> ok
```
If we want we can add more to this in the future, like timing and generated tags.

buoyad:
Hm, actually, it's not as easy as I thought. At first I thought we could create them outside the `chainActionHelper` and then set the prefix inside, but sagas could step on each others' toes and the prefixes could get mixed up. So I think the best we can do is to make the `SagaLogger` outside and create new closures inside to use every time. Do you think that's worth it?

shared/util/saga.js
buoyad: Nope! Thanks

shared/util/saga.js
buoyad: Just messy, fixed!

buoyad:
- [ ] Merge after release

PR #16802, opened 2019-04-01
recover username cli for email or phone

joshblum:


joshblum:
cc @keybase/react-hackers for js change

protocol/avdl/keybase1/account.avdl
```diff
@@ -58,9 +58,9 @@ protocol account {
   GetLockdownResponse getLockdownMode(int sessionID);
   void setLockdownMode(int sessionID, boolean enabled);
 
-  // Call an API handler that will send an e-mail to the user with their
+  // Call an API handler that will send an e-mail or sms to the user with their
   // Keybase username.
-  void recoverUsername(int sessionID, string email);
+  void recoverUsername(int sessionID, string email, PhoneNumber phone);
```
zapu: it will be either username or phone, right? so maybe we can use a variant here

joshblum: the endpoint allows for one or both right now

go/client/cmd_account_recover_username.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2017 Keybase, Inc. All rights reserved. Use of
```
zapu: `2019`

go/client/cmd_account_recover_username.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func NewCmdAccountRecoverUsernameRunner(g *libkb.GlobalContext) *CmdAccountRecoverUsername {
+	return &CmdAccountRecoverUsername{Contextified: libkb.NewContextified(g)}
+}
+
+func (c *CmdAccountRecoverUsername) ParseArgv(ctx *cli.Context) error {
+	c.email = ctx.String("email")
+	c.phone = keybase1.PhoneNumber(ctx.String("phone"))
+	if len(c.email) == 0 && len(c.phone) == 0 {
```
zapu: do we want people using both at the same time? i know this is technically supported by the API endpoint, but it seems really weird to me

go/client/cmd_account_recover_username.go
```diff
@@ -0,0 +1,73 @@
...
+type CmdAccountRecoverUsername struct {
+	libkb.Contextified
+	email string
+	phone keybase1.PhoneNumber
+}
+
+func NewCmdAccountRecoverUsername(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	return cli.Command{
+		Name:  "recover-username",
+		Usage: "Recover username via email or sms.",
```
zapu: for the few occurrences of "SMS" I found in Keybase, it's always been uppercased

joshblum:
depends on https://github.com/keybase/keybase/pull/3655

go/client/cmd_account_recover_username.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func NewCmdAccountRecoverUsernameRunner(g *libkb.GlobalContext) *CmdAccountRecoverUsername {
+	return &CmdAccountRecoverUsername{Contextified: libkb.NewContextified(g)}
+}
+
+func (c *CmdAccountRecoverUsername) ParseArgv(ctx *cli.Context) error {
+	c.email = ctx.String("email")
+	c.phone = keybase1.PhoneNumber(ctx.String("phone"))
+	if (len(c.email) > 0 && len(c.phone) > 0) || (len(c.email) == 0 && len(c.phone) == 0) {
```
zapu: If you want to validate `c.phone`, there is a function: `libkb.IsPossiblePhoneNumber` that does a simple regexp to do entry-level validation on phone numbers, but it expects the format without `+` because it's used for SBS assertions.  It's fine to keep it though as is, but if it at least checked if it has a `+` and is numeric we could give a faster response to the user trying this command.

joshblum: that sounds good, none of the existing phone number handlers validated so I added checkers there as well.

PR #16808, opened 2019-04-02
Revert "fix native Text underline prop"

buoyad:
Reverts keybase/client#16789


buoyad:
@keybase/react-hackers

PR #16813, opened 2019-04-02
Handle autoreset in login flow / provisioning

pzduniak:


pzduniak:
Pending the serverside patch, I'll fix the remaining tests afterwards.

joshblum:
cc @patrickxb or @maxtaco

go/engine/login_provision.go
```diff
@@ -715,17 +718,30 @@ func (e *loginProvision) chooseDevice(m libkb.MetaContext, pgp bool) (err error)
...
+				return err
 			}
+		}
 
-			// they have pgp keys, so try that:
-			return e.tryPGP(m)
+		m.G().Log.Info(`
+The only way to provision this device is with access to one of your existing
+devices. You can try again later, or if you have lost access to all your
+existing devices you can reset your account and start fresh:`)
```
patrickxb: I think you should be using the UI to communicate with the user, otherwise things like this won't show up in the GUI.  This should be in ProvisionUI.

go/engine/account_reset.go
```diff
@@ -108,8 +115,18 @@ func (e *AccountReset) Run(mctx libkb.MetaContext) (err error) {
 		return err
 	}
 
+	if self {
+		eventType, readyTime, err := e.checkStatus(m)
+		if err != nil {
+			return err
+		}
+		if eventType != 0 && err == nil {
```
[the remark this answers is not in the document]
pzduniak: too many iterations :)

go/engine/account_reset.go
```diff
@@ -215,12 +212,19 @@ func (e *AccountReset) resetPrompt(m libkb.MetaContext, eventType int, readyTime
 		return nil
 	}
 
-	// Notify the user how much time is left
-	if err := m.UIs().LoginUI.DisplayResetProgress(m.Ctx(), keybase1.DisplayResetProgressArg{
-		Text: fmt.Sprintf(
+	// Notify the user how much time is left / if they can already reset
```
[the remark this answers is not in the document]
pzduniak: I don't understand, they haven't reset their account yet, it's just a notification that they _can_ perform if if they get into the flow using `keybase login` rather than `keybase account reset-start` (which we might remove).

go/engine/login_provision.go
pzduniak: It's a CLI-only thing, this text was previously displayed by CmdLogin.errProvisionUnavailable() - I kept it this way for now because we don't have any design stuff just yet. It will be functional as-is, we just won't have this explanation (we'll just see "Please enter the Keybase password for %s (%d+ characters) if you know it"). I don't know if it's worth changing it just yet.

pzduniak: ... and the more important aspect of the "lack of feedback in the GUI" is the sole fact that we don't show anything there _after_ you join the ARP. I could probably display it using the progress GUI call?

Since you're here, a somewhat relevant question: how will all of this work if the GUI doesn't implement the prompt methods in LoginUI? Will the calls error? I wanted to ask that stuff last week but we couldn't even get the actual reseting running.

go/engine/account_reset.go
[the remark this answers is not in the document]
patrickxb: sorry, it's just that "can already" together just doesn't make sense.  either "they can reset" or "they already reset".

The notificationText "Your account reset is ready!" sounds a little weird to me too.  It's like they ordered an account reset at a restaurant and it's ready for pickup.  How about just "Log in to finish resetting your account." or something like that.

go/engine/login_provision.go
patrickxb: They will be forced to handle any new functions added to a UI interface.  I believe `flow` will block it until they do.  When that happens, just reach out to @keybase/react-hackers and someone there will fix it up on their side.

go/engine/account_reset.go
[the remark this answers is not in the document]
joshblum: @pzduniak `keybase account reset-start` will likely supersede `keybase account reset` command (and be exposed in prod)  but not be removed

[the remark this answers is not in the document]
pzduniak: I've already fixed `keybase account reset-start` in this branch, wrt the wording I'll update it later, focusing on the actual broken tests now. I thought it sounded OK (I can find plenty of examples in newspapers etc.) but I'll trust your gut here.

pzduniak:
This branch's CI is pending the merge of https://github.com/keybase/keybase/pull/3676 - can we get an agreement on whether the session approach for this reset flow is OK? I'd like to make the tests pass in CI.

go/engine/login_provision.go
joshblum: until the GUI implements the flow and we want to enable autoresets on prod we should hide the new functionality behind feature flags

joshblum: @pzduniak if you don't move this over to provisionUI you should add a TODO with a ticket where it will be done

PR #16816, opened 2019-04-02
fix folder view filter and debounce updates

songgao:


shared/fs/common/folder-view-filter.js
```diff
@@ -6,31 +6,37 @@ import * as Kb from '../../common-adapters'
...
+  _update = text => {
+    this.setState({filter: text})
+    this.props.onUpdate(text)
+  }
 
   _input = React.createRef()
 
-  _onBlur = () => {
-    this.setState({editing: false})
-    this.props.onUpdate('')
```
songgao: This was the bug. We were sending the update in _onBlur, which happens when click on a row happens. So just move this into the Escape handler and only clear the filter if user presses escape when in the filter.

PR #16818, opened 2019-04-02
Allow images from localhost

mlsteele:
Allow images from `http://localhost`. `http://127.0.0.1` is already there. This lets icons on profiles work in dev. The service serves up localhost links using `SiteURILookup`.


mlsteele:
@chrisnojima

mlsteele:
Too risky, going another route

PR #16820, opened 2019-04-02
giphy previews on mobile

mmaxim:
Make one WebView to house all the previews.


go/chat/attachment_httpsrv.go
```diff
@@ -269,6 +294,98 @@ func (r *AttachmentHTTPSrv) serveUnfurlAsset(w http.ResponseWriter, req *http.Re
...
+	if err != nil {
+		r.makeError(context.TODO(), w, http.StatusInternalServerError, "failed to decode convID: %s",
+			err)
+		return
+	}
+	if err := r.G().ChatHelper.SendTextByID(ctx, convID, tlfName, url); err != nil {
+		r.makeError(context.TODO(), w, http.StatusInternalServerError, "failed to send giphy url: %s",
+			err)
+	}
+	ui, err := r.G().UIRouter.GetChatUI()
```
joshblum: log if err?

go/chat/attachment_httpsrv.go
```diff
@@ -269,6 +294,98 @@ func (r *AttachmentHTTPSrv) serveUnfurlAsset(w http.ResponseWriter, req *http.Re
...
+}
+
+func (r *AttachmentHTTPSrv) serveGiphyGallery(w http.ResponseWriter, req *http.Request) {
+	ctx := globals.ChatCtx(context.Background(), r.G(), keybase1.TLFIdentifyBehavior_CHAT_GUI, nil,
+		NewSimpleIdentifyNotifier(r.G()))
+	defer r.Trace(ctx, func() error { return nil }, "serveGiphyGallery")()
+	key := req.URL.Query().Get("key")
+	divStyle, err := url.QueryUnescape(req.URL.Query().Get("divstyle"))
+	if err != nil {
+		r.Debug(ctx, "serveGiphyGallery: failed to decode divstyle: %s", err)
```
joshblum: ~should `divStyle`/`vidStyle` be set to `""` or some sensible default if there is an error ?~ see comment below

shared/chat/conversation/giphy/index.native.js
```diff
@@ -1,8 +1,45 @@
...
+import * as Kb from '../../../common-adapters'
+import {NativeWebView} from '../../../common-adapters/native-wrappers.native'
+import * as Styles from '../../../styles'
 import type {Props} from './index.types'
 
-const GiphySearch = (props: Props) => {
-  return null
+class GiphySearch extends React.Component<Props, State> {
+  render() {
+    const vidStyle = 'height: 100%25 ;'
```
joshblum: i think this should be hardcoded on the results page instead of passing through as a URL param. 1. in makes the code cleaner since you don't have to parse it back out and 2. i feels weird/wrong to be echoing client params into HTML, can you set the `Styles.globalColors.black_20` in react move everything else to goland?

go/chat/commands/giphy.go
```diff
@@ -55,13 +53,21 @@ func (s *Giphy) getQuery(text string) *string {
 	return query
 }
 
+func (s *Giphy) getLimit() int {
+	limit := 25
+	if s.G().GetAppType() == libkb.MobileAppType {
```
joshblum:
```suggestion
	if s.G().IsMobileAppType()  {
```

shared/chat/conversation/giphy/index.native.js
mmaxim: If it is in Go, it makes changing it very time consuming, since you need to recompile everytime you want to tweak something. This way I can just reload the JS and get results right away.

PR #16827, opened 2019-04-03
Make scankeys process server synced keys after local keys (fix double prompt)

zapu:
This lets us prompt once and cache both tsec and pps if there is a combination of LKSec keys and TSec keys.


zapu:
crap i forgot, sorry. i'm on this

PR #16830, opened 2019-04-03
Various bugfixes for cancelling proof creation

buoyad:
Fixes a few cases where we could get stuck with an orphaned `startProof` rpc.
- Fix cancelling generic proofs while the service is polling for them (i.e. after you've been redirected to the site)
- Fix cancelling any proof on mobile when you navigate away via a screen gesture (sidestepping `onBack` calling `cancelProof`)

r? @keybase/react-hackers  cc @mlsteele for go changes


protocol/avdl/keybase1/prove_ui.avdl
```diff
@@ -53,6 +53,11 @@ protocol proveUi {
   // Used by parameterized proofs only
   void checking(int sessionID, string name);
 
+  // The proof is being polled, ask whether to continue checking.
+  // Returns whether to keep checking.
+  // Used by parametrized proofs only.
+  boolean continueChecking(int sessionID, string name);
```
mlsteele: `string name` look unused

shared/actions/profile/proofs.js
```diff
@@ -145,9 +146,32 @@ function* addProof(state, action) {
...
+  const watchForNavUp = yield Saga._fork(function*() {
+    // TODO remove this when nav1 is removed
+    // nav2 should use navigationOptions.gesturesEnabled = false
+    // to disable sidestepping custom behavior on back
+    while (true) {
+      yield Saga.take(RouteTreeGen.navigateUp)
+      const state = yield Saga.selectState()
+      const path = getPath(state.routeTree.routeState)
+      if (
+        ![
```
buoyad: navigating up is probably always invalid here; the point of this is just to make sure we don't leave a hanging RPC and the user unable to make new proofs.

protocol/avdl/keybase1/prove_ui.avdl
buoyad: yep! do you think passing in `0` as sessionID in `engine/prove` is better?

mlsteele: yes and better yet you can, in engine/prove, call `m.UIs().ProveUI.ContinueChecking(m.Ctx())`. And change `ContinueChecking` in service and interfaces to take no arg.

shared/actions/profile/proofs.js
```diff
@@ -145,9 +146,32 @@ function* addProof(state, action) {
     }
   })
 
+  const watchForNavUp = yield Saga._fork(function*() {
+    // TODO remove this when nav1 is removed
```
cjb: We're probably not going to notice this TODO later, but if you put a mention of flags.useNewRouter in here that'd show up when we grep to remove the feature flag.. even if that mention's inside a comment, I guess!

buoyad: Good point, thanks, I'll add it

buoyad: @cjb also swapped out the call to `checkProof` with a call to `Identify3` when `startProof` finishes

PR #16832, opened 2019-04-03
fix sortbar displaying error

songgao:


shared/reducers/fs.js
```diff
@@ -125,8 +125,11 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
     case FsGen.setFolderViewFilter:
       return state.set('folderViewFilter', action.payload.filter)
     case FsGen.sortSetting:
-      const {path, sortSetting} = action.payload
-      return state.setIn(['pathUserSettings', path, 'sort'], sortSetting)
```
songgao: This was the bug. It resulted in a `I.Map` instead of `I.Record`. Little did I know a year ago!

PR #16835, opened 2019-04-04
have the gui stop the service via RPC instead of shelling out

zanderz:
I can't test this yet because the widget menu hamburger has disappeared on me.
UPDATE: tested it, it works. Hamburger problem is a separate GUI issue.


maxtaco:
Can’t be me but why this change? Is there a jira ticket?

zanderz:
No, I just wanted to make this right for Surya.

heronhaye:
So this will call the ctl.Stop RPC? Ok, looks good to me (idk if you want someone on frontend to review as well though). I'll fixup my PR after this is merged

zanderz:
This works for me now but it should have an official desktop review.

shared/actions/settings.js
```diff
@@ -479,6 +479,9 @@ const loadHasRandomPW = state =>
 // Mark that we are not randomPW anymore if we got a password change.
 const passwordChanged = () => SettingsGen.createLoadedHasRandomPw({randomPW: false})
 
+// $FlowIssue
+const stop = (_, action) => RPCTypes.ctlStopRpcPromise(action.payload.exitCode)
```
buoyad: I think flow complains because the param should be `{exitCode: action.payload.exitCode}`

zanderz: whoops, thanks again Danny!
This should wait till after code freeze though right?

heronhaye: I was kind of hoping to get this in for the release. But if it's risky we can bump it to the next one.

buoyad: Seems a little risky, I'd say let it wait

PR #16836, opened 2019-04-04
Fix bug where profile search can't be closed

akalin-keybase:
Happens when instantiated via 'Browse people' button.

### DIFF
--- a/shared/profile/search/container.js
+++ b/shared/profile/search/container.js
@@ -1,21 +1,45 @@
 // @flow
 import Search from '.'
 import {createShowUserProfile} from '../../actions/profile-gen'
-import {connect} from '../../util/container'
+import {connect, type RouteProps} from '../../util/container'
 
-type OwnProps = {|onClose: () => void|}
+// Either onClose is passed in from bar.js or we're instantiated via a
+// route.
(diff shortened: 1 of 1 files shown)


shared/profile/search/container.js
```diff
@@ -1,21 +1,45 @@
 // @flow
 import Search from '.'
 import {createShowUserProfile} from '../../actions/profile-gen'
-import {connect} from '../../util/container'
+import {connect, type RouteProps} from '../../util/container'
 
-type OwnProps = {|onClose: () => void|}
+// Either onClose is passed in from bar.js or we're instantiated via a
+// route.
+type OwnProps = {|onClose: () => void|} | RouteProps<{}, {}>
```
akalin-keybase: @cjb Ah, I see what you mean here. I only need it for `ownProps.navigateUp`, but maybe I should just use `dispatch(RouteTreeGen.createNavigateUp())`?

cjb: yup!

akalin-keybase: Fixed!

akalin-keybase:
PTAL

akalin-keybase:
bump!

PR #16837, opened 2019-04-04
Make 'add to channel' work with nav2

akalin-keybase:


shared/chat/conversation/info-panel/add-to-channel/container.js
```diff
@@ -29,8 +29,12 @@ const mapStateToProps = (state, {routeProps}) => {
 }
 
 const mapDispatchToProps = dispatch => ({
-  _onSubmit: (conversationIDKey, usernames) =>
-    dispatch(Chat2Gen.createAddUsersToChannel({conversationIDKey, usernames})),
+  _onSubmit: (conversationIDKey, usernames) => {
+    dispatch(Chat2Gen.createAddUsersToChannel({conversationIDKey, usernames}))
+    if (flags.useNewRouter) {
```
buoyad: I don't think this is a good way to make this work with nav2 since we lose the waiting state. Is there a way this component can figure out the RPC call has finished and nav away?

akalin-keybase: Ah, you're right! Did a clearModals in the RPC action instead.

akalin-keybase:
PTAL!

PR #16838, opened 2019-04-04
Use small buttons in nav2 headers

cjb:
@keybase/react-hackers


cjb:
@keybase/react-hackers bumpity

PR #16844, opened 2019-04-04
Use new transfer icon.

adamjspooner:


adamjspooner:
@keybase/react-hackers, bump.

shared/wallets/transaction/index.js
```diff
@@ -37,7 +38,20 @@ const CounterpartyIcon = (props: CounterpartyIconProps) => {
...
+          direction="horizontal"
+          style={collapseStyles([styles.transferIconContainer, {width: size}])}
+        >
+          <Icon
+            color={globalColors.purple2}
+            sizeType={props.detailView ? 'Bigger' : 'Big'}
+            style={collapseStyles([!props.detailView && styles.transferIcon])}
+            type="iconfont-wallet-transfer"
+          />
+        </Box2>
```
songgao: Can we achieve the same result by using `padding="xtiny"` instead of this added Box2?

shared/common-adapters/icon.shared.js
```diff
@@ -75,6 +75,8 @@ export function fontSize(type: IconType): ?Object {
 
 export function typeToFontSize(sizeType: SizeType) {
   switch (sizeType) {
+    case 'Bigger':
+      return Styles.isMobile ? 48 : 36
```
songgao: If we can have `Huge` too that'd be great!

adamjspooner: Let's add that when we need it (in a relevant PR).

shared/wallets/transaction/index.js
adamjspooner: Maybe, but we also need the icon to be a set size and a container with a different width to line it up with the other icons.

shared/common-adapters/icon.shared.js
songgao: (Resolved on chat that we should add it in PRs using it.)

PR #16847, opened 2019-04-04
go/libdokan: Handle .kbfs_unmount logger shutdown

taruti:


jakob223:
Hm, is there an equivalent place in `kbfsfuse` where we should be doing this?

taruti:
Nothing similar in `libfuse`, the kludge was only needed on Windows.

PR #16848, opened 2019-04-04
folder: add offline prop to folder and make component

jakob223:
I made a component and a story shot for the component and a prop to the `Folder` component, but I'm not sure if the scope of this ticket includes calling that prop from somewhere. It seems to me like it probably should, but I'm not really sure how to get started on that or if the framework is in place yet.

Issue: KBFS-4018

### DIFF
--- a/shared/fs/folder/container.js
+++ b/shared/fs/folder/container.js
@@ -9,6 +9,8 @@ import * as FsGen from '../../actions/fs-gen'
 const mapStateToProps = (state, {path}) => ({
   _pathItem: state.fs.pathItems.get(path, Constants.unknownPathItem),
   _username: state.config.username,
+  // TODO: figure out how to get this info
+  offline: false,
   resetBannerType: Constants.resetBannerType(state, path),
   shouldShowSFMIBanner: state.fs.sfmi.showingBanner,
 })
@@ -20,6 +22,7 @@ const mapDispatchToProps = (dispatch, {path}: OwnProps) => ({
 })
 
 const mergeProps = (stateProps, dispatchProps, {path, routePath}) => ({
+  offline: stateProps.offline,
   onAttach: stateProps._pathItem.writable ? dispatchProps.onAttach : null,
   path,
   resetBannerType: stateProps.resetBannerType,
--- a/shared/fs/folder/container.js
+++ b/shared/fs/folder/container.js
@@ -5,12 +5,12 @@ import Folder from '.'
 import * as Types from '../../constants/types/fs'
 import * as Constants from '../../constants/fs'
 import * as FsGen from '../../actions/fs-gen'
+import flags from '../../util/feature-flags'
 
 const mapStateToProps = (state, {path}) => ({
   _pathItem: state.fs.pathItems.get(path, Constants.unknownPathItem),
   _username: state.config.username,
-  // TODO: figure out how to get this info
-  offline: false,
+  offline: !state.fs.kbfsDaemonStatus.online,
   resetBannerType: Constants.resetBannerType(state, path),
   shouldShowSFMIBanner: state.fs.sfmi.showingBanner,
 })
@@ -22,7 +22,11 @@ const mapDispatchToProps = (dispatch, {path}: OwnProps) => ({
 })
 
 const mergeProps = (stateProps, dispatchProps, {path, routePath}) => ({
-  offline: stateProps.offline,
+  offline:
+    flags.kbfsOfflineMode &&
+    stateProps.offline &&
+    Types.getPathLevel(path) > 2 &&
+    stateProps._pathItem.prefetchStatus !== Constants.prefetchComplete,
   onAttach: stateProps._pathItem.writable ? dispatchProps.onAttach : null,
   path,
   resetBannerType: stateProps.resetBannerType,
--- a/shared/fs/folder/container.js
+++ b/shared/fs/folder/container.js
@@ -5,12 +5,11 @@ import Folder from '.'
 import * as Types from '../../constants/types/fs'
 import * as Constants from '../../constants/fs'
 import * as FsGen from '../../actions/fs-gen'
-import flags from '../../util/feature-flags'
 
 const mapStateToProps = (state, {path}) => ({
   _pathItem: state.fs.pathItems.get(path, Constants.unknownPathItem),
   _username: state.config.username,
-  offline: !state.fs.kbfsDaemonStatus.online,
+  offline: Constants.isOfflineUnsynced(state.fs.kbfsDaemonStatus, state.fs.pathItems, path),
   resetBannerType: Constants.resetBannerType(state, path),
   shouldShowSFMIBanner: state.fs.sfmi.showingBanner,
 })
@@ -22,11 +21,7 @@ const mapDispatchToProps = (dispatch, {path}: OwnProps) => ({
 })
 
 const mergeProps = (stateProps, dispatchProps, {path, routePath}) => ({
-  offline:
-    flags.kbfsOfflineMode &&
-    stateProps.offline &&
-    Types.getPathLevel(path) > 2 &&
-    stateProps._pathItem.prefetchStatus !== Constants.prefetchComplete,
+  offline: stateProps.offline,
   onAttach: stateProps._pathItem.writable ? dispatchProps.onAttach : null,
   path,
   resetBannerType: stateProps.resetBannerType,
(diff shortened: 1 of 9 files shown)


shared/fs/folder/offline.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+import * as Kb from '../../common-adapters'
+import React from 'react'
+import * as Styles from '../../styles/index'
+
+// TODO: what is the correct thing to do when don't have any props
+class OfflineFolder extends React.PureComponent<any> {
+  render() {
```
songgao: You can have an empty `Props` like this: `type Props = {||}`

Also, this doesn't have state or any other stuff that we need from a class component, so ideally it should be a function component:

```
const OfflineFolder = (props: Props) => (
  <Kb.Box2 ...
)
```

shared/fs/folder/offline.js
```diff
@@ -0,0 +1,33 @@
...
+    return (
+      <Kb.Box2 direction="vertical" fullHeight={true} fullWidth={true}>
+        <Kb.Box2
+          direction="vertical"
+          style={styles.emptyContainer}
+          fullHeight={true}
+          fullWidth={true}
+          centerChildren={true}
+        >
+          <Kb.Icon type="iconfont-cloud" fontSize={64} color={Styles.globalColors.black_10} />
```
songgao: I'm guessing this needs to be `48` on mobile. But really we should probably have `Huge` as `sizeType` for the `Kb.Icon`. cc @adamjspooner thoughts?

shared/fs/folder/offline.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+import * as Kb from '../../common-adapters'
+import React from 'react'
+import * as Styles from '../../styles/index'
+
+// TODO: what is the correct thing to do when don't have any props
+class OfflineFolder extends React.PureComponent<any> {
+  render() {
+    return (
+      <Kb.Box2 direction="vertical" fullHeight={true} fullWidth={true}>
```
songgao: I know `SelfReset` does that, but I think both this and that should just have a `flex: 1` style instead of having `fullHeight={true}`. The reason is there are other stuff in the same outer container with this, and having `fullHeight={true}` here makes it `width: '100%'` which would squeeze other stuff. (`WithContent` above is a an example.)

I guess an alternative is to move [this](https://github.com/keybase/client/pull/16848/files#diff-6df9d6db4b7ccf7cbc83bcd2ba7ee713R65) into `WithContent`, then it'd be the only children of its parent and `fullHeight` would be fine.

songgao: Also, we need to add the top-bar here, for the the Sync toggle. It's not there yet (in PR https://github.com/keybase/client/pull/16859/files#diff-493e2e04422d21c2ab95f3ec21f9e4dfR24). We'll also need to have a way to disable sort in this case.

shared/fs/folder/container.js
```diff
@@ -9,6 +9,8 @@ import * as FsGen from '../../actions/fs-gen'
 const mapStateToProps = (state, {path}) => ({
   _pathItem: state.fs.pathItems.get(path, Constants.unknownPathItem),
   _username: state.config.username,
+  // TODO: figure out how to get this info
+  offline: false,
```
songgao: This is in `songgao/KBFS-4019-KBFS-4020` too, but I guess we can also connect this in a separate PR after that one's merged.

shared/fs/folder/offline.js
jakob223: I found where to add 'Huge' as an option, but I can't find where the definitions mapping names to numbers are. Where can I find that?

shared/fs/folder/container.js
jakob223: I'd like to connect it but I don't know where to start.

shared/fs/folder/offline.js
jakob223: Oh, saw `typeToFontSize` on another PR.

shared/fs/folder/container.js
songgao: If you rebase this over `songgao/KBFS-4019-KBFS-4020`, this could probably be `kbfsDaemonStatus: state.fs.kbfsDaemonStatus` and then in mergeProps, you can combine this with the prefetchStatus from stateProps._pathItem.

songgao: Oh just realized you edited the comment. Sounds good!

jakob223:
Added top bar (without sort) to the offline folder, and I think I covered all the cases for the `offline` prop.

shared/fs/folder/offline.js
```diff
@@ -0,0 +1,40 @@
...
+import * as Types from '../../constants/types/fs'
+import TopBar from '../top-bar'
+
+type Props = {|
+  path: Types.Path,
+|}
+
+const OfflineFolder = (props: Props) => (
+  <Kb.Box2 direction="vertical" style={styles.contentContainer} fullWidth={true} alignItems={'stretch'}>
+    <Kb.BoxGrow>
```
songgao: It doesn't seem like this `BoxGrow` is necessary here. We need it in the folder view because it needs to know the height and BoxGrow makes a positioned div and gives that. In this case I think `flex: 1` would just work. Have you tried removing `<Kb.BoxGrow>`?

shared/fs/folder/offline.js
```diff
@@ -0,0 +1,40 @@
...
+      <TopBar path={props.path} mode={'offline'} />
+      <Kb.Box2
+        direction="vertical"
+        style={styles.emptyContainer}
+        fullHeight={true}
```
songgao: For the same reason as before, this can't be `fullHeight`. Otherwise it'd try to push the `TopBar`. Just use `flex: 1`.

shared/fs/top-bar/index.js
```diff
@@ -38,7 +39,7 @@ const TopBar = (props: Props) =>
       {Types.getPathLevel(props.path) === 3 && <SyncToggle tlfPath={props.path} />}
       <Kb.Box style={styles.flex} />
       <Loading path={props.path} />
-      <Sort path={props.path} />
+      {props.mode !== 'offline' && <Sort path={props.path} />}
```
songgao: It'd be nice to avoid this sort of plumbing. Perhaps [this](https://github.com/keybase/client/pull/16848/files#diff-088e08f0ca0e38cd16c0b120a3ebc514R25) can be moved into `Constants`, and we can use the same logic in `sort-container.js` where it also deals with `isEmpty`?

jakob223: Good point; changed.

shared/fs/folder/container.js
```diff
@@ -5,12 +5,11 @@ import Folder from '.'
 import * as Types from '../../constants/types/fs'
 import * as Constants from '../../constants/fs'
 import * as FsGen from '../../actions/fs-gen'
-import flags from '../../util/feature-flags'
 
 const mapStateToProps = (state, {path}) => ({
   _pathItem: state.fs.pathItems.get(path, Constants.unknownPathItem),
   _username: state.config.username,
-  offline: !state.fs.kbfsDaemonStatus.online,
+  offline: Constants.isOfflineUnsynced(state.fs.kbfsDaemonStatus, state.fs.pathItems, path),
```
songgao: We should try to do as less as possible in `mapStateToProps` since any update to the redux store needs to iterate through all of these connected components. This is fast so it probably won't make a huge difference, but you already have `_pathItem` above; why now have `kbfsDaemonStatus` here and compute `isOfflineUnsyunced` in `mergeProps`?

jakob223: I considered that - the reason I went with computing it here is that `Constants.isOfflineUnsynced` is called elsewhere, from somewhere we don't have the `pathItem`. The alternative that I had thought of, which seemed worse, was to return all the `pathItems` from `mapStateToProps`, but I elected not to do that since it would make the component render too often.

As a third alternative, I could change `Constants.isOfflineUnsynced` to take a `pathItem` and make the sort bar get the path item when it needs it. I'll do that.

songgao: > As a third alternative, I could change `Constants.isOfflineUnsynced` to take a pathItem and make the sort bar get the path item when it needs it. I'll do that.

Yeah this sounds like the right approach. `Constants.isOfflineUnsynced` only needs a single `pathItem`, so it should just take that instead of `state.fs.pathItems`.

PR #16804, opened 2019-04-01
 kbfs offline in GUI: redux, saga, and new RPCs

songgao:

PR #16850, opened 2019-04-04
SyncStatus component

jakob223:
Not hooked up to anything. Not sure if that was part of the scope of this ticket - if it is, would be happy to work on that.

The animation part was split off into KBFS-4039.

This also might need a few tweaks to get it to pass lint after the base branch songgao/KBFS-4019-KBFS-4020 is merged into master (#16804).

Issue: KBFS-4012


shared/fs/common/index.stories.js
```diff
@@ -248,7 +248,20 @@ const load = () => {
...
     ))
+    .add('Sync Status', () => (
+      <Kb.Box2 direction="vertical" gap="large" gapStart={true} fullWidth={false} alignItems={'center'}>
+        <SyncStatus status={'awaiting-to-sync'} folder={false} />
+        <SyncStatus status={'awaiting-to-upload'} folder={false} />
+        <SyncStatus status={'online-only'} folder={false} />
+        <SyncStatus status={'synced'} folder={false} />
+        <SyncStatus status={'sync-error'} folder={true} />
+        <SyncStatus status={'uploading'} folder={false} />
+        <SyncStatus status={0.3} folder={false} />
```
songgao: Just so you know, something like this would also work:

```
['awaiting-to-sync', 'awaiting-to-upload', ......].map(status => <SyncStatus status={status} folder={folder} />)
```

But there's not a lot of props here so what you have is good too. (i.e. not suggesting a change!)

shared/fs/sync-status/index.js
```diff
@@ -0,0 +1,80 @@
...
+      return 'iconfont-upload'
+    case 'online-only':
+      return 'iconfont-cloud'
+    case 'synced':
+      return 'iconfont-proof-good'
+    case 'sync-error':
+      return 'iconfont-exclamation'
+    default:
+      // TODO: handle icons for the pie timer thing
+      return 'iconfont-question-mark'
```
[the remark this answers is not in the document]
jakob223: I tried that and flow complained about it. I could be doing something wrong, but I assumed as a result that flow wasn't smart enough to realize that the switch and the if together were handling all cases.

shared/fs/common/index.stories.js
```diff
@@ -248,7 +249,20 @@ const load = () => {
...
+      </Kb.Box2>
+    ))
 
+  // awaiting-to-sync, awaiting-to-upload, online-only, synced, error,
+  // uploading, syncing
```
songgao: It seems this was left here by accident?

shared/fs/sync-status/container.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+
```
songgao: 🧨

shared/fs/sync-status/container.js
```diff
@@ -0,0 +1,40 @@
...
+    stateProps._kbfsDaemonStatus,
+    stateProps._tlf,
+    stateProps._pathItem,
+    stateProps._uploads,
+    ownProps.path
+  )
+  return {
+    folder: stateProps._pathItem.type === 'folder',
+    status: status,
+  }
```
songgao: Seems like this could be in short-hand form by inlining `status`, but up to you!

shared/fs/sync-status/container.js
jakob223: I don't have that emoji yet!

songgao: Sounds like you needed a macOS upgrade 😉

songgao:
So I started to add this to various places, and realized it'd be nice to move `fs/sync-status` into `fs/common/sync-status` (either inside its own dir or as two files), and export that in `fs/common/index.js`. Also now that we have the feature flag (`kbfsOfflineMode`), we can check for that flag in the container, and return `()=>null` in container if the flag is disabled. Does it sound good?

shared/fs/common/sync-status-container.js
```diff
@@ -28,13 +27,13 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
...
 }
 
 export default namedConnect<OwnProps, _, _, _, _>(
   mapStateToProps,
-  mapDispatchToProps,
-  mergeProps,
-  'SyncStatus'
+    mapDispatchToProps,
+    mergeProps,
+    'SyncStatus'
```
jzila: I don't think this spacing is right.

jakob223:
I have rebased on master and fixed all the comments. Will merge when CI passes.

shared/fs/common/index.js
```diff
@@ -12,3 +12,4 @@ export {default as OpenChat} from './open-chat'
 export {default as NewFolder} from './new-folder'
 export {default as UploadButton} from './upload-button'
 export {default as FolderViewFilter} from './folder-view-filter'
+export {default as SyncStatus} from './sync-status'
```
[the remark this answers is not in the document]
jakob223: Thanks 🎶 , fixed now.

PR #16853, opened 2019-04-05
Android: fix phantom android notifications

jzila:


jzila:
I'm going to merge this since it's an important fix.

PR #16854, opened 2019-04-05
[KBFS-3967] Android: share from cold start

jzila:


jzila:
I _think_ now that I fixed the routing, I might be able to get away with removing the "on start" code, leaving only the emitter. Will test though.

shared/react-native/android/app/src/main/java/io/keybase/ossifrage/modules/IntentHandler.java
```diff
@@ -163,15 +167,14 @@ public String getName() {
     }
 
     @ReactMethod
-    public void handlePushNotification(String convID, String payload, Integer membersType,
```
mmaxim: Why does this get deleted, was it unused?

jzila: Yep, unused.

PR #16857, opened 2019-04-05
add "you are offline" banner for KBFS offline mode

songgao:
also add a bunch of stories for offline view.


buoyad:
- With the feature flag off the banner is always up
- I got a black bar after coming back online the first time I tried: `ERROR CODE 218 - tls: use of closed connection in method keybase.1.SimpleFS.simpleFSFolderSyncConfigAndStatus`. Think it's related?

songgao:
The first one is I forgot to add that flag here after I added the flag. Good catch! Just pushed a commit to handle that. Second one seems to be because we didn't handle errors in offline state properly. I'll created KBFS-4058 to go over them.

PR #16858, opened 2019-04-05
add Kb.Switch

songgao:
Design: 

https://zpl.io/anYqBvQ

https://zpl.io/aM9AwxJ


shared/common-adapters/switch.stories.js
```diff
@@ -0,0 +1,69 @@
...
+        on={this.state.on}
+        onClick={this._onClick}
+        style={{width: 320}}
+      />
+    )
+  }
+}
+
+const load = () => {
+  Sb.storiesOf('Common', module).add('Switch', () => (
```
buoyad: Not scrolling on desktop, `scrollViewDecorator`?

shared/common-adapters/switch.js
```diff
@@ -0,0 +1,77 @@
...
+  disabled?: ?boolean,
+  label: string,
+  on: boolean,
+  onClick: () => void,
+  style?: ?Styles.StylesCrossPlatform,
+|}
+
+const Switch = React.forwardRef<Props, Kb.ClickableBox>((props: Props, ref) => (
+  <Kb.ClickableBox
+    onClick={props.onClick}
```
buoyad: Need to disable this on `props.disabled`

shared/common-adapters/switch.js
```diff
@@ -0,0 +1,77 @@
...
+  ClickableBox,
+  ProgressIndicator,
+  Text,
+}
+
+type Props = {|
+  align?: ?('left' | 'right'), // default to 'left'
+  color?: ?('blue' | 'green'), // default to 'blue'
+  disabled?: ?boolean,
+  label: string,
```
buoyad: would be nice if this were `string | React.Node` for labels like this:
image

shared/common-adapters/switch-toggle.native.js
```diff
@@ -0,0 +1,65 @@
+// @flow
+import React from 'react'
+import {NativeAnimated, NativeEasing} from './native-wrappers.native'
+import * as Styles from '../styles'
+import {type Props} from './switch-toggle'
+
+class SwitchToggle extends React.PureComponent<Props, {||}> {
```
buoyad: `checkbox.native` looks like this, but is only ever blue left-aligned. it'd be cool if Checkbox used this on mobile so they looked the same. cc @cecileboucheron what do you think?

shared/common-adapters/switch.stories.js
```diff
@@ -0,0 +1,69 @@
...
+    <Kb.Box2 direction="vertical" gap="large" gapStart={true} centerChildren={true} fullWidth={true}>
+      {['left', 'right'].map(align =>
+        ['blue', 'green'].map(color =>
+          [true, false, 'disabled'].map(i => {
+            const initialEnabled = !!i
+            const label = `${align} aligned; ${initialEnabled ? 'enabled' : 'disabled'}; ${color}${
```
buoyad: two `disabled`s in the labels is confusing
```suggestion
            const label = `${align} aligned; ${initialEnabled ? 'on' : 'off'}; ${color}${
```

shared/common-adapters/switch-toggle.native.js
```diff
@@ -0,0 +1,65 @@
...
+  _offset = new NativeAnimated.Value(this._getOffset())
+  _animation = null
+
+  _getOffset() {
+    return this.props.on ? enabledOffset : disabledOffset
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    if (prevProps.on !== this.props.on) {
```
buoyad: should the rest of this function move inside this block too?

shared/common-adapters/switch-toggle.native.js
songgao: Yeah I was wondering about that too but forgot. Thanks for bringing it up!

jzila:
How are the settings switches implemented right now? Should we refactor those to this common adapter?

songgao:
> How are the settings switches implemented right now?

@jzila We don't have that anywhere now. The only thing we have is a CheckBox, which is a toggle on mobile and checkbox on desktop. So I added a new common adapter component `Kb.Switch` in this PR and the Files sync switch is actually using that.

jzila:
Got it, and that's a toggle everywhere, right?

So do we want to switch all the checkboxes to this form?

songgao:
> Got it, and that's a toggle everywhere, right?

With `Kb.Switch`, yes.

> So do we want to switch all the checkboxes to this form?

Not sure about switching checkbox on desktop; that'd be up to design. What Danny was suggesting was to switch the mobile part of the CheckBox component (which is already a toggle) into the same style of this new `Kb.Switch` component. They look a bit different but the behavior is basically the same.

jzila:
@cecileboucheron WDYT about moving all the desktop checkboxes to the new Switch design?

shared/common-adapters/switch.js
songgao: Sounds good; I'm gonna also add a `labelSubtitle?: ?string` so it'll cover common cases.

songgao:
@buoyad ready for another look; thanks! I also fixed styles for long text on mobile.

songgao:
Talked to @cecileboucheron on chat; will move the mobile CheckBox over to new style, and keep the checkbox on desktop. So I'm gonna untag review for now.

shared/common-adapters/index.stories.js
```diff
@@ -71,6 +72,7 @@ const load = () => {
     saveIndicator,
     sectionList,
     standardScreen,
+    switchStories,
```
buoyad: just `switch` for consistency

songgao: unfortunately `switch` is a reserved keyword 😅

PR #16859, opened 2019-04-05
add sync toggle for Files

songgao:


jzila:
@songgao shouldn't this be guarded by a flag?

songgao:
Yeah good call; will do before merging!

songgao:
this needs more tweaks, so I'm untagging review for now.

songgao:
actually the tweaks will be on kb.switch, so this is still ready for review.

jakob223:
This works for me on desktop, but I'm getting the following confusing behavior on iOS. Not sure if this is due to a problem with the branch or a problem with my setup.

jakob223:
Oops, image upload issue. Here it is.
image

songgao:
Hmm just tried over here again and it worked for me. Presumably this is your test account? Is it in any teams?

jakob223:
🤦‍♂️ thanks song.. my test account turns out not to be in any teams. Sorry for the false ⏰ ..

PR #16861, opened 2019-04-05
Fix errors when you're importing a private key while a public key is already published

pzduniak:


go/engine/pgp_import_key_test.go
```diff
@@ -91,6 +91,107 @@ func TestPGPImportAndExport(t *testing.T) {
...
+			Secret: false,
+			Query:  fp.String(),
+		},
+	})
+	if err := RunEngine2(m, xe); err != nil {
+		t.Fatal(err)
+	}
+	if len(xe.Results()) != 1 {
+		t.Fatalf("Expected 1 key back out")
+	}
```
zapu: if you want to minimize the code a bit here, using `require` pkg is fine anywhere now, even if given file is not using it already. but fine to leave it as is, it's already written after all

go/engine/pgp_import_key_test.go
```diff
@@ -91,6 +91,107 @@ func TestPGPImportAndExport(t *testing.T) {
...
+
+	// Run import on dev2
+	eng, err = NewPGPKeyImportEngineFromBytes(dev2.G, []byte(key), false)
```
zapu: so this is where it was failing before?

pzduniak: Correct

go/engine/pgp_import_key_test.go
pzduniak: it was just a massive copypaste, I thought about at least making this thing a func but it works as-is and I don't think we'll ever touch this again

go/engine/pgp_import_key_test.go
pzduniak: So basically without the changes it `key/add`'d and failed on the server as the key already exists. The SaveLKS stuff was callled properly so it worked, this change is mostly so that the users don't get confused by what's happening and have a way to do this flow.

pzduniak:
@zapu rdy for final review, tests are ok

maxtaco:
@pzduniak  -- just tried this on the CLI and it doesn't work

zapu:
Created Y2K-129. I think it's different issue though - what we have here is a user who have local secret key but wants to have server synced secret key.

pzduniak:
Yeah, this just stripped errors in this particular already working flow where you only had a public key but wanted to push its secret part.

PR #16862, opened 2019-04-05
Add Env.ForceSecretStoreFile, allow even on prod and on all platforms

zapu:


go/libkb/secret_store_darwin.go
```diff
@@ -222,9 +222,8 @@ func (k KeychainSecretStore) clearSecret(mctx MetaContext, account keychainSlott
 }
 
 func NewSecretStoreAll(mctx MetaContext) SecretStoreAll {
-	if mctx.G().Env.DarwinForceSecretStoreFile() {
-		// Allow use of file secret store for development/testing
-		// on MacOS.
+	if mctx.G().Env.ForceSecretStoreFile() {
+		// Allow use of file secret store for development/testing on MacOS.
```
heronhaye: Also mention linux if people don't want to pollute their keyring.

heronhaye: I'm going to implement the same thing in my PR for forcing file based with the same variable btw

go/libkb/env.go
```diff
@@ -1763,9 +1763,12 @@ func (e *Env) ModelessWantsSystemd() bool {
...
 }
 
-func (e *Env) DarwinForceSecretStoreFile() bool {
-	return (e.GetRunMode() == DevelRunMode &&
-		os.Getenv("KEYBASE_SECRET_STORE_FILE") == "1")
+func (e *Env) ForceSecretStoreFile() bool {
+	// By default use system-provided secret store (like MacOS Keychain), but
+	// allow users to fall back to file-based store for testing and debugging.
+	return e.GetBool(false,
+		func() (bool, bool) { return e.getEnvBool("KEYBASE_SECRET_STORE_FILE") },
```
heronhaye: Can we put this in config as well? So if a user has an issue with their keychain we can tell them to just flip the config flag on and not worry about env vars

zapu: fair point. I was going to not do a config variable not to encourage people to use files if they can avoid it, but it would be good to have one

go/libkb/secret_store_darwin.go
zapu: but that would be in different file, like `secret_store_linux.go`, right? darwin is exclusively macos/ios

heronhaye: Yeah

heronhaye: oh I didn't see filename, then dw about this

go/libkb/config.go
```diff
@@ -883,6 +883,10 @@ func (f *JSONConfigFile) GetExtraNetLogging() (bool, bool) {
 	return f.GetBoolAtPath("extra_net_logging")
 }
 
+func (f *JSONConfigFile) GetForceSecretStoreFile() (bool, bool) {
+	return f.GetBoolAtPath("force_less_safe_secret_store_file")
```
zapu: @heronhaye added config var

heronhaye:
@zapu anything blocking this from being merged?

PR #16870, opened 2019-04-05
Singleflight LoadUser, logging

mlsteele:
Singleflighting LoadUser reduces the number of UserChanged notifications that get sent to the frontend. Previously, two concurrent LoadUser's each thought they had discovered a new id version. LoadUser is allowed 30 seconds waiting on the lock before it bails.


mlsteele:
⛄️

PR #16875, opened 2019-04-05
implement subTitle for floating-menu

songgao:


songgao:
Thanks John; I'll wait until a release branch.

PR #16879, opened 2019-04-05
Show a spinner in place of profile buttons, so that we don't say "Accept" while we're loading

cjb:
@keybase/react-hackers


mmaxim:
Are the Android changes necessary?

cjb:
@mmaxim Nope, removed, thanks.

PR #16885, opened 2019-04-08
Fs: some optimizations to speedup or reduce component updates

songgao:
1. In `Rows`'s container, memoize `sortRowItems` result, so if we get state updates that don't involve pathItem changes, we can skip sorting. 
1.1 Reduce the debounce delay for folder filter to 100ms since computing the results have lower overhead now.
2. Make the Fs reducer smarter by doing some value comparison when we get updates about pathItem changes, and try to reuse the old pathItem reference when the content doesn't change. This way, components relying on `state.fs.get(blahblah)` in `mapStateToProps` only triggers updates when its content actually changes.
2.2 Add some tests for it.

### DIFF
--- a/shared/fs/row/rows-container.js
+++ b/shared/fs/row/rows-container.js
@@ -5,163 +5,164 @@ import * as Types from '../../constants/types/fs'
 import * as RowTypes from './types'
 import * as Constants from '../../constants/fs'
 import {isMobile} from '../../constants/platform'
-import {
-  sortRowItems,
-  type SortableStillRowItem,
-  type SortableEditingRowItem,
-  type SortableUploadingRowItem,
-  type SortableRowItem,
-} from './sort'
+import {sortRowItems, type SortableRowItem} from './sort'
 import Rows from './rows'
 import {asRows as topBarAsRow} from '../top-bar'
+import {memoize} from '../../util/memoize'
 
 type OwnProps = {|
   path: Types.Path, // path to the parent folder containering the rows
   routePath: I.List<string>,
   destinationPickerIndex?: number,
-  headerRows?: ?Array<RowTypes.RowItemWithKey>,
+  headerRows?: ?Array<RowTypes.HeaderRowItem>,
 |}
 
-const getEditingRows = (
-  edits: I.Map<Types.EditID, Types.Edit>,
-  parentPath: Types.Path
-): Array<SortableEditingRowItem> =>
-  edits
-    .filter(edit => edit.parentPath === parentPath)
-    .toArray()
-    .map(([editID, edit]) => ({
-      editID,
-      editType: edit.type,
-      key: `edit:${Types.editIDToString(editID)}`,
-      name: edit.name,
-      // fields for sortable
-      rowType: 'editing',
-      type: 'folder',
-    }))
+const getEditingRows = memoize(
+  (edits: I.Map<Types.EditID, Types.Edit>, parentPath: Types.Path): I.List<RowTypes.EditingRowItem> =>
+    I.List(
+      edits
+        .filter(edit => edit.parentPath === parentPath)
+        .toArray()
+        .map(([editID, edit]) => ({
+          editID,
+          editType: edit.type,
+          key: `edit:${Types.editIDToString(editID)}`,
+          name: edit.name,
+          // fields for sortable
+          rowType: 'editing',
+          type: 'folder',
+        }))
+    )
+)
 
-const getStillRows = (
-  pathItems: I.Map<Types.Path, Types.PathItem>,
-  parentPath: Types.Path,
-  names: Array<string>
-): Array<SortableStillRowItem> =>
-  names.reduce((items, name) => {
-    const item = pathItems.get(Types.pathConcat(parentPath, name), Constants.unknownPathItem)
-    const path = Types.pathConcat(parentPath, item.name)
-    return [
-      ...items,
-      {
-        key: `still:${name}`,
-        lastModifiedTimestamp: item.lastModifiedTimestamp,
-        name: item.name,
-        path,
-        // fields for sortable
-        rowType: 'still',
-        type: item.type,
-      },
-    ]
-  }, [])
+const getStillRows = memoize(
+  (
+    pathItems: I.Map<Types.Path, Types.PathItem>,
+    parentPath: Types.Path,
+    names: I.Set<string>
+  ): I.List<RowTypes.StillRowItem> =>
+    I.List(
+      names.toArray().reduce((items, name) => {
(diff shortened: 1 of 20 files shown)


songgao:
bump!

shared/fs/row/rows-container.js
```diff
@@ -5,163 +5,164 @@ import * as Types from '../../constants/types/fs'
...
+  ])
+
+const getTlfRowsFromTlfs = memoize(
+  (tlfs: I.Map<string, Types.Tlf>, tlfType: Types.TlfType): I.List<SortableRowItem> =>
+    I.List().withMutations(list =>
+      tlfs.reduce(
+        (rows, {isIgnored, isNew}, name) =>
+          isIgnored
+            ? rows
+            : rows.push({
```
jzila: Doesn't this return the length? I think you want `rows.concat`, or the spread operator like it was before.

songgao: Ha; took me a while to understand how this code works ... this is `I.List` now, not and `Array`. So `.push` actually returns a new list: https://immutable-js.github.io/immutable-js/docs/#/List/push

songgao:
Although now I'm realizing this should check equality of the `prefetchStatus` field too. So please hold off reviewing. Will tag review again when it's ready!

songgao:
@jzila ready again; please take a look at the last commit. Thanks!

shared/reducers/__tests__/fs.js
```diff
@@ -148,13 +150,64 @@ describe('fs reducer', () => {
...
             }),
           ],
         ]),
       })
     )
-    expect(stage1.pathItems).not.toBe(state0.pathItems)
-    expect(getFolderOrFail(stage1.pathItems, folder0Path).children).toEqual(I.Set(['file1']))
+    expect(state1.pathItems).not.toBe(state0.pathItems)
+    expect(getFolderOrFail(state1.pathItems, folder0Path).children).toEqual(I.Set(['file1']))
+    expect(getFolderOrFail(state1.pathItems, folder0Path).prefetchStatus).toBe(Constants.prefetchNotStarted)
```
jzila: you can compare `I.Record` like this?

shared/reducers/fs.js
```diff
@@ -9,10 +9,18 @@ import * as Types from '../constants/types/fs'
...
+  newPathItemFromAction: Types.PathItem
+): Types.PathItem => {
+  if (!oldPathItem || oldPathItem.type !== newPathItemFromAction.type) {
+    return newPathItemFromAction
   }
+  // Reuse prefetchStatus if they equal in value.
+  // $FlowIssue
+  const newPathItem = newPathItemFromAction.update('prefetchStatus', newPrefetchStatus =>
+    newPrefetchStatus.equals(oldPathItem.prefetchStatus) ? oldPathItem.prefetchStatus : newPrefetchStatus
+  )
```
jzila: What does this get us? If they're equal, then why not let it be set? You're calling `update` anyway so you get a new object no matter what...

shared/reducers/fs.js
```diff
@@ -102,6 +110,47 @@ const updatePathItem = (oldPathItem?: ?Types.PathItem, newPathItem: Types.PathIt
...
+        .set('resetParticipants', oldTlf.resetParticipants)
+        .set('waitingForParticipantUnlock', oldTlf.waitingForParticipantUnlock)
+        .set('youCanUnlock', oldTlf.youCanUnlock)
+        .set('syncConfig', oldTlf.syncConfig)
+    )
+  )
+}
+
+const updateTlfList = (oldTlfList: Types.TlfList, newTlfList: Types.TlfList): Types.TlfList =>
+  newTlfList.withMutations(list => list.map((tlf, name) => updateTlf(oldTlfList.get(name), tlf)))
```
jzila: I don't think `withMutations` helps us here, since there's only one operation underneath it.

shared/reducers/fs.js
```diff
@@ -102,6 +110,47 @@ const updatePathItem = (oldPathItem?: ?Types.PathItem, newPathItem: Types.PathIt
...
+    return newTlfDontClearSyncConfig
+  }
+  if (
+    !I.is(newTlfDontClearSyncConfig.syncConfig, oldTlf.syncConfig) &&
+    !(
+      newTlfDontClearSyncConfig.syncConfig &&
+      oldTlf.syncConfig &&
+      newTlfDontClearSyncConfig.syncConfig.mode === 'partial' &&
+      oldTlf.syncConfig.mode === 'partial' &&
+      newTlfDontClearSyncConfig.syncConfig.enabledPaths.equals(oldTlf.syncConfig.enabledPaths)
```
jzila: This is extremely difficult to decipher. I'd prefer DeMorgan'ing this:
```javascript
if (
  !I.is(...) && (
    !newTlfDontClearSyncConfig.syncConfig ||
    !oldTlf.syncConfig ||
    !newTlfDontClearSyncConfig.syncConfig.mode === 'partial' ||
    !oldTlf.syncConfig.mode === 'partial' ||
    !newTlfDontClearSyncConfig.syncConfig.enabledPaths.equals(oldTlf.syncConfig.enabledPaths)
  )
)
```

However, it might be better to make a helper function to do this check, that explains what it's meant to represent.

shared/reducers/fs.js
```diff
@@ -181,13 +230,13 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
-      return state.set(
-        'tlfs',
-        Constants.makeTlfs({
-          private: action.payload.private,
-          public: action.payload.public,
-          team: action.payload.team,
-        })
+      return state.update('tlfs', tlfs =>
+        tlfs.withMutations(tlfs =>
+          tlfs
```
jzila: Let's use a different name for the two `tlfs`, to make it unambiguous.

shared/reducers/__tests__/fs.js
songgao: `toBe` is reference equality. So this is saying they should be exactly the same object.

shared/reducers/fs.js
songgao: This re-uses the old `prefetchStatus` record if they equal in value. Update doesn't actually make a new immutable record unless we are giving it a different thing.

```
> r.a
123
> r.set('a', 123) === r
true
```

jzila: OK but if you're using the same references everywhere for a given status, then it shouldn't matter, right?

songgao: For enabled/disabled, you are right it wouldn't matter. But for partially enabled ones, it's not the same reference everywhere. So I thought I might as well do it right here in the first place.

jzila: OK fair enough, thanks for explaining! Mind putting a comment?

songgao:
Thanks John!

PR #16889, opened 2019-04-08
Checking if current device is revoked in CanLogout

zapu:
If we are unable to check HasRandomPW while logging out, check if we are revoked.

Checking HasRandomPW status code would be sufficient (API session error) but this would be server trust, and also prone to other misc errors. And we are trying really hard not to log user out if they don't have passphrase.

### DIFF
--- a/go/service/user.go
+++ b/go/service/user.go
@@ -4,7 +4,6 @@
 package service
 
 import (
-	"errors"
 	"fmt"
 	"sort"
 	"time"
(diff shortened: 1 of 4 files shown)


go/service/user.go
```diff
@@ -638,6 +639,29 @@ func (h *UserHandler) LoadHasRandomPw(ctx context.Context, arg keybase1.LoadHasR
 	return ret.RandomPW, err
 }
 
+func isActiveDeviceRevoked(mctx libkb.MetaContext) (res bool, err error) {
```
zapu: @mlsteele is there an easier way?

zapu:
It's a simple but rather important change - comments appreciated!

We are always erring on the side of caution here, we don't want to log someone out if we are not sure if they can log back in. There is always a workaround by doing `keybase logout --force` which skips all checks, but during normal `keybase logout` it's better to be safe than sorry.

go/service/user.go
maxtaco: `checkDeviceValidForUID`

go/service/user.go
```diff
@@ -651,6 +675,15 @@ func (h *UserHandler) CanLogout(ctx context.Context, sessionID int) (res keybase
 	})
 
 	if err != nil {
+		isRevoked, err2 := isActiveDeviceRevoked(libkb.NewMetaContext(ctx, h.G()))
```
maxtaco: streamline error handling here and make it more go-y.  don't have nested-if error conditions

zapu: I wanted it to be nested here so this is only checked if call to hasRandomPW fails, because hasRandomPW can be cached and works offline. So if `CheckCurrentUIDDeviceID` is only called when hasRandomPW fails, the whole RPC still have the ability to work offline (assuming `hasRandomPW` can find a cached value).

Unless there is some other control flow simplification that I'm missing

zapu:
OK this RPC turned out to have more issues than I expected, so we are not trying to rush it for the release. Stranded users can still use `keybase logout --force`, although it's rather hard to get in this state, because when you become revoked or reset, you get a gregor message that logs you out.

go/service/user.go
```diff
@@ -651,6 +651,18 @@ func (h *UserHandler) CanLogout(ctx context.Context, sessionID int) (res keybase
 	})
 
 	if err != nil {
+		// Couldn't check Random PW. Maybe we are logged in as invalid user or
+		// revoked device. If so, green-light logout.
+		if checkUIDErr := libkb.CheckCurrentUIDDeviceID(libkb.NewMetaContext(ctx, h.G())); checkUIDErr != nil {
+			switch checkUIDErr.(type) {
+			case libkb.DeviceNotFoundError, libkb.UserNotFoundError, libkb.KeyRevokedError, libkb.NoDeviceError, libkb.NoUIDError:
```
joshblum: nit: dropping a newline between these errors makes it easier to read imo

zapu:
@joshblum @maxtaco made the CanLogout check more straightforward but potentially more expensive - `libkb.CheckCurrentUIDDeviceID` is called every time, not only when HasRandomPW errors out. This is much simplier and less error prone with regards to HasRandomPW cache.

Also added a test for the revoked device + canLogout case.

PR #16890, opened 2019-04-08
Show back button on provisioning code page in nav1

buoyad:
r? @keybase/react-hackers


buoyad:
This fixes it on desktop too, it was broken in nav1 in general.

PR #16893, opened 2019-04-08
Fix 'startProof finished' showing on profile after startProof finishes

buoyad:
and having to click 'Ok tweeted! Check for it!' and analogs twice when making non-generic proofs. r? @keybase/react-hackers cc @mlsteele

### DIFF
--- a/shared/actions/profile/proofs.js
+++ b/shared/actions/profile/proofs.js
@@ -253,7 +253,7 @@ function* addProof(state, action) {
     assertion: state.config.username,
     guiID: Tracker2Constants.generateGUIID(),
     inTracker: false,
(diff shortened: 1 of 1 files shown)


shared/actions/profile/proofs.js
```diff
@@ -281,6 +281,9 @@ function* addProof(state, action) {
     })
     yield Saga.put(ProfileGen.createUpdateSigID({sigID}))
     logger.info('Start Proof done: ', sigID)
+    if (!genericService) {
+      yield Saga.put(ProfileGen.createCheckProof())
```
buoyad: This fixes having to click the button twice

PR #16895, opened 2019-04-08
don't allow save for audio files on mobile

songgao:


shared/constants/fs.js
```diff
@@ -526,8 +526,18 @@ export const viewTypeFromMimeType = (mime: ?Types.Mime): Types.FileViewType => {
   return 'default'
 }
 
-export const isMedia = (pathItem: Types.PathItem): boolean =>
-  pathItem.type === 'file' && ['image', 'av'].includes(viewTypeFromMimeType(pathItem.mimeType))
+export const canSaveMedia = (pathItem: Types.PathItem): boolean => {
+  if (pathItem.type !== 'file' || !pathItem.mimeType) {
+    return false
+  }
+  const mime: Types.Mime = pathItem.mimeType
```
jzila: I don't think you need the type here, it should be implicit from the `mimeType`.

songgao: Hrm initially I tried `pathItem.type === 'file' && !!pathItem.mimeType && ( ... || ...)` and flow freaked out. Looks like this this separate assignment it indeed works already. 🤷🏻‍♂️

PR #16896, opened 2019-04-08
Squash tls conn closed/protocol shutdown errors

joshblum:
Got a blackbar from this earlier during `MarkLocalAsRead`, is this the correct place to squash?


mmaxim:
This is very strange, I'm not sure how that happens.

PR #16897, opened 2019-04-08
block_tree: parallelize block-readying

strib:
Use the pipeliner for controlled parallelization when readying all the blocks at the same level of a file.

This speeds up large file copies (about 19% for a 4GB file), though doesn't have much effect on benchmarks of lots of small-file copies.

(I have some ideas about doing multiple `blockTree` readies all at once to help with the small-file benchmarks, but it's a lot more complicated and this seemed like an easy, but unrelated, win for big files.  Detailed tracing shows that maybe 33% of overall time is spent readying blocks.)

### DIFF
--- a/go/kbfs/data/block_tree.go
+++ b/go/kbfs/data/block_tree.go
@@ -16,7 +16,7 @@ import (
 	"github.com/keybase/client/go/kbfs/tlf"
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/client/go/protocol/keybase1"
-	"github.com/keybase/pipeliner"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/block_tree.go
```diff
@@ -1054,16 +1071,30 @@ func (bt *blockTree) readyHelper(
...
+			return nil
+		}
+
+		pipeliner := pipeliner.NewPipeliner(maxParallelReadies)
+		for i := 0; i < len(pathsFromRoot); i++ {
+			err := pipeliner.WaitForRoom(ctx)
+			if err != nil {
+				return nil, err
+			}
+			go func(i int) { pipeliner.CompleteOne(worker(i)) }(i)
```
jzila: I'd prefer an approach with a fixed number of goroutines iterating through the work, rather than launching one for each item. Just seems like unnecessary overhead.

strib: I was just trying out max's new "clean" way of doing things.  I think goroutine overhead is basically zero, but I don't mind doing it with explicit channels, so I pushed the suggested change (and rebased onto master).  Please take another look!

jzila: Yeah I read Max's post, but it goes against the canonical suggestion in the [pipelines](https://blog.golang.org/pipelines) post, which is to fan-in to and fan-out from a pool of goroutines. The difference from threads is that you can have hundreds or thousands instead of tens.

At the very least, goroutines add ~4kB of memory overhead each.

strib: Sure, but the pipeliner way has the same number of concurrent goroutines as the new one, so memory overhead is the same.  (Moving the stackpointer back and forth for each new goroutine doesn't seem like it would matter.)

jzila: > Moving the stackpointer back and forth for each new goroutine doesn't seem like it would matter.

Good point, I hadn't thought of it that way.

PR #16900, opened 2019-04-08
Reintroduce android share from cold start

jzila:


jzila:
OK tested on iPhone, works great.

PR #16904, opened 2019-04-08
Proof suggestions newness bool

mlsteele:
might be used to put a "New" badge on "Add other identities"


go/service/user.go
```diff
@@ -382,6 +382,9 @@ func (h *UserHandler) ProofSuggestions(ctx context.Context, sessionID int) (ret
 		if foldPriority > 0 && suggestion.Priority >= foldPriority {
 			ret.ShowMore = true
 			suggestion.BelowFold = true
+			if !ret.MoreIsNew && len(suggestion.Metas) > 0 {
```
joshblum: could be?
```suggestion
			if len(suggestion.Metas) > 0 {
```

mlsteele:
the button is hardcoded "new" for now

PR #16911, opened 2019-04-08
Promote new rows

mlsteele:


shared/profile/generic/proofs-list/container.js
```diff
@@ -20,18 +20,31 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => ({
   onCancel: dispatchProps.onCancel,
   onClickLearn: () => openURL('https://keybase.io/docs/proof_integration_guide'),
   providerClicked: dispatchProps.providerClicked,
-  providers: stateProps._proofSuggestions
+  providers: promoteNew(stateProps._proofSuggestions
```
cjb: Could consider doing something like:
```js
import {flatten, partition} from 'lodash-es'

providers: flatten(partition(stateProps._proofSuggestions, 'new'))
```

mlsteele: partition returns a tuple and flatten takes an array. If only flatten took a $ReadOnlyArray

shared/profile/generic/proofs-list/container.js
```diff
@@ -20,18 +21,22 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => ({
...
       name: s.pickerText,
       new: s.metas.some(({label}) => label === 'new'),
     }))
-    .toArray(),
+    .toArray()),
   title: 'Prove your...',
 })
 
+const promoteNew = (rows) => {
+  return flatten(partition(rows, (row) => row.new).map(identity))
```
cjb: Can replace row => row.new with just 'new'.

PR #16913, opened 2019-04-08
libkb: fix unstubbing of v2 userchains

maxtaco:
- the setup:
  - the most obvious problem we had here is that if you had followed someone with a V1 link, and then updated it with a V2 link, but some other user on your same cache managed to load a UPAK, you might not see the second link. It turned out that the UPAK would show the stubbed links, even if loaded for self.
  - there is some server-side unstubbing for self, but that gets baffled if the cache is poisoned by the other user
- the solution:
  - fix the SigChainLoader.fullChain machinery that @pzduniak put in. It didn't really work before because it didn't go back and reload stubbed links; so now, if you ask for an unstubbed load, and you have stubbed links, those links will be clobbered (after they are checked for the right hashses)
  - fix UPAK disk cache to not clobber unstubbed links with stubbed links
  - fix UPAK memory cache to also not clobber
  - plumb the new options up to ID2, which need these options when pulling the "me" user to check for track links
- add a test


go/libkb/sig_chain.go
```diff
@@ -1358,10 +1428,6 @@ func (l *SigChainLoader) Load() (ret *SigChain, err error) {
 	if current, err = l.CheckFreshness(); err != nil {
 		return nil, err
 	}
-	stage("CheckStubsIfFull")
-	if current && l.full {
-		current = l.HasStubs()
-	}
```
maxtaco: this didn't work when i fixed the underlying UPAK machinery

maxtaco:
cc @pzduniak on the PR

go/libkb/loaduser.go
```diff
@@ -49,9 +71,9 @@ type LoadUserArg struct {
 }
 
 func (arg LoadUserArg) String() string {
-	return fmt.Sprintf("{UID:%s Name:%q PublicKeyOptional:%v NoCacheResult:%v Self:%v ForceReload:%v ForcePoll:%v StaleOK:%v AbortIfSigchainUnchanged:%v CachedOnly:%v}",
+	return fmt.Sprintf("{UID:%s Name:%q PublicKeyOptional:%v NoCacheResult:%v Self:%v ForceReload:%v ForcePoll:%v StaleOK:%v AbortIfSigchainUnchanged:%v CachedOnly:%v %v}",
```
joshblum:
```suggestion
	return fmt.Sprintf("{UID:%s Name:%q PublicKeyOptional:%v NoCacheResult:%v Self:%v ForceReload:%v ForcePoll:%v StaleOK:%v AbortIfSigchainUnchanged:%v CachedOnly:%v StubMode:%v}",
```

go/libkb/sig_chain.go
```diff
@@ -514,6 +516,15 @@ func (sc *SigChain) Store(m MetaContext) (err error) {
 	return nil
 }
 
+func (sc *SigChain) checkUnstubs(low int, unstubs map[keybase1.Seqno]LinkID) error {
+	for _, link := range sc.chainLinks[low:] {
```
joshblum: sanity check all of the unstubs are found?

go/libkb/sig_chain.go
```diff
@@ -1113,7 +1127,51 @@ func (l *SigChainLoader) LoadLinksFromStorage() (err error) {
...
 	l.links = links
+
+	// Now that we've set the links array above, we're going to potentially discard the stubbed links,
+	// and then set a mapping of (seqno -> linkID) so that the server doesn't lie on an unstubbing.
+	l.maybeDiscardStubbedLinks()
+
+	return
+}
+
+// maybeDiscardStubbedLinks will look at the sigchain links loaded fro the DB and will maybe throw
```
joshblum:
```suggestion
// maybeDiscardStubbedLinks will look at the sigchain links loaded from the DB and will maybe throw
```

go/libkb/loaduser.go
maxtaco: Nope, it has a magic `String()` function!

go/libkb/sig_chain.go
maxtaco: I was thinking about that...sure....

go/libkb/upak_loader.go
```diff
@@ -99,14 +125,42 @@ func (u *CachedUPAKLoader) ClearMemory() {
...
+	} else if !found {
+		u.G().VDL.CLogf(ctx, VLog0, "| missed disk cache")
+	} else if tmp.MinorVersion == UPK2MinorVersionCurrent {
+		u.G().VDL.CLogf(ctx, VLog0, "| hit disk cache (v%d)", tmp.MinorVersion)
+		ret = &tmp
+	} else {
+		u.G().VDL.CLogf(ctx, VLog0, "| found old minor version %d, but wanted %d; will overwrite with fresh UPAK", tmp.MinorVersion, UPK2MinorVersionCurrent)
+	}
+	return ret
+}
```
mlsteele: v5 is a year outmoded. This could have a bug or future changes to v6 could leave it with a bug. How about dropping v5->v6 support?

mlsteele:
Why have two slots for each uid's upak? That could lead to a split cache alice+nostub and alice+stub are at different seqnos. What about single storage. If you load with nostub but the storage has stubbed links then reload. If you load with stub then use the storage.

go/libkb/sig_chain.go
```diff
@@ -1255,8 +1321,21 @@ func (l *SigChainLoader) selfUID() (uid keybase1.UID) {
...
+	l.dirtyTail, err = l.chain.LoadFromServer(l.M(), srv, l.selfUID(), l.stubMode)
+
+	if err != nil {
+		return err
+	}
+
+	err = l.chain.checkUnstubs(low, l.unstubs)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
maxtaco: :+1:

protocol/avdl/keybase1/upk.avdl
```diff
@@ -86,6 +86,9 @@ protocol UPK {
       union { null, string } stellarAccountID;
       map<UID, RemoteTrack> remoteTracks;
       union { null, ResetSummary } reset;
+      // Introduced in 2.6 of UPAKs, so not all 2.6 stored UPAKs have this. We'll therefore
+      // assume that stored upaks are stubbed by default.
+      boolean unstubbed;
```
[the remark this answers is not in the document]
maxtaco: :+1:

go/libkb/upak_loader.go
maxtaco: great idea. I'll just delete those lines

maxtaco:
CI is flake

maxtaco:
@mlsteele  ok to merge this?

PR #16914, opened 2019-04-08
Fix invalid username permaspinner in generic proofs

buoyad:
Fixes a bug where this button would permaspin if you entered an invalid username
Keybase_DEV

r? @keybase/react-hackers


buoyad:
cc @mlsteele

PR #16915, opened 2019-04-08
New label right side

mlsteele:
image
image

Not sure why the right padding is bigger. Don't want to break Button

cc @cecileboucheron

### DIFF
--- a/shared/profile/user/index.js
+++ b/shared/profile/user/index.js
@@ -202,6 +202,7 @@ export class BioTeamProofs extends React.PureComponent<BioTeamProofsProps> {
           onClick={this.props.onAddIdentity}
           style={styles.addIdentityButton}
           type="Secondary"
+          childrenRight={true}
         >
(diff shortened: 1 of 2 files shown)


shared/profile/user/index.js
```diff
@@ -202,6 +202,7 @@ export class BioTeamProofs extends React.PureComponent<BioTeamProofsProps> {
           onClick={this.props.onAddIdentity}
           style={styles.addIdentityButton}
           type="Secondary"
+          childrenRight={true}
         >
           <Kb.Meta backgroundColor={Styles.globalColors.blue} title="NEW" style={styles.newMeta} />
```
buoyad: Include the label as text in `children` and remove the `label` prop. Extra padding on the right is from `styles.newMeta`

mlsteele:
waiting on https://github.com/keybase/client/pull/16920

PR #16916, opened 2019-04-08
Confirm dialog on folder deletion

jakob223:
Still got a few kinks to work out, in that I'm pretty sure I'm not doing the `waitingKey` stuff right. But it's working and ready for review.

Issue: KBFS-3987

### DIFF
--- a/shared/constants/fs.js
+++ b/shared/constants/fs.js
@@ -20,6 +20,9 @@ import {type TypedActions} from '../actions/typed-actions-gen'
 import flags from '../util/feature-flags'
 
 export const syncToggleWaitingKey = 'fs:syncToggle'
+export const sendLinkToChatFindConversationWaitingKey = 'fs:sendLinkToChatFindConversation'
+export const sendLinkToChatSendWaitingKey = 'fs:sendLinkToChatSend'
+export const deleteWaitingKey = 'fs:delete'
 
 export const defaultPath = Types.stringToPath('/keybase')
 
--- a/shared/fs/common/path-item-action/choose-view.js
+++ b/shared/fs/common/path-item-action/choose-view.js
@@ -11,6 +11,7 @@ import Confirm from './confirm-container'
 type OwnProps = {|
   floatingMenuProps: FloatingMenuProps,
   path: Types.Path,
+  position?: 'row' | 'header',
   routePath: I.List<string>,
 |}
 
@@ -20,7 +21,14 @@ const mapStateToProps = state => ({
 
 const ChooseView = props => {
   if (props.view === 'root' || props.view === 'share') {
-    return <Menu routePath={props.routePath} path={props.path} floatingMenuProps={props.floatingMenuProps} />
+    return (
+      <Menu
+        routePath={props.routePath}
+        path={props.path}
+        position={props.position}
+        floatingMenuProps={props.floatingMenuProps}
+      />
+    )
   } else if (props.view === 'confirm-save-media' || props.view === 'confirm-send-to-other-app') {
     return <Confirm path={props.path} floatingMenuProps={props.floatingMenuProps} />
   } else {
--- a/shared/fs/common/path-item-action/menu-container.js
+++ b/shared/fs/common/path-item-action/menu-container.js
@@ -11,10 +11,13 @@ import flags from '../../../util/feature-flags'
 import Menu from './menu'
 import type {FloatingMenuProps} from './types'
 import {getRootLayout, getShareLayout} from './layout'
+import * as RouteTreeGen from '../../../actions/route-tree-gen'
+import {fsTab} from '../../../constants/tabs'
 
 type OwnProps = {|
   floatingMenuProps: FloatingMenuProps,
   path: Types.Path,
+  position?: 'row' | 'header',
   routePath: I.List<string>,
 |}
 
@@ -27,7 +30,7 @@ const mapStateToProps = (state, {path}) => ({
   _view: state.fs.pathItemActionMenu.view,
 })
 
-const mapDispatchToProps = (dispatch, {path, routePath}: OwnProps) => ({
+const mapDispatchToProps = (dispatch, {path, position, routePath}: OwnProps) => ({
   _cancel: (key: string) => dispatch(FsGen.createCancelDownload({key})),
   _confirmSaveMedia: (toCancel: ?string) =>
     dispatch(FsGen.createSetPathItemActionMenuView({view: 'confirm-save-media'})),
@@ -35,8 +38,12 @@ const mapDispatchToProps = (dispatch, {path, routePath}: OwnProps) => ({
     dispatch(FsGen.createSetPathItemActionMenuView({view: 'confirm-send-to-other-app'})),
   _copyPath: () => dispatch(ConfigGen.createCopyToClipboard({text: Constants.escapePath(path)})),
   _delete: () => {
-    dispatch(FsGen.createDeleteFile({path}))
-    dispatch(Constants.makeActionForOpenPathInFilesTab(Types.getPathParent(path), routePath))
+    dispatch(
+      RouteTreeGen.createNavigateTo({
+        parentPath: [fsTab],
+        path: [{props: {path, position}, selected: 'reallyDelete'}],
+      })
+    )
   },
   _download: () => dispatch(FsGen.createDownload({key: Constants.makeDownloadKey(path), path})),
   _ignoreTlf: () => dispatch(FsGen.createFavoriteIgnore({path})),
@@ -142,6 +149,7 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
     ignoreTlf: layout.ignoreTlf ? c(dispatchProps._ignoreTlf) : null,
     moveOrCopy: flags.moveOrCopy && layout.moveOrCopy ? c(dispatchProps._moveOrCopy) : null,
     pathItemType: stateProps._pathItem.type,
+    position: ownProps.position,
     saveMedia: layout.saveMedia ? getSaveMedia(stateProps, dispatchProps, c) : null,
     showInSystemFileManager:
       layout.showInSystemFileManager && stateProps._sfmiEnabled
--- a/shared/fs/really-delete/container.js
+++ b/shared/fs/really-delete/container.js
@@ -1,0 +1,59 @@
+// @flow
+import * as FsGen from '../../actions/fs-gen'
+import * as RouteTreeGen from '../../actions/route-tree-gen'
+import * as Container from '../../util/container'
+import ReallyDelete from '.'
+import {anyWaiting} from '../../constants/waiting'
+import * as Types from '../../constants/types/fs'
+import * as Constants from '../../constants/fs'
+
+type OwnProps = Container.RouteProps<{path: Types.Path, position?: 'row' | 'header'}, {}>
+
+const mapStateToProps = state => ({
+  _deleting: anyWaiting(state, Constants.deleteWaitingKey),
+  title: 'Confirmation',
+})
+
+const mapDispatchToProps = (dispatch, ownProps) => {
+  const path = Container.getRouteProps(ownProps, 'path')
+  const position = Container.getRouteProps(ownProps, 'position')
+  return {
+    _onFinishDelete: () => {
+      // If this is a header menu, then we're deleting the folder we're in,
+      // and we need to navigate up twice.
+      if (position === 'header') {
+        dispatch(Constants.makeActionForOpenPathInFilesTab(Types.getPathParent(path)))
+        dispatch(RouteTreeGen.createNavigateUp())
+      } else {
+        dispatch(RouteTreeGen.createNavigateUp())
+      }
+    },
+    onBack: () => dispatch(RouteTreeGen.createNavigateUp()),
+    onDelete: () => {
+      if (path !== Constants.defaultPath) {
+        dispatch(FsGen.createDeleteFile({path}))
+      }
+    },
+  }
+}
+
+const mergeProps = (stateProps, dispatchProps, ownProps) => {
+  const path = Container.getRouteProps(ownProps, 'path')
+  return {
+    _deleting: stateProps._deleting,
+    _onFinishDelete: dispatchProps._onFinishDelete,
+    onBack: stateProps._deleting ? () => {} : dispatchProps.onBack,
+    onDelete: dispatchProps.onDelete,
+    path: path,
+    title: stateProps.title,
+  }
+}
+
+export default Container.compose(
+  Container.connect<OwnProps, _, _, _, _>(
+    mapStateToProps,
+    mapDispatchToProps,
+    mergeProps
+  ),
+  Container.safeSubmit(['onDelete'], ['_deleting'])
+)(ReallyDelete)
--- a/shared/fs/really-delete/index.js
+++ b/shared/fs/really-delete/index.js
@@ -1,0 +1,39 @@
+// @flow
+import * as React from 'react'
+import * as Constants from '../../constants/fs'
+import * as Kb from '../../common-adapters'
+import * as Styles from '../../styles'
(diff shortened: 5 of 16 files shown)


jzila:
Can you fix CI and conflicts? I'll review once that is done.

shared/fs/really-delete/container.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  onBack: () => dispatch(RouteTreeGen.createNavigateUp()),
+  onDelete: () => {
+    const path = Container.getRouteProps(ownProps, 'path')
+    dispatch(FsGen.createDeleteFile({path}))
+    // TODO: I think I should be closing this using the waiting key instead,
+    //  but I'm not sure how that works.
+    dispatch(RouteTreeGen.createNavigateUp())
```
jzila: I don't think this is relevant for nav2. @songgao what's the right way to navigate out of a menu in nav2?

shared/fs/really-delete/container.js
```diff
@@ -0,0 +1,46 @@
...
+    _deleting: anyWaiting(state, Types.deleteFolderWaitingKey(path)),
+    path: path,
+    title: 'Confirmation',
+  }
+}
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  onBack: () => dispatch(RouteTreeGen.createNavigateUp()),
+  onDelete: () => {
+    const path = Container.getRouteProps(ownProps, 'path')
```
jzila: The standard way to do this is to make this function:
```javascript
_onDelete: (path: Types.Path) => dispatch(FsGen.createDeleteFile({path}))

...
const mergeProps = (stateProps, dispatchProps, ownProps) => ({
  onDelete: () => _onDelete(Container.getRouteProps(ownProps, 'path')),
  ...
})

File: shared/fs/really-delete/index.js [hunk 3]
```diff
@@ -0,0 +1,42 @@
...
+const Spinner = HeaderOnMobile(_Spinner)
+
+const Header = (props: Props) => <Icon type="iconfont-trash" sizeType="Big" />
+
+const _ReallyDeleteFile = (props: Props) => (
+  <ConfirmModal
+    confirmText={`Yes, delete it.`}
+    // TODO: this is technically false, in that time travel exists. Change
+    //  this text, probably?
+    description={`There's no trash can - "${props.name}" will be gone forever.`}
```
jzila: From a UI perspective it's defacto true, until we make history rollback possible in the UI.

shared/constants/fs.js
```diff
@@ -21,6 +21,7 @@ import flags from '../util/feature-flags'
 
 export const sendLinkToChatFindConversationWaitingKey = 'fs:sendLinkToChatFindConversation'
 export const sendLinkToChatSendWaitingKey = 'fs:sendLinkToChatSend'
+export const deleteFolderWaitingKey = (path: Types.Path) => `fs:deleteFolder:${Types.pathToString(path)}`
```
songgao: There's only one confirm modal shown at a time, so we can just use one single waitingKey here. Then it'll be a constant rather than a function.

shared/fs/common/path-item-action/menu-container.js
```diff
@@ -34,10 +36,19 @@ const mapDispatchToProps = (dispatch, {path, routePath}: OwnProps) => ({
...
   },
+  _deleteFolder: () => {
+    dispatch(
+      RouteTreeGen.createNavigateTo({
+        parentPath: [fsTab],
+        path: [{props: {path}, selected: 'reallyDelete'}],
+      })
+    )
+    dispatch(Constants.makeActionForOpenPathInFilesTab(Types.getPathParent(path), routePath))
+  },
```
songgao: IIRC, we want the confirmation for files as well, not just folders. So files/folders are actually the same.

shared/fs/really-delete/container.js
songgao: I think this should work; there's also a way to clear all modals if you have more than one layer, but this is just one layer so navigateUp works too.

shared/fs/really-delete/container.js
```diff
@@ -0,0 +1,47 @@
...
+    _deleting: (path: Types.Path) => anyWaiting(state, Constants.deleteFolderWaitingKey(path)),
+    title: 'Confirmation',
+  }
+}
+
+const mapDispatchToProps = dispatch => ({
+  _onDelete: (path: Types.Path) => {
+    dispatch(FsGen.createDeleteFile({path}))
+    // TODO: I think I should be closing this using the waiting key instead,
+    //  but I'm not sure how that works.
```
songgao: You could fire `navigateUp` in `componentDidUpdate` when waiting state clears. See `TriggerActionsWhenNeeded` in send-link-to-chat. It doesn't use the waiting key, but the logic should be similar.

shared/fs/really-delete/container.js
```diff
@@ -0,0 +1,47 @@
...
+import ReallyDelete from '.'
+import {anyWaiting} from '../../constants/waiting'
+import * as Types from '../../constants/types/fs'
+import * as Constants from '../../constants/fs'
+
+type OwnProps = Container.RouteProps<{path: Types.Path}, {}>
+
+const mapStateToProps = state => {
+  return {
+    _deleting: (path: Types.Path) => anyWaiting(state, Constants.deleteFolderWaitingKey(path)),
```
songgao: If you make waitingKey a string (i.e. not dependent on `path`), then this doesn't need to be a function anymore.

shared/fs/really-delete/index.js
```diff
@@ -0,0 +1,42 @@
...
+  </MaybePopup>
+)
+const Spinner = HeaderOnMobile(_Spinner)
+
+const Header = (props: Props) => <Icon type="iconfont-trash" sizeType="Big" />
+
+const _ReallyDeleteFile = (props: Props) =>
+  !!props.path && (
+    <ConfirmModal
+      confirmText={`Yes, delete it.`}
```
songgao: `confirmText="Yes, delete it."`, same below.

jakob223: This seems more readable to me than escaping the quotes and manually concatenating the path in.

jzila: I think @songgao is saying that you can leave everything the same, except not escape to JS to put the string in.
```
<ConfirmModal confirmText="Yes, delete it." />
```

jakob223: er, I changed this one. I'm referring to the below. e.g. ```prompt={`Are you sure you want to delete "${Types.pathToString(this.props.path)}"?`}```

shared/fs/really-delete/container.js
jakob223: I've changed it to do this.

shared/fs/common/path-item-action/menu-container.js
```diff
@@ -34,8 +36,13 @@ const mapDispatchToProps = (dispatch, {path, routePath}: OwnProps) => ({
   _confirmSendToOtherApp: () =>
     dispatch(FsGen.createSetPathItemActionMenuView({view: 'confirm-send-to-other-app'})),
   _copyPath: () => dispatch(ConfigGen.createCopyToClipboard({text: Constants.escapePath(path)})),
-  _delete: () => {
-    dispatch(FsGen.createDeleteFile({path}))
+  _deleteFile: () => {
```
songgao: Should this be `_delete` instead of `_deleteFile`?

shared/fs/common/path-item-action/menu-container.js
```diff
@@ -34,8 +36,13 @@ const mapDispatchToProps = (dispatch, {path, routePath}: OwnProps) => ({
...
+    dispatch(
+      RouteTreeGen.createNavigateTo({
+        parentPath: [fsTab],
+        path: [{props: {path}, selected: 'reallyDelete'}],
+      })
+    )
     dispatch(Constants.makeActionForOpenPathInFilesTab(Types.getPathParent(path), routePath))
```
songgao: This should be moved to inside the confirm modal, after delete succeeds. Otherwise it'd navigate to the parent path even when delete doesn't happen.

shared/fs/really-delete/container.js
```diff
@@ -0,0 +1,44 @@
...
+type OwnProps = Container.RouteProps<{path: Types.Path}, {}>
+
+const mapStateToProps = state => {
+  return {
+    _deleting: anyWaiting(state, Constants.deleteWaitingKey),
+    title: 'Confirmation',
+  }
+}
+
+const mapDispatchToProps = dispatch => ({
```
songgao: This actually accepts the `ownProps` as well:

```
const mapDispatchToProps = (dispatch, ownProps) => ({
...
```

So you can get the `path` here instead of passing from `mergeProps`.

Another thing to consider is, if `path` is not present in route props, we'd end up having the default path here, which is no fun to delete! So perhaps in the action here we should do something like `path !== Constants.defaultPath && dispatch(...`.

shared/fs/common/index.stories.js
```diff
@@ -19,6 +19,7 @@ import {type OwnProps as PathItemIconOwnProps} from './path-item-icon-container'
 import {type OwnProps as PathItemInfoOwnProps} from './path-item-info-container'
 import SyncStatus from './sync-status'
 import PieSlice from './pie-slice'
+import ReallyDelete from '../really-delete'
```
songgao: Nit: Would you mind either moving `fs/really-delete` to `fs/common/really-delete`, or make a separate stories file in `fs/really-delete/index.stories.js`? It'd be nice to keep this story file only for fs/common components.

shared/fs/really-delete/container.js
```diff
@@ -0,0 +1,44 @@
...
+import * as Container from '../../util/container'
+import ReallyDelete from '.'
+import {anyWaiting} from '../../constants/waiting'
+import * as Types from '../../constants/types/fs'
+import * as Constants from '../../constants/fs'
+
+type OwnProps = Container.RouteProps<{path: Types.Path}, {}>
+
+const mapStateToProps = state => {
+  return {
```
songgao: Could use shorthand form `=>({` ?

shared/fs/really-delete/index.js
```diff
@@ -0,0 +1,49 @@
...
+const _Spinner = (props: Props) => (
+  <MaybePopup onClose={props.onBack}>
+    <Box
+      style={{...globalStyles.flexBoxColumn, alignItems: 'center', flex: 1, padding: globalMargins.xlarge}}
+    >
+      <ProgressIndicator style={{width: globalMargins.medium}} />
+    </Box>
+  </MaybePopup>
+)
+const Spinner = HeaderOnMobile(_Spinner)
```
songgao: I could't find where this is used. Did you plan to use this somewhere?

shared/fs/really-delete/index.js
```diff
@@ -0,0 +1,49 @@
...
+  path: Types.Path,
+  title: string,
+}
+
+const _Spinner = (props: Props) => (
+  <MaybePopup onClose={props.onBack}>
+    <Box
+      style={{...globalStyles.flexBoxColumn, alignItems: 'center', flex: 1, padding: globalMargins.xlarge}}
+    >
+      <ProgressIndicator style={{width: globalMargins.medium}} />
```
songgao: `Kb.ProgressIndicator` has `type="Small"` for `medium` size.

shared/fs/really-delete/index.js
```diff
@@ -0,0 +1,49 @@
...
+  onBack: () => void,
+  onDelete: () => void,
+  path: Types.Path,
+  title: string,
+}
+
+const _Spinner = (props: Props) => (
+  <MaybePopup onClose={props.onBack}>
+    <Box
+      style={{...globalStyles.flexBoxColumn, alignItems: 'center', flex: 1, padding: globalMargins.xlarge}}
```
songgao: Move these into `styles`?

shared/fs/really-delete/index.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+import * as React from 'react'
+import * as Constants from '../../constants/fs'
+import {Box, ConfirmModal, HeaderOnMobile, Icon, MaybePopup, ProgressIndicator} from '../../common-adapters'
+import {globalStyles, globalMargins, globalColors} from '../../styles'
```
songgao: Mind using `import * as Kb` and `import * as Styles` here? We prefer this style for common-adapters and Styles in new code.

shared/fs/really-delete/index.js
cjb: And prefer Kb.Box2 for new code instead of Kb.Box?

shared/fs/common/path-item-action/menu-container.js
jakob223: Hmm, I realized that I hadn't tested deleting the folder you're currently browsed to, and so when I moved this I tested it [before and after moving this line], and it errors the GUI out. It gets an error in an `OopsNoAccess`, which maybe it tries to make when the folder it's in stops existing. Any thoughts on how to avoid this?

shared/fs/really-delete/index.js
jakob223: I think I just copied this assuming it was boilerplate when I made this thing.. will remove, thanks for catching.

shared/fs/common/path-item-action/menu-container.js
songgao: Good catch! There are two issues: 1) I missed the Oops (and another place) when doing the nav2 work. Sorry! We can't use `routeProps` directly in nav2. Here's a fix: `songgao/fix-nav2-routeProps` feel free to take it in. 2) It seems we need to navigateUp twice if we are inside the folder/file being deleted. This can be plumbed through routeProps when we push the navigationAppend from "Delete". As a result, path-item-action needs to know if it's part of a row or a header. So this would need some plumbing.

I guess it's fine to merge this if we put "delete" behind a feature flag. But if you want to address it in this PR before merge, I'm OK with that too.

Sorry for the bad estimate! This is definitely more than 2 points.

jakob223: I'll address it in this PR. Thanks for the helpful explanations!  🙏

shared/fs/really-delete/index.js
```diff
@@ -0,0 +1,39 @@
...
+  render() {
+    return (
+      !!this.props.path && (
+        <Kb.ConfirmModal
+          confirmText="Yes, delete it."
+          description="It will be gone forever."
+          header={<Kb.Icon type="iconfont-trash" sizeType="Big" color={Styles.globalColors.red} />}
+          onCancel={this.props.onBack}
+          onConfirm={this.props.onDelete}
+          prompt={`Are you sure you want to delete "${Types.pathToString(this.props.path)}"?`}
```
songgao: I think this should use `Types.getPathName` instead of `Types.pathToString`: https://zpl.io/awJWZXg

shared/fs/common/path-item-action/menu-container.js
jakob223: Now that (1) is fixed (thanks!), as soon as the folder you're in gets deleted, an `Oops` appears because the folder we're in doesn't exist. I can't figure out how to clear it from the `ReallyDelete`.

I've plumbed through to be able to tell if it's going to happen / if the pathItemAction is in the header.

shared/fs/common/path-item-action/choose-view.js
```diff
@@ -11,6 +11,7 @@ import Confirm from './confirm-container'
 type OwnProps = {|
   floatingMenuProps: FloatingMenuProps,
   path: Types.Path,
+  position?: 'row' | 'header',
```
songgao: Just realized we might have done the similar thing: https://github.com/keybase/client/pull/17120/files#diff-b8167efbb256878db6c5717a2b0c1a94R15 I might change 'header' to 'screen' when rebasing mine, since this used in the file preview screen too (i.e. more than header)

PR #16936, opened 2019-04-10
make nav icons smaller in nav2 desktop + mobile

buoyad:
@cecileboucheron this is desktop only right? 

r? @keybase/react-hackers


cecileboucheron:
@buoyad No they're meant also for mobile.

buoyad:
@keybase/react-hackers @cecileboucheron changed to include mobile too

PR #16944, opened 2019-04-10
Fix passphrase prompt on PGP decrypt

pzduniak:
It's just a typo so I don't think we need tests for this. Thanks Patrick!

### DIFF
--- a/go/engine/scankeys.go
+++ b/go/engine/scankeys.go
@@ -314,7 +314,8 @@ func (s *ScanKeys) unlockByID(m libkb.MetaContext, id uint64) openpgp.EntityList
 			Reason:   unlockReason,
 			SecretUI: m.UIs().SecretUI,
 		}
-		unlocked, err := skb.PromptAndUnlock(m, parg, nil, s.me)
+		secretStore := libkb.NewSecretStore(m.G(), s.me.GetNormalizedName())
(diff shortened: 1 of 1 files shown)


go/engine/scankeys.go
```diff
@@ -314,7 +314,8 @@ func (s *ScanKeys) unlockByID(m libkb.MetaContext, id uint64) openpgp.EntityList
 			Reason:   unlockReason,
 			SecretUI: m.UIs().SecretUI,
 		}
-		unlocked, err := skb.PromptAndUnlock(m, parg, nil, s.me)
+		secretStore := libkb.NewSecretStore(m.G(), s.me.GetNormalizedName())
```
patrickxb: why this instead of `m.G().SecretStore()`?

patrickxb:
There is a test for this that does look like it is checking that there was a passphrase prompt.  See `TestPGPDecryptWithSyncedKey`.  It might need to change.

go/engine/scankeys.go
pzduniak: SecretStore returns SecretStoreLocked, which is a wrapper for SecretStoreAll. NewSecretStore takes m.G().SecretStore and returns their per-user store.

patrickxb: ok

PR #16949, opened 2019-04-10
Surya/core 10475/stop rpc rework

heronhaye:


shared/actions/platform-specific/index.desktop.js
```diff
@@ -245,7 +245,7 @@ const onConnected = () => {
       argv: process.argv,
       clientType: RPCTypes.commonClientType.guiMain,
       desc: 'Main Renderer',
-      pid: process.pid,
+      pid: SafeElectron.getRemote().process.pid,
```
heronhaye: get pid of main electron process instead of renderer

go/install/stop_linux.go
```diff
@@ -0,0 +1,53 @@
...
+		//      shut down its child processes.
+		err := stopPID(mctx, client.Details.Pid)
+		if err != nil {
+			mctx.Debug("Error killing client %+v: %s", client, err)
+		}
+	}
+
+	if mctx.G().Env.GetRunMode() == libkb.ProductionRunMode {
+		// NOTE killall only inspects the first 15 characters; we need to use pkill -f
+		err := stopProcessInexactMatch(mctx, "keybase-redirector")
```
maxtaco: Doesn't `keybase-redirector` run as root? Will this actually work?

go/install/stop_linux.go
```diff
@@ -0,0 +1,53 @@
...
+	defer cancel()
+	output, err := exec.CommandContext(ctx, "pkill", "-f", process).CombinedOutput()
+	mctx.Debug("Output (pkill -f %s): %s", process, string(output), err)
+	return err
+}
+
+func stopPID(mctx libkb.MetaContext, pid int) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	output, err := exec.CommandContext(ctx, "kill", fmt.Sprintf("%d", pid)).CombinedOutput()
```
maxtaco: should we specify the actual signal we want (rather than `kill`'s default)?

go/install/stop_linux.go
```diff
@@ -0,0 +1,53 @@
...
+		if err != nil {
+			mctx.Debug("Error killing keybase-redirector: %s", err)
+		}
+	}
+}
+
+func stopProcessInexactMatch(mctx libkb.MetaContext, process string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	output, err := exec.CommandContext(ctx, "pkill", "-f", process).CombinedOutput()
```
maxtaco: same as below, with explicit signal type

go/install/stop_windows.go
```diff
@@ -0,0 +1,20 @@
+package install
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/keybase/client/go/libkb"
+)
+
+func StopAllButService(mctx libkb.MetaContext) {
```
maxtaco: does this function actually do anything?  seems a bit like a noop

maxtaco:
Also, there are a lot of test-cases to consider, mainly around updating and auto-updating the app, and having it restart properly.

heronhaye:
>Some thoughts. First, this is a high-risk change. It might subtly break the update process that might get people stuck in the middle of their update cycle. We should let it bake for a long time on all platforms.
Sure, but can we get it in without the shell->rpc change on Linux first?

>Second, I guess we're kinda hosed in the case where the service is crashed. There's no way to make the RPC and no way therefore to kill kbfs. This could be maybe why it was done the way it originally was (sorry not to think of it earlier). I'm not saying we should abandon this approach, it's just something to think about it.
I mentioned this in the original PR. I can make it support both "standalone" client and RPC, it already does this on Mac.

>Third, if Electron is starting off this process off via RPC, and expects to be kill'ed as a result, maybe it should fallback to exiting if it hasn't exited after ~15 seconds?
Exiting just electron, or trying to kill keybase/kbfs itself?

go/install/stop_linux.go
heronhaye: no it's run by the user with suid

go/install/stop_windows.go
heronhaye: apparently opening that file causes kbfsdokan to unmount

maxtaco:
> Sure, but can we get it in without the shell->rpc change on Linux first?

Don't understand.

> I mentioned this in the original PR. I can make it support both "standalone" client and RPC, it already does this on Mac.

OK!

> Exiting just electron, or trying to kill keybase/kbfs itself?

I was thinking just killing itself (since that's pretty easy).

go/install/stop_windows.go
maxtaco: That's right, I remember this

heronhaye:
>Don't understand

So the main "scary" change is changing what that rpc/cli does on Mac and WIndows right? I can revert that and leave the gui running shell commands if you think that's safer. I'd just change what the Linux command is doing to stop everything instead of just the service, no code change for mac/windows.

>I was thinking just killing itself (since that's pretty easy).

It already does this

PR #16950, opened 2019-04-10
Fix a couple people page TODOs in nav2

buoyad:
- Edit profile -> goes to profile now
- Add paper key
- Add user avatar

r? @keybase/react-hackers


shared/actions/people.js
```diff
@@ -12,6 +12,9 @@ import {peopleTab} from '../constants/tabs'
 import {getPath} from '../route-tree'
 import flags from '../util/feature-flags'
 
+// set this to true to have all todo items show up all the time
+const debugTodo = false
```
buoyad: This should help with debugging these in the future

PR #16951, opened 2019-04-10
Prevent user from being able to push secret keys if they have a random password

pzduniak:


go/engine/pgp_import_key.go
```diff
@@ -234,6 +240,17 @@ func (e *PGPKeyImportEngine) Run(m libkb.MetaContext) (err error) {
 	return nil
 }
 
+func (e *PGPKeyImportEngine) checkRandomPassword(mctx libkb.MetaContext) error {
+	random, err := libkb.LoadHasRandomPw(mctx, keybase1.LoadHasRandomPwArg{})
+	if err != nil {
+		return err
+	}
+	if random {
+		return libkb.NewPushSecretWithoutPasswordError("A password is required to upload secret keys to the server.")
```
zapu: After second thought, I don't think we need specific error type here, considering it will not be exportable anyway.

Also let's change the message to "You need to set your password first before uploading secret keys"

zapu:
Needs a frontend review @keybase/reacthackers

shared/reducers/profile.js
```diff
@@ -75,6 +75,8 @@ export default function(state: Types.State = initialState, action: ProfileGen.Ac
       })
     case ProfileGen.updatePgpPublicKey:
       return state.merge({pgpPublicKey: action.payload.publicKey})
+    case ProfileGen.updatePromptForStore:
+      return state.merge({promptForStore: action.payload.promptForStore})
```
buoyad: Could this field have a more specific name? `promptForStore` is confusing because of the `store`.

pzduniak:
My flow setup has to be broken, I didn't see any of this :/

shared/.flowconfig
```diff
@@ -42,6 +42,7 @@ esproposal.class_static_fields=enable
 esproposal.class_instance_fields=enable
 experimental.const_params=true
 esproposal.optional_chaining=enable
+merge_timeout=180
```
buoyad: what's this for?

pzduniak: >also increased merge_timeout in flowconfig because it's impossible to work with flow on Windows otherwise
Basically the codebase is so massive than on systems with slow IO (ie. WSL) flow won't ever start as the "merge" process timeouts after ~115s. After I did this, it seems to be working fine, although that was enough to convince me to move to Linux.

PR #16964, opened 2019-04-11
Non-interactive, automated keybase login provisioning with paper keys

pzduniak:
I don't like how I had to do some of the plumbing but I didn't want to pollute the engine root level namespace that much. Maybe we should just do a NewLoginWithArg instead?

### DIFF
--- a/go/client/cmd_login.go
+++ b/go/client/cmd_login.go
@@ -6,6 +6,7 @@ package client
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"strings"
 	"time"
 
@@ -23,6 +24,14 @@ func NewCmdLogin(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command
 		cli.BoolFlag{
 			Name:  "s, switch",
 			Usage: "switch out the current user for another",
+		},
+		cli.BoolFlag{
+			Name:  "paperkey-from-stdin",
+			Usage: "Automatically provision using a paper key provided in stdin",
+		},
+		cli.StringFlag{
+			Name:  "machine-name",
+			Usage: "Machine name used in automated provisioning",
 		},
 	}
 	cmd := cli.Command{
@@ -50,10 +59,14 @@ type CmdLogin struct {
 	libkb.Contextified
 	Username     string
 	doUserSwitch bool
-	clientType   keybase1.ClientType
-	cancel       func()
-	done         chan struct{}
-	SessionID    int
+
+	paperkeyFromStdin bool
+	machineName       string
+
+	clientType keybase1.ClientType
+	cancel     func()
+	done       chan struct{}
+	SessionID  int
 }
 
 func NewCmdLoginRunner(g *libkb.GlobalContext) *CmdLogin {
@@ -89,6 +102,14 @@ func (c *CmdLogin) Run() error {
 		c.cancel = nil
 	}()
 
+	var paperKey string
+	if c.paperkeyFromStdin {
+		paperKey, err = c.readPaperKeyFromStdin()
+		if err != nil {
+			return err
+		}
+	}
+
 	err = client.Login(ctx,
 		keybase1.LoginArg{
 			Username:     c.Username,
@@ -96,6 +117,9 @@ func (c *CmdLogin) Run() error {
 			ClientType:   c.clientType,
 			SessionID:    c.SessionID,
 			DoUserSwitch: c.doUserSwitch,
+
+			PaperKey:    paperKey,
+			MachineName: c.machineName,
 		})
 	c.done <- struct{}{}
 
@@ -134,6 +158,8 @@ func (c *CmdLogin) ParseArgv(ctx *cli.Context) error {
 		}
 	}
 	c.doUserSwitch = ctx.Bool("switch")
+	c.paperkeyFromStdin = ctx.Bool("paperkey-from-stdin")
+	c.machineName = ctx.String("machine-name")
 	return nil
 }
 
@@ -163,6 +189,19 @@ func (c *CmdLogin) Cancel() error {
 		}
 	}
 	return nil
+}
+
+func (c *CmdLogin) readPaperKeyFromStdin() (string, error) {
+	src := &StdinSource{}
+	if err := src.Open(); err != nil {
+		return "", err
+	}
+	defer src.Close()
+	data, err := ioutil.ReadAll(src)
+	if err != nil {
+		return "", err
+	}
+	return string(data), nil
 }
 
 func (c *CmdLogin) errNoSyncedKey() error {
--- a/go/client/cmd_login.go
+++ b/go/client/cmd_login.go
@@ -30,8 +30,8 @@ func NewCmdLogin(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command
 			Usage: "Automatically provision using a paper key provided in stdin",
 		},
 		cli.StringFlag{
-			Name:  "machine-name",
-			Usage: "Machine name used in automated provisioning",
+			Name:  "device-name",
+			Usage: "Device name used in automated provisioning",
 		},
 	}
 	cmd := cli.Command{
@@ -61,7 +61,7 @@ type CmdLogin struct {
 	doUserSwitch bool
 
 	paperkeyFromStdin bool
-	machineName       string
+	deviceName        string
 
 	clientType keybase1.ClientType
 	cancel     func()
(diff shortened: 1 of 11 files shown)


go/engine/login_provision.go
```diff
@@ -686,6 +706,45 @@ func (e *loginProvision) chooseDevice(m libkb.MetaContext, pgp bool) (err error)
 	devices := partitionDeviceList(ckf.GetAllActiveDevices())
 	sort.Sort(devices)
 
+	if e.arg.PaperKey != "" {
```
maxtaco: please factor this out into a subfunction

go/engine/login_provision.go
```diff
@@ -508,6 +510,24 @@ func (e *loginProvision) deviceName(m libkb.MetaContext) (string, error) {
 		names = upk.AllDeviceNames()
 	}
 
+	// Fully non-interactive flow
+	if e.arg.MachineName != "" {
```
maxtaco: please factor this out into a subfunction

go/engine/login_provision.go
```diff
@@ -43,6 +43,10 @@ type loginProvisionArg struct {
 	DeviceType string // desktop or mobile
 	ClientType keybase1.ClientType
 	User       *libkb.User
+
+	// Used for non-interactive provisioning
+	PaperKey    string
+	MachineName string
```
maxtaco: `DeviceName` rather than `MachineName`

go/engine/login.go
```diff
@@ -144,6 +147,9 @@ func (e *Login) Run(m libkb.MetaContext) (err error) {
 		DeviceType: e.deviceType,
 		ClientType: e.clientType,
 		User:       ueng.User(),
+
+		PaperKey:    e.PaperKey,
+		MachineName: e.MachineName,
```
maxtaco: `DeviceName` not `MachineName`

go/client/cmd_login.go
```diff
@@ -165,6 +191,19 @@ func (c *CmdLogin) Cancel() error {
 	return nil
 }
 
+func (c *CmdLogin) readPaperKeyFromStdin() (string, error) {
```
maxtaco: use the technique we use in `cmd_oneshot` and maybe factor out that function to use in both places.

go/client/cmd_login.go
```diff
@@ -24,6 +25,14 @@ func NewCmdLogin(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command
 			Name:  "s, switch",
 			Usage: "switch out the current user for another",
 		},
+		cli.BoolFlag{
+			Name:  "paperkey-from-stdin",
+			Usage: "Automatically provision using a paper key provided in stdin",
+		},
+		cli.StringFlag{
+			Name:  "machine-name",
```
maxtaco: `device-name`

go/client/cmd_login.go
```diff
@@ -24,6 +25,14 @@ func NewCmdLogin(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command
...
+			Usage: "Machine name used in automated provisioning",
```
maxtaco: `Device name`

go/client/cmd_login.go
pzduniak: Cool, I wanted to ask about it but I started this early in the morning. The stdin stuff seemed really messy considering that it sometimes went into interactive mode.

go/client/cmd_login.go
```diff
@@ -134,9 +158,26 @@ func (c *CmdLogin) ParseArgv(ctx *cli.Context) error {
...
+
 	return nil
 }
 
+func (c *CmdLogin) getOption(ctx *cli.Context, s string) string {
+	v := ctx.String(s)
+	if len(v) > 0 {
+		return v
+	}
+	envVarName := fmt.Sprintf("KEYBASE_%s", strings.ToUpper(strings.Replace(s, "-", "_", -1)))
```
maxtaco: you know what, let's kill the environment variable idea here, since no other cmd_foo is doing it. Sorry about that. So either via the flag, or via stdin....

go/engine/login_test.go
```diff
@@ -3703,11 +3703,73 @@ func TestProvisioningWithSmartPunctuationDeviceName(t *testing.T) {
...
+		LogUI:       tc2.G.UI.GetLogUI(),
+		SecretUI:    fu.NewSecretUI(),
+		LoginUI:     provLoginUI,
+		GPGUI:       &gpgtestui{},
+	}
+	eng := NewLogin(tc2.G, libkb.DeviceTypeDesktop, fu.Username, keybase1.ClientType_CLI)
+	eng.PaperKey = loginUI.PaperPhrase
+	eng.DeviceName = "a different device name"
+	m2 := NewMetaContextForTest(tc2).WithUIs(uis2)
+	if err := RunEngine2(m2, eng); err != nil {
```
maxtaco: prefer `require.NoError` these days

go/engine/login_test.go
```diff
@@ -3703,11 +3703,73 @@ func TestProvisioningWithSmartPunctuationDeviceName(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	assertNumDevicesAndKeys(tc, fu, 3, 6)
+	if err := AssertProvisioned(tc2); err != nil {
+		t.Fatal(err)
```
maxtaco: ditto

go/engine/login_test.go
```diff
@@ -3703,11 +3703,73 @@ func TestProvisioningWithSmartPunctuationDeviceName(t *testing.T) {
...
+	}
+	if provLoginUI.CalledGetEmailOrUsername != 0 {
```
maxtaco: prefer `require.Equal`

go/engine/login_provision.go
```diff
@@ -747,6 +776,45 @@ func (e *loginProvision) chooseDevice(m libkb.MetaContext, pgp bool) (err error)
...
+	// ... then match it to the paper keys that can be used with this account
+	for _, d := range devices {
+		if d.Type != libkb.DeviceTypePaper {
+			continue
+		}
+		if prefix != *d.Description {
+			continue
+		}
+
+		// use the KID to find the uid, deviceID and deviceName
```
maxtaco: can only do this once, no need to do it every time through the loop, since it doesn't vary with `d`

go/client/cmd_login.go
```diff
@@ -24,6 +25,14 @@ func NewCmdLogin(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command
 			Name:  "s, switch",
 			Usage: "switch out the current user for another",
 		},
+		cli.StringFlag{
+			Name:  "paperkey",
+			Usage: "DANGEROUS: automatically provision using this paper key",
+		},
+		cli.StringFlag{
+			Name:  "devicename",
+			Usage: "Device name used in automated provisioning",
```
maxtaco: update the online docs to describe what these two params are, how they work together, etc.

go/engine/login_provision.go
pzduniak: It only occurs once, when the key is matched.

maxtaco: can you then refactor this function then to make the flow clearer?  So a little loop to find the key.  Then return `NoPaperKeysError{}` if not found. Then the logic that's currently inside the loop

go/client/cmd_login.go
pzduniak: I like it though! Lots of CI stuff is done using environment variables. I'd totally keep it.

go/client/cmd_login.go
pzduniak: Where should I put that stuff? There's no existing document afaik

maxtaco: Whatever is output in `keybase login --help` should be expanded upon.

maxtaco:
Note there's a legit flow error in CI

go/engine/login_provision.go
```diff
@@ -792,27 +793,32 @@ func (e *loginProvision) preloadedPaperKey(m libkb.MetaContext, devices []*libkb
...
-			case libkb.NotFoundError:
+		matchedDevice = d
+		break
+	}
+
+	if matchedDevice == nil {
+		return libkb.NoPaperKeysError{}
+	}
+
+	// use the KID to find the uid, deviceID and deviceName
```
maxtaco: :+1:

pzduniak:
I've merged master so many times, all remaining failures are flakes.

PR #16966, opened 2019-04-11
fix banner underlining actions on mobile

buoyad:


buoyad:
For hotfix, there's a better fix for this that can go in master

PR #16971, opened 2019-04-11
implement Kb.WithTooltip on mobile

songgao:


shared/common-adapters/with-tooltip.native.js
```diff
@@ -1,7 +1,146 @@
...
+          style={this.props.containerStyle}
+          ref={
+            // $FlowIssue
+            this._clickableRef
+          }
+          className={this.props.className}
+        >
+          <Kb.ClickableBox onClick={this._onClick}>{this.props.children}</Kb.ClickableBox>
+        </Kb.Box>
+        <Kb.Animated from={{}} to={{opacity: this.state.visible ? 1 : 0}}>
```
buoyad: I think this or one if its children should have `pointerEvents="none"`. You can't scroll / tap on anything that's below the tooltip area
tooltip

shared/common-adapters/with-tooltip.native.js
```diff
@@ -1,7 +1,146 @@
...
+            top: constrainTop(c.top + c.height),
+            visible: true,
+          })
+        : this.setState({
+            left: constrainLeft(c.left + c.width / 2 - t.width / 2),
+            top: constrainTop(c.top - t.height),
+            visible: true,
+          }) // default to top center
+
+      this.props.setTimeout(() => this.setState({visible: false}), 3000)
```
buoyad: This needs a `this._mounted` check

songgao:
bump!

shared/common-adapters/with-tooltip.native.js
```diff
@@ -1,7 +1,151 @@
...
+    if (!this.props.showOnPressMobile) {
+      return <Kb.Box style={this.props.containerStyle}>{this.props.children}</Kb.Box>
+    }
+
+    return (
+      <>
+        <Kb.Box
+          style={this.props.containerStyle}
+          ref={
+            // $FlowIssue
```
buoyad: are these `flowIssue`s because the `createRef` calls don't have a type param?

songgao: Thanks Danny; tried it and I think another issue is `.measure` is only available from `View` but not the platform agnostic `Box`. So I'm gonna replace all `Box`es with `View`s too.

PR #16974, opened 2019-04-11
simplefs: add a method and cmd for setting the fs debug level

strib:
cc @patrickxb on the vdebug changes.

To improve KBFS performance, we want to turn off most logging by default, but still let the user set it at runtime.  To do that, we'll move a bunch of logging to vdebug, and then this new command/RPC will let users set it at runtime.

That required making the vdebug logger goroutine-safe, and while I was in there I cleaned up the string constants a bit.

Issue: KBFS-4053


go/kbfs/libkbfs/config_local.go
```diff
@@ -916,10 +917,11 @@ func (c *ConfigLocal) MakeLogger(module string) logger.Logger {
 
 // MakeVLogger implements the logMaker interface for ConfigLocal.
 func (c *ConfigLocal) MakeVLogger(module string) *libkb.VDebugLog {
-	// No need to lock since c.loggerFn is initialized once at
-	// construction. Also resetCachesWithoutShutdown would deadlock.
```
patrickxb: This `resetCachesWithoutShutdown` deadlock is no longer an issue?

strib: That comment had been copy/pasted from the function above this one, and doesn't apply to it.

go/client/cmd_simplefs_set_debug_level.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2018 Keybase, Inc. All rights reserved. Use of
```
songgao: 2019

go/client/cmd_simplefs_set_debug_level.go
```diff
@@ -0,0 +1,68 @@
...
+	libkb.Contextified
+	level string
+}
+
+// NewCmdSimpleFSSetDebugLevel creates a new cli.Command.
+func NewCmdSimpleFSSetDebugLevel(
+	cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	return cli.Command{
+		Name:         "set-debug-level",
+		ArgumentHelp: "<integer_level>",
```
songgao: I wonder if it'd make sense to just have this accept strings `mobile`, 'vlog1', etc.

go/libkb/vdebug.go
```diff
@@ -32,26 +35,57 @@ const (
 	VLog1
 	VLog2
 	VLog3
+
+	VLogNoneString       = "mobile"
```
songgao: Is the plan to have "mobile" equivalent to None forever?

go/client/cmd_simplefs_set_debug_level.go
strib: I didn't want to expose those to users when we ask them to run this command.  Numbers seemed better to me.

go/libkb/vdebug.go
strib: No clue -- I'm just rearranging what was already there.  I don't think it matters right now.

PR #16975, opened 2019-04-11
add labelTooltip to Kb.Switch

songgao:


songgao:
@keybase/react-hackers bump!

songgao:
Oh good point; I guess it makes sense on desktop. On mobile that's the clicking area so can't really do that. Will update.

jakob223:
tooltip

nitty nit: the tooltip is to the left of center on large texts. Feel free to decide it's not worth fixing if it's too hard.

songgao:
> the tooltip is to the left of center on large texts.

That was an intentional change. My thought was that this way the tooltip would appear right above the actual switch. So if user hovers on the switch (as you mentioned) it'd be easier to see. If it's right aligned, the tooltip appears on the right side. Let me know if you don't agree!

PR #16981, opened 2019-04-11
Fix flake in TestChatSearchInbox, fix don't query for aliases

joshblum:
i think this will address the flake people were seeing, if not the extra logging should help track it down


joshblum:
added the fix for correctly parsing the query so we don't also search for aliases of the search query

PR #16984, opened 2019-04-11
give Files on mobile a promotion

songgao:


jzila:
@songgao I think we also need:
https://github.com/keybase/client/blob/01d5ad90e4fc0f43b88692abcf6b76836fc25b43/shared/actions/platform-specific/push.native.js#L64-L70

and:
https://github.com/keybase/client/blob/01d5ad90e4fc0f43b88692abcf6b76836fc25b43/shared/actions/config/index.js#L322

songgao:
Hmm AFAIK nav2 is not supposed to use `navigateTo`. Have you tested this on a nav2 master build?

jzila:
@songgao share text isn't hooked up, ignore that one. The other ones need the nav1 route changed though.

jzila:
@songgao also I think you're right, that shouldn't be `navigateTo` anyway, it should be `createSwitchRouteDef`.

songgao:
> The other ones need the nav1 route changed though.

Hmm I thought we were only gonna do it for nav2? There were issues with nav1 team routing which was why we switched them back.

mmaxim:
@songgao This isn't using the right icon in the nav bar, the Files icon is too big.

songgao:
@mmaxim oh sorry just saw this. Fixed in master.

PR #16990, opened 2019-04-12
Surya/core 10632/linux key splitting

heronhaye:
Follow-up to https://github.com/keybase/client/issues/16564


go/libkb/secret_store_secretservice.go
```diff
@@ -58,8 +62,28 @@ func (s *SecretStoreSecretService) maybeRetrieveSingleItem(mctx MetaContext, srv
...
+	return fmt.Sprintf("ring/%s", username)
+}
+
+func (s *SecretStoreRevokableSecretService) secretlessEKStore(mctx MetaContext, username string) SecretlessErasableKVStore {
+	return NewSecretlessFileErasableKVStore(mctx, s.ekstoreDir(mctx, username))
+}
+
+func (s *SecretStoreRevokableSecretService) ekstore(mctx MetaContext, username string, keyringSecret []byte) ErasableKVStore {
+	keygen := func(mctx MetaContext, noise NoiseBytes) (xs [32]byte, err error) {
+		h := hkdf.New(sha256.New, append(noise[:], keyringSecret...), nil, []byte("Keybase-Derived-LKS-SecretBox-1"))
```
heronhaye: salt is nil, info is context string. uses entropy from both noise and the keyring's secret

heronhaye:
Also we discussed two changes: one to use the hardlink method for atomicness, and one to use chmod on windows. Will investigate both more and make another PR

go/libkb/disable_backup_default.go
```diff
@@ -0,0 +1,9 @@
+// +build !darwin
+
+package libkb
+
+import ()
```
joshblum:
```suggestion
```

go/libkb/secret_store_secretservice.go
```diff
@@ -58,8 +62,28 @@ func (s *SecretStoreSecretService) maybeRetrieveSingleItem(mctx MetaContext, srv
 	return &item, nil
 }
 
-func (s *SecretStoreSecretService) RetrieveSecret(mctx MetaContext, username NormalizedUsername) (secret LKSecFullSecret, err error) {
-	defer mctx.TraceTimed("SecretStoreSecretService.RetrieveSecret", func() error { return err })()
+func (s *SecretStoreRevokableSecretService) ekstoreDir(mctx MetaContext, username string) string {
```
joshblum: ek usually refers to `ephemeralKey`throughout the codebase, could you change the naming up here?

go/libkb/secret_store_secretservice.go
joshblum: can you move the context string to https://github.com/keybase/client/blob/master/go/libkb/constants.go#L602

go/libkb/secret_store_secretservice.go
```diff
@@ -78,15 +102,29 @@ func (s *SecretStoreSecretService) RetrieveSecret(mctx MetaContext, username Nor
...
 	}
-	secretBytes, err := srv.GetSecret(*item, *session)
+	keyringSecret, err := srv.GetSecret(*item, *session)
 	if err != nil {
 		return LKSecFullSecret{}, err
 	}
+
+	ekstore := s.ekstore(mctx, string(username), keyringSecret)
+	var secretBytes []byte
+	err = ekstore.Get(mctx, "key", &secretBytes)
```
joshblum: can you move `"key"` to a const

go/libkb/erasable_kv_store.go
```diff
@@ -44,22 +43,29 @@ const cryptoVersion = 1
 const noiseSuffix = ".ns"
 const storageSubDir = "eraseablekvstore"
 
-func getStorageDir(mctx libkb.MetaContext, subDir string) string {
+func getStorageDir(mctx MetaContext, subDir string) string {
 	base := mctx.G().Env.GetDataDir()
 	// check for iOS
-	if runtime.GOOS == "darwin" && mctx.G().GetAppType() == libkb.MobileAppType {
+	if runtime.GOOS == "darwin" && mctx.G().GetAppType() == MobileAppType {
```
joshblum:
```suggestion
	if runtime.GOOS == "darwin" && mctx.G().IsMobileAppType() {
```

go/libkb/secret_store_secretservice.go
joshblum: drop as a comment in the code?

heronhaye:
thanks for the quick review. Made changes and sanity check checked out.

maxtaco:
Mentioned in person, cases to test:
* kex login
* logout / login
* login after noise file intentionally deleted or corrupted
* paper key provisioning

maxtaco:
Sorry Surya to intervene, can I take one last look at this PR?

heronhaye:
Sure

go/libkb/passphrase_login.go
```diff
@@ -251,8 +245,31 @@ func PassphraseLoginPromptThenSecretStore(m MetaContext, usernameOrEmail string,
...
-	return StoreSecretAfterLoginWithLKSWithOptions(m, n, lks, nil)
+func LoadAdvisorySecretStoreOptionsFromRemote(mctx MetaContext) (options SecretStoreOptions) {
+	options = DefaultSecretStoreOptions()
+
+	var ret struct {
+		AppStatusEmbed
+		RandomPW bool `json:"random_pw"`
+	}
+	err := mctx.G().API.GetDecode(mctx, APIArg{
+		Endpoint:    "user/has_random_pw",
```
zapu: there is a libkb function that calls this but also does cache. I don't know if it would work here though

go/libkb/secret_store_secretservice.go
```diff
@@ -58,8 +63,58 @@ func (s *SecretStoreSecretService) maybeRetrieveSingleItem(mctx MetaContext, srv
...
+	plaintextKeygen := func(mctx MetaContext, noise NoiseBytes) (xs [32]byte, err error) {
+		return sha256.Sum256(noise[:]), nil
+	}
+	return NewFileErasableKVStore(mctx, "ring-identifiers", plaintextKeygen)
+}
+
+func (s *SecretStoreRevokableSecretService) RetrieveSecret(mctx MetaContext, username NormalizedUsername) (secret LKSecFullSecret, err error) {
+	defer mctx.TraceTimed("SecretStoreRevokableSecretService.RetrieveSecret", func() error { return err })()
+
+	identifierKeystore := s.identifierKeystore(mctx)
```
maxtaco: Ok, so let me get this straight.  There's a 1:1 mapping between EraseableKVKeystores and home directories. So we can lookup up a (homedir,username) -> (identifier) mapping without ambiguity by using this second KV store. Can we have some nice commentary about this?

BTW, remind me, how does the eraseableKVStore get erased? Is it different rules for this stuff then it is for ephemeral keys?

go/libkb/passphrase_login.go
```diff
@@ -251,8 +245,31 @@ func PassphraseLoginPromptThenSecretStore(m MetaContext, usernameOrEmail string,
...
+		SessionType: APISessionTypeREQUIRED,
+	}, &ret)
+	if err != nil {
+		mctx.Warning("Failed to load advisory secret store options from remote: %s", err)
+		// If there was an API error, just return the default options.
+		return options
```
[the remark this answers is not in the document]
heronhaye: this just indicates whether we want to store it in the file-based or try to do keyring first. I thought I made the change to default to file-based in case of errors but I didn't push it up yet, will do in next commit. Basically I didn't want the secret store to fail for some ephemeral network failure here.

go/ephemeral/device_ek_storage.go
```diff
@@ -65,9 +64,44 @@ func getLogPrefix(mctx libkb.MetaContext) string {
...
+	skey, err := encKey.SecretSymmetricKey(libkb.EncryptionReasonErasableKVLocalStorage)
+	if err != nil {
+		return fkey, err
+	}
+
+	copy(fkey[:], skey[:])
+	return fkey, nil
+}
+
+func deviceEKKeygen(mctx libkb.MetaContext, noiseBytes libkb.NoiseBytes) (fkey [32]byte, err error) {
```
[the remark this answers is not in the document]
heronhaye: This is just copy/pasted from original code with no changes, I believe it makes the key by xoring the hash of the noise file with the device key.

go/libkb/passphrase_login.go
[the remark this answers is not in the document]
joshblum: I thought the default options would trigger the file based storage

go/libkb/secret_store_secretservice.go
heronhaye: Yes. Will add comments. It doesn't get erased automatically by anything, only when we `ClearSecret` which happens on logout and unclicking "remember my passphrase".

heronhaye:
>I think I could use a little bit more high-level description as to what's going on, as indicated in the review. 

Treating the erasablekvstore as a black box, we have the secret from the passphrase stream `s`. We generate a new 32-byte key `k` and a 32-byte identifier `n`. Create two erasablekvstores,  a keyringstore where the erasableKV SecretBox key is `hkdf(NoiseBytes++k)`, and a identifierstore where the key is `hkdf(NoiseBytes)`. We store `k` in the keyringstore and `n` in the identifierstore.

Then to get the secret, we lookup the identifier (which depends only on the username and the home directory), use that to get `k` from the keyring, and then finally get the secret from the erasablekvstore now that we know `k`.

For clearing secrets, we clear the keyring erasable kv store (just shredding files). If it fails we continue. Then we get and then delete the identifier from the identifier erasable kv store. If this fails we don't continue, because we don't know which one to delete in the keyring. Finally we delete the entry in the keyring using the username and the identifier.

>Did all of those cases test out? 
Yes, when I made an account with nopw, provisioning a new device in all those cases skipped the keyring (though it was enabled and running) and stored the secret in file-based system. And when using the keyring store, when I deleted or modified any component (noise file, keyfile, identifier, keyring), I was logged out and asked to log back in when I restarted service.

>Can we get some unit tests on some of the new functionality, especially to test the error cases? 
Sure. By error cases do you mean when files are corrupted? Should I just check that it returns an error afterwards?

>As for how to avoid moving eraseablekvstore back into libkb, can we make a new directory for secret_store  which depends on libkb and eraseablekvstore? This shouldn't be too bad  since most of libkb interoperates with the secret store via interface  anyways.
Ok, that makes sense. (cc @joshblum)
Should I do it in this PR or another one?

maxtaco:
> Sure. By error cases do you mean when files are corrupted? Should I just check that it returns an error afterwards?

I was just thinking wherever there are major error cases, to try to come up with some unit test coverage to make sure we get the right behavior.

> Should I do it in this PR or another one?

Maybe this one so there's less churn in the code?

joshblum:
@maxtaco we had to move `eraseablekv` store into `libkb` because we instantiate the secret store in `libkb/globals.go`. the interface plumbing won't work unless we do the instantiation in `service/main.go` but we weren't sure if `g` would depend on the secret store being set without the service.

I ran into this issue for the leveldb initialization as well since non-service things that use `g` and need it instantiated. In chat there is a `types` package which defines interfaces, `globals` which has the chat global context so we could separate `libkb` out into several packages to avoid this, but it would be a huge refactor...

maxtaco:
ok to leave erasable KV store as is :'(

heronhaye:
Okay, I added some unit tests. This is going to fail CI until jzila adds gnome-keyring to the AMI though. As far as documentation goes, I'll write something detailed up on the docs page.

PR #16997, opened 2019-04-12
New Button

buoyad:
Flow issues left:
# ~23~ 0


buoyad:
@keybase/react-hackers this is ready for review!

- Redo part of `Button`'s API
  - Remove `backgroundMode`
  - Add `mode` and `backgroundColor`, and change the set of `type`s available
- Update styles to match new designs
  - Add `button.css` for desktop hover effects
- Migrate all buttons in the app

I looked at a lot of the buttons in the app, but not all of them, so this will need some time to bake. 

I was thinking of adding icon-only button support baked in here, but we don't have many of them and usually only needs one style change to work (`minWidth: undefined`). I applied that in a couple places.

- [ ] Update storyshots before merge

cc @keybase/design PTAL. Storybook has all the cases from the zeplin sheet.

shared/common-adapters/button.css
```diff
@@ -0,0 +1,56 @@
...
+  border: 1px solid rgba(77, 142, 255, 0.2);
+}
+.button__border_red:hover {
+  border: 1px solid rgba(255, 77, 97, 0.2);
+}
+.button__border_green:hover {
+  border: 1px solid rgba(55, 189, 153, 0.2);
+}
+.button__border_purple:hover {
+  border: 1px solid rgba(112, 78, 186, 0.2);
```
adamjspooner: The spec says these should be 30%.

shared/common-adapters/button.css
```diff
@@ -0,0 +1,56 @@
...
+  background-color: rgba(255, 77, 97, 0.1);
+}
+.button__underlay_green:hover {
+  background-color: rgba(55, 189, 153, 0.1);
+}
+.button__underlay_purple:hover {
+  background-color: rgba(112, 78, 186, 0.1);
+}
+.button__underlay_black:hover {
+  background-color: rgba(0, 0, 0, 0.1);
```
adamjspooner: The spec says these should be 5%.

buoyad:
Thanks @adamjspooner, updated!

adamjspooner:
LGTM!

buoyad:
This also removes an `alignSelf: center` that was on button, which can cause things like this:
image
fixed by adding `alignItems: center` on the parent or similar. There was a comment to remove it, and I agree that it's better for button to respect `alignItems` from the parent than to do it's own thing.

shared/chat/conversation/messages/exploding-explainer-dialog/index.js
```diff
@@ -30,7 +30,7 @@ export const ExplodingExplainerElement = (props: Props) => (
       </Kb.Text>
       <Kb.Box style={styles.buttonBox}>
         <Kb.Button
-          type="Primary"
+
```
cjb: Hm, are these lines adding a new empty whitespace line? Diff's a bit confusing.

buoyad: urgh, I did find+replace and I guess VS code does this. I can try to fix it before merge, does it make this too hard to review?

shared/profile/confirm-or-pending/index.js
```diff
@@ -57,7 +57,7 @@ const ConfirmOrPending = (props: Props) => {
             </Kb.Text>
           )}
         </>
-        <Kb.Button type="Primary" onClick={props.onCancel} label="Reload profile" />
+        <Kb.Button  onClick={props.onCancel} label="Reload profile" />
```
cjb: Ah, looks like you needed to remove the trailing space in the find/replace

shared/chat/conversation/messages/exploding-explainer-dialog/index.js
cjb: Not a big deal for the review, fixing before merge is fine!

buoyad:
Thanks @cjb!

- Ran `prettier --write` on files with buttons
- Updated storyshots
> Snapshot Summary
 › 426 snapshots updated from 1 test suite.

Merging once green!

PR #17002, opened 2019-04-13
 faster native tab bar

chrisnojima:
Faster tab bar by using a touchable without feedback and use press in
@keybase/react-hackers


chrisnojima:
~5ms -> 2.6ms

PR #17006, opened 2019-04-15
Add "Folders Syncing..." in the top-right of the GUI on desktop

jakob223:
CC @strib is this RPC fine for you?

This shouldn't merge until KBFS-4068 is completed.

Issue: KBFS-4049


protocol/avdl/keybase1/notify_fs.avdl
```diff
@@ -20,6 +20,9 @@ protocol NotifyFS {
   @lint("ignore")
   void FSSyncStatusResponse(FSSyncStatus status, int requestID) oneway;
 
+  @lint("ignore")
+  void FSOverallSyncStatusChanged(FSOverallSyncStatus status) oneway;
```
strib: Is there a chance in the future that you'll need a notification about the local disk usage as well, to display the "out of space" warning?  If so, then I think using a `FolderSyncStatus` here would be better (instead of defining a new type) because it also gives the local disk stats.

protocol/avdl/keybase1/kbfs_common.avdl
```diff
@@ -101,4 +101,10 @@ protocol kbfsCommon {
     array<string> syncingPaths;
     union { null, Time} endEstimate;
   }
+
+  record FSOverallSyncStatus {
+    int64 syncingBytes;
```
strib: These two fields could use either new names, or some comments. I can't figure out which one is supposed to be the total number of bytes we know about that needed syncing, and the amount of those that we've synced so far.  (Or is it the amount we still have left to sync?)

In `PrefetchProgress`, we use:

```
    int64 bytesTotal;
    int64 bytesFetched;
```

which I think it more clear.

protocol/avdl/keybase1/notify_fs.avdl
jakob223: @songgao

protocol/avdl/keybase1/kbfs_common.avdl
jakob223: ok, will rename to that if we don't switch to the other struct entirely.

shared/fs/common/sync-status.js
```diff
@@ -73,7 +72,11 @@ function getTooltip(props: Props): string {
 // TODO: make syncing state actually animate
 const SyncStatus = (props: Props) => (
   <Kb.WithTooltip text={getTooltip(props)}>
-    <Kb.Icon type={getIcon(props)} sizeType={'Small'} color={getColor(props)} />
+    {typeof props.status === 'number' ? (
+      <PieSlice degrees={360 * props.status} animated={true} />
+    ) : (
+      <Kb.Icon type={getIcon(props)} sizeType={'Small'} color={getColor(props)} />
```
buoyad: `{'Small'}` -> `"Small"`

shared/router-v2/header/index.desktop.js
```diff
@@ -61,6 +67,7 @@ class Header extends React.PureComponent<Props> {
               color={this.props.allowBack ? Styles.globalColors.black_50 : Styles.globalColors.black_10}
               onClick={this.props.onPop}
             />
+            <SyncingFolders />
```
buoyad: this should be gated on `flags.kbfsOfflineMode`

shared/fs/common/sync-status.js
```diff
@@ -23,7 +24,7 @@ function getIcon(props: Props): Kb.IconType {
     case 'sync-error':
       return 'iconfont-exclamation'
     default:
-      // TODO: handle icons for the pie timer thing
+      // This case should never be reached.
```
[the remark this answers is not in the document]
jakob223: I only intend to handle all `SyncStatus`s other than `number`s in this function.

go/keybase/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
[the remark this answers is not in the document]
jakob223: oops

protocol/avdl/keybase1/notify_fs.avdl
jakob223: switched to `FolderSyncStatus`.

songgao: Sorry didn't check email/GH last couple days. Sounds good to me!

shared/fs/common/pie-slice.js
```diff
@@ -0,0 +1,84 @@
...
+    borderBottomRightRadius: 6,
+    borderTopRightRadius: 6,
+    left: 6,
+  },
+  rotateContainer: {
+    height: 12,
+    left: 0,
+    position: 'absolute',
+    width: 12,
+    zIndex: -1,
```
songgao: Is there a way we can do this without zIndex? e.g. can we change the order of boxes to have them layered correctly? These are contagious and hard to remove. A couple weeks ago most of them were removed from our code.

shared/router-v2/header/syncing-folders.js
```diff
@@ -0,0 +1,41 @@
...
+        Syncing Folders...
+      </Kb.Text>
+    </Kb.Box2>
+  )
+
+const mapStateToProps = state => {
+  if (state.fs.syncingFoldersProgress.bytesTotal === 0) {
+    return {progress: 1}
+  }
+  return {progress: state.fs.syncingFoldersProgress.bytesFetched / state.fs.syncingFoldersProgress.bytesTotal}
```
songgao: It probably doesn't make a huge difference in this particular case, but it'd be nice to just return `state.fs.syncingFoldersProgress` here and do this work in `mergeProps`. `mapStateToProps` needs to be fast as all connected components need to be iterated through on state updates.

shared/fs/common/pie-slice.js
```diff
@@ -0,0 +1,84 @@
+// @flow
+import * as Kb from '../../common-adapters'
+import * as React from 'react'
+import * as Styles from '../../styles/index'
+
+type Props = {|
+  degrees: number,
+  animated?: boolean,
```
songgao: It seems this is only not true in stories, where you're using a separate `Kb.Animated` component. If you want to test animation in stories, I think having a stateful component in the storybook is a better way than using `Kb.Animated` directly in story, like [this](https://github.com/keybase/client/blob/91136b97814eacb8153c9e7b2d29b0cb92c92c01/shared/common-adapters/switch.stories.js#L32). WDYT?

jakob223: I left this as an option in case the animation is too slow or too CPU-intensive to use in the sync status icons, so we can remove it easily.

shared/fs/common/pie-slice.js
jakob223: Now that I think about it again, there is a way and it's about as simple. Will fix.

shared/fs/common/pie-slice.js
songgao: Ah I see. Makes sense. Would you mind still changing the story to not use a separate `Kb.Animated`? So that it uses exactly the same behavior as in the app.

jakob223: Okay, I've changed the story so that instead of having a separate animated one, there is a button on each one to animate it forward by 72 degrees.

shared/fs/common/index.stories.js
```diff
@@ -130,6 +130,24 @@ const pathItemActionCommonProps = {
...
 const pieSlices = [0, 20, 90, 179, 180, 181, 270, 359, 360]
 
+class PieSliceWrapper extends React.PureComponent<
+  {
+    initialDegrees: number,
+  },
+  {degrees: number}
+> {
+  state = {degrees: this.props.initialDegrees}
+  _onClick = () => this.setState(({degrees}) => ({degrees: (degrees + 72) % 361}))
```
songgao: Did you mean `% 360`?

jakob223: No. 360 is full, 0 is empty.

PR #17007, opened 2019-04-15
message replies

mmaxim:
Implements message replies in the service and UI, replacing the "Quote" option.

**Go**
1. Refactor `Sender` interface by allowing for options to be sent to `Send` and `Prepare` from AVDL types. 
2. Thread these new options through the `Outbox` so they can be used with `NonblockingSender`.
3. Add reply information to text body and supersedes field when given a reply message ID parameter,
4. Add `ReplyFiller` which will fill in the reply message when it can. ~~Right now this charges forward on errors, but this might not be the right play~~ changed. 
5. Hook up cache updates for edit/deletes/explosions of replied to messages.

**JS**
1. Add new component for displaying a reply within a text message. 
2. Add new component for previewing a reply while composing a message.
3. Add new highlight modes when jumping to a selected message. Use `flash` for replies.


go/chat/storage/storage.go
```diff
@@ -1114,6 +1142,53 @@ func (s *Storage) updateUnfurlTargetOnDelete(ctx context.Context, convID chat1.C
...
+	}
+	defer s.Trace(ctx, func() error { return nil }, "getExplodedReplies: num: %d", len(exploded))()
+	var replies []chat1.MessageID
+	for _, msg := range exploded {
+		if !msg.IsValid() {
+			continue
+		}
+		replies = append(replies, msg.Valid().ServerHeader.Replies...)
+	}
+	replyMap := make(map[chat1.MessageID]chat1.MessageUnboxed)
```
joshblum: why not call `updateRepliesAffected` here?

go/chat/replyto.go
```diff
@@ -0,0 +1,195 @@
...
+	// Modify messages
+	for _, msg := range msgs {
+		if replyToID := f.getReplyTo(msg); replyToID != nil {
+			st, err := msg.State()
+			if err != nil {
+				continue
+			}
+			replyTo, found := replyMap[*replyToID]
+			if !found {
+				continue
```
joshblum: should you add the msg to res here anyway?

go/chat/sender.go
```diff
@@ -484,15 +484,42 @@ func (s *BlockingSender) resolveOutboxIDEdit(ctx context.Context, uid gregor1.UI
...
+		header.Supersedes = *replyTo
+		return chat1.MessagePlaintext{
+			ClientHeader: header,
+			MessageBody: chat1.NewMessageBodyWithText(chat1.MessageText{
+				Body:     msg.MessageBody.Text().Body,
+				Payments: msg.MessageBody.Text().Payments,
+				ReplyTo:  replyTo,
+			}),
+			SupersedesOutboxID: msg.SupersedesOutboxID,
+		}
```
joshblum: can you add a `default:` log here so we can catch unexpected behavior?

go/chat/sender.go
```diff
@@ -985,6 +1022,17 @@ func (s *Deliverer) Start(ctx context.Context, uid gregor1.UID) {
...
 			convID := obr.ConvID
+			vis := keybase1.TLFVisibility_PRIVATE
+			if obr.Msg.ClientHeader.TlfPublic {
+				vis = keybase1.TLFVisibility_PUBLIC
+			}
+			conv := newBasicUnboxConversationInfo(convID, chat1.ConversationMembersType_IMPTEAMNATIVE, nil,
+				vis)
+			msgs, err := NewReplyFiller(s.G()).Fill(ctx, uid, conv,
+				[]chat1.MessageUnboxed{chat1.NewMessageUnboxedWithOutbox(obr)})
+			if err != nil && len(msgs) > 0 {
```
joshblum: log error? log if `len(msg) != 1`?

go/chat/replyto.go
```diff
@@ -0,0 +1,195 @@
...
+		fetcher: fetcher,
+	}
+}
+
+func (f *fillerReplyMsgs) add(msgID chat1.MessageID) {
+	f.msgIDs[msgID] = true
+}
+
+func (f *fillerReplyMsgs) fill(ctx context.Context, uid gregor1.UID, conv types.UnboxConversationInfo) (res []chat1.MessageUnboxed, err error) {
+	var msgIDs []chat1.MessageID
```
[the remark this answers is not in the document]
mmaxim: It's just flattening a map.

go/chat/storage/storage.go
mmaxim: Indeed!

go/chat/replyto.go
```diff
@@ -0,0 +1,197 @@
...
+		f.Debug(ctx, "Fill: failed to supersede replies: %s", err)
+		return res, err
+	}
+	replyMap := make(map[chat1.MessageID]chat1.MessageUnboxed)
+	for _, msg := range allMsgs {
+		replyMap[msg.GetMessageID()] = msg
+	}
+
+	// Modify messages
+	for _, msg := range msgs {
```
joshblum: why not replace a message at the index if there is a reply, cleans up the logic around making sure everything is appended and covers any future error cases so we never omit a message here

mmaxim: Sure

shared/chat/conversation/messages/wrapper/index.js
```diff
@@ -43,6 +43,7 @@ export type Props = {|
   authorIsAdmin: ?boolean,
   authorIsOwner: ?boolean,
   centeredOrdinal: boolean,
+  centeredOrdinalHighlightMode: Types.CenterOrdinalHighlightMode,
```
buoyad: Could you drive this without the additional `centeredOrdinal` prop? The container can always pass in `none` if this isn't the right ordinal

mmaxim: Sure

shared/chat/conversation/messages/text/container.js
```diff
@@ -1,22 +1,47 @@
...
   const editInfo = Constants.getEditInfo(state, ownProps.message.conversationIDKey)
   const isEditing = !!(editInfo && editInfo.ordinal === ownProps.message.ordinal)
   return {isEditing}
 }
-const mapDispatchToProps = dispatch => ({})
+
+const mapDispatchToProps = (dispatch, {message}: OwnProps) => ({
+  _onReplyClick: messageID =>
+    dispatch(
+      Chat2Gen.createLoadMessagesCentered({
```
buoyad: Could this clear out the `centeredOrdinalInfo` for this convo? I think that'd make it so messages can flash more than once per mount

shared/chat/conversation/reply-preview/index.js
```diff
@@ -0,0 +1,98 @@
...
+import * as Styles from '../../../styles'
+
+export type Props = {|
+  onCancel: () => void,
+  text: string,
+  username: string,
+|}
+
+const ReplyPreview = (props: Props) => {
+  return (
```
buoyad: This should exit if the message we're replying to explodes / is deleted
image

shared/chat/conversation/messages/text/index.js
```diff
@@ -4,22 +4,52 @@ import * as Types from '../../../../constants/types/chat2'
...
 import * as Styles from '../../../../styles'
 
+type ReplyProps = {
+  edited: boolean,
+  onClick: () => void,
+  text: string,
+  username: string,
+}
+
+const Reply = (props: ReplyProps) => {
```
buoyad: Could this say if the replied-to message exploded or was deleted?

mmaxim: It leaves a placeholder for exploded, and is gone for deleted.

shared/chat/conversation/messages/text/container.js
mmaxim: Done

shared/chat/conversation/reply-preview/index.js
mmaxim: I'll fix on the server

shared/chat/conversation/messages/text/index.js
mmaxim: Fixed

shared/chat/conversation/messages/text/index.js
```diff
@@ -29,7 +30,13 @@ const Reply = (props: ReplyProps) => {
               {props.username}
             </Kb.Text>
           </Kb.Box2>
```
buoyad: :sheep: don't think this top line needs to be here if `props.deleted`

shared/chat/conversation/messages/text/container.js
```diff
@@ -9,16 +9,32 @@ type OwnProps = {|
...
   }
-  return {
-    edited: replyTo.hasBeenEdited,
-    onClick: () => onReplyClick(replyTo.id),
-    text: replyTo.text.stringValue(),
-    username: replyTo.author,
+  switch (replyTo.type) {
+    case 'text':
+      return {
+        deleted: false,
```
buoyad: I think this being `replyTo.exploded` would be an improvement.

PR #17009, opened 2019-04-15
simplefs: add RPC for fetching all sync configs and statuses

strib:
The CLI command output looks like this:

```sh
$ keybase fs sync show
private/strib,jzila
  Syncing enabled
  Status: fully synced

private/strib,songgao
  Syncing enabled
  Status: fully synced

team/keybase.staff_v8
  Syncing configured for this subpath:
  	candidates
		Status: sync in progress
		62.92 MB/75.80 MB (83.01%)
		Estimated time remaining: 2m16s

796.42 GB (87.06%) of the local disk available for caching.
```

Issue: KBFS-4047


go/kbfs/simplefs/simplefs.go
```diff
@@ -2268,6 +2276,69 @@ func (k *SimpleFS) SimpleFSSetFolderSyncConfig(
...
+		[]keybase1.FolderSyncConfigAndStatusWithFolder, len(tlfIDs))
+
+	for i, tlfID := range tlfIDs {
+		config, err := k.config.KBFSOps().GetSyncConfig(ctx, tlfID)
+		if err != nil {
+			return keybase1.SyncConfigAndStatusRes{}, err
+		}
+
+		if config.Mode == keybase1.FolderSyncMode_DISABLED {
+			// Shouldn't happen.
```
jakob223: Since this shouldn't happen, do we not care that the output will look a little strange if this happens?
Maybe we should at least log something?

strib: I turned it into a panic just to make sure we find out if it ever happens.

PR #17013, opened 2019-04-16
Fix filename word wrapping in widget

zapu:


buoyad:
[`overflow-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) seems better for this. That will wrap in the middle of the word only if there isn't another break point available. what do you think @songgao / @jzila?

shared/menubar/files.desktop.js
```diff
@@ -36,6 +36,13 @@ type FilesPreviewProps = {|
   userTlfUpdates: Array<UserTlfUpdateRowProps>,
 |}
 
+// Force word wrapping in long filenames so widget does not expand horizontally.
+const wordWrapFilename = Styles.platformStyles({
+  isElectron: {
+    wordBreak: 'break-word',
+  },
+})
```
jzila: Can you put this in `styles` at the bottom?

zapu: yes! sorry I didn't notice there is a style dict

jzila:
> [`overflow-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) seems better for this. That will wrap in the middle of the word only if there isn't another break point available. what do you think @songgao / @jzila?

@buoyad I can't tell if `overflow-wrap` actually does anything.

zapu:
Screenshot_2019-04-16_18-44-51

screenshot of how this wrapping works on this branch. the last long filename there (starting with `8D6`) has no hyphens or spaces, so it's broken arbitrarily. Otherwise this seems to try to break words at hyphens or spaces. First filename has two very long parts separated by space, so each of these parts is broken wherever, and there's a line break on first space.

jzila:
@zapu have you tried `word-break: break-all`?

zapu:
Screenshot_2019-04-16_18-51-21 break all

i wanted to go with break-word because it seems to keep current behavior in filenames that are "breakable", but does last-resort break-anywhere for all other filenames.

This one seems to look better but I'm afraid this is only in the degenerate case of very long filenames.

jzila:
I'll leave it to @adamjspooner to decide which behavior we want.

adamjspooner:
I'm surprised we missed this in our last pass. I prefer `break-all`, but it's just a preference. Either is better than the current behavior.

buoyad:
@jzila I haven't tried it so can't claim it works but that MDN article makes it sound nice. `wordBreak` is fine with me too

PR #17015, opened 2019-04-16
Out of Space banners

jakob223:
Add banner components for running out of space, for in the widget and in the file rows.

Issue: KBFS-4013
Issue: KBFS-3963


shared/fs/banner/out-of-space-banner/index.js
```diff
@@ -0,0 +1,43 @@
...
+const Banner = (props: Props) => (
+  <Kb.Box2 direction="vertical" fullWidth={true} centerChildren={true} style={styles.banner}>
+    <Kb.Text type="BodySemibold" style={styles.text}>
+      You are out of storage space and some folders could not be properly synced. Make some space and &nbsp;
+      <Kb.Text onClick={props.onRetry} underline={true} type="BodySemibold" style={styles.text}>
+        retry the sync
+      </Kb.Text>
+      .
+    </Kb.Text>
+  </Kb.Box2>
```
songgao: Sorry for not mentioning this earlier, but we should use `Kb.Banner` here, like Errs: https://github.com/keybase/client/blob/master/shared/fs/common/errs-container.js so that if style changes it's easy to fix it in common-adapters

shared/fs/banner/out-of-space-banner/container.js
```diff
@@ -0,0 +1,37 @@
...
+const ConnectedBanner = namedConnect<OwnProps, _, _, _, _>(
+  mapStateToProps,
+  mapDispatchToProps,
+  (s, d, o) => ({...o, ...s, ...d}),
+  'OutOfSpaceBanner'
+)(Banner)
+
+export default ConnectedBanner
+
+// TODO do we actually need this?
```
songgao: Not sure; My understanding is this banner will stay there even if user scrolls down in the folder, which means this won't be part of Rows. If that's correct, then we don't need this (and we can remove the `List2` comment at the beginning of the component file).

@cecileboucheron could you confirm that the banner [here](https://zpl.io/2EPL8on) doesn't scroll with the folder content?

shared/fs/banner/widget-banner/index.js
```diff
@@ -0,0 +1,58 @@
...
+      ) : (
+        <Kb.Text type="BodySemibold" style={styles.text}>
+          You are out of storage space. Make some space then{' '}
+          <Kb.Text onClick={props.onRetry} underline={true} type="BodySemibold" style={styles.text}>
+            retry the sync
+          </Kb.Text>{' '}
+          or unsync some folders.
+        </Kb.Text>
+      )}
+    </Kb.Box2>
```
songgao: Same here; can we use `Kb.Banner` here? Also let's move this to `menubar/` as that's where the `OutOfDate` banner is as well.

shared/fs/banner/out-of-space-banner/index.js
jakob223: Ooh, `Kb.Banner` looks useful!

shared/fs/banner/out-of-space-banner/container.js
songgao: So as Cécile mentioned in chat, this should not be part of the row on desktop, but part of the row on mobile. i.e. we still need these `asRows` stuff as well as keeping height constant.

jakob223: And this is why I shouldn't force push so often 🙈

I've put back said code.

jakob223:
TODO:

- [ ]  fix height issue
- [ ]  get the out-of-space status from an RPC and put it in the state

jakob223:
Decided to wait for KBFS implementation of sending remaining space before parsing it into state.

jakob223:
Okay, this PR now contains both out-of-space and the modified offline banner, and mostly connects them [except for connecting the out-of-space data which we don't have yet anyway].

Issue: KBFS-3968
Issue: KBFS-4013

shared/fs/nav-header/main-banner/container.js
```diff
@@ -0,0 +1,33 @@
...
+
+const mapStateToProps = state => ({
+  _kbfsDaemonStatus: state.fs.kbfsDaemonStatus,
+  _name: state.config.username,
+  _overallSyncStatus: null,
+})
+
+const mapDispatchToProps = dispatch => ({
+  _onRetry: name => () =>
+    dispatch(FsGen.createLoadPathMetadata({path: Types.stringToPath('/keybase/private' + name)})),
```
songgao: Mind adding a comment here that this load triggers a retry?

shared/menubar/space-warning.js
```diff
@@ -0,0 +1,26 @@
...
+        props.bannerType === 'warning'
+          ? 'You have less than 1 GB of' + ' storage space. Make some space, or unsync some folders.'
+          : 'You are' + ' out of storage space. Unsync some folders, or make some space then'
+      }
+      color={props.bannerType === 'warning' ? 'blue' : 'red'}
+      actions={[...(props.onRetry ? [{onClick: props.onRetry, title: 'retry' + ' the sync.'}] : [])]}
+    />
+  )
+
+export default Banner
```
songgao: Looks like this isn't actually used in the menubar. Did you mean to add it somewhere?

shared/menubar/space-warning.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+import * as React from 'react'
+import * as Kb from '../common-adapters'
+
+export type BannerType = 'warning' | 'failure' | 'none'
```
songgao: Perhaps we can just use `Types.MainBannerType` here (and maybe rename it into something more generic like `Types.BannerType`)

shared/menubar/space-warning.js
jakob223: No, since we don't have the data for it yet, I didn't add this one or the other to the widget yet. I figure that's part of KBFS-4073.

shared/menubar/space-warning.js
jakob223: That's a different thing. MainBannerType is offline vs out of space - this is out of space vs low on space.

shared/menubar/space-warning.js
jakob223: I could add a comment to that effect if you like.

shared/menubar/space-warning.js
songgao: I see. Makes sense. Thanks!

Not suggesting a change here but I'm thinking perhaps we can just have the BannerType cover all them, and in Files or menubar, just don't show anything if it's not a banner we want to show (e.g. low space in main window). Not sure if it'd actually be better though since we have only these banners. So really just thinking out loud.

jakob223: I feel like it would be annoying to have it be all one type, since there can be different banners in different places, if e.g. you're offline and have >1GB of space available.

PR #17018, opened 2019-04-16
add assetsearch to stellard rpc interface

xgess:
for https://github.com/keybase/keybase/pull/3712


protocol/avdl/stellar1/remote.avdl
```diff
@@ -294,4 +294,7 @@ protocol remote {
     array<PaymentSummary> pendingPayments;
   }
   DetailsPlusPayments detailsPlusPayments(keybase1.UserVersion caller, AccountID accountID);
+
+  // search by either an assetCode or an issuerAccountID to get back matches from stellard/horizon
```
zapu: can you search by both at the same time?

xgess:
```suggestion
  // search by either an assetCode or an issuerAccountID (or both) to get back matches from stellard/horizon
```

PR #17019, opened 2019-04-16
Simpler Role Picker

MarcoPolo:
This is ready to review @keybase/react-hackers 

Replace all instances of Role Picker with a new simpler role picker. This new role picker is just a normal component you place in your render function and pass in the appropriate props. It should be flexible enough to use in a variety of ways and anywhere you need to use a role picker. It's already replaced all existing instances of the old role picker. The old role picker is gone. 

Callers will technically have to do a bit more work, but luckily this work is type protected. Compare this with the old solution which required careful untyped coordination across multiple screens.

### Context
The old role picker relies on screen and passing data back from different screens to a parent screen from callbacks. This is really brittle, and is currently broken when you try to add someone to a team (and change their role) from the profile tab. (and it was broken in a bunch of other ways too)

### Solution
Create A new component that replaces the existing role picker. It has no state and is not connected.
The new Role Picker exposes a helper that turns it into a floating item on top of anything (most commonly a button).


cjb:
@MarcoPolo The example looks good! But I think there's quite some complexity to the old props that the example use doesn't cover.  e.g. these props:
```
  allowOwner?: boolean,
  allowAdmin?: boolean,
  ownerDisabledExp?: string,
  pluralizeRoleName?: boolean,
  showNotificationCheckbox?: boolean,
  sendNotificationChecked?: boolean,
```
It's these that made me tend to prefer using the controlled role picker to the uncontrolled one, because often you can't find out e.g. whether "allowOwner" should be true without interrogating state, and sometimes even dispatching to load team operations, and this seems like a lot to hand off to some totally unrelated component that is in general uninterested in team capabilities.  You could imagine bugs based on the different parent containers of the same uncontrolled rolepicker doing things to set up these props slightly differently to each other.

Anyway, mainly just encouraging us to think about the worst-case load on the parent connected component and whether it's manageable, and if not whether there are small changes we could make to make it be.

MarcoPolo:
excellet points @cjb.

for the allowOwner/Admin we could probably have a pure function in constants/teams that takes `State` and returns a map of `{[key: role]: (DisabledReason: string)}` (which is what this new simple role picker takes).

for these props: 
```
  showNotificationCheckbox?: boolean,
  sendNotificationChecked?: boolean,
```
I'd argue that this is actually a good reason that this component shouldn't be connected! These props have nothing to do with "what role do you want?". This should be handled by something else. (I think this is only used in the let-someone-in-who's-requested-to-join flow.)

I know in the past we've been fairly quick to adopt connected components, but I think they introduce a lot of complexity, code, and make the components a lot more powerful than they otherwise would be. Sometimes we need that extra power, but most of the time it's probably overkill.

cjb:
@MarcoPolo Cool, sounds like a good plan.

cjb:
Looking good! Some quick desktop styling comments from the "someone else's profile->...->Add to team..." menu

Screen Shot 2019-04-23 at 9 30 24 PM

* role row is squashed up against Add button

* when you change from e.g. reader to writer, the buttons scoot over horizontally because reader is a horizontally-longer word -- perhaps put a minWidth on the "Make <role>" button that covers the longest length?

* I wonder whether we still need to show "Send chat notification" on these.  I think in the live app we'd just left the checkbox off most of the time and had it turned on with no way to turn it off.  @cecileboucheron, what do you think?

* It'd be cool if, in the screenshot, the role modal was on top of the cancel/add buttons behind it, as it is in Zeplin.  Definitely doesn't need to be pixel-perfect or anything, but just somewhere on top of the buttons so you aren't looking at two different Cancels sounds better.

* On "Teams->someteam->Add to people...", it looks like there's no default role selected.  Maybe we should just always default to writer?

* Perhaps clicking outside the dropdown area should close the dropdown?  It doesn't (in the screenshot scenario) and I was kind of expecting it to though I can go either way.
  
(Oh, and we have a release coming up and I don't consider most of these to be blockers.. probably only the first bullet point is.)

cecileboucheron:
Agreed with @cjb's feedback above. A bit more feedback from design:
* Default to Writer when adding people or letting people in
image
* We do still need the checkbox. Two changes: it should be checked by default, and the copy should be more clear, let's do "Announce them in #general|" in big teams and "Announce them in team chat" in small teams.
* This red mention should just be grey (BodySmall) and not transparent + let's close the sentence with a `.`.
image
* The `x` to close the modal window behind should stil work

MarcoPolo:
> It'd be cool if, in the screenshot, the role modal was on top of the cancel/add buttons behind it, as it is in Zeplin. Definitely doesn't need to be pixel-perfect or anything, but just somewhere on top of the buttons so you aren't looking at two different Cancels sounds better.

The role picker does live on top of the thing that spawned it. In the screen shot's case it lives on top on the dropdown button. In other cases it lives on top of the add button.

cecileboucheron:
Mobile feedback: 
* the radio button doesn't work for me. It only works when clicking the label.
* Cancel action should be moved to the top left in a header bar
* Make blue button full-width

MarcoPolo:
I think I addressed all the comments except for Cecile's

> It should probably be using the modal window component on mobile now? The page would be sliding from the bottom on iOS and have a header bar, etc.

Since this isn't a modal or a screen (on purpose), this isn't feasible right now.

cecileboucheron:
Mobile switch could use some left-spacing:
image

Other than that lgtm!

PR #17024, opened 2019-04-17
New email/phone flow: actions, sagas, reducers, store

cjb:
@keybase/react-hackers 

This'll be used by the upcoming PR to integrate the new signup flow.  We rely on notifications (e.g.EngineGen.keybase1NotifyPhoneNumberPhoneNumberAdded) to trigger reloading the state from the service rather than doing it ourselves in the reducer, and Core's gonna add more notification types to cover all the mutation cases.


mmaxim:
I think core can make this more convenient than it is right now by just making a "phone numbers updated" notification so we don't need all these handlers in JS. The store just has a list right now, so the service can just keep sending JS the right contents of that list without the proliferation of a bunch of notifications and handlers.

cc @zapu

zapu:
I agree, sending the entire list over - which will be short, at most 10 numbers which is our hard limit - with every update seems much simpler. Thanks!

buoyad:
@cjb what's the status of this?

jakob223:
@cjb if this branch is abandoned, can I close this PR?

zapu:
My commit (in this PR) adds new notification handlers that the server is already sending, see:
https://github.com/keybase/client/pull/17024/commits/194e7fdcb1a31bbacd3d538f088a3601f287245a#diff-b6e80ce81e6eef0d5f1b02f21723ef61R123
https://github.com/keybase/client/pull/17024/commits/194e7fdcb1a31bbacd3d538f088a3601f287245a#diff-32791bac650c83dd6a5131f097fb5b15R107

it would be good to get this in

buoyad:
@zapu your commit was included with https://github.com/keybase/client/pull/17881.

zapu:
thanks!

PR #17027, opened 2019-04-17
faster avatar

chrisnojima:
Some speedups for avatar
- [x] skip loading call if we aren't using a placeholder
- [x] inline some classes so we can skip mounting
- [x] rAF the load call

### DIFF
--- a/shared/common-adapters/avatar.js
+++ b/shared/common-adapters/avatar.js
@@ -110,37 +110,42 @@ class SharedAskForUserData {
       this._userLastReq = {}
     }
   }
-  _makeCalls = throttle(() => {
-    if (!this._dispatch) {
-      return
-    }
-    const now = Date.now()
-    const oldEnough = now - this._cacheTime
-    const usernames = Object.keys(this._userQueue).filter(k => {
-      const lr = this._userLastReq[k]
-      if (!lr || lr < oldEnough) {
-        this._userLastReq[k] = now
-        return true
+  _makeCalls = throttle(
+    () => {
+      if (!this._dispatch) {
+        return
       }
-      return false
-    })
-    const teamnames = Object.keys(this._teamQueue).filter(k => {
-      const lr = this._teamLastReq[k]
-      if (!lr || lr < oldEnough) {
-        this._teamLastReq[k] = now
-        return true
+      const now = Date.now()
+      const oldEnough = now - this._cacheTime
+      const usernames = Object.keys(this._userQueue).filter(k => {
+        const lr = this._userLastReq[k]
+        if (!lr || lr < oldEnough) {
+          this._userLastReq[k] = now
+          return true
+        }
+        return false
+      })
+      const teamnames = Object.keys(this._teamQueue).filter(k => {
+        const lr = this._teamLastReq[k]
+        if (!lr || lr < oldEnough) {
+          this._teamLastReq[k] = now
+          return true
+        }
+        return false
+      })
+      this._teamQueue = {}
+      this._userQueue = {}
+      if (usernames.length || teamnames.length) {
+        requestAnimationFrame(() => {
+          console.log('aaa making dispatch for avatars', usernames)
+          usernames.length && this._dispatch(ConfigGen.createLoadAvatars({usernames}))
+          teamnames.length && this._dispatch(ConfigGen.createLoadTeamAvatars({teamnames}))
+        })
       }
-      return false
-    })
-    this._teamQueue = {}
-    this._userQueue = {}
-    if (usernames.length) {
-      this._dispatch(ConfigGen.createLoadAvatars({usernames}))
-    }
-    if (teamnames.length) {
-      this._dispatch(ConfigGen.createLoadTeamAvatars({teamnames}))
-    }
-  }, 200)
+    },
+    100,
+    {leading: false}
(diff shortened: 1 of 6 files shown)


shared/common-adapters/avatar.js
```diff
@@ -186,27 +184,28 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps) => {
...
-  ])
+  const style = Styles.isMobile
+    ? ownProps.style
+    : Styles.collapseStyles([
+        ownProps.style,
+        onClick && Styles.platformStyles({isElectron: Styles.desktopStyles.clickable}),
+      ])
 
   let url = stateProps._urlMap ? urlsToImgSet(stateProps._urlMap.toObject(), ownProps.size) : null
+  let load
```
chrisnojima: only load if we're showing a placeholder

shared/common-adapters/avatar.js
```diff
@@ -231,44 +228,12 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps) => {
   }
 }
 
-class AvatarConnector extends React.PureComponent<Props> {
```
chrisnojima: remove this extra class that just does the loading logic

shared/common-adapters/avatar.render.desktop.js
```diff
@@ -5,6 +5,14 @@ import * as Styles from '../styles'
 import type {Props} from './avatar.render'
 
 class AvatarRender extends React.PureComponent<Props> {
+  componentDidUpdate(prevProps: Props) {
```
chrisnojima: removed from shim class above. now use rAF

shared/common-adapters/avatar.render.native.js
```diff
@@ -32,77 +29,63 @@ const sizeToTeamBorderRadius = {
 // Android doesn't handle background colors border radius setting
 const backgroundOffset = 1
 
-class Background extends React.PureComponent<{loaded: boolean, loadingColor: any, borderRadius: number}> {
```
chrisnojima: removing extra classes and just using a helper function

shared/common-adapters/avatar.render.native.js
```diff
@@ -32,77 +29,63 @@ const sizeToTeamBorderRadius = {
...
-}
+const border = ({borderColor, borderRadius}) => (
+  <Kb.Box style={[styles.borderBase, {borderColor, borderRadius}]} />
+)
 
 class AvatarRender extends React.PureComponent<Props, State> {
-  state: State = {loaded: false}
+  constructor(props: Props) {
+    super(props)
+    this.state = {loaded: !!this.props.load}
```
chrisnojima: only load if we need to load

chrisnojima:
@keybase/react-hackers speeds up avatar a little

shared/common-adapters/avatar.render.native.js
```diff
@@ -32,77 +29,63 @@ const sizeToTeamBorderRadius = {
...
-    return (
-      <NativeImage
-        source={this.props.url}
-        onLoadEnd={this.props.onLoadEnd}
-        style={[styles[`image:${this.props.size}`], {borderRadius, opacity}]}
-      />
-    )
-  }
-}
+const background = props => (
```
buoyad: Seems like `PureComponent` would render less than this, do you want to use `React.memo` for these?

shared/common-adapters/avatar.render.native.js
```diff
@@ -112,44 +95,42 @@ class AvatarRender extends React.PureComponent<Props, State> {
...
-            <Border
-              borderColor={this.props.borderColor || globalColors.black_10}
-              borderRadius={borderRadius}
-            />
-          )}
+      <Kb.ClickableBox onClick={this.props.onClick} feedback={false} style={containerStyle}>
+        <Kb.Box style={containerStyle}>
+          {!this.props.skipBackground &&
+            (!this.props.skipBackgroundAfterLoaded || !this.state.loaded) &&
+            background({
```
buoyad: Capitalize and treat like components

chrisnojima: These are not components on purpose

buoyad: Ooooh ok TIL the difference. Why would we ever use jsx over this for functional components? I guess for hooks, now. Well anyway sounds good.

chrisnojima: this isn't really a sub component where some of its props might change. so making it a real component like it used to do just creates more overhead. if you look at desktop its all just one function. this means if the parent renders this does too but thats what happens anyways practically and its a lot faster to just call it and not have this also mount

shared/common-adapters/avatar.js
```diff
@@ -112,37 +110,42 @@ class SharedAskForUserData {
...
+          this._teamLastReq[k] = now
+          return true
+        }
+        return false
+      })
+      this._teamQueue = {}
+      this._userQueue = {}
+      if (usernames.length || teamnames.length) {
+        requestAnimationFrame(() => {
+          console.log('aaa making dispatch for avatars', usernames)
```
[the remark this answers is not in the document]
chrisnojima: done

PR #17032, opened 2019-04-17
pipeline search, reindex modes

joshblum:
patch does the following:

- splits the search process into a pipeline of first reading the index off disk to calculate the current indexed percentage and then searching the conv. this gets the first hits back to the UI much faster than blocking on reading every index before searching
- splits the reindexing modes into pre/post search sync/async and uses `POSTSEARCH_SYNC` as the default for the UI. as long as the search is not cancelled the UI will continue to have conversations reindexed and will return any new results 
- runs selective sync during startup after a short delay since it was unlikely to trigger on mobile

### DIFF
--- a/go/chat/search/indexer.go
+++ b/go/chat/search/indexer.go
@@ -87,10 +87,15 @@ func (idx *Indexer) Start(ctx context.Context, uid gregor1.UID) {
 	}
 	idx.started = true
 	ticker := libkb.NewBgTicker(time.Hour)
-	// run one quickly
-	idx.SelectiveSync(ctx, uid, false /*forceReindex */)
 	go func() {
 		idx.Debug(ctx, "starting SelectiveSync bg loop")
+
+		// run one quickly
+		select {
+		case <-time.After(libkb.DefaultBgTickerWait):
+			idx.SelectiveSync(ctx, uid, false /*forceReindex */)
+		}
+
 		for {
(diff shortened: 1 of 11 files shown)


go/chat/search/indexer.go
```diff
@@ -548,115 +607,183 @@ func (idx *Indexer) Search(ctx context.Context, uid gregor1.UID, query string, o
...
+					return nil, err
+				}
+			// reindex all of the conversations and research to populate any new results
+			case chat1.ReIndexingMode_POSTSEARCH_SYNC:
+				// ignore any fully indexed convs since we respect
+				// opts.MaxConvsSearched
+				if convIdx.PercentIndexed(conv.Conv) == 100 {
+					continue
+				}
+				convIdx, totalPercentIndexed, err = idx.reindexConvWithUIUpdate(ctx, conv.Conv, uid,
```
mmaxim: I wonder if this whole thing could be in another goroutine in the errgroup.

joshblum: yeah that sounds like a good idea, i'll leave the postsearch async as is and rework post search sync to be in a thread

joshblum: punting on this for simplicity, main benefit is when initially read out the indices from disk

joshblum:
merging over flake

PR #17036, opened 2019-04-17
android: Missing break statement causing fallthrough

jzila:
r? @cjb @MarcoPolo


jzila:
Going to merge this because it's fixing a bug.

PR #17037, opened 2019-04-17
add os info and processes to logdump

heronhaye:


go/libkb/log_send.go
```diff
@@ -573,3 +570,40 @@ func (l *LogSendContext) mergeExtendedStatus(status string) string {
 	}
 	return string(fullStatus)
 }
+
+func keybaseProcessList() string {
+	ret := ""
+	osinfo, err := ioutil.ReadFile("/etc/os-release")
```
maxtaco: what about windows?

maxtaco: can this be made to work on other platforms reliably?

heronhaye: No, this is only a Linux thign. I added this because in the case of Linux, it's usually helpful to know what distro the user is on. For windows/mac if we wanted to get the os version we'd have to do something more platform specific.

maxtaco: Right, is there something similar to do per-platform here?

heronhaye: Sure, it looks like the executables `sw_vers` on osx and `systeminfo` on windows. I can add these in + ask people to test.

maxtaco: Sure, maybe you should also have a mac machine you can use with a windows VM.

heronhaye: zapu says `systeminfo` takes over 10 seconds on windows. there's a `ver` but it's a shell built-in, not a binary so it can't be called from go. But `sw_vers` on Mac works well, ok to just add that one?

maxtaco: But as it turns out, we need this most on windows. Why not wait 10s?

heronhaye: Okay, we can do that as well.

heronhaye: ok so on windows we have something like ```
   "OSVersion": "Microsoft Windows [Version 10.0.17763.379]",
```
and on OSX something like
```   "OSVersion": "10.13.6-17G65",
```

Ok to not add anything else then?

maxtaco: which tab does this show up in?

heronhaye: Status

go/libkb/log_send.go
```diff
@@ -529,20 +538,10 @@ func (l *LogSendContext) LogSend(statusJSON, feedback string, sendLogs bool, num
 		if mergeExtendedStatus {
 			statusJSON = l.mergeExtendedStatus(statusJSON)
 		}
-	} else {
-		kbfsLog = ""
```
heronhaye: I also got rid of this else branch, as far as I can tell it doesn't actually do anything.

heronhaye:
@maxtaco  ready to look at again

PR #17041, opened 2019-04-17
libfs: handle the .kbfs_status file as a wrapped file

strib:
This creates a generic libfs framework for handling TLF special files (at any level of the hierarchy inside the TLF), and uses it to serve per-TLF `.kbfs_status` files.  So now we can do this:

```
$ keybase fs read /keybase/public/chris/.kbfs_status
{
  "Staged": false,
  "BranchID": "00000000000000000000000000000000",
  "HeadWriter": "chris",
  "DiskUsage": 9054562209,
  "RekeyPending": false,
  "LatestKeyGeneration": -1,
  "FolderID": "7f4f5b66a042b95736b6ab7c98afd317",
  "Revision": 30887,
  "LastGCRevision": 30886,
  "MDVersion": 3,
  "RootBlockID": "01b4f034998d056df9066f0d9e4efa9b5897e2b5475dcb22c31b7c81f518a19f8d",
  "SyncEnabled": false,
  "PrefetchStatus": "TriggeredPrefetch",
  "UsageBytes": 209330339990,
  "ArchiveBytes": 27287819224,
  "LimitBytes": 268435456000,
  "GitUsageBytes": 2480949046,
  "GitArchiveBytes": 170731329,
  "GitLimitBytes": 107374182400,
  "LocalTimestamp": "2019-03-25T10:11:13.01239706-07:00",
  "DirtyPaths": null,
  "Unmerged": null,
  "Merged": null
}
```

### DIFF
--- a/go/kbfs/libfs/wrapped_read_file.go
+++ b/go/kbfs/libfs/wrapped_read_file.go
@@ -1,0 +1,130 @@
+// Copyright 2019 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libfs
+
+import (
+	"context"
+	"io"
+	"time"
+
+	"github.com/keybase/client/go/kbfs/libkbfs"
+	"github.com/keybase/client/go/logger"
+	"github.com/pkg/errors"
+	billy "gopkg.in/src-d/go-billy.v4"
+)
+
+const (
+	// Debug tag ID for an individual FS operation.
+	ctxFSOpID = "FSID"
+)
+
+type ctxFSTagKey int
+
+const (
+	ctxFSIDKey ctxFSTagKey = iota
+)
+
+// wrappedReadFile is a read-only file that serves data from a given
+// `reader` function.
+type wrappedReadFile struct {
+	name   string
+	reader func(context.Context) ([]byte, time.Time, error)
+	log    logger.Logger
+
+	// TODO: proper locking for `nextRead` if we ever use it outside
+	// of libkbfs node-wrapping (which always uses `ReadAt`).
+	nextRead int
+}
+
+var _ billy.File = (*wrappedReadFile)(nil)
+
+func (wrf *wrappedReadFile) getDataAndTime() ([]byte, time.Time) {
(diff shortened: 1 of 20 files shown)


go/kbfs/libfs/wrapped_read_file.go
```diff
@@ -0,0 +1,130 @@
...
+	return 0, errors.New("wrapped read files can't be written")
+}
+
+func (wrf *wrappedReadFile) Read(p []byte) (n int, err error) {
+	data, _ := wrf.getDataAndTime()
+
+	if wrf.nextRead >= len(data) {
+		return 0, io.EOF
+	}
+	n = copy(p, data[wrf.nextRead:])
```
jzila: Doesn't this risk that `getDataAndTime` could change between calls to `Read`?

Same comment re: `ReadAt`, `Seek`, `Len`, and `GetInfo`.

go/kbfs/libgit/autogit_node_wrappers.go
```diff
@@ -147,16 +147,26 @@ func (rdn *repoDirNode) ShouldCreateMissedLookup(
...
+			am:        rdn.am,
+			gitRootFS: rdn.gitRootFS,
+			repo:      rdn.repo,
+			hash:      plumbing.NewHash(commit),
+		}
+		f := rcn.GetFile(ctx)
+		if f == nil {
+			rdn.am.log.CDebugf(ctx, "Error getting commit file")
+			return rdn.Node.ShouldCreateMissedLookup(ctx, name)
+		}
```
jzila: Nice fix!

go/kbfs/libfs/wrapped_read_file.go
strib: Yes, but so could any file you're reading on a file system (when it's modified by some other process).  (And the previous libfuse-based code had the same issue, where it read the file once on a stat and then again on each read, so I don't think this introduces anything new.)

This is also the reason we dial the kernel caching duration to 0 for all of this data.

jzila: OK fair enough.

PR #17043, opened 2019-04-17
logs

heronhaye:
@zapu if you wanted to look (but don't worry about it if you're busy). It's mostly adding logging and making CLI command nicer to use.

### DIFF
--- a/go/libkb/interfaces.go
+++ b/go/libkb/interfaces.go
@@ -701,9 +701,9 @@ type TeamAuditor interface {
 type TeamBoxAuditor interface {
 	AssertUnjailedOrReaudit(m MetaContext, id keybase1.TeamID) (didReaudit bool, err error)
 	IsInJail(m MetaContext, id keybase1.TeamID) (bool, error)
-	RetryNextBoxAudit(m MetaContext) (err error)
-	BoxAuditRandomTeam(m MetaContext) (err error)
-	BoxAuditTeam(m MetaContext, id keybase1.TeamID) (err error)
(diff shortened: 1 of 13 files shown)


go/libkb/interfaces.go
```diff
@@ -701,9 +701,9 @@ type TeamAuditor interface {
 type TeamBoxAuditor interface {
 	AssertUnjailedOrReaudit(m MetaContext, id keybase1.TeamID) (didReaudit bool, err error)
 	IsInJail(m MetaContext, id keybase1.TeamID) (bool, error)
-	RetryNextBoxAudit(m MetaContext) (err error)
-	BoxAuditRandomTeam(m MetaContext) (err error)
-	BoxAuditTeam(m MetaContext, id keybase1.TeamID) (err error)
+	RetryNextBoxAudit(m MetaContext) (attempt *keybase1.BoxAuditAttempt, err error)
+	BoxAuditRandomTeam(m MetaContext) (attempt *keybase1.BoxAuditAttempt, err error)
+	BoxAuditTeam(m MetaContext, id keybase1.TeamID) (attempt *keybase1.BoxAuditAttempt, err error)
```
zapu: do we need to return pointers here?

also these will not be shared by the same object that's in the box audit logs, I noticed that BoxAuditAttempt has some pointers in there, so there might still be some unwanted data sharing. Should we ensure we use deep copy to store these?

heronhaye: not really, it was to distinguish the case where the attempt was never even attempted. but I can make a field inside attempt for that

zapu: ah then my point is moot

PR #17044, opened 2019-04-17
Multi-user support in keybase passphrase recover

pzduniak:
~still pending tests~


go/engine/login_with_paperkey_test.go
```diff
@@ -121,8 +121,42 @@ func TestLoginWithPaperKeyLoggedInAndLocked(t *testing.T) {
 	AssertDeviceKeysLock(&tc, true)
 }
 
+// full flow, login with username
```
joshblum: can you test an invalid username? what about a valid username that is not provisioned on this device? do we handle the error correctly in the cli?

pzduniak:
Work will continue in #17165

PR #17045, opened 2019-04-17
simplefs: serve the overall .kbfs_status file

strib:
Based on #17041.

Reuse as much SimpleFS code as possible by creating a new `RootFS` to open and make a file representing the overall status file, and use that when a root `/keybase` path is given to SimpleFS.

Example:
```
$ keybase fs read /keybase/.kbfs_status 
{
  "CurrentUser": "strib",
  "IsConnected": true,
  "UsageBytes": 209330339990,
  "ArchiveBytes": 27287819224,
  "LimitBytes": 268435456000,
...
```

Issue: KBFS-4028

### DIFF
--- a/go/kbfs/libfs/root_fs.go
+++ b/go/kbfs/libfs/root_fs.go
@@ -132,7 +132,9 @@ func (rfs *RootFS) Readlink(_ string) (target string, err error) {
 
 // Chroot implements the billy.Filesystem interface for RootFS.
 func (rfs *RootFS) Chroot(_ string) (newFS billy.Filesystem, err error) {
-	return nil, errors.New("RootFS cannot read links")
+	// Don't allow chroot'ing anywhere outside of the root FS since we
+	// haven't yet implemented folderlist browsing.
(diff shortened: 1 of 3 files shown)


go/kbfs/libfs/root_fs.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// Readlink implements the billy.Filesystem interface for RootFS.
+func (rfs *RootFS) Readlink(_ string) (target string, err error) {
+	return "", errors.New("RootFS cannot read links")
+}
+
+// Chroot implements the billy.Filesystem interface for RootFS.
+func (rfs *RootFS) Chroot(_ string) (newFS billy.Filesystem, err error) {
+	return nil, errors.New("RootFS cannot read links")
```
jzila: `RootFS cannot chroot` (but also, why not?)

strib: Because it could only chroot up from the root FS into say a folder list like `/private`, or all the way into the TLF, and I don't see a need to implement all that right now. Seems better to just prevent it.

PR #17048, opened 2019-04-18
index memcache, debounce

joshblum:
- adds a in memory lru cache to the index store
- adds a short delay before running the search in case we can canceled by a subsequent search.

i'm not sure the delay buys us anything, wouldn't it be better for the GUI to debounce it's calls or for the server to keep some state instead of delay every search 50ms? unless i misunderstood the ticket..

### DIFF
--- a/go/chat/search/storage.go
+++ b/go/chat/search/storage.go
@@ -60,6 +60,12 @@ func newStore(g *globals.Context) *store {
 	}
 }
 
+func (s *store) ClearCache() {
+	s.Lock()
+	defer s.Unlock()
+	s.lru.Purge()
+}
+
 func (s *store) memKey(convID chat1.ConversationID, uid gregor1.UID) string {
 	return fmt.Sprintf("idx:%s:%s", uid, convID)
 }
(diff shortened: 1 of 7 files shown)


mmaxim:
I don't want the UI to be in charge of anything related to how the search is coordinated. I also want the name search to go on every keypress, but the name search can be sagotaged by the index search starting/stopping. 50ms seems like it shouldn't matter, but if it does we can re-evaluate.

go/chat/search/storage.go
```diff
@@ -46,17 +48,32 @@ func newStore(g *globals.Context) *store {
 	dbFn := func(g *libkb.GlobalContext) *libkb.JSONLocalDb {
 		return g.LocalChatDb
 	}
+	nlru, err := lru.New(1000)
```
mmaxim: Are you sure this isn't too big?

joshblum:
oh makes sense, forgot about the conv search aspect

go/chat/search/storage.go
joshblum: could dial it back a bit if we see an issue on mobile but was hoping we fit most inboxes since otherwise the cache is useless and we'll just thrash on searches

joshblum: dropped down to 250 and moved to the indexer instead of storage so we just use it on searches

PR #17050, opened 2019-04-18
Rename subteam dialog

buoyad:
~Desktop nav2 only - mobile nav2 will come along for the ride with new header design.~ Desktop + mobile / nav1 + nav2

Adds a pencil icon to the right of the teamname on the team page for subteams you can rename.
image
image

r? @keybase/react-hackers


shared/teams/rename-team/index.js
```diff
@@ -0,0 +1,181 @@
...
+    if (!this.props.waiting && prevProps.waiting && !this.props.error) {
+      // finished, go back
+      this.props.onSuccess(this._newFullName())
+    }
+  }
+
+  _newFullName = () => [this._prefix, this.state.newName].join('.')
+  _onChangeText = newName => this.setState({newName})
+  _disabled = () => this.state.newName.length < 2
+  _validateTeamname = () => {
```
buoyad: The service errors for these are essentially the same with a bunch of unfriendly codes and stuff. This should catch the most common errors so they can have nice messages.

cjb:
Maybe show the resulting full team name a little more explicitly?  Create subteam puts up a "This team will be named **team.sub**" string while you type -- could put that in place of the current "Subteam of" string?  Or it could all be done inside the input field with placeholder somehow.  No need to put a lot of effort in, just a little more than there is now to show that you only need to (and can only) be writing the new sub name in there.

shared/teams/routes.js
```diff
@@ -161,5 +161,10 @@ export const newModalRoutes = {
     getScreen: () => require('./team/really-remove-member/container').default,
     upgraded: true,
   },
+  teamRename: {
+    // TODO add to nav1 routes too
```
cjb: Can never tell whether a TODO is meant for before or after the merge :)  If after, perhaps file a ticket and put the ticket ID in this comment?  And if before, put up a checkbox in the PR desc?

buoyad: Meant for before merge, done

buoyad:
Thanks @cjb, I added the teamname preview and made it work everywhere (nav1/2/desktop/mobile). PTAL

PR #17052, opened 2019-04-18
identify3: fix unhandled key track mode for eldest reset

maxtaco:
- also add some debug


go/identify3/adapter.go
```diff
@@ -333,6 +333,7 @@ func (i *UIAdapter) displayKey(mctx libkb.MetaContext, key keybase1.IdentifyKey)
 	if key.PGPFingerprint == nil {
 		return
 	}
+	mctx.Debug("ID3: displayKey(key: %+v)", key)
```
mmaxim: Is this needed?

maxtaco: i'd really like it, since it'll say for sure if this is the right fix.  I'm only like 75% sure. We only hit this spam when you view someone's profile, not on team load, let's say

PR #17053, opened 2019-04-18
OnDbNuke hook

joshblum:
Adds `DbNukeHooks` to `GlobalContext` to unify clearing of caches.

### DIFF
--- a/go/stellar/global.go
+++ b/go/stellar/global.go
@@ -27,7 +27,9 @@ func ServiceInit(g *libkb.GlobalContext, walletState *WalletState, badger *badge
 	if g.Env.GetRunMode() != libkb.ProductionRunMode {
 		stellarnet.SetClientAndNetwork(horizon.DefaultTestNetClient, build.TestNetwork)
 	}
-	g.SetStellar(NewStellar(g, walletState, badger))
+	s := NewStellar(g, walletState, badger)
+	g.SetStellar(s)
+	g.AddLogoutHook(s, "stellar")
 }
 
 type Stellar struct {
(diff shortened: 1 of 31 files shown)


go/stellar/global.go
```diff
@@ -27,7 +27,9 @@ func ServiceInit(g *libkb.GlobalContext, walletState *WalletState, badger *badge
 	if g.Env.GetRunMode() != libkb.ProductionRunMode {
 		stellarnet.SetClientAndNetwork(horizon.DefaultTestNetClient, build.TestNetwork)
 	}
-	g.SetStellar(NewStellar(g, walletState, badger))
+	s := NewStellar(g, walletState, badger)
+	g.SetStellar(s)
+	g.AddLogoutHook(s, "stellar")
```
joshblum: cc @mlsteele @patrickxb is there anything we want to clear in stellar on a dbnuke? from a quick pass I didn't see anything obvious

joshblum:
nope that happens separately on logout

go/stellar/global.go
patrickxb: stellar doesn't use leveldb.

joshblum: @patrickxb goal is to have the dbnuke hook clear any disk/memory caches not just leveldb

patrickxb: then you would want to do what happens in service/main.go:OnLogout:

            log("resetting wallet state on logout")
            if d.walletState != nil {
                    d.walletState.Reset(m)                                                    
            }

on that Service.walletState variable.

PR #17058, opened 2019-04-18
add required styles for lineclamp and use in chat + teams header

buoyad:
This works nicely in the headers. I also tried it in some random places around the app with different parent styles - it worked everywhere I checked. No external styles required :tada: 

r? @keybase/react-hackers


adamjspooner:
This fixes the bug where long unfurls weren't properly clamped too. Nice!

PR #17063, opened 2019-04-18
Fix some buttons

buoyad:
r? @keybase/react-hackers @cecileboucheron


shared/wallets/send-form/footer/index.js
```diff
@@ -86,8 +87,10 @@ const Footer = (props: Props) => {
             />
           )}
           {!!props.onClickSend &&
-            (props.disabled ? (
```
buoyad: @adamjspooner this `disabled` is also when the form is invalid, I added a more specific bool here

PR #17071, opened 2019-04-19
continue on errors and roundup if we exit on sync

mmaxim:


go/chat/search/indexer.go
```diff
@@ -851,6 +853,11 @@ func (idx *Indexer) Search(ctx context.Context, uid gregor1.UID, query string, o
 				hitMap[convIDStr] = *convHits
 			}
 		}
+		if totalPercentIndexed >= 95 && opts.ReindexMode == chat1.ReIndexingMode_POSTSEARCH_SYNC {
```
joshblum: we shouldn't do this, it will just mask real bugs.

PR #17072, opened 2019-04-19
Stellar: add, remove, change limit of trustlines

zapu:
This PR adds the following RPCs for Stellar:
  - addTrustlineLocal(accountID, asset[, limit])
  - deleteTrustlineLocal(accountID, asset)
  - changeTrustlineLimitLocal(accountID, asset, limit)

They will be used to manage trustlines for Stellar accounts. AccountID argument can be skipped, primary account will be used then. `addTrustlineLocal` takes an optional `limit` argument, when it's omitted, default max limit is used.

The RPCs create trustline operations using `stellarnet.CreateTrustlineTransaction` and `stellarnet. DeleteTrustlineTransaction`. Internally, same Stellar operation type is used. Delete is just a `ChangeTrustOp` with limit=0. On stellard side, `ChangeTrustline` RPC is used to send the signed transaction.

Command line commands are provided for all 3 operations.

I didn't add `ListTrustlines` RPC because trustlines are already returned by `BalancesLocal` and `GetAccountAssetsLocal`. But we can add one that returns only custom asset trustlines if that's desired.


protocol/avdl/stellar1/local.avdl
```diff
@@ -519,6 +519,12 @@ protocol local {
   // airdropRegisterLocal registers a user for an airdrop.
   void airdropRegisterLocal(int sessionID, boolean register);
 
+  // Trustline operations:
+  // In addTrustline, limit is optional, if not supplied, max limit will be used.
+  void addTrustlineLocal(int sessionID, AccountID accountID, Trustline trustline, string limit);
+  void deleteTrustlineLocal(int sessionID, AccountID accountID, Trustline trustline);
+  void changeTrustlineLimitLocal(int sessionID, AccountID accountID, Trustline trustline, string limit);
```
patrickxb: I imagine frontend is going to need `listTrustlines`.  I know it is in the balances reply, but to keep it simple for their logic, let's add `listTrustlines` that just returns the non-XLM trustlines for an account.

zapu: ok this shouldn't be too hard to do - thanks!

PR #17073, opened 2019-04-19
go1.12.4 fmt

heronhaye:
Checked out master, switched to 1.12.4, ran make fmt, make lint, and make vet.
Seems new vet rules have been added.

```
go vet ./...
# github.com/keybase/client/go/chat/signencrypt/example
../../code/src/github.com/keybase/client/go/chat/signencrypt/example/main.go:25:3: fail call has possible formatting directive %s
# github.com/keybase/client/go/libkb
../../code/src/github.com/keybase/client/go/libkb/passphrase_stream.go:124:9: Sprintf format %x has arg ps.LksClientHalf() of wrong type github.com/keybase/client/go/libkb.LKSecClientHalf
../../code/src/github.com/keybase/client/go/libkb/passphrase_stream_test.go:68:4: Errorf format %x has arg dk.LksClientHalf() of wrong type github.com/keybase/client/go/libkb.LKSecClientHalf
# github.com/keybase/client/go/service
../../code/src/github.com/keybase/client/go/service/account.go:115:2: struct field Status repeats json tag "status" also at ../libkb/api.go:45
make: *** [Makefile:8: vet] Error 2
```


heronhaye:
todo: copy over git hash into pre-commit config after merging https://github.com/keybase/pre-commit-golang/pull/3

jzila:
@heronhaye I think you need to `make` using go 1.12.4 in the protocol directory.

heronhaye:
@jzila hm, I did that but didn't get a diff.

jzila:
@heronhaye maybe you need to reinstall the binaries.

jzila:
nevermind, it's failing just because CI isn't up to date.

jzila:
I just triggered a rebuild of this with the new instances. Hopefully it passes!

go/kbhttp/srv.go
```diff
@@ -103,6 +104,7 @@ func NewSrv(log logger.Logger, listenerSource ListenerSource) *Srv {
 	return &Srv{
 		log:            log,
 		listenerSource: listenerSource,
+		doneCh:         make(chan struct{}),
```
joshblum: should this be initialized in the `Start` method so the server could be started/stopped multiple times?

go/kbhttp/srv.go
```diff
@@ -152,11 +155,12 @@ func (h *Srv) Addr() (string, error) {
...
 // Stop stops listening on the server's listener source.
-func (h *Srv) Stop() {
+func (h *Srv) Stop() <-chan struct{} {
 	h.Lock()
 	defer h.Unlock()
 	if h.server != nil {
 		h.server.Close()
 		h.server = nil
 	}
+	return h.doneCh
```
joshblum: this will hang if the server never started

jzila: That should be fine. Right now we only block in tests.

go/kbhttp/srv.go
jzila: Is that a usage pattern we've ever done? I can make it do that but it won't address the blocking on a nil channel.

joshblum: we usually have that flow in chat so it stuck out

go/kbhttp/srv.go
joshblum: could close before returning if server is nil

jzila:
@keybase/kbfs-hackers PTAL at the KBFS changes.

PR #17082, opened 2019-04-19
Use user's first name instead of username in "Hi !" string

cjb:
@keybase/react-hackers CC @keybase/design


shared/router-v2/tab-bar/container.js
```diff
@@ -38,6 +38,7 @@ const getBadges = memoize(b => b.toObject())
 
 const mergeProps = (stateProps, dispatchProps, ownProps) => ({
   badgeNumbers: getBadges(stateProps._badgeNumbers),
+  firstname: stateProps.fullname.split(' ')[0],
```
heronhaye: does not work for asian languages

PR #17085, opened 2019-04-19
restrict windows updater process search

heronhaye:


heronhaye:
forgot to change this

PR #17086, opened 2019-04-19
fs: hide syncing folders when offline

jakob223:


jakob223:
untagging review while I fix what caused CI to fail..

PR #17088, opened 2019-04-20
prefetcher: cancel all prefetches for a given TLF

strib:
Previously, when changing a sync config, we would toggle the whole prefetcher to make sure the old requests were canceled.  But this can block for a really long time if there are a lot of requests outstanding.  We were doing this to make sure that when a block no longer needs to be prefetched (due to the config change), it wouldn't continue to be prefetched.  But we can handle this a different way, without all the blocking.

Instead, When a sync config is disabled, or a partial sync config is changed, just cancel all the prefetches associated with that TLF ID. It should be relatively rare, so that it doesn't matter much that we have to loop through all the prefetches when it happens.

Do it synchronously so we can make sure all the requests are canceled before the disk cache is scrubbed.

Issue: KBFS-4070

### DIFF
--- a/go/kbfs/libkbfs/config_local.go
+++ b/go/kbfs/libkbfs/config_local.go
@@ -1390,7 +1390,7 @@ func (c *ConfigLocal) openConfigLevelDB(configName string) (*LevelDb, error) {
 
 func (c *ConfigLocal) loadSyncedTlfsLocked() (err error) {
 	defer func() {
-		c.MakeLogger("").CDebugf(nil, "Loading synced TLFs: %+v", err)
+		c.MakeLogger("").CDebugf(nil, "Loaded synced TLFs: %+v", err)
 	}()
 	syncedTlfs := make(map[tlf.ID]FolderSyncConfig)
 	syncedTlfPaths := make(map[string]bool)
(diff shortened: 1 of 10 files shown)


go/kbfs/libkbfs/config_local.go
```diff
@@ -1389,6 +1389,9 @@ func (c *ConfigLocal) openConfigLevelDB(configName string) (*LevelDb, error) {
 }
 
 func (c *ConfigLocal) loadSyncedTlfsLocked() (err error) {
+	defer func() {
+		c.MakeLogger("").CDebugf(nil, "Loading synced TLFs: %+v", err)
```
jzila: s/Loading/Loaded

go/kbfs/libkbfs/prefetcher_test.go
```diff
@@ -2176,3 +2177,106 @@ func TestPrefetcherWithCanceledDedupBlocks(t *testing.T) {
...
+	select {
+	case <-getA1:
+	case <-ctx.Done():
+		t.Fatal(ctx.Err())
+	}
+
+	t.Log("Cancel the first TLF's prefetches")
+	go func() {
+		notifySyncCh(t, prefetchSyncCh)
+	}()
```
jzila: Any particular reason this isn't
```go
go notifySyncCh(t, prefetchSyncCh)
```

strib: Nope, good call.

PR #17089, opened 2019-04-20
Remaining integration of AccountReset into keybase login

pzduniak:
Effectively this flow is untested, just like other non-authenticated autoreset/enter flows as a nonauth account reset request is the same as not being in the pipeline at all. I initially had password recovery in this branch but stripped it as I'll need both this branch and the multiuser recovery to make it all work.


go/engine/login_provisioned_device.go
```diff
@@ -248,8 +260,35 @@ func (e *LoginProvisionedDevice) run(m libkb.MetaContext) (err error) {
 	return nil
 }
 
-func (e *LoginProvisionedDevice) connectivityWarning(m libkb.MetaContext) {
+func (e *LoginProvisionedDevice) suggestRecovery(mctx libkb.MetaContext) error {
+	// Whatever happens here won't result in a session, so we can mark this engine as a noop.
+	e.resetPending = true
+
+	enterReset, err := mctx.UIs().LoginUI.PromptResetAccount(mctx.Ctx(), keybase1.PromptResetAccountArg{
+		Text: "If you have forgotten your password and either lost all of your devices, or if you " +
```
joshblum: I think that this display text should be separated from the RPC argument and left to the caller to decide. It would be better if the arg contained a reason enum specifying the part of the flow the prompt originated from.

go/engine/login_provision.go
```diff
@@ -748,29 +749,50 @@ func (e *loginProvision) chooseDevice(m libkb.MetaContext, pgp bool) (err error)
...
+				"like to request a reset of your account?",
+		})
+		if err != nil {
+			return err
+		}
+
+		if !enterReset {
+			// Although we didn't enter the pipeline, we don't want the flow to try logging in,
+			// so we mark it as if we have just joined the pipeline.
+			e.resetPending = true
```
joshblum: can you make a second state variable and then expose a method on the engine which the callers use to determine if there is a pending reset or if the reset was skipped? there's not need to overload this one variable, I think it will be confusing in the long run.

pzduniak: I've renamed resetPending and created methods exporting it on engines that get queried on it. wdyt now?

go/client/ui.go
```diff
@@ -819,7 +819,23 @@ func (l LoginUI) DisplayPrimaryPaperKey(_ context.Context, arg keybase1.DisplayP
 }
 
 func (l LoginUI) PromptResetAccount(ctx context.Context, arg keybase1.PromptResetAccountArg) (bool, error) {
-	return l.parent.PromptYesNo(PromptDescriptorResetAccount, arg.Text, libkb.PromptDefaultNo)
+	var msg string
+	switch arg.Kind {
+	case keybase1.ResetPromptType_COMPLETE:
+		msg = "Would you like to complete the reset of your account?"
+	case keybase1.ResetPromptType_ENTER_PROVISION:
+		msg = "The only way to provision this device is with access to one of your existing " +
```
joshblum: prefer string literal ` `` `

protocol/avdl/keybase1/login_ui.avdl
```diff
@@ -8,6 +8,20 @@ protocol loginUi {
   void displayPaperKeyPhrase(int sessionID, string phrase);
   void displayPrimaryPaperKey(int sessionID, string phrase);
 
-  boolean promptResetAccount(int sessionID, string text);
+  enum ResetPromptType {
+    COMPLETE_0,        // reset completion if the countdown is done
+    ENTER_PROVISION_1, // no devices available in provisioning, suggest entering
```
joshblum: maybe `ENTER_NO_DEVICES` instead?

pzduniak:
Extensively manually tested, no issues whatsoever, works as expected.

PR #17102, opened 2019-04-22
Contact syncing

zapu:
Big part of it is transforming phone contact list, where each contact can have multiple numbers/addresses, into a contact list that will be displayed in the UI.

There are few rules:
1) If there is a contact with 1 or more components and none of them resolve, display all components in the list.
2) If a component match to a user, return that user.
3) Deduplicate users, so even if many components match to some user, they shall only be displayed once.


zapu:
@heronhaye @joshblum ready for first look to what's going on here - it's very much undone but it's better if any interface changes arise sooner rather than later

go/phonenumbers/contacts.go
```diff
@@ -0,0 +1,116 @@
...
+				phoneComps = append(phoneComps, phoneToContact{
+					contactIndex:   contactI,
+					componentIndex: compI,
+				})
+			}
+		}
+	}
+
+	// contactIndex -> true for all contacts that have at least one compoonent resolved.
+	contactsFound := make(map[int]bool)
```
zapu: this should probably be a set of uids to dedup users, instead of contacts

go/contacts/contacts.go
```diff
@@ -0,0 +1,151 @@
...
+			// component or another contact.
+			return
+		}
+		contact := contacts[toContact.contactIndex]
+		component := contact.Components[toContact.componentIndex]
+
+		usersFound[lookupRes.UID] = true
+
+		res = append(res, keybase1.ResolvedContact{
+			Name:         lookupRes.Username, // if found, return username
```
joshblum: see the comment in avdl about the `Name` field

protocol/avdl/keybase1/contacts.avdl
```diff
@@ -0,0 +1,44 @@
...
+        string name; // "Bob Goodman"
+        array<ContactComponent> components;
+    }
+    record ResolvedContact {
+        string name; // "Bob Goodman"
+        int contactIndex; // to associate the resolution to input contact entry
+        ContactComponent component;
+
+        // Outputs:
+        union { null, string } err;
```
joshblum: unused?

protocol/avdl/keybase1/contacts.avdl
```diff
@@ -0,0 +1,44 @@
...
+        union { null, RawPhoneNumber } phoneNumber;
+        union { null, EmailAddress } email;
+        string label; // "", "Home", "Work"
+    }
+    record Contact {
+        string name; // "Bob Goodman"
+        array<ContactComponent> components;
+    }
+    record ResolvedContact {
+        string name; // "Bob Goodman"
```
joshblum: talked with @cecileboucheron, and we decided that this field should default to the contact name from the contact list, fallback to email if empty and otherwise fall back to phonenumber so that a user can know the mapping between a keybase user and what they called the user in their contact list

joshblum: also, if we fall back to phone or email we should use the `ContactComponent.Label` field in the formatting if `len(Contact.Components) > 0`(zeplin designs show an example)

protocol/avdl/keybase1/contacts.avdl
```diff
@@ -0,0 +1,44 @@
...
+        int contactIndex; // to associate the resolution to input contact entry
+        ContactComponent component;
+
+        // Outputs:
+        union { null, string } err;
+        boolean resolved;
+        // If resolved:
+        UID uid; // uid
+        string username; // keybase username
+        string fullName; // keybase full name
```
joshblum: i think we can drop this given how we want the `Name` field to work.

protocol/avdl/keybase1/contacts.avdl
```diff
@@ -0,0 +1,44 @@
...
+    // of (un)resolved contacts.
+    //
+    // If a contact finds a keybase resolution for any of its ContactComponents,
+    // one ResolvedContact is returned with the contact name, ContactComponent
+    // granting its resolution, and Keybase data.
+    //
+    // If a contact cannot be resolved for any of its component, each
+    // ContactComponent is transformed into ResolvedContact and returned. They are
+    // separate because each one can be used in the UI to allow for SBS chat.
+    array<ResolvedContact> lookupContactList(int sessionID, array<Contact> contacts, RegionCode userRegionCode);
```
joshblum: the output should be sorted according either according to `First, Last` or `Last, First` (if this preference is accessible via the contact import, it should be an input to the rpc)

protocol/avdl/keybase1/contacts.avdl
```diff
@@ -0,0 +1,44 @@
...
+    // ---------------------------
+
+    record ContactComponent {
+        // One of:
+        union { null, RawPhoneNumber } phoneNumber;
+        union { null, EmailAddress } email;
+        string label; // "", "Home", "Work"
+    }
+    record Contact {
+        string name; // "Bob Goodman"
```
joshblum: I think this should split to `first` and `last` fields if the native import has it separated that way so we can sort the output to match the user's normal view of their contact list

go/contacts/contacts.go
```diff
@@ -0,0 +1,151 @@
...
+		for i, k := range phoneRes {
+			if !k.Found {
+				continue
+			}
+
+			insertResult(k, phoneComps[i])
+		}
+	}
+
+	if len(emails) > 0 {
```
joshblum: given how the `Name` field will be calculated, emails should be inserted first

go/contacts/contacts.go
```diff
@@ -0,0 +1,183 @@
...
+	FillUsernames(libkb.MetaContext, []keybase1.ProcessedContact)
+}
+
+// ResolveContacts resolves contacts with cache for UI. See API documentation
+// in phone_numbers.avdl
+//
+// regionCode is optional, user region should be provided if it's known. It's
+// used when resolving local phone numbers, they are assumed to be local to the
+// user, so in the same region.
+func ResolveContacts(mctx libkb.MetaContext, provider ContactsProvider, contacts []keybase1.Contact,
```
heronhaye: can we put a CTraceTimed on this one and LookupEmails etc?

go/contacts/contacts.go
```diff
@@ -0,0 +1,179 @@
...
+		// And now that we have Keybase names and following information, make a
+		// decision about displayName and displayLabel.
+		for i, v := range res {
+			if !v.Resolved || v.Uid.IsNil() {
+				// Sanity check - should only have resolveds now.
+				return res, errors.New("found unresolved contact in display name processing")
+			}
+
+			res[i].DisplayName = v.Username
+			if v.Following && v.FullName != "" {
```
heronhaye: put this in a different function?
did this displaylabel strategy come from design?

go/contacts/lookup.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/keybase/client/go/phonenumbers"
+	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/client/go/uidmap"
+)
+
+type BulkLookupContactsProvider struct {
+}
+
+func (c *BulkLookupContactsProvider) LookupPhoneNumbers(mctx libkb.MetaContext, numbers []keybase1.RawPhoneNumber,
+	userRegion keybase1.RegionCode) (res []ContactLookupResult, err error) {
```
zapu: @heronhaye there already is a provider that uses real API. But the cache layer is missing.

go/contacts/contacts.go
```diff
@@ -0,0 +1,186 @@
...
+					contactIndex:   contactI,
+					componentIndex: compI,
+				})
+			}
+		}
+	}
+
+	mctx.Debug("Going to look up %d emails and %d phone numbers", len(emails), len(phoneNumbers))
+
+	// contactIndex -> true for all contacts that have at least one compoonent resolved.
```
joshblum:
```suggestion
	// contactIndex -> true for all contacts that have at least one component resolved.
```

go/contacts/contacts.go
```diff
@@ -0,0 +1,186 @@
...
+
+		res = append(res, keybase1.ProcessedContact{
+			ContactIndex: toContact.contactIndex,
+			ContactName:  contact.Name,
+			Component:    component,
+			Resolved:     true,
+			Uid:          lookupRes.UID,
+			Following:    true, // assume following=true for now because this creates better display label.
+
+			// following, username (TODO???), and full name are filled later.
```
joshblum: we shouldn't fill in `following` from the server right? the client should run identify on the uid?

go/contacts/contacts.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if len(emails) > 0 {
+		emailRes, err := provider.LookupEmails(mctx, emails)
+		if err != nil {
+			return res, err
+		}
+
+		for i, k := range emailRes {
+			if !k.Found {
```
joshblum:
```
if k.Found{
   insert(...)
}
```
 seems cleaner

go/contacts/contacts.go
zapu: yes, but I'm debating whether it should be client looking up its following list somehow or whether API sever should return this info in the `LookupBulk*` endpoints

zapu: oh, I don't think we need to ID here... this is all server trust on first use anyway

PR #17103, opened 2019-04-22
Add textContentType="emailAddress" to signup inputs

buoyad:
Doesn't do anything on simulator, waiting for xcode to update so I can test on device. r? @keybase/react-hackers 

- [x] Test on device before merge

### DIFF
--- a/shared/login/signup/common.js
+++ b/shared/login/signup/common.js
@@ -12,6 +12,7 @@ import {
   ButtonBar,
 } from '../../common-adapters'
 import {styleSheetCreate, isMobile, globalMargins, globalColors} from '../../styles'
+import flags from '../../util/feature-flags'
 
 type Props = {
   children: React.Node,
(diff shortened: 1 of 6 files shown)


buoyad:
So, it didn't work. Fixed by https://github.com/keybase/react-native/pull/17 and updated `yarn.lock` here.

shared/login/signup/common.js
```diff
@@ -29,7 +31,13 @@ export const Wrapper = (props: Props) => (
     >
       {props.children}
     </Box2>
-    <HeaderHocHeader onBack={props.onBack} headerStyle={styles.header} />
+    {flags.useNewRouter ? (
```
buoyad: `useNewRouter` wraps `HeaderHocHeader` with a `SafeAreaViewTop` which breaks `headerStyle`. Add with an extra box here to fix.

buoyad:
@keybase/react-hackers bump

PR #17109, opened 2019-04-22
paramproof logo_key display config, fuzzy filtering, fixed proof service selector

pzduniak:


shared/desktop/renderer/index.html.template
```diff
@@ -7,7 +7,7 @@
     object-src 'self';
     font-src 'self' <%= htmlWebpackPlugin.options.isDev ? 'http://localhost:4000' : '' %>;
     media-src http://127.0.0.1:*;
-    img-src 'self' data: http://127.0.0.1:* https://keybase.io/ https://pbs.twimg.com/ https://avatars.githubusercontent.com/ https://s3.amazonaws.com/keybase_processed_uploads/ <%= htmlWebpackPlugin.options.isDev ? 'http://localhost:4000' : '' %>;
+    img-src 'self' data: http://127.0.0.1:* http://localhost:* https://keybase.io/ https://pbs.twimg.com/ https://avatars.githubusercontent.com/ https://s3.amazonaws.com/keybase_processed_uploads/ <%= htmlWebpackPlugin.options.isDev ? 'http://localhost:4000' : '' %>;
```
mlsteele: Drop this change. @chrisnojima points out changing the CSP could be risky so we can https://github.com/keybase/client/pull/17106 instead

protocol/avdl/keybase1/user.avdl
```diff
@@ -160,6 +160,7 @@ protocol user {
   // A proof the user doesn't have.
   record ProofSuggestion {
     string key;
+    string logoKey; // Logo to use, "mastodon"
```
mlsteele: Gui doesn't need logoKey so it'd be nice to keep it out of the RPC interface. May work to move `.LogoKey` from `keybase1.ProofSuggestion` to `service.ProofSuggestion`

shared/desktop/renderer/index.html.template
pzduniak: oops, forgot about it

pzduniak: will remove after remaining issues are sorted out so that I can still test

go/libkb/services.go
```diff
@@ -140,6 +140,18 @@ func (t *BaseServiceType) GetAPIArgKey() string {
 
 func (t *BaseServiceType) IsDevelOnly() bool { return false }
 
+func (t *BaseServiceType) GetLogoKey() string {
+	t.Lock()
+	defer t.Unlock()
+	if t.displayConf == nil {
+		return ""
```
mlsteele: It'd be nice if this returned the service key `t.Key()` as a fallback. But that might be annoying so fine without

shared/desktop/renderer/index.html.template
```diff
@@ -22,4 +22,4 @@
         <% if (!htmlWebpackPlugin.options.isDev) { %><script>window.eval = global.eval = function () { throw new Error('Sorry, this app does not support window.eval().') }</script>
         <% } %>
         <%= htmlWebpackPlugin.files.js.filter(js => new RegExp(`\\W${htmlWebpackPlugin.options.name}\\W`).exec(js)).map(js => `<script src="${js}"></script>`).join('\n') %> </body>
-</html>
+</html>
```
mlsteele: don't need this change

go/libkb/services.go
pzduniak: yeaaah I don't know how plumbing that would work, let's keep it as is for now and change it for a more reasonable fallback if something ever breaks

go/identify3/adapter.go
```diff
@@ -264,6 +264,7 @@ func (i *UIAdapter) rowPartial(mctx libkb.MetaContext, proof keybase1.RemoteProo
 			}
 		}
 		row.ProofURL = i.makeSigchainViewURL(mctx, proof.SigID)
+		iconKey = serviceType.GetLogoKey()
```
xgess: this line is crashing my service locally when `serviceType` is nil. it's a test user using a service type that no longer exists in the configs because my local setup is a little janky. but it seems like this could also reappear in prod if we ever remove a config (which we definitely might could do).

mlsteele: Does https://github.com/keybase/client/pull/17169 fix that?

PR #17112, opened 2019-04-22
optimize disk attempt

joshblum:
attempt to optimize the msgpack encode/decode process. i'm not sure there is a practical difference


joshblum:
tested locally, merging over CI checks

PR #17114, opened 2019-04-22
implement Kb.SearchFilter

songgao:


songgao:
@keybase/react-hackers bump!

bump

shared/common-adapters/search-filter.js
```diff
@@ -0,0 +1,317 @@
...
+  Icon,
+  NewInput,
+  PlainInput,
+  ProgressIndicator,
+  Text,
+}
+
+type Props = {|
+  icon?: ?IconType,
+  focusOnMount?: ?boolean,
```
MarcoPolo: You don't have to make everything `foo?: ?bar`. That's for the specific case where: you can optionally supply foo, and if you do you can even pass in null/undefined.

I think what you want for most of these is: 
```foo?: bar```
You can optional supply `foo`, but if you do, then you **must** supply a `bar`

File: shared/common-adapters/search-filter.js [hunk 2]
```diff
@@ -0,0 +1,317 @@
...
+    this.setState({focused: false})
+    this.props.onBlur && this.props.onBlur()
+  }
+  _onFocus = () => {
+    this.setState({focused: true})
+  }
+
+  _focus = () => {
+    if (this.state.focused) {
+      // Can't just make a null for Kb.ClickableBox as that'd cause
```
MarcoPolo: I don't understand this comment.

shared/common-adapters/search-filter.js
```diff
@@ -0,0 +1,317 @@
...
+        </Kb.Text>
+      )
+    }
+    const iconColor = this.props.negative ? Styles.globalColors.white_75 : Styles.globalColors.black_50
+    const iconSizeType = this.props.type === 'Full-width' ? 'Default' : 'Small'
+    const hotkeyText =
+      this.props.hotkey && !this.props.onClick && !Styles.isMobile
+        ? ` (${Platforms.shortcutSymbol}+${this.props.hotkey.toUpperCase()})`
+        : ''
+    const content = (
```
MarcoPolo: It might be nice to split this up into helper functions since it's pretty big. Not react components, just functions.

shared/common-adapters/search-filter.js
```diff
@@ -0,0 +1,317 @@
...
+}
+
+type Props = {|
+  icon?: ?IconType,
+  focusOnMount?: ?boolean,
+  negative?: ?boolean,
+  onChange: (text: string) => void,
+  placeholderText: string,
+  style?: ?Styles.StylesCrossPlatform,
+  type: 'Small' | 'Full-width' | 'Mobile',
```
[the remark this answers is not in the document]
songgao: The "Mobile" is different from Small or FullWidth. If we add `fullWidth`, it'd be ignored by mobile, which I guess is fine. One reason I didn't do `fullWidth` was that with `Box2`, `fullWidth` means `width: 100%` but in this case it's actually `flexGrow`. But the "Full-width" came from design, so I thought I'd make it a bit different. Perhaps it should just be `big?: ?boolean`?

shared/common-adapters/search-filter.js
songgao: I was under the impression that it'd be useful in the situation where we have 

```
type PropsA = {
  foo?: ?bar,
}

type PropsB = {
  foo?: ?bar,
}
```

Then when component B renders a component A, it can just do `<A foo={props.foo />`. If it were `foo?: bar`, then it can't right?

shared/common-adapters/search-filter.js
songgao: When you pass in a `null` as `onClick` to `Kb.ClickableBox`, it becomes a different thing, causing the children to unmount and mount again, losing its focus state. So instead, always pass in `this._focus` as `onClick` but do nothing if we're already focused. This comment should probably be near that `onClick`?

shared/common-adapters/search-filter.js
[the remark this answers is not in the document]
MarcoPolo: You shouldn't be thinking about how it's implemented underneath. If it matches what you think of when you think "FullWidth" than I think it's fine to have a fullWidth flag.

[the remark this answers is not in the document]
songgao: The thing is when I think of fullWidth I think of `100%`. For example, if I have siblings to a Box2, I know I can't just give it a `fullWidth`. But this is different; it *can* have siblings.

shared/common-adapters/search-filter.js
MarcoPolo: I'm actually not sure about that, I'd have to try it out.

shared/common-adapters/search-filter.js
MarcoPolo: yeah that would be better. Maybe just say `keep a constant prop for onClick to avoid render thrashing`

PR #17115, opened 2019-04-22
prefetcher: fix some overall status bugs, and print overall status on CLI

strib:
We always need to pass the same request into the overall sync status checking functions, and that request has to already have its action set to be syncing.  We also need to make sure we count the block's own after we finish it processing it for prefetching.

So update the action first, and count the fetched bytes in a defer.

Also, fix a bug where `RootMetadata.TlfID()` could be called on a nil instance.

Finally, because I forgot to do so before, have simplefs populate the overall sync status, and have the CLI print it out.


go/kbfs/libkbfs/prefetcher.go
```diff
@@ -873,6 +882,305 @@ func (p *blockPrefetcher) handleStatusRequest(req *prefetchStatusRequest) {
 	}
 }
 
+func (p *blockPrefetcher) handlePrefetchRequest(req *prefetchRequest) {
```
strib: This function is copy-pasted from the original `run()` function, except where noted.

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -873,6 +882,305 @@ func (p *blockPrefetcher) handleStatusRequest(req *prefetchStatusRequest) {
...
+	// Ensure the block is in the right cache.
+	b, err := p.getBlockSynchronously(ctx, req, req.action.SoloAction())
+	if err != nil {
+		p.log.CWarningf(ctx, "error requesting for block %s: "+
+			"%+v", req.ptr.ID, err)
+		// There's nothing for us to do when there's an error.
+		return
+	}
+
+	// Update the priority and action of any existing
```
strib: This block of code, down to line 968, is new.

go/kbfs/simplefs/simplefs.go
```diff
@@ -2376,7 +2376,19 @@ func (k *SimpleFS) SimpleFSSyncConfigAndStatus(
...
+		p := k.config.BlockOps().Prefetcher().OverallSyncStatus()
+		res.OverallStatus.PrefetchProgress =
+			k.prefetchProgressFromByteStatus(p)
+		if p.SubtreeBytesTotal == p.SubtreeBytesFetched ||
+			p.SubtreeBytesTotal == 0 {
+			res.OverallStatus.PrefetchStatus = keybase1.PrefetchStatus_COMPLETE
+		} else {
+			res.OverallStatus.PrefetchStatus =
+				keybase1.PrefetchStatus_IN_PROGRESS
+		}
```
songgao: Hmm do we never have `keybase1.PrefetchStatus_NOT_STARTED`?

strib: I guess if all of the individual folders are NOT_STARTED, then the overall one should be not started.  I'll fix it up.

strib: (But as it is, if there are no TLFs configured, it will be NOT_STARTED since that's the default.)

strib:
Merging over unrelated gpg failure.

PR #17116, opened 2019-04-22
config: clean sync block cache on a restart

strib:
A previous sync disable may not have finished cleaning out all the blocks for that TLF before the process was killed, so make sure there aren't any leftover blocks from unsynced TLFs in the cache.

### DIFF
--- a/go/kbfs/libkbfs/disk_block_cache_wrapped.go
+++ b/go/kbfs/libkbfs/disk_block_cache_wrapped.go
@@ -521,6 +521,48 @@ func (cache *diskBlockCacheWrapped) GetTlfSize(
 	return size, nil
 }
 
+// GetTlfSize implements the DiskBlockCache interface for
+// diskBlockCacheWrapped.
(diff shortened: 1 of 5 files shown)


go/kbfs/libkbfs/disk_block_cache_wrapped.go
```diff
@@ -521,6 +521,72 @@ func (cache *diskBlockCacheWrapped) GetTlfSize(
...
 
+// GetTlfSize implements the DiskBlockCache interface for
+// diskBlockCacheWrapped.
+func (cache *diskBlockCacheWrapped) GetTlfIDs(
+	ctx context.Context, cacheType DiskBlockCacheType) (
+	tlfIDs []tlf.ID, err error) {
+	cache.mtx.RLock()
+	defer cache.mtx.RUnlock()
+
+	if cacheType != DiskBlockWorkingSetCache {
```
songgao: Might use a comment here to mention it's either `DiskBlockSyncCache` or `DiskBlockAnyCache` here? same below.

PR #17118, opened 2019-04-23
prefetcher: send overall sync status notifications

strib:
Based on #17115, review that one first.

Send once-a-second overall sync status notifications to the service, which routes them to the GUI.

Issue: KBFS-4082


go/kbfs/libkbfs/prefetcher.go
```diff
@@ -208,6 +212,34 @@ func newBlockPrefetcher(retriever BlockRetriever,
...
+	var status keybase1.FolderSyncStatus
+	status.PrefetchProgress = p.overallSyncStatus.ToProtocolProgress(
+		p.config.Clock())
+	if p.overallSyncStatus.SubtreeBytesTotal ==
+		p.overallSyncStatus.SubtreeBytesFetched ||
+		p.overallSyncStatus.SubtreeBytesTotal == 0 {
+		status.PrefetchStatus = keybase1.PrefetchStatus_COMPLETE
+	} else {
+		status.PrefetchStatus = keybase1.PrefetchStatus_IN_PROGRESS
+	}
```
songgao: Would it make sense to have a helper function that covers these bytes comparison too, since it's used in [simplefs.go](https://github.com/keybase/client/pull/17118/files#diff-b9b12d1b48b0bfd42d3ebb71a8bf0ffeR2365) as well?

strib: Done!

go/kbfs/libkbfs/reporter_kbpki.go
```diff
@@ -271,12 +273,33 @@ func (r *ReporterKBPKI) NotifySyncStatus(ctx context.Context,
...
+func (r *ReporterKBPKI) NotifyOverallSyncStatus(
+	ctx context.Context, status keybase1.FolderSyncStatus) {
+	select {
+	case r.notifyOverallSyncBuffer <- status:
+	default:
+		// If this represents a "complete" status, we can't drop it.
+		// Instead launch a goroutine to make sure it gets sent
+		// eventually.
+		if status.PrefetchStatus == keybase1.PrefetchStatus_COMPLETE {
+			go func() { r.notifyOverallSyncBuffer <- status }()
```
jzila: Should we have a shutdown channel here? I'm worried about these goroutines leaking, or panicking racily due to a write to a closed channel when `Shutdown` is called.

Alternatively, what if this buffer was an `InfiniteChannel`? Then it'd never block on write, and we wouldn't have to use this goroutine approach.

strib: For what it's worth, this reporter type is not used in tests, only in production, and so I don't think `Shutdown()` will ever be called in a situation where there could be outstanding requests.

But sure, I'll add a shutdown channel just to make it a little cleaner.

jzila: What about the infinite channel idea?

strib: For one, it still has the close race issue.  For another, this particular channel purposely has a size of 1 to avoid unnecessary "incomplete" messages from being sent.  So it would be hard to turn it into an infinite channel while keeping that nice property.  For another another, it's annoying to do the casting away from an `interface{}` with an infinite channel.

jzila: I see, so you're using the blocking of the channel to filter out incompletes. We could throttle that with a timer instead, but I see the appeal of the current approach.

strib:
Ugh a lot of unrelated failures.  Merging over them.

PR #17120, opened 2019-04-23
implement folder view filter as designed

songgao:


shared/fs/common/path-item-action/choose-view.js
```diff
@@ -10,8 +10,8 @@ import Confirm from './confirm-container'
 
 type OwnProps = {|
   floatingMenuProps: FloatingMenuProps,
+  mode: 'row' | 'screen',
   path: Types.Path,
-  position?: 'row' | 'header',
```
songgao: @jakob223 I ended up changing the name into `mode` as well, as `mode` is also used in `path-item-info` and I thought this'd be more consistent.

shared/fs/common/path-item-action/layout.js
```diff
@@ -77,8 +102,8 @@ const getRawLayout = (path: Types.Path, pathItem: Types.PathItem, me: string): L
         showInSystemFileManager: !isMobile,
         // share menu items
         // eslint-disable-next-line sort-keys
-        sendAttachmentToChat: flags.sendAttachmentToChat && isMobile && pathItem.type === 'file', // desktop uses separate button
-        sendLinkToChat: isMobile && Constants.canSendLinkToChat(parsedPath), // desktop uses separate button
+        sendAttachmentToChat: flags.sendAttachmentToChat && pathItem.type === 'file', // desktop uses separate button
```
jakob223: Is removing `isMobile` accidental, or is the comment now a lie?

shared/fs/common/path-item-action/menu.js
```diff
@@ -57,6 +59,30 @@ const hideMenuOnClick = (onClick: (evt?: SyntheticEvent<>) => void, hideMenu: ()
...
+          onClick: hideMenuOnClick(props.newFolder, hideMenu),
+          title: 'New folder',
+        },
+      ]
+    : []),
+  ...(props.openChatTeam
+    ? [
+        {
+          onClick: hideMenuOnClick(props.openChatTeam, hideMenu),
+          title: 'Chat with users in this team',
```
jakob223: I feel like "chat with the users in this team" more accurately describes this action to me and I'd prefer to add the word "the" there unless this is particularly space constrained.

shared/fs/common/path-item-action/menu.js
```diff
@@ -57,6 +59,30 @@ const hideMenuOnClick = (onClick: (evt?: SyntheticEvent<>) => void, hideMenu: ()
...
+        },
+      ]
+    : []),
+  ...(props.openChatNonTeam
+    ? [
+        {
+          onClick: hideMenuOnClick(props.openChatNonTeam, hideMenu),
+          title: 'Chat with users in this folder',
```
jakob223: ditto with "the users" here.

shared/fs/common/upload-button.js
```diff
@@ -21,15 +21,27 @@ const UploadButton = Kb.OverlayParentHOC(props => {
     return <Kb.Button small={true} onClick={props.openAndUpload('both')} label="Upload" style={props.style} />
   }
   if (isIOS) {
-    return <Kb.Icon type="iconfont-new" padding="tiny" onClick={props.pickAndUpload('mixed')} />
+    return (
+      <Kb.Icon
+        type="iconfont-upload"
```
jakob223: This icon super jumps out at me here - maybe because I'm used to only having this as an in-progress icon, or maybe because it's big and blue - but it feels like it's more important than the rest of the screen. Can it be black instead maybe? Or did this come from design and I should comment somewhere in zeplin along these lines?

shared/fs/nav-header/mobile-header.js
```diff
@@ -15,26 +16,44 @@ type Props = {|
...
+        direction="vertical"
+        fullWidth={true}
+        style={Styles.collapseStyles([styles.container, this.props.bannerType === 'offline' && styles.blue])}
+      >
+        {this.props.bannerType === 'offline' && <Kb.Banner text="You are offline." color="blue" />}
+        {this.state.filterExpanded ? (
+          <Kbfs.FolderViewFilter path={this.props.path} onBlur={this._filterDone} />
+        ) : (
+          <Kb.Box2 direction="horizontal" fullWidth={true} style={styles.expandedTopContainer}>
+            <Kb.BackButton badgeNumber={0 /* TODO */} onClick={this.props.onBack} />
```
jakob223: What's the badge supposed to contain? do you plan to do it in this PR?

shared/fs/common/path-item-action/layout.js
songgao: Comment is now a lie and I got rid of them in https://github.com/keybase/client/pull/17164

shared/fs/common/upload-button.js
songgao: Not on zeplin but this indeed came from design; @cecileboucheron wdyt? Should it remain blue?

shared/fs/nav-header/mobile-header.js
songgao: This is actually from previous commit so not planning to address it in this PR. Not sure what it should be yet. Perhaps notifications from previous screen?

songgao:
Thanks @jakob223 ! I can't repro 1; hitting "Enter" just does nothing for me. 2) is a good catch! It's even worse when you filter in a TLF list and hit back, where it just shows nothing and no filter bar ...

shared/fs/nav-header/mobile-header.js
jakob223: Do back buttons normally have badges on them? I feel like the TODO in the code is useless without any sort of explanation, so if nobody knows what it was even for we should just delete the `badgeNumber` from the `BackButton`.

shared/fs/common/path-item-action/menu.js
songgao: I tried adding "the" but it overflew into two lines. So not sure. Would you mind if I merge without the added "the", and discuss with #design?

jakob223:
I experienced #1 in iOS simulator, iPhone 5s, 12.0, if that helps.

shared/fs/nav-header/mobile-header.js
songgao: It's in design actually: https://zpl.io/amygjBK , so perhaps we'll still need it in the future. I'll make a ticket for it.

jakob223: Can you stick the ticket number in the TODO?

songgao:
Oh! Didn't realize it was in sim. Will try to fix

shared/fs/common/index.stories.js
```diff
@@ -163,24 +168,35 @@ const load = () => {
         gapStart={true}
         style={{paddingLeft: Styles.globalMargins.medium}}
       >
+        <Kb.Text type="Body">Row mode</Kb.Text>
+        <PathItemAction
+          path={Types.stringToPath('/keybase/private/meatball/folder/treat')}
+          routePath={I.List()}
+          {...pathItemActionCommonProps}
```
jzila: This should come after the `mode` row. Same below for all the `PathItemAction`s

shared/constants/fs.js
```diff
@@ -1134,6 +1134,9 @@ export const makeActionsForShowSendAttachmentToChat = (
 export const splitFileNameAndExtension = (fileName: string) =>
   ((str, idx) => [str.slice(0, idx), str.slice(idx)])(fileName, fileName.lastIndexOf('.'))
 
+export const isFolder = (path: Types.Path, pathItem: Types.PathItem) =>
+  Types.getPathLevel(path) <= 3 || pathItem.type === 'folder'
```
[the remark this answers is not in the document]
songgao: It's about both. If pathLevel <=3, we know for sure it's a folder. Otherwise we need to look at the pathItem.

shared/fs/common/folder-view-filter-icon.js
```diff
@@ -0,0 +1,53 @@
...
+import * as Styles from '../../styles'
+import * as React from 'react'
+import flags from '../../util/feature-flags'
+
+type Props = {|
+  onClick: () => void,
+  path: Types.Path,
+  pathItem: Types.PathItem,
+  style?: ?Styles.StylesCrossPlatform,
+|}
```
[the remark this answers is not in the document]
songgao: I went the other way with `$Diff` because the container is provide these props so I felt that's clearer.

shared/constants/fs.js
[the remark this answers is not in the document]
jzila: But why don't we _only_ look at the pathItem?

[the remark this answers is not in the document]
songgao: Because we usually don't have a pathItem for TLF root before going into it, and we never have a pathItem for TLF list (e.g. /keybase/team)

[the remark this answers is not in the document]
jzila: We don't? We used to...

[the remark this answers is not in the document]
songgao: Yeah we removed that in one of the refactors I think. Outside a TLF we don't have the same set of metadata (e.g. lastModifiedTime, lastWriter, etc.) so it doesn't feel proper to have the same data structure back the data.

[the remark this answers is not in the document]
jzila: But now we're forced to do this weird hybrid thing anyway...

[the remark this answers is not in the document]
songgao: I think it's fine as it properly represents the underlying data. Note that TLF roots also have their own metadata, and not requiring pathItem for them makes it possible to delay constructing a FBO for each when user is in e.g. /keybase/private

PR #17122, opened 2019-04-23
Go1.12 symlink detection vs Dokan issue, pass test for now

taruti:
Go symlink detection was changed again, and now it misidentifies things on Dokan. 
Fix properly later.


taruti:
gofmt makes jenkins fails.

jzila:
Go ahead and merge without CI.

taruti:
I pushed it to https://github.com/keybase/client/pull/17073 so it can be merged together with the other changes.

PR #17126, opened 2019-04-23
prefetcher: fix dumb prefetch cancel bug, add some more canceling

strib:
A dumb typo was preventing prefetches from being properly disabled.

Also cancel any ongoing edit-history based prefetches when syncing is enabled, and check for context cancellation right away when processing prefetch requests.

### DIFF
--- a/go/kbfs/libkbfs/config_local.go
+++ b/go/kbfs/libkbfs/config_local.go
@@ -1567,7 +1567,7 @@ func (c *ConfigLocal) setTlfSyncState(tlfID tlf.ID, config FolderSyncConfig) (
 func (c *ConfigLocal) SetTlfSyncState(
 	ctx context.Context, tlfID tlf.ID, config FolderSyncConfig) (
 	<-chan error, error) {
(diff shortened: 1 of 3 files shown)


strib:
> I don't think I understand why we're cancelling those prefetches whenever there might be a sync prefetch scheduled for those same blocks.

@jakob223 if syncing is being enabled with this `SetSyncConfig` call, it won't get started until the end of the function, after the working set prefetches are canceled (and we wait for them to complete).  If syncing was already enabled, then they shouldn't be happening yet.  I guess we don't strictly have to cancel them since the any full or partial sync should just override them and move the blocks into the right cache, but it makes the logs messier and probably wastes CPU to have them going on in the background for no reason.

> you'll have a conflict with your PR to move out handlePrefetchRequest

@jzila yeah known, I'll deal with that when rebasing the other PR, no problem.

I also just added a small test flake fix to this PR, so that we only cancel the prefetches when not in test mode.  Otherwise it messes with very specific ordering requirements on some prefetch-related tests.

PR #17129, opened 2019-04-23
fix gregor client stop

joshblum:


joshblum:
merging over bad CI since changes were vendored in kbweb

PR #17132, opened 2019-04-23
prefetcher: check for and handle status requests constantly

strib:
`keybase fs show` depends on being able to request the prefetch status from the prefetcher, which in turn requires sending a message on a channel and waiting for it to be processed by the main processing goroutine.  But in some cases that goroutine can be busy for tens of seconds, e.g. when it is processing a block with 1000s of child pointers, each one involving a disk cache operation.

So this just processes any pending status requests (which are quick) inline during every iteration of that child processing, to guarantee a quick return.  This is safe, since the status request is read-only and isn't interfering with the prefetching logic at all.

### DIFF
--- a/go/kbfs/libkbfs/prefetcher.go
+++ b/go/kbfs/libkbfs/prefetcher.go
@@ -644,6 +644,33 @@ func (p *blockPrefetcher) request(ctx context.Context, priority int,
 	return 0, 0, 0
 }
 
+func (p *blockPrefetcher) handleStatusRequest(req *prefetchStatusRequest) {
+	pre, isPrefetchWaiting := p.prefetches[req.ptr.ID]
+	if !isPrefetchWaiting {
+		req.ch <- PrefetchProgress{}
+	} else {
+		req.ch <- pre.PrefetchProgress
+	}
+}
+
+// handleCriticalRequests should be called periodically during any
+// long prefetch requests, to make sure we handle critical requests
+// quickly.  These are requests that are required to be run in the
+// main processing goroutine, but shouldn'won't interfere with
+// whatever request we're in the middle of.
+func (p *blockPrefetcher) handleCriticalRequests() {
+	for {
+		// Fulfill any status requests since the user could be waiting
+		// for them.
+		select {
+		case req := <-p.prefetchStatusCh.Out():
+			p.handleStatusRequest(req.(*prefetchStatusRequest))
+		default:
+			return
+		}
+	}
+}
+
 func (p *blockPrefetcher) prefetchIndirectFileBlock(
 	ctx context.Context, parentPtr data.BlockPointer, b *data.FileBlock,
 	kmd libkey.KeyMetadata, lifetime data.BlockCacheLifetime, isPrefetchNew bool,
@@ -659,6 +686,8 @@ func (p *blockPrefetcher) prefetchIndirectFileBlock(
 		numBlocks += b
 		numBytesFetched += f
 		numBytesTotal += t
+
+		p.handleCriticalRequests()
 	}
 	return numBlocks, numBytesFetched, numBytesTotal, len(b.IPtrs) == 0
 }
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/prefetcher.go
```diff
@@ -644,6 +644,33 @@ func (p *blockPrefetcher) request(ctx context.Context, priority int,
...
+		req.ch <- PrefetchProgress{}
+	} else {
+		req.ch <- pre.PrefetchProgress
+	}
+}
+
+// handleCriticalRequests should be called periodically during any
+// long prefetch requests, to make sure we handle critical requests
+// quickly.  These are requests that are required to be run in the
+// main processing goroutine, but shouldn'won't interfere with
```
jzila: shouldn'won't is a new one. Is that 85% or 95% certainty?

go/kbfs/libkbfs/prefetcher.go
```diff
@@ -719,6 +750,8 @@ func (p *blockPrefetcher) prefetchDirectDirBlock(
 		numBlocks += b
 		numBytesFetched += f
 		numBytesTotal += t
+
+		p.handleCriticalRequests()
```
jzila: Why are these all in the individual handlers instead of just before the return in `handlePrefetch`, or just after `handlePrefetch` is called in the `run` loop?

go/kbfs/libkbfs/prefetcher.go
strib: Oops will fix, thanks.

go/kbfs/libkbfs/prefetcher.go
strib: `handlePrefetch()` isn't called in a loop, it's just called once per request.  Do you mean `request()`?  I guess I just liked being more explicit about it, and `request()` didn't seem like a natural place for it; I guess I was worried that `request()` was too generic and could be reasonably called from elsewhere in the code.   If you want I'm happy to change it though.

jzila: `prefetch{Indirect,Direct}{Dir,File}Block()` are called in `handlePrefetch()`, which is in turn called in the `run()` loop. If you put a single `handleCriticalRequests` call in either of `handlePrefetch()` or `run()` after the call to `handlePrefetch`, you can replace these three `handleCriticalRequests()` calls with one.

strib: The point of this is that it's called inside the `for` loop inside those `prefetch{Indirect,Direct}{Dir,File}Block()` calls.  It can't just be a single call in `run()` (in fact, there already is one there).

jzila: Oh I see. What's the benefit here though? The calls inside the for loop are all async, we should be able to preempt them with a high-priority critical request.

strib: The call isn't completely async -- it does a `checkCaches` call inline.  Add 2000 of those up, and you're talking about 10s of seconds inside that for loop.

strib: (because of the leveldb calls)

jzila: Hmm maybe we should eliminate the disk cache check in `checkCaches`, allowing the primary queue in `block_retrieval_queue` to do that as part of the call out to `bserver_remote`.

strib: Maybe but that seems like a larger discussion as it would slow down normal block accesses a bit by forcing them to go through the normal queue when they're already cached.

jzila: OK I agree it's a larger discussion, but it seems like the right call. I don't love this critical request handling in the depths of a prefetch request being handled. The current design necessitates it, which is a strong signal for me that we should change the current design.

strib: Sounds good, feel free to open a ticket.  We could also just add a signal to `blockRetrievalQueue.Request()` to skip the disk cache check just for this given request, and use that in `prefetcher.request()`.

PR #17133, opened 2019-04-23
Add retries to LookupUser calls

pzduniak:
I think we should just add a bunch of "quick" retry params on all GET requests but eh, this one should be good for now. I followed the local user lookup flow and it seems that this is the only significant request that can fail.

### DIFF
--- a/go/libkb/merkle_client.go
+++ b/go/libkb/merkle_client.go
@@ -738,10 +738,13 @@ func (mc *MerkleClient) lookupPathAndSkipSequenceHelper(m MetaContext, q HTTPArg
 	}
 
 	apiRes, err = m.G().API.Get(m, APIArg{
-		Endpoint:       "merkle/path",
-		SessionType:    APISessionTypeNONE,
-		Args:           q,
-		AppStatusCodes: []int{SCOk, SCNotFound, SCDeleted},
+		Endpoint:        "merkle/path",
+		SessionType:     APISessionTypeNONE,
+		Args:            q,
+		AppStatusCodes:  []int{SCOk, SCNotFound, SCDeleted},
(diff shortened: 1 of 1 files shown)


patrickxb:
When did GET merkle/path fail?

go/libkb/merkle_client.go
```diff
@@ -738,10 +738,13 @@ func (mc *MerkleClient) lookupPathAndSkipSequenceHelper(m MetaContext, q HTTPArg
...
-		SessionType:    APISessionTypeNONE,
-		Args:           q,
-		AppStatusCodes: []int{SCOk, SCNotFound, SCDeleted},
+		Endpoint:        "merkle/path",
+		SessionType:     APISessionTypeNONE,
+		Args:            q,
+		AppStatusCodes:  []int{SCOk, SCNotFound, SCDeleted},
+		RetryCount:      3,
+		InitialTimeout:  4 * time.Second,
+		RetryMultiplier: 1.1,
```
mlsteele: what's multiplier for?

patrickxb: it changes the timeouts for subsequent requests.

mlsteele: yeah I just wonder since this is the only non-default one in the app why that's desirable

mlsteele:
Here's where it failed https://keybase.atlassian.net/browse/CORE-10601

go/libkb/merkle_client.go
pzduniak: I took it from "home"... whatever that is. No multipler is OK too, we mostly care about retry in sensitive situations here (for example what happened in the UI). This is most likely what we would have to do if we were to add support for poor network conditions.

PR #17135, opened 2019-04-23
Reposition open meta tag on teams list.

adamjspooner:


shared/teams/main/index.js
```diff
@@ -184,6 +186,7 @@ const styles = Styles.styleSheetCreate({
   },
   kerning: {letterSpacing: 0.2},
   maxWidth: {maxWidth: '100%'},
+  openMeta: {alignSelf: 'center'},
```
songgao: I wonder if it'd work to just add `alignItems="center"` in the containing Box2.

adamjspooner: It doesn't because the Meta component has `alignSelf: 'flex-start'`. So, this is needed to override that.

PR #17138, opened 2019-04-24
Very small perf tweak to help box2

MarcoPolo:
@keybase/react-hackers 

Small tweak to `intersperseFn` which is used by Box2 when you ask for gaps.

On my machine it makes box2 with 100 children 10% faster. In practice it probably won't be so dramatic since I don't think we use box2 with a lot of children and gap. When the children count is small it's basically the same thing as the reduce version.

It does have fewer allocations which could help with GC /shrug

### DIFF
--- a/shared/util/arrays.js
+++ b/shared/util/arrays.js
@@ -6,13 +6,13 @@ export function intersperseFn<A, B>(
   arr: Array<A>
 ): Array<A | B> {
   if (arr.length === 0) {
-    return []
+    // $FlowIssue an array with no length is a valid type for any Array
+    return arr
   }
 
-  return arr.slice(1).reduce(
-    (acc, x, i, a) => {
-      return acc.concat([separatorFn(i, x, a), x])
-    },
-    [arr[0]]
-  )
+  let toReturn = [arr[0]]
+  for (let i = 1; i < arr.length; i++) {
+    toReturn.push(separatorFn(i, arr[i], arr), arr[i])
+  }
+  return toReturn
 }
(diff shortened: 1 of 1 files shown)


shared/util/arrays.js
```diff
@@ -6,13 +6,13 @@ export function intersperseFn<A, B>(
   arr: Array<A>
 ): Array<A | B> {
   if (arr.length === 0) {
-    return []
+    // $FlowIssue an array with no length is a valid type for any Array
+    return arr
   }
```
chrisnojima: Could try and make a local module const that's returned (this is fine also. Making a new array each time likely bad)

shared/util/arrays.js
```diff
@@ -6,13 +6,13 @@ export function intersperseFn<A, B>(
...
 
-  return arr.slice(1).reduce(
-    (acc, x, i, a) => {
-      return acc.concat([separatorFn(i, x, a), x])
-    },
-    [arr[0]]
-  )
+  let toReturn = [arr[0]]
```
chrisnojima: Not that I think we lose a lot of perf here ever but likely faster
toReturn = new Array(arr. length)
For...
toReturn[i] = ...

MarcoPolo: I forgot you could size hint the array!

buoyad:
@MarcoPolo the `letterSpacing` in storyshots was fixed in master

MarcoPolo:
Proof of correctness: https://runkit.com/marcopolo/5cbfdc169fa32800123da8fe

PR #17130, opened 2019-04-23
Add app action icons.

adamjspooner:
@cjb, this is ready for review.

PR #17057, opened 2019-04-18
hide menubar on linux

heronhaye:

PR #17140, opened 2019-04-24
Frameless Windows and Linux

cjb:
@keybase/react-hackers r?

This PR combines a few things:

* @adamjspooner's #17130, which adds minimize/maximize/close window icons

* @heronhaye's #17057, which hides the menubar on Linux, expanded to hide it on Windows too

* `frame: false` turned on for Windows and Linux

* Hook up the minimize/maximize/close icons using the `BrowserWindow` API fns.

.. with the end result that macOS is unchanged, and Windows and Linux both use the same hand-drawn nav2 window management icons, and are both frameless and menubarless.


heronhaye:
Looks good on i3.

On xfce4 i'm not able to drag the window by the titlebar so it's kind of stuck (I can't drag it on i3 either but I can use the wm's native movement controls there). And I also can't resize it by dragging on the border, it doesn't do anything.

Also, the minimize and maximize buttons don't do anything on either i3 or xfce4. I'm not sure maximize and minimize really make sense on Linux across different wms (the buttons in the Window menu never did anything either). But I think at least the maximize button should maximize on xfce4. Probably an electron issue though?

This is more of a design question but would it be better to have those controls on the left like in Mac so it's more consistent with the Mac app?

Finally, in my PR I can press "Alt" to access the file menu but I can't do that in this branch for some reason, but I don't see any reason why that should be the case from the code.

Let me know if you need me to test anything else, excited about this change!

cc @zapu

zapu:
The "active zone" of the buttons should be larger, also a button background colour change on hover would be more natural for Windows users.

I recorded a video in `/keybase/team/keybase/cjb_titlebar.mp4`.

adamjspooner:
I agree with @zapu's feedback. Let's mimic the hit target sizes and hover colors (black over gray and white over red) of other Windows apps.

zapu:
Also on Windows you can't resize using top corner anymore, which is rather unusual for a Windows application. Maybe it's similar on some Linux desktops as well.

cjb:
@adamjspooner Thanks, how's this?

winapp

adamjspooner:
Looking good! Can we move them (or make them larger) so the top edge is aligned with the top edge of the app? Can we move them closer together so there's no gap between them? Also, the close hover should be white on a red background. Take a look at @zapu's video for more details. Let's get as close to that as possible.

cjb:
Cool, I'll try.  Weird, `<Kb.Icon style={{color: grey, hoverColor: white}}>` isn't changing the color to white on hover.  Do you want globalColors.red or globalColors.red_20 or something else?  The full red seems aggressive to me but I guess Windows just looks that way.

cjb:
Ready for another look, how's this? (That's white on red, just the GIF palette being too small.)

newapp

adamjspooner:
Looks great! @cecileboucheron, thoughts? Do we want to push it all the way to the right (so the red box is fully in the upper right corner)?

zapu:
This is getting better with every iteration!

Some thoughts:
Video for reference from Windows: `/keybase/team/keybase/cjb_windows_2.mp4`

- The buttons have to be pushed all the way to the right, at least on Windows - that's what the native apps (without custom titlebars) do. Custom titlebars emulate this as well. See VSCode in the reference video.
- Something broke and minimize button maximizes now. Not sure if that was different when I tried this branch last time, maybe it was always wrong.
- The maximize button icon changes when window is maximized - see VSCode icon in reference video.
- The "hot zones" should still be larger - they should be done in a way where there is no space between buttons. See reference video - I can drag window by dragging titlebar from between two buttons. For reference, in VSCode (or any other app), you can't do it. You can miss the button "visually" by a fairly large margin and still get a button click.
- Window resizing is not correct. Somehow I can resize the window using top corner but only in the area where the buttons are. See the recorded video. Windows and (most of) Linux uses are used to being able to drag and resize using any corner of the window.

zapu:
I've added another video that shows the resizing issue: `/keybase/team/keybase/cjb_window_resizing.mp4`

zapu:
Also there is an issue with "Search" bar as well as icons/buttons on the top of conversation (search / files / info). You can't click these anymore because they act as a bar that drags the entire window. Same on Linux XFCE and Windows 10. I have a recording here: `/home/zapu/Videos/cjb_panel_drag.mp4`

cecileboucheron:
I trust you all on design decisions :)

cjb:
@zapu Thanks! Another look?

* Pushed the buttons over to the edge

* Fixed minimize button

* I can use a different icon when maximized if @adamjspooner wants to export another one (can push it directly to this branch), no strong opinion from me.

* Made hot zones larger, should be no gaps now.

* Inconsistent resizing is probably an Electron bug/behavior, don't see any fix

* The issues with unclickable buttons on headers are unrelated, we've got tickets for them, a style needs to be applied for anything in the nav area that ought to be clickable.

So I think this should be ready to go unless we want to change the maximized icon.

pzduniak:
2 things on Gnome 2.28 on Ubuntu 18.04:

1. Buttons are not offset to the right here (although it's not as much of an issue here I guess)
2. Login title bar is absolutely messed up and offset by like 10px to the bottom:

image

cjb:
@pzduniak What is "Login title bar" and where is it in the screenshot?

pzduniak:
Title bar is the thing we're removing, so by "login title bar" I meant the 3 window buttons and the back button. They're moved to the bottom quite significantly, you can't grab anything above them to move the window.

pzduniak:
Maybe that helps?
image

cjb:
@pzduniak I think `git pull origin cjb/nav2-bar-desktop-logged-out-routes` on top of this PR branch should fix the logged out view, want to give it a try?

pzduniak:
image

Works!

heronhaye:
checked out latest, some more notes. thanks again for working on this!

- need https://keybase.atlassian.net/browse/DESKTOP-9492 before release
- I can't Ctrl + or - anymore to zoom

minor issues that would be nice to fix but not necessary or that important
- resizing on top edge of window. right now you can only do this from above the control icons - I think it should be the opposite and that part of the top bar next to the control icons should be for the control, so you can move your mouse all the way to the top right of your screen and select X instead of resize icon
- the color of the icons themselves changes from light to dark grey when hovering over, but the hitbox doesn't match the size of the background box. To repro: put your cursor on the very edge of the maximize box such that the background box is highlighted. then move your cursor a bit closer to the center and the square maximize icon gets darker. ideally, the hitboxes would be the same or the  
- when I from floating mode to windowed mode and back, it loses the original shape of the window and defaults to right half of my screen. in discord, it remembers the original floating size and position.

cjb:
Thanks! Agree those don't sound like blockers right now, so I'll merge these in and work on 9492.  The extra icon color change is because the icon component itself changes its hoverCover when it's hovered.

PR #17142, opened 2019-04-24
Change phone number notifications

zapu:
Do one UI notification when phone numbers change instead of 3

this is a PR into CJB's PR.


zapu:
@cjb do you think we can merge this some time after release? I added additional notifications for phone and email events.

PR #17146, opened 2019-04-24
Wallets nav2 header

buoyad:
- Factor `SendButton` and `DropdownButton` from the original header into wallets common components

image

r? @keybase/react-hackers


shared/constants/wallets.js
```diff
@@ -532,6 +532,9 @@ export const getAccounts = (state: TypedState) => state.wallets.accountMap.value
 
 export const getSelectedAccount = (state: TypedState) => state.wallets.selectedAccount
 
+export const getSelectedAccountData = (state: TypedState) =>
+  state.wallets.accountMap.get(getSelectedAccount(state), makeAccount())
```
chrisnojima: should use unknownAccount so its not a new object every time

shared/wallets/common/buttons/index.js
```diff
@@ -0,0 +1,137 @@
...
+  onSendToKeybaseUser: () => void,
+  onSendToStellarAddress: () => void,
+  onSendToAnotherAccount: () => void,
+  small?: boolean,
+|}
+
+class _SendButton extends React.PureComponent<SendProps & Kb.OverlayParentProps> {
+  _menuItems = [
+    {
+      onClick: () => this.props.onSendToKeybaseUser(),
```
chrisnojima: don't need new closures here, can just be `this.props.onSendToKeybaseUser` same below

shared/wallets/common/buttons/index.js
```diff
@@ -0,0 +1,137 @@
...
+  small?: boolean,
+|}
+
+class _DropdownButton extends React.PureComponent<DropdownProps & Kb.OverlayParentProps> {
+  render() {
+    const onShowSecretKey = this.props.onShowSecretKey
+    const _menuItems = [
+      onShowSecretKey
+        ? {
+            onClick: () => onShowSecretKey(),
```
chrisnojima: same here

shared/wallets/common/buttons/index.js
```diff
@@ -0,0 +1,137 @@
...
+          </Kb.Button>
+        </Kb.Box2>
+        <Kb.FloatingMenu
+          attachTo={this.props.getAttachmentRef}
+          closeOnSelect={true}
+          items={_menuItems}
+          onHidden={this.props.toggleShowingMenu}
+          visible={this.props.showingMenu}
+          position="bottom center"
+          positionFallbacks={[]}
```
chrisnojima: we should allow this to be null so it doesn't thrash every time (can ticket)

shared/wallets/nav-header/add-account.js
```diff
@@ -0,0 +1,64 @@
...
+    dispatch(
+      RouteTreeGen.createNavigateAppend({path: [{props: {showOnCreation: true}, selected: 'linkExisting'}]})
+    )
+  },
+})
+
+export default Container.namedConnect<{||}, _, _, _, _>(
+  () => ({}),
+  mapDispatchToProps,
+  (_1, d, _2) => d,
```
chrisnojima: can just be `(_, d)`

shared/wallets/nav-header/container.js
```diff
@@ -0,0 +1,48 @@
+// @flow
+import * as Container from '../../util/container'
+import * as Constants from '../../constants/wallets'
+import * as Types from '../../constants/types/wallets'
+import * as RouteTreeGen from '../../actions/route-tree-gen'
+import {HeaderTitle as _HeaderTitle, HeaderRightActions as _HeaderRightActions} from '.'
+
+const mapStateToPropsHeaderTitle = state => ({
+  _account: Constants.getSelectedAccountData(state),
+  username: state.config.username || '',
```
chrisnojima: username is always a string

shared/wallets/common/buttons/index.js
buoyad: It's allowed to be undefined

shared/constants/wallets.js
buoyad: ah forgot about that, I checked for `emptyAccount`

buoyad:
Thanks @chrisnojima, made changes PTAL

shared/wallets/common/buttons/index.js
buoyad: Made desktop-9481

PR #17149, opened 2019-04-24
TLF sync config change notifications from kbfs to GUI

taruti:


taruti:
gofmt issues in proto for 1.12...

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -8852,6 +8852,10 @@ func (fbo *folderBranchOps) SetSyncConfig(
 		fbo.log.CDebugf(ctx, "Starting full deep sync")
 		_ = fbo.kickOffRootBlockFetch(ctx, md)
 	}
+
+	fbo.config.Reporter().Notify(ctx, syncConfigChangeNotification(
```
strib: Nothing above actually checks and returns early if the new config is the same as the old config, so this could send unnecessary notifications.  I don't care much either way, so if that matters for the GUI we should maybe add a check like that above.

strib:
Looks good in Go-land other than my comment.  Not sure about the JS line so a react-hacker should look.

go/kbfs/libkbfs/folder_branch_ops.go
taruti: If the new config is identical to the old one should we bail out at top? Setting the config also bails out prefetches, potentially triggers mark-and-sweep and starts new prefetching, even if the config is identical to the old one.

shared/util/kbfs-notifications.js
```diff
@@ -162,6 +162,7 @@ export function kbfsNotification(notification: FSNotification, notify: any, stat
     // independently.
     [kbfsCommonFSNotificationType.initialized]: '',
     [kbfsCommonFSNotificationType.connection]: '',
+    [kbfsCommonFSNotificationType.syncConfigChanged]: 'Synchronization config changed',
```
jzila: Isn't this going to give notifications even if you just changed the config directly from the UI? Also this text isn't really clear. I'd say for now we can just make it blank.

go/kbfs/libkbfs/folder_branch_ops.go
taruti: I ended up just rebasing and adding the check to the notify call.

shared/util/kbfs-notifications.js
taruti: I'll make it blank.

taruti: That is keep it commented out so it will be not displayed.

taruti:
After rebasing CI passes again.

PR #17157, opened 2019-04-24
workaround RN bug that broke pasting into confirm new password input

buoyad:
It seems if the placeholder takes up the entire width of the textinput, the context menu doesn't appear. Workaround by shortening "Confirm new password" to "Confirm"

- [x] Check other password inputs


buoyad:
@keybase/react-hackers ready for review

PR #17163, opened 2019-04-24
nav header: universalize offline banner

jakob223:
Move this banner into `router-v2` since it should appear throughout the app.

r? @keybase/react-hackers 

Issue: KBFS-3968


mmaxim:
This isn't going to be showing an offline banner on chat is it?

mmaxim:
If so, I would greatly prefer it stayed on the files tab only.

PR #17164, opened 2019-04-25
a bunch of tweaks for Files

songgao:


shared/chat/selectable-big-team-channel-container.js
```diff
@@ -23,15 +22,14 @@ const mapStateToProps = (state, {conversationIDKey}) => {
 const mapDispatchToProps = () => ({})
 
 const mergeProps = (stateProps, dispatchProps, ownProps) => {
-  const isSelected = ownProps.isSelected && !Styles.isMobile
   return {
     channelname: stateProps.channelname,
-    isSelected,
+    isSelected: ownProps.isSelected,
```
mmaxim: This can't just be changed back like this. If you really want this behavior just add another prop to control what it does on mobile, I don't want the inbox selection stuff to get worse to support this feature.

songgao: @mmaxim I'm changing these back in this PR for now. Tried to figure out how it's related to the chat search but mobile chat search seems broken on master.

songgao: As I said in the comment below, I couldn't try the chat search out on master. What happens when you had this?

songgao: I'm gonna try to fix the chat search on master

songgao: @mmaxim Turns out I just needed to `gobuild`. I don't it's the right answer to override `isSelected` in this component as this is the correct "selected" style. I pushed another commit to get the same behavior by doing the mobile check in `chat/inbox-search/index.js`. PTAL thanks!

shared/fs/common/path-item-action/header.js
```diff
@@ -33,36 +32,48 @@ class Header extends React.PureComponent<Props> {
...
+                styles.stylesNameText,
+                {color: Constants.getPathTextColor(this.props.path)},
+              ])}
+            />
+          </Kb.WithTooltip>
+          {this.props.type === 'file' && (
+            <Kb.Text type="BodySmall">{Constants.humanReadableFileSize(this.props.size)}</Kb.Text>
+          )}
+          {this.props.type === 'folder' && (
+            <Kb.Text type="BodySmall">
```
jakob223: I feel like it might be slightly clearer to move this logic out into a function. But it's possible I'm just not used to how many ternary operators are standard in react world.

shared/fs/common/path-item-action/header.js
```diff
@@ -33,36 +32,48 @@ class Header extends React.PureComponent<Props> {
...
-        <Kb.Box2 direction="horizontal" style={styles.nameTextBox}>
-          <CommaSeparatedName
-            type="BodySmallSemibold"
-            name={Types.getPathName(this.props.path)}
-            elementStyle={Styles.collapseStyles([
-              styles.stylesNameText,
-              {color: Constants.getPathTextColor(this.props.path)},
-            ])}
-          />
+      <Kb.Box
```
jakob223: Why a `<Box>`? I was told to only use `Box2` in new code - is that the case?

shared/fs/common/path-item-action/menu-container.js
```diff
@@ -46,7 +46,7 @@ const mapDispatchToProps = (dispatch, {path, mode, routePath}: OwnProps) => ({
     dispatch(
       RouteTreeGen.createNavigateTo({
         parentPath: [fsTab],
-        path: [{props: {path, mode}, selected: 'reallyDelete'}],
+        path: [{props: {mode, path}, selected: 'reallyDelete'}],
```
jakob223: oops

shared/fs/common/path-item-action/header.js
songgao: I think the general rule of thumb is if you need a box that's a flex column or flex row, or need to otherwise implement other Box2 features yourself, use Box2. Otherwise, can still use Box.

In this case, we need this dummy `onClick` prop which Box2 doesn't have yet, so it'd have to be a Box. If Box2 had `onClick`, it could have been just on the Box2 below.

PR #17165, opened 2019-04-25
Passphrase recovery rework

pzduniak:


go/client/cmd_passphrase_recover.go
```diff
@@ -35,78 +38,46 @@ func NewCmdPassphraseRecoverRunner(g *libkb.GlobalContext) *CmdPassphraseRecover
...
 		return err
 	}
 
-	// Run the main recovery engine. At this point the user should be logged in with
-	// unlocked keys. This has the potential to issue all sorts of prompts
-	// but given that we should now be logged in and unlocked, it shouldn't
-	// issue any prompts.
-	return passphraseChange(c.G(), newChangeArg(pp, true))
-
 	// BUG the user sometimes ends up recovered and unlocked, but logged out after all this.
```
joshblum: Is this still true?

go/client/cmd_passphrase_recover.go
```diff
@@ -118,8 +89,8 @@ func (c *CmdPassphraseRecover) GetUsage() libkb.Usage {
 	}
 }
 
-func (c *CmdPassphraseRecover) errNoUID() error {
-	return errors.New(`Can't recover without a UID.
+func (c *CmdPassphraseRecover) errNotProvisioned() error {
```
joshblum: are these errors still used/needed?

go/client/ui.go
```diff
@@ -846,6 +846,32 @@ func (l LoginUI) DisplayResetProgress(ctx context.Context, arg keybase1.DisplayR
 	return nil
 }
 
+func (l LoginUI) ExplainDeviceRecovery(ctx context.Context, arg keybase1.ExplainDeviceRecoveryArg) error {
+	switch arg.Kind {
+	case keybase1.DeviceType_DESKTOP:
+		l.parent.Printf("On %s go to Settings > Your account, and change your password.\n", arg.Name)
+		return nil
+	case keybase1.DeviceType_MOBILE:
+		l.parent.Printf("On %s go to the app menu (the last tab) > Change password, and proceed as instructed.\n", arg.Name)
```
joshblum:
```suggestion
		l.parent.Printf("On %q, go to the "menu > Change password" to change your password.\n", arg.Name)
```

go/client/ui.go
```diff
@@ -846,6 +846,32 @@ func (l LoginUI) DisplayResetProgress(ctx context.Context, arg keybase1.DisplayR
 	return nil
 }
 
+func (l LoginUI) ExplainDeviceRecovery(ctx context.Context, arg keybase1.ExplainDeviceRecoveryArg) error {
+	switch arg.Kind {
+	case keybase1.DeviceType_DESKTOP:
+		l.parent.Printf("On %s go to Settings > Your account, and change your password.\n", arg.Name)
```
joshblum:
```suggestion
		l.parent.Printf("On %q, go to "Settings > Your account" to change your password.\n", arg.Name)
```

go/client/ui.go
```diff
@@ -846,6 +846,32 @@ func (l LoginUI) DisplayResetProgress(ctx context.Context, arg keybase1.DisplayR
...
+		return nil
```
joshblum: nit: don't need a return in each case if you `return nil` at the end

go/engine/passphrase_recover.go
```diff
@@ -0,0 +1,291 @@
...
+
+	// Reorder the devices for the list
+	devices := partitionDeviceList(ckf.GetAllActiveDevices())
+	sort.Sort(devices)
+
+	// Choose an existing device
+	expDevices := make([]keybase1.Device, len(devices))
+	idMap := make(map[keybase1.DeviceID]*libkb.Device)
+	for i, d := range devices {
+		// Don't show paper keys if the device hasn't been configured
```
joshblum:
```suggestion
		// Don't show paper keys if the user has not provisioned on this device
```

go/client/cmd_passphrase_recover.go
pzduniak: Readded them, legacy flow didn't change much so it should stay as-is.

joshblum: can you add a TODO to move these to UI calls when `legacyProvisioning` is removed/refactored for the GUI?

go/engine/passphrase_recover.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	e.arg.Username = normalized.String()
+
+	// Check if the passed username is in the map
+	_, ok := usernamesMap[normalized]
+	e.usernameFound = ok
+	return nil
+}
+
+func (e *PassphraseRecover) legacyRecovery(mctx libkb.MetaContext) (err error) {
```
joshblum: drop a TODO to remove/refactor this

go/engine/passphrase_recover.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+func (e *PassphraseRecover) loginWithPaperKey(mctx libkb.MetaContext) (err error) {
+	// First log in using the paper key
+	loginEng := NewLoginWithPaperKey(mctx.G(), e.arg.Username)
+	if err := RunEngine2(mctx, loginEng); err != nil {
+		return err
+	}
+
+	if err := e.changePassword(mctx); err != nil {
```
joshblum: what happens if you cancel the passphrase recovery here? shouldn't the whole thing abort instead of the logout/login attempt?

go/engine/passphrase_recover.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// Prereqs returns engine prereqs
+func (e *PassphraseRecover) Prereqs() Prereqs {
+	return Prereqs{}
+}
+
+// RequiredUIs returns the required UIs.
+func (e *PassphraseRecover) RequiredUIs() []libkb.UIKind {
+	return []libkb.UIKind{}
```
joshblum: `ProvisionUI`, `LoginUI` and `SecretUI` are all required

go/client/ui.go
```diff
@@ -846,6 +846,32 @@ func (l LoginUI) DisplayResetProgress(ctx context.Context, arg keybase1.DisplayR
 	return nil
 }
 
+func (l LoginUI) ExplainDeviceRecovery(ctx context.Context, arg keybase1.ExplainDeviceRecoveryArg) error {
```
joshblum: can you handle `arg.Name == ""`? prompt wouldn't really make sense as is. https://github.com/keybase/client/pull/17165/files#diff-d759c186442347fe38ee29952fd5174dR265

go/client/cmd_passphrase_recover.go
pzduniak: I'll wait with this for a meeting so that we have a consensus on how we should structure this. IIRC we want to work on this in core, so it might make sense to bundle it up with the frontend impl.

go/engine/passphrase_recover.go
pzduniak: It does abort, Logout is just a step required to clean up after LoginWithPrivateKey.

go/client/ui.go
pzduniak: Can you have a device without Description (ie. name)? I thought it's not possible.

joshblum: i don't think you can but it's a sanity check just like checking `Description != nil`

go/client/cmd_passphrase_recover.go
joshblum: right i'm just saying to mark it with a ticket so we can remember to clean it up.

go/engine/passphrase_recover.go
joshblum: If we cancel the prompt no error is returned, don't we attempt to login instead of logout?

PR #17166, opened 2019-04-25
Each tab has a stack nav2 ux change. 

chrisnojima:
- [x] New nav ux. Each tab has a stack. All routes can be pushed to any stack
- [x] Before we bookkept a single pending convo. Instead of that let's just always make a convo when going to it. This really simplifies a lot of code
- [x] iphonex
- [x] iphone
- [x] android
- [x] desktop (next pr)
- [x] undo plist/version hack


shared/actions/platform-specific/index.native.js
```diff
@@ -203,7 +202,7 @@ function* persistRoute(state, action) {
   }
 
   const path = action.payload.path
-  const top = path[path.length - 1]
+  const top = path[path.length - 2] // real top is the root of the tab (aka chatRoot) and not the tab itself
```
chrisnojima: routes changed slightly

shared/common-adapters/loading-line.native.js
```diff
@@ -1,61 +1,73 @@
 // @flow
-import Box from './box'
-import React, {Component} from 'react'
-import {NativeAnimated, NativeEasing} from './native-wrappers.native'
+import * as React from 'react'
+import {ReAnimated, ReAnimatedEasing} from './mobile.native'
```
chrisnojima: use reanimated to avoid bridge traffic

shared/constants/platform.native.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import {Dimensions, Platform, NativeModules} from 'react-native'
 import {cachesDirectoryPath} from '../util/file.native'
+import * as iPhoneXHelper from 'react-native-iphone-x-helper'
```
chrisnojima: this was a dep that i needed to pull in for something that i removed but the code is tiny and more specific than ours so keeping it

shared/app/main.native.js
```diff
@@ -36,13 +36,9 @@ class Main extends React.Component<Props> {
   }
 
   render() {
-    if (this.props.showPushPrompt) {
```
chrisnojima: if we don't do this then when we show the push prompt it thrashes the entire stack. for nav2 this means the ref to the navigator gets lost so we can lose sending actions to it!

shared/common-adapters/loading-line.native.js
```diff
@@ -1,61 +1,73 @@
...
+    R.cond(state.finished, [
+      // we stop
+      R.stopClock(clock),
+
+      // set flag ready to be restarted
+      R.set(state.finished, 0),
+      // same value as the initial defined in the state creation
+      R.set(state.position, -1),
+
+      // very important to reset this ones !!! as mentioned in the doc about timing is saying
```
MarcoPolo: Reword?

chrisnojima: This is just copy and paste. And it's a wip. Let's not review

shared/app/main.native.js
```diff
@@ -69,7 +65,8 @@ class Main extends React.Component<Props> {
             />
           </Kb.NativeKeyboardAvoidingView>
         )}
-      </React.Fragment>
+        {this.props.showPushPrompt && <PushPrompt />}
```
chrisnojima: stops router from being unmounted

shared/chat/conversation/container.js
```diff
@@ -1,27 +1,66 @@
...
 
+let NavigationEvents
+if (flags.useNewRouter) {
+  NavigationEvents = require('@react-navigation/core').NavigationEvents
+} else {
+  NavigationEvents = () => null
+}
+
 class Conversation extends React.PureComponent<SwitchProps> {
+  _onDidFocus = () => {
```
chrisnojima: convo now has to tell the system when it's the visible conversation so mark as read work correctly on tab switches

chrisnojima:
@keybase/react-hackers lets start the review of this. i need to double check android and also desktop isn't working. i'll likely make a branch off of this to continue working on that stuff. we want this in asap

shared/actions/platform-specific/index.native.js
```diff
@@ -203,20 +202,27 @@ function* persistRoute(state, action) {
   }
 
   const path = action.payload.path
-  const top = path[path.length - 1]
-  if (!top) return
+  const tab = path[path.length - 2] // real top is the root of the tab (aka chatRoot) and not the tab itself
```
MarcoPolo: I don't understand this logic. How will it always be true that the tab will the be penultimate item?

shared/actions/platform-specific/index.native.js
```diff
@@ -287,6 +293,16 @@ function* loadStartupDetails() {
   const initialShare = yield Saga._fork(getStartupDetailsFromShare)
   const [routeState, link, push, share] = yield Saga.join(routeStateTask, linkTask, initialPush, initialShare)
 
+  // Clear last value to be extra safe bad things don't hose us forever
+  yield Saga._fork(() => {
+    RPCTypes.configSetValueRpcPromise({
+      path: 'ui.routeState2',
+      value: {isNull: false, s: ''},
```
MarcoPolo: This seems weird. What's the point of `isNull` if you're just passing an empty string?

shared/chat/conversation/container.js
```diff
@@ -42,7 +55,12 @@ class Conversation extends React.PureComponent<SwitchProps> {
         // To solve this we render a blank screen on mobile conversation views with "noConvo"
         return isMobile ? null : <NoConversation />
       case 'normal':
-        return <Normal conversationIDKey={this.props.conversationIDKey} isPending={this.props.isPending} />
+        return (
+          <>
+            <NavigationEvents onDidFocus={this._onDidFocus} onWillBlur={this._onWillBlur} />
```
MarcoPolo: :sheep: :sheep: At some point would be nice if we could type the context that's required by this component, because I think this will runtime error if navigation doesn't exist in the context. Or, alternately, we can explicitly pass navigation in.

shared/chat/conversation/list-area/normal/index.native.js
```diff
@@ -203,7 +203,6 @@ class ConversationList extends React.PureComponent<Props> {
             keyExtractor={this._keyExtractor}
             // Limit the number of pages rendered ahead of time (which also limits attachment previews loaded)
             windowSize={5}
-            removeClippedSubviews={true}
```
buoyad: removed because true by default?

shared/constants/chat2/index.js
```diff
@@ -207,9 +200,7 @@ export const isUserActivelyLookingAtThisThread = (
 
   let chatThreadSelected = false
   if (flags.useNewRouter) {
-    const routePath = Router2.getVisiblePath()
-    chatThreadSelected =
-      routePath[routePath.length - 1]?.routeName === (isMobile ? 'chatConversation' : 'tabs.chatTab')
+    chatThreadSelected = true // conversationIDKey === selectedConversationIDKey is the only thing that matters in the new router
```
MarcoPolo: :sheep: add `// TODO delete when we fully commit to flags.useNewRouter`

shared/router-v2/router.native.js
```diff
@@ -42,34 +45,99 @@ const defaultNavigationOptions = {
...
+  (s, _, o) => ({
+    badgeNumber: s.badgeNumber,
+    focused: o.focused,
+    routeName: o.routeName,
+  })
+)(TabBarIcon)
+
+const TabNavigator = createBottomTabNavigator(
+  tabs.reduce((map, tab) => {
+    map[tab] = createStackNavigator(Shim.shim(routes), {
```
MarcoPolo: I *think* making each tab a stack navigator will have implications on how `constants/router2.findFullRoute` work. I can check...

shared/router-v2/router.shared.js
```diff
@@ -67,49 +70,45 @@ export const oldActionToNewActions = (action: any, navigation: any) => {
...
       if (action.payload.path) {
         const p = action.payload.path.last
           ? action.payload.path.last()
           : action.payload.path[action.payload.path.length - 1]
 
-        // a chat, we want people/inbox/chat
+        // a chat, we want inbox/chat
         if (p === 'chatConversation') {
-          sa.push(StackActions.push({params: undefined, routeName: 'tabs.chatTab'}))
+          sa.push(NavigationActions.navigate({params: undefined, routeName: 'tabs.chatTab'}))
```
MarcoPolo: this will pop up to the tabs.chatTab, right?

shared/router-v2/router.shared.js
```diff
@@ -67,49 +70,45 @@ export const oldActionToNewActions = (action: any, navigation: any) => {
...
     }
     case RouteTreeGen.clearModals: {
       const numModals = getNumModals(navigation)
       return numModals ? [StackActions.pop({n: numModals})] : []
     }
     case RouteTreeGen.navigateUp:
-      return [StackActions.pop()]
+      return [NavigationActions.back()]
     case RouteTreeGen.navUpToScreen: {
       const fullPath = Constants._getFullRouteForNavigator(navigation.state)
```
MarcoPolo: Can't you simplify this action with navactions.back?

MarcoPolo:
What happened with all the pending/search logic?

shared/actions/platform-specific/index.native.js
chrisnojima: updated

shared/chat/conversation/list-area/normal/index.native.js
chrisnojima: removed as this seemed problematic for a bunch of ppl and i wanted to see if it changed anything

shared/actions/platform-specific/index.native.js
chrisnojima: we don't care about this isNull capability of this api. we're fine with it just being a string

shared/router-v2/router.shared.js
chrisnojima: this switches the tabs

shared/router-v2/router.shared.js
chrisnojima: i don't understand

shared/chat/conversation/list-area/normal/index.native.js
buoyad: I think the default value is true, you may want to set it to `false` instead

chrisnojima: by default its not defined and it has a comment saying you shouldn't use it
https://github.com/facebook/react-native/blob/e4392b773ce5a17257c93c5c8c5ac097ca60eb88/Libraries/Lists/VirtualizedList.js#L192

buoyad: K that prop plumbs to the one in `ScrollView`, their docs are wrong
https://github.com/facebook/react-native/blob/e4392b773ce5a17257c93c5c8c5ac097ca60eb88/Libraries/Components/ScrollView/ScrollView.js#L531-L532
anyway sounds good

chrisnojima: still says 'Experimental' and i've just seen a bunch of github issues where ppl complained so i think lets turn it off and see if it makes any difference either way

shared/actions/chat2/index.js
```diff
@@ -1599,183 +1573,52 @@ const confirmScreenResponse = (_, action) => {
...
+      conversationIDKey,
+      reason: 'previewResolved',
+    })
+  })
 }
 
 const startupInboxLoad = state =>
   state.config.username ? Chat2Gen.createInboxRefresh({reason: 'bootstrap'}) : undefined
 
 const changeSelectedConversation = (state, action) => {
```
buoyad: Leaving a channel via the info panel doesn't nav you to the inbox anymore

PR #17168, opened 2019-04-25
Speed up ProofSuggestions

mlsteele:
Bring `ProofSuggestions` from ~600ms to ~200ms on my laptop. On all calls but the first in each run of the service, the bulk of the time is now in `libkb.LoadMe`, as it should be. Time was being spent parsing and loading services many times. Now that only happens after initial load if the merkley hash changes.

### DIFF
--- a/go/externals/services.go
+++ b/go/externals/services.go
@@ -118,7 +118,7 @@ func (p *proofServices) SuggestionFoldPriority() int {
 }
 
 func (p *proofServices) loadServiceConfigs() {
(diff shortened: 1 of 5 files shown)


mlsteele:
Not sure this really fixes the bug I saw though cause on my phone it can take 13s

mlsteele:
that'd be cool. Is there a good spot to store the processed artifact

joshblum:
not sure, just remember that was the other caller of this store. we could move into another ticket as well

mlsteele:
skipping CI too many flakes

PR #17176, opened 2019-04-25
Fix the nav bar on loggedOut routes

cjb:
@keybase/react-hackers 

On macOS, when logged out, the sidebar isn't rendered and so the nav2 back arrow is at the left edge of the screen and overlaps the system buttons (minimize, maximize, close).  When you're logged in it doesn't overlap because it's offset by the sidebar.

To fix this, we put a 20px paddingTop on the nav2 header when you're logged out.  But the nav2 header is also the draggable area, so this broke dragging on the top of the window when logged out -- the draggable area doesn't start until 20px down.

The best thing I could think to do is tell the header whether we're in the state of logged out and on macOS, and let it relatively offset the back arrow down below the system buttons in that case.

### DIFF
--- a/shared/router-v2/header/index.desktop.js
+++ b/shared/router-v2/header/index.desktop.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react'
 import * as Kb from '../../common-adapters'
+import * as Platform from '../../constants/platform'
 import * as Styles from '../../styles'
 import SyncingFolders from './syncing-folders'
 import flags from '../../util/feature-flags'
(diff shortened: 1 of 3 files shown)


shared/router-v2/header/index.desktop.js
```diff
@@ -53,6 +54,10 @@ class Header extends React.PureComponent<Props> {
       showDivider = false
     }
 
+    const backArrowStyle = {
+      ...(this.props.allowBack ? styles.icon : styles.disabledIcon),
+      ...(!this.props.loggedIn && Platform.isDarwin ? {position: 'relative', top: 30} : {}),
```
buoyad: Comment about why would be nice

shared/router-v2/router.desktop.js
```diff
@@ -71,7 +71,7 @@ class AppView extends React.PureComponent<any> {
           style={selectedTab ? styles.contentArea : styles.contentAreaLogin}
         >
           {scene}
-          <Header options={descriptor.options} onPop={() => childNav.pop()} allowBack={index !== 0} />
+          <Header loggedIn={selectedTab} options={descriptor.options} onPop={() => childNav.pop()} allowBack={index !== 0} />
```
buoyad: `!!selectedTab`. Seems a little fragile? Maybe `nameToTab` should have a note about it.

PR #17178, opened 2019-04-25
Use markdown to show chat descriptions

buoyad:
r? @keybase/react-hackers


buoyad:
@cjb maybe that should be the default behavior? Text could do `break-all` when `lineClamp={1}` and `break-word` for >1.

cjb:
Sounds good!

buoyad:
The storyshot changes were daunting - I ticketed DESKTOP-9515

PR #17186, opened 2019-04-26
fix lineClamp for long big team headers

buoyad:
Fix is removing `display: 'inline'` which conflicts with recently fixed lineClamp styles. The rest of this diff is fixing inbox stories + adding a long big teamname. The stories don't show in storyshots because of `react-virtualized-auto-sizer`. Fix coming in separate PR. r? @keybase/react-hackers

### DIFF
--- a/shared/chat/inbox/index.stories.js
+++ b/shared/chat/inbox/index.stories.js
@@ -248,30 +248,30 @@ const mapPropProviderProps = {
 
   // Big Team B
   bigTeamBHeader: {
-    teamname: 'techtonica',
+    teamname: 'techtonica.long.team.name.with.ellipsis',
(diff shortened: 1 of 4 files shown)


shared/chat/inbox/index.stories.js
```diff
@@ -324,7 +324,7 @@ const mapPropProviderProps = {
  * Uses either conversationIDKey or teamname as a key in mapPropProviderProps
  */
 const getPropProviderProps = own => {
-  if (own.conversationIDKey) {
+  if (own.conversationIDKey && own.conversationIDKey !== 'EMPTY') {
```
buoyad: fixes inbox storybook

PR #17188, opened 2019-04-26
Service support for FindPaymentPath, SendPathLocal, SendPathCLI

patrickxb:
`keybase wallet send-path-payment` works.

I'm sure the interface will change a bit for frontend once they get to integrating, but basics work.


protocol/avdl/stellar1/local.avdl
```diff
@@ -382,6 +382,8 @@ protocol local {
     boolean pending;
   }
 
+  SendPaymentResLocal sendPathLocal(AccountID source, string recipient, PaymentPath path, string note, string publicNote);
```
mlsteele: then `sendPaymentLocal` could be reused with a protocol addition of `union { null, PaymentPath } path`

protocol/avdl/stellar1/common.avdl
```diff
@@ -155,4 +155,13 @@ protocol common {
...
   }
+
+  record PaymentPath {
+    string sourceAmount;
+    string sourceAmountMax;
+    Asset sourceAsset;
+    array<Asset> path;
+    string destinationAmount;
+    Asset destinationAsset;
+  }
```
mlsteele: If `PaymentPath` were
```
{
  string sourceAmount;
  string sourceAmountMax;
  Asset sourceAsset;
  array<Asset> path;
}
```

protocol/avdl/stellar1/local.avdl
```diff
@@ -526,23 +528,35 @@ protocol local {
...
     // Direct: tx ID of the payment tx
     // Relay : tx ID of the funding payment tx
     TransactionID txID;
   }
 
+  SendResultCLILocal sendCLILocal(string recipient, string amount, Asset asset, string note,
+                                  string displayAmount, string displayCurrency, boolean forceRelay,
+                                  string publicNote, AccountID fromAccountID);
+
+  SendResultCLILocal sendPathCLILocal(AccountID source, string recipient, PaymentPath path, string note, string publicNote);
```
mlsteele: same deal as `sendPaymentLocal`

go/stellar/stellar.go
```diff
@@ -696,6 +696,130 @@ func sendPayment(mctx libkb.MetaContext, walletState *WalletState, sendArg SendP
...
+		return res, err
+	}
+
+	sp, unlock := NewSeqnoProvider(mctx, walletState)
+	defer unlock()
+
+	sig, err := stellarnet.PathPaymentTransaction(senderSeed, to, sendArg.Path.SourceAsset, sendArg.Path.SourceAmountMax, sendArg.Path.DestinationAsset, sendArg.Path.DestinationAmount, AssetSliceToAssetBase(sendArg.Path.Path), sendArg.PublicMemo, sp, nil, baseFee)
+	if err != nil {
+		return res, err
+	}
```
mlsteele: `sendPathPayment` looks on the whole the same as `sendPayment` except for this here `PathPaymentTransaction` and `walletState.SubmitPathPayment`. `sendPayment` already switches `PaymentXLMTransaction` by `if !funded {`. What about deduping `sendPathPayment` and `sendPayment` by taking an optional path-start?

timebounds are nil

go/stellar/stellar.go
```diff
@@ -2133,3 +2257,27 @@ func AirdropStatus(mctx libkb.MetaContext) (stellar1.AirdropStatus, error) {
...
+
+	query := stellar1.PaymentPathQuery{
+		Source:           sourceEntry.AccountID,
+		Destination:      stellar1.AccountID(recipient.AccountID.String()),
+		SourceAsset:      sourceAsset,
+		DestinationAsset: destinationAsset,
+		Amount:           amount,
+	}
+	return remoter.FindPaymentPath(mctx, query)
```
mlsteele: if the remote server swapped out same values like destination account could that trick the client?

go/stellar/stellar.go
patrickxb: I also didn't want to mess with sendPayment.  It works.  This is new and an "advanced" feature that few users will use.

There are other differences too, like request clearing, in chat send notifications, etc.

I think it's fine for them to be split, especially now as a new feature.

I'd be more in favor of taking some of the common functionality out of both of them into little functions that they both call.

mlsteele: It could be done in a way that would almost provably not change what sendPayment executes when `path == nil`. It may be hard to tell going forward the difference between purposeful differences and accidental desync. Which wouldn't be the case with `if path == nil`s. It'd be easy to split later but harder to merge. I approve whatever decision you make.

go/stellar/stellar.go
patrickxb: FindPaymentPath does not return the destination account.

If the remote server modified the argument to use a different destination account for the call to find the path, either the destination account would have a trustline to the asset or not.  That's the only difference.  Horizon doesn't even look at the destination account (I filed a ticket about this), but stellarnet checks that it has a trustline to the destination asset.  So worst case a malicious server would either return a path that wouldn't work when submitted or an error for a path that would work.  Neither of these would result in a loss of funds, just bad user experience on an advanced feature.

The call to SendPathPayment requires a recipient, which is looked up to get the account ID on the client at the time that the path payment transaction is constructed.

patrickxb:
CI errors are now isolated to kbfs-server

PR #17189, opened 2019-04-26
Paramproof wrap long username

mlsteele:
it was sticking out of its box


shared/profile/generic/enter-username/index.js
```diff
@@ -305,9 +305,12 @@ const styles = Styles.styleSheetCreate({
     opacity: 0.75,
   },
   paddingRightTiny: {paddingRight: Styles.globalMargins.tiny},
-  placeholder: {
-    color: Styles.globalColors.black_40,
-  },
+  placeholder: Styles.platformStyles({
```
buoyad: this is also used by the invisible text that makes the service suffix sliding over work. can you make a separate style for the one in `Unreachable` that we want to wrap?

PR #17191, opened 2019-04-26
SelectiveSync fixes

joshblum:
- increase delay to selective sync, especially important on mobile
- drop use of `ConvLoader` in `reindexConv` to reduce complexity, instead check mobile background state in `Indexer` directly
- separate index tokenization version from on disk format version, cleanup old disk formats
- add `PokeSync`, which triggers selective sync to run after a search completes
- add `CancelSync`, canceling any running sync at the start of search

### DIFF
--- a/go/chat/search/indexer.go
+++ b/go/chat/search/indexer.go
@@ -45,7 +45,7 @@ func NewIndexer(g *globals.Context) *Indexer {
 		DebugLabeler: utils.NewDebugLabeler(g.GetLog(), "Search.Indexer", false),
 		store:        newStore(g),
 		pageSize:     defaultPageSize,
-		stopCh:       make(chan chan struct{}),
+		stopCh:       make(chan chan struct{}, 10),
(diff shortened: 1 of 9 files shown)


joshblum:
will add selectivesync cancel/queue during searches

go/chat/search/indexer.go
```diff
@@ -84,32 +98,94 @@ func (idx *Indexer) Start(ctx context.Context, uid gregor1.UID) {
...
+	for {
 		select {
-		case <-time.After(libkb.DefaultBgTickerWait):
-			idx.SelectiveSync(ctx, uid, false /*forceReindex */)
-		case <-stopCh:
+		case <-idx.cancelSyncCh:
+			cancelSync()
+		case <-idx.pokeSyncCh:
+			attemptSync(ctx)
+		case <-after:
```
mmaxim: Is this right? Seems like we go every 10 seconds with this.

PR #17192, opened 2019-04-26
Inline dropdown for role picker tweak

MarcoPolo:
@keybase/react-hackers 
Screen Shot 2019-04-26 at 1 30 16 PM

### DIFF
--- a/shared/common-adapters/box.desktop.js
+++ b/shared/common-adapters/box.desktop.js
@@ -5,7 +5,7 @@ import type {Box2Props} from './box'
 
 class Box extends React.PureComponent<any> {
   render() {
-    const {forwardedRef, ...rest} = this.props
+    const {forwardedRef, inline, ...rest} = this.props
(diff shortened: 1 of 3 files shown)


shared/common-adapters/box.desktop.js
```diff
@@ -5,7 +5,7 @@ import type {Box2Props} from './box'
 
 class Box extends React.PureComponent<any> {
   render() {
-    const {forwardedRef, ...rest} = this.props
+    const {forwardedRef, inline, ...rest} = this.props
```
chrisnojima: why is this here?

MarcoPolo: whoops. I did that as a temp hack. will fix

MarcoPolo: I added inline prop to button box, and styled by default forwards all props so I was getting an error in react because div was getting a boolean.

PR #17193, opened 2019-04-26
rework search index

mmaxim:


go/chat/search/wrapper.go
```diff
@@ -0,0 +1,247 @@
...
+		false /* resolveSupersedes*/, &reason)
+	if err != nil {
+		// Log but ignore error
+		w.Debug(ctx, "fetchSupersededMsgs: unable to get fetch messages: %v", err)
+		return nil
+	}
+	return supersededMsgs
+}
+
+// addTokensLocked add the given tokens to the index under the given message
```
joshblum:
```suggestion
// addTokens adds the given tokens to the index under the given message
```

go/chat/search/indexer.go
```diff
@@ -172,7 +268,18 @@ func (idx *Indexer) Add(ctx context.Context, convID chat1.ConversationID, uid gr
 	defer idx.Trace(ctx, func() error { return err },
 		fmt.Sprintf("Indexer.Add convID: %v, msgs: %d", convID.String(), len(msgs)))()
 	defer idx.consumeResultsForTest(convID, err)
-	return idx.store.add(ctx, convID, uid, msgs)
+	defer func() {
+		if err == nil {
```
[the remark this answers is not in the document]
mmaxim: Not sure it matters, we are mostly here if there is an actual change to be made right?

go/bind/keybase.go
```diff
@@ -350,6 +350,7 @@ func Reset() error {
 // ForceGC Forces a gc
 func ForceGC() {
 	fmt.Printf("Flushing global caches\n")
+	kbChatCtx.Indexer.ClearMemory()
```
joshblum: why not add to `FlushCaches`?

go/chat/search/indexer.go
```diff
@@ -70,6 +81,73 @@ func (idx *Indexer) SetReindexCh(ch chan chat1.ConversationID) {
...
+			idx.Debug(ctx, "Flush: failed to put conv index: %s: err: %s", convID, err)
+		}
+		lock.Release(ctx)
+	}
+}
+
+func (idx *Indexer) flushLoop(stopCh chan struct{}) {
+	for {
+		select {
+		case <-time.After(30 * time.Second):
```
joshblum: can you make this 0 in tests so we have coverage

go/chat/search/indexer.go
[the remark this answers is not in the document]
joshblum: yeah probably fine, it's possible we don't modify the index at all but more of an edge case

go/chat/search/indexer.go
```diff
@@ -201,20 +330,34 @@ type reindexOpts struct {
...
 	}
 	minIdxID := missingIDs[0]
 	maxIdxID := missingIDs[len(missingIDs)-1]
 
 	convID := conv.GetConvID()
 	defer idx.Trace(ctx, func() error { return err },
-		fmt.Sprintf("Indexer.reindex: conv: %v, minID: %v, maxID: %v, numMissing: %v",
+		fmt.Sprintf("reindexConv: conv: %v, minID: %v, maxID: %v, numMissing: %v",
 			rconv.GetName(), minIdxID, maxIdxID, len(missingIDs)))()
+	defer func() {
```
joshblum: let's ditch this i don't think it's worth to log every time unless we're debugging

go/chat/search/storage.go
```diff
@@ -130,9 +130,11 @@ func (s *store) getConvIndex(ctx context.Context, convID chat1.ConversationID, u
 			}
 		}
 		if err != nil {
+			s.Debug(ctx, "getConvIndex: failed to get index: %s", err)
 			if derr := s.deleteLocked(ctx, convID, uid); derr != nil {
 				s.Debug(ctx, "unable to delete: %v", derr)
 			}
+			err = nil
```
joshblum: we never really expect errors here i think it would be better to return them so we can a log if it happens in the wild

PR #17195, opened 2019-04-26
SendLinkToChat: sort channel list

jakob223:
@joshblum said he would look at adding a parameter for whether to only get channels the user is a member of to the `GetTLFConversationsLocal` RPC. I'm going to wait for that before doing the other half of the work here.

Issue: KBFS-3762

### DIFF
--- a/shared/fs/send-link-to-chat/container.js
+++ b/shared/fs/send-link-to-chat/container.js
@@ -115,16 +115,18 @@ const mergeProps = (stateProps, {onCancel, _onSent, _send, _selectChannel}, ownP
 
   // big team
 
-  const channels = stateProps._sendLinkToChat.channels.reduce(
-    (channels, channelname, convID) => [
-      ...channels,
-      {
-        channelname,
-        convID,
-      },
-    ],
-    []
-  )
+  const channels = stateProps._sendLinkToChat.channels
+    .reduce(
+      (channels, channelname, convID) => [
+        ...channels,
+        {
+          channelname,
+          convID,
+        },
+      ],
+      []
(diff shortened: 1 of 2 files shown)


joshblum:
@jakob223 actually you can just filter on `InboxUIItem.MemberStatus == chat1.ConversationMemberStatus_ACTIVE` so no need for an RPC change, misspoke before

shared/fs/send-link-to-chat/container.js
```diff
@@ -115,16 +115,20 @@ const mergeProps = (stateProps, {onCancel, _onSent, _send, _selectChannel}, ownP
...
+        ...channels,
+        {
+          channelname,
+          convID,
+        },
+      ],
+      []
+    )
+    .sort(function(a, b) {
+      return a.channelname.localeCompare(b.channelname, 'en', {sensitivity: 'base'})
```
jzila: `'en'` is wrong here, it should be `undefined`.

shared/fs/send-link-to-chat/container.js
```diff
@@ -115,16 +115,20 @@ const mergeProps = (stateProps, {onCancel, _onSent, _send, _selectChannel}, ownP
...
+      (channels, channelname, convID) => [
+        ...channels,
+        {
+          channelname,
+          convID,
+        },
+      ],
+      []
+    )
+    .sort(function(a, b) {
```
jzila: This can be an arrow function:
```javascript
.sort((a, b) => a.channelname.localCompare(b.channelname, ...))
```

PR #17196, opened 2019-04-26
Fix routing error when we show the push prompt

MarcoPolo:
@keybase/react-hackers Fun issue. The `PushPrompt` Unmounted the navigator which caused it to lose all state and drop messages from actions.


MarcoPolo:
@chrisnojima is that up yet?

MarcoPolo:
Closing since it was fixed in Nojima's PR

PR #17197, opened 2019-04-26
replace Oops popups with in screen views; fix delete popup routing

songgao:


songgao:
@keybase/react-hackers bump!

shared/constants/types/fs.js
```diff
@@ -459,13 +467,13 @@ export type _State = {|
   loadingPaths: I.Map<Path, I.Set<string>>,
   localHTTPServerInfo: LocalHTTPServer,
   destinationPicker: DestinationPicker,
-  sendLinkToChat: SendLinkToChat,
```
jakob223: lol why does flow allow this sort of thing 🤦‍♂

songgao:
> Oops screens seem to come with black bars when they're files you don't have permission to see. We should suppress the black bars.

Is this the blackbar from `SimpleFSSimpleFSFolderSyncConfigAndStatusRpcPromise`? If so it's fixed in the latest commit as well. The reason is `.catch` wasn't handled before.

> Also the nonexistent user issue that I mentioned in chat could be handled here if you like.

fixed!

> Do you want to delete the delete folder feature flag (or at least turn it on by default) now?

Seems like a good idea to wait until the delete card/banner is done, and move it to `isAdmin`. Then after a design sign off we can get rid of the flag. How does it sound?

PR #17198, opened 2019-04-27
fix selectable in SearchFilter on desktop

songgao:
I fixed the draggable thing in a previous PR, but just realized that didn't actually make the test in the box selectable. This PR fixes that.

### DIFF
--- a/shared/common-adapters/search-filter.js
+++ b/shared/common-adapters/search-filter.js
@@ -217,7 +217,7 @@ class SearchFilter extends React.PureComponent<Props, State> {
         onMouseLeave={this._mouseLeave}
         onClick={
           this.props.onClick ||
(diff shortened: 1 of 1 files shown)


shared/common-adapters/search-filter.js
```diff
@@ -217,9 +217,9 @@ class SearchFilter extends React.PureComponent<Props, State> {
         onMouseLeave={this._mouseLeave}
         onClick={
           this.props.onClick ||
-          // Can't just make a null for Kb.ClickableBox here when focused, as
-          // that'd cause PlainInput to be re-constructed.
-          this._focus
+          // On moile we can't just make a null for Kb.ClickableBox here when
```
jzila: s/moile/mobile

PR #17199, opened 2019-04-27
search filter to:@mention/--sent-to

joshblum:
we could also allow multiple `to:` to be present, right not we just allow one


joshblum:
merging over kbfs flakes

joshblum:
merging over kbfs flakes

PR #17215, opened 2019-04-29
re-structure index to take more advantage of LevelDB 

mmaxim:
Stores the index in a new format, where each term and alias is given a key in LevelDB (after being sent through HMAC). This way we don't need to load entire maps of data into memory to perform searches. Also does a few other things:

1.) Makes requirements stricter for tokens and alias length.
2.) Index in recency order in post search.
3.) Fixes an error in the index % calc.


go/chat/search/md.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"unsafe"
+
+	"github.com/keybase/client/go/protocol/chat1"
+)
+
+const indexMetadataVersion = 3
+
+type indexMetadata struct {
+	SeenIDs map[chat1.MessageID]chat1.EmptyStruct `codec:"s"`
```
joshblum: store as array on disk? some for terms/aliases

go/chat/search/utils.go
```diff
@@ -39,15 +39,15 @@ var stripSeps = []string{
...
-	if len(token) <= 2 {
+	if len(token) < minTokenLength {
 		return nil
 	}
 	for i := range token {
-		if i <= 2 {
+		if i < minTokenLength {
 			continue
 		}
 		// Skip any prefixes longer than 20 to limit the index size.
```
joshblum:
```suggestion
		// Skip any prefixes longer than `maxPrefixLength` to limit the index size.
```

go/chat/search_test.go
```diff
@@ -525,19 +524,11 @@ func TestChatSearchInbox(t *testing.T) {
 			}
 		}
 
-		verifyIndex := func(expectedIndex *chat1.ConversationIndex) {
```
joshblum: could we add a function to the indexer/store that rebuilds the old style map (or something similar) to keep this checking in? these tests covered the indexer storage in a addition to the search logic but now we have no verification of storage

go/chat/search/session.go
```diff
@@ -105,28 +104,18 @@ func (s *searchSession) percentIndexedLocked() int {
...
+func (s *searchSession) searchConv(ctx context.Context, convID chat1.ConversationID) (msgIDs []chat1.MessageID, err error) {
+	defer s.indexer.Trace(ctx, func() error { return err }, fmt.Sprintf("searchConv convID: %s", convID))()
 	var allMsgIDs mapset.Set
 	for token := range s.tokens {
 		matchedIDs := mapset.NewThreadUnsafeSet()
-
-		// first gather the messages that directly match the token
-		for msgID := range convIdx.Index[token] {
-			matchedIDs.Add(msgID)
+		idMap, err := s.indexer.store.GetHits(ctx, s.uid, convID, token)
```
joshblum: could just have this return a `mapset` directly

go/chat/search/storage.go
```diff
@@ -13,32 +16,51 @@ import (
...
+var refAliasEntry = newAliasEntry()
+
 type store struct {
+	utils.DebugLabeler
 	lockTab *libkb.LockTable
 	globals.Contextified
-	utils.DebugLabeler
-	encryptedDB *encrypteddb.EncryptedDB
+	keyFn      func(ctx context.Context) ([32]byte, error)
+	aliasCache *lru.Cache
```
joshblum: these should get cleared `OnLogout`/`OnDbNuke`

go/chat/search/storage.go
```diff
@@ -48,204 +70,350 @@ func newStore(g *globals.Context) *store {
...
+	}, nil
+}
+
+func (s *store) aliasKey(ctx context.Context, dat string) (res libkb.DbKey, err error) {
+	material, err := s.keyFn(ctx)
+	if err != nil {
+		return res, err
+	}
+	hasher := hmac.New(sha256.New, material[:])
+	hasher.Write([]byte(dat))
```
joshblum: should have `uid` here too

go/chat/search/storage.go
```diff
@@ -48,204 +70,350 @@ func newStore(g *globals.Context) *store {
...
-		ret.Metadata.SeenIDs[msgID] = chat1.EmptyStruct{}
+	return s.edb.Put(ctx, key, te)
+}
+
+func (s *store) putAliasEntry(ctx context.Context, alias string, ae *aliasEntry) error {
+	key, err := s.aliasKey(ctx, alias)
+	if err != nil {
+		return err
 	}
+	s.aliasCache.Remove(key)
```
joshblum: `Put`?

go/chat/search/storage.go
```diff
@@ -48,204 +70,350 @@ func newStore(g *globals.Context) *store {
...
-	for alias, tokenMap := range idx.Alias {
-		tokens := make([]string, len(tokenMap))
-		jj := 0
-		for token := range tokenMap {
-			tokens[jj] = token
-			jj++
-		}
-		entry.Alias[j] = chat1.AliasTuple{Alias: alias, Tokens: tokens}
-		j++
+type addTokenBatch struct {
```
joshblum: why not have the `Put` methods just write to memory and then have periodic flushing to disk? would be a bigger win when reindexing a bunch of pages of messages

shared/actions/chat2/index.js
```diff
@@ -1374,7 +1374,7 @@ function* threadSearch(state, action) {
           isRegex: false,
           maxConvsHit: 0,
           maxConvsSearched: 0,
-          maxHits: Constants.inboxSearchMaxTextMessages,
+          maxHits: 1000,
```
joshblum: revert?

mmaxim: This is intended, otherwise we only search 25 messages here.

go/chat/search_test.go
mmaxim: Is there any value in doing that check outside of making sure the searches do the right thing?

go/chat/search/storage.go
mmaxim: Don't want any chance that MD gets out of sync with these keys.

go/chat/search/md.go
mmaxim: Don't think it matters.

go/chat/search_test.go
joshblum: just to sanity check and to make sure the structures are as we expect, I could see a search not returning a result even if some alias data remained present for example

mmaxim: Ok I think I will push that to a different PR, I think this test more or less lets you know this works, and I think it will be non-trivial work to get this working like that again.

go/chat/search/session.go
mmaxim: I guess, seemed like just something this guy was using though.

joshblum: don't have to...

go/chat/search_test.go
joshblum: yup that sounds good

go/chat/search/indexer.go
```diff
@@ -477,7 +490,7 @@ func (idx *Indexer) indexConvWithProfile(ctx context.Context, conv types.RemoteC
 		return res, err
 	}
 	res.DurationMsec = gregor1.ToDurationMsec(time.Now().Sub(startT))
-	res.IndexSizeDisk = 0
+	res.IndexSizeDisk = int(md.Size())
 	res.IndexSizeMem = 0
```
joshblum: this should be `md.Size()` and the disk size just the raw level db value

PR #17217, opened 2019-04-29
Remove ShouldUseParameterizedProofs switch

mlsteele:


mlsteele:
Not directly. That feature flag is used to enable sock puppets to try proof types ahead of all users. For example theqrl used some sock puppets with the flag to try out their integration on prod.

PR #17218, opened 2019-04-29
Keep the chat headline emoji small when on its own

cjb:
@keybase/react-hackers 

Before above, after below:

Screen Shot 2019-04-29 at 8 20 26 AM


cjb:
Thanks, will do.

cjb:
@buoyad How's this?  Feels like my prop naming game is off today, feel free to recommend something else :)

PR #17221, opened 2019-04-29
SFMIBanner: disable showing on linux

jakob223:
Disable showing the system file manager integration banner on linux, even
when it thinks it should be showing. I'm not confident that this is the right
approach, but the user log did not enlighten me as to why the banner was
appearing at all, so this seemed like the simplest approach.

Issue: KBFS-4084

### DIFF
--- a/shared/fs/folder/container.js
+++ b/shared/fs/folder/container.js
@@ -4,6 +4,7 @@ import {namedConnect} from '../../util/container'
 import Folder from '.'
 import * as Types from '../../constants/types/fs'
 import * as Constants from '../../constants/fs'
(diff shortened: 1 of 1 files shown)


songgao:
I think the problem is we have incorrect data in the redux store. On Linux the sfmi status should always be "enabled": https://github.com/keybase/client/blob/cf93a9467422e16103098d95f6537c36df5c20cd/shared/constants/fs.js#L260

It might be because we are triggering refreshDriverStatus on rpcStatus change. It should either not be triggered on Linux, or the handler should do nothing on Linux.

PR #17222, opened 2019-04-29
Ease up on ExperimentalGenericProofs feature flag

mlsteele:
Hit `FeatureFlagSet` once per ProofSuggestions call instead of once per proof type. Don't know if the cpu effect matters but it will stop logging this line so many times
```
2019-04-29T11:59:12.385349-04:00 ▶ [DEBU keybase features.go:150] 10c3 Feature (cached) "experimental_generic_proofs" -> false [tags:FEAT=xPPKIV9ht5tl,PV=vnYBZ07NgUCC]
```

### DIFF
--- a/go/libkb/services.go
+++ b/go/libkb/services.go
@@ -185,6 +185,18 @@ func (t *BaseServiceType) CanMakeNewProofs(mctx MetaContext) bool {
 	return !t.displayConf.CreationDisabled
 }
 
+func (t *BaseServiceType) CanMakeNewProofsSkipFeatureFlag(mctx MetaContext) bool {
+	t.Lock()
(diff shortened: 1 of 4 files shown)


go/libkb/services.go
```diff
@@ -185,6 +185,18 @@ func (t *BaseServiceType) CanMakeNewProofs(mctx MetaContext) bool {
 	return !t.displayConf.CreationDisabled
 }
 
+func (t *BaseServiceType) CanMakeNewProofsSkipFeatureFlag(mctx MetaContext) bool {
```
joshblum: prefer `CanMakeNewProofs` just having a bool arg to toggle the behavior so we don't dupe the code

PR #17223, opened 2019-04-29
Fix wallet disclaimer modal display on nav2

buoyad:
- [ ] ~Maybe second commit should be replaced with something more like how nav1 does it (the benefit w/ the old way is that navigating to the wallet tab is valid even if you haven't accepted disclaimer)~ nope

### DIFF
--- a/shared/actions/json/route-tree.json
+++ b/shared/actions/json/route-tree.json
@@ -2,7 +2,8 @@
   "prelude": [
     "import * as Types from '../constants/types/route-tree'",
     "import * as Constants from '../constants/route-tree'",
-    "import * as RCConstants from '../route-tree'"
+    "import * as RCConstants from '../route-tree'",
+    "import * as Tabs from '../constants/tabs'"
   ],
   "actions": {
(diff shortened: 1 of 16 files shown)


buoyad:
@keybase/react-hackers this is ready! Switches some modals over to `modalRoutes`. 

The bulk of this is redoing the accept disclaimer flow for wallets. Now when you do a wallets thing without having accepted the disclaimer we'll push the disclaimer modal on top of whatever screen you were on. We nav into the wallets tab only after accepting.

shared/actions/json/route-tree.json
```diff
@@ -51,5 +52,9 @@
       "_description": "ONLY used by the new nav. Navigates up to this route if it already exists, noops otherwise.",
       "routeName": "string"
     },
+    "switchTab": {
+      "_description": "ONLY used by the new nav. Switch to a different tab.",
+      "tab": "Tabs.DesktopTab"
```
chrisnojima: lets make this not just desktop

PR #17224, opened 2019-04-29
Stop propagation of inline dropdown

MarcoPolo:
@keybase/react-hackers 

In this case inline dropdown was inside a checkbox component. The event from clicking the dropdown propagated up to the checkbox component. I made the fix inside inline dropdown because I don't think you ever want the click to propagate and having callers do this seems tedious & error prone.


cjb:
I think @songgao hit this recently, CC'ing so that Song could remove workarounds

songgao:
Thanks for copying. This seems to be a different component though. My issue was specifically for a floating-menu spawned from a clickable inside ListItem2

PR #17232, opened 2019-04-29
WIP: Each tab has a stack navigator desktop

chrisnojima:
@keybase/react-hackers this applies these changes to desktop


chrisnojima:
- [ ] check push

buoyad:
Switching back to a tab always resets you to the root of the tab - intended?

chrisnojima:
Nope will fix

buoyad:
Navigating into a chat on another tab goes to the full inbox-and-conversation, I didn't expect to see the inbox

chrisnojima:
That's on purpose

buoyad:
Sure, just pointing out it was a weird experience for me

chrisnojima:
fixed tab clear @buoyad

chrisnojima:
- [ ] tab click to root

buoyad:
The weird back behavior is inconsistent but here's one repro:

1. Fresh start app on people tab
2. Nav into a profile
3. Switch to teams tab
4. Hit back -> end up back on people tab

Sometimes you end up on the profile, sometimes on the root. If you skip step 2 step 4 doesn't nav anywhere.

shared/router-v2/header/index.desktop.js
```diff
@@ -99,7 +99,7 @@ class Header extends React.PureComponent<Props, State> {
               type="iconfont-arrow-left"
               style={backArrowStyle}
               color={iconColor}
-              onClick={this.props.allowBack || this.props.loggedIn ? this.props.onPop : null}
```
buoyad: cc @cjb I removed this so the back button would be disabled at the root of the tab. Do we need this for something?

cjb: Nope!

PR #17233, opened 2019-04-29
Electron 5.0.6

chrisnojima:
- [x] Updates electron to 5.0.6


cjb:
@chrisnojima Shall we get this in now? And, I see 5.0.5 came out with a "Fixed issue on macOS Catalina where the tray icon would be highlighted incorrectly and sometimes invisible. #18852".

PR #17235, opened 2019-04-29
path-item-action: don't include write actions on unwritable objects

jakob223:
Also fix word wrapping in the move or copy dialog.

r? @keybase/react-hackers 

Issue: KBFS-3833

### DIFF
--- a/shared/fs/common/path-item-action/layout.js
+++ b/shared/fs/common/path-item-action/layout.js
@@ -72,7 +72,7 @@ const getRawLayout = (
         ...empty,
         ...(mode === 'screen'
           ? {
-              newFolder: pathItem.type === 'folder',
+              newFolder: pathItem.writable,
               openChatNonTeam: parsedPath.kind === 'group-tlf',
               openChatTeam: parsedPath.kind === 'team-tlf',
             }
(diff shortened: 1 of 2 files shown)


shared/fs/common/path-item-action/layout.js
```diff
@@ -95,7 +95,7 @@ const getRawLayout = (
             }
           : {}),
         copyPath: true,
-        delete: pathItem.type === 'file' || flags.enableDeleteFolder,
+        delete: (pathItem.writable && pathItem.type === 'file') || flags.enableDeleteFolder,
```
songgao: Seems like pathItem.writable should be outside the parentheses and the flag should be inside

shared/fs/common/path-item-action/layout.js
```diff
@@ -72,7 +72,7 @@ const getRawLayout = (
         ...empty,
         ...(mode === 'screen'
           ? {
-              newFolder: pathItem.type === 'folder',
+              newFolder: pathItem.writable && pathItem.type === 'folder',
```
songgao: `pathItem.type === 'folder'` shouldn't actually be necessary as we know for sure a TLF root is a folder; Not sure why I did that in the first place; maybe copy/paste laziness. But not a big deal / harmless.

songgao: Also I think https://github.com/keybase/client/pull/17235/files#diff-924982ffd1e49c940aa1080b466eb57fR92 needs this too (where the `type === 'folder'` is needed).

PR #17236, opened 2019-04-29
actions/fs: don't do driver refreshes on linux

jakob223:
(cc @jzila since you reviewed the previous version of this)

Issue: KBFS-4084


jzila:
OH wait actually

jzila:
You can make the `Saga.chainAction` not do anything on Linux.

jakob223:
@jzila how do I do that?

jzila:
@jakob223: guard this line with an `if (!isLinux)`:
https://github.com/keybase/client/blob/9206032e1df1cd133c84e2b1e76f02afbb9be121/shared/actions/fs/platform-specific.desktop.js#L388-L391

jakob223:
Heh - I kinda forgot that was actual JS code - I tend to think of it as boilerplate in some mysterious language that looks a lot like JS but only has one variety of statement.

PR #17237, opened 2019-04-29
jenkins: Add GPG env var to represent new gpgv2 binary

jzila:


jzila:
Different error in CI. GPG tests now pass.

PR #17240, opened 2019-04-29
Surya/core 10748/persistent team cache

heronhaye:
TODO

- [x] Mirror freezing functionality for FTL (currently, does not do any deletion)
- [x] Tests for freezing FTL
- [x] Tombstoning + tests

### DIFF
--- a/go/teams/loader.go
+++ b/go/teams/loader.go
@@ -111,14 +111,49 @@ func (l *TeamLoader) Load(ctx context.Context, lArg keybase1.LoadTeamArg) (res *
 	return l.load1(ctx, me, lArg)
 }
 
-func (l *TeamLoader) Delete(ctx context.Context, teamID keybase1.TeamID) (err error) {
-	defer l.G().CTraceTimed(ctx, fmt.Sprintf("TeamLoader#Delete(%v)", teamID), func() error { return err })()
+func newFrozenChain(chain *keybase1.TeamSigChainState) keybase1.TeamSigChainState {
+	return keybase1.TeamSigChainState{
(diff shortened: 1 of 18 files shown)


go/teams/loader.go
```diff
@@ -111,14 +111,27 @@ func (l *TeamLoader) Load(ctx context.Context, lArg keybase1.LoadTeamArg) (res *
...
 	lock := l.locktab.AcquireOnName(ctx, l.G(), teamID.String())
 	defer lock.Release(ctx)
-
-	return l.storage.Delete(libkb.NewMetaContext(ctx, l.G()), teamID, teamID.IsPublic())
+	mctx := libkb.NewMetaContext(ctx, l.G())
+	td, frozen := l.storage.Get(mctx, teamID, teamID.IsPublic())
+	if frozen || td == nil {
+		return nil
+	}
+	frozenChain := keybase1.TeamSigChainState{
```
heronhaye: This seems prone to future bugs. In particular, this is enough to satisfy the TeamDataGeneric interface which needs ID() and Public() but later the storage might use another chain value in the future which would result in a crash/bad data when trying to get data that has been cleared out.

One idea (other than a test) is to make a new "frozen team data" type that would implement ID() and Public(), and store that in the db instead of a stubbed out full `TeamData`. But I think it would require clearing the team data caches, so not great.

heronhaye:
ok @maxtaco, this is ready to look at again. I added tombstoning which permanently marks it as deleted in the cache and also mirrored functionality and tests for FTL. Tombstoning is somewhat of a scary change so won't merge until after release of course.

PR #17241, opened 2019-04-29
fix panic if logged in account is deleted

pzduniak:
truly impressive work in this pullreq


mlsteele:
Argh why is this possible

PR #17242, opened 2019-04-29
fix KeyboardAvoidingView for Files tab by plumbing through a headerHeight to nav2

songgao:
We are providing a custom header for the Files tab, and it breaks KeyboardAvoidingView. So set an explicit `keyboardVerticalOffset`.

Not sure if it's the right way to do it, but this works for me.

### DIFF
--- a/shared/fs/nav-header/mobile-header.js
+++ b/shared/fs/nav-header/mobile-header.js
@@ -7,7 +7,6 @@ import * as Types from '../../constants/types/fs'
 import {namedConnect} from '../../util/container'
 import Actions from './actions'
 import * as Kbfs from '../common'
-import {isIPhoneX} from '../../constants/platform'
 import flags from '../../util/feature-flags'
 
 type BannerType = 'none' | 'offline'
@@ -58,8 +57,7 @@ class MobileHeader extends React.PureComponent<Props, State> {
   }
 }
 
-const statusBarHeight = isIPhoneX ? 45 : 20
-export const height = statusBarHeight + 88
+export const height = Styles.statusBarHeight + 88
 
 const styles = Styles.styleSheetCreate({
   blue: {
@@ -70,7 +68,7 @@ const styles = Styles.styleSheetCreate({
     height,
     maxHeight: height,
     minHeight: height,
-    paddingTop: statusBarHeight,
+    paddingTop: Styles.statusBarHeight,
   },
   expandedTitleContainer: {
     backgroundColor: Styles.globalColors.white,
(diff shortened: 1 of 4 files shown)


shared/fs/nav-header/mobile-header.js
```diff
@@ -58,13 +58,19 @@ class MobileHeader extends React.PureComponent<Props, State> {
   }
 }
 
+const statusBarHeight = isIPhoneX ? 45 : 20
```
chrisnojima: I believe this is in styles already

shared/router-v2/shim.native.js
```diff
@@ -11,18 +11,23 @@ const shimNewRoute = (Original: any) => {
   class ShimmedNew extends React.PureComponent<any, void> {
     static navigationOptions = Original.navigationOptions
     render() {
+      const navigationOptions =
+        typeof Original.navigationOptions === 'function'
+          ? Original.navigationOptions(this.props.navigation)
+          : Original.navigationOpptions
```
chrisnojima: Typo

shared/router-v2/shim.native.js
```diff
@@ -11,18 +11,23 @@ const shimNewRoute = (Original: any) => {
   class ShimmedNew extends React.PureComponent<any, void> {
     static navigationOptions = Original.navigationOptions
     render() {
+      const navigationOptions =
+        typeof Original.navigationOptions === 'function'
+          ? Original.navigationOptions(this.props.navigation)
```
chrisnojima: The params should be different. Should be an object.  See the nav docs

songgao: Oh! that's why I had issue with navigation being undefined. Thanks!

shared/fs/nav-header/mobile-header.js
songgao: https://github.com/keybase/client/blob/3e6b7e6ebd9598f34d53c6d7c3034009b5368d0c/shared/styles/index.native.js#L51

doesn't seem to consider iPhoneX yet, so I added it to the new commit.

songgao: (talked on chat; rebased over nojima's branch)

shared/router-v2/shim.native.js
```diff
@@ -11,18 +11,23 @@ const shimNewRoute = (Original: any) => {
...
           behavior={Styles.isIOS ? 'padding' : undefined}
+          keyboardVerticalOffset={(navigationOptions && navigationOptions.headerHeight) || undefined}
         >
           {body}
         </Kb.NativeKeyboardAvoidingView>
       )
 
       // don't make safe areas
-      if (Original.navigationOptions && Original.navigationOptions.underNotch) {
+      if (navigationOptions && Original.navigationOptions.underNotch) {
```
chrisnojima: should be navigationOptions.underNotch

PR #17252, opened 2019-04-30
loading state for friends tabs + fix for header colors

chrisnojima:
- [x] This changes the background color behavior on profiles. Now the color is stuck to the top always
- [x] The search bar uses the white form
- [x] The back button on profiles is white
- [x] Correct loading state for loading up followers instead of showing it as zero


chrisnojima:
cc: @cecileboucheron

shared/profile/user/index.js
```diff
@@ -39,7 +39,7 @@ export type Props = {|
   username: string,
 |}
 
-const colorTypeToStyle = type => {
+export const colorTypeToStyle = (type: 'red' | 'green' | 'blue') => {
```
buoyad: `BackgroundColorType`

shared/profile/user/index.js
```diff
@@ -299,21 +293,21 @@ class User extends React.Component<Props, State> {
         />
       )
     }
-    const loading = this.props.followersCount == null || this.props.followingCount == null
+    const loading = !this.props.followers || !this.props.following
```
buoyad: I see `Followers (null)` flash in during profile load.

shared/profile/user/index.js
chrisnojima: actually not exported anymore so i don't need the type

shared/profile/user/index.js
```diff
@@ -350,13 +344,20 @@ class User extends React.Component<Props, State> {
...
     let chunks = this.state.width ? chunk(friends, itemsInARow) : []
-    if (chunks.length === 0 && this.props.followingCount !== null && this.props.followingCount !== null) {
-      chunks.push({
-        text: this.state.selectedFollowing
-          ? `${this.props.userIsYou ? 'You are' : `${this.props.username} is`} not following anyone.`
-          : `${this.props.userIsYou ? 'You have' : `${this.props.username} has`} no followers.`,
-        type: 'noFriends',
-      })
+    if (chunks.length === 0) {
+      if (this.props.following && this.props.following) {
```
MarcoPolo: ?

chrisnojima: should be followers, copy and paste of the incorrect code above on line 353

PR #17254, opened 2019-04-30
inbox index percentage, bug fixes

joshblum:
- calculate the index percentage across the whole inbox instead averaging across conversations
- fix `--reforce-reindex`
- check context canceling in outer search stages since we sometimes just log errors


joshblum:
merging over kbweb flakes

PR #17258, opened 2019-04-30
Add APISessionTypeOPTIONAL to user/lookup calls

zapu:
If the server checks for readability of users in user/lookup calls, we need to present a valid session so API can serve us deleted users we are eligible to see.


zapu:
Do not merge - do not use OPTIONAL here, do something else instead.

Work in progress.

PR #17262, opened 2019-04-30
add bio info to uidmapper/inboxuiitem

joshblum:
- adds bio info to chat participants to display in 1-1 chats.
- depends on https://github.com/keybase/keybase/pull/3789 for server changes

cc @mmaxim


protocol/avdl/keybase1/common.avdl
```diff
@@ -328,6 +328,7 @@ protocol Common {
   record FullNamePackage {
     FullNamePackageVersion version;
     FullName fullName;
+    string bio;
```
zapu: will it still be able to use existing on-disk cache? It should just load FullNamePackages with `bio` being empty, right?

mmaxim:
I actually didn’t really want to do this ticket like this. I think it would be a lighter weight change to just have an RPC that hits the server every time that the UI loads async, lets discuss

zapu:
Sgtm, separate endpoint sounds good

PR #17265, opened 2019-04-30
make Files tab header align with others

songgao:


shared/styles/index.native.js
```diff
@@ -48,7 +48,7 @@ export const globalStyles = {
   ...util,
 }
 
-export const statusBarHeight = iPhoneXHelper.getStatusBarHeight()
+export const statusBarHeight = iPhoneXHelper.getStatusBarHeight(true)
```
songgao: This needs to be `true` for it to return `44` instead of `30`.

songgao:
Thanks!

PR #17268, opened 2019-05-01
offline data plumbing

songgao:
1. use the new API for fetching SyncConfig for all sync-enabled TLFs
2. drive GUI data reloads using the overall notifications API


shared/actions/fs/index.js
```diff
@@ -54,52 +54,58 @@ const getSyncConfigFromRPC = (tlfName, tlfType, config: RPCTypes.FolderSyncConfi
...
-      .map(tlfType => tlfListToGetSyncConfigPromise(state, tlfType))
-      // $FlowIssue hasn't learnt about .flat yet.
-      .flat()
-  ).then(results => {
-    const payloadMutable = results.reduce(
-      (payload, {syncConfig, tlfType, tlfName}) => ({
-        ...payload,
-        [tlfType]: payload[tlfType].set(tlfName, syncConfig),
-      }),
+  RPCTypes.SimpleFSSimpleFSSyncConfigAndStatusRpcPromise().then(({folders}) => {
```
jakob223: Maybe we should also set all the other TLFs as "not synced" here?

shared/actions/fs/index.js
```diff
@@ -54,52 +54,58 @@ const getSyncConfigFromRPC = (tlfName, tlfType, config: RPCTypes.FolderSyncConfi
...
         private: I.Map().asMutable(),
         public: I.Map().asMutable(),
         team: I.Map().asMutable(),
       }
     )
     return FsGen.createTlfSyncConfigsLoaded({
-      private: payloadMutable.private.asImmutable(),
-      public: payloadMutable.public.asImmutable(),
-      team: payloadMutable.team.asImmutable(),
+      private: mutable.private.asImmutable(),
```
jakob223: why the rename? I thought the old name was more descriptive.

shared/fs/row/load-files-when-needed.js
```diff
@@ -9,52 +9,67 @@ type OwnProps = {|
...
-  loadFolderList,
+const mergeProps = (s, d, {path}) => ({
+  ...s,
+  ...d,
   path,
 })
 
 type Props = {|
-  loadFolderList: () => void,
+  loadFolderList: (setRefreshTag: boolean) => void,
```
jakob223: Instead of being a `boolean`, could this be a `'set-refresh-tag' | 'no-set-refresh-tag'` or something, so that the code calling `_load(true)` and such is more legible?

shared/actions/fs/index.js
songgao: Oh; I rewrote this function and didn't consciously rename it. Will change it back.

shared/fs/row/load-files-when-needed.js
songgao: I felt like it was fine as this is in the same file. But I guess it wouldn't hurt to define some constants inside the file.

jakob223: /shrug I don't have strong feelings.

PR #17273, opened 2019-05-01
Update font

cecileboucheron:
For some reason icons are all negative... Need to figure out why tomorrow.
image

cc @adamjspooner I went ahead and uploaded @aimee-keybase's changes to the font - hoping it will make it to the release.


adamjspooner:
This is okay now, right?

PR #17277, opened 2019-05-01
Put mastodon in sigchain links

mlsteele:
image

### DIFF
--- a/go/externals/proof_service_generic_social.go
+++ b/go/externals/proof_service_generic_social.go
@@ -269,7 +269,11 @@ func (t *GenericSocialProofServiceType) GetPrompt() string {
 }
 
 func (t *GenericSocialProofServiceType) ToServiceJSON(username string) *jsonw.Wrapper {
-	return t.BaseToServiceJSON(t, username)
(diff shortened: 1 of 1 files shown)


go/externals/proof_service_generic_social.go
```diff
@@ -269,7 +269,11 @@ func (t *GenericSocialProofServiceType) GetPrompt() string {
 }
 
 func (t *GenericSocialProofServiceType) ToServiceJSON(username string) *jsonw.Wrapper {
-	return t.BaseToServiceJSON(t, username)
+	ret := t.BaseToServiceJSON(t, username)
+	if strings.HasPrefix(strings.ToLower(t.DisplayGroup()), "mastodon") {
```
joshblum: should this use the group instead so new groups can add this as well?

mlsteele: Yeabut the group is "Mastodon instance". The group is doubling as the text to show in the UI. "mastodon" seems like a cleaner data value. I'm not too worried about another group

PR #17278, opened 2019-05-01
Move forgot pw autoreset suggestion from LoginProvisionedDevice to Login

pzduniak:


go/engine/login.go
```diff
@@ -103,6 +104,10 @@ func (e *Login) Run(m libkb.MetaContext) (err error) {
 	// First see if this device is already provisioned and it is possible to log in.
 	loggedInOK, err = e.loginProvisionedDevice(m, e.username)
 	if err != nil {
+		// Suggest autoreset if user failed to log in and we're provisioned
+		if _, ok := err.(libkb.PassphraseError); ok {
+			return e.suggestRecoveryForgotPassword(m)
+		}
```
maxtaco: print the error before entering this block (line 111)

patrickxb:
Is there a test for this?

pzduniak:
This particular flow is in fact not tested (as a non-authenticated pipeline status is the same as not being in the pipeline at all) but we're working towards full coverage, #17165 has some prerequisites for this.

patrickxb:
Ok, I'd strongly advocate for anything regarding login and provisioning to have tests before being merged into master.  

login_test.go is the largest test file in the engine directory because login failures are bad for the health of keybase.

patrickxb:
(if this fixes something that is about to be released, go ahead and merge it, but please make a test as soon as possible...)

PR #17291, opened 2019-05-01
fix search delegation

joshblum:
- adds `--conv` to inbox search cli
- fixes render bugs on search delegation with cli
- triggers delegate search if query < `search.MinTokenLength`
- fixes bugs around delegation for name searches and streaming back thread results
~- triggers `SelectiveSync` in RPC method instead of indexer so we can trigger/cancel a sync when the search is delegated~

outstanding on GUI: 
- [x] show inbox index percentage on hover
- [x] fix duplication of results on thread search in GUI


mmaxim:
Don't think so, I am fine having it run in thread search, but I don't see the point in the full inbox. It just makes everything slower when typing, and makes this code more complicated.

PR #17293, opened 2019-05-01
log before we close the ch

joshblum:
cc @heronhaye i think this will fix the test flake where we log after shutdown


joshblum:
merging over kbfs flake

PR #17294, opened 2019-05-01
Fix marking as read on desktop. notifications go to the right place

chrisnojima:
With the changes to how conversation selected works desktop mark as read was busted
Notifications will now take you to the chat root and select the right convo vs pushing anywhere


shared/actions/chat2/index.js
```diff
@@ -1153,13 +1153,14 @@ function* desktopNotify(state, action) {
         const onClick = () => {
           resolve(
             Saga.sequentially([
+              Saga.put(RouteTreeGen.createSwitchTab({tab: 'tabs.chatTab'})),
```
chrisnojima: switch, go to root, choose convo

shared/actions/chat2/index.js
```diff
@@ -2992,7 +2993,7 @@ function* chat2Saga(): Saga.SagaGenerator<any, any> {
       Chat2Gen.updateReactions,
       ConfigGen.changedFocus,
       ConfigGen.changedActive,
-      a => typeof a.type === 'string' && a.type.startsWith(RouteTreeGen.typePrefix),
```
chrisnojima: we used to watch all routing actions, now we just listen for a specific action

shared/actions/platform-specific/push.native.js
```diff
@@ -127,8 +127,9 @@ function* handleLoudMessage(notification) {
 
   const {conversationIDKey, unboxPayload, membersType} = notification
 
+  yield Saga.put(RouteTreeGen.createSwitchTab({tab: 'tabs.chatTab'}))
```
chrisnojima: switch and go to root

shared/actions/tracker2.js
```diff
@@ -193,7 +193,7 @@ const showUser = (_, action) => {
     reason: '',
   })
 
-  if (flags.useNewRouter) {
+  if (flags.useNewRouter && !action.payload.skipNav) {
```
chrisnojima: profile on remount would call reload which does this which can try and push a screen. that should never nav so add a param

shared/chat/inbox/container/index.js
```diff
@@ -37,8 +37,12 @@ const mapStateToProps = state => {
 }
 
 const mapDispatchToProps = dispatch => ({
+  // a hack to have it check for marked as read when we mount as the focus events don't fire always
   _onInitialLoad: (conversationIDKeys: Array<Types.ConversationIDKey>) =>
     dispatch(Chat2Gen.createMetaNeedsUpdating({conversationIDKeys, reason: 'initialTrustedLoad'})),
+  _onMountedDesktop: () => {
```
chrisnojima: ideally we'd use navigationevents to know when its shown but that seems to not fire consistently. TODO figure that out later. so on inbox mount on desktop we just do an extra check

shared/constants/chat2/index.js
```diff
@@ -201,7 +201,11 @@ export const isUserActivelyLookingAtThisThread = (
 
   let chatThreadSelected = false
   if (flags.useNewRouter) {
-    chatThreadSelected = true // conversationIDKey === selectedConversationIDKey is the only thing that matters in the new router
+    if (isMobile) {
+      chatThreadSelected = true // conversationIDKey === selectedConversationIDKey is the only thing that matters in the new router
+    } else {
+      chatThreadSelected = Router2.getVisibleScreen()?.routeName === 'chatRoot'
```
chrisnojima: on desktop in nav2 we do maintain the selected convo

chrisnojima:
@keybase/react-hackers

shared/router-v2/tab-bar/container.desktop.js
```diff
@@ -36,6 +37,9 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
     if (ownProps.selectedTab === Tabs.peopleTab && tab !== Tabs.peopleTab) {
       dispatch(PeopleGen.createMarkViewed())
     }
+    if (ownProps.selectedTab === Tabs.chatTab && tab !== Tabs.chatTab) {
```
buoyad: This is checking for when we switch away from the chat tab, intended?

chrisnojima: nope fixed

shared/actions/platform-specific/push.native.js
```diff
@@ -127,8 +127,9 @@ function* handleLoudMessage(notification) {
...
+  yield Saga.put(RouteTreeGen.createNavUpToScreen({routeName: 'chatRoot'}))
```
buoyad: There's a delay on the previously selected convo before it switches over to the new one (too long in dev, on a profile build it's not too bad). I think fine for now, but maybe switching this to do a `reset` later instead would help

PR #17297, opened 2019-05-01
libfuse: Fix logic to detect ignorable fusermount unmount error.

heronhaye:
I did this incorrectly initially. The error was just "exit status 1", I needed to look in stderr for the real error message from fusermount.

Also, this code kind of feels messy in general with the error checking and handling. Let me know if there's a better way and I'll fix it up.

### DIFF
--- a/go/kbfs/libfuse/mounter.go
+++ b/go/kbfs/libfuse/mounter.go
@@ -7,6 +7,7 @@
 package libfuse
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


go/kbfs/libfuse/mounter.go
```diff
@@ -92,28 +93,36 @@ func fuseMountDir(dir string, platformParams PlatformParams) (*fuse.Conn, error)
...
 	// Try normal unmount
 	switch runtime.GOOS {
 	case "darwin":
 		_, err = exec.Command("/sbin/umount", dir).Output()
 	case "linux":
-		_, fusermountErr := exec.Command("fusermount", "-u", dir).Output()
+		fusermountOutput, fusermountErr := exec.Command("fusermount", "-u", dir).CombinedOutput()
 		// Only clean up mountdir on a clean unmount.
 		if fusermountErr == nil {
+			m.log.Info("Successfully unmounted.")
```
strib: We don't end log messages or errors with punctuation in KBFS (here and two places below).

heronhaye: fixed. So the rule is no punctuation for message and errors, but yes punctuation for comments? Is there a rule about capitalization as well?

PR #17299, opened 2019-05-01
cr_chains: when identifying types, handle self-contained subdirs

strib:
If a directory was both created and deleted within a branch, and a file within that directory was created, set-attr'd, and deleted within the branch as well, we had a bug where CR couldn't complete because it choked on trying to find the deleted file's entry in the deleted parent block, and the parent block couldn't be fetched from anywhere.

In this case, just assume the file has been deleted, and return early.

Also adds a test that repros the failure.

Finally, tracking down this issue was made harder by the fact that the logs contained error messages for seemingly-normal block IDs, so I burned a lot of time trying to figure out if the blocks were somehow deleted from the servers.  So this PR also makes temporary block IDs (the ones that persist in the `setAttrOp` of the removed file) obvious by giving them an invalid ID prefix.

Issue: KBFS-4114

### DIFF
--- a/go/kbfs/kbfsblock/id.go
+++ b/go/kbfs/kbfsblock/id.go
@@ -100,19 +100,30 @@ func (id *ID) HashType() kbfshash.HashType {
 	return id.h.GetHashType()
 }
 
-// MakeTemporaryID generates a temporary block ID using a CSPRNG. This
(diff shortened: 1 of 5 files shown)


strib:
Merging over unrelated libkb failures.

PR #17304, opened 2019-05-02
team mention autocomplete

mmaxim:
@malgorithms works like we discussed today.

### DIFF
--- a/shared/chat/conversation/input-area/normal/container.js
+++ b/shared/chat/conversation/input-area/normal/container.js
@@ -67,6 +67,7 @@ const mapStateToProps = (state, {conversationIDKey}: OwnProps) => {
     showWalletsIcon: Constants.shouldShowWalletsIcon(state, conversationIDKey),
     suggestChannels: Constants.getChannelSuggestions(state, teamname),
     suggestCommands: Constants.getCommands(state, conversationIDKey),
+    suggestTeams: Constants.getTeams(state),
     suggestUsers: Constants.getParticipantSuggestions(state, conversationIDKey),
     typing: Constants.getTyping(state, conversationIDKey),
     unsentText,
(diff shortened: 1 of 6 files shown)


shared/constants/chat2/meta.js
```diff
@@ -452,3 +452,10 @@ export const getConversationRetentionPolicy = (
 
 export const isDecryptingSnippet = (meta: Types.ConversationMeta) =>
   meta.trustedState === 'requesting' || meta.trustedState === 'untrusted'
+
+export const getTeams = (state: TypedState): Array<string> => {
+  const teamMap = state.chat2.metaMap.reduce((m, meta) => {
+    return meta.teamname ? m.set(meta.teamname, true) : m
```
chrisnojima: Just pass an array to reduce and push

shared/chat/conversation/input-area/normal/container.js
```diff
@@ -180,6 +181,9 @@ const mergeProps = (stateProps, dispatchProps, ownProps: OwnProps): Props => ({
   showWalletsIcon: stateProps.showWalletsIcon,
   suggestChannels: stateProps.suggestChannels,
   suggestCommands: stateProps.suggestCommands,
+  suggestTeams: stateProps.suggestTeams
```
chrisnojima: This likely should be memoized. Merge probably gets called a ton especially while typing

shared/constants/chat2/meta.js
mmaxim: This is to de-dup the names, since you can have multiple convs with same team name.

shared/chat/conversation/input-area/normal/container.js
mmaxim: Yeah I was concerned about this, although I don't think typing is a factor since this input is uncontrolled.

shared/constants/chat2/meta.js
chrisnojima: Can't you just use general

mmaxim: Ah yeah good point

PR #17307, opened 2019-05-02
In scankeys, add all synced secret key, not just one at random

zapu:
Non-determinism came from here: https://github.com/keybase/client/blob/master/go/libkb/sync_secret.go#L182 

`ss.keys.PrivateKeys` is a map of strings and iterating over it is randomized.

I added new methods like `GetSyncedSecretKeys` that have "signular" `~Key` methods equivalent because there are multiple flows that assumed only one key - including PGP provisioning :( - and I don't want to change all of this right now.

### DIFF
--- a/go/engine/scankeys.go
+++ b/go/engine/scankeys.go
@@ -59,7 +59,7 @@ func NewScanKeys(m libkb.MetaContext) (sk *ScanKeys, err error) {
 		return nil, fmt.Errorf("loadme error: %s", err)
 	}
 
-	// if user provided, then load their local keys, and their synced secret key:
+	// if user provided, then load their local keys, and their synced secret keys:
 	synced, err := sk.me.GetSyncedSecretKeys(m)
 	if err != nil {
 		return nil, fmt.Errorf("getsyncedsecret err: %s", err)
--- a/go/engine/scankeys.go
+++ b/go/engine/scankeys.go
@@ -60,7 +60,7 @@ func NewScanKeys(m libkb.MetaContext) (sk *ScanKeys, err error) {
 	}
 
 	// if user provided, then load their local keys, and their synced secret key:
-	synced, err := sk.me.GetSyncedSecretKey(m)
+	synced, err := sk.me.GetSyncedSecretKeys(m)
 	if err != nil {
 		return nil, fmt.Errorf("getsyncedsecret err: %s", err)
 	}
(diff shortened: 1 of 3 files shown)


zapu:
Fixes https://github.com/keybase/keybase-issues/issues/3373

0xdevalias:
Any idea when this might land in a stable release channel?

StephanX:
I also just encountered this.  Using version `4.7.2-20191026044208+6fc2e969b4 `

zapu:
Hey, is this still broken for you @StephanX ? Can you send a log using `keybase log send` command?

zapu:
along with more info e.g. what action have you tried and which key you expected to work. If it's private, and you are willing to only share with keybase, you can put it in as "log feedback". thank you.

StephanX:
Sure, logid: 2be52c7ca739c21bf406b61c 

I created a gpg key `5D030726DF141AE3` a few days ago, using the mac version of gpgtools.org , version:
```
gpg (GnuPG) 2.2.17
libgcrypt 1.8.5
```
Perhaps of note, I created a new UID version of the original key to change the name and password (ref: https://coderwall.com/p/tx_1-g/gpg-change-email-for-key-in-pgp-key-servers ) 

I then installed keybase for the first time today, also mac client version, and imported my pgp key.  Finally, I attempted to use keybase to decrypt a fyle with `keybase pgp decrypt` and received the error `ERROR decrypt error: unable to find a PGP decryption key for this message`

Happy to offer more debugging info, thanks for the quick response!

zapu:
Hey, sorry for not getting back to you, I was out of office.

Looks like this key was revoked and another one was added: https://keybase.io/stephanx/sigchain#42f610201df16afe14daa8bc838a009511bf031625c166f59a6dae194a8e8b8d0f

What's your status now? PGP commands are only available for keys that are published on your Keybase user. If you have a key that's not attached to Keybase profile, it's better to just use GPG program.

Let me know if I misunderstood what's going here.

joe-niland:
Hi @zapu 

I have the same issue as @StephanX I think.

For example running this gets the same error as mentioned above:

```
keybase pgp encrypt -m "message" | keybase pgp decrypt
▶ ERROR decrypt error: unable to find a PGP decryption key for this message
```

How can I figure out where it's going wrong?

Thanks
Joe

maxtaco:
@joe-niland likely you don't have the private key on the local machine.

joe-niland:
Thanks @maxtaco

It is in my GPG keychain but I can't remember if/how I imported it into keybase.

Is it a case of running `keybase select --import`?

I will try this later.

PR #17311, opened 2019-05-02
fix header title styling

chrisnojima:
@keybase/react-hackers 
note if you just want a string just use title

### DIFF
--- a/shared/settings/index.js
+++ b/shared/settings/index.js
@@ -65,9 +65,7 @@ const Connected = compose(
 
 // $FlowIssue lets fix this
 Connected.navigationOptions = {
-  header: undefined,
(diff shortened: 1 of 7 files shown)


shared/settings/index.js
```diff
@@ -65,9 +65,7 @@ const Connected = compose(
 
 // $FlowIssue lets fix this
 Connected.navigationOptions = {
```
buoyad: There should still be a blank header right?
image

chrisnojima: i thought it looked better w/o this blank space. @cecileboucheron ?

chrisnojima:
@buoyad i accidentally pushed the android fix here, cecile is fine with the settings change

PR #17318, opened 2019-05-02
Fix block conversation routing

buoyad:
The problem was `findFullRoute` assumed there could only be one modal. r? @keybase/react-hackers


shared/chat/manage-channels/edit-channel.js
```diff
@@ -107,7 +107,7 @@ class _EditChannel extends React.Component<Props, State> {
           />
         </Box>
         <Box style={_bottomRowStyle}>
-          {!isMobile && this.props.showDelete && (
+          {!isMobile && this.props.showDelete && !this.props.deleteRenameDisabled && (
```
buoyad: Unrelated fix. You could try to delete `#general` and get a black bar because of this

PR #17321, opened 2019-05-02
get the device ID from active device, and make sure to use the correct UID

mmaxim:


go/service/gregor.go
```diff
@@ -749,10 +755,12 @@ func (g *gregorHandler) OnConnect(ctx context.Context, conn *rpc.Connection,
 	if err != nil {
 		return fmt.Errorf("failed to get gregor client: %s", err.Error())
 	}
-	uid, token, nist, err := g.authParams(ctx)
+	uid, deviceID, token, nist, err := g.authParams(ctx)
 	if err != nil {
 		return err
 	}
+	gcli.User = uid
+	gcli.Device = deviceID
```
maxtaco: Can't multiple threads have control of this object at the same time? And we don't have a lock here, so it might race.

mmaxim: Can't be any worse.

maxtaco: make CORE ticket to fix this by always passing through UID/deviceID

PR #17322, opened 2019-05-02
Fix linux autorestarting issues

heronhaye:


packaging/linux/systemd/keybase.gui.service
```diff
@@ -33,6 +33,10 @@ ExecStartPost=/usr/bin/env systemctl --user unset-environment KEYBASE_AUTOSTART
 
 # We don't set Restart=, since a GUI crash is probably visible to the user.
 
+# Only kill main Electron process, which kills the others. But if that fails,
+# systemd will send a SIGKILL to all Electron processes.
+KillMode=process
```
heronhaye: systemd was sending a SIGTERM to every electron process at the same time, and it did not like that, occasionally crashing (and since we don't set restart=always for gui, it didn't restart successfully). This change makes it so it only SIGTERMs the main process and then SIGKILLs everything later if it doesn't work.

heronhaye: This still does not explain why KBFS was not starting up but that seems better with this change somehow. Will do some more testing.

shared/desktop/app/node.desktop.js
```diff
@@ -116,8 +116,12 @@ const handleCrashes = () => {
       })
 
       if (win.webContents) {
-        win.webContents.on('crashed', e => {
-          console.log('Browser window crashed: ', e)
+        win.webContents.on('crashed', (e, killed) => {
```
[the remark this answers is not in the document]
heronhaye: ref: https://github.com/electron/electron/issues/7591

packaging/linux/run_keybase
```diff
@@ -56,12 +56,12 @@ start_systemd() {
   # systemctl calls don't have it set.
   systemctl --user import-environment "KEYBASE_AUTOSTART"
 
-  # The keybase.gui.service unit has keybase.service and kbfs.service as
-  # dependencies, so we don't have to list them here. But including them lets
-  # us report an error if they fail to start. Also prefer `restart` to `start`
-  # so that we don't race against the service shutting down.
-  # kbfs.service will attempt to start the redirector, but it isn't an error if
-  # it fails to start, which happens if it is disabled.
+  # The keybase.gui.service unit has keybase.service as dependencies, so we
```
strib: "as dependencies" -> "as a dependency"

packaging/linux/run_keybase
```diff
@@ -56,12 +56,12 @@ start_systemd() {
...
+  # don't have to list them here. But including them lets us report an error if
```
strib: s/them/it/g

packaging/linux/systemd/keybase.gui.service
cjb: Why don't we set `restart=always`?

heronhaye: Jack's decision, "since a GUI crash is probably visible to the user."

I think I agree as well, not 100% certain

heronhaye:
merging w/ cnojima's lgtm

packaging/linux/systemd/keybase.gui.service
heronhaye: cc @eli-schwartz who may be interested, I think arch community uses the same unit files

eli-schwartz: GUI crashes will at least mean the icon in the corner disappears, yes?

Restart=always is completely wrong anyhow, it means if you click "Quit", then because the program exited successfully but it wasn't `systemctl stop` which stopped it, it gets restarted. Restart=on-failure would cause it to restart on crashes and other abnormal exits, if that is something considered to be desirable.

PR #17326, opened 2019-05-02
Fix adding yourself to a subteam with new role picker

cjb:
@keybase/react-hackers CC @MarcoPolo 

Adding yourself to a subteam wasn't possible, because getDisabledReasonsForRolepicker noticed that you aren't an admin and refused to show you the options.  In the implicit admin case of a subteam, you don't have any role on the team at all, but you're still allowed to manage members.

This PR makes it so that the ultimate test of whether you can perform the add comes from the service (via canUserPerform.manageMembers).  If manageMembers says you can't, then the existing code seems correct to explain why to the user.

### DIFF
--- a/shared/constants/teams.js
+++ b/shared/constants/teams.js
@@ -327,6 +327,11 @@ const getDisabledReasonsForRolePicker = (
   teamname: Types.Teamname,
   memberToModify: ?string
 ): Types.DisabledReasonsForRolePicker => {
+  const canManageMembers = getCanPerform(state, teamname).manageMembers
+  if (canManageMembers) {
+    // If you're an implicit admin, the tests below will fail for you, but you can still change roles.
+    return {}
+  }
   const members = getTeamMembers(state, teamname)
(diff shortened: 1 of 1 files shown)


shared/constants/teams.js
```diff
@@ -327,6 +327,11 @@ const getDisabledReasonsForRolePicker = (
   teamname: Types.Teamname,
   memberToModify: ?string
 ): Types.DisabledReasonsForRolePicker => {
+  const canManageMembers = getCanPerform(state, teamname).manageMembers
+  if (canManageMembers) {
+    // If you're an implicit admin, the tests below will fail for you, but you can still change roles.
+    return {}
```
buoyad: If this is a subteam, `"Subteams cannot have owners."`

buoyad: Guess this applies to the rest of this function too. Want to split into a separate ticket?

cjb:
@buoyad Thanks, how's this look?

PR #17328, opened 2019-05-02
fix headers on signup / provision

chrisnojima:


shared/login/signup/common.js
```diff
@@ -31,13 +30,7 @@ export const Wrapper = (props: Props) => (
...
       {props.children}
     </Box2>
-    {flags.useNewRouter ? (
-      <Box style={styles.header}>
-        <HeaderHocHeader onBack={props.onBack} headerStyle={styles.header} />
-      </Box>
-    ) : (
-      <HeaderHocHeader onBack={props.onBack} headerStyle={styles.header} />
-    )}
+    {!flags.useNewRouter && <HeaderHocHeader onBack={props.onBack} headerStyle={styles.header} />}
```
chrisnojima: header is in the router and not on these screens now

shared/provision/code-page/index.js
```diff
@@ -125,7 +125,7 @@ class CodePage2 extends React.Component<Props, State> {
               }
             />
           </Kb.Box2>
-          {(!flags.useNewRouter || Styles.isMobile) && (
+          {(!flags.useNewRouter || this.props.currentDeviceAlreadyProvisioned) && (
```
chrisnojima: only need the header if we're coming in from add device

shared/provision/set-public-name/index.js
```diff
@@ -15,7 +15,7 @@ type Props = {|
 
 const SetPublicName = (props: Props) => {
   return (
-    <Box2 direction="vertical" fullWidth={true} fullHeight={true} gap="medium">
+    <Box2 direction="vertical" fullWidth={true} fullHeight={true} gap="medium" gapStart={true}>
```
chrisnojima: not enough space on top

shared/router-v2/router.native.js
```diff
@@ -153,7 +153,20 @@ const LoggedInStackNavigator = createStackNavigator(
 const LoggedOutStackNavigator = createStackNavigator(
   {...Shim.shim(loggedOutRoutes)},
   {
-    defaultNavigationOptions,
+    defaultNavigationOptions: {
```
chrisnojima: make the logged out stack have cancel buttons so signup / provision work correctly

chrisnojima:
@keybase/react-hackers

PR #17329, opened 2019-05-02
libkbfs init: behave nicer when levelDBs won't open

jakob223:
- CR DB can be null now
 - disk metadata store replaced with noop DB when busy

Issue: KBFS-4107

### DIFF
--- a/go/kbfs/libkbfs/config_local.go
+++ b/go/kbfs/libkbfs/config_local.go
@@ -1110,8 +1110,10 @@ func (c *ConfigLocal) Shutdown(ctx context.Context) error {
 	if bms != nil {
 		bms.Shutdown()
 	}
-	if err := c.conflictResolutionDB.Close(); err != nil {
-		errorList = append(errorList, err)
+	if c.conflictResolutionDB != nil {
+		if err := c.conflictResolutionDB.Close(); err != nil {
+			errorList = append(errorList, err)
+		}
 	}
 	kbfsServ := c.kbfsService
 	if kbfsServ != nil {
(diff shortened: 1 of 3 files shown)


go/kbfs/libkbfs/config_local.go
```diff
@@ -1423,7 +1425,9 @@ func (c *ConfigLocal) MakeBlockMetadataStoreIfNotExists() (err error) {
 	}
 	c.blockMetadataStore, err = newDiskBlockMetadataStore(c)
 	if err != nil {
-		// TODO: open read-only instead KBFS-3659
+		// TODO (KBFS-3659): when we can open levelDB read-only,
+		//  do that instead of returning a Noop version
```
strib: End comment in period.

go/kbfs/libkbfs/init.go
```diff
@@ -835,17 +835,12 @@ func doInit(
...
-		//                   this, instead of failing the init.
-		/*
-			notification := &keybase1.FSNotification{
-				StatusCode:       keybase1.FSStatusCode_ERROR,
-				NotificationType: keybase1.FSNotificationType_INITIALIZED,
-				ErrorType:        keybase1.FSErrorType_DISK_CACHE_ERROR_LOG_SEND,
-			}
-			defer config.Reporter().Notify(ctx, notification)
-		*/
+		notification := &keybase1.FSNotification{
```
strib: I don't think we need the notification -- this will lead to a confusing toast popup for the user.  I think you can keep it commented out.

PR #17338, opened 2019-05-02
More linux autorestart fixes CORE-10818

heronhaye:
You suggested this when I wrote this code, but couldn't do it due to another bug in the service. I fixed the service issue a while ago, and adding it seems to makes the restart better:

I did a lot more debugging and it seems that before postinstall even restarts the services with systemd, something else (very mysterious) is shutting keybase.gui and kbfs down outside of systemd. That exits cleanly but later when systemctl goes to restart, they are already stopped so won't be restarted.

It doesn't happen with --direct.

But it's a very odd failure mode, if I build the package and then just install it repeatedly, or run post_install.sh manually, everything works perfectly. It's only when I build and install with the same command that it fails. I think maybe it's affected by the increased CPU usage relating to compiling the app.

### DIFF
--- a/packaging/linux/post_install.sh
+++ b/packaging/linux/post_install.sh
@@ -106,9 +106,6 @@ safe_restart_systemd_services() {
         fi
 
         if systemd_unit_active_for "$user" "kbfs.service"; then
-            # TODO: CORE-9789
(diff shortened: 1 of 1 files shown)


heronhaye:
going to merge this

PR #17339, opened 2019-05-02
prefetcher: send status notification right away when out of space

strib:
When the sync cache is out of space and prefetching has stopped, send an immediate sync status notification with the `OutOfSyncSpace` field set to true.

Also populate it on polling requests in simplefs, and show it in `fs sync show`.

There was also a bug in the `DiskBlockCache` that wasn't checking the space usage correctly for the sync cache.  The `backpressureDiskLimiter` doesn't actually track the bytes used by the sync cache, because it never wants to reject bytes from being saved.  So `DoesCacheHaveSpace` wasn't working correctly for the sync cache, since the limit reported by the tracker was always 0.  Instead use the reported max limit from the tracker to calculate the current usage.

Issue: KBFS-4121

### DIFF
--- a/go/kbfs/libkbfs/util.go
+++ b/go/kbfs/libkbfs/util.go
@@ -336,3 +336,44 @@ func cleanOldTempStorageRoots(config Config) {
 		log.CDebugf(ctx, "Done cleaning old storage roots")
 	}
 }
+
+// GetLocalDiskStats returns the local disk stats, according to the
+// disk block cache.
+func GetLocalDiskStats(ctx context.Context, dbc DiskBlockCache) (
+	bytesAvail, bytesTotal int64) {
+	if dbc == nil {
+		return 0, 0
+	}
+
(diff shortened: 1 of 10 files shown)


go/kbfs/libkbfs/util.go
```diff
@@ -336,3 +336,44 @@ func cleanOldTempStorageRoots(config Config) {
...
+			int64(status.LocalDiskBytesTotal)
+	}
+	return 0, 0
+}
+
+// FillInDiskSpaceStatus fills in the `OutOfSyncSpace` and local disk
+// space fields of the given status.  `status.PrefetchStatus` should
+// be populated before this function is called.
+func FillInDiskSpaceStatus(
+	ctx context.Context, status *keybase1.FolderSyncStatus,
```
jzila: Super nitpicky, but you could enforce the precondition by explicitly passing in `PrefetchStatus` to this function.

PR #17341, opened 2019-05-03
inject meta from new conversation into map

mmaxim:
Hidden conversations can't be restored with new conv now, this fixes that.


mmaxim:
KBFS failure in CI

PR #17344, opened 2019-05-03
fix showing inbox when coming from push / previous load

chrisnojima:
@keybase/react-hackers this fixes you seeing the inbox when you start up and land on a prev convo or come from a push
I added a new action for this. I do want to clean this action and the switch route def (already ticketed) when we deprecate nav1

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -380,7 +380,7 @@ const onIncomingMessage = (state, incoming) => {
 }
 
 // Helper to handle incoming inbox updates that piggy back on various calls
-const chatActivityToMetasAction = (payload: ?{+conv?: ?RPCChatTypes.InboxUIItem}) => {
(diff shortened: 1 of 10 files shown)


buoyad:
1. Be at inbox > convo > info panel & background
2. Tap push
3. End up at inbox > convo > convo

chrisnojima:
@buoyad updated
cc: @mmaxim i made it so when we create a new convo , we never make a meta delete. if its a brand new convo we'd make it and then immediately delete

mmaxim:
I'm a touch worried about that change, I can't remember if the service filters out empty conversations before you get to that point. If it doesn't, then you could be showing random blank convos that people are creating against you.

chrisnojima:
strangely if you go through the new team building flow it seems to never happen but if you go through the profile chat it does. i'll have to see why

PR #17346, opened 2019-05-03
fix convert to team flow

chrisnojima:
@keybase/react-hackers this fixes going from an adhoc chat to a team


chrisnojima:
@cjb per chat lets not even bother with nav1 anymore. it's going away and we're not going to test it anymore

PR #17351, opened 2019-05-03
Fix test flakes

joshblum:
we weren't shutting down chat modules in systests causing panics if anything was logged post test.


go/chat/inboxsource.go
```diff
@@ -479,33 +479,34 @@ func (s *HybridInboxSource) Clear(ctx context.Context, uid gregor1.UID) error {
...
+	defer s.Trace(ctx, func() error { return nil }, "Start")()
 	s.baseInboxSource.Start(ctx, uid)
 	s.Lock()
 	defer s.Unlock()
-	s.Debug(ctx, "Start")
 	if s.started {
 		return
 	}
-	s.stopCh = make(chan struct{})
+	s.stopCh = make(chan chan struct{}, 10)
```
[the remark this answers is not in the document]
joshblum: 👍 good point

go/libkb/constants.go
```diff
@@ -722,3 +721,5 @@ const (
 )
 
 const ProfileProofSuggestions = true
+
+const ShutdownChanDefaultSize = 10
```
jzila: But more specifically, why `10`? Also you replaced some numbers smaller than 10 with this.

joshblum: 10 was arbitrary but I'm going to replace the recent changes to address the flakes with the new pattern we discussed anyway.

joshblum:
we also were not waiting for id3 to shutdown: https://ci.keyba.se/job/client/job/PR-17390/3/execution/node/3182/log/

joshblum:
merging over CI... systests have been timing out a lot recently i don't think this change is hurting it but will keep an eye on things

PR #17354, opened 2019-05-03
DESKTOP-9594: Fix broken upload menu

jzila:


songgao:
Seems like this broke Darwin and iOS. This line (and the Darwin one above) needs to be changed too: https://github.com/keybase/client/blob/15ba88bc0deb360808a32f2f28d9e55738a1bd32/shared/fs/common/upload-button.js#L24

PR #17358, opened 2019-05-03
Fix showing new display currency after a change

cjb:
@keybase/react-hackers 

Not sure why this broke, though this change works for me.  The dropdown is controlled through Redux, and Redux only updates when we get new data from the service, and nothing was causing that to happen.  I suppose it might be the case that there used to be a service notification that triggered our refresh that isn't there anymore?

### DIFF
--- a/shared/actions/wallets.js
+++ b/shared/actions/wallets.js
@@ -506,7 +506,7 @@ const changeDisplayCurrency = (state, action) =>
       currency: action.payload.code, // called currency, though it is a code
     },
     Constants.changeDisplayCurrencyWaitingKey
-  )
+  ).then(_ => WalletsGen.createLoadDisplayCurrency({accountID: action.payload.accountID}))
 
 const changeAccountName = (state, action) =>
   RPCStellarTypes.localChangeWalletAccountNameLocalRpcPromise(


buoyad:
I think we should be getting an `accountDetailsUpdate`

cjb:
@patrickxb @mlsteele It looks like when we call `changeDisplayCurrencyLocal()` we receive an `notify.accountDetailsUpdate` immediately afterwards, but that notification contains the currency pre-change rather than post-change.  Seems like a regression.  Calling `getDisplayCurrencyLocal` afterwards returns the post-change one, so that's the workaround in this PR.

This is a release blocker, so we could either wait for an RPC fix ASAP, or merge this workaround for the release.  Take a look and let us know?

patrickxb:
I would be surprised if there has been any change here, I don't know of any code that has gone in to change it.  But I'll take a look right now.

patrickxb:
We have a fix incoming.  It is not a regression, the code hasn't changed since January.  But we think there is still a frontend bug.

@zapu has details.

I would not block any release on this...seems very minor.

cjb:
@zapu what's the frontend bug?

I think we're treating it as blocker because it makes it look like the set currency option doesn't work at all.

PR #17362, opened 2019-05-04
fix keyboard hiding

chrisnojima:
@keybase/react-hackers 
cc: @buoyad 
turns out the floating teams picker uses floating box and not overlay so i moved the keyboard hiding down a level and added a flag


chrisnojima:
@keybase/react-hackers bump

PR #17368, opened 2019-05-06
be more conservative about resolving mentions

mmaxim:
The point of this PR is to use local caches (and sender information) to resolve @ mentions. If we can't do it, show UI to ask the user to do it explicitly.


go/uidmap/hardcoded_uids.go
```diff
@@ -21,7 +21,7 @@ var usernameSortOrder []uint16
 var usoOnce sync.Once
 
 func findInit() {
-	once.Do(func() {
+	usoOnce.Do(func() {
```
maxtaco: can you use `once` here and `usoOnce` in the other init?  `uso` = "usernameSortOrder"

go/uidmap/hardcoded_uids.go
```diff
@@ -21,14 +21,14 @@ var usernameSortOrder []uint16
...
-	usoOnce.Do(func() {
+	once.Do(func() {
 		initUsernameOffsets()
 		initUsernames()
 	})
 }
 
 func usernameFindInit() {
-	once.Do(initUsernameSort)
+	usoOnce.Do(initUsernameSort)
```
maxtaco: Thanks!!!

mmaxim:
cc @malgorithms @heronhaye

chrisnojima:
- [ ] after release

go/chat/inboxsource.go
```diff
@@ -845,6 +849,20 @@ func (s *HybridInboxSource) Search(ctx context.Context, uid gregor1.UID, query s
 	return res, nil
 }
 
+func (s *HybridInboxSource) IsTeam(ctx context.Context, uid gregor1.UID, item string) (res bool, err error) {
+	defer s.Trace(ctx, func() error { return err }, "IsTeam")()
+	_, convs, err := s.createInbox().ReadAll(ctx, uid, true)
```
joshblum: why not use `Read` with `MembersTypes`  set? otherwise this bypasses the query caching right?

go/uidmap/hardcoded_uids.go
joshblum: if we're merging this PR after the release should we pick this change out?

go/chat/inboxsource.go
mmaxim: No this just gives whatever we have in cache back.

go/uidmap/hardcoded_uids.go
mmaxim: No, this is the only thing that uses it.

PR #17369, opened 2019-05-06
fix iOS/macOS Files tab where upload popups spontaneously

songgao:


songgao:
> so we aren't creating a new function each time one of these is rendered.

It'd actually be the same. Whenever `mergeProps` is called, `render` is called. So in terms of number of functions it'd be the same. But agree this should better be done in the container as the component should just display things. Will push a follow-up commit.

PR #17382, opened 2019-05-06
CachedContactsProvider, integrate with contacts/lookup

zapu:
Support new `contacts/lookup` endpoint that's being reviewed here: https://github.com/keybase/keybase/pull/3819

Most of the tests are "offline" tests that use mock providers. Testing for the endpoint itself happens in kbweb tests. There is one systest to see if everything is wired up properly but it doesn't check all possible flows, just the happy path.

Add new `CachedContactsProvider` that saves resolutions in leveldb map.


go/contacts/contacts.go
```diff
@@ -35,96 +29,83 @@ func ResolveContacts(mctx libkb.MetaContext, provider ContactsProvider, contacts
 		return res, nil
 	}
 
-	// Collect phone numbers and emails from Contact list. Iterate through
```
zapu: This function (`ResolveContacts`) is almost entirely new - it's best to just go to that file and look rather than looking at the diff.

go/contacts/cache.go
```diff
@@ -12,13 +15,121 @@ type CachedContactsProvider struct {
...
+
+type CachedLookupResult struct {
+	ContactLookupResult
+	Resolved bool
+	cachedAt time.Time
+}
+
+type ContactsCache struct {
+	Lookups map[string]CachedLookupResult
+	Version struct {
```
heronhaye: what is minor? backwards compatible changes?

go/contacts/cache.go
```diff
@@ -12,13 +15,121 @@ type CachedContactsProvider struct {
...
+	if len(emails)+len(numbers) == 0 {
+		return res, nil
+	}
+
+	now := mctx.G().Clock().Now()
+
+	var conCache ContactsCache
+	cacheKey := libkb.DbKey{
+		Typ: libkb.DBContactResolution,
+		Key: fmt.Sprintf("%v-%s", mctx.CurrentUID(), userRegion),
```
heronhaye: why do we need either of these things? in case user region changes?

heronhaye: could blast away the old one

go/contacts/lookup.go
```diff
@@ -0,0 +1,80 @@
...
+		lookups = append(lookups, lookupArg{PhoneNumber: string(v)})
+	}
+	for _, v := range emailsContacts {
+		lookups = append(lookups, lookupArg{Email: string(v)})
+	}
+
+	payload := make(libkb.JSONPayload)
+	payload["contacts"] = lookups
+	if !userRegionCode.IsNil() {
+		payload["user_region_code"] = userRegionCode
```
heronhaye: can you ask cjb if we can actually get this info from mobile apis?

go/libkb/db.go
```diff
@@ -20,6 +20,7 @@ const (
 	DBTeamChain         = 0x10
 	DBUserPlusAllKeysV1 = 0x19
 
+	DBContactResolution              = 0xb0
```
heronhaye: skipped 0xba?

go/service/contacts.go
```diff
@@ -121,5 +77,6 @@ func (h *ContactsHandler) LookupContactList(ctx context.Context, arg keybase1.Lo
 	provider := &contacts.CachedContactsProvider{
 		Provider: &bulkLookupContactsProvider{},
 	}
+	// provider := &bulkLookupContactsProvider{}
```
heronhaye: rm

go/libkb/db.go
zapu: I didn't know which one to use, they've all been pretty arbitrary

heronhaye: Just decrement by 1, I think the times we've skipped have been errors. Also might want to think about making this key 2 bytes soon

go/contacts/cache.go
zapu: yes, i felt inspired by something that Miles did in team cache where he could update team cache to new version using info that was already there: https://github.com/keybase/client/pull/17353/files#diff-4c4b8b19dca791c68066641d12b1df46R146

I don't know if it ever becomes useful though.

go/libkb/db.go
zapu: I remember now. i put it temporarily here and I was going to stuck it somewhere in the middle where there is space, like 0xc9. I don't think that matters a lot, though,  I'm changing it to 0xba.

heronhaye:
I'm still unsure about the DB key of uid-regioncode. The db is already per-user, and whatever frontend api gives us the regioncode might change for no reason. I think it could be better to hardcode it like "cache" or something.

Also, since this info is sensitive, what do you think about putting this in an encrypteddb?

LGTM otherwise

PR #17386, opened 2019-05-06
fs: FilenameText component

jakob223:
- make a `FilenameText` component that ellipsizes before the file extension
 - use it everywhere we have filenames ellipsizing
 - update stories

 - [x] either fix the gap between the ellipsis and the extension or decide it's okay
 - [x] release

Issue: KBFS-4099


shared/fs/common/filename-text.js
```diff
@@ -0,0 +1,76 @@
...
+    o.path ? Types.getPathName(o.path) : o.filename || ''
+  )
+  return {
+    fileExtension,
+    fileNameWithoutExtension,
+    selectable: o.selectable,
+    style: o.style,
+    type: o.type,
+  }
+}
```
songgao: Here you don't need state from the redux store, or dispatch any actions. So there's no need to have a redux container. You can just do this inside the component. It saves Redux from iterating over this container each time a state update happens, and saves us from having to have those story providers for this.

shared/constants/fs.js
```diff
@@ -1151,8 +1151,14 @@ export const makeActionsForShowSendAttachmentToChat = (
...
-export const splitFileNameAndExtension = (fileName: string) =>
-  ((str, idx) => [str.slice(0, idx), str.slice(idx)])(fileName, fileName.lastIndexOf('.'))
+export const splitFileNameAndExtension = (fileName: string) => {
+  const idx = fileName.lastIndexOf('.')
+  if (idx === -1 || fileName.length - idx > 10) {
+    return [fileName, '']
+  } else {
+    return [fileName.slice(0, idx), fileName.slice(idx)]
+  }
+}
```
songgao: It feels a bit weird to me to returns an array with second string being empty. Probably better to have an object or I.Record type for this. That being said, this is only used inside the component, so I think it should stay in that file and the function interface would not matter as much.

shared/fs/common/filename-text.js
```diff
@@ -0,0 +1,76 @@
...
+
+type Props = {|
+  fileExtension: string,
+  fileNameWithoutExtension: string,
+  selectable?: boolean,
+  style?: Styles.StylesCrossPlatform,
+  type: TextType,
+|}
+
+const FilenameText = (props: Props) => (
```
songgao: How about just `Filename`?

jakob223:
@songgao what do you think of this gap?
gap
I couldn't come up with any way to get rid of it.

songgao:
Thanks for the screencast! Meant to try myself but forgot. That seems to be a side-effect of character spacing being fixed. Note the space is only there when you have less than a character's space. I can't think of a good way to solve it. @cecileboucheron wdyt? Is it fine to leave it this way?

cecileboucheron:
That space's fine with me!

shared/local-debug.desktop.js
```diff
@@ -26,7 +26,7 @@ let config = {
   reduxSagaLoggerMasked: true, // Print saga debug info masked out
   showDevTools: false, // Show devtools on start
   skipAppFocusActions: false, // dont emit actions when going foreground/background, helpful while working on other actions stuff
-  skipSecondaryDevtools: true, // Don't show devtools for menubar/trackers etc
+  skipSecondaryDevtools: false, // Don't show devtools for menubar/trackers etc
```
songgao: Did you mean to flip this?

PR #17388, opened 2019-05-06
Fix creating an account from the send form

buoyad:
Also: use `resetStack` in `navigateToAccount` - it was originally ported over before we had `resetStack`. Currently it always pushes a wallet screen even if you were already on one. r? @keybase/react-hackers


buoyad:
Oh also fix a bug in `SendButton` where the `from` account ID wouldn't be set correctly because of some stale functions.

PR #17389, opened 2019-05-06
turn on inline requires

chrisnojima:
@keybase/react-hackers this turns on inline requires, this will be the default in the near future and they suggest trying it. it seems fine to me on ios and android


MarcoPolo:
What does inline requires do?

chrisnojima:
https://facebook.github.io/react-native/blog/#faster-app-launches-with-inline-requires

PR #17390, opened 2019-05-06
cleaner-config cli

joshblum:
add `keybase db cleaner-config` command to get/set the maximum cache size for the local databases. in action:

```
$ keybase db cleaner-config
main db current size: 371kB, configured max size: 2.0GB
chat db current size: 53kB, configured max size: 1.0GB
```


joshblum:
closing this out for now pending design discussions. Sounds like we want a unified setting across all client caches (lvldb, attachments, kbfs)

PR #17391, opened 2019-05-06
Retry pkg/check api call

mlsteele:


mlsteele:
Actually I don't think an error here would surface visibly to the user. So this is unnecessary. But still seems like an ok idea.

PR #17399, opened 2019-05-06
menubar: get KBFS status from widget ifneedbe

jakob223:
If we are not sure if we're connected to KBFS, ask instead of spinning forever.

Issue: DESKTOP-9627

### DIFF
--- a/shared/actions/fs/platform-specific.desktop.js
+++ b/shared/actions/fs/platform-specific.desktop.js
@@ -356,6 +356,7 @@ const openAndUpload = (state, action) =>
   )
 
 const loadUserFileEdits = (state, action) =>
+  state.fs.kbfsDaemonStatus.rpcStatus === 'connected' &&
   RPCTypes.SimpleFSSimpleFSUserEditHistoryRpcPromise().then(writerEdits =>
     FsGen.createUserFileEditsLoaded({
       tlfUpdates: Constants.userTlfHistoryRPCToState(writerEdits || []),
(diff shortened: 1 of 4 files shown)


shared/menubar/index.desktop.js
```diff
@@ -46,22 +47,23 @@ class MenubarRender extends React.Component<Props, State> {
...
   attachmentRef = React.createRef<Kb.Icon>()
 
-  _refreshUserFileEditsIfPossible = () =>
+  _refreshUserFileEditsOrWaitForKbfsDaemon = () =>
     this.props.loggedIn &&
-    this.props.kbfsDaemonStatus.rpcStatus === 'connected' &&
-    this.props.refreshUserFileEdits()
+    (this.props.kbfsDaemonStatus.rpcStatus === 'connected'
+      ? this.props.refreshUserFileEdits()
+      : this.props.waitForKbfsDaemon())
```
songgao: Note that this alone won't trigger the actual load. I think you also want `kbfsDaemonRpcStatusChanged` [here](https://github.com/keybase/client/blob/f3cca64128fe659bcdafa144720488b60a04de8e/shared/actions/fs/platform-specific.desktop.js#L394) so that once we're connected, we'd load it. Similar to [this](https://github.com/keybase/client/blob/f3cca64128fe659bcdafa144720488b60a04de8e/shared/actions/fs/platform-specific.desktop.js#L388-L390).

shared/actions/fs/platform-specific.desktop.js
```diff
@@ -391,7 +391,10 @@ function* platformSpecificSaga(): Saga.SagaGenerator<any, any> {
     )
   }
   yield* Saga.chainAction<FsGen.OpenAndUploadPayload>(FsGen.openAndUpload, openAndUpload)
-  yield* Saga.chainAction<FsGen.UserFileEditsLoadPayload>(FsGen.userFileEditsLoad, loadUserFileEdits)
+  yield* Saga.chainAction<FsGen.UserFileEditsLoadPayload | FsGen.KbfsDaemonRpcStatusChangedPayload>(
+    [FsGen.userFileEditsLoad, FsGen.kbfsDaemonRpcStatusChanged],
+    loadUserFileEdits
```
songgao: Sorry for not mentioning this earlier, but this would trigger it even on disconnected. Might wanna do something like `state.fs.kbfsDaemonStatus.rpcStatus === 'connected' &&` [here](https://github.com/keybase/client/blob/64eca01005d033f2b90ab63143fb2811c8fcf971/shared/actions/fs/platform-specific.desktop.js#L358)

PR #17402, opened 2019-05-06
Raise trackbreaks in CLI

mlsteele:


go/chat/teams.go
```diff
@@ -656,9 +663,17 @@ func (t *ImplicitTeamsNameInfoSource) EncryptionKey(ctx context.Context, name st
...
+	if err != nil {
 		return res, ni, err
 	}
+	if len(idFails) > 0 {
+		var suffix string
+		if len(idFails) > 1 {
+			suffix = fmt.Sprintf(" (and %v other users)", len(idFails)-1)
+		}
+		return res, ni, fmt.Errorf("Failed to identify %v%v", idFails[0].User.Username, suffix)
+	}
```
mlsteele: this is the different part

mmaxim:
I'm surprised this is needed, we don't get an error back from identify in CLI mode?

mlsteele:
The CLI is using GUI mode. I'll look into that. But this is needed anyway since the reset banner can be absent in the GUI.

mlsteele:
Right, in CLI mode we get no error but do get `TLFIdentifyFailure` from `ImplicitTeamsNameInfoSource.identify`. But there are several layers of identify result munging so I'm confused again.

mlsteele:
Ah, the reason for that `CHAT_GUI` comes across in the logs while sending a message on the CLI
```
ImplicitTeamsNameInfoSource: identify(ireset1,st103, CHAT_GUI)
```
Is it's from a push activity
```
PushHandler: chat activity: action newMessage```

mlsteele:
Going with https://github.com/keybase/client/pull/17563 instead of blocking sends from gui

PR #17410, opened 2019-05-07
check filter in filtered container

buoyad:
Got a log where this was undefined. Might be dead code since inbox search went in. ¯\\\_(ツ)\_/¯ r? @keybase/react-hackers


songgao:
As far as I know, `conversation-list` should be switching to `Kb.SearchFilter`. So whatever fixes this one in the short-term would be fine I assume.

songgao:
(currently it's used in sendAttachmentToChat in the Files tab)

PR #17411, opened 2019-05-07
restore convid to chat json api

joshblum:
reverts parts of https://github.com/keybase/client/pull/15738 and https://github.com/keybase/client/pull/15598

- removes `SkipUnbox` parameter from the list method so we neverhave unverified convIDs returned
- adds the convID param back to existing methods, `join`, `leave`, and `searchinbox` (optional) now have this param as well.


joshblum:
merging over existing kbfs-server fail with upgrade tests

PR #17413, opened 2019-05-07
Surya/core 10832/nopw cmdsignup

heronhaye:


heronhaye:
engine/logout.go
engine/account_delete.go
engine/paperprovision.go
engine/deprovision.go
engine/login.go
engine/login_provision.go
engine/signup.go

zapu:
are these other engines we need to change?

go/client/util.go
```diff
@@ -0,0 +1,26 @@
...
+
+func ensureSetPassphraseFromRemote(mctx libkb.MetaContext) error {
+	if !mctx.G().ActiveDevice.Valid() {
+		return nil
+	}
+	cli, err := GetUserClient(mctx.G())
+	if err != nil {
+		return err
+	}
+	ret, err := cli.CanLogout(mctx.Ctx(), 0)
```
zapu: btw this already does the `!h.G().ActiveDevice.Valid()` call

heronhaye: good catch, i'll remove it here then

heronhaye:
merging over autoreset flake

PR #17414, opened 2019-05-07
Profile screen for non-users

chrisnojima:
Supports non-user screens in profile
Search for 'asd' on github to see a profile

### DIFF
--- a/shared/actions/tracker2.js
+++ b/shared/actions/tracker2.js
@@ -1,6 +1,7 @@
 // @flow
 import * as Tracker2Gen from './tracker2-gen'
 import * as ProfileGen from './profile-gen'
+import * as WaitingGen from './waiting-gen'
 import * as EngineGen from './engine-gen-gen'
 import * as Saga from '../util/saga'
 import * as Constants from '../constants/tracker2'
@@ -143,6 +144,13 @@ function* load(state, action) {
       })
     )
   } catch (err) {
+    if (err.code === RPCTypes.constantsStatusCode.scresolutionfailed) {
+      // not a user, handle differently, don't show reloadable
+      yield Saga.put(WaitingGen.createClearWaiting({key: Constants.profileLoadWaitingKey}))
+      yield Saga.put(
+        Tracker2Gen.createUpdateResult({guiID: action.payload.guiID, reason: null, result: 'notAUserYet'})
+      )
+    }
     // hooked into reloadable
     logger.error(`Error loading profile: ${err.message}`)
   }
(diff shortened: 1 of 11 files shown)


shared/actions/tracker2.js
```diff
@@ -143,6 +144,11 @@ function* load(state, action) {
       })
     )
   } catch (err) {
+    if (err.code === RPCTypes.constantsStatusCode.scresolutionfailed) {
```
chrisnojima: mark as not a user

shared/common-adapters/reload.js
```diff
@@ -123,10 +124,14 @@ export type OwnProps = {|
   reloadOnMount?: boolean,
   title?: string,
   waitingKeys: string | Array<string>,
+  errorFilter?: RPCError => boolean,
```
chrisnojima: allow reloadable to skip some errors

chrisnojima:

chrisnojima:
there isn't a convenient way to get the proof icons so i deferred that

shared/profile/user/container.js
```diff
@@ -30,15 +30,18 @@ const mapStateToProps = (state, ownProps) => {
...
+  const followersCount =
+    (!notAUser && state.tracker2.usernameToDetails.getIn([username, 'followersCount'])) || 0
+  const followingCount =
+    (!notAUser && state.tracker2.usernameToDetails.getIn([username, 'followingCount'])) || 0
 
   return {
     _assertions: d.assertions,
     _suggestionKeys: userIsYou ? state.tracker2.proofSuggestions : null,
-    backgroundColorType: headerBackgroundColorType(d.state, followThem),
+    backgroundColorType: notAUser ? 'blue' : headerBackgroundColorType(d.state, followThem),
```
buoyad: :sheep: `headerBackgroundColorType` could do this

shared/profile/user/index.js
```diff
@@ -98,9 +100,22 @@ const Proofs = p => {
...
 
+  let proveIt = null
+
+  if (p.notAUser) {
+    const [name, service] = p.username.split('@')
+    proveIt = (
+      <Kb.Text
+        type="BodySmall"
+        style={styles.proveIt}
+      >{`Tell ${name} to join Keybase and prove their ${service}.`}</Kb.Text>
```
buoyad: don't need the `{``}` around this

shared/profile/user/index.js
```diff
@@ -536,6 +558,9 @@ export const styles = Styles.styleSheetCreate({
       paddingRight: Styles.globalMargins.medium,
     },
   }),
+  proveIt: {
```
buoyad: gap `Box2` above instead

chrisnojima: that puts gaps between the proofs which we don't want (i tried that first)

PR #17420, opened 2019-05-07
Don't hold ActiveDevice write lock during api call

patrickxb:


go/libkb/active_device.go
```diff
@@ -377,21 +377,39 @@ func (a *ActiveDevice) valid() bool {
 	return a.signingKey != nil && a.encryptionKey != nil && !a.uv.IsNil() && !a.deviceID.IsNil() && a.deviceName != ""
 }
 
-func (a *ActiveDevice) Ctime(m MetaContext) (keybase1.Time, error) {
-	a.Lock()
-	defer a.Unlock()
+func (a *ActiveDevice) Ctime(m MetaContext) (ctime keybase1.Time, err error) {
```
maxtaco: Thanks for jumping on this fix. I got in the habit of refactoring whenever I had multiple returns in a function that each had to release a lock separately.  It's so easy to break this sort of code in the future...

patrickxb: I know, it wasn't very pretty.  I pushed a new version.

maxtaco: thank you!

patrickxb:
CI error is with a twitter proof failure to find a div (i.e. unrelated), merging...

PR #17423, opened 2019-05-07
folder_branch_ops: add a test method for forcing a stuck conflict

strib:
This allows us to force a TLF into a stuck conflict state, where CR will not work to automatically resolve the issue.  It works by:

1. Writing a no-op `gcOp` to the main folder view, and flushing it to the server.
2. Reverting that op locally, to go back one revision.
3. Putting CR into a mode where it auto-fails every resolution.
4. Making 11 files one-by-one, to trigger CR this many times and get into a stuck state.
5. Then putting CR back into a normal mode, but it will still be "stuck" because of all the other failed attempts.

This also adds a `fs force-conflict` command for devel-only builds, to force it into this state for testing purposes.

cc: @jakob223 -- I basically took what you implemented in a test and put it in a proper `folderBranchOps` method.  I didn't both rewriting your test though, since it explicitly uses two separate users which isn't needed here.

Issue: KBFS-4136


go/client/cmd_simplefs_force_conflict.go
```diff
@@ -0,0 +1,70 @@
...
+
+// NewCmdSimpleFSForceConflict creates a new cli.Command.  Right now
+// it just triggers a "stuck" conflict, but maybe in the future we'll
+// want to add more types of manual conflicts.
+func NewCmdSimpleFSForceConflict(
+	cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	return cli.Command{
+		Name:         "force-conflict",
+		ArgumentHelp: "<path-to-folder>",
+		Usage:        "forces a conflict in the given folder ",
```
jakob223: trailing space intentional?

go/client/cmd_simplefs_force_conflict.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Run runs the command in client/server mode.
+func (c *CmdSimpleFSForceConflict) Run() error {
+	cli, err := GetSimpleFSClient(c.G())
+	if err != nil {
+		return err
+	}
+
+	return cli.SimpleFSForceStuckConflict(context.TODO(), c.path)
```
jakob223: why TODO and not Background?

go/kbfs/libkbfs/conflict_resolver.go
```diff
@@ -109,6 +109,7 @@ type ConflictResolver struct {
 	lockNextTime  bool
 	canceledCount int
 
+	failModeLock    sync.RWMutex
 	failModeForTest failModeForTest
```
jakob223: maybe rename this since it's not for tests only per se anymore?  No strong feelings though.

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -759,6 +759,120 @@ func (fbo *folderBranchOps) clearConflictView(ctx context.Context) (
...
+		return nil
+	}()
+	if err != nil {
+		return err
+	}
+
+	// Set CR to always fail.
+	oldMode := fbo.cr.getFailModeForTest()
+	fbo.cr.setFailModeForTest(alwaysFailCR)
+	defer func() { fbo.cr.setFailModeForTest(oldMode) }()
```
jakob223: It doesn't feel like the locking is doing much good on `failModeForTest` if we don't take the lock during this entire time - someone else could set the fail mode to fail always, then we set it to fail always and take their fail always as the thing to put back. Then they're done and come turn it off, and we come turn it back on. There isn't any other code accessing this, but it still feels misleading to have the locks like this if they're not effective.

WDYT?

go/client/cmd_simplefs_force_conflict.go
strib: This is copy-pasted from your ClearConflicts command!  But I think all commands use `TODO()`.  Not 100% sure, but I think eventually we'd want to make contexts that have debug IDs in them, and we just haven't yet.

go/kbfs/libkbfs/conflict_resolver.go
strib: I mean, it's still only used in testing.  Would "ForTesting" be better?

go/kbfs/libkbfs/folder_branch_ops.go
strib: The lock is just intended to prevent test data races on the access of that specific variable.  If we happen to have simultaneous get and set calls (if CR is somehow running the background already when we call this, for example).

strib:
> Also, since we now have isStuck, do you want to change conflict_resolver.go to check isStuck when deciding whether or not to proceed CRing, so that we don't have the possibility of having two different logics in 2 different places checking the same thing?

I thought about this, but it's hard to get rid of the complexity in `recordStartResolve()` because after the "stuck" check it still depends on the `db`, `key`, and `conflictsSoFar` variables, so I'd have to change `isStuck()` to return all those variables, which somehow seemed grosser than just duplicating the code.  I'm pretty on the fence though, so if that seems better to you I'll be happy to do it.  We can also make two helper functions, `isStuck()` and `isStuckWithConflicts()` or something, and the former one can call the latter but just return the bool and drop everything else.

go/client/cmd_simplefs_force_conflict.go
jakob223: LOL. Sounds reasonable to me.

go/kbfs/libkbfs/conflict_resolver.go
jakob223: Sure.

jakob223:
> > Also, since we now have isStuck, do you want to change conflict_resolver.go to check isStuck when deciding whether or not to proceed CRing, so that we don't have the possibility of having two different logics in 2 different places checking the same thing?
> 
> I thought about this, but it's hard to get rid of the complexity in `recordStartResolve()` because after the "stuck" check it still depends on the `db`, `key`, and `conflictsSoFar` variables, so I'd have to change `isStuck()` to return all those variables, which somehow seemed grosser than just duplicating the code. I'm pretty on the fence though, so if that seems better to you I'll be happy to do it. We can also make two helper functions, `isStuck()` and `isStuckWithConflicts()` or something, and the former one can call the latter but just return the bool and drop everything else.

:| it does seem like neither option is clearly better - I think the two-function approach seems the best overall.

strib:
Merging over unrelated core CI failure in kex2.

PR #17428, opened 2019-05-08
improvements on online/offline detection for Files

songgao:
Go changes (@keybase/kbfs-hackers):

* `KBFSOps().Status()` used to return error when user is logged out. This error is ignored in `libfs/status_file.go` which is why we don't see an error when reading from CLI when logged out. When GUI calls that though, the error gets propagated through into a black bar. So instead just do the ignoring part in `libkbfs`.

* Add a `SimpleFSCheckReachability` and call `MDServer().CheckReachability()` there.

* In `CheckReachability()`, it used to only force a reconnect (i.e. throwaway existing TCP conn) if Dial fails. But the reconnect has a backoff timer which makes things slow when we get back online. So this PR also makes `CheckReachability` "fast forwards" any existing reconnect (which would typically be started from a previous call where Dial failed), which essentially skips the random backoff timer. This makes the online detection instant when, for example, user plug in an Ethernet cable and the machine gets an IP.

FE changes (@keybase/react-hackers):

* Change the part that hooks into the OS networking API so that it not only triggers an event, but also plumbs through about if it's online or offline.

* Add `ConfigGen.osNetworkStatusChanged`, and fire that when OS network status changes, instead of just the gregor reachability event.

* Trigger a `ConfigGen.osNetworkStatusChanged` at init with the correct state, to make the added `config.osNetworkOnline` correct at start. This action is fired with `isInit: true`, which is used in next point.

* On `ConfigGen.osNetworkStatusChanged`, if `!isInit`, trigger `CheckReachability` through `SimpleFS`, which tests if we can dial to the MD server. The reason why we don't do that at init is we don't want to override the random backoff timer that we put in back in time to avoid surge server load when we make a new release.


songgao:
CI is having issues. I did a `go test` in `libkbfs` locally and it passed. That should cover the last failure seems irrelevant. So I'll just merge this now.

go/kbfs/libkbfs/kbfs_ops.go
```diff
@@ -1057,18 +1058,26 @@ func (fs *KBFSOpsStandard) Status(ctx context.Context) (
...
-		var quErr error
-		_, usageBytes, archiveBytes, limitBytes,
-			gitUsageBytes, gitArchiveBytes, gitLimitBytes, quErr =
-			fs.quotaUsage.GetAllTypes(
-				ctx, quotaUsageStaleTolerance/2, quotaUsageStaleTolerance)
-		if quErr != nil {
-			// The error is ignored here so that other fields can still be populated
-			// even if this fails.
-			fs.log.CDebugf(ctx, "Getting quota usage error: %v", quErr)
+	switch err.(type) {
```
jzila: `errors.Cause(err).(type)`

go/kbfs/libkbfs/kbfs_ops.go
```diff
@@ -1057,18 +1058,26 @@ func (fs *KBFSOpsStandard) Status(ctx context.Context) (
...
+			_, usageBytes, archiveBytes, limitBytes,
+				gitUsageBytes, gitArchiveBytes, gitLimitBytes, quErr =
+				fs.quotaUsage.GetAllTypes(
+					ctx, quotaUsageStaleTolerance/2, quotaUsageStaleTolerance)
+			if quErr != nil {
+				// The error is ignored here so that other fields can still be populated
+				// even if this fails.
+				fs.log.CDebugf(ctx, "Getting quota usage error: %v", quErr)
+			}
 		}
```
jzila: Shouldn't we at least log if we aren't connected?

shared/actions/fs/index.js
```diff
@@ -909,11 +909,20 @@ const waitForKbfsDaemon = (state, action) => {
...
+const updateKbfsDaemonOnlineStatus = (state, action) =>
+  state.fs.kbfsDaemonStatus.rpcStatus === 'connected' && state.config.osNetworkOnline
+    ? RPCTypes.SimpleFSSimpleFSAreWeConnectedToMDServerRpcPromise().then(connectedToMDServer =>
+        FsGen.createKbfsDaemonOnlineStatusChanged({online: connectedToMDServer})
+      )
+    : Promise.resolve(FsGen.createKbfsDaemonOnlineStatusChanged({online: false}))
+
+// We don't trigger the reachability check at init. Reachability checks cause
+// any pending "reconnect" fire right away, and overrides any random back-off
+// timer we have at process restart (which is there to avoid surge server
```
jzila: s/surge/surging

PR #17443, opened 2019-05-08
Disable secret store priming in tests by default

joshblum:


zapu:
hm, i'd feel so much better if this wasn't exposed to users via environment on command line. but if that's needed it's fine

joshblum:
@zapu agreed, disabled cmdline/json config

joshblum:
merging over kbfs flake

PR #17447, opened 2019-05-08
two small fixes for master

songgao:


shared/fs/common/filename.js
```diff
@@ -38,7 +38,7 @@ const Filename = (props: Props) => {
       >
         {fileNameWithoutExtension}
       </Kb.Text>
-      {fileExtension && (
+      {!!fileExtension && (
```
songgao: @jakob223 sorry for missing this in review. Problem here is when `fileExtension` is an empty string, it'd just end up in there, and on Android strings have to be inside the `Text` component.

jakob223: oops 😬

PR #17448, opened 2019-05-08
libkbfs: set conflict state for each TLF in the favorites

strib:
This PR adds a conflict state to each `keybase1.Folder` object returned with the favorites list from SimpleFS.

Rather than loop through all possible favorites, it uses the journal manager to figure out which TLFs are currently in conflict, compares that list manually against the favorites list, and marks the ones that match.

It also adds a conflict status to the TLF `.kbfs_status` file, as well as the set of conflicting TLFs to the overall `.kbfs_status` file.

Notifications to the GUI that favorites have changed will come in a future PR.

Issue: KBFS-4137

### DIFF
--- a/go/kbfs/libkbfs/interfaces.go
+++ b/go/kbfs/libkbfs/interfaces.go
@@ -428,6 +428,11 @@ type KBFSOps interface {
 	// updated (to eliminate the need for polling this method).
 	FolderStatus(ctx context.Context, folderBranch data.FolderBranch) (
 		FolderBranchStatus, <-chan StatusUpdate, error)
+	// FolderConflictStatus is alightweight method to return the
+	// conflict status of a particular folder/branch.  (The conflict
+	// status is also available in `FolderBranchStatus`.
+	FolderConflictStatus(ctx context.Context, folderBranch data.FolderBranch) (
+		keybase1.FolderConflictType, error)
 	// Status returns the status of KBFS, along with a channel that will be
 	// closed when the status has been updated (to eliminate the need for
 	// polling this method). Note that this channel only applies to
--- a/go/kbfs/libkbfs/interfaces.go
+++ b/go/kbfs/libkbfs/interfaces.go
@@ -428,9 +428,9 @@ type KBFSOps interface {
 	// updated (to eliminate the need for polling this method).
 	FolderStatus(ctx context.Context, folderBranch data.FolderBranch) (
 		FolderBranchStatus, <-chan StatusUpdate, error)
-	// FolderConflictStatus is alightweight method to return the
+	// FolderConflictStatus is a lightweight method to return the
 	// conflict status of a particular folder/branch.  (The conflict
-	// status is also available in `FolderBranchStatus`.
+	// status is also available in `FolderBranchStatus`.)
 	FolderConflictStatus(ctx context.Context, folderBranch data.FolderBranch) (
 		keybase1.FolderConflictType, error)
 	// Status returns the status of KBFS, along with a channel that will be
--- a/go/kbfs/libkbfs/kbfs_ops.go
+++ b/go/kbfs/libkbfs/kbfs_ops.go
@@ -267,12 +267,62 @@ func (fs *KBFSOpsStandard) GetFavorites(ctx context.Context) (
 
 // GetFavoritesAll implements the KBFSOps interface for
 // KBFSOpsStandard.
-func (fs *KBFSOpsStandard) GetFavoritesAll(ctx context.Context) (keybase1.
-	FavoritesResult, error) {
+func (fs *KBFSOpsStandard) GetFavoritesAll(ctx context.Context) (
+	keybase1.FavoritesResult, error) {
 	timeTrackerDone := fs.longOperationDebugDumper.Begin(ctx)
 	defer timeTrackerDone()
 
-	return fs.favs.GetAll(ctx)
+	favs, err := fs.favs.GetAll(ctx)
+	if err != nil {
+		return keybase1.FavoritesResult{}, err
+	}
+
+	// Add the conflict status for any folders in a conflict state to
+	// the favorite struct.
+	journalManager, err := GetJournalManager(fs.config)
+	if err != nil {
+		// Journaling not enabled.
+		return favs, nil
+	}
+	conflicts, err := journalManager.GetJournalsInConflict(ctx)
+	if err != nil {
+		return keybase1.FavoritesResult{}, err
+	}
+
+	if len(conflicts) == 0 {
+		return favs, nil
+	}
+	conflictMap := make(map[ConflictJournalRecord]tlf.ID, len(conflicts))
+	for _, c := range conflicts {
+		conflictMap[ConflictJournalRecord{Name: c.Name, Type: c.Type}] = c.ID
+	}
+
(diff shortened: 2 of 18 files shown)


go/kbfs/libkbfs/interfaces.go
```diff
@@ -428,6 +428,11 @@ type KBFSOps interface {
 	// updated (to eliminate the need for polling this method).
 	FolderStatus(ctx context.Context, folderBranch data.FolderBranch) (
 		FolderBranchStatus, <-chan StatusUpdate, error)
+	// FolderConflictStatus is alightweight method to return the
```
jakob223: typo: alightweight

go/kbfs/libkbfs/interfaces.go
```diff
@@ -428,6 +428,11 @@ type KBFSOps interface {
...
+	// conflict status of a particular folder/branch.  (The conflict
```
jakob223: no close paren

go/kbfs/libkbfs/kbfs_ops.go
```diff
@@ -267,12 +267,62 @@ func (fs *KBFSOpsStandard) GetFavorites(ctx context.Context) (
...
+	if len(conflicts) == 0 {
+		return favs, nil
+	}
+	conflictMap := make(map[ConflictJournalRecord]tlf.ID, len(conflicts))
+	for _, c := range conflicts {
+		conflictMap[ConflictJournalRecord{Name: c.Name, Type: c.Type}] = c.ID
+	}
+
+	found := 0
+	for i, f := range favs.FavoriteFolders {
```
jakob223: Should we consider also doing this for new folders? It would be confusing to end up in a conflict state due to CLI shenanigans and then not be able to see it in the GUI.

strib: I don't think there is a way for new folders to be edited without appearing in the favorites list.  By "CLI shenanigans" do you mean writing to the TLF via a terminal?  Any access in the terminal automatically puts a TLF on your favorites list.

jakob223: I meant `keybase fs` and/or FUSE, I guess.
>  Any access in the terminal automatically puts a TLF on your favorites list.
In that case, then I agree it's not necessary.

strib: Yeah, `keybase fs` works the same way, TLFs get auto-added to favorites.

shared/constants/rpc.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 import {favoriteFolderType, type FolderType} from './types/rpc-gen'
+
```
songgao: 🔥

protocol/avdl/keybase1/favorite.avdl
```diff
@@ -33,6 +38,9 @@ protocol favorite {
     array<User> resetMembers;
 
     union { null, Time } mtime;
+
+    // Is the local view of this folder hosed due to stuck CR?
+    FolderConflictType conflictType;
```
songgao: Hmm this `Folder` type is for both RPC param to reference a TLF, and as result from RPC to describe a TLF's metadata. It'd be nice if we can at some point separate the two as it doesn't make much sense to have resetMembers, conflict, etc. when referring to a TLF from GUI.

go/kbfs/libkbfs/kbfs_ops.go
```diff
@@ -267,12 +267,62 @@ func (fs *KBFSOpsStandard) GetFavorites(ctx context.Context) (
 
 // GetFavoritesAll implements the KBFSOps interface for
 // KBFSOpsStandard.
-func (fs *KBFSOpsStandard) GetFavoritesAll(ctx context.Context) (keybase1.
-	FavoritesResult, error) {
+func (fs *KBFSOpsStandard) GetFavoritesAll(ctx context.Context) (
```
songgao: Is the plan to use this for the SimpleFS interface for TLF metadata? I think we'll need some more than what's currently in `keybase1.Folder`: https://github.com/keybase/client/blob/dea76af9287733957df1f2123d605dd534d937d6/shared/constants/types/fs.js#L70-L89

I guess these could be filled by simpleFS though.

go/kbfs/libkbfs/kbfs_ops.go
```diff
@@ -267,12 +267,62 @@ func (fs *KBFSOpsStandard) GetFavorites(ctx context.Context) (
...
+		return keybase1.FavoritesResult{}, err
+	}
+
+	if len(conflicts) == 0 {
+		return favs, nil
+	}
+	conflictMap := make(map[ConflictJournalRecord]tlf.ID, len(conflicts))
+	for _, c := range conflicts {
+		conflictMap[ConflictJournalRecord{Name: c.Name, Type: c.Type}] = c.ID
+	}
```
songgao: Is there a reason not to just make a map (instead of slice) in `getJournalsInConflictLocked`?

go/kbfs/libkbfs/kbfs_ops.go
strib: I have no idea, I don't know where you currently get the TLF metadata.  Note that you don't just get a `Folder`, you get a `FavoriteResult`, which categorizes the favorite vs. new vs. ignored stuff already.  I don't know where this other stuff comes from.

protocol/avdl/keybase1/favorite.avdl
strib: Noted, feel free to make a ticket if it's important to you.  Seems like it would require some JS-side work I'm not comfortable doing.

go/kbfs/libkbfs/kbfs_ops.go
strib: It's a slice because it's not obvious what the key should be when it's returned from `JournalManager`, and when it's put in the JSON `JournalManager` status.  Makes more sense as a list in both those places in my opinion.  Here it needs to be a map just because we're looping through the favorites and need to do a comparison, but that's not a general use case and the key+value used here is specific for that purpose.

go/kbfs/libkbfs/kbfs_ops.go
songgao: Oh that's right; that solves the ignored/new things. So that leaves us just the reset-related stuff I think. Currently they are from API server `kbfs/favorite/list`

go/kbfs/libkbfs/kbfs_ops.go
songgao: Sounds good; thanks for the explanation!

protocol/avdl/keybase1/favorite.avdl
songgao: Cool; made KBFS-4157

go/kbfs/libkbfs/kbfs_ops.go
jakob223: this interface is the same as the one provided by that API endpoint - the reset stuff is in `resetMembers` IIRC.

songgao: Sweet; I'll look into plumbing this into GUI. I seem to have a de javu now and I think you've said that days ago. Sorry for forgetting that!

PR #17449, opened 2019-05-08
Connect out-of-space banners, add them to widget, plumb things

jakob223:
Also, I prettified some files that my editor kept prettifying and I got tired of checking back out.

Issue: KBFS-4129


shared/menubar/remote-container.desktop.js
```diff
@@ -75,9 +79,8 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
   return {
     ...stateProps,
     ...dispatchProps,
-    refreshUserFileEdits: dispatchProps.refreshUserFileEdits,
+    onRetrySync: dispatchProps._onRetrySync(stateProps.username),
```
songgao: Did you mean to remove `refreshUserFileEdits and waitForKbfsDaemon`?

shared/menubar/space-warning.js
```diff
@@ -1,26 +1,59 @@
...
-          : 'You are' + ' out of storage space. Unsync some folders, or make some space then'
-      }
-      color={props.bannerType === 'warning' ? 'blue' : 'red'}
-      actions={[...(props.onRetry ? [{onClick: props.onRetry, title: 'retry' + ' the sync.'}] : [])]}
-    />
-  )
+  _close = () => {
+    this.setState(() => ({
+      hidden: true,
+    }))
```
songgao: The state will be gone once the component is unmounted (which I think happens when the widget hides). Did you mean to have it come back next time the widget is open?

shared/menubar/remote-container.desktop.js
jakob223: they're already included with `...dispatchProps`.

shared/menubar/space-warning.js
jakob223: I did :/ - any suggestions for that?

songgao: Oh I think I might have misunderstood. It sounds like the current behavior is it doesn't come back when the widget is closed and re-opened, and you want that to happen?

jakob223: I did mean for the state to come back so that the banner does not come back.

songgao: Oh I see. I think instead of using component state, we'd just need to move it into the redux store. The SFMI banner uses a similar mechanism.

jakob223:
@songgao I moved the state to the redux store and merged in the other PR.

shared/reducers/fs.js
```diff
@@ -525,17 +525,35 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
+          )
       )
     case FsGen.setDriverStatus:
       return state.update('sfmi', sfmi => sfmi.set('driverStatus', action.payload.driverStatus))
     case FsGen.showSystemFileManagerIntegrationBanner:
       return state.update('sfmi', sfmi => sfmi.set('showingBanner', true))
     case FsGen.hideSystemFileManagerIntegrationBanner:
       return state.update('sfmi', sfmi => sfmi.set('showingBanner', false))
+    case FsGen.hideDiskSpaceBanner:
+      return state.update('overallSyncStatus', status => status.set('diskSpaceBannerHidden', false))
```
songgao: 🧐is this supposed to be `true`? If so seems like the one in the `overallSyncStatusChanged` case above should be reversed too

jakob223: It does seem that way. I swear I tested this ._.

jakob223: The one in the above case is correct though. Hidden should be set only if both of the following hold:
 - hidden was already set
 - we are still in warning state

shared/reducers/fs.js
```diff
@@ -532,14 +532,7 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
           )
-          .update('diskSpaceStatus', oldDiskSpaceStatus => {
-            // TODO: allow for 'warning' status
-            const newDiskSpaceStatus = action.payload.outOfSpace ? 'error' : 'ok'
-            if (newDiskSpaceStatus !== oldDiskSpaceStatus && oldDiskSpaceStatus !== 'error') {
-              notifyDiskSpaceStatus(newDiskSpaceStatus)
-            }
-            return newDiskSpaceStatus
-          })
+          .set('diskSpaceStatus', action.payload.outOfSpace ? 'error' : 'ok')
```
songgao: do we want to swap this for enums now?

songgao:
I think this looks good to me now, but still needs the rebase. Sorry for failing to get through this before the ts move.

jakob223:
np, rebase shouldn't be hard, TS is running in my veins :P

PR #17454, opened 2019-05-08
fix my embarrassing bug where refreshTag was never set for folderListLoad

songgao:


songgao:
I think story failure is because I forgot to update providers. Will fix when I get a chance.

PR #17460, opened 2019-05-09
bot autolog support

pzduniak:


pzduniak:
used by https://github.com/keybase/keybase-bot/pull/139

go/client/cmd_status.go
```diff
@@ -169,7 +169,11 @@ func (c *CmdStatus) load() (*fstatus, error) {
 		status.Service.Running = false
 	} else {
 		status.Service.Running = true
-		status.Service.Log = filepath.Join(extStatus.LogDir, libkb.ServiceLogFileName)
+		if logPath := c.G().Env.GetLogFile(); logPath != "" {
+			status.Service.Log = logPath
```
joshblum: this can't be right as is, at the very least the filename name need to be present. did you try a log send? why is this needed?

go/client/cmd_log_send.go
```diff
@@ -205,6 +209,7 @@ func (c *CmdLogSend) ParseArgv(ctx *cli.Context) error {
 	}
 	c.noConfirm = ctx.Bool("no-confirm")
 	c.numBytes = ctx.Int("n")
+	c.feedback = ctx.String("feedback")
```
joshblum: shouldn't `getFeedback` be modified to skip prompted if this is provided?

pzduniak: The idea is that you could still get interactivity with some prefixed note. Probably would never get used so I can remove it.

go/client/cmd_status.go
pzduniak: logPath is path to the logFile. I could rename it to `logFilePath` to make it clearer. Check the keybase-bot pullreq to see how it's used. Without this change LogDir has nothing to do with what we're passing to --log-file, so you end up sending logs with no service logs (only status).

pzduniak: ... this is still quite hacky, I could move GetLogFile() resolving to something that service returns here so that we get the proper path. As-is this version is good for telling the CLI to "send file x as service logs".

go/client/cmd_log_send.go
joshblum: should just skip `getFeedback` if this is set

go/client/cmd_status.go
joshblum: this should happen in `GetExtendedStatus`. When `--log-file` is set does that only affect the service log file? what happens to all of the other logs? Are they still rooted at `g.Env.GetLogDir()`?

pzduniak: All the other logs stay as-is. I don't know why this is implemented this way, looks like it was reworked a while ago and everything is happy with the status quo so noone reworked it.

pzduniak: ... actual logging seems to use .GetEffectiveLogFile(), probably worth running it

pzduniak: ok this should be perfect

joshblum:
merging over test flake to bring into https://github.com/keybase/client/pull/17464

PR #17463, opened 2019-05-09
packaging: only kill main electron process in run_keybase

heronhaye:


heronhaye:
We did this in the systemd unit files, this makes it so run_keybase follows the same logic as well

PR #17464, opened 2019-05-09
refactor to status pkg

joshblum:
refactors logic for getting the status of keybase from the `client` package to `status` so the a `keybase1.FullStatus` is a single RPC call the GUI can use. The output changed slightly when using the protocol definitions (casing and a few extra fields) but nothing was lost. tested sending with and without a service. 

https://github.com/keybase/keybase/pull/3842


go/status/status.go
```diff
@@ -0,0 +1,244 @@
...
+	if uid := cr.GetUID(); uid.Exists() {
+		res.Registered = true
+		res.User = libkb.NewUserThin(cr.GetUsername().String(), uid).Export()
+	}
+	res.SessionIsValid = mctx.ActiveDevice().Valid()
+	res.LoggedIn = res.SessionIsValid
+	return res, nil
+}
+
+func GetClientStatus(mctx libkb.MetaContext) (res []keybase1.ClientStatus) {
```
heronhaye: is this different from `libkb.GetClientStatus`?

joshblum: oops no this is supposed to be removed since `install` needs this function

joshblum:
merging over TestSetPhoneNumber flake

go/status/status_nix.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// osVersionAndBuild returns OS version, and build too on some platforms
+func osVersionAndBuild() (string, string, error) {
+	productVersion, err := execToString("uname", []string{"-mrs"})
+	if err != nil {
+		return "", "", err
+	}
+
+	buildVersion, err := execToString("lsb_release", []string{"-sd"})
```
jzila: This broke kbfs-server tests.

PR #17467, opened 2019-05-09
add version to font to hopefully fix the windows installer

chrisnojima:
@keybase/react-hackers


shared/desktop/yarn-helper/font.js
```diff
@@ -133,9 +132,11 @@ function updateIconFont(web) {
     },
     (error, result) => (error ? fontsGeneratedError(error) : fontsGeneratedSuccess(web, result))
   )
+  console.log('Created new webfont')
```
buoyad: just `font` I think, or check if `web` and change the message

chrisnojima: sure, this was me moving it down

PR #17471, opened 2019-05-09
libkbfs: send notifications when favorites, and CR stuck state, change

strib:
This PR adds a notification that the GUI can listen for, which will be fired when the favorite state changes and should be refreshed.

This needs to happen both when favorites are added/deleted, and when the CR stuck status changes for a local TLF view.

Tested using the command line dev command `keybase show-notifications`.

@jakob223 and @songgao should probably both review.

Issue: KBFS-4137

### DIFF
--- a/go/kbfs/libkbfs/favorites.go
+++ b/go/kbfs/libkbfs/favorites.go
@@ -298,18 +298,21 @@ func (f *Favorites) closeReq(req *favReq, err error) {
 
 // sendChangesToEditHistory notes any deleted favorites and removes them
 // from this user's kbfsedits.UserHistory.
-func (f *Favorites) sendChangesToEditHistory(oldCache map[favorites.Folder]favorites.Data) {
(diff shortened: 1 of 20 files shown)


go/kbfs/libkbfs/favorites.go
```diff
@@ -339,6 +342,7 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 	// quickly when offline.
 	needFetch := (req.refresh || f.favCache == nil) && !req.clear
 	wantFetch := f.config.Clock().Now().After(f.cacheExpireTime) && !req.clear
+	changed := false
```
songgao: Heh just realized I made the almost exact change in my PR: https://github.com/keybase/client/pull/17530/files#diff-5919763ea12fd1848e70a4ca1d6ce411R446

I'll incorporate yours after yours is merged

PR #17475, opened 2019-05-10
Fix mobile-only lockdown account UX

cjb:
@keybase/react-hackers CC @buoyad @patrickxb 

We've been storing whether a Stellar account is in `mobileOnlyMode` and using that to decide whether to allow sends -- if you're on a desktop device and the account's in mobileOnlyMode, you can't send.

Two problems with this:

* There are other disallowed operations: set as default and deleting the account.

* The criteria aren't correct: it's actually "if you're on a desktop, or a mobile device that's less than 7 days old" then these operations are disallowed.

So, this PR moves `disabledDueToMobileOnly` to instead test (and be called) `thisDeviceIsLockedOut`, and then expands the use from the send form to set as default and delete as well.  The test is done using `!account.accountModeEditable` to mean that this device is locked down.

@buoyad, you mentioned a confusing error on the send form itself -- I didn't change that, but since the Send button now isn't available if it's not going to work, I think we can leave that as-is and trust that we're unlikely to see it, does that make sense?

### DIFF
--- a/shared/constants/wallets.js
+++ b/shared/constants/wallets.js
@@ -612,6 +612,9 @@ export const getCurrencyAndSymbol = (state: TypedState, code: string) => {
 
 export const getAcceptedDisclaimer = (state: TypedState) => state.wallets.acceptedDisclaimer
 
+export const getThisDeviceIsLockedOut = (state: TypedState, accountID: Types.AccountID) =>
+  getAccount(state, accountID).mobileOnlyEditable
+
 export const balanceChangeColor = (delta: Types.PaymentDelta, status: Types.StatusSimplified) => {
   let balanceChangeColor = Styles.globalColors.black
   if (delta !== 'none') {
--- a/shared/constants/wallets.js
+++ b/shared/constants/wallets.js
@@ -612,8 +612,12 @@ export const getCurrencyAndSymbol = (state: TypedState, code: string) => {
 
 export const getAcceptedDisclaimer = (state: TypedState) => state.wallets.acceptedDisclaimer
 
+// The device is read-only if it's a desktop device and mobile-only is on, or it's a mobile device
+// that's less than seven days old.  TODO: Use a `deviceReadOnly` field insstead once it exists.
 export const getThisDeviceIsLockedOut = (state: TypedState, accountID: Types.AccountID) =>
-  getAccount(state, accountID).mobileOnlyEditable
+  Styles.isMobile
+    ? getAccount(state, accountID).mobileOnlyEditable
+    : state.wallets.mobileOnlyMap.get(accountID, false)
 
 export const balanceChangeColor = (delta: Types.PaymentDelta, status: Types.StatusSimplified) => {
   let balanceChangeColor = Styles.globalColors.black
--- a/shared/constants/wallets.js
+++ b/shared/constants/wallets.js
@@ -616,7 +616,7 @@ export const getAcceptedDisclaimer = (state: TypedState) => state.wallets.accept
 // that's less than seven days old.  TODO: Use a `deviceReadOnly` field insstead once it exists.
 export const getThisDeviceIsLockedOut = (state: TypedState, accountID: Types.AccountID) =>
   Styles.isMobile
-    ? getAccount(state, accountID).mobileOnlyEditable
+    ? !getAccount(state, accountID).mobileOnlyEditable
     : state.wallets.mobileOnlyMap.get(accountID, false)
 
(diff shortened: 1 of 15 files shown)


shared/wallets/common/buttons/index.js
```diff
@@ -47,8 +47,10 @@ const _SendButton = (props: Kb.PropsWithOverlay<SendProps>) => {
       />
     </>
   )
-  return props.disabledDueToMobileOnly ? (
-    <Kb.WithTooltip text="This is a mobile-only account.">{button}</Kb.WithTooltip>
+  return props.thisDeviceIsLockedOut ? (
+    <Kb.WithTooltip text="You can only send from a mobile device less than 7 days old.">
```
patrickxb: more than 7 days old...

shared/wallets/wallet/header/index.js
```diff
@@ -85,6 +86,9 @@ const Header = (props: Props) => {
         />
         <DropdownButton />
       </Kb.Box2>
+      {props.thisDeviceIsLockedOut && Styles.isMobile && (
+        <Kb.Text type="BodySmall">You can only send from a mobile device less than 7 days old.</Kb.Text>
```
patrickxb: more than 7 days old

patrickxb:
I'm a little confused.  Where do you figure out `thisDeviceIsLockedOut`?

One note: accountModeEditable will always be false on desktop devices.  Desktop devices are not allowed to toggle the mobile-only setting.  It is also false on mobile devices that are less than 7 days old.

cjb:
@patrickxb Thanks! Here:
```js
export const getThisDeviceIsLockedOut = (state: TypedState, accountID: Types.AccountID) =>
  getAccount(state, accountID).mobileOnlyEditable
```
.. where mobileOnlyEditable is based on `WalletAccountLocal.accountModeEditable`.

Sounds like that's wrong, because it doesn't account for desktop being allowed to Send when mobile only is off, but not allowed to edit account mode.

So perhaps this would work:
```js
export const getThisDeviceIsLockedOut = (state: TypedState, accountID: Types.AccountID) => {
  if (isMobile) {
    return getAccount(state, accountID).mobileOnlyEditable
  } else {
    return !state.wallets.mobileOnlyMap.get(accountID, false)
  }
}
```
.. or perhaps we should ask you to put a `deviceIsLockedOut` field onto WalletAccountLocal.  Any preference?

patrickxb:
Your second function looks good to me, but we can move that logic into the service.  How about a field name with something like `deviceReadOnly`, because it is allowed to see tx history and tx details.

We do have a "CanSubmitTx" field as well, which is true if your balance is above the base fee.  Do you use that for anything?

cjb:
`deviceReadOnly` sounds great, thanks.

We do use `canSubmitTx`, but only to gate whether this device is allowed to change inflation destination, it seems.

cjb:
The PR should be correct as-is now, and we can move to `deviceReadOnly` after it exists on the service-side.

Thanks!

shared/constants/wallets.js
```diff
@@ -612,6 +612,13 @@ export const getCurrencyAndSymbol = (state: TypedState, code: string) => {
 
 export const getAcceptedDisclaimer = (state: TypedState) => state.wallets.acceptedDisclaimer
 
+// The device is read-only if it's a desktop device and mobile-only is on, or it's a mobile device
+// that's less than seven days old.  TODO: Use a `deviceReadOnly` field insstead once it exists.
```
chrisnojima: 'instead'

shared/constants/wallets.js
```diff
@@ -612,6 +612,13 @@ export const getCurrencyAndSymbol = (state: TypedState, code: string) => {
...
+export const getThisDeviceIsLockedOut = (state: TypedState, accountID: Types.AccountID) =>
+  Styles.isMobile
+    ? getAccount(state, accountID).mobileOnlyEditable
```
buoyad: `mobileOnlyEditable` means this device **can** change the setting right?

shared/wallets/wallet/header/index.js
```diff
@@ -85,6 +86,9 @@ const Header = (props: Props) => {
         />
         <DropdownButton />
       </Kb.Box2>
+      {props.thisDeviceIsLockedOut && (
+        <Kb.Text type="BodySmall">You can only send from a mobile device more than 7 days old.</Kb.Text>
```
buoyad: `center={true}`, wraps on the iphone 5S

PR #17476, opened 2019-05-10
override cmd usage if running in standalone mode

pzduniak:
fixes a bug where sometimes standalone commands panic becuase certain components aren't initialized

### DIFF
--- a/go/libkb/globals.go
+++ b/go/libkb/globals.go
@@ -721,6 +721,14 @@ func (u Usage) UseKeyring() bool {
 	return u.KbKeyring || u.GpgKeyring
 }
 
+var ServiceUsage = Usage{
+	Config:     true,
+	KbKeyring:  true,
+	GpgKeyring: true,
+	API:        true,
+	Socket:     true,
+}
+
 func (g *GlobalContext) ConfigureCommand(line CommandLine, cmd Command) error {
 	usage := cmd.GetUsage()
 	return g.Configure(line, usage)
@@ -730,6 +738,10 @@ func (g *GlobalContext) Configure(line CommandLine, usage Usage) error {
 	g.SetCommandLine(line)
 	if err := g.ConfigureLogging(); err != nil {
 		return err
+	}
+	if g.Env.GetStandalone() {
+		// If standalone, override the usage to be the same as in a service
+		usage = ServiceUsage
 	}
 
 	if err := g.ConfigureUsage(usage); err != nil {
--- a/go/service/main.go
+++ b/go/service/main.go
@@ -1153,13 +1153,7 @@ func NewCmdService(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Comma
 }
 
 func (d *Service) GetUsage() libkb.Usage {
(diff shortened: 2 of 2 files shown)


go/libkb/globals.go
```diff
@@ -731,6 +739,10 @@ func (g *GlobalContext) Configure(line CommandLine, usage Usage) error {
 	if err := g.ConfigureLogging(); err != nil {
 		return err
 	}
+	if g.Env.GetStandalone() {
+		// If standalone, override the usage to be the same as in a service
+		usage = ServiceUsage
```
heronhaye: This doesn't seem right, why do we need to configure the API for doing a `config get`?

pzduniak: Because we're running a complete service! In this particular case, the service tries to sync up the configured user and fails because API is not initialized. An alternative fix would be to rework all possible code paths to be aware that some modules might not be initialized.

heronhaye: But can we not do that? For example ` echo hello | keybase --standalone base62 encode` works without any API calls. So should config get.

pzduniak: Counterpoint: base62 does not interact with the service. It would also crash if it did.

The following commands share this property:
1) base62
2) chat api listen, ctl commands, show-notification - don't make sense in standalone mode, usually explicitly disabled
4) internal kbfs utils - interact with kbfs processes directly
5) installation / configuration utils that only work on osx/windows and interact with the OS.

The following commands will crash in standalone mode:
- everything `config`
- `dismiss pgp-storage`
- some of the things above probably indirectly interact with the service, namely kbfs stuff

Additionally I feel that some commands might run into weird edge cases when we build more and more functionality in the initialization etc. The flow which breaks is the following one:
1) service init triggers d.SetupCriticalSubServices() in service/main.go:329
2) which triggers ephemeral.ServiceInit on line 334 in the same file
3) which then tries to load local user's upak during eklib's initialization

3 possible fixes:
1) add API to Usage of every command that doesn't have it and isn't explicitly disabled in standalone mode
2) make sure that the API is initialized in the eklib initialization code (effectively this will mean that standalone mode will always have an initialized API and there will be another place that runs ConfigureAPI other than ConfigureUsage)
3) this pullreq, which has only a single caveat - some commands that don't interact with the service might end up making client requests because eklib _always_ gets triggered

heronhaye: I think that running a `config get` shouldn't trigger anything ephemeral at all, it should work offline. Talking to service is fine but it doesn't need to talk to network. Can we skip `ephemeral.ServiceInit` if in standalone mode? cc @joshblum for ephemeral details

joshblum: let's leave ephemeral in serviceinit since even in standalone expired keys can be deleted

patrickxb: So many things do not work in standalone mode.  Nobody uses it outside of keybase.  I don't think we should put any effort into it.

patrickxb: That being said, I agree that any form of `keybase config get` should work offline.

PR #17477, opened 2019-05-10
LogSend RPC for gui

joshblum:
refactors logic out of the `keybase log send` command into the `status` package to expose a simple RPC for the gui to call. 

- [x] verify mobile

### DIFF
--- a/go/client/cmd_status.go
+++ b/go/client/cmd_status.go
@@ -92,7 +92,7 @@ func (c *CmdStatus) outputJSON(status *keybase1.FullStatus) error {
 func (c *CmdStatus) outputTerminal(status *keybase1.FullStatus) error {
 	extStatus := status.ExtStatus
 	dui := c.G().UI.GetDumbOutputUI()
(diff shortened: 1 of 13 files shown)


go/status/log_send.go
```diff
@@ -4,27 +4,21 @@
 package status
 
 import (
-	"archive/tar"
```
joshblum: sorry about the huge diff here. the helper functions were moved without change to `status/utils.go`

go/client/cmd_status.go
```diff
@@ -92,7 +92,7 @@ func (c *CmdStatus) outputJSON(status *keybase1.FullStatus) error {
 func (c *CmdStatus) outputTerminal(status *keybase1.FullStatus) error {
 	extStatus := status.ExtStatus
 	dui := c.G().UI.GetDumbOutputUI()
-	dui.Printf("Username:      %s\n", status.CurStatus.User.Username)
```
joshblum: @jzila this was the crasher for kbfs-server tests

jzila: Thanks!

PR #17479, opened 2019-05-10
Deprecate nav 1

chrisnojima:
@keybase/react-hackers this deprecated nav1


shared/constants/fs.js
```diff
@@ -1061,7 +1048,7 @@ export const makeActionsForDestinationPickerOpen = (
     index,
     path,
   }),
-  flags.useNewRouter || !routePath
+  !routePath
```
songgao: This seems wrong. Should just take the thing after `?`

chrisnojima: so the existing code is wrong is what you're saying?

songgao: No existing code is right. Note that's a `||` not `&&`, so with nav2 we never hit the `:` case.

PR #17482, opened 2019-05-10
drop gregor connection on desktop power events like we do on mobile

mmaxim:
We get bizarre problems where the internal retry loop in the RPC library seems to get stuck on resume on Windows. As a result, after coming out of sleep on Windows, there is a good chance the connection to Gregor just won't get re-made. Rather than trying to figure that out, instead hook into the power events system, and drop and reconnect the entire connection object like we do in the mobile app.

### DIFF
--- a/go/service/gregor.go
+++ b/go/service/gregor.go
@@ -266,13 +266,35 @@ func (g *gregorHandler) Init() {
 	go g.syncReplayThread()
 }
 
+const (
+	monitorConnect int = iota
+	monitorDisconnect
+	monitorNoop
+)
+
 func (g *gregorHandler) monitorAppState() {
 	// Wait for state updates and react accordingly
 	state := keybase1.MobileAppState_FOREGROUND
+	suspended := false
 	for {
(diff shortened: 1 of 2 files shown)


mmaxim:
cc @maxtaco

go/service/gregor.go
```diff
@@ -266,21 +266,43 @@ func (g *gregorHandler) Init() {
...
+			if !suspended {
+				monitorAction = monitorConnect
+			} else {
+				monitorAction = monitorDisconnect
+			}
+		}
+		switch monitorAction {
+		case monitorConnect:
 			// Make sure the URI is set before attempting this (possible it isn't in a race)
 			if g.uri != nil {
```
mlsteele: This isn't new but it's not clear to me that `g.uri` isn't a potential data race.

mmaxim:
I'm not going to fix CI for this, some compile error in systests (not sure how that could get in there), and kbfs-server fail.

go/service/gregor.go
mmaxim: Probably, but in practice this is just fixed to a single value.

PR #17483, opened 2019-05-10
Out of space notification + icon + in widget

jakob223:
Probably review #17449 first.

Issue: KBFS-4073


shared/reducers/fs.js
```diff
@@ -116,6 +117,25 @@ const updatePathItem = (
...
+    case 'warning':
+      // TODO: propogate correct number of GB from setting
+      NotifyPopup('Disk Space Low', {body: 'You have less than 1 GB of storage space left.'})
+      break
+    case 'ok':
+      break
+    default:
+      Flow.ifFlowComplainsAboutThisFunctionYouHaventHandledAllCasesInASwitch(diskSpaceStatus)
+  }
+}
```
songgao: The reducer doesn't seem like the right place for this. Can we move it into Saga?

shared/constants/types/notifications.js
```diff
@@ -2,8 +2,16 @@
...
+export type NotificationKeys = 'kbfsUploading' | 'outOfSpace'
+export type BadgeType =
+  | 'regular'
+  | 'update'
+  | 'badged'
+  | 'error'
+  | 'uploading'
+  | 'badged-update'
+  | 'badged-uploading'
+  | 'badged-error'
```
songgao: I'm a bit confused. Does this PR add a second icon to the widget, so there could be a badged count and a status icon at the same time? If so it seems like we could have get rid of badged state, and the `badged-` prefix, and rely on `desktopAppBadgeCount` to determine if we should show the number. Then `BadgeType` just controls the status icon. How does that sound?

shared/actions/fs/index.js
```diff
@@ -935,6 +935,15 @@ const onFSOverallSyncSyncStatusChanged = (state, action) =>
...
   })
 
+let prevOutOfSpace = false
+const updateMenubarIconOnStuckSync = (state, action) => {
+  const outOfSpace = action.payload.params.status.outOfSyncSpace
+  if (outOfSpace !== prevOutOfSpace) {
+    prevOutOfSpace = outOfSpace
+    return NotificationsGen.createBadgeApp({key: 'outOfSpace', on: outOfSpace})
+  }
+}
```
songgao: It'd be nice if we could do this sort of deduplication in Go side.

shared/constants/types/notifications.js
jakob223: Yeah, it lets there be both at once. That seems reasonable, I think.

shared/actions/fs/index.js
jakob223: The Go side is sending us overall status notifications once per second, which include the current space status and the current sync status. Currently the space status is just a boolean value for whether it's out of space or not - it doesn't seem worth it to me to complicate that to (out of space, not out of space, unchanged).

shared/reducers/fs.js
jakob223: You're right, I'll move it.

shared/menubar/remote-proxy.desktop.js
```diff
@@ -8,23 +8,96 @@ import * as SafeElectron from '../util/safe-electron.desktop'
...
+    componentDidMount() {
+      this._updateBadges()
+
+      if (isDarwin && SafeElectron.getSystemPreferences().subscribeNotification) {
+        SafeElectron.getSystemPreferences().subscribeNotification(
+          'AppleInterfaceThemeChangedNotification',
+          () => {
+            this._updateBadges()
+          }
+        )
```
songgao: Seems like we should unsubscribe in `componentDidUnmount`

shared/menubar/space-warning.js
```diff
@@ -43,13 +43,16 @@ class SpaceWarning extends React.PureComponent<Props, State> {
...
+              ? 'Your computer has less than 1 GB of storage space. Make some' +
+                ' space, or unsync some folders.'
+              : 'Your computer is out of space. Unsync some folders, or make some space and'
           }
           color={this.props.diskSpaceStatus === 'warning' ? 'blue' : 'red'}
           actions={[
             ...(this.props.onRetry ? [{onClick: this.props.onRetry, title: 'retry' + ' the sync.'}] : []),
           ]}
+          narrow={true}
+          style={{minHeight: 50}}
```
songgao: Is this to add extra padding to top/bottom? If so is there a reason not to just make the padding larger in Kb.Banner?

shared/actions/fs/index.js
```diff
@@ -935,6 +936,35 @@ const onFSOverallSyncSyncStatusChanged = (state, action) =>
...
+      Flow.ifFlowComplainsAboutThisFunctionYouHaventHandledAllCasesInASwitch(diskSpaceStatus)
+  }
+}
+
+let prevOutOfSpace = false
+const updateMenubarIconOnStuckSync = (state, action) => {
+  const outOfSpace = action.payload.params.status.outOfSyncSpace
+  if (outOfSpace !== prevOutOfSpace) {
+    prevOutOfSpace = outOfSpace
+    // TODO once go side sends info: low on space warning
```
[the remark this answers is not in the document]
jakob223: I will do this as part of KBFS-4175

shared/menubar/space-warning.js
jakob223: This was a workaround because the padding already present was being hidden behind the things above and below the banner in the widget.

songgao: Do things above and below have `fullWidth={true}`? Either way, I guess this is fine for now.

shared/actions/fs/index.js
```diff
@@ -988,6 +989,35 @@ const onFSOverallSyncSyncStatusChanged = (state, action) =>
...
+      NotifyPopup('Disk Space Low', {body: 'You have less than 1 GB of storage space left.'})
+      break
+    case 'ok':
+      break
+    default:
+      Flow.ifFlowComplainsAboutThisFunctionYouHaventHandledAllCasesInASwitch(diskSpaceStatus)
+  }
+}
+
+let prevOutOfSpace = false
```
jzila: @jakob223 why is this a file-scoped variable instead of being in the state?

jakob223: Hmm, I'm not really sure I have a good reason. Feel free to move it to the state, or I can put up a PR for that in a bit if you want.

jzila: No worries, I'll do it in my notification PR

PR #17484, opened 2019-05-10
engine: fix merkle audit retries before checkpoint

maxtaco:


maxtaco:
cc @xgess this was your bug

xgess:
:cool: thanks :)

PR #17487, opened 2019-05-10
Support for SEP7 at service/cli levels

patrickxb:
This adds support for the SEP7 URI protocol.

Basic idea is that upon receiving one of these URIs, the client will call

    validateStellarURILocal(int sessionID, string inputURI);

which makes sure that the URI is signed by a key that matches the origin_domain in the URI.

The client will then present some detailed information to the user about the request and ask for confirmation (and any needed inputs like possibly the amount).  If the user approves it, the client will call one of

    approveTxURILocal
    approvePayURILocal
    approvePathURILocal

at which point the service will sign the transaction and either submit it to the stellar network or post it to a callback URL.

In the case that a non-XLM asset is in the URI, the client will make a findPaymentPath call first and confirm the max send amount with the user before calling approvePathURILocal.

Note that there is nothing in here registering the app as a handler of web+stellar: URIs.  That will be handled by Desktop when they integrate.  There is a CLI command `keybase wallet handle-uri <uri>` that can be used.  There is also a tool in tools called `sep7uri` that can make these URIs.


go/client/cmd_wallet_handle_uri.go
```diff
@@ -0,0 +1,301 @@
...
+
+	ui.Printf(c.green("URI validated.") + "\n\n")
+	ui.Printf("Keybase validated the URI you submitted.  It was signed by\n\n")
+	ui.Printf("\t%s\n\n", ColorString(c.G(), "underline", v.OriginDomain))
+	if v.Message != "" {
+		ui.Printf("Message: %q\n\n", v.Message)
+	}
+	if v.Memo != "" {
+		if v.MemoType == "MEMO_TEXT" {
+			ui.Printf("Public memo: %q\n\n", v.Memo)
```
mlsteele: Is this susceptible to Antonio-style trickery with tty control characters in the memo? Maybe that was remedied in ui.Printf, I don't remember.

go/client/cmd_wallet_handle_uri.go
```diff
@@ -0,0 +1,301 @@
...
+	ui.Printf("\t%s\n\n", v.Xdr)
+	ui.Printf("Transaction summary:\n\n")
+	if v.Summary.Source != "" {
+		ui.Printf("\tSource account: %s\n", v.Summary.Source)
+	}
+	if v.Summary.Fee > 0 {
+		ui.Printf("\tFee: %d stroops\n", v.Summary.Fee)
+	}
+	if v.Summary.Memo != "" {
+		ui.Printf("\tMemo: %q (%s)\n", v.Summary.Memo, v.Summary.MemoType)
```
mlsteele: memo question again

go/stellar/stellarsvc/service.go
```diff
@@ -595,10 +599,309 @@ func (s *Server) BatchLocal(ctx context.Context, arg stellar1.BatchLocalArg) (re
...
+		}
+
+		_, senderSeed, err := stellar.LookupSenderSeed(mctx)
+		if err != nil {
+			return "", err
+		}
+		var memoText string
+		if vp.MemoType == "MEMO_TEXT" {
+			memoText = vp.Memo
+		}
```
mlsteele: Does this silently drop non-text memos?

go/client/cmd_wallet_handle_uri.go
patrickxb: This CLI command is temporary because we don't have frontend support yet.

go/stellar/stellarsvc/service.go
patrickxb: No, there is an error earlier (and a ticket for us to support all types).

go/client/cmd_wallet_handle_uri.go
mlsteele: can it be unlisted then if it might have that problem?

go/stellar/stellarsvc/service.go
mlsteele: Seems like an easy place to forget. What about a switch with a default return error? Or at least a comment with the ticket number

patrickxb: the `handle-uri` has a comment with a ticket number, i'll add here...the ticket is in the current sprint, waiting for this to be approved :) ...

patrickxb:
```
    if v.Memo != "" {
                if v.MemoType == "MEMO_TEXT" {
                        ui.Printf("Public memo: %q\n\n", v.Memo)
                } else {
                        // CORE-10865 will remedy this situation...
                        ui.Printf("Request contains a public memo with an unsupported type (%s): %q\n", v.MemoType, v.Memo)
                        ui.Printf("At this time, Keybase does not support that memo type.\n")

                        // shouldn't proceed if we can't include it in the transaction.
                        return nil
                }
        }
```

PR #17491, opened 2019-05-10
fs: start CR banner

jakob223:
- also add feature flag

Issue: KBFS-4149

### DIFF
--- a/shared/fs/banner/conflict-banner-container.js
+++ b/shared/fs/banner/conflict-banner-container.js
@@ -1,0 +1,50 @@
+// @flow
+import * as React from 'react'
+import * as Constants from '../../constants/fs'
+import * as Types from '../../constants/types/fs'
+import {namedConnect} from '../../util/container'
+import ConflictBanner, {getHeight} from './conflict-banner'
+import * as RowTypes from '../row/types'
+import openUrl from '../../util/open-url'
+
+type OwnProps = {|path: Types.Path, conflictState: Types.ConflictState|}
+
+const mapStateToProps = (state, {path}: OwnProps) => ({
+  _tlf: Constants.getTlfFromPath(state.fs.tlfs, path),
+})
+
+const mapDispatchToProps = dispatch => ({
+  onFeedback: () => {},
+  onFinishResolving: () => {},
+  onHelp: () => openUrl('https://keybase.io/docs/kbfs/understanding_kbfs#conflict_resolution'),
+  onSeeOtherView: () => {},
+  onStartResolving: () => {},
+})
+
+const mergeProps = (s, d, o) => ({
+  ...s,
+  ...d,
+  ...o,
+  isUnmergedView: Constants.isUnmergedView(o.path),
+})
+
+const ConnectedBanner = namedConnect<OwnProps, _, _, _, _>(
+  mapStateToProps,
+  mapDispatchToProps,
+  mergeProps,
+  'ConflictBanner'
+)(ConflictBanner)
+
+export default ConnectedBanner
+
+export const asRows = (path: Types.Path, conflictState: Types.ConflictState): Array<RowTypes.HeaderRowItem> =>
+  conflictState === 'none'
+    ? []
+    : [
+        {
+          height: getHeight(conflictState),
+          key: 'conflict-banner',
+          node: <ConnectedBanner path={path} conflictState={conflictState} />,
+          rowType: 'header',
+        },
+      ]
--- a/shared/fs/banner/conflict-banner-container.js
+++ b/shared/fs/banner/conflict-banner-container.js
@@ -1,13 +1,11 @@
 // @flow
-import * as React from 'react'
 import * as Constants from '../../constants/fs'
 import * as Types from '../../constants/types/fs'
 import {namedConnect} from '../../util/container'
-import ConflictBanner, {getHeight} from './conflict-banner'
-import * as RowTypes from '../row/types'
+import ConflictBanner from './conflict-banner'
 import openUrl from '../../util/open-url'
 
-type OwnProps = {|path: Types.Path, conflictState: Types.ConflictState|}
+type OwnProps = {|path: Types.Path|}
 
 const mapStateToProps = (state, {path}: OwnProps) => ({
   _tlf: Constants.getTlfFromPath(state.fs.tlfs, path),
@@ -22,10 +20,10 @@ const mapDispatchToProps = dispatch => ({
 })
 
 const mergeProps = (s, d, o) => ({
-  ...s,
   ...d,
-  ...o,
+  conflictState: s._tlf.conflict.state,
   isUnmergedView: Constants.isUnmergedView(o.path),
+  tlfName: s._tlf.name,
 })
 
 const ConnectedBanner = namedConnect<OwnProps, _, _, _, _>(
@@ -36,15 +34,3 @@ const ConnectedBanner = namedConnect<OwnProps, _, _, _, _>(
 )(ConflictBanner)
 
 export default ConnectedBanner
-
-export const asRows = (path: Types.Path, conflictState: Types.ConflictState): Array<RowTypes.HeaderRowItem> =>
-  conflictState === 'none'
-    ? []
-    : [
-        {
-          height: getHeight(conflictState),
-          key: 'conflict-banner',
-          node: <ConnectedBanner path={path} conflictState={conflictState} />,
-          rowType: 'header',
-        },
-      ]
--- a/shared/fs/banner/conflict-banner-container.js
+++ b/shared/fs/banner/conflict-banner-container.js
@@ -1,22 +1,20 @@
 // @flow
 import * as Constants from '../../constants/fs'
 import * as Types from '../../constants/types/fs'
+import * as FsGen from '../../actions/fs-gen'
 import {namedConnect} from '../../util/container'
 import ConflictBanner from './conflict-banner'
 import openUrl from '../../util/open-url'
 
 type OwnProps = {|path: Types.Path|}
 
-const mapStateToProps = (state, {path}: OwnProps) => ({
-  _tlf: Constants.getTlfFromPath(state.fs.tlfs, path),
-})
-
-const mapDispatchToProps = dispatch => ({
+const mapStateToProps = () => ({})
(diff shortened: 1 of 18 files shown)


shared/fs/banner/conflict-banner.js
```diff
@@ -0,0 +1,102 @@
...
+      return (
+        'Your changes to this folder conflict with changes made to this' +
+        ' folder on the server. We are trying to fix it automatically, but' +
+        ' you may need to manually resolve the conflict. This is not' +
+        ' supposed to happen!'
+      )
+    case 'in-manual-resolution':
+      return isUnmerged
+        ? 'This is your local view of the conflicted folder.'
+        : "This is the server's view of the conflicted folder."
```
songgao: Several thoughts on the wording:

1. I wonder if we should avoid mentioning "server", not only because of #design's opinion, but also that it'd probably give a wrong impression that server can see the content somehow. How about something like "other devices's view".

2. I feel like `in-conflict-stuck` should say "you need to manually resolve" instead of "you may need to manually resolve". `in-conflict-not-stuck` should probably not mention manual resolution. I think neither should say "This is not supposed to happen" as that sounds like a bug, but I can be convinced otherwise.

3. It seems `in-manual-resolution` could use some more text -- something like suggesting user to click "finish resolving" when they're done when in server view, and "go to server view to resolve it" when in local view.

shared/fs/folder/index.js
```diff
@@ -30,6 +33,7 @@ const WithContent = (props: Props) => (
         routePath={props.routePath}
         headerRows={[
           ...resetBannerAsRows(props.path, props.resetBannerType),
+          ...conflictBannerAsRows(props.path, props.conflictState),
```
songgao: I'm inclined to having this as a fixed banner rather than a row that scrolls with the content. WDYT?

shared/fs/banner/conflict-banner.js
songgao: Also, since conflict is in TLF level, perhaps it'd be helpful to say "/keybase/private/jakob223" instead of "this folder". Related, we could probably make it more clear that user is currently in a subfolder of the conflict TLF, so they don't mistakenly mark a resolution as finished when they are only done with a particular dir within the TLF.

shared/fs/banner/conflict-banner-container.js
```diff
@@ -0,0 +1,50 @@
...
+
+const mapStateToProps = (state, {path}: OwnProps) => ({
+  _tlf: Constants.getTlfFromPath(state.fs.tlfs, path),
+})
+
+const mapDispatchToProps = dispatch => ({
+  onFeedback: () => {},
+  onFinishResolving: () => {},
+  onHelp: () => openUrl('https://keybase.io/docs/kbfs/understanding_kbfs#conflict_resolution'),
+  onSeeOtherView: () => {},
```
songgao: How about just do `onSeeServerView` and `onSeeLocalView` here? I think the explicit ones are easier to follow.

shared/fs/banner/conflict-banner-container.js
```diff
@@ -0,0 +1,50 @@
...
+  onStartResolving: () => {},
+})
+
+const mergeProps = (s, d, o) => ({
+  ...s,
```
songgao: I don't think you need `...s` here.

shared/fs/banner/conflict-banner-container.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+import * as React from 'react'
+import * as Constants from '../../constants/fs'
+import * as Types from '../../constants/types/fs'
+import {namedConnect} from '../../util/container'
+import ConflictBanner, {getHeight} from './conflict-banner'
+import * as RowTypes from '../row/types'
+import openUrl from '../../util/open-url'
+
+type OwnProps = {|path: Types.Path, conflictState: Types.ConflictState|}
```
songgao: `conflictState` is available from `_tlf`, so it seems we should just have the banner accept a `path`.

shared/fs/banner/conflict-banner.js
jakob223: > I think neither should say "This is not supposed to happen" as that sounds like a bug, but I can be convinced otherwise
This isn't supposed to happen, though - and we want to know about it when it happens. If we don't say something like that, I feel like users may not have any reason to click "let us know".

jakob223: > 

Also, since conflict is in TLF level, perhaps it'd be helpful to say "/keybase/private/jakob223" instead of "this folder". Related, we could probably make it more clear that user is currently in a subfolder of the conflict TLF, so they don't mistakenly mark a resolution as finished when they are only done with a particular dir within the TLF.

Hmm, this is a good point. I wonder if we should only let them click finish from the TLF itself.. WDYT of that?

jakob223: > `in-conflict-not-stuck` should probably not mention manual resolution

I think it should, because I imagine that most CR conflicts will continue to be unresolved after 10 tries if they are unresolved after one try. Maybe we should lower the limit from 10 to 5 or something too..

shared/fs/folder/index.js
jakob223: Yeah I think you're right.. we should have a more standard system for dealing with all our banners.. I feel like I've been just sticking them semirandomly into different levels of the component hierarchy depending on where they appear in the UI, but this seems suboptimal.

shared/fs/banner/conflict-banner-container.js
jakob223: I need it for the `conflictState`. Though I might move collecting that into the container now that this is not going to be a row anymore..

shared/fs/banner/conflict-banner-container.js
jakob223: It was done like this so that `asRows` would always have the right height. But since it's not going to scroll, we can clean it up :)

shared/fs/banner/conflict-banner.js
songgao: > This isn't supposed to happen, though - and we want to know about it when it happens. If we don't say something like that, I feel like users may not have any reason to click "let us know".

The "not stuck" situation isn't really not supposed to happen right? It happens all the time; but just get resolved automatically quickly.

> Hmm, this is a good point. I wonder if we should only let them click finish from the TLF itself.. WDYT of that?

That sounds good with me. I think we should be careful to not mention the term TLF though -- most users don't know what it is and we haven't introduced that concept in GUI before.

> I think it should, because I imagine that most CR conflicts will continue to be unresolved after 10 tries if they are unresolved after one try. Maybe we should lower the limit from 10 to 5 or something too..

Related to the first point; I was under the impression that this happens underneath automatically all the time. For example when you're behind the server and have the local chages in the journal, and they get "rebased" to latest version before being pushed. So it seems to me most of the "not stuck" conflict will never need manual intervention. I could be wrong though; maybe @strib could clarify

jakob223: As I understand it, the CR state would not change to `in-conflict-not-stuck` in that automatic "rebase" situation. Only once the CR fails once and we record a panic or an error, would we enter the conflict state. At that point CR might try again (when another change happens in the TLF, or when KBFS restarts after having died due to a panic) and maybe succeed if it failed due to running out of memory, though likely fail for the same reason as before. Once this trying again has happened 10 times, we enter `in-conflict-stuck` and auto-fail CR.

jzila:
ping, what do we need to do to take this across the finish line?

jakob223:
I read back through all the unresolved feedback and I think I've fixed everything. @songgao what do you think?

jzila:
Just need to address the conflicts.

songgao:
Sorry for the lack of follow-up on this. I was waiting for Strib's input on that not-stuck thing. Just summoned him in-person.

shared/fs/banner/conflict-banner.js
strib: @songgao is right, it switches to `in-conflict-not-stuck` as soon as it notices a conflict, and before it gets stuck.  But, there might not be a notification to the UX in that case until it becomes stuck, so you'd only see it if you happen to poll the favorites during that time.  I don't recommend having a badge or a banner in that state right now, as it doesn't require any user interaction and should be a very brief state, unless it becomes stuck.

shared/fs/banner/conflict-banner.tsx
```diff
@@ -0,0 +1,92 @@
...
+const getMessage = (tlf: string, conflictState: Types.ConflictState, isUnmerged): string => {
+  switch (conflictState) {
+    case Types.ConflictState.InConflictStuck:
+      return (
+        `Your changes to ${tlf} conflict with changes made to this` +
+        ' folder on another device. Automatic conflict resolution has failed,' +
+        ' so you need to manually resolve the conflict. This is not' +
+        ' supposed to happen!'
+      )
+    case Types.ConflictState.InConflictNotStuck:
```
strib: Yeah let's not show anything in this case.  This state _is_ supposed to happen, whenever there's a conflict.

jakob223: Cool, sorry perpetuating my misunderstanding @songgao. Will fix then resolve conflicts.

jakob223:
Ok, @songgao / @jzila I think this is ready to merge. What say you?

shared/fs/banner/conflict-banner.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const onSeeGlobalView = {onClick: props.onSeeOtherView, title: 'See the global view'}
+  const onSeeLocalView = {onClick: props.onSeeOtherView, title: 'See local changes'}
+
+  let actions = []
+  switch (props.conflictState) {
+    case Types.ConflictState.InConflictStuck:
+      actions = [startRes, feedbackAction, helpAction]
+      break
+    case Types.ConflictState.InConflictNotStuck:
+      actions = [startRes, feedbackAction, helpAction]
```
songgao: This and the line in `None` case could be gone now right?

jakob223: we need to return something in those cases, I'd rather keep the `Let us know` in the event of a future bug where we show the "this is not supposed to happen" banner.

PR #17492, opened 2019-05-10
Don't let confirm form header get too squished on mobile

buoyad:
image

r? @keybase/react-hackers


buoyad:
Also upped the height a bit on desktop and fixed a memo wrapping problem.

PR #17497, opened 2019-05-11
conflict_resolver: don't panic on an empty CR DB

strib:
Perhaps the user is running two kbfsfuse instances, and couldn't open the DB on one of them so it's nil.

Issue: KBFS-4160


strib:
{quote}
Hm, any guesses on why I didn't hit this bug when testing multiple KBFS at once after enabling the CRDB in this case?
{quote}

@jakob223 it only happens when you explicitly ask for `.kbfs_status`, which maybe you didn't do in your testing?

PR #17504, opened 2019-05-12
Fix errors in text

junderw:
There are nations with populations less than 10 recognized by the UN.

Stellar employees use more electricity than 10 people living in a small rural country.

Therefore the text is an error.


junderw:
Unless we have some data on the electricity consumption of each Stellar employee and the building they occupy...

In which case a proper comparison to the country with the smallest energy consumption can be made.

Is there data on this subject?

junderw:
https://twitter.com/ofnumbers/status/1128866873318223872?s=09

Looks like Stellar should maybe use a little more electricity... :-/

dabura667:
Yeah, it seems like Stellar has some sort of daddy issues...

It would be like anti-vaxxers citing some study and saying "OUR KIDS ARE ALL NATURAL AND IMMUNE SYSTEMS ARE STRONG WITH ESSENTIAL OILS!!!" as if it's some sort of burn on non-anti-vaxxers.

It's not a good look on Stellar, they should stop it. And Keybase should encourage them by merging this pr.

dabura667:
bump

junderw:
Bumping.

This factual error in the text should be fixed.

PR #17512, opened 2019-05-13
feedback tab on desktop

joshblum:
Issue: CORE-10790
Issue: KBFS-4166


shared/settings/feedback/index.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+export const getOtherErrorInfo = (err: Error) => {
```
chrisnojima: what is this file for. we don't want to mask the index.desktop/index.native files

shared/settings/feedback/index.desktop.js
```diff
@@ -0,0 +1,125 @@
+// @flow
+
+import React, {Component} from 'react'
```
chrisnojima: this and the native one should be unified and not split

shared/settings/feedback/container.desktop.js
```diff
@@ -0,0 +1,125 @@
...
+  title: 'Feedback',
+})
+
+const connected = compose(
+  connect<OwnProps, _, _, _, _>(
+    mapStateToProps,
+    mapDispatchToProps,
+    (s, d, o) => ({...s, ...d})
+  ),
+  HOCTimers
```
chrisnojima: if you're tracking mounted yourself due to your promises you don't need hoctimers

shared/settings/feedback/container.desktop.js
```diff
@@ -0,0 +1,125 @@
...
+        sendLogs={this.state.sendLogs}
+        onChangeSendLogs={this._onChangeSendLogs}
+      />
+    )
+  }
+}
+
+const mapStateToProps = state => {
+  return {
+    chat: extraChatLogs(state),
```
chrisnojima: this will render thrash

shared/settings/feedback/container.desktop.js
```diff
@@ -0,0 +1,125 @@
...
+import {writeLogLinesToFile} from '../../util/forward-logs'
+import {extraChatLogs, type State, type Props} from './utils'
+import * as RPCTypes from '../../constants/types/rpc-gen'
+
+type OwnProps = RouteProps<{}, {}>
+
+class FeedbackContainer extends React.Component<Props, State> {
+  mounted = false
+
+  state = {
```
chrisnojima: if you unify the index you don't need this container w/ state

shared/settings/feedback/container.desktop.js
```diff
@@ -0,0 +1,125 @@
...
+
+  _onSendFeedback = () => {
+    this.setState({sending: true, sentFeedback: false})
+    this.props.setTimeout(() => {
+      const maybeDump = this.state.sendLogs ? this._dumpLogs() : Promise.resolve('')
+      maybeDump
+        .then(() => {
+          logger.info(`Sending ${this.state.sendLogs ? 'log' : 'feedback'} to daemon`)
+          const extra = this.state.sendLogs ? {...this.props.status, ...this.props.chat} : this.props.status
+          return RPCTypes.configLogSendRpcPromise({
```
chrisnojima: we never make rpc calls directly in components

shared/settings/feedback/container.desktop.js
joshblum: this was copied from the native file, what should happen here instead?

shared/settings/feedback/index.js
joshblum: oops this was meant to be in the `util.js` file but will merge the desktop/native index files

shared/settings/feedback/container.desktop.js
joshblum: this was copied from the native file, is there a reason it was there to begin with?

jakob223:
@joshblum feel free to take a look and make sure this still looks right

chrisnojima:
👀

shared/reducers/settings.js
```diff
@@ -105,6 +105,14 @@ function reducer(state: Types.State = initialState, action: SettingsGen.Actions)
       return state.update('email', email => email.merge({error: action.payload.error}))
     case SettingsGen.waitingForResponse:
       return state.merge({waitingForResponse: action.payload.waiting})
+    case SettingsGen.sendFeedbackError:
+      return state.update('feedback', feedback => feedback.set('error', action.payload.error))
+    case SettingsGen.setFeedback:
+      return state.update('feedback', feedback => feedback.set('feedback', action.payload.feedback))
+    case SettingsGen.setSendLogs:
+      return state.update('feedback', feedback => feedback.set('sendLogs', action.payload.sendLogs))
+    case SettingsGen.sentFeedback:
```
jakob223: I realized I never actually trigger this action. Will add to the action.

shared/settings/feedback/index.js
```diff
@@ -0,0 +1,150 @@
...
+  onChangeFeedback: (nextValue: ?string) => void,
+}
+type State = {
+  showSuccessBanner: boolean,
+}
+
+class Feedback extends React.Component<Props, State> {
+  state = {showSuccessBanner: false}
+  componentDidUpdate(prevProps: Props) {
+    if (prevProps.sending && !this.props.sending) {
```
chrisnojima: collapse into
```
if (prevProps.sending !== this.props.send) {
  this.setState({showSuccessBanner: !this.props.sending})
}
```

shared/settings/feedback/index.js
```diff
@@ -0,0 +1,150 @@
...
+    } = this.props
+    return (
+      <Kb.Box2 direction="vertical" fullWidth={true} alignItems="center">
+        <Kb.ScrollView>
+          {this.state.showSuccessBanner && (
+            <Kb.Box2
+              alignItems="center"
+              direction="horizontal"
+              fullWidth={true}
+              style={{
```
chrisnojima: lets not inline all these styles, use styleSheetCreate instead

shared/actions/json/settings.json
```diff
@@ -121,6 +121,18 @@
...
       "checkPasswordIsCorrect": "?boolean"
-    }
+    },
+    "setFeedback": {
+      "feedback": "?string"
+    },
+    "setSendLogs": {
+      "sendLogs": "boolean"
+    },
+    "sendFeedback": {},
```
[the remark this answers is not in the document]
jakob223: should I keep track of the feedback and the checkbox in the component's own state? or is there some way of retrieving them from the subcomponents when dispatching the `sendFeedback` action?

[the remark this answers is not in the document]
chrisnojima: local state. nothing else cares about this. the component is just keeping track so it can submit it to the rpc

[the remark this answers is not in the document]
jakob223: The KBFS CR banner contains a link that wants to pre-fill some feedback and default the checkbox to on - is there a way to do that without putting it in the global store?

Given that, do you still think this should move to a local store?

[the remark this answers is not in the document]
jakob223: (resolved offline: will move to local state, use navigationProps for the other thing.

shared/actions/json/settings.json
```diff
@@ -121,6 +121,14 @@
     },
     "loadedCheckPassword": {
       "checkPasswordIsCorrect": "?boolean"
-    }
+    },
+    "sendFeedback": {
+      "feedback": "?string",
```
chrisnojima: should just be a string

PR #17513, opened 2019-05-13
delete: simplify delete flow, change upload banner copy

jakob223:
Issue: KBFS-4112

### DIFF
--- a/shared/fs/footer/upload.desktop.js
+++ b/shared/fs/footer/upload.desktop.js
@@ -53,10 +53,10 @@ const Upload = ({showing, files, fileName, totalSyncingBytes, timeLeft, debugTog
           <Kb.Text key="files" type="BodySemibold" style={styles.textOverflow}>
             {files
               ? fileName
-                ? `Encrypting and uploading ${fileName}...`
-                : `Encrypting and uploading ${files} files...`
+                ? `Encrypting and updating ${fileName}...`
+                : `Encrypting and updating ${files} files...`
               : totalSyncingBytes
-              ? 'Encrypting and uploading...'
+              ? 'Encrypting and updating files...'
               : 'Done!'}
           </Kb.Text>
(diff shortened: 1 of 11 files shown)


shared/fs/footer/upload.desktop.js
```diff
@@ -53,10 +53,10 @@ const Upload = ({showing, files, fileName, totalSyncingBytes, timeLeft, debugTog
           <Kb.Text key="files" type="BodySemibold" style={styles.textOverflow}>
             {files
               ? fileName
-                ? `Encrypting and uploading ${fileName}...`
-                : `Encrypting and uploading ${files} files...`
+                ? `Encrypting and updating ${fileName}...`
+                : `Encrypting and updating ${files} files...`
```
songgao: This should be "Encrypting and updating ${files} items...".

shared/fs/footer/upload.desktop.js
```diff
@@ -53,10 +53,10 @@ const Upload = ({showing, files, fileName, totalSyncingBytes, timeLeft, debugTog
...
               : totalSyncingBytes
-              ? 'Encrypting and uploading...'
+              ? 'Encrypting and updating files...'
```
songgao: Same here; I think we should drop "files" or replace with "items"

shared/util/feature-flags.js.flow
```diff
@@ -4,7 +4,6 @@ export type FeatureFlags = {|
   airdrop: boolean,
   chatIndexProfilingEnabled: boolean,
   dbCleanEnabled: boolean,
-  enableDeleteFolder: boolean,
```
songgao: I think design would need a QA pass before we can release this, and likely we'll need to tweak the modal popup. If you get rid of this flag in this PR, could you follow-up with design and make sure we either get it right before next release?

PR #17514, opened 2019-05-13
fs: open finder after SFMI enable

jakob223:
It's still not coming to the forefront but I think that's a different problem

Issue: KBFS-4145

### DIFF
--- a/shared/actions/fs/platform-specific.desktop.js
+++ b/shared/actions/fs/platform-specific.desktop.js
@@ -169,13 +169,17 @@ const fuseStatusToActions = (previousStatusType: 'enabled' | 'disabled' | 'unkno
     ? [
         FsGen.createSetDriverStatus({
           driverStatus: Constants.makeDriverStatusEnabled({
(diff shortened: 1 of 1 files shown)


shared/actions/fs/platform-specific.desktop.js
```diff
@@ -169,13 +169,16 @@ const fuseStatusToActions = (previousStatusType: 'enabled' | 'disabled' | 'unkno
     ? [
         FsGen.createSetDriverStatus({
           driverStatus: Constants.makeDriverStatusEnabled({
-            dokanOutdated: status.installAction === 2,
+            dokanOutdated: status.installAction === 2, // 2 == UPGRADE
```
MarcoPolo: I think you meant to use this:
https://github.com/keybase/client/blob/970bb5d56e9ad815846eb3b2a0e246b84232f6d8/shared/constants/types/rpc-gen.js#L490

`status.installAction === RPCTypes.installInstallAction.upgrade`

PR #17517, opened 2019-05-13
Add withSafeNavigation HOC

buoyad:
- Adds a param `fromKey` to `navigateUp` and `navigateAppend` and bail if the active screen's key doesn't match
- Adds `withSafeNavigation` that injects wrapped versions of those actions so component doesn't have to get key manually
- Make header back button do safe nav on desktop

Things this doesn't do:

- Give these wrapped fcns to route screen components too
  - I tried to do this at the shim layer and ran into an error:
    ```TypeError: No "routes" found in navigation state. Did you try to pass the navigation prop of a React component to a Navigator child?```
    Only happened when going to the wallets or settings tabs on desktop (i.e. tabs with subnavigators). The error pointed to [here](https://reactnavigation.org/docs/en/custom-navigators.html#navigator-navigation-prop) which suggests that react navigation does some introspection to switch what `props.navigation` is when the child is a navigator, so wrapping it in a shim breaks it.
  - I don't think this is a big loss, as using the HOC is easy, and it may even be better not to have the component's `OwnProps` depend on where it is in the tree

~Note: this doesn't pass flow because `navigateUp()` needs to take an empty object now. I didn't want to add a ton of noise to this diff, so I'll make it pass in a second PR.~    fixed in 47653e2

r? @keybase/react-hackers


shared/router-v2/router.shared.js
```diff
@@ -71,6 +72,15 @@ export const oldActionToNewActions = (action: any, navigation: any, allowAppendD
         }
       }
 
+      if (action.payload.fromKey) {
+        const {fromKey} = action.payload
+        const activeKey = getActiveKey(navigation.state)
+        if (fromKey !== activeKey) {
+          console.log('Skipping append on wrong screen')
```
MarcoPolo: logger.warn?

shared/util/safe-navigation.js
```diff
@@ -0,0 +1,53 @@
...
+        <Component
+          ref={forwardedRef}
+          {...rest}
+          navigateAppend={this._navigateAppend}
+          navigateUp={this._navigateUp}
+        />
+      )
+    }
+  }
+  const WithForwardRef = React.forwardRef((props, ref) => (
```
MarcoPolo: :sheep: Might be nice to name this? https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools

shared/util/safe-navigation.js
```diff
@@ -0,0 +1,53 @@
...
+    _navigateAppend = ({path, replace}) =>
+      RouteTreeGen.createNavigateAppend({fromKey: getActiveKey(this.props.navigation.state), path, replace})
+
+    _navigateUp = () => RouteTreeGen.createNavigateUp({fromKey: getActiveKey(this.props.navigation.state)})
+
+    render() {
+      const {navigation, forwardedRef, ...rest} = this.props
+      return (
+        <Component
+          ref={forwardedRef}
```
MarcoPolo: Just curious, do we currently use a ref to the component anywhere?

buoyad: We don't

PR #17522, opened 2019-05-13
Pending state while waiting for a convo to resolve

chrisnojima:
@keybase/react-hackers this partially resurrects the idea of a pending waiting conversation but makes it much looser / simpler.
We already had a sentry value for this in constants so we just set it when we're going to resolve a conversation and select it. In that mode we show a simple 'loading...' text so it doesn't get annoying when if it flashes
Unlike before we don't do any kind of mapping between pending convos and resolving them, we just select the convo when its done and if we're looking at the pending we replace it

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -1570,8 +1570,13 @@ const confirmScreenResponse = (_, action) => {
 // We always make adhoc convos and never preview it
 const previewConversationPersonMakesAConversation = (state, action) =>
   !action.payload.teamname &&
-  action.payload.participants &&
-  Chat2Gen.createCreateConversation({participants: action.payload.participants})
+  action.payload.participants && [
+    Chat2Gen.createSelectConversation({
+      conversationIDKey: Constants.pendingWaitingConversationIDKey,
+      reason: 'justCreated',
+    }),
+    Chat2Gen.createCreateConversation({participants: action.payload.participants}),
+  ]
 
 // We preview channels
 const previewConversationTeam = (state, action) => {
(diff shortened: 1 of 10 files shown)


shared/actions/chat2/index.js
```diff
@@ -1568,10 +1568,19 @@ const confirmScreenResponse = (_, action) => {
...
 
 // We always make adhoc convos and never preview it
-const previewConversationPersonMakesAConversation = (state, action) =>
-  !action.payload.teamname &&
-  action.payload.participants &&
-  Chat2Gen.createCreateConversation({participants: action.payload.participants})
+const previewConversationPersonMakesAConversation = (state, action) => {
+  const participants = action.payload.participants
+  return (
+    !action.payload.teamname &&
```
buoyad: :sheep: destructure with `participants`

shared/actions/chat2/index.js
```diff
@@ -2064,8 +2077,20 @@ const navigateToInbox = (state, action, logger) => {
...
+  let replace = false
+
+  const visible = Router2Constants.getVisibleScreen()
+  // looking at the pending screen?
+  if (
+    visible?.routeName === 'chatConversation' &&
+    visible?.params?.conversationIDKey === Constants.pendingWaitingConversationIDKey
+  ) {
+    replace = true
+  }
```
buoyad: oh, a back button fix here regressed with deprecating nav1
https://github.com/keybase/client/pull/17439/files#diff-62d6b6a8baeb0bb4a699ed1965bcbf00R2110

shared/chat/conversation/messages/special-top-message.js
```diff
@@ -79,6 +91,7 @@ type OwnProps = {
 const mapStateToProps = (state, ownProps: OwnProps) => {
   const hasLoadedEver = state.chat2.messageOrdinals.get(ownProps.conversationIDKey) !== undefined
   const meta = Constants.getMeta(state, ownProps.conversationIDKey)
+  const waitingPending = ownProps.conversationIDKey === Constants.pendingWaitingConversationIDKey
```
buoyad: :sheep: `pendingWaiting` for consistency

shared/actions/chat2/index.js
chrisnojima: fixed and pushed tnx

PR #17523, opened 2019-05-13
Hide 'review payments' link on error banner if this is an exchange rate err

buoyad:
r? @keybase/react-hackers

### DIFF
--- a/shared/wallets/banner/index.js
+++ b/shared/wallets/banner/index.js
@@ -2,7 +2,7 @@
 import * as React from 'react'
 import {Box2, Text} from '../../common-adapters'
 import type {Background} from '../../common-adapters/text'
-import {backgroundModeToColor, backgroundModeToTextColor, collapseStyles, globalMargins, styleSheetCreate} from '../../styles'
+import * as Styles from '../../styles'
 
 type Props = {
   background: Background,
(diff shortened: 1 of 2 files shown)


shared/wallets/banner/index.js
```diff
@@ -35,7 +43,7 @@ const Banner = (props: Props) => (
         </Text>
       )}
     </Text>
-    {props.sendFailed && (
+    {props.sendFailed && props.onAction && (
```
buoyad: I only changed this and L34, the rest is prettier / `* as Styles`

shared/wallets/confirm-form/container.js
```diff
@@ -64,23 +49,37 @@ export default connect<OwnProps, _, _, _, _>(
   mapStateToProps,
   mapDispatchToProps,
   (stateProps, dispatchProps) => {
+    const {_built, _sentPaymentError} = stateProps
+    const exchangeRateChanged = _sentPaymentError && _sentPaymentError.toLowerCase().includes('exchange rate')
+    const banners = (_sentPaymentError
```
buoyad: I noticed this was thrashing because it was created in `mapStateToProps`

PR #17530, opened 2019-05-13
mtime-based TLF sorting and use SimpleFS fav API

songgao:
Commits are individually reviewable.

@keybase/kbfs-hackers for Go code, and @keybase/react-hackers for JS.


songgao:
I think I know what went wrong with tests. Will fix. Untagging kbfs hackers now

go/kbfs/libkbfs/favorites_test.go
```diff
@@ -163,23 +171,30 @@ func TestFavoritesAddRemoveAdd(t *testing.T) {
...
 	}
 
+	config.mockKbpki.EXPECT().FavoriteDelete(gomock.Any(), folder1).
+		Return(nil)
+	config.mockKbpki.EXPECT().FavoriteList(gomock.Any()).Return(keybase1.FavoritesResult{}, nil)
 	if err := f.Delete(ctx, fav1.Folder); err != nil {
 		t.Fatalf("Couldn't delete favorite: %v", err)
 	}
 
-	// A second add shouldn't result in a KBPKI call
```
songgao: I couldn't think of a reason why the second add shouldn't result in a KBPKI call. That sounds like a bug we had. So I got rid of this comment and added an `FavoriteAdd`. But shout if anyone thinks I'm wrong here!

go/kbfs/libkbfs/favorites.go
```diff
@@ -342,7 +350,40 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
...
+	for _, fav := range req.toAdd {
+		// FOR DISCUSSION: we don't check this in toDel because it could be
+		// out-of-date. So perhaps we should get rid of this too? Relevant
+		// question, if a favorite is deleted from another device, do we get
+		// notification from core for cache invalidation?
+		_, present := f.favCache[fav.Folder]
+		if !fav.Created && present {
+			continue
+		}
+		//
```
songgao: Question for @keybase/kbfs-hackers in the comment above.

strib: I believe this check for adds is critical and we should definitely leave it in.  We've had issues in the past with spamming the API server with adding the same favorite multiple times.  We don't have the same problem with deletes, because after the user deletes it, they aren't accessing the folder again.  But with adds, we could be going through this code on basically every folder access.

Yes, favorite deletes from another device result in a notification to this device.

go/kbfs/libkbfs/favorites.go
```diff
@@ -342,7 +350,40 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
...
+	for _, fav := range req.toDel {
+		// Since our cache isn't necessarily up-to-date, always delete
+		// the favorite.
+		folder := fav.ToKBFolder(false)
+		err := kbpki.FavoriteDelete(req.ctx, folder)
+		if err != nil {
+			return err
+		}
+		f.config.UserHistory().ClearTLF(tlf.CanonicalName(fav.Name), fav.Type)
+		needFetch = true
```
strib: Why do we need to fetch in this case?  Don't we only need to fetch in order to get the mtime for favorited TLFs?

go/kbfs/libkbfs/favorites_test.go
strib: YOU'RE WRONG!  (Sorry, you asked for a shout.)  For the reasons I outlined above -- we don't want to spam the server with multiple identical add calls.  Why does your code change this though, the check for not duplicating add calls is still there as far as I can tell, right?

go/kbfs/libkbfs/favorites_test.go
```diff
@@ -261,6 +277,7 @@ func TestFavoritesListFailsDuringAddAsync(t *testing.T) {
 }
 
 func TestFavoritesControlUserHistory(t *testing.T) {
+	t.Skip()
```
strib: Is this something you meant to address before putting up for review?

go/kbfs/libkbfs/favorites_test.go
songgao: haha thanks. I actually don't know why it used to pass. Note that there's a Delete right above this, on the same folder, which also went to the API server. So this needs another RPC to add it back, right?

go/kbfs/libkbfs/favorites.go
songgao: Sounds good. I was mostly worried about the races. But with notifications it seems the race is not possible. Will steal your response here as code comment.

go/kbfs/libkbfs/favorites_test.go
songgao: whooops I marked all the broken ones with this first. I guess I missed one. will fix.

songgao:
> Question: does the GUI need invalidation notifications when the mtime of an existing favorite changes? Seems like that would need a bit more work -- the edit history code could invalidate favorites or update the mtime directly perhaps.

hmm without such notification, does it mean as long as there's no other favorites changes, mtime can be cached up to a week? That sounds bad. If it's more work, perhaps we can also just make it expire more often? Edit history triggered invalidation sounds good to me too.

go/kbfs/libkbfs/favorites.go
songgao: I was trying to avoid editing the cache directly and always derive it from an RPC to core. I feel like it's easier to reason about. This just goes to service and is not expensive anyway right?

go/kbfs/libkbfs/favorites_test.go
strib: Oohhh, sorry, I just searched for "// A second add" in `favorites_test.go` and was looking at the first result in `TestFavoritesAddTwice`, so was responding based on that.

In this case there's already a `Times(2)` on the add call above -- I guess the comment was just copy and pasted, and wrong. So with your code, it's being called 3 times now?  That seems incorrect too.

strib:
> hmm without such notification, does it mean as long as there's no other favorites changes, mtime can be cached up to a week? That sounds bad. If it's more work, perhaps we can also just make it expire more often? Edit history triggered invalidation sounds good to me too.

Yes, the mtimes would be cached up to a week, which is obviously bad.  If the GUI isn't doing something to force a refresh whenever switching into that tab or something, we should probably hook into the edit history to make it do something.  But I'm ok with that being a separate PR as long as we get it in before releasing this to the public.

go/kbfs/libkbfs/favorites.go
strib: I see.  I guess it's not expensive for regular users, but might be for something like keybasepub if it ever deleted favorites.  Plus it just adds another round trip time for no reason.  I think I'd prefer to just delete it from the cache here and not refresh if there was no other change.  I can't see how that would cause any problems.

go/kbfs/libkbfs/favorites_test.go
songgao: Got it. That makes sense. That `.Times(2)` is removed actually, so sounds like this is correct as-in!

songgao: (I think I removed that `Times(2)` at some point when fixing the tests but forgot about that when I made the comment here, which is why I thought we had a bug)

strib: Ah got it.

songgao:
OK; I'd prefer to get this in first and do that in a separate PR. If we fail to get that one in before next release we can feature flag off mtime based sorting in GUI.

go/kbfs/libkbfs/favorites.go
songgao: OK will do

songgao:
@strib ready for another look!

songgao:
@keybase/react-hackers bump! Go side is approved.

songgao:
> seems weird to me that "older first" shows empty TLFs at the top. Any thoughts on this?

I think this is because those have a Mtime of 0. Good point, and I guess we can also make them always appear in the end. But then that gives an impression that they don't exist if you have a lot of TLFs. So I'm a bit torn on which way to go.

> I'm getting some bouncing-around behavior as sync states change. Might be a sync bug? If you're not experiencing this, I'm willing to call it that and debug it with offline mode.

Hmm by bouncing-around do you mean the affected TLF bounces up *and* down? We don't have the Mtime-triggered cache invalidation yet, so that _could_ be related. It seems we could wait until that's in, and it'll be easier to debug.

songgao:
@jakob223 I was suspecting the reducer to have a bug that could have caused the "bouncing" behavior that you saw, but didn't find any. Although I did realize I missed `tlf.conflict` when updating `updateTlf`, so pushed another commit for it. PTAL thanks!

PR #17532, opened 2019-05-13
dont load messages if we're just showing up due to a nav switch

chrisnojima:
@keybase/react-hackers 
before if you select a convo it'd try and reload it. but if we go to reactjis or info panel and click back it'd do this which is pretty annoying

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -922,6 +922,10 @@ function* loadMoreMessages(state, action, logger) {
     case Chat2Gen.selectConversation:
       key = action.payload.conversationIDKey
       reason = action.payload.reason || 'selected'
+      if (reason === 'focused') {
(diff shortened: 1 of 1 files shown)


mmaxim:
Should try this scrolled up as a result of a search, it might break it. There is an assumption in the reducer that when select conversation happens, that we load messages. This will break that assumption.

mmaxim:
In other words, why generate this action at all with this reason?

chrisnojima:
This only affects this action with this reason

mmaxim:
In the reducer for this action there is this:
```
  // ignore non-changing
      if (state.selectedConversation === action.payload.conversationIDKey) {
        return state
      }
```
Perhaps it is more consistent to use the same logic in the saga.

chrisnojima:
Let's discuss tomorrow

PR #17533, opened 2019-05-13
wrap long usernames/teamnames in convo menu header

buoyad:
r? @keybase/react-hackers


buoyad:
@keybase/react-hackers bump

PR #17534, opened 2019-05-13
folder_block_ops: defer/re-apply dir updates for non-syncing files

strib:
A user hit this scenario:

* Create file X.
* Sync the creation of that file.
* Write a whole bunch of other files in the same directory.
* Start a new `SyncAll`.
* Then another write comes in for file X, during the `SyncAll`. The write doesn't get deferred because file X isn't being synced. But the directory metadata update related to the file gets lost after the `SyncAll` completes and the metadata operations are cleared, leading to the file seeming to be truncated to 0 bytes again.

So instead, during a `SyncAll` make sure we track all the directory updates for files that aren't currently being synced, and re-apply them after the sync completes successfully.

Also add a test that repros the issue without the fix in place.

Issue: KBFS-4165

### DIFF
--- a/go/kbfs/libkbfs/folder_block_ops.go
+++ b/go/kbfs/libkbfs/folder_block_ops.go
@@ -216,7 +216,9 @@ type folderBlockOps struct {
 
 	// dirtyDirs track which directories are currently dirty in this
 	// TLF.
-	dirtyDirs map[data.BlockPointer][]data.BlockInfo
+	dirtyDirs          map[data.BlockPointer][]data.BlockInfo
+	dirtyDirsSyncing   bool
+	deferredDirUpdates []func(lState *kbfssync.LockState) error
 
 	// dirtyRootDirEntry is a DirEntry representing the root of the
 	// TLF (to be copied into the RootMetadata on a sync).
@@ -1485,6 +1487,25 @@ func (fbo *folderBlockOps) updateEntryLocked(ctx context.Context,
 	} else {
 		_ = fbo.makeDirDirtyLocked(lState, parentPath.TailPointer(), unrefs)
 	}
+
+	// If we're in the middle of syncing the directories, but the
+	// current file is not yet being synced, we need to re-apply this
+	// update after the sync is done, so it doesn't get lost after the
+	// syncing directory block is readied.  This only applies to dir
+	// updates being caused by file changes; other types of dir writes
+	// are protected by `folderBranchOps.syncLock`, which is held
+	// during `SyncAll`.
+	if fbo.dirtyDirsSyncing && !fbo.doDeferWrite {
+		fbo.log.CDebugf(ctx, "Deferring update entry during sync")
+		n := fbo.nodeCache.Get(file.TailRef())
+		fbo.deferredDirUpdates = append(
+			fbo.deferredDirUpdates, func(lState *kbfssync.LockState) error {
+				file := fbo.nodeCache.PathFromNode(n)
+				return fbo.updateEntryLocked(
+					ctx, lState, kmd, file, de, includeDeleted)
+			})
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


go/kbfs/libkbfs/folder_block_ops.go
```diff
@@ -1485,6 +1487,25 @@ func (fbo *folderBlockOps) updateEntryLocked(ctx context.Context,
...
 	}
+
+	// If we're in the middle of syncing the directories, but the
+	// current file is not yet being synced, we need to re-apply this
+	// update after the sync is done, so it doesn't get lost after the
+	// syncing directory block is readied.  This only applies to dir
+	// updates being caused by file changes; other types of dir writes
+	// are protected by `folderBranchOps.syncLock`, which is held
+	// during `SyncAll`.
+	if fbo.dirtyDirsSyncing && !fbo.doDeferWrite {
```
jakob223: Is there a race whereby `fbo.dirtyDirsSyncing` is `false` here, but `true` when the update would normally happen above, and so it gets wiped out but we don't notice?

go/kbfs/libkbfs/folder_block_ops.go
```diff
@@ -216,7 +216,9 @@ type folderBlockOps struct {
 
 	// dirtyDirs track which directories are currently dirty in this
 	// TLF.
-	dirtyDirs map[data.BlockPointer][]data.BlockInfo
+	dirtyDirs          map[data.BlockPointer][]data.BlockInfo
+	dirtyDirsSyncing   bool
+	deferredDirUpdates []func(lState *kbfssync.LockState) error
```
jakob223: Why did you decide to make this a slice of arbitrary functions? It doesn't seem like the wrong choice per se, but it seems hard to debug.

strib: The place that calls the function doesn't have all the state it needs about nodes and such, so it's either make a closure like this, or buildout an explicit param struct that saves all the arguments and pass that around instead.  Either seems valid to me, and this was easier and in line with other defer-like functions in this file.  Not sure it impacts debuggability but happy to discuss more if you feel strongly about it.

go/kbfs/libkbfs/folder_block_ops.go
strib: No, `fbo.dirtyDirsSyncing` is protected by the same lock that is required for calling this function, so it can't change between the above code and here.  (If you're asking something else, let me know -- not sure if I understood the question completely.)

jakob223: I think I slightly misunderstood the purpose of `dirtyDirsSyncing` when I was writing this, and am now convinced the race that I was thinking of does not exist.

strib:
Bump for @jzila / @songgao review -- it would be nice to get this in soonish to bake before the next release.

jzila:
On it.

go/kbfs/libkbfs/folder_block_ops.go
```diff
@@ -1485,6 +1487,25 @@ func (fbo *folderBlockOps) updateEntryLocked(ctx context.Context,
...
+	// syncing directory block is readied.  This only applies to dir
+	// updates being caused by file changes; other types of dir writes
+	// are protected by `folderBranchOps.syncLock`, which is held
+	// during `SyncAll`.
+	if fbo.dirtyDirsSyncing && !fbo.doDeferWrite {
+		fbo.log.CDebugf(ctx, "Deferring update entry during sync")
+		n := fbo.nodeCache.Get(file.TailRef())
+		fbo.deferredDirUpdates = append(
+			fbo.deferredDirUpdates, func(lState *kbfssync.LockState) error {
+				file := fbo.nodeCache.PathFromNode(n)
```
jzila: Is there a chance this will no longer be in the cache?

go/kbfs/libkbfs/folder_block_ops.go
```diff
@@ -2422,6 +2457,16 @@ func (fbo *folderBlockOps) clearAllDirtyDirsLocked(
 	}
 	fbo.dirtyDirs = make(map[data.BlockPointer][]data.BlockInfo)
 	fbo.dirtyRootDirEntry = nil
+	fbo.dirtyDirsSyncing = false
```
jzila: This is all locked so it doesn't matter, but IMO the `dirtyDirsSyncing` and `deferredDirUpdates` assignments should be together. After the for loop below.

go/kbfs/libkbfs/folder_block_ops.go
strib: Nope -- it only gets garbage collections when there are no more references to it, and the closure itself is holding a reference at least.  Plus the node for the file being written is held by the caller somewhere, and it will have a reference to its parent as well.

go/kbfs/libkbfs/folder_block_ops.go
strib: It has to be done together _before_ the loop, not after, since the deferred functions rely on it being false.  But I'll move the other one up and just copy the value there to a temp var for iterating.

jzila: So it does, thanks.

PR #17537, opened 2019-05-13
stub out twitter pvl responses

pzduniak:


pzduniak:
tested out by adding `127.0.0.1 mobile.twitter.com` to `/etc/hosts`

go/libkb/api_stub_html.go
```diff
@@ -0,0 +1,577 @@
+package libkb
+
+var htmlAPIStubs = map[string]string{
```
zapu: might as well try to integrate `vcr` in client codebase, what do you think @patrickxb ?

patrickxb: sure, it's already used in client for stellar.

but whatever's easiest.  if this is done, then that's fine...

(also, i know we already have/had this somewhere for github and maybe twitter...maybe it was deleted)

patrickxb: oh, it's this api_stub.go file.  

@zapu i say just keep using this api_stub system for now and move on to better things...

zapu: sounds good. I didn't know it was used in client, I couldn't find it. I thought it was only in stellard and stellarnet

pzduniak: @patrickxb merge ok? we don't have to wait for Max if you think this solution is fine.

go/libkb/api_stub_html.go
```diff
@@ -0,0 +1,577 @@
+package libkb
```
patrickxb: this file needs a comment header, and a build tag so this doesn't get into production builds (like in api_stub.go)

pzduniak: corrected, wdyt now

PR #17543, opened 2019-05-14
Deactivate chat input when minWriterRole is insufficient.

adamjspooner:

### DIFF
--- a/shared/actions/chat2/index.js
+++ b/shared/actions/chat2/index.js
@@ -669,8 +669,8 @@ const onChatSetConvSettings = (_, action, logger) => {
   const role = newRole && TeamsConstants.teamRoleByEnum[newRole]
   const cannotWrite = conv?.convSettings?.minWriterRoleInfo?.cannotWrite
   logger.info(`got new minWriterRole ${role || ''} for convID ${conversationIDKey}`)
-  if (role && role !== 'none' && cannotWrite !== null) {
-    return Chat2Gen.createSaveMinWriterRole({conversationIDKey, cannotWrite, role})
+  if (role && role !== 'none' && cannotWrite !== undefined) {
+    return Chat2Gen.createSaveMinWriterRole({cannotWrite, conversationIDKey, role})
   }
   logger.warn(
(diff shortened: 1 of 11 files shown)


shared/chat/conversation/input-area/normal/platform-input.desktop.js
```diff
@@ -217,7 +222,8 @@ class _PlatformInput extends React.Component<PlatformInputPropsInternal, State>
                 ) : (
                   <Kb.Icon
                     className="timer"
-                    onClick={this._toggleShowingMenu}
+                    colorOverride={this.props.cannotWrite ? Styles.globalColors.black_20 : null}
+                    onClick={this.props.cannotWrite ? undefined : this._toggleShowingMenu}
```
adamjspooner: This is to prevent the hover styles.

shared/actions/chat2/index.js
```diff
@@ -665,15 +665,12 @@ const onChatSetConvRetention = (_, action, logger) => {
...
-    conv.convSettings &&
-    conv.convSettings.minWriterRoleInfo &&
-    conv.convSettings.minWriterRoleInfo.role
+  const newRole = conv?.convSettings?.minWriterRoleInfo?.role
   const role = newRole && TeamsConstants.teamRoleByEnum[newRole]
+  const cannotWrite = conv?.convSettings?.minWriterRoleInfo?.cannotWrite
   logger.info(`got new minWriterRole ${role || ''} for convID ${conversationIDKey}`)
-  if (role && role !== 'none') {
-    return Chat2Gen.createSaveMinWriterRole({conversationIDKey, role})
+  if (role && role !== 'none' && cannotWrite !== null) {
```
buoyad: just `&& cannotWrite`

adamjspooner: Even though it can be `false`?

buoyad: oops, right. then I think you should check for `undefined` (according to [here](https://github.com/tc39/proposal-optional-chaining#base-case))

PR #17547, opened 2019-05-14
kbfs: data plumbing for offline sync notification settings

jzila:
KBFS-4014


go/kbfs/libkbfs/init.go
```diff
@@ -228,7 +228,7 @@ func DefaultInitParams(ctx Context) InitParams {
 		DiskCacheMode:                  DiskCacheModeLocal,
 		DiskBlockCacheFraction:         0.10,
 		SyncBlockCacheFraction:         1.00,
-		Mode:                           InitDefaultString,
```
jakob223: intentional? did `go fmt` rules change at 1.12 and we're only formatting files as they're committed to?

go/kbfs/libkbfs/interfaces.go
```diff
@@ -139,6 +139,10 @@ type blockRetrieverGetter interface {
 	BlockRetriever() BlockRetriever
 }
 
+type settingsDBGetter interface {
```
jakob223: Does this need to be its own type instead of just being another method on the `Config` interface?

go/kbfs/libkbfs/settings_db.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+const (
+	// Where in config.StorageRoot() we store settings information.
+	settingsDBDir           = "kbfs_settings"
+	settingsDBVersionString = "v1"
+	settingsDBName          = "kbfsSettings.leveldb"
+
+	// Settings keys
+	notificationThresholdKey = "notificationThreshold"
```
jakob223: maybe a slightly more descriptive name would be good here, e.g. `spaceAvailableNotificationThreshold` like below, or something? no strong feelings though.

go/kbfs/libkbfs/settings_db.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return si.UID
+}
+
+func getSettingsDbKey(uid keybase1.UID, key string) []byte {
+	return append([]byte(uid), []byte(key)...)
+}
+
+// Get overrides (*LevelDb).Get to cache values in memory.
+func (db *SettingsDB) Get(key []byte, ro *opt.ReadOptions) ([]byte, error) {
```
jakob223: should `Put` and `Get` be exported? It seems like callers are not expected to use them.

go/kbfs/libkbfs/settings_db.go
```diff
@@ -0,0 +1,142 @@
...
+var ErrNoSettingsDB = errors.New("no settings DB")
+
+var errNoSession = errors.New("no session")
+
+// SettingsDB stores KBFS user settings for a given device.
+type SettingsDB struct {
+	*LevelDb
+	sessionGetter idutil.CurrentSessionGetter
+
+	cache map[string][]byte
```
jakob223: Why is there a cache here? Is this DB accessed often enough for it to be necessary?

shared/actions/fs/index.js
```diff
@@ -135,6 +135,20 @@ const setTlfSyncConfig = (state, action) =>
     })
   )
 
+const loadSettings = (state, action) =>
+  RPCTypes.SimpleFSSimpleFSSettingsRpcPromise()
```
jakob223: Should we catch the error here so the user doesn't mistake no settings DB for the settings are loading?

I'm not really sure about the conventions around catching RPC errors though, so I may be totally wrong.

shared/actions/fs/index.js
```diff
@@ -135,6 +135,20 @@ const setTlfSyncConfig = (state, action) =>
...
+      FsGen.createSettingsLoaded({
+        settings: Constants.makeSettings({
+          spaceAvailableNotificationThreshold: settings.spaceAvailableNotificationThreshold,
+        }),
+      })
+    )
+
+const setSpaceNotificationThreshold = (state, action) =>
+  RPCTypes.SimpleFSSimpleFSSetNotificationThresholdRpcPromise({spaceAvailableNotificationThreshold: action.payload.spaceAvailableNotificationThreshold})
+    .then
```
jakob223: did you intend to do something after `then`?

go/kbfs/libkbfs/settings_db.go
jzila: I anticipate it being accessed on every disk cache access. But levelDB might already have a cache.

shared/actions/fs/index.js
jzila: Nope nice catch.

go/kbfs/libkbfs/interfaces.go
jzila: no reason not to have it be its own type

go/kbfs/libkbfs/init.go
jzila: I think that's accurate.

go/kbfs/libkbfs/settings_db.go
jzila: They are already exported, as per the comment. I'm just overriding them here.

go/kbfs/libkbfs/settings_db.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, err
+	}
+
+	return openLevelDB(stor)
+}
+
+func openSettingsDB(config Config) *SettingsDB {
+	db, err := openSettingsDBInternal(config)
+	if err != nil {
+		config.MakeLogger("").CWarningf(context.Background(),
```
songgao: how about `MakeLogger("SETDB")`?

go/kbfs/libkbfs/settings_db.go
```diff
@@ -0,0 +1,142 @@
...
+// SetNotificationThreshold sets the notification threshold setting for the
+// logged-in user.
+func (db *SettingsDB) SetNotificationThreshold(
+	ctx context.Context, threshold int64) error {
+	uid := db.getUID(ctx)
+	if uid == keybase1.UID("") {
+		return errNoSession
+	}
+	return db.Put(getSettingsDbKey(uid, notificationThresholdKey),
+		[]byte(strconv.FormatInt(threshold, 10)), nil)
```
songgao: There's also `strconv.Itoa` but no need to change.

shared/actions/fs/index.js
```diff
@@ -135,6 +135,19 @@ const setTlfSyncConfig = (state, action) =>
...
+    .then(settings =>
+      FsGen.createSettingsLoaded({
+        settings: Constants.makeSettings({
+          spaceAvailableNotificationThreshold: settings.spaceAvailableNotificationThreshold,
+        }),
+      })
+    )
+
+const setSpaceNotificationThreshold = (state, action) =>
+  RPCTypes.SimpleFSSimpleFSSetNotificationThresholdRpcPromise({threshold: action.payload.spaceAvailableNotificationThreshold})
```
songgao: should probably have `.then(() => FsGen.createSettingsLoad())` here

jzila: I was debating how to do that, whether we wanted to handle it in the reducer or do a full settings load.

songgao: I think we should handle it through the RPC. If we do it in the reducer, there's a chance we'll have inconsistent data.

jzila: I mean whether to handle it via a `.then(() => createSomeNewAction)` and `.catch(createRevertStateAction)`, or do a full settings load.

PR #17552, opened 2019-05-14
Account settings with phone/email storybook

buoyad:
image

Also:
* Fix actions addon in native storybook
* Add `decoration` prop to menu items for stuff that goes to the right of the title

Part 2 will be "confirm remove email/phone" dialog

- [ ] Storyshots before merge

r? @keybase/react-hackers

### DIFF
--- a/shared/common-adapters/floating-menu/menu-layout/index.native.js
+++ b/shared/common-adapters/floating-menu/menu-layout/index.native.js
@@ -31,12 +31,14 @@ const MenuRow = (props: MenuRowProps) => (
     {props.view || (
       <>
         <Box2 direction="horizontal" fullWidth={true} centerChildren={true}>
+          {props.decoration && <Box style={styles.flexOne} />}
           <Text center={true} type="BodyBig" style={styleRowText(props)}>
             {props.title}
           </Text>
           {props.newTag && (
             <Meta title="New" size="Small" backgroundColor={Styles.globalColors.blue} style={styles.badge} />
           )}
+          {props.decoration && <Box style={styles.flexOne}>{props.decoration}</Box>}
         </Box2>
         {!!props.subTitle && (
           <Text center={true} type="BodyTiny">
@@ -118,6 +120,9 @@ const styles = Styles.styleSheetCreate({
   },
   flexGrow: {
     flexGrow: 1,
+  },
+  flexOne: {
+    flex: 1,
   },
   itemContainer: {
     ...Styles.globalStyles.flexBoxColumn,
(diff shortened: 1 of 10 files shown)


shared/common-adapters/floating-menu/menu-layout/index.native.js
```diff
@@ -31,12 +31,14 @@ const MenuRow = (props: MenuRowProps) => (
     {props.view || (
       <>
         <Box2 direction="horizontal" fullWidth={true} centerChildren={true}>
+          {props.decoration && <Box style={styles.flexOne} />}
```
buoyad: this is a nice solution for having a centered item and something on the side without having to set `width`s. A box with `flex: 1` on both sides will make sure the center item stays centered.

shared/settings/account/index.js
```diff
@@ -0,0 +1,120 @@
...
+
+const AccountSettings = (props: Props) => {
+  let passwordLabel = props.hasPassword ? 'Change password' : 'Set a password'
+  if (props.hasPassword && Styles.isMobile) {
+    passwordLabel = 'Change'
+  }
+  return (
+    <Kb.ScrollView>
+      <Kb.Box2 direction="vertical" fullWidth={true} fullHeight={true}>
+        {/* Email + Phone */}
```
chrisnojima: if its worth commenting then i think its worth breaking out into its own component in this file

shared/settings/account/index.js
```diff
@@ -0,0 +1,120 @@
...
+  hasPassword: boolean,
+  onAddEmail: () => void,
+  onAddPhone: () => void,
+  onDeleteAccount: () => void,
+  onSetPassword: () => void,
+|}
+
+const AccountSettings = (props: Props) => {
+  let passwordLabel = props.hasPassword ? 'Change password' : 'Set a password'
+  if (props.hasPassword && Styles.isMobile) {
```
chrisnojima: 🐑 
```
const passwordLabel = props.hasPassword ? `Change${Styles.isMobile ? '' : ' password'}` : 'Set a password'
```
or
```
let passwordLabel
if (props.hasPassword) {
  passwordLabel = Styles.isMobile ? 'Change' : 'Change password'
} else {
 passwordLabel = 'Set a password'
}
```

i find logic to choose and then logic to overwrite more confusing

shared/settings/account/email-phone-row.js
```diff
@@ -0,0 +1,174 @@
...
+    top: -2,
+  },
+  badgeMenuItem: {
+    alignSelf: 'center',
+    marginLeft: 'auto',
+  },
+  container: {
+    height: Styles.isMobile ? 48 : 40,
+  },
+  flexOne: {
```
chrisnojima: this should go into globalStyles

buoyad:
@chrisnojima updated PTAL

PR #17555, opened 2019-05-15
Delete Team support in the GUI

cjb:
@keybase/react-hackers 

Mostly straightforward, with some things to note:

* The service calls out to `teamsUi` for confirmation, so I added a `customResponseIncomingCallMap` to confirm automatically

* I wanted to use the `<ConfirmModal />` common-adapter, but I also wanted to disable the delete button until all three checkboxes in the confirm form are checked, so I changed the common-adapter to disable its confirm button if `onConfirm` is null, and passed that through.

* There should be a `canUserPerform.deleteTeam` to handle deciding whether to show this (CORE-10908).  In the meantime I added an "isOwner && hasNoSubteams` check in the JS.

* The UX copy mentions "team members will be notified", but I'm not sure that actually happens (CORE-10916).  I could remove that line from the checkboxes while we decide what the notification should look like, or this PR could hang out until then if that's a blocker.


shared/constants/teams.js
```diff
@@ -319,6 +320,10 @@ const getTeamMemberCount = (state: TypedState, teamname: Types.Teamname): number
 const isLastOwner = (state: TypedState, teamname: Types.Teamname): boolean =>
   isOwner(getRole(state, teamname)) && !isMultiOwnerTeam(state, teamname)
 
+const canDeleteTeam = (state: TypedState, teamname: Types.Teamname): boolean =>
+  // FIXME: Use getCanPerform().deleteTeam once it exists: CORE-10908.
+  isOwner(getRole(state, teamname)) && !teamHasSubteams(state, teamname)
```
buoyad: We should be able to delete subteams too - I'm not sure what the rules are for that (do you need to be an owner of the root team?)

shared/teams/team/menu-container.js
```diff
@@ -36,6 +37,13 @@ const mapDispatchToProps = (dispatch, {teamname}: OwnProps) => ({
         path: [{props: {makeSubteam: true, name: teamname}, selected: 'teamNewTeamDialog'}],
       })
     ),
+  onDeleteTeam: () =>
+    dispatch(
+      RouteTreeGen.createNavigateTo({
```
buoyad: just append to `teamDeleteTeam`, no need to support nav1 anymore

cjb:
@chrisnojima @buoyad Ready for another look! The deletedTeams badgeState went in and I turn it into a:
```js
      case 'deletedTeam':
        const {deletedBy, teamName} = item.team
        return (
          <Kb.Banner
            color="blue"
            key={'deletedTeamBannerFor' + teamName}
            text={`The ${teamName} team was deleted by ${deletedBy}.`}
          />
        )
```

shared/actions/teams.js
```diff
@@ -1429,11 +1450,11 @@ const teamsSaga = function*(): Saga.SagaGenerator<any, any> {
     'getChannelInfo'
   )
   yield* Saga.chainAction<TeamsGen.GetChannelsPayload>(TeamsGen.getChannels, getChannels, 'getChannels')
-  yield* Saga.chainGenerator<ConfigGen.BootstrapStatusLoadedPayload, TeamsGen.GetTeamsPayload, TeamsGen.LeftTeamPayload>(
-    [ConfigGen.bootstrapStatusLoaded, TeamsGen.getTeams, TeamsGen.leftTeam],
-    getTeams,
-    'getTeams'
-  )
+  yield* Saga.chainGenerator<
+    ConfigGen.BootstrapStatusLoadedPayload,
```
chrisnojima: these should be `|` together

shared/teams/delete-team/index.js
```diff
@@ -0,0 +1,108 @@
...
+export type Props = {
+  onBack: () => void,
+  onDelete: () => void,
+  teamname: string,
+}
+
+const Header = (props: Props) => (
+  <>
+    <Kb.Avatar teamname={props.teamname} size={64} />
+    <Kb.Icon type="icon-team-delete-28" style={{marginRight: -60, marginTop: -20, zIndex: 1}} />
```
chrisnojima: don't think zindex is necessary

shared/teams/delete-team/index.js
```diff
@@ -0,0 +1,108 @@
...
+  box1: boolean,
+  box2: boolean,
+  box3: boolean,
+|}
+type CheckboxesProps = {|
+  onAllowDelete: boolean => void,
+|}
+
+class Checkboxes extends React.Component<CheckboxesProps, CheckboxesState> {
+  state = {
```
chrisnojima: i think its confusing to have this state and the state above in the parent where you're having to mange that by using didUpdate. I think it'd be simpler if this thing just took props for box1/2/3 and had click handlers from the parent and canDelete goes away and is just `this.state.box1 && this.state.box2 && this.state.box3`

cjb:
@chrisnojima Thanks, ready for another look!

PR #17558, opened 2019-05-15
Change user/lookup API call for deleted users

zapu:


zapu:
I think there will be similar issue here:

https://github.com/keybase/client/blob/master/go/libkb/resolve.go#L333-L334

PR #17567, opened 2019-05-15
improve regex for link parsing

mmaxim:


go/chat/utils/utils.go
```diff
@@ -2075,9 +2075,22 @@ func DecorateBody(ctx context.Context, body string, offset, length int, decorati
 	return res, added
 }
 
-var linkRegexp = xurls.RelaxedAtDomain()
+var linkRegexp = xurls.Relaxed()
+var linkRelaxedRegexpIndex = 0
```
joshblum: comment as to what these are would be good

joshblum: `linkRelaxedGroupIndex`

PR #17572, opened 2019-05-15
[Android HOTFIX] Fix paste on TextInput

MarcoPolo:
@keybase/react-hackers @chrisnojima 

This fixes paste by extending RN's TextInput this lets us avoid adding this logic to a fork. But it should be noted that this is implicitly tied to the implementation of the specific version of RN Text Input. If RN changes how TextInput works this will likely also need to be updated...

### DIFF
--- a/shared/common-adapters/input.native.js
+++ b/shared/common-adapters/input.native.js
@@ -4,16 +4,12 @@
 import React, {Component} from 'react'
 import Box from './box'
 import Text, {getStyle as getTextStyle} from './text'
-// import {NativeTextInput} from './native-wrappers.native'
+import NativeTextInput from './kb-text-input.native'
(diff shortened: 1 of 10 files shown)


shared/common-adapters/input.native.js
```diff
@@ -4,7 +4,7 @@
 import React, {Component} from 'react'
 import Box from './box'
 import Text, {getStyle as getTextStyle} from './text'
-import {NativeTextInput} from './native-wrappers.native'
```
chrisnojima: You should change this in native wrappers so it's pulled in by everything. We already patch native components there

MarcoPolo: 👍

Issue #17573, opened 2019-05-16
Can't access one of my teams (team link before user key revocation)


htdvisser:
I can't access one of my teams on any device. The CLI returns the following error:

```
ERROR proof error for proof 'team link before user key revocation': seqno 16 > 15
```

I think this could be related to another issue that @mlsteele resolved with https://github.com/keybase/client/blob/master/go/teams/loader2.go#L159-L168 in #12620.

my team id: `ttn.ops.deployments.web`
my log id: 5d356d8ff7012a714b1e981c

maxtaco:
We will take a look. Likely a bug and a serious one at that. Thank you!

maxtaco:
It looks like a server error allowed a revoked device to rotate this team. We'll have to fix the server error and hardcode an exception for this particular case. Can you wait a release cycle or is it urgent? We're very sorry about this, still investigating.

maxtaco:
(Good though that the clients catch the server error, as they should)

htdvisser:
Thanks for looking into this, @maxtaco. 

This particular team wasn't very actively used, so I can't say for sure when this issue first came up. That also means that a fix is not urgent, so we can wait for it to land in your regular release cycle.

Let me know if you need any additional information from us.

maxtaco:
Ok great! Should be fixed in the next release then, in about ~2 weeks. We try to release every 3 weeks but we have a company offsite next week, so that might slow us down a bit. Thank you and sorry for the bug.

htdvisser:
Sorry for leaving this issue open. It's been fixed, thanks!

cc: @mlsteele

PR #17583, opened 2019-05-16
teams: add exception for a rotate-after-revoke bug in the wild

maxtaco:
- the first was on the server, but add client support to clean up the mess


maxtaco:
Addresses #17573

PR #17585, opened 2019-05-16
strip go binaries on mobile

heronhaye:
https://blog.filippo.io/shrink-your-go-binaries-with-this-one-weird-trick/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -222,10 +222,7 @@ helpers.rootLinuxNode(env, {
                       // other than Go tests on Windows,
                       // add a `hasGoChanges` check here.
                       dir("go/keybase") {
-                        bat "go build --tags=production"
-                      }
-                      dir("go/keybase") {
-                        bat "go build"
+                        bat "go build -ldflags \"-s -w\" --tags=production"
                       }
                       testGo("test_windows_go_", getPackagesToTest(dependencyFiles))
                     }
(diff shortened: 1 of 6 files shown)


joshblum:
@heronhaye any reason not to do this on other desktop platforms?

heronhaye:
nope, just wanted to see on one first. there used to

joshblum:
I tried an iOS sim build and it seemed fine, I think it’s worth trying for
some admin builds

heronhaye:
The thing to make sure is if we get a segfault we still have full stacktrace in logs. It might be different on simulator vs real device too. Maybe we can make a panic button?

heronhaye:
Also, KBFS prints some traces from goroutines in their logs - should check those too.

joshblum:
will try this out:
```
$ git diff
diff --git a/go/keybase/main.go b/go/keybase/main.go
index 433fc4aa05..987da684ff 100644
--- a/go/keybase/main.go
+++ b/go/keybase/main.go
@@ -47,6 +47,17 @@ func keybaseExit(exitCode int) {
 }

 func main() {
+       go func() {
+               select {
+               case <-time.After(time.Second * 15):
+                       debug.PrintStack()
+               }
+               select {
+               case <-time.After(time.Second * 15):
+                       panic("panic")
+               }
+       }()
+
        // Preserve non-critical errors that happen very early during
        // startup, where logging is not set up yet, to be printed later
        // when logging is functioning.
```

joshblum:
log id `d9562ca381b476889c90ea1c`

heronhaye:
@jzila CI changes ok?

PR #17589, opened 2019-05-16
add deleteTeam to canUserPerform, cleanup

joshblum:


cjb:
@joshblum Could you describe the semantics you used? I haven't tried running this but I'd expect to be able to delete a team that I just created and has no other owners, and to be able to delete a subteam.

joshblum:
@cjb sorry i was mixing up not being able to leave a team without other owners.. it doesn't make sense for deletion. Thanks for catching!

go/teams/service_helper.go
```diff
@@ -1580,52 +1573,59 @@ func CanUserPerform(ctx context.Context, g *libkb.GlobalContext, teamname string
...
+	// team settings
+	ret.ListFirst = isImplicitAdmin
+	ret.JoinTeam = isNone && isImplicitAdmin
+	ret.SetPublicityAny = isAdmin || isImplicitAdmin
+	ret.ManageMembers = isAdmin || isImplicitAdmin
+	ret.ManageSubteams = isAdmin || isImplicitAdmin
+	ret.SetTeamShowcase = isAdmin || isImplicitAdmin
+	ret.ChangeOpenTeam = isAdmin || isImplicitAdmin
+	ret.ChangeTarsDisabled = isAdmin || isImplicitAdmin
+	ret.EditTeamDescription = isAdmin
```
mlsteele: Implicit admins ought to be able to edit team descriptions. I don't know how the team description actually works though. If it's based on chat then nevermind.

mlsteele:
You can't delete a team that has subteams
image

go/teams/service_helper.go
joshblum: I don't think it's via chat, just a prexisting bug

mlsteele:
To list subteams there's `team.chain().ListSubteams()`. But that may return an incomplete set due to stubbed links if the team wasn't loaded with NeedAdmin. So you could load the team again with NeedAdmin if `team.chain().HasAnyStubbedLinks()`.

joshblum:
@cjb what is the UX supposed to be if there are subteams? Do you want another bool to indicate this so the user knows why they can't delete?

joshblum:
maybe it would be better to show the user the option even if there are subteams so we don't have to reload the team on the canperform handler. If the user tries to delete the team`TeamListSubteamsRecursive` with `forceRepoll=false` would be run on the confirm dialogue. If the delete RPC fails with `keybase1.StatusCode_SCTeamHasLiveChildren`  `TeamListSubteamsRecursive` is run again with `forceRepoll=true`

cjb:
@joshblum would it be okay to just use forceRepoll=true when the confirm dialogue's loading?  I don't think teams are gonna get deleted often.

joshblum:
@cjb that's fine, the CLI calls with `forceRepoll=true` in both cases: https://github.com/keybase/client/blob/master/go/client/cmd_team_delete.go#L64-L78

so we'll leave the permission check as is and GUI will call the `TeamListSubteamsRecursive` before attempting the `TeamDelete` rpc?

go/teams/service_helper.go
joshblum: tested locally, implicit admins can change the description

cjb:
> so we'll leave the permission check as is and GUI will call the TeamListSubteamsRecursive before attempting the TeamDelete rpc?

👍

PR #17590, opened 2019-05-16
reorder and retry loading for new transactions in payment notifications

xgess:
Payments that come from outside of keybase (and to a lesser extent, from non-primary keybase accounts) are harder to lookup, and horizon is often several seconds behind the keybase notification system. This can cause the client to report a new balance but not show the transaction for multiple minutes. 

This change makes loading a payment from a notification synchronous. And move this step to before refreshing our local state on the account.

### DIFF
--- a/go/stellar/loader.go
+++ b/go/stellar/loader.go
@@ -244,7 +244,21 @@ func (p *Loader) runRequests() {
 	}
 }
 
-func (p *Loader) loadPayment(id stellar1.PaymentID) {
+func (p *Loader) LoadPaymentSync(ctx context.Context, paymentID stellar1.PaymentID) {
+	mctx := libkb.NewMetaContext(ctx, p.G())
+	defer mctx.TraceTimed(fmt.Sprintf("LoadPaymentSync(%s)", paymentID), func() error { return nil })()
+
+	for i := 1; i <= 100; i++ {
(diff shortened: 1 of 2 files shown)


go/stellar/loader.go
```diff
@@ -244,7 +244,21 @@ func (p *Loader) runRequests() {
...
+	mctx := libkb.NewMetaContext(ctx, p.G())
+	defer mctx.TraceTimed(fmt.Sprintf("LoadPaymentSync(%s)", paymentID), func() error { return nil })()
+
+	for i := 1; i <= 100; i++ {
+		err := p.loadPayment(paymentID)
+		if err == nil {
+			break
+		}
+		mctx.Debug("error on attempt %d to load payment %s: %s. sleep and retry.", i, paymentID, err)
+		time.Sleep(1 * time.Second)
```
patrickxb: what about using libkb/backoff.go?

100 attempts sounds like a lot...

patrickxb:
Nice work.  Yes, stellarmond often beats horizon ingestion.  I thought we had something else in place for that, but maybe not in this path...

go/stellar/loader.go
xgess: done. in this case, the fastest i've seen it load is a few seconds, but it's typically closer to 7.

PR #17591, opened 2019-05-16
add push token to status

joshblum:
example log send id `0a775d4db19334a37e3c391c`

### DIFF
--- a/shared/constants/settings.js
+++ b/shared/constants/settings.js
@@ -76,6 +76,8 @@ export const makeState: I.RecordFactory<Types._State> = I.Record({
   waitingForResponse: false,
 })
 
+export const getPushTokenForLogSend = (state: Object) => ({pushToken: state.push.token})
+
 export const getExtraChatLogsForLogSend = (state: Object) => {
   const chat = state.chat2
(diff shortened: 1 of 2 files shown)


shared/constants/settings.js
```diff
@@ -75,6 +75,12 @@ export const makeState: I.RecordFactory<Types._State> = I.Record({
   waitingForResponse: false,
 })
 
+export const getPushTokenForLogSend = (state: Object) => {
+  return I.Map({
```
chrisnojima: why make a map and convert it to an object?

joshblum: i was following the existing call below: https://github.com/keybase/client/pull/17591/files#diff-1d9ec48879ceb528287d545487be4738R89

chrisnojima: thats cause it had hierarchical immutable structures, which this does not. can be
```
=> ({pushToken: state.push.token})
```

joshblum: got it, thanks!

joshblum:
@chrisnojima this is ready

PR #17592, opened 2019-05-16
fix audio. clean up and unused file

chrisnojima:
@keybase/react-hackers fixes losing external audio while scrolling past a gif


chrisnojima:
also deleted an unused and dangerous file

PR #17602, opened 2019-05-16
Report secret store failures

mlsteele:


go/libkb/osv_windows.go
```diff
@@ -0,0 +1,12 @@
+// +build windows
+
+package libkb
+
+// OSVersionAndBuild returns OS version, and build too on some platforms
+func OSVersionAndBuild() (string, string, error) {
+	productVersion, err := execToString("cmd", []string{"/c", "ver"})
```
zapu: I thought @heronhaye worked on something like this before, maybe it could be used?

heronhaye: Looks like we have it in the status package and in the install package. Could both depend on a libkb helper but doesn't really matter.

zapu: looks like it is used, sorry! but factored out to libkb?

heronhaye: Ah didn't see rest of PR. Looks fine to me

go/libkb/secret_store.go
```diff
@@ -316,6 +317,11 @@ func (s *SecretStoreLocked) SetOptions(mctx MetaContext, options *SecretStoreOpt
 // store, retrieve, and then delete a secret with an arbitrary name. This should
 // be done before provisioning or logging in
 func PrimeSecretStore(mctx MetaContext, ss SecretStoreAll) (err error) {
+	defer func() {
+		if err != nil {
+			go reportPrimeSecretStoreFailure(mctx.BackgroundWithLogTags())
```
zapu: it will also get called in few of the tests but this is probably fine

go/libkb/osv_windows.go
mlsteele: In order to call it from libkb it had to move either into libkb or another package. Due to libkb->status->libkb

mlsteele:
it passed CI ! 🎈

PR #17603, opened 2019-05-16
init storage hooks when starting chat modules

joshblum:
prevents a deadlock if a hook is trying to be added during login/logout


joshblum:
(ci flake unrelated)

PR #17604, opened 2019-05-16
small fixes

chrisnojima:
@keybase/react-hackers i was doing some memory leak analysis and found some small things that should be changed

### DIFF
--- a/shared/chat/conversation/messages/wrapper/exploding-meta/index.js
+++ b/shared/chat/conversation/messages/wrapper/exploding-meta/index.js
@@ -31,6 +31,7 @@ class ExplodingMeta extends React.Component<Props, State> {
   state = {mode: 'none'}
   tickerID: TickerID
   sharedTimerID: SharedTimerID
+  sharedTimerKey: string
 
   componentDidMount() {
     this._hideOrStart()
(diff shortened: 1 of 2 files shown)


shared/chat/conversation/messages/wrapper/exploding-meta/index.js
```diff
@@ -44,8 +45,9 @@ class ExplodingMeta extends React.Component<Props, State> {
     if (this.props.exploded && !prevProps.exploded) {
       this.setState({mode: 'boom'})
       SharedTimer.removeObserver(this.props.messageKey, this.sharedTimerID)
+      this.sharedTimerKey = this.props.messageKey
```
chrisnojima: we should stash what was actually passed and not this prop based thing since that could change and our unregister could be incorrect

shared/chat/conversation/messages/wrapper/exploding-meta/index.js
```diff
@@ -80,6 +82,7 @@ class ExplodingMeta extends React.Component<Props, State> {
     }
     const interval = getLoopInterval(difference)
     if (interval < 1000) {
+      removeTicker(this.tickerID)
```
chrisnojima: always clean first

shared/chat/inbox/row/small-team/index.js
```diff
@@ -58,7 +58,8 @@ class SmallTeam extends React.PureComponent<Props, State> {
 
   _onMouseLeave = () => this.setState({isHovered: false})
   _onMouseOver = () => this.setState({isHovered: true})
-  _onShowMenu = (showMenu: boolean) => this.setState({showMenu})
+  _onForceShowMenu = () => this.settate({showMenu: true})
```
chrisnojima: use property and not inline closure

jzila: s/this.settate/this.setState, here and on the line below.

chrisnojima:
@keybase/react-hackers bump

PR #17613, opened 2019-05-17
prefetch hasrandompw state on login

heronhaye:


heronhaye:
Tested on both nopw and pw, in master the key is not found but in this branch the randompw key is in the db.

heronhaye:
merging over bech32 test failure that is fixed in master

PR #17614, opened 2019-05-17
log send on desktop: fix reducer, center

joshblum:


shared/reducers/settings.js
```diff
@@ -108,7 +108,9 @@ function reducer(state: Types.State = initialState, action: SettingsGen.Actions)
     case SettingsGen.feedbackSent:
       return state.update('feedback', feedback => feedback.set('error', action.payload.error))
     case SettingsGen.sendFeedback:
-      return state.update('feedback', feedback => feedback.set('error', null))
+      return state.update('feedback', feedback =>
+        feedback.merge({error: null, feedback: action.payload.feedback, sendLogs: action.payload.sendLogs})
```
jakob223: I took the feedback and the logs out of the redux state, they're just used in the action now. I guess I forgot to take them out of the redux state though, just never touch them anymore. Regardless, they get put in the action in `actions/settings.js` so they don't need to be touched here.

joshblum: the action was still referencing the state, i'll fix that instead

jakob223: whoops 🤦‍♂

PR #17615, opened 2019-05-17
fs: public folder redesign, except for avatars

jakob223:
Issue: KBFS-4172


shared/constants/fs.js
```diff
@@ -601,6 +595,17 @@ export const getTlfFromTlfs = (tlfs: Types.Tlfs, tlfType: Types.TlfType, name: s
...
+export const isOwnPublic = (p: Types.Path, username: string) => {
+  const parsedPath = parsePath(p)
+  switch (parsedPath.kind) {
+    case 'group-tlf':
+    case 'in-group-tlf':
+      return parsedPath.writers.contains(username)
+    default:
+      return false
+  }
+}
```
songgao: IIRC "your own public TLF" means you're the only writer there, so just `.contains` wouldn't suffice.

But perhaps we should show it in all public TLFs? @cecileboucheron wdyt?

shared/fs/banner/public-reminder.js
```diff
@@ -0,0 +1,47 @@
...
+    default:
+      return ''
+  }
+}
+
+class PublicBanner extends React.Component<Props, State> {
+  state = {closed: false}
+
+  _onClose = () => this.setState({closed: true})
+
```
songgao: I think this works only because we never have scenario where we go directly from one public TLF to another -- in which case this component will get be kept mounted and maintain the closed state, thus fail to bring back the banner. Seems we could have a `componentDidUpdate` here to check if path change results in TLF change, just to be safe.

We could also add something like "lastDismissedPublicBannerTlf" in the redux store, depending on when exactly the banner is supposed to come back. For example, should it come back if user is in the Files tab, goes to chat, and comes back to the Files tab and land in the same public folder? Should it come back if user goes back out to the TLF list from a public folder, then back into the same public folder?

shared/fs/banner/public-reminder.js
```diff
@@ -0,0 +1,47 @@
...
+  render() {
+    if (this.state.closed) {
+      return null
+    }
+    const url = getURL(this.props.path)
+    return (
+      <Kb.Banner
+        color="yellow"
+        text="Everything you upload in here can be viewed by everyone at "
+        actions={[{onClick: () => openUrl(url), title: url + '.'}]}
```
songgao: If you do [the thing that determines if the banner should show up] inside this component, you'd never need to worry about showing his while url is `''`. I don't feel very strongly about it though so up to you!

shared/constants/fs.js
jakob223: I think we should show it in any public TLFs you have edit access to - hence `contains`. But interested to hear what Cécile thinks.

shared/fs/banner/public-reminder.js
jakob223: Hmm, this is a good point. I'll put it in the redux store.

> For example, should it come back if user is in the Files tab, goes to chat, and comes back to the Files tab and land in the same public folder?   
I guess no here.

> Should it come back if user goes back out to the TLF list from a public folder, then back into the same public folder?
Not sure here, I'd be okay with yes. WDYT?

shared/constants/fs.js
songgao: Ah I see. In that case you could also use `writable` in the pathItem, which covers edge cases like when you access it through a social assertion handle as yourself.

That aside, it seems to me this would be a place to educate user about these stuff being open to the world, _and_ to let them know about keybase.pub, which is why I thought maybe we should show it for all. Design has final call though!

shared/fs/banner/public-reminder.js
songgao: Not sure! @cecileboucheron wdyt?

shared/constants/fs.js
cecileboucheron: Ah yeah good call, let's show it in all the public TLFs where you're a writer. So shared public folders too.

shared/fs/banner/public-reminder.js
jakob223: hmm, so upon trying to implement this I am finding that the redux solution adds a lot of complication and doesn't actually seem to yield a consistent solution that I can think of. I think I'm going to go with the simpler state version with `componentDidUpdate` for now.

jakob223:
Untagging review while I fix a mobile thing..

shared/fs/banner/public-reminder.js
songgao: I think it still depends on the behavior we want. If we want the "dismissed" state persist through screens, using the state wouldn't work as we'd lose states after unmount happens.

jakob223:
Ok, @songgao, I've rebased this on master, converted to TS, and fixed the behavior of the close button to what you described above. Can you take another look when you have a chance?

shared/fs/banner/public-reminder.tsx
```diff
@@ -0,0 +1,83 @@
...
+  if (props.show) {
+    return (
+      props.hidden && (
+        <Kb.Banner
+          color="yellow"
+          text="Everything you upload in here can be viewed by everyone at "
+          actions={[
+            {
+              onClick: () => openUrl(props.url),
+              title: props.url + '.',
```
songgao: Why the `.` in the end? That'd become part of the underlined clickable.

shared/fs/banner/public-reminder.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+type StateProps = {_lastClosedTlf: string; _writable: boolean}
+const mapStateToProps = (state, ownProps: OwnProps) => ({
+  _lastClosedTlf: state.fs.lastPublicBannerClosedTlf,
+  _writable: state.fs.pathItems.get(ownProps.path, Constants.unknownPathItem).writable,
+})
+
+type DispatchProps = {_onClose: (string) => () => void}
+const mapDispatchToProps = dispatch => ({
+  _onClose: tlf => () => dispatch(FsGen.createSetLastPublicBannerClosedTlf({tlf})),
```
songgao: Could also compute `tlf` here from `ownProps`. I suppose you did it this way so `parsePath` is only called once? Perhaps we could memoize-one `parsePath`.

shared/fs/banner/public-reminder.tsx
```diff
@@ -0,0 +1,83 @@
...
+              onClick: () => openUrl(props.url),
+              title: props.url + '.',
+            },
+          ]}
+          onClose={props.onClose}
+        />
+      )
+    )
+  } else {
+    props.onLoad()
```
songgao: Yeah, this needs to be in lifecycles or `useEffect` rather than `render()`.

An alternative is to have the connector pass in simpler actions (i.e. `onSetLastClosed` and `clearLastClosed`), and have this component whether we should set or clear or do nothing.

shared/fs/banner/public-reminder.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+const mergeProps = (s, d, o: OwnProps) => {
+  const parsedPath = Constants.parsePath(o.path)
+  switch (parsedPath.kind) {
+    case Types.PathKind.GroupTlf:
+    case Types.PathKind.InGroupTlf:
+      return {
+        hidden: s._lastClosedTlf !== parsedPath.tlfName,
+        onClose: d._onClose(parsedPath.tlfName),
+        onLoad: s._lastClosedTlf === '' ? () => {} : d._onClose(''),
```
songgao: This would cause the banner to come back when you navigate inside a public TLF too right? Seems like we'd need to compare against this with current tlfName too.

shared/fs/nav-header/mobile-header.tsx
```diff
@@ -16,6 +16,22 @@ type State = {
...
+    case 'group-tlf':
+    case 'in-group-tlf':
+      if (parsedPath.tlfType === 'public') {
+        return true
+      }
+      break
+    default:
+  }
+  return false
+}
```
songgao: There's also `Constants.getPathVisibility`

shared/fs/nav-header/title.tsx
```diff
@@ -84,10 +84,29 @@ const Breadcrumb = Kb.OverlayParentHOC(
...
+          <Kb.Box2 direction="horizontal">
+            <Kb.Meta title="public" backgroundColor={Styles.globalColors.green} />
+          </Kb.Box2>
+        )
+      }
+      break
+    default:
+  }
+  return null
+}
```
songgao: Same here. If you prefer to use `showPublicTag` above, it should probably go into `constants/fs` so it can be used here too.

shared/fs/banner/public-reminder.tsx
jakob223: Yeah, but there's no way to include a `.` after the actions with `Banner`, and there's a `.` in the design.

shared/fs/banner/public-reminder.tsx
jakob223: also, reusing `_onClose` to clear the state. How about renaming `_onClose` to `setLastPublicBannerClosedTlf` or something?

shared/fs/banner/public-reminder.tsx
jakob223: This only gets called when `show` is false. Maybe rename to `onLoadNonBannerFolder` or something?

shared/fs/nav-header/mobile-header.tsx
jakob223: 😮

shared/common-adapters/avatar-line.tsx
```diff
@@ -0,0 +1,96 @@
...
+        .map(username => (
+          <Kb.Avatar
+            size={props.size}
+            username={username}
+            key={username}
+            borderColor="white"
+            style={styles.avatar}
+          />
+        ))
+        .reverse()}
```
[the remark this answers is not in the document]
jakob223: This is so that the last avatar is underneath the first one.

PR #17618, opened 2019-05-17
Add lock in contact cache

zapu:


go/contacts/cache.go
```diff
@@ -110,6 +113,15 @@ func (c *CachedContactsProvider) LookupAll(mctx libkb.MetaContext, emails []keyb
 
 	now := mctx.G().Clock().Now()
 
+	// This is a rather long-lived lock, because normally it will be held
+	// through the entire duration of the lookup, but:
+	// - We don't expect this to be called concurrently, or repeatedly, without
+	//   user's interaction.
+	// - We want to avoid looking up the same assertion multiple times (burning
+	//   through the rate limit), while keeping the locking strategy simple.
```
heronhaye: So the idea is the first one will store the cache and the second lookup will just read from cache and not burn limit?

PR #17629, opened 2019-05-18
[KBFS-4014] offline sync settings UI

jzila:
Tagging kbfs-hackers since there are some KBFS fixes as well that need review. Ideally @songgao and/or @jakob223 would take a look.

This PR has draft code for the mobile settings page, but that's not complete. I'll open a new PR for that.


shared/common-adapters/dropdown.js
```diff
@@ -84,7 +86,9 @@ class Dropdown<N: React.Node> extends React.Component<Props<N> & OverlayParentPr
             {this.props.items.map((i, idx) => (
               <ClickableBox
                 key={idx}
-                onClick={() => {
+                onClick={(evt: SyntheticEvent<>) => {
+                  evt.stopPropagation && evt.stopPropagation()
+                  evt.preventDefault && evt.preventDefault()
```
jzila: @buoyad I'm having trouble getting this to actually prevent default. Any suggestions?

shared/settings/files/index.js
```diff
@@ -1,28 +1,110 @@
...
+  spaceAvailableNotificationThreshold: number,
+  onChangedSyncNotifications: (number) => void,
+  onEnableSyncNotifications: () => void,
+  onDisableSyncNotifications: () => void,
 |}
 
+export const allowedNotificationThresholds = [
+  100 * 1024 ** 2,
+  1024 ** 3,
+  10 * 1024 ** 3,
```
jakob223: maybe add one more option in like the 2-5 GB range? I feel like I'd want to set it somewhere around there personally.

shared/settings/files/index.js
```diff
@@ -1,28 +1,110 @@
...
 import RefreshDriverStatusOnMount from '../../fs/common/refresh-driver-status-on-mount'
+import RefreshSettings from './refresh-settings'
+import flags from '../../util/feature-flags'
 
 type Props = {|
+  areSettingsLoading: boolean,
   driverStatus: Types.DriverStatus,
-  onEnable: () => void,
   onDisable: () => void,
+  onEnable: () => void,
```
jakob223: I'm confused why you moved this down - if it's for alphabetical order you should probably swap `onEnableSyncNotifications` down too.

shared/settings/files/index.js
```diff
@@ -1,28 +1,110 @@
...
+    selectedBoxStyle={styles.syncNotificationDropdownItem}
+    disabled={props.areSettingsLoading || props.spaceAvailableNotificationThreshold === 0}
+  />
+)
+
+const ThresholdDropdown = Platform.isMobile ? ThresholdDropdownMobile : ThresholdDropdownDesktop
+
+const SyncNotificationSetting = (props: Props) => (
+  <Kb.Box2 direction="horizontal" alignItems="center">
+    <Kb.Text type="Body">Warn me if I only have less than </Kb.Text>
```
jakob223: "only have less than" feels redundant to me. Maybe cut the 'only'?

shared/settings/files/index.js
```diff
@@ -1,28 +1,110 @@
...
+          onDone={this._hide}
+          onSelect={this._select}
+        />
+      </>
+    )
+  }
+}
+
+const ThresholdDropdownDesktop = (props: Props) => (
+  <Kb.Dropdown
```
jakob223: The place the dropdown appears (not actually below the dropdown box, but centered on it) feels weird to me.

Also, the text in it is way smaller than the box itself - it feels to me like the expanded box should be the same size as the box pre-expansion. If it stays big, the text should probably be centered in it.

Finally, there should be margins around the text in the dropped-down box - right now the text is touching the borders.

no margins

shared/settings/files/index.js
```diff
@@ -71,12 +175,30 @@ const styles = Styles.styleSheetCreate({
...
   },
+  syncContent: {
+    paddingLeft: Styles.globalMargins.xsmall,
+    paddingTop: Styles.globalMargins.medium,
+  },
+  syncNotificationCheckbox: {
+    alignItems: 'center',
+  },
+  syncNotificationDropdownItem: {
+    width: 2 * Styles.globalMargins.xlarge,
```
jakob223: why is this so big?

shared/settings/files/index.js
jzila: Copy is from design. I'll change it though to "Warn me if I have less than <n> storage space remaining".

shared/settings/files/index.js
jzila: It's actually pretty small, that's how we get the root dropdown to look so small.

shared/settings/files/index.js
jzila: I was having trouble figuring out how to style this.

jzila:
@jakob223 fixes implemented.

shared/settings/files/index.tsx
```diff
@@ -202,7 +203,11 @@ const styles = Styles.styleSheetCreate({
     alignItems: 'center',
   },
   syncNotificationDropdownItem: {
-    width: 2 * Styles.globalMargins.xlarge,
+    alignItems: 'center',
+    paddingLeft: Styles.globalMargins.small,
+  },
+  syncNotificationDropdownOverlay: {
+    width: Styles.globalMargins.xlarge + Styles.globalMargins.medium,
```
jakob223: Why is this expressed as a sum of margins instead of as a number?

jzila: I always try to express dimensions as linear combinations of margins so it looks "good" relative to all the other dimensions we use, but usually I change it to an actual amount after the fact. I can do that here.

PR #17632, opened 2019-05-19
Fix nav2 header displaying on desktop

cjb:
@keybase/react-hackers 

Looks like React components stopped being functions and started being objects, perhaps in the newer React that came along with newer RN.  Haven't had a chance to investigate further but this PR looks fine on master to me.


chrisnojima:
https://github.com/keybase/client/pull/17635

PR #17640, opened 2019-05-20
types/fs: support '//' in paths more

jakob223:
This is a response to an issue raised by @heronhaye about broken breadcrumbs.

Perhaps we should also trim `/` from the end of paths when coming from e.g. a chat link.

### DIFF
--- a/shared/constants/types/fs.js
+++ b/shared/constants/types/fs.js
@@ -529,7 +529,8 @@ export const getPathFromRelative = (tlfName: string, tlfType: TlfType, inTlfPath
   '/keybase/' + tlfType + '/' + tlfName + '/' + inTlfPath
 export const stringToEditID = (s: string): EditID => s
 export const editIDToString = (s: EditID): string => s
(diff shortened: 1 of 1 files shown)


jakob223:
@songgao suggested in chat that I move this logic into `stringToPath` for now, with a possible change when KBFS-4177 is done.

shared/constants/types/fs.js
```diff
@@ -547,7 +547,7 @@ export const getPathParent = (p: Path): Path =>
         .slice(0, -1)
         .join('/')
 export const getPathElements = memoize<Path, void, void, void, _>(
-  (p: Path): Array<string> => (!p ? [] : p.split('/').slice(1))
+  (p: Path): Array<string> => (!p ? [] : p.split('/').filter(w => w))
```
songgao: 👾.filter

songgao: oops sorry just realized you had removed review

PR #17653, opened 2019-05-20
Skip TestCancelResetPipeline systest for now

pzduniak:


pzduniak:
ok maybe we dont need reviews for skips :^)

PR #17667, opened 2019-05-21
TS - People

buoyad:
r? @keybase/react-hackers


shared/tsconfig.json
```diff
@@ -1,21 +1,16 @@
...
-        "node_modules",
-        "react-native",
-        "desktop/dist"
-    ]
+  "compilerOptions": {
+    "outDir": "./dist/",
+    "sourceMap": true,
+    "noImplicitAny": false,
+    "module": "commonjs",
+    "moduleResolution": "node",
```
buoyad: Only added this, the rest is formatting. This is to support `import moment from 'moment'` and should match our reality better in general

MarcoPolo: o cool! I had a different solution that used the `"allowSyntheticDefaultImports": true,` flag. I guess this is better?

MarcoPolo: O hmm, I think we actually need this as well

buoyad: Yeah, I saw that one first too

shared/people/index.shared.tsx
```diff
@@ -7,10 +6,13 @@ import Todo from './todo/container'
 import FollowNotification from './follow-notification'
 import Announcement from './announcement/container'
 import FollowSuggestions from './follow-suggestions'
-import {type Props} from '.'
+import {Props} from '.'
 import AirdropBanner from '../wallets/airdrop/banner/container'
 
-export const itemToComponent: (Types.PeopleScreenItem, Props) => React.Node = (item, props) => {
+export const itemToComponent: (arg0: Types.PeopleScreenItem, props: Props) => React.ReactNode = (
```
adamjspooner: Why `arg0` and not `item`?

buoyad: That was from the converter, I'll change

buoyad:
@keybase/react-hackers rebased. PTAL

PR #17675, opened 2019-05-22
TS util/

MarcoPolo:
@keybase/react-hackers Transform `util/`


shared/tsconfig.json
```diff
@@ -8,7 +8,14 @@
         "jsx": "preserve",
         "incremental": true,
         "outDir": "./.tsOut/emit",
-        "tsBuildInfoFile": "./.tsOut/cache"
+        "allowSyntheticDefaultImports": true,
```
MarcoPolo: I might back this change out in favor of @buoyad 's version.

shared/tsconfig.json
```diff
@@ -8,7 +8,14 @@
...
+        "tsBuildInfoFile": "./.tsOut/cache",
+        "baseUrl": ".",
+        "paths": {
```
MarcoPolo: The rn-fetch-blob types seem broken. This is a workaround to override that type. It looks like this commit is wrong (https://github.com/joltup/rn-fetch-blob/commit/8bc5cebddaddc2e6d6d27dcb248d2e9bb24aecaf)

shared/globals.d.ts
```diff
@@ -0,0 +1,20 @@
+declare var __DEV__: boolean
```
MarcoPolo: Globals declaration file for our globals

shared/yarn.lock
```diff
@@ -10716,6 +10723,11 @@ mksnapshot@^0.3.4:
     fs-extra "0.26.7"
     request "2.x"
 
+moment@*:
```
cjb: Maybe this is accidental?

MarcoPolo: I don't know how to properly add dependencies it seems.

shared/util/hidden-string.tsx
```diff
@@ -8,11 +7,11 @@ class HiddenString {
     this._value = () => stringValue
   }
 
-  toString(): string {
+  toString = (): string => {
```
chrisnojima: doesn't need to be an arrow right

shared/util/kbfs-notifications.tsx
```diff
@@ -7,13 +6,13 @@ import {
...
-import type {TypedState} from '../constants/reducer'
+import { FSNotification } from '../constants/types/rpc-gen';
+import { TypedState } from '../constants/reducer';
 
 type DecodedKBFSError = {
   title: string,
-  body: string,
-}
+  body: string
+};
```
chrisnojima: maybe you didn't run prettier

shared/yarn.lock
MarcoPolo: I removed it and ran `yarn` and it got added again. It looks like this is because of the `@types/moment` dependency. We have other `@*` types in this file too.

shared/tsconfig.json
MarcoPolo: Just kidding, it seems like we need this too.

PR #17678, opened 2019-05-22
TS git

buoyad:
r? @keybase/react-hackers

### DIFF
--- a/shared/git/row/index.tsx
+++ b/shared/git/row/index.tsx
@@ -1,37 +1,36 @@
-// @flow
 import * as React from 'react'
 import * as Kb from '../../common-adapters'
 import * as Styles from '../../styles'
 
 type Props = {
-  canDelete: boolean,
-  canEdit: boolean,
-  channelName: ?string,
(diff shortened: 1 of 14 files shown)


shared/git/row/index.tsx
```diff
@@ -1,37 +1,36 @@
...
-  onClickDevice: () => void,
-  onShowDelete: () => void,
-  onChannelClick: (SyntheticEvent<>) => void,
-  onToggleChatEnabled: () => void,
-  onToggleExpand: () => void,
-  openUserTracker: (username: string) => void,
-  _onOpenChannelSelection: () => void,
+  canDelete: boolean
+  canEdit: boolean
+  channelName: string | null
```
jzila: General comment: why not switch to `channelName?: string`?

buoyad: It wouldn't be a strictly correct conversion, so if the converter did that we'd likely have more tsc errors to deal with. Maybe in TS that'll end up being easier to use, not sure

PR #17679, opened 2019-05-22
TS login

chrisnojima:


chrisnojima:
just a heads up i made jest ignore tsOut so it doesn't try and test the output tsc
also touched the jenkins file re: how it installs yarn. i think its safe to run yarn modules

PR #17681, opened 2019-05-23
Handle unused errors

joshblum:
https://staticcheck.io/docs/ seems like a useful tool and may be worth integrating into CI

ran `$ staticcheck -checks "SA4006" ./...  | grep "this value of err"` to handle unused errors.

### DIFF
--- a/go/engine/passphrase_recover.go
+++ b/go/engine/passphrase_recover.go
@@ -251,6 +251,9 @@ func (e *PassphraseRecover) changePassword(mctx libkb.MetaContext) (err error) {
 	// We either have no server keys or the user is OK with resetting them
 	// Prompt the user for a new passphrase.
 	passphrase, err := e.promptPassphrase(mctx)
+	if err != nil {
+		return err
+	}
 
 	// ppres.Passphrase contains our new password
 	// Run passphrase change to finish the flow
--- a/go/kbfs/libkbfs/tlf_journal_test.go
+++ b/go/kbfs/libkbfs/tlf_journal_test.go
@@ -532,7 +532,8 @@ func testTLFJournalBlockOpDiskByteLimit(t *testing.T, ver kbfsmd.MetadataVer) {
 	md := config.makeMD(kbfsmd.RevisionInitial, kbfsmd.ID{})
 	err = tlfJournal.doOnMDFlushAndRemoveFlushedMDEntry(
 		ctx, kbfsmd.ID{}, &RootMetadataSigned{RootMetadataSigned: kbfsmd.RootMetadataSigned{MD: md.bareMd}})
-	require.NoError(t, err)
+	require.Error(t, err)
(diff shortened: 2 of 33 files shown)


joshblum:
@jzila 

```
rn-story-mode........................................(no files to check)Skipped
eslint...............................................(no files to check)Skipped
tsc..................................................(no files to check)Skipped
prettier.............................................(no files to check)Skipped
go fmt...................................................................Passed
go vet...................................................................Passed
go lint..................................................................Failed
hookid: go-lint

go/kbfs/simplefs/simplefs_test.go:137:1: context.Context should be the first parameter of a function
go/kbfs/simplefs/simplefs_test.go:174:1: context.Context should be the first parameter of a function

```

strib:
@joshblum not sure what tool this is, but we explicitly mask out that golint ctx check for tests in https://github.com/keybase/client/blob/a19fd3247f50b28058dbcd6ecc56616a26cf09d7/go/Makefile#L36

I don't mind changing the code this way, but presumably there will be a ton of other lint failures for the other errors we filter out.

joshblum:
@strib this just came from the precommit hook not the tool, could revert the changes and skip the precommit checks if it will pass CI

go/kbfs/libkbfs/tlf_journal_test.go
```diff
@@ -532,6 +532,7 @@ func testTLFJournalBlockOpDiskByteLimit(t *testing.T, ver kbfsmd.MetadataVer) {
 	md := config.makeMD(kbfsmd.RevisionInitial, kbfsmd.ID{})
 	err = tlfJournal.doOnMDFlushAndRemoveFlushedMDEntry(
 		ctx, kbfsmd.ID{}, &RootMetadataSigned{RootMetadataSigned: kbfsmd.RootMetadataSigned{MD: md.bareMd}})
+	require.NoError(t, err)
```
joshblum: @strib is an error expected here? https://ci.keyba.se/job/client/job/PR-17681/4/execution/node/2354/log/

strib: I took a quick look and yeah it seems like an error is expected here, it's doing something kinda funny for test purposes.

joshblum: thx!

joshblum:
@jzila ptal

go/engine/passphrase_recover.go
```diff
@@ -251,6 +251,9 @@ func (e *PassphraseRecover) changePassword(mctx libkb.MetaContext) (err error) {
 	// We either have no server keys or the user is OK with resetting them
 	// Prompt the user for a new passphrase.
 	passphrase, err := e.promptPassphrase(mctx)
+	if err != nil {
+		return err
+	}
```
jzila: Was this a bug?

joshblum: yes i don't think i the error was intentionally ignored

PR #17686, opened 2019-05-23
fix infinite recursion, redundant statements

joshblum:


joshblum:
merging over `TestBech32FeatureFlag`, was fixed in https://github.com/keybase/keybase/pull/3876

PR #17694, opened 2019-05-24
Support for public text memo in wallet api

patrickxb:


patrickxb:
CI failure unrelated, merging over it.

PR #17703, opened 2019-05-24
TS Actions

MarcoPolo:
@keybase/react-hackers


shared/actions/teams.tsx
```diff
@@ -1334,15 +1362,18 @@ const badgeAppForTeams = (state, action) => {
...
+    // TODO ts-migration remove any
+    const toLoad: I.Set<any> = newTeamRequestsSet.subtract(existingNewTeamRequestsSet)
     const loadingCalls = toLoad.map(teamname => TeamsGen.createGetDetails({teamname})).toArray()
     actions = actions.concat(loadingCalls)
   }
 
   // if the user wasn't on the teams tab, loads will be triggered by navigation around the app
   actions.push(
     TeamsGen.createSetNewTeamInfo({
+      // @ts-ignore a legit error. todo fix
```
MarcoPolo: @cjb I think this is a legit error from your PR. You forgot to add this in the types

shared/actions/chat2/index.tsx
```diff
@@ -662,12 +701,40 @@ const onChatSetConvRetention = (_, action, logger) => {
   })
 }
 
-const onChatSetConvSettings = (_, action, logger) => {
+const onChatSetConvSettings = (_, action: EngineGen.Chat1NotifyChatChatSetConvSettingsPayload, logger) => {
   const {conv, convID} = action.payload.params
   const conversationIDKey = Types.conversationIDToKey(convID)
-  const newRole = conv?.convSettings?.minWriterRoleInfo?.role
+  const newRole = // Auto generated from flowToTs. Please clean me!
```
chrisnojima: i think we need to fix this

shared/actions/chat2/index.tsx
```diff
@@ -2631,15 +2783,15 @@ const unsentTextChanged = (state, action) => {
   })
 }
 
-const onGiphyResults = (state, action) => {
+const onGiphyResults = (state, action: EngineGen.Chat1ChatUiChatGiphySearchResultsPayload) => {
```
chrisnojima: do we really have to type these actions also?

shared/package.json
```diff
@@ -23,7 +23,7 @@
     "build-dev": "yarn run _helper build-dev",
     "build-emoji": "yarn run _node common-adapters/markdown/generate-emoji-parser.js",
     "build-prod": "yarn run _helper build-prod",
-    "build-actions": "yarn run _node desktop/yarn-helper/action-creator-creator",
+    "build-actions": "yarn run _node --extensions \".tsx\" desktop/yarn-helper/action-creator-creator.tsx",
```
chrisnojima: i don't think its necessarily required converting all these files over unless its worth it

shared/util/flow.tsx
```diff
@@ -1 +1,2 @@
-export const ifFlowComplainsAboutThisFunctionYouHaventHandledAllCasesInASwitch = (e: never) => {}
+// @ts-ignore codemod-issue
+export const ifFlowComplainsAboutThisFunctionYouHaventHandledAllCasesInASwitch = (e: never): never => {}
```
chrisnojima: i think we should toss this file

shared/actions/chat2/index.tsx
MarcoPolo: I changed it to   `const cannotWrite = conv && conv.convSettings && conv.convSettings.minWriterRoleInfo && conv.convSettings.minWriterRoleInfo.cannotWrite || null` which is technically different than the nullish coalescing, but fine for this I think.

shared/util/flow.tsx
MarcoPolo: We still need something like this for a similar reason. I was going to rename it in another pr.

shared/package.json
MarcoPolo: It was easy so why not /shrug. but good to know for future cases

shared/actions/chat2/index.tsx
MarcoPolo: Yes. One of the differences in TS is that it won't infer function parameter types based on usage. (except in anonymous inline closures).

example: https://www.typescriptlang.org/play/#src=const%20numbers%20%3D%20%5B1%2C%202%2C%203%2C%204%2C%205%5D%0A%0Anumbers.map(a%20%3D%3E%20a%20%2B%201)%0A%0Aconst%20add1Untyped%20%3D%20(a)%20%3D%3E%20a%20%2B%201%0Aconst%20add1Typed%20%3D%20(a%3A%20number)%20%3D%3E%20a%20%2B%201%0A%0A%0Anumbers.map(add1Untyped)%0Anumbers.map(add1Typed)

shared/actions/fs/index.tsx
```diff
@@ -154,7 +161,7 @@ const loadTlfSyncConfig = (state, action) => {
     .catch(makeUnretriableErrorHandler(action, tlfPath))
 }
 
-const setTlfSyncConfig = (state, action) =>
+const setTlfSyncConfig = (state, action: FsGen.SetTlfSyncConfigPayload) =>
```
MarcoPolo: Basically just added the type of actions here. @jakob223 @songgao

shared/actions/teams.tsx
```diff
@@ -1039,6 +1057,7 @@ function* setPublicity(state, action) {
           {
             name: teamname,
             settings: {
+              // @ts-ignore codemod fix when constants is typed
```
cjb: fwiw constants is typed now

shared/actions/teams.tsx
```diff
@@ -968,7 +981,12 @@ function* createChannel(_, action, logger) {
     }
 
     // Dismiss the create channel dialog.
-    if (Router2Constants.getVisibleScreen()?.routeName === 'chatCreateChannel') {
+    if (
+      // Auto generated from flowToTs. Please clean me!
+      (Router2Constants.getVisibleScreen() === null || Router2Constants.getVisibleScreen() === undefined
+        ? undefined
+        : Router2Constants.getVisibleScreen().routeName) === 'chatCreateChannel'
```
cjb: I don't think we should merge in unconverted code like this, could be really unsafe

PR #17705, opened 2019-05-24
Bust passphrase stream cache if it's outdated in places prone to failure

pzduniak:
Depends on https://github.com/keybase/keybase/pull/3885

Fixes a ppgen assertion bug where an outdated cached passphrase stream would result in failing kex2 provisions / paperkey generations.


maxtaco:
this looks good, but how does this login_test pass if the Web endpoint hasn't been merged?  I would expect the tests to fail on that endpoint 404'ing

pzduniak:
my phone UI says the last CI never finished

maxtaco:
build 4 and 3 worked, but the engine tests seemed to work

i reran for you since it seems like build 4 flaked

pzduniak:
lmk if i should merge, everything passes locally

maxtaco:
:+1: to merge

PR #17706, opened 2019-05-24
piping urls from the server to the front end

xgess:
* a bunch of boilerplate for fetching some data (external_urls) and verifying it against a hash in the merkle root
* a local RPC for getting stellar exchange urls out of ^ and to the front end

depends on https://github.com/keybase/keybase/pull/3868

### DIFF
--- a/go/libkb/constants.go
+++ b/go/libkb/constants.go
@@ -721,3 +721,8 @@ const (
 )
 
 const ProfileProofSuggestions = true
+
+const (
+	ExternalURLsBaseKey          = "external_urls"
+	ExternalURLsStellarExchanges = "stellar_exchanges"
+)
--- a/go/libkb/constants.go
+++ b/go/libkb/constants.go
@@ -723,6 +723,6 @@ const (
 const ProfileProofSuggestions = true
 
 const (
-	ExternalURLsBaseKey          = "external_urls"
-	ExternalURLsStellarExchanges = "stellar_exchanges"
+	ExternalURLsBaseKey         = "external_urls"
+	ExternalURLsStellarPartners = "stellar_partners"
 )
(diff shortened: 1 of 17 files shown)


protocol/avdl/stellar1/local.avdl
```diff
@@ -728,9 +728,22 @@ protocol local {
...
+    string title;
+    string description;
+    @jsonkey("icon_filename")
+    string iconFilename;
+    @jsonkey("admin_only")
+    boolean adminOnly;
+    string extra; // extra json tbd
+  }
+
+  array<ExchangeUrl> getExchangeUrlsLocal(int sessionID);
```
patrickxb: I think these should be called ExternalURLs...or PartnerURLs.

So they aren't confused with the links we are adding to assets to the stellarx exchange.

xgess:
> Anything frontend calls needs a sessionID.

lol. i figured something like that would be the case :/
ok thanks.

protocol/avdl/stellar1/local.avdl
xgess: done

go/libkb/constants.go
```diff
@@ -723,6 +723,6 @@ const (
 const ProfileProofSuggestions = true
 
 const (
-	ExternalURLsBaseKey          = "external_urls"
-	ExternalURLsStellarExchanges = "stellar_exchanges"
+	ExternalURLsBaseKey         = "external_urls"
+	ExternalURLsStellarPartners = "stellar_exchanges"
```
patrickxb: sorry to be a pain, but can you change the value of the const to `stellar_partners` too?

just don't want to get confused later...

thanks!

xgess: done. https://github.com/keybase/keybase/pull/3891

PR #17709, opened 2019-05-25
fix desktop section list sticky calc

mmaxim:
`throttle` isn't really what you want here, since you can find yourself on a time boundary and end up "missing" a scroll event since the list is still changing. That is, you don't always wait the full 20ms and this calculation is frequently wrong, meaning section headers render when they aren't supposed to. You can see this on the inbox search results, and on the new info panel.

Switching this to `setTimeout` fixes the bug for me, and makes more sense. You basically want to process every scroll event here, but you just want to wait some fixed time for the list to give you the right result in `getVisibleRange`.


chrisnojima:
Mmm. Let's discuss. I think I'd prefer throttle with trailing if that works

mmaxim:
Actually just changing to `debounce` also works, and seems like you might like it better.

mmaxim:
Yeah `debounce` might suffer from the same problems as the `throttle` solutions (I'm not sure how the trailing change affects anything). It seems to work for me, but I think that is just because the failure case is harder to hit. The `setTimeout` change definitely works, so we could go with that.

Why don't we just talk about it in person on Tuesday though, no rush on it.

chrisnojima:
I'd you can repro just try throttle but pass {trailing:true} as the 3rd param

shared/common-adapters/section-list.desktop.js
```diff
@@ -105,11 +105,14 @@ class SectionList extends React.Component<Props, State> {
...
     }
-  }, 20)
+  }
+  _checkStickyDebounced = debounce(this._checkSticky, 20)
+  _checkStickyThrottled = throttle(this._checkSticky, 20)
 
   _onScroll = e => {
     e.currentTarget && this._checkOnEndReached(e.currentTarget)
-    this._checkSticky()
+    this._checkStickyDebounced()
```
chrisnojima: might want to add a comment as this is kinda tricky

PR #17714, opened 2019-05-28
proposal for PTK hidden chain

maxtaco:
- a manager object will manage all puts and fetches into chain17 chains
- FTL and slow loader will continue to load prevs and RKMs as before
- they will call intio the HiddenChainManager to get the PTKs that aren't on chain3 as they load
- do a small refactor of the loader to indicate where we're going to alternate between chain3 and chain17 loads
- next steps are to change the web endpoints to send down needed chain17 links
- @mlsteele I'd like to discuss this in person whenever you get a chance


go/teams/teams.go
```diff
@@ -519,17 +524,34 @@ func (t *Team) Rotate(ctx context.Context) (err error) {
...
+	latestSeqno, err := t.postChangeItem(ctx, section, libkb.LinkTypeRotateKey, mr, payloadArgs)
+	if err != nil {
+		return err
+	}
+	t.notify(ctx, keybase1.TeamChangeSet{KeyRotated: true}, latestSeqno)
+	return nil
+}
+
+func (t *Team) rotatePostHidden(ctx context.Context, section SCTeamSection, mr *libkb.MerkleRoot, payloadArgs sigPayloadArgs) error {
+	return nil
```
mlsteele: return error

protocol/avdl/keybase1/teams.avdl
```diff
@@ -248,6 +283,23 @@ protocol teams {
...
 
+  record Signer {
+    Seqno e; // eldest seqno
+    KID k;   // signer KID
+    UID u;   // signer UID
+  }
+
+  record HiddenTeamChainLink {
+    @jsonkey("m") @mpackkey("m")
+    MerkleRootV2 merkleRoot;
```
mlsteele: What merkle root? Merkle root that this link is first included in? Merkle root it signed over?

maxtaco:
@heronhaye sorry I merged that prior PR by accident. The merges are now in this PR. Ignore the stuff in `teams`. The PR in question is mainly in `sig3`

go/sig3/prot.go
```diff
@@ -0,0 +1,133 @@
...
+	_struct             bool                `codec:",toarray"`
+	Version             SigVersion          `codec:"version"` // should be 3
+	Seqno               Seqno               `codec:"seqno"`
+	Prev                *LinkID             `codec:"prev"`
+	InnerLink           LinkID              `codec:"curr"` // hash of the msgpack of the InnerLink
+	LinkType            LinkType            `codec:"type"` // hash of the msgpack of the previous OuterLink
+	ChainType           ChainType           `codec:"chaintype"`
+	IgnoreIfUnsupported IgnoreIfUnsupported `codec:"ignore_if_unsupported"`
+	// New field for V3; if this link is encrypted, specify the format, nonce and PUK
+	EncryptionParameters *EncryptionParameters `codec:"encryption_parameters"`
```
heronhaye: I don't see this field in proofs/src/sig3.iced?

maxtaco: good point, it shoudl be there

maxtaco: https://github.com/keybase/proofs/pull/143/files

maxtaco:
CI red is related to a bug that @chrisnojima  is fixing, but I'm not sure about the KBFS red

go/sig3/sig3.go
```diff
@@ -0,0 +1,401 @@
...
+	// Inner methods for generic decoding of incoming sig3 links
+	outerPointer() *OuterLink
+	setVerifiedBit()
+	verify() error
+}
+
+// Base struct for sig3 links that contains much of the raw material pulled down or off local storage.
+// Most implementations of sig3 links should include this base class.
+type Base struct {
+	verified bool
```
heronhaye: Does this mean just that the signature is verified against the given kid (which is in Base.verify) or also link-specific things like reverse sigs were verified? e.g., would it be set in RotateKey.verify() if the base sig checks out (because the bit is set at the end of Base.verify()) but the reverse sigs don't?

go/sig3/prot.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+// OuterLink V3 is the third version of Keybase sigchain signatures, it roughly approximates
+// the outer link v2s that we have previously used.
+type OuterLink struct {
+	_struct             bool                `codec:",toarray"`
+	Version             SigVersion          `codec:"version"` // should be 3
+	Seqno               Seqno               `codec:"seqno"`
+	Prev                *LinkID             `codec:"prev"`
+	InnerLink           LinkID              `codec:"curr"` // hash of the msgpack of the InnerLink
```
heronhaye: could we rename to `InnerLinkID` to avoid confusing it with an actual link type?

go/sig3/sig3.go
```diff
@@ -0,0 +1,401 @@
...
+	// Make sure to replace them on the way out of the function, even in an error.
+	defer func() {
+		for j, rs := range reverseSigs {
+			r.rkb.PTKs[j].ReverseSig = rs
+		}
+		r.Base.outer.InnerLink = innerLink
+	}()
+
+	// Verify signatures in the reverse order they were signed, nulling them out
+	// from back to front. We are not going middle-out.
```
heronhaye: :musical_score:

go/sig3/sig3.go
```diff
@@ -0,0 +1,401 @@
...
+	out.setVerifiedBit()
+	return out, nil
+}
+
+func (sig Sig) verify(kid KID, body []byte) error {
+	key := kbcrypto.KIDToNaclSigningKeyPublic([]byte(kid))
+	if key == nil {
+		return newSig3Error("failed to import public key")
+	}
+	msg := kbcrypto.SignaturePrefixSigchain3.Prefix(body)
```
heronhaye: this is for the polyglot issue?

go/sig3/sig3.go
heronhaye: This looks fine but is there a reason we don't null out a copy of the link instead so as not to mutate the receiver? Performance?

go/sig3/prot.go
```diff
@@ -0,0 +1,133 @@
...
+type LinkID [32]byte
+type Seqno = keybase1.Seqno
+type Time = keybase1.Time
+type IgnoreIfUnsupported bool
+type KID = keybase1.BinaryKID
+type TeamID [16]byte
+type PerTeamKeyGeneration = keybase1.PerTeamKeyGeneration
+type Entropy []byte
+type Sig [64]byte
+type PTKType int
```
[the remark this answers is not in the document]
maxtaco: yes.  longer-term. Or maybe admin keys.

go/sig3/sig3.go
```diff
@@ -0,0 +1,401 @@
...
+	outerRaw []byte
+	outer    OuterLink
+	sig      *Sig
+}
+
+// RotateKey is a sig3 link type for a PTK rotation. Handles multiple PTK types being rotated in
+// one link.
+type RotateKey struct {
+	*Base
+	rkb *RotateKeyBody
```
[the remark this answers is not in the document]
maxtaco: just to save the casting

go/sig3/sig3.go
maxtaco: It's all or nothing. It should only be unset in the case of a stubbed link (since we can't verify) or if there was an error in any crypto operations, including outer and inner sigs.

go/sig3/sig3.go
maxtaco: I'd happily use a deep copy operation, but we'd have to write it. I guess we can?

go/sig3/sig3.go
maxtaco: yes

go/sig3/prot.go
maxtaco: ok

maxtaco: done

go/sig3/sig3.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+// Seqno returns the seqno of this link, as reported in the outer link itself.
+func (b *Base) Seqno() Seqno {
+	return b.outer.Seqno
+}
+
+func (b Base) verify() error {
+	if (b.sig == nil) != (b.inner == nil) {
+		return newSig3Error("need sig and inner, or neither, but can't have one and not the other")
```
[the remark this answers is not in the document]
maxtaco: fixed here and one more place.

maxtaco:
Many thanks @heronhaye !!!!

go/sig3/sig3.go
maxtaco: I also wonder if the deep copy should keep the aliasing from `RotateKeyBody` into `Base.inner.Body`... Maybe let's punt for now and revisit in the future.

PR #17715, opened 2019-05-28
use isolatedModules

chrisnojima:
@keybase/react-hackers 

i noticed webpack complaining about some exports

babel assumes this flag is on so we should turn it on in our tsconfig
https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats

You have to re-export in a different way due to this sometimes
https://github.com/microsoft/TypeScript/issues/28481


shared/tsconfig.json
```diff
@@ -1,19 +1,18 @@
...
-    "incremental": true,
-    "outDir": "./.tsOut/emit",
     "allowSyntheticDefaultImports": true,
-    "tsBuildInfoFile": "./.tsOut/cache",
     "baseUrl": ".",
-    "paths": {
-      "rn-fetch-blob": ["custom-d.ts/rn-fetch-blob.d.ts"]
-    }
+    "incremental": true,
+    "isolatedModules": true,
```
chrisnojima: just added isolatedModules and sorted this list

PR #17717, opened 2019-05-28
Display "advanced" txs in the CLI

patrickxb:
We currently skip all non-payment transactions in history, details.

This allows us to show a text summary of "advanced" transactions like creating trustlines, offers, etc. Or multiple operation transactions.

It is only enabled when the CLI calls it since we don't have frontend support yet for rendering these.

There's a stellard PR incoming for the server-side.


protocol/avdl/stellar1/remote.avdl
```diff
@@ -108,6 +108,9 @@ protocol remote {
     string sourceAmountMax;
     string sourceAmountActual;
     Asset sourceAsset;
+    boolean isAdvanced;     // true for non-payment transactions
+    string summaryAdvanced;   // one-line summary of the advanced tx
+    array<string> operations;   // string representations of each operation in the tx
```
mlsteele: This looks like a makeshift variant inside a real variant. Are there issues with adding a new PaymentSummary variant?

patrickxb: Yes, I didn't want a 4th cursor to manage and deal with merging issues.  

Keeping this here solves the problem of looking for X payments in the transactions results as well.

mlsteele: A PaymentSummary variant would still use `horizonCursor` as they only come from that source. What's the X payments issue? I think another variant would be easier to work with. Especially since this way most of `PaymentSummaryStellar`'s fields like `to` and `amount` hold zero values.

patrickxb: I disagree.  I don't think this is that challenging to understand in the data structure.  There is already variance for path payments.

The cursor I'm talking about is the cursor that we maintain that has 3 parts currently, one for network, one for direct, one for relay.  Adding a 4th variant would require a 4th part to that cursor, a 4th thing to merge when getting recent payments.  The current system is already complicated enough and doesn't need any other complications.

The X payments issue is this:

            // Scan through pages of horizon transactions to gather `limit` payments.                                 
        // But only check so many pages so we don't get stuck in this loop.
        const maxPages = 10
        for i := 0; i < maxPages; i++ {
                // Scan large pages of transactions no matter how many payments we want
                // because payments can be sparse in transactions pages.

currently for people with lots of non-payments, this fails.  With this change, it works.

mlsteele: Ok. I think swaths of zero-valued fields are apt to be used in a context where it's not obvious they're blank. But if I can't convince you that's fine, this PR makes sense. I'll leave these thoughts here:

I don't see why a new cursor would be necessary. The cursor points into a data source which is a stream of `PaymentSummaryStellar` in this PR. And would be a stream of `PaymentSummaryStellar | PaymentSummaryOther` with another variant. The same cursor part would page through it.

Re X payments: `recentPaymentsNetwork` would return a list of `PaymentSummaryStellar | PaymentSummaryOther`. (It's Go so []interface{} that gets reconciled higher up)

PR #17719, opened 2019-05-28
Remove dead code: ListTrackers

mlsteele:


shared/constants/types/rpc-gen.js.flow
```diff
@@ -0,0 +1,4103 @@
+// @flow strict
```
buoyad: Run `make clean` before `make` to avoid generating these files. Or use `yarn update-protocol`

mlsteele: thanks

mlsteele: Is running protocol's `make` and not `yarn update-protocol` no longer cool?

buoyad: `make` is a little less cool; update-protocol updates FE generated stuff so CI doesn't have to tell you to

mlsteele: A suggestion for `update-protocol` would be to use `make` instead of `make clean && make`, since the point of `make` is to be faster

PR #17729, opened 2019-05-28
Login cleanup scope reduction, restore original session if failed with switch

pzduniak:
Proof of concept, most likely will need a test for some of these flows that trigger it. Probably should fix all of these switch issues.


go/engine/login_provision.go
```diff
@@ -1192,7 +1192,7 @@ func (e *loginProvision) cleanup(m libkb.MetaContext) {
 
 	// the best way to cleanup is to logout...
 	m.Debug("an error occurred during provisioning, logging out")
-	m.G().Logout(m.Ctx())
+	m.G().LogoutWithSecretKill(m, e.arg.User.GetNormalizedName(), true)
```
maxtaco: just delete this

go/engine/login_provision.go
```diff
@@ -1192,7 +1192,7 @@ func (e *loginProvision) cleanup(m libkb.MetaContext) {
 
 	// the best way to cleanup is to logout...
 	m.Debug("an error occurred during provisioning, logging out")
-	m.G().Logout(m.Ctx())
```
maxtaco: there are 3 more of these to delete

go/libkb/globals.go
```diff
@@ -307,11 +309,10 @@ func (g *GlobalContext) logoutSecretStore(mctx MetaContext, username NormalizedU
 	delete(g.switchedUsers, username)
 }
 
-func (g *GlobalContext) LogoutWithSecretKill(mctx MetaContext, killSecrets bool) (err error) {
+func (g *GlobalContext) LogoutWithSecretKill(mctx MetaContext, username NormalizedUsername, killSecrets bool) (err error) {
```
maxtaco: Rename to `LogoutUsernameWithSecretKill`

go/engine/login.go
```diff
@@ -125,7 +134,7 @@ func (e *Login) Run(m libkb.MetaContext) (err error) {
 	// clear out any existing session:
 	m.Debug("clearing any existing login session with Logout before loading user for login")
 	// If the doUserSwitch flag is specified, we don't want to kill the existing session
-	m.G().LogoutWithSecretKill(m, !e.doUserSwitch)
+	m.G().LogoutWithSecretKill(m, libkb.NewNormalizedUsername(e.username), !e.doUserSwitch)
```
maxtaco: this is the wrong user to logout, you want the current username.

go/libkb/globals.go
maxtaco: Maybe also have a fuction `LogoutCurrentUserWithSecretKill`

maxtaco: That way you don't have to change `Logout` or `ClearStateForSwitchUsers`

go/engine/login_provision.go
pzduniak: Only 1 more with _fatal consequences_, the remaining calls are actually somewhat required (and hard to rework) but they are fixed with the restoreSession fn.

PR #17736, opened 2019-05-29
fix tsc

chrisnojima:
@keybase/react-hackers merging now


shared/__mocks__/file-mock.tsx
```diff
@@ -1,4 +1,4 @@
 if (!__STORYBOOK__) {
   throw new Error('Invalid load of mock')
 }
-module.exports = {}
```
chrisnojima: wants a normal module export now

PR #17739, opened 2019-05-29
clean up removal of null propagation in useNativeFrame

heronhaye:


shared/actions/platform-specific/index.desktop.js
```diff
@@ -345,9 +345,11 @@ const setUseNativeFrame = state =>
   SafeElectron.getIpcRenderer().send('setAppState', {useNativeFrame: state.settings.useNativeFrame})
 
 function* initializeUseNativeFrame() {
+  const useNativeFrame = new AppState().state.useNativeFrame
   yield Saga.put(
     SettingsGen.createOnChangeUseNativeFrame({
-      enabled: new AppState().state.useNativeFrame ?? defaultUseNativeFrame,
+      enabled:
```
chrisnojima: these can be `useNativeFrame || defaultUseNativeFrame` (and below)

heronhaye: wouldn't this override if user set it to false?

chrisnojima: ok, we can leave it as is

PR #17743, opened 2019-05-29
ts store. add global dfs and rename some things

chrisnojima:
@keybase/react-hackers 
i also moved some things around. tried hot loading on desktop/native and it seems fine


shared/app/index.native.tsx
```diff
@@ -7,40 +7,30 @@ import {GatewayProvider} from 'react-gateway'
 import {Provider} from 'react-redux'
 import {makeEngine} from '../engine'
 
-// @ts-ignore
```
chrisnojima: added a @types/webpack for this

shared/app/index.native.tsx
```diff
@@ -7,40 +7,30 @@ import {GatewayProvider} from 'react-gateway'
...
 module.hot &&
-  // @ts-ignore
   module.hot.accept(() => {
     console.log('accepted update in shared/index.native')
   })
 
-class Keybase extends Component<any> {
-  store: any
+let store
```
chrisnojima: just keep this in the module and not the component

PR #17748, opened 2019-05-29
TS search

chrisnojima:
@keybase/react-hackers also adds lodash types


shared/chat/conversation/messages/wrapper/exploding-height-retainer/index.native.tsx
```diff
@@ -169,7 +169,7 @@ class EmojiTower extends React.Component<
       return
     }
     if (!this.state.running) {
-      this.setState({running: true}, this._update())
+      this.setState({running: true}, this._update)
```
chrisnojima: i believe this is the intention cc: @buoyad

MarcoPolo:
👀

shared/profile/user/index.tsx
```diff
@@ -352,7 +354,7 @@ class User extends React.Component<Props, State> {
       } else {
         chunks.push({
           text: 'Loading...',
-          type: 'loading',
+          type: 'loading' as 'loading',
```
MarcoPolo: It would be nice to update our eslint so we can use `'loading' as const` here instead

shared/search/helpers.tsx
```diff
@@ -7,14 +6,14 @@ import {debounce} from 'lodash-es'
...
 type OwnProps = {
-  onChangeSearchText: ?(s: string) => void,
-  search: (term: string, service: Types.Service) => void,
-  selectedService: Types.Service,
-  searchResultIds: Array<Types.SearchResultId>,
-  selectedSearchId: ?Types.SearchResultId,
-  onUpdateSelectedSearchResult: (id: ?Types.SearchResultId) => void,
-  onAddUser: (id: Types.SearchResultId) => void,
-  searchResultTerm: string,
+  onChangeSearchText: (s: string) => void | null
```
MarcoPolo: bug in code mod, this should be ((s: string) => void) | null. (bug has been fixed)

chrisnojima:
@keybase/react-hackers this changed enought to review the commits since merging recompose in

chrisnojima:
actually i'm just going to get this in. i'm just mostly casting to any to get it to pass short term. merged master in , going in when green

PR #17753, opened 2019-05-29
kbfs_ops: populate the SyncConfig field for favorites

strib:
And add a test.

Also remove some leftover, unnecessary debugging.

Issue: KBFS-4140


go/kbfs/libkbfs/kbfs_ops.go
```diff
@@ -300,6 +300,43 @@ func (fs *KBFSOpsStandard) GetFavoritesAll(ctx context.Context) (
...
+		}
+
+		fb := data.FolderBranch{Tlf: id, Branch: data.MasterBranch}
+		_, h, err := fs.GetRootNodeMetadata(ctx, fb)
+		if err != nil {
+			return keybase1.FavoritesResult{}, err
+		}
+
+		name := string(h.GetCanonicalName())
+		for i, fav := range favs.FavoriteFolders {
```
jzila: Instead of doing a linear search every time, can you create a set/map one time above this loop and just index into it? Some people have lots of favorites and this could hurt performance.

strib: It would only hurt performance if they also had a lot of synced TLFs, and your suggestion would hurt performance a little if they only had one (overhead for the map creation) and makes the code more complex (I'll have to define a new type for the map key, I think).  I don't know what the right answer is performance-wise.  I'll give it a try and see how it looks.

jzila: I suspect this would be a win with like 3+ synced TLFs.

strib: Cool I pushed a new commit that does this optimization.  Let me know if it looks preferable to you.

strib:
cc: @songgao, you can get the sync configs in one RPC now.

songgao:
Thanks!

PR #17758, opened 2019-05-30
 update babel

chrisnojima:
- [x] ios
- [x] android
- [x] desktop
- [x] ci


chrisnojima:
@keybase/react-hackers this updates babel so our linting will allow `blah as const`

PR #17760, opened 2019-05-30
use patched react-native-screens to fix keyboard getting stuck open

buoyad:
See https://github.com/kmagiera/react-native-screens/pull/77 and https://github.com/keybase/react-native-screens/commit/0465a795e7bb59637a76857f11e71c2882341eb3. r? @keybase/react-hackers

### DIFF
--- a/shared/yarn.lock
+++ b/shared/yarn.lock
@@ -5452,6 +5452,11 @@ dateformat@^2.0.0:
   resolved "https://registry.yarnpkg.com/dateformat/-/dateformat-2.2.0.tgz#4065e2013cf9fb916ddfd82efb506ad4c6769062"
   integrity sha1-QGXiATz5+5Ft39gu+1Bq1MZ2kGI=
 
+debounce@^1.2.0:
+  version "1.2.0"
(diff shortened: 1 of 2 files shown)


shared/yarn.lock
```diff
@@ -12772,11 +12772,17 @@ react-native-safe-module@^1.1.0:
   dependencies:
     dedent "^0.6.0"
 
-react-native-screens@1.0.0-alpha.22, "react-native-screens@^1.0.0 || ^1.0.0-alpha":
+"react-native-screens@^1.0.0 || ^1.0.0-alpha":
   version "1.0.0-alpha.22"
```
chrisnojima: why does this still exist

buoyad: Dunno, I `yarn install`ed and this stayed. is there a better command to update the lockfile?

buoyad:
@chrisnojima `react-native-screens` is still there because `react-navigation` depends on it. `debounce` was added because `react-native-screens` depends on it.

shared/yarn.lock
chrisnojima: lets use resolutions to force it over

PR #17761, opened 2019-05-30
More @types

chrisnojima:


.gitignore
```diff
@@ -45,6 +45,7 @@ shared/node_modules
 shared/.static-storybook
 shared/fonts-for-web
 shared/constants/types/rpc*js*
+shared/report.*.json
```
chrisnojima: crash dumps from ts

shared/actions/fs/index.tsx
```diff
@@ -271,18 +271,21 @@ function* folderList(_, action: FsGen.FolderListLoadPayload | FsGen.EditSuccessP
   }
 
   try {
+    // @ts-ignore TS is correct here. TODO fix we're passing buffers as strings
```
chrisnojima: @songgao

shared/app/main.native.tsx
```diff
@@ -42,7 +42,11 @@ class Main extends React.Component<Props> {
     )
   }
 }
-const ViewForGatewayDest = (props: any) => <View {...props} />
+class ViewForGatewayDest extends React.Component {
```
chrisnojima: wants something refable

shared/chat/inbox/index.desktop.tsx
```diff
@@ -49,7 +49,7 @@ class Inbox extends React.PureComponent<T.Props, State> {
     }
 
     if (listRowsResized) {
-      this._list && this._list.resetAfterIndex(0)
+      this._list && this._list.resetAfterIndex(0, true)
```
chrisnojima: the type is wrong, this is optional but we can just pass it

chrisnojima:
@keybase/react-hackers this adds `@types` for most things that are available. the RN ones conflict globally w/ node and others so I just left those out for now. There doesn't seem like a good workaround currently, maybe when we do per-platform builds we can change that

PR #17763, opened 2019-05-30
folder_branch_ops: new ctx for root block fetch on new sync config

strib:
Otherwise `ctx` can be canceled by the caller immediately after `SetSyncConfig` returns; in that case the root block fetch would fail and the prefetch would never be kicked off.


strib:
@songgao this should fix the problem you saw yesterday.

PR #17768, opened 2019-05-30
Cleanup expansions

MarcoPolo:
@keybase/react-hackers @cjb 

Also adds a new helper type called `Unpacked` which can unpack a type wrapped in another type.

i.e. `Unpack<Promise<U>> == U`


shared/actions/chat2/index.tsx
```diff
@@ -1482,11 +1482,7 @@ const onInboxSearchSelect = (state, action: Chat2Gen.InboxSearchSelectPayload) =
   if (!conversationIDKey) {
     return
   }
-  const query = action.payload.query
-    ? action.payload.query // Auto generated from flowToTs. Please clean me!
-    : selected === null || selected === undefined
-    ? undefined
-    : selected.query
+  const query = action.payload.query ? action.payload.query : selected && selected.query
```
mmaxim: This isn't right, needs parens after `:`

MarcoPolo: o hang on, I think this is right (prettier changes it)
```
true ? 'a' : 'foo' && 'bar' // => 'a'
false ? 'a' : 'foo' && 'bar' // => 'bar'
```

mmaxim: Hmm I guess it is, I tried something earlier in the console and thought it wasn't, but trying again now it seems like it's fine. Sorry!

shared/actions/config/index.tsx
```diff
@@ -465,23 +462,25 @@ function* criticalOutOfDateCheck() {
   // check every hour
   while (true) {
     try {
-      const s = yield* Saga.callPromise(RPCTypes.configGetUpdateInfo2RpcPromise, {})
+      const s: Unpacked<ReturnType<typeof RPCTypes.configGetUpdateInfo2RpcPromise>> = yield* Saga.callPromise(
```
buoyad: Can we get the typing from `callPromise`?

MarcoPolo: not really. callPromise returns an effect. The effect is processed by the saga (which gets it from the yield) and returns the value of the promise. For this type to work automatically tsc would have to understand that interplay (which isn't feasible right now)

PR #17773, opened 2019-05-31
updates prettier

chrisnojima:
@keybase/react-hackers also fixes our scripts to work on tsc files etc


shared/devices/paper-key/container.tsx
```diff
@@ -22,8 +22,6 @@ const mergeProps = (stateProps, dispatchProps) => ({
 })
 
 // @ts-ignore codemode issue
-export default Container.connect<OwnProps, _, _, _, _>(
-  mapStateToProps,
-  mapDispatchToProps,
-  mergeProps
-)(PaperKey)
+export default Container.connect<OwnProps, _, _, _, _>(mapStateToProps, mapDispatchToProps, mergeProps)(
```
MarcoPolo: We can remove the generics and the ts-ignore

chrisnojima: lets do taht as another pass

kingleo103:
Hmmmm I fucks with it but not really $gangshit

PR #17775, opened 2019-05-31
TS leftovers

chrisnojima:


chrisnojima:
@keybase/react-hackers bump this is the leftovers. after this we should be all ts for the app code

PR #17776, opened 2019-05-31
Don't ts-ignore connect(), just remove type generics

cjb:
@keybase/react-hackers


cjb:
@MarcoPolo What are we gonna do with the leftover `type OwnProps = {..}`?  Should they be removed or will they be used by a later typing of connect?

jakob223:
If you put the `OwnProps` as the type in one of `mapStateToProps` / `mapDispatch..` / `mergeProps`, then TS will infer the type which is useful for actually type checking things that use the component being defined.

cjb:
@keybase/react-hackers bump

PR #17777, opened 2019-05-31
Cocoapods iOS

buoyad:


buoyad:
@keybase/react-hackers this is ready. Switches us to use cocoapods to manage native iOS dependencies in anticipation of our using unimodules.

- I opted to check in `Pods/` so people don't have to have cocoapods installed unless they're changing deps. Also so the buildbot doesn't need to interact with this.

- [ ] See what's up with buildbot generating `Pods/` changes

buoyad:
K @chrisnojima this has the jsc changes too

chrisnojima:
- [ ] add react-native/android to gitignore

PR #17779, opened 2019-05-31
TS: strict config

chrisnojima:
@keybase/react-hackers this adds a new strict config we can use to grow the strictness of ts
just run
`yarn tsc-strict-watch` and it'll run the stricter config. I don't want to make CI take longer right now so i'm not adding it to that but that could be something we do in the future


shared/package.json
```diff
@@ -46,6 +46,7 @@
     "storybook-build": "cross-env-shell build-storybook -c .storybook -o .static-storybook",
     "rn-storybook": "cross-env-shell storybook start -p 7007",
     "tsc-watch": "node --max-old-space-size=4096 ./node_modules/.bin/tsc --watch",
+    "tsc-strict-watch": "node --max-old-space-size=4096 ./node_modules/.bin/tsc --watch --project ./tsconfig.strict.json",
```
buoyad: Can we add one that doesn't watch?

PR #17740, opened 2019-05-29
TS common-adapters subfolders, misc

jakob223:
- [x] finish fixing `Text`

 - [x] `markdown/`
 - [x] `__test__/`
 - [x] `background-repeat-box/`
 - [x] `confirm-modal/`
 - [x] `floating-box/`
 - [x] `floating-menu/`
 - [x] `header-hoc/`
 - [x] `name-with-icon/`
 - [x] `overlay/`
 - [x] `usernames/`

Issue: DESKTOP-9814

PR #17780, opened 2019-05-31
TS common-adapters: Animated thru Icon

jakob223:
This also includes fixes for @buoyad's comments on #17740, but by the time I made those changes enough other things had been TSed that there were a lot more errors on that branch, and it didn't seem worth it to fix that up in a separate place from this.


jakob223:
bump @keybase/react-hackers

PR #17782, opened 2019-05-31
redact paper keys in log sends

heronhaye:


go/status/log_send.go
```diff
@@ -64,6 +66,37 @@ type LogSendContext struct {
...
+		} else if idx-start+1 == serialPaperKeyWordThreshold {
+			for jdx := start; jdx <= idx; jdx++ {
+				words[jdx] = redactedReplacer
+			}
+			didRedact = true
+		} else if idx-start+1 > serialPaperKeyWordThreshold {
+			words[idx] = redactedReplacer
+		}
+	}
+	if didRedact {
```
jakob223: seems a little unfortunate that we lose all the punctuation here.

heronhaye: I made a commit to address this. Don't know if it could be cleaner though.

go/status/log_send.go
```diff
@@ -72,10 +72,23 @@ const redactedReplacer = "[REDACTED]"
 const serialPaperKeyWordThreshold = 5
 
 func redactPotentialPaperKeys(s string) string {
-	words := strings.Split(noncharacterRxx.ReplaceAllLiteralString(s, " "), " ")
+	doubleDelimited := noncharacterRxx.ReplaceAllFunc([]byte(s), func(x []byte) []byte {
+		return []byte{'~', '~', '~', x[0], '~', '~', '~'} // regexp is single char so we can take first elem
+	})
+	allWords := strings.Split(string(doubleDelimited), "~~~")
+	var checkWords []string
+	wordmap := make(map[int]int)
```
jakob223: Could you put a comment on this saying what it is? It took a few read throughs to figure out what it is.

tintinweb:
@jzila @heronhaye @jakob223 @maxtaco should I be concerned about this If I was using `keybase log send`  before this fix landed? Where would I find this in the release notes?

jzila:
@tintinweb this was shipped in v4.1.0: https://github.com/keybase/client/commits/v4.1.0?after=952fee6c593f42b72831510c7a4b1f5399ae41f3+76

We delete all logs within some timeframe on the order of weeks to months (not sure of the exact number right now), so I wouldn't be concerned about any logs sticking around that were before this change.

maxtaco:
My recollection too, is that this PR had two major intents:

  * preventing people from pasting paper keys as **feedback**
  * preventing transmission of paper keys via command line arguments to `keybase oneshot`

We never put paper keys in logs, that would have been horrible.

heronhaye:
In addition, we already strip sensitive arguments from commands like `keybase oneshot` and `keybase encrypt`. This is just extra protection in case we forget about a case (but it's mainly for the feedback, as Max mentioned--people would send paper keys in their feedback message).

PR #17786, opened 2019-06-02
allow pluses at the end of the device name regexp

pzduniak:
https://github.com/keybase/keybase/pull/3904


zapu:
need to update checkers test:
```
--- FAIL: TestCheckDeviceName (0.00s)
    checkers_test.go:59: 
        	Error Trace:	checkers_test.go:59
        	Error:      	Not equal: 
        	            	expected: true
        	            	actual  : false
        	Test:       	TestCheckDeviceName
```

also due to how device names are checked wrt normalization, it would still not be possible to have a device named e.g. "iPhoneXS" and then add a new one called "iPhoneXS+". But I think this is fine and desired.

PR #17790, opened 2019-06-03
Change GetSecretKeyLocked to support multiple server synced keys

zapu:
`GetSecretKeyLocked` shows non-deterministic behaviour when fetching multiple server synced keys. Add new `(u *User) SyncedSecretKeyWithSka` method to get all server synced keys with `GetSyncedSecretKeys` and use `SecretKeyArg` to select one of them to return.

This fixes `keybase pgp export -s -q KEYID` when user has more than one server synced secret key.

### DIFF
--- a/go/engine/pgp_import_key_test.go
+++ b/go/engine/pgp_import_key_test.go
@@ -25,7 +25,7 @@ func TestPGPImportAndExport(t *testing.T) {
 
 	// try all four permutations of push options:
 
-	fp, _, key := armorKey(t, tc, u.Email)
+	fp, _, key := genPGPKeyAndArmor(t, tc, u.Email)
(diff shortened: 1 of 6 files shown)


go/engine/pgp_import_key_test.go
```diff
@@ -351,7 +351,7 @@ func numPrivateGPGKeys(g *libkb.GlobalContext) (int, error) {
 	return index.Len(), nil
 }
 
-func armorKey(t *testing.T, tc libkb.TestContext, email string) (libkb.PGPFingerprint, keybase1.KID, string) {
+func genPGPKeyAndArmor(t *testing.T, tc libkb.TestContext, email string) (libkb.PGPFingerprint, keybase1.KID, string) {
```
zapu: Renamed `armorKey` to `genPGPKeyAndArmor` to make it clear that it generates a new key each call.

PR #17792, opened 2019-06-03
Better connect typing

chrisnojima:
This PR attempts to improve the typing of our connected components. It does the following

#### Replace the DefinitelyTyped version of react-redux with our more constrained version (https://github.com/keybase/client/pull/17792/files#diff-eb7a4ecd6c78a0c812ac1d6733e3832bR165). 
1. Like our previous form of connect where you have to explicitly give mergeProps w/ OwnProps
1. If the type isn't correct it generates a type we can use to debug what the issue is (see below on a debugging flow)

#### Replace the DefinitelyTyped version of redux with our more constrained version (https://github.com/keybase/client/pull/17792/files#diff-2dad49ff38936f4c2e325a29fcdea655R8). 
1. Enforces that only our generated actions are valid actions
1. Enforces that state is TypedState
1. Enforces dispatch only dispatches our typed actions

### How the connect typing works now

As before we pull in connect/namedConnect from `utils/container`.
Typically we export default the connector like

```
export default Container.namedConnect(
  mapStateToProps,
  mapDispatchToProps,
  (stateProps, dispatchProps, ownProps: OwnProps) => ({
...
}),
'MyComponent')(MyComponent)
```

mapStateToProps looks like
`(state: TypedState, ownProps: OwnProps)`
mapDispatchToProps looks like
`(dispatch: TypedDispatch, ownProps: OwnProps)`
mergeProps looks like
```
( stateProps: ReturnType<typeof mapStateToProps>,
  dispatchProps: ReturnType<typeof mapDispatchToProps>,
  ownProps: OwnProps)
```

If you inline the functions in the mergeProps TS will infer the parameters for you (`TypedState/TypedDispatch/ReturnType<...>`) so the least boilerplate form is

```
export default Container.namedConnect(
  (state, ownProps: OwnProps) => ({}),
  (dispatch, ownProps: OwnProps) => ({}),
  (stateProps, dispatchProps, ownProps: OwnProps) => ({}),
'MyComponent')(MyComponent)
```
 above, state, dispatch, stateProps, dispatchProps are typed correctly for you. you just have to supply OwnProps

I think its fine to inline them if they're small. If they're larger then we should leave them out and maybe mostly inline mergeProps

#### Debugging containers

If TS complains that the container isn't compatible with the component its wrapping it now will complain that there is `no callable signature`. This is because the type of connect will now give you diagnostic info instead of a real type if it fails (https://github.com/keybase/client/pull/17792/files#diff-eb7a4ecd6c78a0c812ac1d6733e3832bR154).

The easiest way to debug is change a typical connect to
```
const C = Container.namedConnect(...
export default C
```

and introspect C in your editor. It will show you keys it thinks are missing and the type of the MergedProps and Wrapped component.

A common error I saw while fixing up the containers is mergeProps deriving some strict constants like 
```
status: stateProps.something ? 'good' : 'bad'
```
where the Wrapped component had
` type Props = {status: 'good' | 'bad'}`
and the Merged props were
`type {status: string}`
a workaround was to extract the type from props
`type StatusType = Props['status']`
and use that in a cast / guard


chrisnojima:
@MarcoPolo lets discuss

chrisnojima:
@MarcoPolo bump

chrisnojima:
@keybase/react-hackers

shared/override-d.ts/react-redux/index.d.ts
```diff
@@ -0,0 +1,648 @@
...
+
+// KB added
+export type ConnectedComponentType<TMergedProps, TOwnProps> = <C extends ComponentType<any>>(
+  component: C
+) => TMergedProps extends React.ComponentProps<C>
+  ? ConnectedComponentClass<C, TOwnProps>
+  : [
+      // just to help you debug what's going on
+      Exclude<keyof GetProps<C>, keyof TMergedProps>,
+      Exclude<keyof GetProps<C>, keyof GetProps<C>>,
```
MarcoPolo: ? Isn't this always going to return an empty set?

Also maybe it would be nice to return a string literal to make clear this is "Debug Info"

shared/chat/conversation/messages/message-popup/payment/container.tsx
```diff
@@ -237,7 +246,7 @@ const RequestPaymentPopup = Container.connect(
   requestMapStateToProps,
   requestMapDispatchToProps,
   requestMergeProps
-)(PaymentPopup)
+)(PaymentPopup) as any
```
MarcoPolo: What happened here?

shared/override-d.ts/react-redux/index.d.ts
chrisnojima: oh sorry this was a copy and paste error, it's supposed to be TMergedProps first param

shared/chat/conversation/messages/message-popup/payment/container.tsx
chrisnojima: the type was complicated and i didn't want to fix it up short term

PR #17793, opened 2019-06-03
fix FsConstants.makeUUID typing

songgao:


songgao:
It was working before. I think both Buffer and string works. Just that our RPCTypes types those as strings. There's some small discussion in #kbfs

PR #17795, opened 2019-06-03
Reduce homedir priorities test flakes, standardize (ctx, t) order in test helpers

jakob223:
The former due to some CI flakes, the latter because `go-lint` was being annoying about the former.

### DIFF
--- a/go/kbfs/libkbfs/kbfs_ops_test.go
+++ b/go/kbfs/libkbfs/kbfs_ops_test.go
@@ -186,9 +186,8 @@ func kbfsOpsInit(t *testing.T) (mockCtrl *gomock.Controller,
 	return mockCtrl, config, ctx, cancel
 }
 
-func kbfsTestShutdown(
-	t *testing.T, mockCtrl *gomock.Controller, config *ConfigMock,
-	ctx context.Context, cancel context.CancelFunc) {
+func kbfsTestShutdown(ctx context.Context,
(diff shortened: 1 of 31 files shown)


go/kbfs/libkbfs/kbfs_ops_test.go
```diff
@@ -186,9 +186,8 @@ func kbfsOpsInit(t *testing.T) (mockCtrl *gomock.Controller,
 	return mockCtrl, config, ctx, cancel
 }
 
-func kbfsTestShutdown(
-	t *testing.T, mockCtrl *gomock.Controller, config *ConfigMock,
-	ctx context.Context, cancel context.CancelFunc) {
+func kbfsTestShutdown(ctx context.Context,
```
strib: Small quibble to have `ctx` on the next line down, and keep the rest of the params on lines less than 80 chars.

go/kbfs/libkbfs/kbfs_ops_test.go
```diff
@@ -258,23 +257,23 @@ func kbfsOpsInitNoMocks(t *testing.T, users ...kbname.NormalizedUsername) (
 	return config, session.UID, ctx, cancel
 }
 
-func kbfsTestShutdownNoMocks(t *testing.T, config *ConfigLocal,
-	ctx context.Context, cancel context.CancelFunc) {
+func kbfsTestShutdownNoMocks(ctx context.Context, t *testing.T,
```
strib: You didn't change this but would be nice to move these params down to the next line too, while you're here.

PR #17797, opened 2019-06-03
Remove isolatedModules

MarcoPolo:
@keybase/react-hackers 

Without this flag our update time in tsc watch is < 1s for small changes.


MarcoPolo:
For now making a new tsconfig and yarn command `yarn tsc-fast`. If you get an error from CI or a normal `yarn tsc` but not in `yarn tsc-fast`, let me know!

Note: You must delete your old cache `rm -rf .tsOut` before using this or you will get wrong errors.

PR #17798, opened 2019-06-04
Move tsouts

MarcoPolo:
Keep em in one place @keybase/react-hackers


shared/.tsOuts/README.md
```diff
@@ -0,0 +1 @@
+Where ts cache files live
```
MarcoPolo: Needed so git recreates this folder

PR #17805, opened 2019-06-04
Use upgradeable_secret_store with secret_store_file for android 22

zapu:


go/libkb/secret_store_android.go
```diff
@@ -0,0 +1,50 @@
...
+	if mctx.G().Env.ForceSecretStoreFile() {
+		// Allow use of file secret store on Android, for debugging or use with
+		// Termux (https://termux.com/).
+		return secFile
+	}
+
+	secAndroid := &secretStoreAndroid{}
+
+	var androidOsVersion int64
+	if v, err := strconv.ParseInt(mctx.G().MobileOsVersion, 10, 32); err != nil {
```
heronhaye: may be better to do this in bind

go/libkb/secret_store_android.go
```diff
@@ -0,0 +1,50 @@
...
+		androidOsVersion = v
+	}
+	shouldUpgradeOpportunistically := func() bool {
+		return true
+	}
+	shouldStoreInFallback := func(options *SecretStoreOptions) SecretStoreFallbackBehavior {
+		if androidOsVersion <= 22 {
+			// Use file based secret store on old Android version (or when
+			// version detection didn't work).
```
heronhaye: might mention it has to do with nil value, not obvious I think

go/libkb/secret_store_upgradeable.go
```diff
@@ -20,50 +33,82 @@ func NewSecretStoreUpgradeable(a, b SecretStoreAll, shouldUpgradeOpportunistical
...
+			mctx.Debug("ClearSecret error=<nil> for %s from store B", username)
 		} else {
 			mctx.Debug("Failed to clear secret for %s from secretstore b: %s", username, clearBErr)
 		}
 		return nil
 	}
 
-	mctx.Warning("Failed to reach system keyring (%s), falling back to file-based secret store.", err1)
+	if fallbackBehavior == SecretStoreFallbackBehaviorNever {
+		mctx.Warning("Failed to reach system keyring (store A: %s), do not falling back to store B because of fallback behavior.", err1)
```
heronhaye:
```suggestion
		mctx.Warning("Failed to reach system keyring (store A: %s), do not fall back to store B because of fallback behavior.", err1)
```

go/libkb/secret_store_android.go
zapu: of course i messed this up, heh
should be
`if v, err := strconv.ParseInt(mctx.G().MobileOsVersion, 10, 32); err == nil`

Issue #17670, opened 2019-05-21
Subteam rename permission issue


Lumenaut:
Here's the scenario...

1 - Create a team;
2 - Create a subteam;
3 - Add members to the subteam;
4 - Rename the subteam;
5 - Members of the subteam don't have access/permissions to post again, yet the renamed subteam shows the correct number of members.

This is the message shown at the Keybase app:

<img width="1180" alt="Screenshot 2019-05-21 at 17 15 20" src="https://user-images.githubusercontent.com/33041885/58114546-db68f600-7bef-11e9-9906-472965def46f.png">

PR #17807, opened 2019-06-04
Do not offer underprivileged admins to rename team

mlsteele:
Admins of a subteam can’t rename a subteam. You need admin of the parent for that.


joshblum:
addresses https://github.com/keybase/client/issues/17670

PR #17811, opened 2019-06-04
macOS notarization

mmaxim:
Major things

1. Add new options to `codesign` in all the places we invoke it. The options are `--timestamp` which adds a secure timestamp to the sig, and `--options runtime` which turns on the hardened runtime.
2. Add a new entitlements file for `codesign` which turns off the parts of the hardened runtime that stop an app from using memory intended for use by a JIT. This is memory marked as both executable and writable. The second option in that file is because Electron (the version we are using at least) doesn't mark the memory allocated as `MAP_JIT`, so we need to let the OS know we are cool with that.
3. Generate a new kbfuse.bundle with the new sig options. The binaries for these are just in source control (probably because they are extremely difficult to build from scratch), so luckily all I had to do was regen the sig. 
4. Hack up `kbnm` to workaround a Go bug that would mark the macOS SDK at 10.7 under most circumstance. The only way I could figure out how to solve this is to force it to link against packages from the rest of the app (I don't know why this works). 
5. Add a new step to `package_darwin.sh` that submits for notarization and waits for a response before continuing. You can pass `NONOTARIZE=1` to `build_app.sh` (and hence the build bot) to skip it.


mmaxim:
cc @chrisnojima

go/kbnm/hostmanifest/whitelist_darwin.go
```diff
@@ -6,9 +6,12 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+
+	"github.com/keybase/client/go/chat/attachments"
 )
 
 func wrapWriteErr(err error, hostsPath string) error {
+	attachments.LinkNoop()
```
chrisnojima: Should comment why we do this

osx/Keybase.xcodeproj/project.pbxproj
```diff
@@ -1399,7 +1399,7 @@
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				STRIP_INSTALLED_PRODUCT = NO;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
-				SWIFT_VERSION = 3.0;
+				SWIFT_VERSION = 4.0;
```
chrisnojima: Is that needed

mmaxim: Xcode doesnt build that version anymore, it's an error.

PR #17814, opened 2019-06-04
Update app window icons.

adamjspooner:


shared/router-v2/header/index.desktop.tsx
```diff
@@ -70,6 +71,18 @@ const SystemButtons = () => (
...
+  }
+  // We need to forceUpdate when maximizing and unmaximizing the window to update the
+  // app icon on Windows and Linux.
+  _registerWindowEvents() {
+    if (Platform.isDarwin) return
+    const win = BrowserWindow.getFocusedWindow()
+    if (!win) return
+    win.on('maximize', () => this.forceUpdate())
+    win.on('unmaximize', () => this.forceUpdate())
+  }
```
songgao: I think we need to unsubscribe in `componentWillUnmount()` as well.

shared/router-v2/header/index.desktop.tsx
```diff
@@ -70,6 +71,18 @@ const SystemButtons = () => (
...
+    this._registerWindowEvents()
+  }
+  // We need to forceUpdate when maximizing and unmaximizing the window to update the
+  // app icon on Windows and Linux.
+  _registerWindowEvents() {
+    if (Platform.isDarwin) return
+    const win = BrowserWindow.getFocusedWindow()
+    if (!win) return
+    win.on('maximize', () => this.forceUpdate())
+    win.on('unmaximize', () => this.forceUpdate())
```
buoyad: Deregister these in `componentWillUnmount`

adamjspooner: Is it just `win.off('event')`? I'm not finding it in the docs.

buoyad: Looks like it's `removeListener('event', handler)`. I think you can just pass `this.forceUpdate` instead of `() => this.forceUpdate()` to all these and that'll do the trick.

adamjspooner: Thanks!

adamjspooner: @buoyad, it didn't work without passing the anonymous function.

PR #17816, opened 2019-06-04
tweaks on the Oops screen

songgao:


jakob223:
Issue: HOTPOT-30

shared/fs/simple-screens/oops.tsx
```diff
@@ -93,9 +113,21 @@ const Oops = (props: Props) => {
   }
 }
 
-export default Oops
+const mapDispatchToProps = (dispatch, props: OwnProps) => ({
+  openParent: () =>
+    dispatch(
+      RouteTreeGen.createNavigateAppend({
+        path: [{props: {path: Types.getPathParent(props.path)}, selected: 'main'}],
```
jakob223: Hmm, I clicked a link in chat, to get to a fake folder, then I clicked this button, and then I pressed "back". The back button took me back to the error page. I understand why that happened but I think maybe it should take me all the way back to chat. I could go either way though. WDYT?

songgao: Good question; I thought about that but didn't think it was easy to define. You could land in this screen not only from chat but files too -- for example something get deleted while you are in there, or your access to a team is removed when you are in the team. We could probably keep popping route until it's not an oops screen anymore but that could might complicated to understand as user. Also note that if we navigateUp twice, there'd be two animations flying by on mobile.

We can always loop design in and see how they wanna change it. Merging this for now!

PR #17818, opened 2019-06-04
md_ops: ignore writer-check error if the merkle root is too old

strib:
If a TLF was last written by a device that was revoked too long ago (i.e., the service doesn't support checking its Merkle root because it's too old), there's nothing much we can do but trust the server. Note that we already do this for other Merkle checks -- this one was just missing until now.

### DIFF
--- a/go/kbfs/libkbfs/md_ops.go
+++ b/go/kbfs/libkbfs/md_ops.go
@@ -375,7 +375,6 @@ func (md *MDOpsStandard) checkRevisionCameBeforeMerkle(
 	verifyingKey kbfscrypto.VerifyingKey, irmd ImmutableRootMetadata,
 	root keybase1.MerkleRootV2, timeToCheck time.Time) (err error) {
 	ctx = context.WithValue(ctx, ctxMDOpsSkipKeyVerification, struct{}{})
-
 	kbfsRoot, merkleNodes, rootSeqno, err :=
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/md_ops.go
```diff
@@ -739,15 +738,19 @@ func (mbtc merkleBasedTeamChecker) IsTeamWriter(
...
+				"the revoked key: %+v", err)
+	} else if err != nil {
 		return false, err
+	} else {
+		// TODO(CORE-8199): pass in the time for the writer downgrade.
+		err = mbtc.md.checkRevisionCameBeforeMerkle(
+			ctx, mbtc.rmds, verifyingKey, mbtc.irmd, root, time.Time{})
+		if err != nil {
+			return false, err
+		}
```
jzila: I find this somewhat difficult to read. How about?

```go
switch errors.Cause(err).(type) {
case nil:
	// TODO(CORE-8199): pass in the time for the writer downgrade.
	err = mbtc.md.checkRevisionCameBeforeMerkle(
		ctx, mbtc.rmds, verifyingKey, mbtc.irmd, root, time.Time{})
	if err != nil {
		return false, err
	}
case libkb.MerkleClientError:
	if m.IsOldTree() {
		mbtc.md.vlog.CLogf(...)
	} else {
		return false, err
	}
default:
	return false, err
}
```

PR #17819, opened 2019-06-04
favorites: adjust TLF mtimes based on the edit history times

strib:
To avoid counting GC and git modifications in the favorite mtimes used to sort folders, cross-check them with the time of of the corresponding edit history.  If the favorites time is newer, we can safely set it to the edit history time.  If we don't yet know the edit history of the TLF, that means it's not in the most recent edit history list, and we can just set the time to be below the oldest known time, to keep it below the most-recently-edited TLFs in the sorted list.

Apparently the favorite mtime is incorrectly in seconds, rather than in milliseconds, so some math is required.

Issue: KBFS-4200

### DIFF
--- a/go/kbfs/libkbfs/favorites.go
+++ b/go/kbfs/libkbfs/favorites.go
@@ -300,6 +300,56 @@ func (f *Favorites) closeReq(req *favReq, err error) {
 	}
 }
 
+func (f *Favorites) crossCheckWithEditHistory() {
+	// NOTE: Ideally we would wait until all edit activity processing
+	// had completed in the FBO before we do these checks, but I think
+	// in practice when the mtime changes, it'll be the edit activity
+	// processing that actually kicks off these favorites activity,
+	// and so that particular race won't be an issue.  If we see
+	// weirdness here though, it might be worth revisiting that
+	// assumption.
+
+	// Fix the favorite times to be the latest known edit history
+	// times, if possible.  If not possible, set them to be lower than
+	// the minimum known time, if they're not already.
+	var minTime keybase1.Time
+	uh := f.config.UserHistory()
+	tlfsWithNoHistory := make(map[favorites.Folder]favorites.Data)
+	for fav, data := range f.favCache {
+		h := uh.GetTlfHistory(tlf.CanonicalName(fav.Name), fav.Type)
+		if h.ServerTime == 0 {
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/favorites.go
```diff
@@ -300,6 +300,56 @@ func (f *Favorites) closeReq(req *favReq, err error) {
...
+	// had completed in the FBO before we do these checks, but I think
+	// in practice when the mtime changes, it'll be the edit activity
+	// processing that actually kicks off these favorites activity,
+	// and so that particular race won't be an issue.  If we see
+	// weirdness here though, it might be worth revisiting that
+	// assumption.
+
+	// Fix the favorite times to be the latest known edit history
+	// times, if possible.  If not possible, set them to be lower than
+	// the minimum known time, if they're not already.
```
jakob223: Can you mention in this comment the reason the times might be wrong (i.e. GC, etc.)?

go/kbfs/libkbfs/favorites.go
```diff
@@ -300,6 +300,56 @@ func (f *Favorites) closeReq(req *favReq, err error) {
...
+		if h.ServerTime == 0 {
+			if data.TlfMtime != nil {
+				tlfsWithNoHistory[fav] = data
+			}
+			continue
+		}
+		if minTime == 0 || h.ServerTime < minTime {
+			minTime = h.ServerTime
+		}
+		// Seems like the mtime in the favorites list is incorrectly
```
jakob223: maybe we should fix this? If you don't want to because it includes frontend code, make a ticket and song or I could take a whack at it?

strib: @heronhaye said he'd fix this, so I'll hold off on merging this until that's in.

heronhaye: Should be fixed in kbweb master, let me know if not

strib:
@jakob223 did you mean to approve this, or were you wanting another review?

jakob223:
I meant to approve in the event that you are not waiting for the `kbweb` 1000x fix. Since that fix is in, though, can you revise this to account for that?

strib:
Will do.

strib:
@jakob223 done and ready for a thumb!

PR #17821, opened 2019-06-05
First pass at removing unnecessary @ts-ignore comments

cjb:
@keybase/react-hackers 

Unlike Flow, TS isn't able to determine whether a ts-ignore line is useless: that is, whether it's silencing an error that wouldn't actually exist if the comment was missing.  We used ts-ignore a lot to mean "this error is temporary because we haven't migrated some files from Flow to TS yet" during the migration.

So I wrote a script that removes all ts-ignores, runs tsc once, and then reinserts the comments that used to be on lines that have become errors.  We can keep running the script occasionally.

Stats: Removed 259 out of 547 ts-ignore lines on master (47%).

You can verify that the removals are safe by noticing that tsc still reports 0 errors after applying this PR.

### DIFF
--- a/shared/actions/__tests__/add-new-device.tsx
+++ b/shared/actions/__tests__/add-new-device.tsx
@@ -211,7 +211,6 @@ describe('reply with device type', () => {
     expect(() =>
       makeInit({
         initialStore: {
-          // @ts-ignore codemod issue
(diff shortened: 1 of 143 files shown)


cjb:
> Where's the script

Will tidy up and PR it separately.

PR #17824, opened 2019-06-05
Plumb mobile os version to go side

zapu:
Rationale: we need to know Android version on go side, because secret store behavior will be different on old (android <=22) systems.


zapu:
Android builds and seems to work, I couldn't test iOS though

mmaxim:
I thought you had a Mac somewhere? If you do it would be great to test it first

zapu:
I had some issues in building ios, but might as well figure it out now if I'm going to do more frontend work in near future. thanks!

zapu:
Tried it with xcode, fixed few issues, seems to work at least on simulators / emulators - confirmed with logcat on android side and in xcode terminal that right version string is getting plumbed.

zapu:
Rebased, @mmaxim i think this is ready for review

shared/android/app/src/main/java/io/keybase/ossifrage/util/MobileOSVersion.java
```diff
@@ -0,0 +1,7 @@
+package io.keybase.ossifrage.util;
+
+public final class MobileOSVersion {
+    public static String Get() {
+        return Integer.toString(android.os.Build.VERSION.SDK_INT);
```
buoyad: Can we get this without adding this util?

zapu: Yes - alternative is adding doing `Integer.toString(android.os.Build.VERSION.SDK_INT)` in both call sites of `initOnce`. So in `KeybasePushNotificationListenerService.java` and `MainActivity.java`. I can change it if that's preferable. Thanks!

PR #17825, opened 2019-06-05
use typeRoots. remove allowJs/checkJs

chrisnojima:
@keybase/react-hackers changes the tsconfig slightly


shared/tsconfig.base.json
```diff
@@ -1,21 +1,25 @@
...
     "jsx": "preserve",
     "module": "commonjs",
-    "noImplicitAny": false,
+    "noFallthroughCasesInSwitch": true,
+    "outDir": "./.tsOuts/.tsOut/emit",
     "paths": {
-      "rn-fetch-blob": ["custom-d.ts/rn-fetch-blob.d.ts"],
-      "popper": ["custom-d.ts/popper.d.ts"],
-      "framed-msgpack-rpc": ["custom-d.ts/framed-msgpack-rpc.d.ts"]
+      "rn-fetch-blob": ["custom-d.ts/rn-fetch-blob/index.d.ts"],
```
chrisnojima: the def in node_modules is wrong so we need to swap it out

shared/tsconfig.base.json
```diff
@@ -1,21 +1,25 @@
...
     },
     "sourceMap": true,
     "target": "es2017",
-    "outDir": "./.tsOuts/.tsOut/emit",
-    "tsBuildInfoFile": "./.tsOuts/.tsOut/cache"
+    "tsBuildInfoFile": "./.tsOuts/.tsOut/cache",
+    "typeRoots": [
```
chrisnojima: otherwise we have a local version of `@types` in custom-d.ts

PR #17826, opened 2019-06-05
Use expo-sms for team invite by phone

buoyad:
Based on #17777. r? @keybase/react-hackers


shared/package.json
```diff
@@ -125,6 +125,7 @@
     "emoji-datasource-apple": "4.1.0",
     "emoji-mart": "2.9.2",
     "exif-js": "2.3.0",
+    "expo-sms": "^5.0.0",
```
chrisnojima: exact

PR #17829, opened 2019-06-05
Implement conflict banner link to send feedback

jakob223:
Issue: Y2K-42

### DIFF
--- a/shared/actions/fs/index.tsx
+++ b/shared/actions/fs/index.tsx
@@ -986,12 +986,7 @@ const waitForKbfsDaemon = (state, action: ConfigGen.InstallerRanPayload | FsGen.
 const startManualCR = (state, action) =>
   RPCTypes.SimpleFSSimpleFSClearConflictStateRpcPromise({
     path: Constants.pathToRPCPath(action.payload.tlfPath),
-  }).then(() =>
-    FsGen.createTlfCrStatusChanged({
-      status: Types.ConflictState.InManualResolution,
-      tlfPath: action.payload.tlfPath,
-    })
-  ) // TODO: deal with errors
+  }).then(() => FsGen.createFavoritesLoad())
 
 const updateKbfsDaemonOnlineStatus = (
   state,
--- a/shared/fs/banner/conflict-banner-container.tsx
+++ b/shared/fs/banner/conflict-banner-container.tsx
@@ -43,7 +43,7 @@ const mergeProps = (s, d, o: OwnProps) => ({
   ...d,
   conflictState: s._tlf.conflict.state,
   isUnmergedView: Constants.isUnmergedView(o.path),
-  tlfPath: Types.pathToTlfPath(o.path),
+  tlfPath: Constants.getTlfPath(o.path),
 })
 
 const ConnectedBanner = namedConnect<OwnProps, {}, DispatchProps, Props, {}>(
--- a/shared/fs/banner/conflict-banner-container.tsx
+++ b/shared/fs/banner/conflict-banner-container.tsx
@@ -1,6 +1,8 @@
 import * as Constants from '../../constants/fs'
 import * as Types from '../../constants/types/fs'
+import * as SettingsConstants from '../../constants/settings'
 import * as FsGen from '../../actions/fs-gen'
+import * as RouteTreeGen from '../../actions/route-tree-gen'
 import {namedConnect} from '../../util/container'
 import ConflictBanner, {Props} from './conflict-banner'
 import openUrl from '../../util/open-url'
@@ -9,7 +11,9 @@ type OwnProps = {
   path: Types.Path
 }
 
-const mapStateToProps = (state, ownProps: OwnProps) => ({})
+const mapStateToProps = (state, ownProps: OwnProps) => ({
+  _tlf: Constants.getTlfFromPath(state.fs.tlfs, ownProps.path),
+})
 type DispatchProps = {
   onFeedback: () => void
   onFinishResolving: () => void
@@ -18,7 +22,17 @@ type DispatchProps = {
   onStartResolving: () => void
 }
 const mapDispatchToProps = (dispatch, ownProps: OwnProps) => ({
-  onFeedback: () => {},
+  onFeedback: () =>
+    dispatch(
+      RouteTreeGen.createNavigateAppend({
+        path: [
+          {
+            props: {feedback: `Conflict Resolution failed in \`${ownProps.path}\`.\n`},
+            selected: SettingsConstants.feedbackTab,
+          },
+        ],
+      })
+    ),
   onFinishResolving: () => {},
   onHelp: () => openUrl('https://keybase.io/docs/kbfs/understanding_kbfs#conflict_resolution'),
   onSeeOtherView: () => {},
@@ -29,7 +43,7 @@ const mergeProps = (s, d, o: OwnProps) => ({
   ...d,
   conflictState: s._tlf.conflict.state,
   isUnmergedView: Constants.isUnmergedView(o.path),
-  tlfName: s._tlf.name,
+  tlfPath: Types.pathToTlfPath(o.path),
 })
 
 const ConnectedBanner = namedConnect<OwnProps, {}, DispatchProps, Props, {}>(
--- a/shared/reducers/fs.tsx
+++ b/shared/reducers/fs.tsx
@@ -602,24 +602,7 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
         // This should not happen.
         return state
       }
-      return state.update('tlfs', tlfs =>
-        tlfs.update(parsedPath.tlfType, tlfList =>
-          tlfList.update(parsedPath.tlfName, tlf =>
-            tlf.update('conflict', tlfConflict => {
-              if (
-                tlfConflict.state === 'in-manual-resolution' &&
-                (newState === 'in-conflict-stuck' || newState === 'in-conflict-not-stuck')
-              ) {
-                // If the conflict is being manually resolved, ignore new
-                // conflicts that come in.
-                return tlfConflict
-              } else {
-                return tlfConflict.set('state', newState)
-              }
-            })
-          )
-        )
-      )
+      return state.setIn(['tlfs', parsedPath.tlfType, parsedPath.tlfName, 'conflict', 'state'], newState)
     case FsGen.setPathSoftError:
       return state.update('softErrors', softErrors =>
         softErrors.update('pathErrors', pathErrors =>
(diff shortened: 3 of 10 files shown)


shared/settings/feedback/container.desktop.tsx
```diff
@@ -18,9 +18,10 @@ const mapDispatchToProps = dispatch => ({
   onSendFeedback: (feedback, sendLogs) => dispatch(SettingsGen.createSendFeedback({feedback, sendLogs})),
 })
 
-const mergeProps = (s, d, o) => ({
+const mergeProps = (s, d, o: OwnProps) => ({
   ...s,
   ...d,
+  feedback: o.routeProps.get('feedback'),
```
songgao: I think `OwnProps.routeProps` is going away. Should use `getRouteProps` instead https://github.com/keybase/client/blob/49eb317d732f264a69a65ac45ac33c4fc6be3d7a/shared/fs/container.tsx#L37

shared/settings/feedback/index.tsx
```diff
@@ -37,6 +38,10 @@ class Feedback extends React.Component<Props, State> {
     sendLogs: true,
     showSuccessBanner: false,
   }
+  componentWillMount(): void {
```
songgao: This is legacy now: https://reactjs.org/docs/react-component.html#legacy-lifecycle-methods

`componentDidMount()` would work I think

shared/fs/banner/conflict-banner-container.tsx
```diff
@@ -18,7 +22,17 @@ type DispatchProps = {
...
+    dispatch(
+      RouteTreeGen.createNavigateAppend({
+        path: [
+          {
+            props: {feedback: `Conflict Resolution failed in \`${ownProps.path}\`.\n`},
+            selected: SettingsConstants.feedbackTab,
+          },
+        ],
+      })
+    ),
```
songgao: Hmm this doesn't seem to be used anywhere. Did you forget to add it to `actions` in the component? Also note that we don't have the feedback screen on desktop, so would probably need a `isMobile` guard there.

shared/settings/feedback/index.tsx
jakob223: You're not supposed to `setState` in `componentDidMount` because that triggers a double render. Looks like I need a `constructor()`.

shared/fs/banner/conflict-banner-container.tsx
jakob223: We do have the feedback screen on desktop now, I worked on that so we could do this.

I'm confused by "doesn't seem to be used anywhere". It's listed as a routeProp in the feedback container, and passed there. It's included in the `feedbackAction` in the component.

jakob223:
@songgao I've addressed all of your feedback. I'm going to wait to rebase until the other PR is merged and you've looked at this again.

shared/fs/banner/conflict-banner-container.tsx
songgao: I meant to say `onFeedback` wasn't called in the component, but looks like I just missed it. My bad!

Sounds good; feedback screen in desktop sounds awesome!

shared/settings/feedback/index.tsx
```diff
@@ -31,12 +32,11 @@ type State = {
...
 
 class Feedback extends React.Component<Props, State> {
-  state = {
-    email: null,
-    feedback: '',
-    sendLogs: true,
-    showSuccessBanner: false,
+  constructor(props) {
+    super(props)
+    this.state = {email: null, feedback: this.props.feedback, sendLogs: true, showSuccessBanner: false}
```
songgao: FWIW you could also do this and get rid of the constructor itself:

```
class Feedback ... {
  state = {
    ...
  }
}
```

shared/reducers/fs.tsx
```diff
@@ -595,6 +595,31 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
             : driverStatus
         )
       )
+    case FsGen.tlfCrStatusChanged:
+      const parsedPath = Constants.parsePath(action.payload.tlfPath)
+      const newState = action.payload.status
+      if (parsedPath.kind === Types.PathKind.Root || parsedPath.kind === Types.PathKind.TlfList) {
+        // This should not happen.
+        return state
+      }
```
songgao: `in-group-tlf` and `in-team-tlf` also shouldn't happen right? So could probably test for `group-tlf` and `team-tlf`.

shared/reducers/fs.tsx
```diff
@@ -595,6 +595,31 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
+      return state.update('tlfs', tlfs =>
+        tlfs.update(parsedPath.tlfType, tlfList =>
+          tlfList.update(parsedPath.tlfName, tlf =>
+            tlf.update('conflict', tlfConflict => {
+              if (
+                tlfConflict.state === 'in-manual-resolution' &&
+                (newState === 'in-conflict-stuck' || newState === 'in-conflict-not-stuck')
+              ) {
+                // If the conflict is being manually resolved, ignore new
+                // conflicts that come in.
```
songgao: If I understand correctly, the only state that InManualResolution can transit into is Finishing. After that, this TLF is deleted since all the needed data has been supposedly copied into the main one where the conflict state is cleared.

shared/actions/fs/index.tsx
```diff
@@ -988,6 +988,16 @@ const waitForKbfsDaemon = (state, action: ConfigGen.InstallerRanPayload | FsGen.
     })
 }
 
+const startManualCR = (state, action) =>
+  RPCTypes.SimpleFSSimpleFSClearConflictStateRpcPromise({
+    path: Constants.pathToRPCPath(action.payload.tlfPath),
+  }).then(() =>
+    FsGen.createTlfCrStatusChanged({
```
songgao: I don't think this action should be triggered. Once ClearConflictState is called, the original conflicted TLF splits into two. Then we'll actually have two TLF entries in favorites list for the same TLF, one for the server view, and one for local branch. See [this](https://github.com/keybase/client/blob/master/protocol/avdl/keybase1/simple_fs.avdl#L373-L388). Looks like we should trigger favoritesLoad here.

shared/settings/feedback/index.tsx
jakob223: Can I? I'm using `this.props.feedback` here so I think I can't.

shared/reducers/fs.tsx
jakob223: I only tested for these two because that way `parsedPath` is definitely a type that has a `tlfPath`. I think I'd rather this be permissive and work if given an in-TLF path. WDYT?

shared/reducers/fs.tsx
jakob223: but if we're in the middle of resolving a conflict, and then another one appears on the same TLF, what should we do?

shared/actions/fs/index.tsx
jakob223: Yeah, you're right. Will fix.

shared/settings/feedback/index.tsx
songgao: Yup; we have a bunch of them:

```
%   rg 'state = \{.*this\.props' | head -n1
devices/index.tsx:  state = {revokedExpanded: this.props._stateOverride ? this.props._stateOverride.revokedExpanded : false}
```

shared/reducers/fs.tsx
songgao: `tlfPath` is supposed to be just paths to TLFs I think; I'd actually rather be strict here than causing confusion down the road.

shared/reducers/fs.tsx
songgao: You mean after the TLF already branched into two? Then conflict can indeed happen again, but that'd be only on the server-view one, and it'd just be able to branch into a third TLF. The KBFS daemon would give it a different suffix in this case, so all three TLFs show up in favorites fine.

shared/settings/feedback/index.tsx
songgao: (You don't have to change this though; just wanted to make sure you know that!)

jakob223: Cool! I think your way is cleaner then.

shared/reducers/fs.tsx
jakob223: Let's say that you have a TLF, `jakob223,songgao`. You get a conflict in it:
 - jakob223,songgao [in-conflict-stuck]
and you press "Start conflict resolution". Now you have 2 TLFs:
 - jakob223,songgao [in-manual-resolution]
 - jakob223,songgao (conflict #1) [in-manual-resolution]
and now you get a second conflict. I claim that `jakob223,songgao` should not switch from `in-manual-resolution` to `in-conflict-stuck` because that would hinder the ability of the user to finish resolving the first conflict. This amounts to forcing them to resolve the first conflict before another appears.

WDYT?

songgao: it'd be like this:

```
jakob223,songgao
```

gets stuck

```
jakob223,songgao [conflicted]
```

hit start manual resolve

```
jakob223,songgao
jakob223,songgao (conflict #1)
```

first one (server view) get conflicted again

```
jakob223,songgao [conflicted]
jakob223,songgao (conflict#1)
```

hit manual resolve in the first one

```
jakob223,songgao
jakob223,songgao (conflict#1)
jakob223,songgao (conflict#2)
```

Note that branched out TLFs (with the conflict suffixes) can never get conflicted again -- they're a local branch and becomes readonly.

jakob223: But since we are only showing the "finish conflict resolution" button on the original / main-line TLF, we don't expose any way for the user to mark the current CR as finished in the 
```
jakob223,songgao [conflicted]
jakob223,songgao (conflict#1)
```
state. Additionally, it's not clear which conflict should be marked as finished when the user clicks "finish" in `jakob223,songgao` in the case below:
```
jakob223,songgao
jakob223,songgao (conflict#1)
jakob223,songgao (conflict#2)
```

songgao: You are right. Not sure what the best answer is here. Perhaps "finish cr" should be in the local view(s).

jakob223: Maybe we should rename "finish" to "delete this conflict view" or something.

songgao: Yeah, that sounds good to me too

jakob223: Cool, made this change. Thanks for all the discussion 👍

songgao:
I'm realizing some of my feedback here are actually already implemented in my plumbing PR. sigh

PR #17830, opened 2019-06-05
force the video control to believe video source is network

mmaxim:
I'm not totally happy with this, I don't understand how we could ever be showing a blank URL in these components. The react-native-video library thinks any URI that doesn't start with `/https?://` to be an asset bundled with the app, and will straight crash of that is not true. Solution here is if we somehow end up with a blank URL in a native video component, then just put some bogus string in there that at least doesn't send the library down the crasher path.

### DIFF
--- a/shared/chat/conversation/messages/wrapper/unfurl/image/video.native.tsx
+++ b/shared/chat/conversation/messages/wrapper/unfurl/image/video.native.tsx
@@ -21,8 +21,9 @@ export class Video extends React.Component<Props, State> {
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


shared/chat/conversation/messages/wrapper/unfurl/image/video.native.tsx
```diff
@@ -21,8 +21,9 @@ export class Video extends React.Component<Props, State> {
   }
 
   render() {
+    const uri = this.props.url.length > 0 ? this.props.url : 'https://'
     const source = {
```
jakob223: Can you add a comment here explaining what you explained in the PR description?

PR #17832, opened 2019-06-05
Support a value prop in plain-input native

cjb:

Maybe someone who's been in `<PlainInput />` more recently can help figure out what's happened here-- we use `<Kb.NewInput value='foo' />` in a bunch of places, and that gets passed through to `<Kb.PlainInput value='foo' />` and ultimately `<RN.NativeTextInput value='foo' />`, but that native component seems to want `text=` instead of `value=`?  Still trying to figure out when/how this broke.

This PR fixes the Stellar QR code scanner not updating the To: public key after a scan.

### DIFF
--- a/shared/common-adapters/kb-text-input.native.tsx
+++ b/shared/common-adapters/kb-text-input.native.tsx
@@ -63,7 +63,7 @@ if (isAndroid) {
           // @ts-ignore no RN types
           onTextInput={this._onTextInput}
           // @ts-ignore no RN types
-          text={this.props.value || this.props.text || this.props.defaultValue || ''}
+          text={this.props.value || this.props.defaultValue || ''}
           children={children}
           // @ts-ignore no RN types
           disableFullscreenUI={this.props.disableFullscreenUI}
--- a/shared/common-adapters/kb-text-input.native.tsx
+++ b/shared/common-adapters/kb-text-input.native.tsx
@@ -24,8 +24,9 @@ if (isAndroid) {
 
       // @ts-ignore no RN types
       props.style = [this.props.style]
-      props.autoCapitalize =
-        UIManager.AndroidTextInput.Constants.AutoCapitalizationType[props.autoCapitalize || 'sentences']
+      props.autoCapitalize = UIManager.getViewManagerConfig(
+        'AndroidTextInput'
+      ).Constants.AutoCapitalizationType[props.autoCapitalize || 'sentences']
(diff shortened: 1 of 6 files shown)


chrisnojima:
The underlying rn component should take value correctly. We need to look at this more

buoyad:
The `value` prop still works, for example edit-profile prefills all your details using it. This seems like a race somewhere, when I put breakpoints in it sometimes works correctly.

buoyad:
@cjb `send-form/participants/to-field` `ToStellarPublicKey` seems wrong. it copies `props.recipientPublicKey` into state but doesn't keep it updated. This used to make it remount: 
https://github.com/keybase/client/blob/b9123d1258ece42e63742ba33b2ae645f08b22cd/shared/wallets/send-form/participants/container.js#L59-L73
Commented out during nav1 deprecation but this broke with nav2 because `getPath` stopped working (so this is in the release).

shared/common-adapters/kb-text-input.native.tsx
```diff
@@ -24,8 +24,9 @@ if (isAndroid) {
 
       // @ts-ignore no RN types
       props.style = [this.props.style]
-      props.autoCapitalize =
-        UIManager.AndroidTextInput.Constants.AutoCapitalizationType[props.autoCapitalize || 'sentences']
+      props.autoCapitalize = UIManager.getViewManagerConfig(
```
buoyad: Got a yellow box that said we should be using this

cjb:
@buoyad Wow, gnarly! Thanks very much for doing all that debugging.

cjb:
Update: @buoyad's bug looks legit, but it's not the bug I was hitting on Android.  The bug I was hitting is that controlled inputs don't work on Android: they ignore their default value, provided as `value=`.  This is a bug we introduced with the longpress subclass for Android, which was the last thing to go into the 4.0.1 release before it went out.  The bug was a simple typo:
```js
-          text={this.props.text || this.props.text || this.props.defaultValue || ''}
+          text={this.props.value || this.props.text || this.props.defaultValue || ''}
```
We didn't see `text={}` in the RN docs because this isn't an RNTextInput, it's a KBTextInput, based on Android's EditText.

@buoyad, does 4c741b2 look ok?  Tested working on Android.

@MarcoPolo Input fix is in this PR.

shared/common-adapters/kb-text-input.native.tsx
MarcoPolo: nice :+1: I don't think this was supported in 0.58

PR #17834, opened 2019-06-06
TS common/index

MarcoPolo:
@keybase/react-hackers 

Most of this is `as const` or doing HOC typing.

- [ ] Test all platforms


shared/teams/team/header/index.tsx
```diff
@@ -109,7 +109,6 @@ const _TeamHeader = (props: Props) => (
               type="iconfont-chat"
               style={Kb.iconCastPlatformStyles(styles.chatIcon)}
               color={Styles.globalColors.white}
-              size={16}
```
cjb: Just checking this is intentional

MarcoPolo: as far as I can tell, it isn't used. In other places I kept it for now. But I'll make sure to test this and maybe remove the other cases as well

PR #17836, opened 2019-06-06
fix TLF rows on mobile, remove public tag from /keybase/public

jakob223:
fixes an issue found by @mmaxim

### DIFF
--- a/shared/constants/fs.tsx
+++ b/shared/constants/fs.tsx
@@ -1047,9 +1047,9 @@ export const getTlfPath = (path: Types.Path): Types.Path | null => {
 }
 
 export const hasPublicTag = (path: Types.Path): boolean => {
-  return (
-    Types.getPathVisibility(path) === Types.TlfType.Public && path !== Types.stringToPath('/keybase/public')
-  )
+  const publicPrefix = '/keybase/public/'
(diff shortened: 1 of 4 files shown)


shared/constants/fs.tsx
```diff
@@ -1046,6 +1046,12 @@ export const getTlfPath = (path: Types.Path): Types.Path | null => {
   return elems.length > 2 ? Types.pathConcat(Types.pathConcat(defaultPath, elems[1]), elems[2]) : null
 }
 
+export const hasPublicTag = (path: Types.Path): boolean => {
+  return (
+    Types.getPathVisibility(path) === Types.TlfType.Public && path !== Types.stringToPath('/keybase/public')
+  )
```
songgao: This is in constants, so could probably be more efficient, e.g. checking for a prefix and making sure the length is larger than `/keybase/public`'s length?

jakob223:
The names were blank on mobile because the extra `Box2` which contains the `<Content>` and the `<Avatars>` was collapsing to size 0 on mobile for some reason. I didn't fully diagnose why, but I figured simpler was better anyway. They were there, but their parent was 0-width.

PR #17837, opened 2019-06-06
Add proxy support to the go client

ddworken:
Adds proxy support to the go client including the ability to disable cert pinning. The general strategy for the proxy support is setting the `HTTP_PROXY` and `HTTPS_PROXY` environment variables such that golang's http library will automatically configure the proxy. Note that despite the names of these variables, [golang does support socks5 proxies via this method](https://go-review.googlesource.com/c/go/+/35488/). This PR: 

* Adds HTTP Connect proxy support
* Adds socks5 proxy support
* Adds the ability to disable certificate pinning via switching the domain to `keybase.io`
* Refactors the current tor support to reuse the socks5 proxy support and remove the dependency on `h12.me/socks`

Currently, all of this functionality is only exposed through `config.json`, CLI flags, and environment variables. I will open another PR in the future to add a UI component that allows configuring proxies. 

All of the code is tested with unit tests. I tested HTTP proxy support via [squid](https://help.ubuntu.com/lts/serverguide/squid.html.en) and socks5 proxy support via [microsocks](https://github.com/rofl0r/microsocks) (including testing username/password auth for each). I used burp suite as an example of a proxy that does TLS MITM.


go/libcmdline/cmdline.go
```diff
@@ -380,6 +380,16 @@ func (p CommandLine) GetTorProxy() string {
 	return p.GetGString("tor-proxy")
 }
 
+func (p CommandLine) GetProxyType() string {
+	return p.GetGString("proxy-type")
+}
+
+func (p CommandLine) IsSSLPinningEnabled() bool {
```
maxtaco: Maybe "TLS" rather than "SSL"?  Or maybe "Cert Pinning"?

go/libcmdline/cmdline.go
```diff
@@ -543,6 +553,11 @@ func (p *CommandLine) PopulateApp(addHelp bool, extraFlags []cli.Flag) {
 			Name:  "debug, d",
 			Usage: "Enable debugging mode.",
 		},
+		cli.BoolFlag{
+			Name:  "disable-ssl-pinning",
+			Usage: "Disable SSL pinning within the app. WARNING: This reduces the security of the app. Do not use " +
```
maxtaco: Maybe say when it should be used?  As in behind an invasive proxy?

go/libcmdline/cmdline.go
```diff
@@ -613,7 +628,23 @@ func (p *CommandLine) PopulateApp(addHelp bool, extraFlags []cli.Flag) {
...
 			Name:  "proxy",
-			Usage: "Specify an HTTP(s) proxy to ship all Web requests over.",
+			Usage: "Specify a proxy to ship all Web requests over; Must be used with --proxy-type; Example: localhost:8080",
+		},
+		cli.StringFlag{
+			Name:  "proxy-type",
+			Usage: fmt.Sprintf("set the proxy type; One of: %s", func() string {
+				// This function retrieves the list of valid proxy types from the dict that tracks them
+				proxyTypes := ""
+				for k := range libkb.ProxyTypeStrToEnum {
```
maxtaco: I'd make a helper here, so as not to make this thing even bigger.

Also, maybe pop the keys out into a slice and then call `strings.Join`, it might be slightly cleaner.

go/libkb/client.go
```diff
@@ -191,20 +191,14 @@ func NewClient(g *GlobalContext, config *ClientConfig, needCookie bool) *Client
...
-				return dialSocksProxy(network, addr)
-			}
-		} else {
-			xprt.Proxy = http.ProxyFromEnvironment
-		}
+	if config != nil && config.RootCAs != nil {
+		xprt.TLSClientConfig = &tls.Config{RootCAs: config.RootCAs}
+	}
+
+	err := EnableProxy(env.GetProxyType(), env.GetProxy())
```
maxtaco: can we achieve the same effect by just fiddling with `xprt` above? It feels kludgy to change this local behavior via global variables. I think from the docs we can

https://golang.org/src/net/http/transport.go?s=3607:10104#L93

go/libkb/client.go
```diff
@@ -191,20 +191,14 @@ func NewClient(g *GlobalContext, config *ClientConfig, needCookie bool) *Client
...
+	if err != nil {
+		// We failed to set an environment variable, something is very wrong
+		panic(err)
```
maxtaco: This isn't great, we very rarely, if ever, do this sort of thing.

go/libkb/constants.go
```diff
@@ -41,12 +41,26 @@ const (
 	NoRunMode         RunMode = kbconst.NoRunMode
 )
 
-var ServerLookup = map[RunMode]string{
-	DevelRunMode:      DevelServerURI,
-	StagingRunMode:    StagingServerURI,
-	ProductionRunMode: ProductionServerURI,
+func ServerLookup(env *Env, mode RunMode) (string, error) {
```
maxtaco: Can we move this function out of constants?  Trying to keep it "constants" only.

go/libkb/env.go
```diff
@@ -541,14 +543,28 @@ func (e *Env) GetServerURI() string {
...
+		return server
 	}
 
 	return e.GetString(
 		func() string { return e.cmd.GetServerURI() },
 		func() string { return os.Getenv("KEYBASE_SERVER_URI") },
 		func() string { return e.GetConfig().GetServerURI() },
-		func() string { return ServerLookup[e.GetRunMode()] },
+		func() string {
+			server, e := ServerLookup(e, e.GetRunMode())
```
maxtaco: Call your error `err` since that's the convention

go/libkb/env.go
```diff
@@ -541,14 +543,28 @@ func (e *Env) GetServerURI() string {
...
+			if e != nil {
+				// ServerLookup only returns an error if the RunMode is bogus. Panic since there is no
+				// way to get a URL in this case
+				panic(e)
```
maxtaco: Again, I'm not crazy about this. I guess previously the program would crash or behave weirdly downstream anyways (maybe by not having a URL prefix). The right answer is to propagate the error all the way up, but that might be ugly in this case.

maxtaco: Eh, it would be horrible. Have `GetServerURI` return `(string, error)`?  About 15 places to fix.

go/libkb/proxy.go
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+package libkb
+
+import "os"
+
+// Represents the different types of supported proxies
+type ProxyType int
+const (
+	No_Proxy ProxyType = iota
```
maxtaco: It's not "Go"-y to use underscores, can you just camel case these instead?

ddworken: Done.

go/libkb/env.go
ddworken: Done

go/libkb/constants.go
ddworken: Done. Moved to `client.go` since that seems like a reasonable place for it.

go/libkb/client.go
ddworken: Yes, it is possible to change it via setting the Proxy function. I initially did it this way because [`s3.go`](https://github.com/keybase/client/blob/master/go/chat/s3/s3.go#L953) and [`search.go`](https://github.com/keybase/client/blob/master/go/chat/giphy/search.go#L114) currently do not use the http client created in `client.go`. By setting it in the environment variables, those http.Clients (and any others that are made in the future) will automatically get configured to use the proxy. 

I could either refactor `s3.go` and `search.go` to use `client.go` as the HTTP client or I could simply modify them to use the proxy directly. What are your thoughts on the best solution here? I'm leaning towards migrating them to reference `client.go` but I'm not sure if there is a reason they aren't using it already.

go/libcmdline/cmdline.go
ddworken: Done

go/libcmdline/cmdline.go
ddworken: Done

go/libcmdline/cmdline.go
ddworken: Changed to cert pinning

go/libkb/client.go
ddworken: Done. I refactored `NewClient` to return an error in order to get rid of the panic

go/libkb/client.go
maxtaco: I wouldn't refactor, I would just modify the other two instances individually. There is likely some good reason they weren't using `client.go` to begin with. Cc @mmaxim  and @patrickxb  on why not.

patrickxb: chat/s3/s3.go is for chat attachment uploads and uses its own http client.  We heavily modified that package for direct to s3 uploads.

I would prefer if it was modified as little as possible as it has been working well for years.

go/libkb/env.go
ddworken: Done.

mmaxim:
Please let me look at this before merge

go/libkb/client.go
ddworken: @patrickxb  What do you think of this solution? I attempted to keep the changes to s3 as minimal as possible but it required some decent changes to wind the global context in in order to get access to the config.

go/chat/s3/s3.go
```diff
@@ -133,7 +136,7 @@ var DefaultAttemptStrategy = AttemptStrategy{
 }
 
 // New creates a new S3.  Optional client argument allows for custom http.clients to be used.
-func New(signer Signer, region Region, client ...*http.Client) *S3 {
+func New(signer Signer, region Region, env *libkb.Env, client ...*http.Client) *S3 {
```
maxtaco: Why not `libkb.GlobalContext` instead?

maxtaco: BTW, make sure you check out this doc: https://github.com/keybase/client/blob/master/go/doc/context.md

go/libkb/api_test.go
```diff
@@ -42,8 +42,13 @@ func TestProductionCA(t *testing.T) {
 	t.Log("WARNING: setting run mode to production, be careful:")
 	tc.G.Env.Test.UseProductionRunMode = true
 
-	if tc.G.Env.GetServerURI() != uriExpected {
-		t.Fatalf("production server uri: %s, expected %s", tc.G.Env.GetServerURI(), uriExpected)
+	serverURI, err := tc.G.Env.GetServerURI()
```
maxtaco: if possible, we prefer the `require` version since it's more succinct.

go/chat/s3/s3.go
patrickxb: I agree I think `libkb.GlobalContext` would be better than `libkb.Env`.

ddworken: Changed to `libkb.GlobalContext`. I winded `Env` through there since it required slightly fewer changes in order to do, but :+1: if we're okay with slightly more changes to the s3 code.

go/libkb/api_test.go
ddworken: Done.

go/chat/sender.go
```diff
@@ -41,7 +41,7 @@ func NewBlockingSender(g *globals.Context, boxer *Boxer, getRi func() chat1.Remo
 		DebugLabeler:      utils.NewDebugLabeler(g.GetLog(), "BlockingSender", false),
 		getRi:             getRi,
 		boxer:             boxer,
-		store:             attachments.NewS3Store(g.GetLog(), g.GetEnv(), g.GetRuntimeDir()),
+		store:             attachments.NewS3Store(g.GlobalContext, g.GetLog(), g.GetEnv(), g.GetRuntimeDir()),
```
maxtaco: Passing `g.GlobalContext` might undo some separation that @patrickxb had in mind, so there might be a slightly better way, not sure though.

patrickxb: It's fine if it's necessary.

go/chat/attachments/store.go
```diff
@@ -104,8 +105,9 @@ type S3Store struct {
 
 // NewS3Store creates a standard Store that uses a real
 // S3 connection.
-func NewS3Store(logger logger.Logger, env *libkb.Env, runtimeDir string) *S3Store {
+func NewS3Store(g *libkb.GlobalContext, logger logger.Logger, env *libkb.Env, runtimeDir string) *S3Store {
```
mmaxim: We can drop passing in logger and env as long as `GlobalContext` is now coming in here

go/chat/unfurl/scraper.go
```diff
@@ -15,13 +16,15 @@ type Scraper struct {
 	utils.DebugLabeler
 	cache      *unfurlCache
 	giphyProxy bool
+	libkb.Contextified
 }
 
-func NewScraper(logger logger.Logger) *Scraper {
+func NewScraper(g *libkb.GlobalContext, logger logger.Logger) *Scraper {
```
mmaxim: Same comment here, you can use the logger on `G` instead of the one that gets passed in,

go/chat/attachments/store.go
ddworken: Done.

go/chat/unfurl/scraper.go
ddworken: Done.

go/chat/attachments/store.go
```diff
@@ -88,7 +88,6 @@ type S3Store struct {
 	utils.DebugLabeler
 
 	g        *libkb.GlobalContext
```
mmaxim: Sorry to keep bothering you with this stuff that isn't really the focus of this PR, but typically what happens here is we use the `libkb.Contextified` struct to provide a function called `G()` to the struct that includes it. If you grep around for that, there are some examples.

ddworken: Done.

mmaxim:
Pending CI

PR #17847, opened 2019-06-07
reset with eldest key reuse detection in identify

pzduniak:


maxtaco:
I shouldn't be the reviewer but I took a quick look and this looks like definitely the right idea.

zapu:
Changes for the test:
```
diff --git a/go/engine/identify2_test.go b/go/engine/identify2_test.go
index 342b31dca5..17a3ce9f0a 100644
--- a/go/engine/identify2_test.go
+++ b/go/engine/identify2_test.go
@@ -1256,6 +1256,7 @@ func TestTrackResetReuseKey(t *testing.T) {
 		})),
 	)
 	require.False(t, idUI.BrokenTracking)
+	require.Empty(t, idUI.DisplayKeyDiffs)
 
 	// Bob tracks Alice
 	trackUser(tcY, fuY, fuX.NormalizedUsername(), libkb.GetDefaultSigVersion(tcX.G))
@@ -1301,12 +1302,14 @@ func TestTrackResetReuseKey(t *testing.T) {
 	require.Equal(t, "1 followed proof failed", err.(libkb.IdentifySummaryError).Problems()[0])
 	require.Len(t, idUI.DisplayKeyDiffs, 1, "key diffs count")
 	require.Equal(t, keybase1.TrackDiffType_NEW_ELDEST, idUI.DisplayKeyDiffs[0].Type, "key diff new eldest")
+	require.False(t, idUI.BrokenTracking) // tracking is not "broken" for this user - it's a key change
 
 	// He should be able to retrack
 	trackUser(tcY, fuY, fuX.NormalizedUsername(), libkb.GetDefaultSigVersion(tcX.G))
 	assertTracking(tcY, fuX.Username)
 
 	// Which should fix the identification
+	idUI = &FakeIdentifyUI{}
 	require.NoError(t, RunEngine2(
 		NewMetaContextForTest(tcY).WithUIs(libkb.UIs{
 			LogUI:      tcY.G.UI.GetLogUI(),
@@ -1319,6 +1322,7 @@ func TestTrackResetReuseKey(t *testing.T) {
 		})),
 	)
 	require.False(t, idUI.BrokenTracking)
+	require.Empty(t, idUI.DisplayKeyDiffs)
 }
 
 var aliceUID = keybase1.UID("295a7eea607af32040647123732bc819")
```

`BrokenTracking` is false for every id run in this case, so it's now also checking `DisplayKeyDiffs`

go/libkb/identify_state.go
```diff
@@ -181,7 +181,8 @@ func (s *IdentifyState) computeKeyDiffs(dhook func(keybase1.IdentifyKey) error)
 	observedEldest := s.u.GetEldestKID()
 	if s.track != nil {
 		trackedEldest := s.track.GetEldestKID()
-		if observedEldest.NotEqual(trackedEldest) {
+		if observedEldest.NotEqual(trackedEldest) ||
+			s.u.GetCurrentEldestSeqno() > s.track.GetTrackedLinkSeqno() {
```
zapu: @patrickxb quick look at a change in how resets in tracked users are detected?

zapu: The assumption is that if someone's eldest_seqno is higher than the seqno of tail link we tracked (embedded in our "track" link), it means they must have reset.

go/engine/device_keygen.go
```diff
@@ -211,6 +215,9 @@ func (e *DeviceKeygen) setup(m libkb.MetaContext) {
 	}
```
zapu: let's add something like
```
if m.G().Env.GetRunMode() != libkb.DevelRunMode && (e.args.naclSigningKeyPair != nil || e.args.naclEncryptionKeyPair != nil) {
    e.runErr = errors.Make("Trying to pass key pair agument to device keygen")
    return 
}
```
in case someone misunderstands and tries to utilize this in prod

go/libkb/identify_state.go
patrickxb: Maybe things have changed, but I thought when a user resets the eldest_seqno goes to 0.  So wouldn't the tracked seqno be higher than the current user seqno?

zapu:
Youre right. It goes to zero before reprovisioning. We should handle this case and also add another test

go/libkb/identify_state.go
pzduniak: `GetCurrentEldestSeqno` returns the seqno that starts the current subchain, so there's no way to figure out if someone was reset but hasn't provisioned yet. That being said for all intents and purposes the account is useless until you provision _something_ and as soon as you're done with that the change will be noticed by this code.

zapu: please hold for a bit, i'll take a look as well. seems to be possible to detect that we are loading in-between-reset user, but i don't know off the bat

pzduniak: I really don't understand the significance of detecting whether account is in a temporary unusable state, as far as I understand the whole model it's not a vulnerability at all.

zapu: I tried this today. When an account is in eldest=0 state, observedEldest is "", so user gets something like 
```
✖ <Account reset! Old key was 0120bd9a998f63961caeffe5e033f413e46bb90e855ae13a96323feb6dbfa1a71cca0a; new key is >
```
in this case.

This is good to go IMO

PR #17848, opened 2019-06-07
Have links to the info panel take you to the right tab

buoyad:
Changes links in chat to "manage notifications" and "manage retention" take you to the settings tab of the info panel. r? @keybase/react-hackers


mmaxim:
I'm glad this was that easy to fix, thanks!

PR #17854, opened 2019-06-07
Add block/unblock action to profile

mlsteele:


shared/profile/block/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+          <Kb.Box style={styles.errorBanner}>
+            <Kb.Text center={!Styles.isMobile} style={styles.errorBannerText} type="BodySemibold">
+              {props.errorMessage}
+            </Kb.Text>
+          </Kb.Box>
+        )}
+        <Kb.Box style={styles.contentContainer}>
+          <Kb.Text center={!Styles.isMobile} style={styles.descriptionText} type="Header">
+            Really block {props.username}?
+          </Kb.Text>
```
mlsteele: Won't merge until design is happy with whatever text ends up here

go/service/user.go
```diff
@@ -623,3 +623,33 @@ func (h *UserHandler) UserCard(ctx context.Context, arg keybase1.UserCardArg) (r
 	}
 	return libkb.UserCard(mctx, uid, arg.UseSession)
 }
+
+func (h *UserHandler) BlockUser(ctx context.Context, username string) (err error) {
+	mctx := libkb.NewMetaContext(ctx, h.G())
+	defer mctx.TraceTimed("UserHandler#BlockUser", func() error { return err })()
+	return h.blockUserHelper(mctx, username, false)
```
xgess: this is a confusing API. you call the `block` thing and pass it `false` to do the blocking and `true` to undo it. '__'  it looks like it's because `user/block` accepts a parameter for `unblock` which seems like the thing that's actually weird. i'd propose keeping that as isolated as possible inside the `blockUserHelper` and changing `unblock` on this method to something like `wantBlocked`.

mlsteele: It's not too late to change the server side. What's the least confusing combo?

xgess:
there's a snapshot error in the tests.

go/service/user.go
xgess: i'm a believer in minimizing the number of `negative`s in parameters, because there will always be a reason to add a negative on top of something later, and triple negatives should be avoided. if i had my druthers, i think i'd try to make this read more declarative. like, alice wants bob to be blocked, or alice wants bob not to be blocked. so either a desired state (enum: "blocked" / "not blocked") or a `true` is for `blocked` and `false` is for `not blocked`. not sure the juice is worth the squeeze on going back to the server. it's more important to me that whatever is weird is totally isolated in one function.

mlsteele:
Good point. It lost its rounded corners somehow. I'll merge this to get minimize conflict risk and then add `error?: string` in another PR.

shared/profile/block/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+          <Kb.Text center={!Styles.isMobile} style={styles.errorBannerText} type="BodySemibold">
+            {props.errorMessage}
+          </Kb.Text>
+        </Kb.Box>
+      )
+    }
+    const modalProps = {
+      confirmText: 'Yes, block them',
+      description:
+        'This will hide them from your followers and suggestions, and prevent them from creating new conversations or teams with you. Note that they may be able to find out that you block them.',
```
cjb: Maybe "that you blocked them"?  Your call.

mlsteele: I tried that. But 'them' wrapped onto its own line and looked silly. Which made @cecileboucheron look like a magician. What do you think?
(I'm merging but still listening)

cecileboucheron:
I think the `?` is missing in the title, and the banner should stick to the top of the modal window, like here, on desktop. 
image

PR #17858, opened 2019-06-07
fix security popup for first sfmi enable on macOS

songgao:
r @keybase/react-hackers / cc @keybase/hotpotatosquad

### DIFF
--- a/shared/fs/banner/system-file-manager-integration-banner/kext-permission-popup-container.tsx
+++ b/shared/fs/banner/system-file-manager-integration-banner/kext-permission-popup-container.tsx
@@ -1,6 +1,7 @@
 import {namedConnect, RouteProps} from '../../../util/container'
 import * as FsGen from '../../../actions/fs-gen'
 import KextPermissionPopup from './kext-permission-popup'
+import * as RouteTreeGen from '../../../actions/route-tree-gen'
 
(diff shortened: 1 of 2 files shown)


shared/fs/banner/system-file-manager-integration-banner/kext-permission-popup-container.tsx
```diff
@@ -1,6 +1,7 @@
 import {namedConnect, RouteProps} from '../../../util/container'
 import * as FsGen from '../../../actions/fs-gen'
 import KextPermissionPopup from './kext-permission-popup'
+import * as RouteTreeGen from '../../../actions/route-tree-gen'
 
 type OwnProps = RouteProps<{}, {}>
```
jzila: You can get rid of `OwnProps` now, right?

jzila:
That's fine, but at the least it's not longer a RouteProps, right?

chrisnojima:
Correct. In the upcoming types connect pr we'll use ownprops (like before) so best to keep it named

songgao:
It's from the router, so if typed I assume it'd still need to be RouteProps?

songgao:
Anyway I'm merging this for now. It's an easy fix either way

PR #17859, opened 2019-06-07
data: implement a per-node string obfuscator

strib:
This obfuscator takes in a secret, and obfuscates a given plaintext
string this way:

* `strHash` = `sha256-hmac(secret, plaintext)`
* Use the first 3 bytes of `strHash` to index into the secwords (bip-0039) list in libkb/secwords.go, getting two words.
* Join those two words with a dash, to get `obfuscatedStr`.
* If this collides with a previously-obfuscated string created by this same node, a numeric suffix like "-2" is appended to the obfuscated string.

In a future PR, there will be one `NodeObfuscator` for each directory's `libkbfs.Node`, for obfuscating entry names within that directory.

cc: @keybase/hotpotatosquad 

Issue: HOTPOT-55

### DIFF
--- a/go/kbfs/data/node_obfuscator.go
+++ b/go/kbfs/data/node_obfuscator.go
@@ -1,0 +1,145 @@
+// Copyright 2019 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package data
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"fmt"
+	"math"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/keybase/client/go/libkb"
+)
+
+const (
+	separator = "-"
+)
+
+// NodeObfuscatorSecret is a byte slice wrapper for a secret that can
+// be passed to `NodeObfuscator`.  It overrides `String` so there's no
+// chance of the secret being accidentally logged.
+type NodeObfuscatorSecret []byte
+
+// String implements the `fmt.Stringer` interface for NodeObfuscatorSecret.
+func (nos NodeObfuscatorSecret) String() string {
+	return fmt.Sprintf("{private %d bytes}", len(nos))
+}
+
+// NodeObfuscator takes a secret, and uses it to create obfuscate
+// strings based on BIP-0039 dictionary words.  It remembers previous
+// obfuscations, and if there happens to be a collision on the
+// obfuscated text for a new plaintext string, it appends a numeric
+// suffix to the new obfuscation.
+type NodeObfuscator struct {
+	secret NodeObfuscatorSecret
+
+	lock     sync.RWMutex
+	obsCache map[string]string
+	usedObs  map[string]bool
+}
+
+var _ Obfuscator = (*NodeObfuscator)(nil)
+
+// NewNodeObfuscator creates a new `NodeObfuscator` instance.
+func NewNodeObfuscator(secret NodeObfuscatorSecret) *NodeObfuscator {
+	return &NodeObfuscator{
+		secret: secret,
+		// Don't initialize caches yet, for memory reasons, in case
+		// they're never used.
+	}
+}
+
+func (no *NodeObfuscator) checkCacheIfExistsLocked(
+	plaintext string) (string, bool) {
+	if no.obsCache == nil {
+		return "", false
+	}
+
+	obs, ok := no.obsCache[plaintext]
+	return obs, ok
+}
+
+func (no *NodeObfuscator) checkCacheIfExists(plaintext string) (string, bool) {
+	no.lock.RLock()
+	defer no.lock.RUnlock()
+	return no.checkCacheIfExistsLocked(plaintext)
+}
+
+func (no *NodeObfuscator) obfuscateWithHasher(
+	plaintext string, hasher func(string) []byte) string {
+	obs, ok := no.checkCacheIfExists(plaintext)
+	if ok {
+		return obs
+	}
+
+	no.lock.Lock()
+	defer no.lock.Unlock()
+	// See if it's been cached since we last released the lock.
+	obs, ok = no.checkCacheIfExistsLocked(plaintext)
+	if ok {
+		return obs
+	}
+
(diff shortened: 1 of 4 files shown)


go/kbfs/data/node_obfuscator.go
```diff
@@ -0,0 +1,145 @@
...
+		obs = strings.Join(
+			[]string{firstWord, secondWord, strconv.Itoa(suffix)}, separator)
+	}
+
+	no.obsCache[plaintext] = obs
+	no.usedObs[obs] = true
+	return obs
+}
+
+func (no *NodeObfuscator) defaultHash(plaintext string) []byte {
```
strib: @maxtaco, @mlsteele: this function will be used to generated hashes for obfuscating a filename in the KBFS logs.  The `secret` here will be a sha256 hash of the TLF's 1st key generation, and `plaintext` will be the name of the file or directory being obfuscated.  Then, the first 3 bytes of the resulting hmac returned by this function will be used to index into the bip-0039 wordlist to get 2 words that can replace the plaintext word.  (Yes, there will be collisions, and that's fine.)  This is an expanded version of an idea @mlsteele wrote up in KBFS-962.

I'd like a sanity check on this to make sure it won't leak any sensitive data, either about the TLF's key (crucial), or the `plaintext` being obfuscated (ideal, but less crucial).  Thanks!

go/kbfs/data/node_obfuscator_test.go
```diff
@@ -0,0 +1,66 @@
...
+	second := libkb.SecWord(1)
+	expectedObs := strings.Join([]string{first, second}, separator)
+	require.Equal(t, expectedObs, obs)
+
+	t.Log("Obfuscate a new plaintext string with the same hash")
+	obs2 := no.obfuscateWithHasher("test2", hasher1)
+	expectedObs2 := strings.Join([]string{first, second, "2"}, separator)
+	require.Equal(t, expectedObs2, obs2)
+
+	t.Log("And one more, with a different hash but the same first 3 bytes")
```
jakob223: s/`3 bytes`/`22 bits`?

go/kbfs/data/node_obfuscator.go
jakob223: I can't think of any way this can be used to leak data either.

mlsteele: A unique derivation string could be used instead of bare sha256(key) to decrease the chances that some other protocol uses sha256(key) for something else

strib: @mlsteele that would be nice, but it needs to be something that's consistent and deterministically generated between devices.  Would you suggest making another key type in the implicit team sigchain for the TLF?

mlsteele: I'm not suggesting using additional team mechanisms.

[Here](https://github.com/keybase/client/blob/333ded3a693b5acff16faf393dc6e5df0091bd70/go/stellar/note.go#L37) in stellar for example [DeriveSymmetricKey](https://github.com/keybase/client/blob/333ded3a693b5acff16faf393dc6e5df0091bd70/go/libkb/naclwrap.go#L636) is used.

strib: @mlsteele I see, thanks!  So basically hmac the key (or the hash of the key) with some hardcoded, app-wide-unique constant.  Good idea, will do that when I code that bit, thanks!

maxtaco: yeah i agree with @mlsteele here. i think the approach is fine but if this is an important key that can be used in other contexts, i would recommend a prefix that won't be used in that other context.

go/kbfs/data/node_obfuscator.go
```diff
@@ -0,0 +1,145 @@
...
+	// Look up two words based on the first three bytes of the mac.
+	// (Each word takes 11 bits to lookup a unique word among the 2048
+	// secret words.)
+	first := int(buf[0])
+	// First 8 bits are from buf[0].
+	first <<= 3
+	// Last 3 bits are the first 3 bits of buf[1].
+	const first3 = 0x7 << 5
+	first |= ((int(buf[1]) & first3) >> 5)
+	if float64(first) >= math.Exp2(11) {
```
jzila: Does `math.Exp2(11)` get optimized into a constant by the compiler? If not, then probably best to just use `2048`.

go/kbfs/data/node_obfuscator.go
```diff
@@ -0,0 +1,145 @@
...
+	// First 5 bits are the last 5 bits of buf[1].
+	const last5 = 0x1f
+	second := int(buf[1]) & last5
+	second <<= 6
+	// Last 6 bits are the first 6 bits of buf[2].
+	const first6 = 0x3f << 2
+	second |= ((int(buf[2]) & first6) >> 2)
+	if float64(second) >= math.Exp2(11) {
+		panic(fmt.Sprintf("Generated %d from 11 bits", second))
+	}
```
jzila: I wonder if it'd be cleaner to pull the first 3 bytes into a `uint32`, then shift out of it twice 11 bits at a time to get the words. That way it's the same algorithm twice instead of this complexity.

go/kbfs/data/node_obfuscator.go
strib: @maxtaco 	yep, already done with a unique prefix.

go/kbfs/data/node_obfuscator.go
strib: I don't know how to know that.  I just wanted the number of bits to be clear when reading the code.  What about:

```go
const max = math.Exp2(11)
if float64(first) >= max {
...
```

?  Seems like that would definitely be a one-time constant.

jzila: `const initializer math.Exp2(11) is not a constant`

jzila: @strib can you just do `1 << 11`?

go/kbfs/data/node_obfuscator.go
strib: Good suggestion, done in #18085.  And that obviates the above comment.

PR #17862, opened 2019-06-10
display file attachments with the filename if they have captions

mmaxim:
cc @keybase/hotpotatosquad


mmaxim:
I'd like to get this in the release as well.

PR #17863, opened 2019-06-10
run_keybase: no squirrel when no colorable tty

heronhaye:
Addresses https://github.com/keybase/keybase-issues/issues/3423.

### DIFF
--- a/packaging/linux/run_keybase
+++ b/packaging/linux/run_keybase
@@ -143,7 +143,9 @@ warn_if_weird_path() {
 }
 
 show_cryptosquirrel() {
-  [ "${KEYBASE_NO_SQUIRREL:-}" != "1" ] && cat /opt/keybase/crypto_squirrel.txt
+  if tput colors &> /dev/null; then
+      [ "${KEYBASE_NO_SQUIRREL:-}" != "1" ] && cat /opt/keybase/crypto_squirrel.txt
+  fi
 }
 
 warn_if_exists_and_unwritable() {
--- a/packaging/linux/run_keybase
+++ b/packaging/linux/run_keybase
@@ -143,7 +143,7 @@ warn_if_weird_path() {
 }
 
 show_cryptosquirrel() {
-  if tput colors &> /dev/null; then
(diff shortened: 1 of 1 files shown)


packaging/linux/run_keybase
```diff
@@ -143,7 +143,9 @@ warn_if_weird_path() {
 }
 
 show_cryptosquirrel() {
-  [ "${KEYBASE_NO_SQUIRREL:-}" != "1" ] && cat /opt/keybase/crypto_squirrel.txt
+  if tput colors &> /dev/null; then
```
strib: I'm not sure about this.  On my machine `man tput` says this unde rthe "EXIT CODES" section:

>           integer
>                 a  value  of 0 is always set, whether or not capname is defined for this terminal type.  To determine if capname
>                 is defined for this terminal type, the user must test the value written to standard output.  A value of -1 means
>                 that capname is not defined for this terminal type.

So it seems like `tput colors` should always have a 0 exit code, and thus always pass this `if` check.  We'd have to check the actual printed integer.

But maybe you tested this and I'm misinterpreting this?  I don't know a good way to test it.

heronhaye: You're right, I think it was only working because it wasn't run in a tty

```
Jun 12 10:09:03 gyre-circling-spirits systemd[557]: Started /usr/bin/tput colors.
Jun 12 10:09:03 gyre-circling-spirits tput[14849]: tput: No value for $TERM and no -T specified
Jun 12 10:09:03 gyre-circling-spirits systemd[557]: run-r86ba475491a848a98ce3adb962680f9a.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Jun 12 10:09:03 gyre-circling-spirits systemd[557]: run-r86ba475491a848a98ce3adb962680f9a.service: Failed with result 'exit-code'.
```

I fixed it to check for error code and greater than 2 (white and black). That being said it doesn't seem to work too reliably, my emacs shell says "-1" although colors seem to be supported anyway.

PR #17873, opened 2019-06-10
Store pwhash and eddsa when signing up fails to allow for signup recovery

zapu:
Store PWHash and EdDSA (generated from stretched random passphrase) in secret store after failed provisioning during NOPW signup. In `login_provision.go`, this will allow us to log user back in to provision their first device. Otherwise, they would have been stuck because they wouldn't know their password.

Passphrase stream parts are stored in two secret store entries, pretending to be LKSecs - pwhash and eddsa are 32-byte each so they match format used by secret store implementation. Identifiers under which they are stored are usernames with `.tmp_eddsa` or `.tmp_pwhash` suffixes.

These secrets are only stored during failed NOPW signup - where SignupJoin works but DeviceWrap doesn't because of network issues or similar. EdDSA and PWhash secrets are also cleared after successful provisioning.

We are expecting the following use case:
1) User tries to signup, because they are on bad connection, `SignupJoin` worked but provisioning didn't.
  a) They are now signed up with no devices, and they don't know their password.
2) User tries to log in using their chosen username, `recoverAfterFailedSignup` logs them in using stored eddsa/pwhash and allows them to provision.
3) eddsa/pwhash secrets are cleared.


go/engine/signup.go
```diff
@@ -321,6 +338,23 @@ func (s *SignupEngine) storeSecret(m libkb.MetaContext, randomPw bool) {
...
+func (s *SignupEngine) storeSecretForRecovery(m libkb.MetaContext) (err error) {
+	defer m.Trace("SignupEngine#storeSecretForRecovery", func() error { return err })()
+
+	if !s.arg.GenerateRandomPassphrase {
+		m.Debug("Not GenerateRandomPassphrase - skipping storeSecretForRecovery")
+		return nil
+	}
+
+	username := s.me.GetNormalizedName()
+	err = libkb.StoreFullPassphraseStream(m, username, s.ppStream)
```
maxtaco: is it the full passphrase stream or just a bigger subslice of it?

go/libkb/passphrase_stream.go
```diff
@@ -87,6 +87,23 @@ func (ps *PassphraseStream) SetGeneration(gen PassphraseGeneration) {
 	ps.gen = gen
 }
 
+func newPassphraseStreamPE(pwhash, eddsa []byte) (*PassphraseStream, error) {
```
maxtaco: confusing abbreviation, maybe it's better for a longer name.

Can you make a type here rather than `[]byte`. I'm remember bug 3964 which was so much effort to recover from. That was a result of using `[]byte`s in the login path rather than the specialized `LKS...` types.

zapu: do you mind if I change typing in another PR? I'd like to introduce more types to `passphrase_stream.go` but this would require additional changes to consumers of that API.

or is there some specific small change I should do here right now?

maxtaco: don't follow. can't you change this signature and then switch to bytes one layer down?

zapu: so we would have something like:
```
type PWHash []byte
```
and in `RetrievePwhashEddsaPassphraseStream`:
```
pwHashSecret, err := ss.RetrieveSecret(mctx, formatPPSSecretStoreIdentifier(username, ssPwhashSuffix))
pps, err := newPassphraseStreamFromPwhAndEddsa(PWHash(pwHashSecret.Bytes()), EdDSA(edDSASecret.Bytes())
```
I thought that does not buy us much

maxtaco: what you have, with two `[]byte` params, is just really dangerous. Get that slightly wrong, either by transposition, or the wrong type of data, and you'll have a very difficult to diagnose, debug and fix bug on your hand. See Bug 3964

maxtaco: thanks!

PR #17874, opened 2019-06-10
favorites: fix shutdown race in two ways

strib:
1. When a favorites operation triggers with a long timeout (like the background favorites refresh done on login), the main favorites loop can get blocked on it, and then when it finishes, it will check for buffered, timer-based requests (once every 5 seconds).  If it finds the buffered channel is closed (because `Shutdown` has already been called) then it returned immediately, without letting any of the requests buffered in the other channel complete, leading to a hang in the `Shutdown` code while waiting on the request group.

2. The above race happens in git because `Shutdown` is called on the critical path.  But there's no reason for git code to be refreshing favorites in the background, because it doesn't even have a proper favorites DB since a fresh directory is used for every operation.  So don't bother refreshing the favorites when in that mode.

Issue: HOTPOT-73

### DIFF
--- a/go/kbfs/libkbfs/interfaces.go
+++ b/go/kbfs/libkbfs/interfaces.go
@@ -1903,9 +1903,9 @@ type InitMode interface {
 	// OldStorageRootCleaningEnabled indicates whether we should clean
 	// old temporary storage root directories.
 	OldStorageRootCleaningEnabled() bool
-	// DoRefreshCachedFavorites indicates whether we should refresh
+	// DoRefreshFavoritesOnInit indicates whether we should refresh
 	// our cached versions of the favorites immediately upon a login.
-	DoRefreshCachedFavorites() bool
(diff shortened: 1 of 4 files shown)


go/kbfs/libkbfs/interfaces.go
```diff
@@ -1903,6 +1903,9 @@ type InitMode interface {
 	// OldStorageRootCleaningEnabled indicates whether we should clean
 	// old temporary storage root directories.
 	OldStorageRootCleaningEnabled() bool
+	// DoRefreshCachedFavorites indicates whether we should refresh
+	// our cached versions of the favorites immediately upon a login.
+	DoRefreshCachedFavorites() bool
```
jakob223: Can we call this something that sounds more like initialization? e.g. `DoRefreshFavoritesOnInit`?

PR #17876, opened 2019-06-10
fix red bar identify warnings in chat

mmaxim:
Do a couple things:

1. Remove background identifier code, it was not being used, and is not needed anymore.
2. Just run `Identify` directly instead of `LookupID` which does not actually do an identify anymore (this is why the red bar didn't clear after accepting a broken proof). The JS does not care about anything other than the users it gets from `CanonicalName` and `Breaks`, `TLFID` is unused so don't bother with it.

cc @keybase/hotpotatosquad


chrisnojima:
- [ ] lets do this after the release

PR #17879, opened 2019-06-10
Add a GUI to control the configured proxy

ddworken:
Blocked on #17837

 Adds a GUI to configure proxy support in Settings --> Advanced.

image
image

The typescript frontend interacts with the golang service over RPC. Refactored `MakeProxy` to take in an `Env` so that it will detect proxy changes without a reboot. Refactored `ProxyType` to make the enum members public.


shared/settings/advanced/index.tsx
```diff
@@ -54,43 +55,83 @@ const UseNativeFrame = (props: Props) => {
...
+    this.setState({disableCertPinning, showCertPinningModal: true})
+  }
+
+  render() {
+    const disabled =
+      this.props.lockdownModeEnabled == null || this.props.hasRandomPW || this.props.settingLockdownMode
+    return (
+      <Kb.Box style={styles.advancedContainer}>
+        {this.state.showCertPinningModal && (
+          <Kb.Box style={{zIndex: 1}}>
```
cjb: Instead of zIndex, dispatch an action that navs to a modal route.

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +249,153 @@ class Developer extends React.Component<Props, State> {
...
+
+    this.state = {
+      address: '',
+      certPinning: true,
+      port: '',
+      proxyType: ProxyTypeToDisplayName['noProxy'],
+      showCertPinningConfirmationModal: false,
+    }
+
+    RPCTypes.configGetProxyDataRpcPromise()
```
cjb: We don't dispatch RPCs from the display components anywhere else: the usual pattern is:

* accept an `onLoadProxyData` from the container component that calls something like `dispatch(ConfigGen.createLoadProxyData())`

* have a saga listening for that action that perform the RPC and then calls `ConfigGen.createProxyDataReceived(proxydata)`

* have a reducer that puts that into the Redux store

* have this component subscribed to the Redux store via the container's mapStateToProps

etc

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +249,153 @@ class Developer extends React.Component<Props, State> {
...
+      .then(this.handleProxyData)
+      .catch(error => console.warn('Error in retrieving proxy data, using default data:', error))
+  }
+
+  handleProxyData = (data: RPCTypes.ProxyData) => {
+    var addressPort = data.addressWithPort.split(':')
```
cjb: const instead of var everywhere

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +249,153 @@ class Developer extends React.Component<Props, State> {
...
+  toggleCertPinning = () => {
+    if (this.state.certPinning) {
+      this.props.confirmDisableCertPinning(() => this.setState({certPinning: false}))
+    } else {
+      this.setState({certPinning: !this.state.certPinning})
+    }
+  }
+
+  saveProxySettings = () => {
+    var proxyData = {
```
cjb: This'd be an `onSaveProxySettings` coming from the container that calls dispatch

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +249,153 @@ class Developer extends React.Component<Props, State> {
...
+          style={{marginBottom: Styles.globalMargins.medium, marginTop: Styles.globalMargins.xlarge}}
+        >
+          <Kb.Switch
+            on={!this.state.certPinning}
+            onClick={this.toggleCertPinning}
+            label="Allow TLS Interception"
+          />
+        </Kb.Box2>
+        <Kb.Button style={{margin: Styles.globalMargins.xsmall}} onClick={this.saveProxySettings}>
+          Save Proxy Settings
```
cjb: Use `<Kb.Button label="Save Proxy Settings" ...>`

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +249,153 @@ class Developer extends React.Component<Props, State> {
...
+      console.warn('Error in saving proxy data:', error)
+    )
+  }
+
+  render() {
+    return (
+      <Kb.Box style={styles.proxyContainer}>
+        <Kb.Divider style={styles.divider} />
+        <Kb.Text center={true} type="BodySmallSemibold" style={styles.text}>
+          Configure a HTTP(s) or SOCKS5 proxy
```
cjb: I'd have a quick chat with @keybase/design about how this should look (and what the wording should be) -- putting it all behind a "Configure proxy settings" button would make sense to me since most people don't need it.

Does this work on mobile too? Attaching a mobile screenshot would help too, after the next round of changes.

cecileboucheron:
Hey @ddworken, what do the 3 buttons do at the top? And how does this look when the setting is set up?

ddworken:
@cecileboucheron The three buttons configure whether or not the proxy is enabled. If "No Proxy" is selected then the app will not use a proxy and it will act as it currently does. This is thedefault setting. The other two buttons are types of proxies and selecting them will configure the app to use a proxy. This is the default settings page:

image

This is an example of a configured settings page:

image

shared/settings/advanced/index.tsx
```diff
@@ -54,43 +56,83 @@ const UseNativeFrame = (props: Props) => {
...
-      {isLinux && <UseNativeFrame {...props} />}
-      {!Styles.isMobile && !isLinux && (
-        <Kb.Box style={styles.openAtLoginCheckboxContainer}>
+type AdvancedState = {
+  disableCertPinning: () => void | undefined
+  showCertPinningModal: boolean
+}
+
+class Advanced extends React.Component<Props, AdvancedState> {
+  constructor(props: Props) {
```
chrisnojima: you can just do
```
state = {
  disableCertPinning: undefined,
  showCertPinningModal: false,
}
``` 
and avoid the constructor

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +250,152 @@ class Developer extends React.Component<Props, State> {
   }
 }
 
+// A list so the order of the elements is fixed
+const ProxyTypeToDisplayNameList = [
+  ['noProxy', 'No Proxy'],
+  ['httpConnect', 'HTTP Connect'],
+  ['socks', 'Socks5'],
+]
+const ProxyTypeToDisplayName = ProxyTypeToDisplayNameList.reduce((obj, item) => {
```
chrisnojima: i'd invert this relationship. just have a map of the short name to long name and have the array be the order of the short names, aka

```
const proxyTypeOrder = ['noProxy', 'httpConnect'...
const proxyTypeToDisplayName = invert(displayNameToProxyType)
```

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +250,152 @@ class Developer extends React.Component<Props, State> {
...
+  obj[item[0]] = item[1]
+  return obj
+}, {})
+const DisplayNameToProxyType = {
```
chrisnojima: shouldn't be capitalized

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +250,152 @@ class Developer extends React.Component<Props, State> {
...
+  showCertPinningConfirmationModal: boolean
+}
+
+type ProxyProps = Props & {confirmDisableCertPinning: (disableCertPinning: () => void) => void}
+
+class ProxySettings extends React.Component<ProxyProps, ProxyState> {
+  constructor(props: ProxyProps) {
+    super(props)
+
+    this.state = {
```
chrisnojima: same here

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +250,152 @@ class Developer extends React.Component<Props, State> {
...
+      address: '',
+      certPinning: true,
+      port: '',
+      proxyType: ProxyTypeToDisplayName['noProxy'],
+      showCertPinningConfirmationModal: false,
+    }
+  }
+
+  componentWillReceiveProps(nextProps: ProxyProps) {
```
chrisnojima: shouldn't use this lifecycle even, use componentDidUpdate instead

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +250,152 @@ class Developer extends React.Component<Props, State> {
...
+
+      var certPinning = nextProps.proxyData.certPinning
+
+      var proxyType = ProxyTypeToDisplayName[RPCTypes.ProxyType[nextProps.proxyData.proxyType]]
+
+      this.setState({address, certPinning, port, proxyType})
+    }
+  }
+
+  handleAddressChange = (address: string) => {
```
chrisnojima: we don't need functions like this, just call setstate

shared/settings/advanced/index.tsx
```diff
@@ -54,43 +56,83 @@ const UseNativeFrame = (props: Props) => {
...
-      {!!props.setLockdownModeError && (
-        <Kb.Text type="BodySmall" style={styles.error}>
-          {props.setLockdownModeError}
-        </Kb.Text>
-      )}
-      {isLinux && <UseNativeFrame {...props} />}
-      {!Styles.isMobile && !isLinux && (
-        <Kb.Box style={styles.openAtLoginCheckboxContainer}>
+type AdvancedState = {
+  disableCertPinning: () => void | undefined
```
chrisnojima: i don't think the state should contain this closure that gets plumbed from the child. there are simpler ways to encode what you're trying to do

shared/settings/advanced/index.tsx
```diff
@@ -54,43 +56,83 @@ const UseNativeFrame = (props: Props) => {
...
+        </Kb.Box>
+        <Kb.Box style={styles.checkboxContainer}>
           <Kb.Checkbox
-            label="Open Keybase on startup"
-            checked={props.openAtLogin}
-            onCheck={props.onSetOpenAtLogin}
+            checked={this.props.hasRandomPW || !!this.props.lockdownModeEnabled}
+            disabled={disabled}
+            label={
+              'Forbid account changes from the website' +
```
chrisnojima: use string templates here instead

shared/settings/advanced/index.tsx
```diff
@@ -54,43 +56,83 @@ const UseNativeFrame = (props: Props) => {
...
+        {!Styles.isMobile && !isLinux && (
+          <Kb.Box style={styles.openAtLoginCheckboxContainer}>
+            <Kb.Checkbox
+              label="Open Keybase on startup"
+              checked={this.props.openAtLogin}
+              onCheck={this.props.onSetOpenAtLogin}
+            />
+          </Kb.Box>
+        )}
+        <ProxySettings {...{...this.props, confirmDisableCertPinning: this.confirmDisableCertPinning}} />
```
chrisnojima: shouldn't pass all props but if you need to you can do
```
{...this.props} confirmDisableCertPinning={this.confirmDisableCertPinning}
```

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Migrated to a new modal route per cjb's feedback.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Switched to just `{...this.props}` since I got rid of the closure via migrating to a modal route.

chrisnojima:
@ddworken just ping me when you're done iterating

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/actions/settings.tsx
```diff
@@ -433,6 +434,27 @@ const loadLockdownMode = state =>
...
+
+const saveProxyData = (_, proxyDataPayload: SettingsGen.SaveProxyDataPayload) =>
+  RPCTypes.configSetProxyDataRpcPromise(proxyDataPayload.payload).catch(err =>
+    logger.warn('Error in saving proxy data', err)
+  )
+
+const setCertificatePinning = (_, certPinningPayload: SettingsGen.SetCertificatePinningPayload) =>
+  RPCTypes.configGetProxyDataRpcPromise(undefined)
+      .then((result: RPCTypes.ProxyData) => {
+        const cloned = JSON.parse(JSON.stringify(result))
```
chrisnojima: you'll lose all typing by doing this, instead
```
const proxyData = {
...result,
certPinning: action.payload.enabled // also rename it to action so it matches everything else
}
```

shared/settings/advanced/index.tsx
```diff
@@ -54,43 +61,45 @@ const UseNativeFrame = (props: Props) => {
...
-      </Kb.Box>
-      {!!props.setLockdownModeError && (
-        <Kb.Text type="BodySmall" style={styles.error}>
-          {props.setLockdownModeError}
-        </Kb.Text>
-      )}
-      {isLinux && <UseNativeFrame {...props} />}
-      {!Styles.isMobile && !isLinux && (
-        <Kb.Box style={styles.openAtLoginCheckboxContainer}>
+class Advanced extends React.Component<Props, {}> {
```
chrisnojima: why is this a component now

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +217,128 @@ class Developer extends React.Component<Props, State> {
...
+        return true // Default value
+      }
+    } else {
+      return !this.props.allowTlsMitmToggle
+    }
+  }
+
+  renderProxySettings() {
+    if (this.state.proxyType === 'noProxy') {
+      return undefined
```
chrisnojima: return null

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +217,128 @@ class Developer extends React.Component<Props, State> {
...
+    }
+    return (
+      <Kb.Box direction="vertical" alignItems="flex-start" style={{width: '40ch'}}>
+        <Kb.Box2
+          direction="vertical"
+          gap="tiny"
+          style={{marginBottom: Styles.globalMargins.small, marginTop: Styles.globalMargins.small}}
```
chrisnojima: all these styles should move into the stylesheet

shared/settings/disable-cert-pinning-modal/container.tsx
```diff
@@ -0,0 +1,28 @@
...
+  onCancel: () => {
+    dispatch(RouteTreeGen.createNavigateUp())
+  },
+  onConfirm: () => {
+    dispatch(createCertificatePinningToggled({toggled: true}))
+    dispatch(RouteTreeGen.createNavigateUp())
+  },
+})
+
+const mergeProps = (stateProps, dispatchProps, ownProps) => ({
```
chrisnojima: remove ownprops

shared/settings/disable-cert-pinning-modal/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+import * as Kb from '../../common-adapters'
+import {globalStyles} from "../../styles";
+import {globalMargins} from "../../styles";
+
+export type Props = {
+  onCancel: () => void
+  onConfirm: () => void
+}
+
+class ConfirmDisableCertPinningModal extends React.Component<Props> {
```
chrisnojima: shouldn't be a component

ddworken: Changed to a function.

shared/settings/disable-cert-pinning-modal/container.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Done.

shared/settings/advanced/index.tsx
ddworken: Changed back. Earlier I had added state and then ended up removing it.

shared/actions/settings.tsx
ddworken: Deleted (sorry about that, that code was dead and I forgot to delete it).

shared/settings/advanced/index.tsx
```diff
@@ -55,40 +62,40 @@ const UseNativeFrame = (props: Props) => {
 }
 
 const Advanced = (props: Props) => {
-  const disabled = props.lockdownModeEnabled == null || props.hasRandomPW || props.settingLockdownMode
+  const disabled =
+    this.props.lockdownModeEnabled == null || this.props.hasRandomPW || this.props.settingLockdownMode
```
chrisnojima: should no longer be this.props, instead just props (and below)

shared/settings/advanced/index.tsx
```diff
@@ -240,9 +359,17 @@ const styles = Styles.styleSheetCreate({
   error: {
     color: Styles.globalColors.red,
   },
+  expandedProxyContainer: {
+    marginTop: Styles.globalMargins.small,
+    width: '40ch',
```
chrisnojima: lets not use these types of css units. it won't work w/ react native.

ddworken:
@cjb As requested here are screenshots of it on Android. I also tested that it does work correctly on Android. 

Screenshot_20190618-141059
Screenshot_20190618-141112
Screenshot_20190618-141133
Screenshot_20190618-141443

cecileboucheron:
* For the confirm screen we should use ConfirmModal, it has the right layout already
* Here it needs more space between the title and the divider
image
* The container box for these should be full-width on mobile, and the inputs too should be full width:
image
* We need more space between the switch and the text here:
image

shared/settings/advanced/index.tsx
```diff
@@ -208,6 +216,136 @@ class Developer extends React.Component<Props, State> {
...
+    }
+    this.setState({proxyType}, cb)
+  }
+
+  renderProxySettings() {
+    if (this.state.proxyType === 'noProxy') {
+      return null
+    }
+    return (
+      <Kb.Box direction="vertical" alignItems="flex-start" style={styles.expandedProxyContainer}>
```
cjb: This "alignItems" doesn't do anything, and there's a warning on the console about it -- it's a property used by style={}.

ddworken:
@cecileboucheron Fixed those things. Here are updated screenshots of both android and desktop: 

Screenshot from 2019-06-18 16-30-31
Screenshot from 2019-06-18 16-30-50
Screenshot from 2019-06-18 16-31-14
Screenshot_20190618-162901
Screenshot_20190618-162914
Screenshot_20190618-162921

cecileboucheron:
Good, let's push down the content below the title now, it feels too close. Maybe 16px more pixels.
image

ddworken:
Screenshot_20190618-175159
Screenshot_20190618-175153
image
image

pzduniak:
@ddworken 2 cents from a guy who unfortunately had to write an enterprise firewall.

The current captions are giving me a "crappy software written by an outsourcing company in Asia" vibes. Purely UI-wise I think the captions should be as following:

1. `No proxy` rather than `No Proxy` (there's lots of `Proxy` in that widget) 
2. `HTTP` rather than `HTTP Connect` - a "HTTP proxy" is a proxy that exposes CONNECT for HTTPS connections and supports other methods with the full URL after the verb, therefore "HTTP Connect" is somewhat redundant.
3 `SOCKS5` rather than `Socks5` - I have no clue why Gnome calls this "Socks", the family of protocols is called "SOCKS", all letters uppercase.

image

There's also a "HTTPS" proxy type which I haven't seen in the wild but from what I've seen it's just the standard HTTP protocol wrapped in a TLS tunnel. Might be worth throwing this in if it's barely any effort?

ddworken:
@pzduniak Thanks for the feedback! I definitely agree with 1&2 and will change those (PR [here](https://github.com/keybase/client/pull/18043)). The reason I want to say "HTTP Connect" rather than just "HTTP" is because "HTTP" really can refer to two different types of proxies. See [this](https://en.wikipedia.org/wiki/Proxy_server#Web_proxy_servers) wikipedia article which has a good description of the other type of "HTTP proxy". So I'm hesitant to change it when we don't actually support the first type of HTTP proxy. Also, it is a bit of a personal pet peeve of mine that everyone simply says http proxy given the hidden ambiguity. Though I do see an argument that we should match what everyone else says, what do you think? 

Also, we do actually support https connect proxies, so I changed the label to more accurately reflect that.

PR #17884, opened 2019-06-10
data: implement PathPartString

strib:
This struct will be used in place of regular strings for parts of KBFS paths.  It uses an obfuscator to output `String()`, so that plaintext is never accidentally marshaled.

It cannot be marshaled directly for JSON; it must first be explicitly converted into plaintext.

Issue: HOTPOT-57

### DIFF
--- a/go/kbfs/data/path_part_string.go
+++ b/go/kbfs/data/path_part_string.go
@@ -1,0 +1,80 @@
+// Copyright 2019 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package data
+
+import (
+	"encoding"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+const (
+	prefixToSkipObfsucation = ".kbfs_"
+	tarGzSuffix             = ".tar.gz"
+
+	// extRestoreMaxLen specifies the max length of an extension, such
+	// that it's restored to the obfuscated string.  It includes the
+	// leading dot.
+	extRestoreMaxLen = 5
+)
+
+// PathPartString returns an obfuscated version of part of a path,
+// preserving the suffixes if they're small enough or if it's a
+// conflicted copy of a file.  Two `PathPartString`s for the asme
+// plaintext data, created using the same `Obfuscator`, should pass
+// equality checks.
+type PathPartString struct {
+	plaintext  string
+	ext        string
+	obsfucator Obfuscator
+}
+
+var _ fmt.Stringer = PathPartString{}
+
+// NewPathPartString creates a new `PathPartString` instance, given a
+// plaintext string representing one part of a path, and a
+// `Obfuscator`.
+func NewPathPartString(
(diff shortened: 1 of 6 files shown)


go/kbfs/data/path_part_string.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"encoding"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+const (
+	prefixToSkipObfsucation = ".kbfs_"
```
jakob223: is there any chance we'll want to expand this to more than one thing such that this should be an array of strings?

go/kbfs/data/path_part_string.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// extRestoreMaxLen specifies the max length of an extension, such
+	// that it's restored to the obfuscated string.  It includes the
+	// leading dot.
+	extRestoreMaxLen = 5
+)
+
+// PathPartString returns an obfuscated version of part of a path,
+// preserving the suffixes if they're small enough or if it's a
+// conflicted copy of a file.  Two `PathPartString`s for the asme
```
jakob223: s/`asme`/`same`

go/kbfs/data/path_part_string.go
```diff
@@ -0,0 +1,80 @@
...
+
+var _ fmt.Stringer = PathPartString{}
+
+// NewPathPartString creates a new `PathPartString` instance, given a
+// plaintext string representing one part of a path, and a
+// `Obfuscator`.
+func NewPathPartString(
+	plaintext string, obfuscator Obfuscator) PathPartString {
+	_, ext := SplitFileExtension(plaintext)
+	if len(ext) > extRestoreMaxLen && ext != tarGzSuffix {
```
jakob223: why make an exception for `.tar.gz` [and not for `.xcodeproj` / `.tar.bz2` / all the other things that are more than 5 characters]? Just seems sorta random.

go/kbfs/data/path_part_string.go
strib: I think it's very low.  in the 4+ years of KBFS, we've only had this one special prefix.  if we ever do, we can always implement the array at that time instead of wasting CPU on it now.

go/kbfs/data/path_part_string.go
strib: Good question, I was just following the logic that @taruti wrote in the original extension parsing for conflict resolution.  I guess it was just an arbitrary compromise at the time.  That code (now in `SplitFileExtension`) would not return `tar.bz2` (just `.bz2`).  It would return `.xcodeproj`, but here I'm just trying to be sensitive to people who just put dots in files names rather than underscores, where the last part of the name might be meaningful.  I'm just guessing that for debugging purposes, only the usual short extensions will be useful to us, and so it's not worth the privacy risk for catching longer ones that aren't specially flagged by existing code already.

I can add a comment to that effect.

PR #17627, opened 2019-05-17
plumb actions and redux store for new conflict protocol changes

songgao:
@jakob223 I created a temporary branch by rebasing `jakob/KBFS-4151-resolve-manually-link` to master, and am opening this PR against that branch so it's clear what's changed. It's mostly plumbing, plus fixing flow errors by changing as less as possible. Let me know what you think!

PR #17885, opened 2019-06-10
plumb actions and redux store for new conflict protocol changes

songgao:
a.k.a. a post-tsc remix of https://github.com/keybase/client/pull/17627


shared/constants/types/fs.tsx
```diff
@@ -82,22 +82,44 @@ export type TlfSyncPartial = I.RecordOf<_TlfSyncPartial>
...
+}
+export type ConflictStateAutomaticResolving = I.RecordOf<_ConflictStateAutomaticResolving>
+
+export type _ConflictStateManualResolvingServerView = {
+  localViewTlfPaths: I.List<Path>
+  type: ConflictStateType.ManualServerView
+}
+export type ConflictStateManualResolvingServerView = I.RecordOf<_ConflictStateManualResolvingServerView>
+
+export type _ConflictStateManualResolvingLocalView = {
```
jakob223: If I split A into A and B, A gets ManualResolvingServer and B gets ManualResolvingLocal. Then what happens when there's a new conflict in A?

I feel like maybe we ought to combine ManualResolvingServer and AutomaticResolving, such that you can be in both states at once, if that makes any sense.

WDYT?

shared/fs/banner/conflict-banner-container.tsx
```diff
@@ -12,25 +13,26 @@ type OwnProps = {
...
   onHelp: () => void
-  onSeeOtherView: () => void
   onStartResolving: () => void
 }
 const mapDispatchToProps = (dispatch, ownProps: OwnProps) => ({
   onFeedback: () => {},
-  onFinishResolving: () => {},
+  onFinishResolving: dispatch(FsGen.createFinishManualConflictResolution({localViewTlfPath: ownProps.path})),
+  onGoToTlf: (tlfPath: Types.Path) =>
+    dispatch(RouteTreeGen.createNavigateAppend({path: [{props: {path: tlfPath}, selected: 'main'}]})),
```
jakob223: It could be cool to route to the corresponding subfolder when you click on the banner in a subfolder of the conflict TLF. Not necessarily a thing for this particular PR, but a cool thing to have in the future if you agree

shared/fs/banner/conflict-banner.tsx
```diff
@@ -5,35 +5,35 @@ import * as Types from '../../constants/types/fs'
...
-            'sure to copy any changes you want to keep into the global view.'
-        : `This is the rest of the world's view of ${tlf}.` +
-            " When you're satisfied with this view, you can click finish."
-    case Types.ConflictState.Finishing:
-      return 'Finishing conflict resolution...'
-    case Types.ConflictState.None:
-    case Types.ConflictState.InConflictNotStuck:
+    case Types.ConflictStateType.ManualServerView:
+      return (
+        `This is the rest of the world's view of ${tlf}.` +
```
jakob223: This message changed in my PR because of the change to where you can click finish from - I can help rebase this after my PR is merged, if you like, since it's my changes causing the problems.

shared/fs/banner/conflict-banner.tsx
```diff
@@ -42,44 +42,43 @@ const getMessage = (tlf: string, conflictState: Types.ConflictState, isUnmerged)
...
-    case Types.ConflictState.InConflictNotStuck:
-    case Types.ConflictState.None:
-      actions = [feedbackAction]
-      break
+      return <Kb.Banner {...commonProps} actions={[onSeeGlobalView, finishRes, feedbackAction, helpAction]} />
+    case Types.ConflictStateType.ManualServerView:
+      const count = props.conflictState.localViewTlfPaths.size
+      const onSeeLocalViews = props.conflictState.localViewTlfPaths.toArray().map((tlfPath, idx) => ({
+        onClick: () => props.onGoToTlf(tlfPath),
+        title: 'See local changes' + (count > 1 ? ` (version ${idx.toString} of ${count}` : ''),
```
jakob223: 👌

shared/fs/browser/index.stories.tsx
```diff
@@ -100,18 +95,6 @@ export default () => {
         />
       </Kb.Box2>
     ))
-    .add('in conflict', () => (
```
jakob223: why did you take away this story?

shared/constants/types/fs.tsx
songgao: > If I split A into A and B, A gets ManualResolvingServer and B gets ManualResolvingLocal. Then what happens when there's a new conflict in A?

I _think_ in that case A would become "automatic resolve" with isStuck being true, then you can start resolve again from A to split into C. @strib could you confirm?

songgao: > I feel like maybe we ought to combine ManualResolvingServer and AutomaticResolving, such that you can be in both states at once, if that makes any sense.

Hmm perhaps ManualResolvingServer and the "none" case should merge instead. Without any more conflict being resolved, ManualResolvingServer is essentially "none" right?

shared/fs/banner/conflict-banner-container.tsx
songgao: AH that's right. Will change!

shared/fs/banner/conflict-banner.tsx
songgao: I don't mind rebasing either; I guess depends on which one gets merged first.

shared/fs/browser/index.stories.tsx
songgao: because the conflict state isn't passed in from the browser. There's one in the banner stories.

shared/constants/types/fs.tsx
jakob223: yeah.. how about merging all three of ManualServer, None, and Automatic, such that at any time you can have (1) a list of conflict copies and (2) a current conflict?

shared/fs/banner/conflict-banner.tsx
jakob223: k. mine is merged now - lmk if rebase causes you issues

shared/constants/types/fs.tsx
songgao: Good point! I've just made this change. PTAL thanks!

songgao:
@jakob223 I also changed the banner color to yellow when we are stuck. Otherwise server view of a manual resolving seems too similar to server view with manual resolving and stuck again.

protocol/avdl/keybase1/favorite.avdl
```diff
@@ -18,27 +18,29 @@ protocol favorite {
...
-  record ConflictAutomaticResolving {
-    boolean isStuck;
-  }
-
-  record ConflictManualResolvingServerView {
+  record FolderServerView {
+    // Set to true if we're trying to automatically resolve conflict in this
+    // folder. Whether or not there's a local view version of this TLF doesn't
+    // affect this.
+    boolean resolvingConflict;
```
strib: I'm confused by this -- if this is supposed to be a "ServerView", how can it ever be resolving a conflict?  If we have a conflict locally and it's not stuck, then the user is not seeing the server view; they're seeing their local conflict view, and currently have no way of seeing the real server view until the conflict resolution is complete, or gets stuck and moved out of the way.  Maybe `FolderServerView` is just the wrong name for this structure?  Is the idea here that it's representing the current local state of the normal path to the TLF?  Maybe `FolderNormalView`?  `FolderNormalPathView`?

songgao: Good point! That must have been why I did it with two separate types in the first place. Sigh. Sorry for not recalling this earlier. Normal view sounds reasonable; will update.

PR #17887, opened 2019-06-11
only show loading spinner at first time loading

songgao:
This is motivated by the issue where when a path is being synced, we'd load folder list every second, which results in the spinner being shown every second. I talked to Cécile and we decided to only show it when first time loading a thing. This means even if a folder content itself changes, we'll just show the changed content and no spinner, which seems fine.

I was slightly wrong when I said this wouldn't be triggered outside admin mode. It actually gets triggered too if user enables sync from CLI. So perhaps it still makes sense to get it into this release.

### DIFF
--- a/shared/fs/top-bar/loading.tsx
+++ b/shared/fs/top-bar/loading.tsx
@@ -2,6 +2,7 @@ import * as React from 'react'
 import {namedConnect} from '../../util/container'
 import * as I from 'immutable'
 import * as Types from '../../constants/types/fs'
+import * as Constants from '../../constants/fs'
 import * as Kb from '../../common-adapters'
 
 type OwnProps = {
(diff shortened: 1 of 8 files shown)


shared/fs/top-bar/loading.tsx
```diff
@@ -10,14 +11,28 @@ type OwnProps = {
...
 const emptySet = I.Set()
 
-const mergeProps = (stateProps, dispatchProps, {path}: OwnProps) => ({
-  loading: stateProps._loadingPaths.get(path, emptySet).size > 0,
-})
+const mergeProps = (stateProps, dispatchProps, {path}: OwnProps) => {
+  // Only show the loading spinner when we are first-time loading a pathItem.
+  // If we already have content to show, just don't show spinner anymore even
+  // if we are loading.
+  if (stateProps._pathItem.type === Types.PathType.Unknown) {
```
buoyad: Looks like this is true for `private`,`public`, and `team`, the spinner is always on for those

songgao: Opops good catch!

shared/fs/top-bar/loading.tsx
```diff
@@ -2,22 +2,56 @@ import * as React from 'react'
 import {namedConnect} from '../../util/container'
 import * as I from 'immutable'
 import * as Types from '../../constants/types/fs'
+import * as Constants from '../../constants/fs'
 import * as Kb from '../../common-adapters'
+import * as Flow from '../../util/flow'
+
+// The behavior is to only show spinner when user first time lands on a screen
+// and when don't have the data thar drives it yet. Since RPCs happen
```
buoyad: `that`

PR #17889, opened 2019-06-11
a bunch of fixes for Files tab

songgao:


shared/common-adapters/list-item2.tsx
```diff
@@ -193,9 +199,6 @@ const styles = Styles.styleSheetCreate({
     right: 0,
     top: 0,
   },
-  hiddenNoHoverAction: {
-    justifyContent: 'flex-end',
```
songgao: @jakob223 This doesn't seem used as the actions are passed in as a box. Let me know if I'm wrong!

shared/common-adapters/list-item2.tsx
```diff
@@ -21,30 +21,37 @@ type Props = {
...
       ':hover .avatar-border': {
+        // TODO: it'd be nice to move this out of this file since list-item2
+        // has nothing todo with avatars. We'd need Kb.AvatarLine to know
+        // about different background and set that on hover though.
         boxShadow: `0px 0px 0px 2px ${Styles.globalColors.blueLighter2} !important`,
       },
-      ':hover .hidden-no-hover': {
-        maxWidth: 64,
+      ':hover .fade': {
+        opacity: 1,
```
songgao: This is where it went wrong for the editing row. The `max-width` is limiting the width of actions for all rows. So instead, don't use max-width for the "appear" case, and use `opacity` instead.

Also instead of a fixed `max-width` in the `:hover` case, use `'unset'`.

shared/common-adapters/list-item2.tsx
```diff
@@ -21,30 +21,37 @@ type Props = {
   body: React.ReactNode
   firstItem: boolean
   action?: React.ReactNode
-  onlyShowActionOnHover?: 'appear' | 'animate' | null
+  onlyShowActionOnHover?: 'fade' | 'grow' | null
```
songgao: @jakob223 I renamed this a bit; let me know if you don't like this!

songgao:
Thanks Nojima. I’ll merge this first and fix the mobile send attachment later

shared/common-adapters/list-item2.tsx
jakob223: This is used - it prevents the actions from moving as they are uncovered in the 'grow' case, and instead keeps them in one place as the avatars move out from on top of them.

PR #17893, opened 2019-06-11
start version 4.2.0

chrisnojima:
@keybase/react-hackers ups to version 420


shared/android/app/build.gradle
```diff
@@ -6,7 +6,7 @@ apply from: '../../node_modules/react-native-unimodules/gradle.groovy'
 import com.android.build.OutputFile
 
 // KB: app version
-def VERSION_NAME = "4.0.0"
```
buoyad: missed this last time?

chrisnojima: i think it was a function of the build changing to app bundling

PR #17897, opened 2019-06-11
External tools and partners: storybook and integration

cjb:

When loading a settings page, if the admin-only `stellarExternalPartners` feature flag is on, also show a list of external tools and partners from the service.  If `%{accountId}` exists in the url then interpolate the account's public key.  If %{username}` exists in the url then supply it in the url, but only if the account is primary.

There's an `iconFilename` field in the payload but I'm ignoring it for now, need to think about whether the service should be supplying those through the local web server.


shared/wallets/wallet/settings/index.tsx
```diff
@@ -168,6 +197,46 @@ class AccountSettings extends React.Component<SettingsProps> {
...
+                <Kb.Box2
+                  direction="vertical"
+                  style={styles.section}
+                  fullWidth={true}
+                  gap="tiny"
+                  gapEnd={true}
+                >
+                  <Kb.Text type="BodySmallSemibold">External tools and partners</Kb.Text>
+                  {props.externalPartners.map((partner, index, array) => {
+                    const showDivider = index !== array.length - 1
```
chrisnojima: if you do the opposite and put the divider in front you don't need to use the array at all, just use index > 0

shared/wallets/wallet/settings/index.tsx
```diff
@@ -168,6 +197,46 @@ class AccountSettings extends React.Component<SettingsProps> {
...
+                    // Note: `props.user` is only the Keybase username if
+                    // this is the primary account. Non-primary accounts
+                    // are not associated with usernames.
+                    const url = partner.url
```
chrisnojima: maybe better for the connector to do this. then it can figure out the url and showDivider

mlsteele:
The snorkler is an odd default for partners without icons.

Maybe hide the "External tools and partners" section when there are no rows.

cjb:
@chrisnojima Thanks, made these changes.

@mlsteele Brought in new Stellar icons from Zeplins so I can use `icon-stellar-logo-grey`.

shared/wallets/wallet/settings/container.tsx
```diff
@@ -29,7 +52,11 @@ const mapStateToProps = state => {
   const canSubmitTx = account.canSubmitTx
   const thisDeviceIsLockedOut = account.deviceReadOnly
   const inflationDest = Constants.getInflationDestination(state, accountID)
-  const externalPartners = Constants.getExternalPartners(state, accountID).toArray()
+  const externalPartners = prepareExternalPartners(
```
chrisnojima: do not make a call like this in mapstate  you'll thrash every time, belongs in merge or needs to be memoized

cjb:
@chrisnojima Thanks, over to mergeProps now.

PR #17900, opened 2019-06-11
New Signup 2 - Route to optional phone number entry after signup

buoyad:
- Add `signedUp` to `notifySession.loggedIn` so we know we just signed up and to route to phone number entry
- Hook up container for `signup/phone-number` and `addPhoneNumber` RPC

- [ ] storyshots before merge

cc @keybase/y2ksquad


buoyad:
@chrisnojima this'll wait until the verification code PR is ready and merged into this.

PR #17905, opened 2019-06-11
[Y2K-36] Files Settings on Mobile

jzila:
This implements settings for files notifications on mobile.

cc @keybase/y2ksquad 

TODO: before merging, I need to revert c4f1598 since it bypasses the offline flag for testing.


jzila:
@jakob223 there is one more coming, that connects the notifications to the new state.

PR #17912, opened 2019-06-12
use safe-navigation in files

songgao:


shared/fs/browser/destination-picker/container.tsx
```diff
@@ -9,34 +9,35 @@ import * as FsGen from '../../../actions/fs-gen'
...
-const getDestinationParentPath = memoize((stateProps: StateProps, ownProps: OwnProps) =>
-  stateProps._destinationPicker.destinationParentPath.get(
-    getRouteProps(ownProps, 'index') || 0,
-    stateProps._destinationPicker.source.type === Types.DestinationPickerSource.MoveOrCopy
-      ? Types.getPathParent(stateProps._destinationPicker.source.path)
-      : Types.stringToPath('/keybase')
-  )
+
+const getIndex = memoize((ownProps: OwnPropsWithSafeNavigation) => ownProps.getParam('index') || 0)
+const getDestinationParentPath = memoize(
```
chrisnojima: i don't think either of these should be memoized.
1. getIndex is fast so this'll just be slower and use more memory
2. getDestinationParentPath is memoizing on state and ownprops and both are likely mutating all the time

PR #17913, opened 2019-06-12
allow full log sends

joshblum:
patch does the following:

- adds a hidden option to send 'full logs' (128mb)
- moves the ios logs into a subdir that has background readable permissions so logs can properly rotate
- keep a second log file on mobile when not admin so we can get logs after a rotation

cc @keybase/hotpotatosquad

### DIFF
--- a/shared/ios/Keybase/Fs.m
+++ b/shared/ios/Keybase/Fs.m
@@ -134,21 +134,12 @@
   NSString* serviceLogFile = skipLogFile ? @"" : [logPath stringByAppendingString:@"/ios.log"];
 
   if (!skipLogFile) {
-      // migrate old log files to the new location
+      // cleanup old log files
       NSFileManager* fm = [NSFileManager defaultManager];
-      NSError* error = nil;
       NSArray<NSString*>* logs = @[@"/ios.log", @"/ios.log.ek"];
       for (NSString* file in logs) {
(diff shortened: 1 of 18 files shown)


shared/ios/Keybase/Fs.m
```diff
@@ -128,12 +128,12 @@ - (NSDictionary*) setupFs:(BOOL)skipLogFile setupSharedHome:(BOOL)setupSharedHom
   // Setup app level directories
   NSString* levelDBPath = [@"~/Library/Application Support/Keybase/keybase.leveldb" stringByExpandingTildeInPath];
   NSString* chatLevelDBPath = [@"~/Library/Application Support/Keybase/keybase.chat.leveldb" stringByExpandingTildeInPath];
-  NSString* logPath = [@"~/Library/Caches/Keybase" stringByExpandingTildeInPath];
+  NSString* logPath = [@"~/Library/Caches/Keybase/logs" stringByExpandingTildeInPath];
```
chrisnojima: we need some kind of cleanup of old logs story if we're moving the logs around

joshblum: ah good catch

joshblum:
@chrisnojima this is ready

joshblum:
merging over flake

PR #17917, opened 2019-06-12
normalize query during upgrade [HOTPOT-124]

joshblum:
makes conv search case insensitive


joshblum:
moved to inboxsource otherwise we could nuke regexps with uppercase chars

PR #17926, opened 2019-06-13
New Signup 3 - verify phone number

buoyad:
- Route to verify screen after `addPhoneNumber`
- Accept code and call to verify
- Exit on success

cc @keybase/y2ksquad

### DIFF
--- a/shared/actions/json/settings.json
+++ b/shared/actions/json/settings.json
@@ -6,9 +6,10 @@
   ],
   "actions": {
     "addPhoneNumber": {
-      "_description": "Add a phone number and kick off a text message with a verification code. If the phone number is already added, just kicks off the verification message.",
+      "_description": "Add a phone number and kick off a text message with a verification code. If `resend` is passed, ignores the other params and uses stashed params from store.",
       "allowSearch": "boolean",
-      "phoneNumber": "string"
+      "phoneNumber": "string",
+      "resend?": "boolean"
     },
     "addedPhoneNumber": {
       "_description": "We just attempted to add a phone number and either got an error or the number is pending verification.",
--- a/shared/actions/json/settings.json
+++ b/shared/actions/json/settings.json
@@ -12,6 +12,7 @@
     },
     "addedPhoneNumber": {
       "_description": "We just attempted to add a phone number and either got an error or the number is pending verification.",
+      "allowSearch": "boolean",
       "error": "string",
       "phoneNumber": "string"
     },
--- a/shared/actions/json/settings.json
+++ b/shared/actions/json/settings.json
@@ -14,7 +14,7 @@
     "addedPhoneNumber": {
       "_description": "We just attempted to add a phone number and either got an error or the number is pending verification.",
       "allowSearch": "boolean",
-      "error": "string",
+      "error?": "string",
       "phoneNumber": "string"
     },
     "verifyPhoneNumber": {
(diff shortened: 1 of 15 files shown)


shared/signup/phone-number/verify.tsx
```diff
@@ -3,105 +3,101 @@ import * as Kb from '../../common-adapters'
...
-    return (
-      <SignupScreen
-        onBack={this.props.onBack}
-        banners={this.props.error ? [<Kb.Banner key="error" color="red" text={this.props.error} />] : []}
-        buttons={[{label: 'Continue', onClick: this.props.onContinue, type: 'Success'}]}
-        titleComponent={
+const VerifyPhoneNumber = (props: Props) => {
+  // Clear data on unmount
+  React.useEffect(() => {
+    return props.onCancel
```
jakob223: did you mean to call `onCancel` instead of returning it? I don't understand what this is doing if not.

shared/signup/phone-number/verify.tsx
```diff
@@ -3,105 +3,101 @@ import * as Kb from '../../common-adapters'
...
+          autoFocus={true}
+          style={styles.input}
+          flexable={true}
+          keyboardType="numeric"
+          onChangeText={onChangeCode}
+          textType="Header"
+          textContentType="oneTimeCode"
+        >
+          <Kb.Text type="Header" style={styles.inputText}>
+            {/* We put this child in Input because some text styles don't work on input itself - the one we need here is letterSpacing */}
```
jakob223: Is that a limitation of the base react/HTML component, or of our `PlainInput`? if the latter, should we modify `PlainInput` to allow this style?

shared/actions/settings.tsx
```diff
@@ -515,11 +519,48 @@ const passwordChanged = () => SettingsGen.createLoadedHasRandomPw({randomPW: fal
...
   RPCTypes.ctlStopRpcPromise({exitCode: action.payload.exitCode})
 
-const addPhoneNumber = (_, action: SettingsGen.AddPhoneNumberPayload, logger) => {
+const addPhoneNumber = (state: TypedState, action: SettingsGen.AddPhoneNumberPayload, logger) => {
   logger.info('adding phone number')
-  const {phoneNumber, allowSearch} = action.payload
+  let {phoneNumber, allowSearch, resend = false} = action.payload
+  if (resend) {
+    logger.info('resending verification code')
+    phoneNumber = state.settings.phoneNumbers.pendingVerification
```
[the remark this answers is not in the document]
jakob223: Having read the rest of the code, this makes sense now - never mind.

shared/signup/phone-number/verify.tsx
buoyad: It's an RN limitation

shared/signup/phone-number/verify.tsx
buoyad: No, this is `componentWillUnmount` a-la hooks. We return the callback that should be called on unmount and updates. I actually think this is a pretty bad pattern, it is also called when props in the second array change, so I imagine it could lead to terrible unexpected behavior if something in the connector changes.

shared/actions/json/settings.json
```diff
@@ -6,15 +6,30 @@
...
       "phoneNumber": "string"
     },
     "verifyPhoneNumber": {
       "_description": "Submit a verification code for a phone number",
       "phoneNumber": "string",
       "code": "string"
     },
+    "verifiedPhoneNumber": {
+      "_description": "We verified a phone number or hit an error.",
+      "error": "string",
```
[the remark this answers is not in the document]
buoyad: Yep

PR #17928, opened 2019-06-13
More strict ts: generated types

chrisnojima:
- [x] before all of the common-adapters were pulled in from the constants/types side due to these icon.constants to get IconType, instead just pull in this type and not everything


shared/constants/types/fs.tsx
```diff
@@ -632,7 +632,7 @@ export const stringToPath = (s: string): Path =>
 export const pathToString = (p: Path): string => (!p ? '' : p)
 export const stringToLocalPath = (s: string): LocalPath => s
 export const localPathToString = (p: LocalPath): string => p
-export const getPathName = (p: Path): string => (!p ? '' : p.split('/').pop())
+export const getPathName = (p: Path): string => (!p ? '' : p.split('/').pop() || '')
```
chrisnojima: @songgao please sanity check

chrisnojima:
@keybase/react-hackers

shared/constants/types/fs.tsx
songgao: This is sane ✅

shared/constants/types/more.tsx
```diff
@@ -1,56 +1,64 @@
...
+  | 'facebook'
+  | 'github'
+  | 'hackernews'
+  | 'http'
+  | 'https'
+  | 'reddit'
+  | 'rooter'
+  | 'twitter'
+  | 'web'
+  | 'rooter'
```
songgao: Any reason why we didn't go with enum here? Not suggesting a change necessarily; just curious

shared/actions/profile/proofs.tsx
```diff
@@ -61,7 +61,7 @@ const recheckProof = (state, action: ProfileGen.RecheckProofPayload) =>
 // only let one of these happen at a time
 let addProofInProgress = false
 function* addProof(state, action: ProfileGen.AddProofPayload) {
-  const service = More.isPlatformsExpandedType(action.payload.platform)
+  const service = More.asPlatformsExpandedType(action.payload.platform)
```
songgao: This can still be the `is` version right?

chrisnojima: no, the type guard wants a boolean. this original function is a cast to the thing or null, so i renamed the old on `as` and made the type guard `is` as i think its more accurate

songgao: Oh right. I missed the `switch (service)` below and thought it was only used in the line immediately below. LGTM

PR #17929, opened 2019-06-13
reply to attachments

mmaxim:
Adds the ability to reply to attachments, and fixes some styling with reply display.

cc @keybase/hotpotatosquad


shared/chat/conversation/messages/text/container.tsx
```diff
@@ -22,14 +22,23 @@ const getReplyProps = (replyTo, onReplyClick) => {
...
+    case 'attachment':
     case 'text':
+      const attachment: Types.MessageAttachment =
+        replyTo.type === 'attachment' && replyTo.attachmentType === 'image' ? replyTo : null
       return replyTo.exploded
         ? deletedProps
         : {
             deleted: false,
             edited: replyTo.hasBeenEdited,
+            imageHeight: attachment ? attachment.previewHeight : undefined,
```
buoyad: :sheep: feels a little redundant to have the switch and also do this. would splitting the cases duplicate too much?

PR #17930, opened 2019-06-13
more ts-strict

chrisnojima:
- [x] pushing forward getting redux in which requires all of typedstate/typedactions to type strict
- [x] constants/chat2 fixes


chrisnojima:
@keybase/react-hackers 
cc: @mmaxim

PR #17933, opened 2019-06-13
add trustline asset component (and some ListItem2 changes)

songgao:
Since this is 1) expandable; 2) icon-less (after talking to Cécile), I had to make some changes to ListItem2 as well to support these.

I'll have a separate PR for the actual trustline modal screen.

Story screencast: /keybase/team/keybase/videos/trustline-asset.mov


songgao:
oops I forgot mobile. Untagging review for now.

songgao:
this is ready! @keybase/react-hackers @keybase/picnicsquad

shared/wallets/trustline/asset.tsx
```diff
@@ -0,0 +1,153 @@
...
+  trusted: boolean // TODO add limit when we support it in GUI
+
+  onAccept: () => void
+  onCollapse: () => void
+  onExpand: () => void
+  onRemove: () => void
+  onViewDetails: () => void
+}
+
+const stopPropagation = memoize(onClick => e => {
```
chrisnojima: Why memoize this

songgao: I was thinking to avoid making new functions if onClick from dispatchProps doesn’t change. Though now I’m realizing this doesn’t work since all the onClicks are using the same memoized function. Guess I’ll just get rid of it

shared/common-adapters/list-item2.tsx
```diff
@@ -58,22 +59,15 @@ const HoverBox = Styles.isMobile
...
-  props.type === 'Small'
-    ? props.statusIcon
-      ? styles.iconSmallWithStatusIcon
-      : styles.iconSmall
-    : props.statusIcon
-    ? styles.iconLargeWithStatusIcon
-    : styles.iconLarge
+// Using margin and keeping the icon on the let using absolute is to work around an issue in RN where
+// the nested views won't grow and still retain their widths correctly
+const iconLeftStyle = (props: Props) => (props.statusIcon ? {left: statusIconWidth} : null)
```
chrisnojima: the previous implementation existed so the styles would be static since these item are reused a lot and there are thousands of them being rendered. let's not make this change which makes this style change on every render

songgao:
@chrisnojima ready for another pass. Thanks!

PR #17934, opened 2019-06-14
[Y2K-35] out of space notifications

jzila:


jzila:
@jakob223 this is the refactor PR. The plumbing PR to hook it all up is coming soon. I'm just going to send a status notification as soon as the available bytes fall below the threshold. The JS stuff does the check to see if that deserves a warning.

shared/actions/fs/index.tsx
```diff
@@ -963,42 +963,59 @@ const checkKbfsServerReachabilityIfNeeded = (state, action: ConfigGen.OsNetworkS
...
+    | NotificationsGen.BadgeAppPayload
+    | FsGen.OverallSyncStatusChangedPayload
+    | FsGen.ShowHideDiskSpaceBannerPayload
+  > = [
+    FsGen.createOverallSyncStatusChanged({
+      diskSpaceStatus,
+      progress: Constants.makeSyncingFoldersProgress(action.payload.params.status.prefetchProgress),
+    }),
+  ]
+  // Only notify about the disk space status if it has changed.
```
jakob223: Do actions guaranteedly get run before reducers?

shared/reducers/fs.tsx
```diff
@@ -519,30 +519,19 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
...
-          .set(
-            'diskSpaceBannerHidden',
-            overallSyncStatus.diskSpaceBannerHidden &&
-              overallSyncStatus.diskSpaceStatus === Types.DiskSpaceStatus.Warning
-          )
+          .set('syncingFoldersProgress', action.payload.progress)
+          .set('diskSpaceStatus', action.payload.diskSpaceStatus)
+      )
+    case FsGen.showHideDiskSpaceBanner:
+      return state.update('overallSyncStatus', overallSyncStatus =>
```
jakob223: nit: could do `setIn`

shared/actions/fs/index.tsx
jzila: It shouldn't matter since `state` is immutable point-in-time.

jakob223: If the reducer runs first, then the check of whether the status has changed will be incorrect (think it hasn't changed even though it has).  Or is there not a reducer running on this trigger?

jzila: OK I understand your question now. I believe Sagas are guaranteed to be called with the pre-reduced state.

jzila: Actually, I'm googling around a bit now and it seems you might be correct. But we've built functionality with this assumption before, and now I'm wondering how that functionality works.

jzila: @jakob223 I further thought about it, and I think it's actually fine here. Since the action that triggers this handler doesn't actually change any state in the reducer. The action that changes state is the one triggered on the line above. But that one is independent of the state variable in this saga handler.

PR #17935, opened 2019-06-14
team: support for hidden chain rotations

maxtaco:
- introduce sig3 for chain17 links
- hidden team rotations implemented, only in test thus far
- hidden team rotations can be digested by the slow loader
- one simple test but lots more tests needed


maxtaco:
@mlsteele just FYI, no review necessary right now. needs a bunch of polish, but it's getting there

PR #17945, opened 2019-06-14
Limited text colors

cecileboucheron:
* Updated color names to reflect new color palette
* Texts now use a limited set of colors. @MarcoPolo and @songgao helped me on this ticket and before fixing all the Type issues, we wanted to make sure this is the best way of doing it. @chrisnojima any opinion?
@keybase/react-hackers 
cc @adamjspooner


shared/common-adapters/text.d.ts
```diff
@@ -11,8 +12,35 @@ type Background =
...
+    | 'white'
+    | 'white_75'
+    | 'white_40'
+    | 'brown_75'
+    | 'orange'
+    | 'transparent'
+  >
+>
+
+export type StylesTextCrossPlatform = StylesCrossPlatform & {color?: AllowedColors}
```
songgao: @MarcoPolo particularly this was added because in places this type is passed through props; not sure if that's the best way to do it.

MarcoPolo: Yeah I think that's fine.

MarcoPolo:
A bit more context,

Cecile wanted to limit the colors that could be used for a Text component. Originally the idea was to pass another prop `color` that would control the color for the text, but I think this idea of constraining the color type in styles is better. The type constraint has several benefits:

* It requires no new code. (Code that would read props.color and do some defaulting stuff. and maybe code that warns if you pass a color in style).
* It will tell you if you're doing it wrong at compile time.
* It will make sure that all existing usages abide by the constraint.

cecileboucheron:
@keybase/react-hackers Bump ^

PR #17947, opened 2019-06-14
minor visual tweaks [HOTPOT-149]

joshblum:
before:
Screen Shot 2019-06-14 at 3 03 22 PM
after:
Screen Shot 2019-06-14 at 2 53 58 PM

cc @keybase/hotpotatosquad

### DIFF
--- a/shared/chat/conversation/messages/system-added-to-team/index.tsx
+++ b/shared/chat/conversation/messages/system-added-to-team/index.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 import * as Kb from '../../../../common-adapters'
 import * as Styles from '../../../../styles'
 import UserNotice from '../user-notice'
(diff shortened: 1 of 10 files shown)


chrisnojima:
i don't think we center text like this.
cc: @cecileboucheron

joshblum:
if not we should change some of the other messages: 
image

chrisnojima:
to further clarify. we center text but we don't center bulleted text

joshblum:
@cecileboucheron updated as discussed in chat:
Screen Shot 2019-06-14 at 5 14 23 PM
Screen Shot 2019-06-14 at 5 21 36 PM
image

joshblum:
@cecileboucheron as discussed over chat timestamp is now in a separate component so these message have a consistent font size/top margin

Screen Shot 2019-06-17 at 10 29 06 AM

PR #17948, opened 2019-06-14
Show advanced transaction summaries

mlsteele:
Some funky transactions. 
1. The frontend throwing errors was a good guide while working on this. But I may not have hit every possible configuration. Should I remove the remaining throws from the components so they never crash?
2. The wording like past tense and that large limit number can be massaged by changing the server. That will come in another PR.
3. For trust lines, producing rows that look like [this](https://user-images.githubusercontent.com/705646/59532524-096fec00-8eb7-11e9-81db-74c06f139c05.png) is not what this PR does. That could be its own ticket

image

Sometimes there's a "Sender", sometimes not.

image


patrickxb:
Looks good!

The main thing would be for the details interface to show a line for each operation when there is a multiple operation transaction.  (stellarx "purchases" have about 3 operations in them, if that's helpful)

shared/wallets/transaction-details/index.tsx
```diff
@@ -226,16 +230,19 @@ const propsToParties = (props: NotLoadingProps) => {
...
-    <Counterparty
-      accountID={counterpartyAccountID}
-      counterparty={props.counterparty}
-      counterpartyMeta={props.counterpartyMeta}
-      counterpartyType={props.counterpartyType}
-      onChat={props.onChat}
-      onShowProfile={props.onShowProfile}
-    />
-  )
+  let counterparty = null
```
cjb: Perhaps `const counterparty = props.counterparty && <Counterparty ...>`?

shared/wallets/transaction-details/index.tsx
```diff
@@ -301,19 +310,25 @@ const TransactionDetails = (props: NotLoadingProps) => {
...
         />
       </Kb.Box2>
       <Kb.Divider />
       <Kb.Box2 direction="vertical" gap="small" fullWidth={true} style={styles.container}>
-        <Kb.Box2 direction="vertical" gap="xtiny" fullWidth={true}>
-          <Kb.Text type="BodySmallSemibold">Sender:</Kb.Text>
-          {sender}
-        </Kb.Box2>
+        {sender && (
+          <Kb.Box2 direction="vertical" gap="xtiny" fullWidth={true}>
```
cjb: Sometimes `{!!sender && (` is necessary to avoid a crash on Android, where the failing && returns a '' that's treated as bare text, that might be there case here too.

shared/wallets/transaction-details/index.tsx
```diff
@@ -301,19 +310,25 @@ const TransactionDetails = (props: NotLoadingProps) => {
...
+            <Kb.Text type="BodySmallSemibold">Sender:</Kb.Text>
+            {sender}
+          </Kb.Box2>
+        )}
 
-        <Kb.Box2 direction="vertical" gap="xxtiny" fullWidth={true}>
-          <Kb.Text type="BodySmallSemibold">Recipient:</Kb.Text>
-          {receiver}
-        </Kb.Box2>
+        {receiver && (
```
cjb: Same.

shared/wallets/transaction-details/index.tsx
cjb: Or `const counterparty = props.counterparty ? <Counterparty ...> : null`

shared/wallets/transaction-details/index.tsx
mlsteele: Thanks! Updated but github doesn't seem to understand.

PR #17955, opened 2019-06-15
post location with /location

mmaxim:
Add /location, an interface for posting your current location using the Google Maps API.

1. The technique uses the same strategy as we do for /giphy, where all requests to the third party are routed through Keybase (without Keybase being able to read them) to semi-anonymize the request. 
2. Use the RN API for getting the location (this could possibly use some optimization). 
3. Threads all the information through a generic unfurl, using the coordinate attached to the text message to inform the `Unfurler` that it should extract a special maps URL.
4. `Scraper` knows about the special URL, and will implement a separate scraping path for maps. 
5. The rest of the packager pipeline is more or less the same as for a normal generic unfurl.


mmaxim:
cc @maxtaco @malgorithms @chrisnojima

go/chat/unfurl/scrape_map.go
```diff
@@ -0,0 +1,47 @@
...
+	acc, err := strconv.ParseFloat(sacc, 64)
+	if err != nil {
+		return res, err
+	}
+	mapURL, err := maps.GetMapURL(ctx, s.G().ExternalAPIKeySource, lat, lon)
+	if err != nil {
+		return res, err
+	}
+	linkURL := maps.GetExternalMapURL(ctx, lat, lon)
+	desc := fmt.Sprintf("Accurate to %.2fm.", acc)
```
joshblum: convert to imperial?

go/chat/unfurl/unfurler.go
```diff
@@ -152,6 +152,14 @@ func (u *Unfurler) extractURLs(ctx context.Context, uid gregor1.UID, convID chat
 	}
 	switch typ {
 	case chat1.MessageType_TEXT:
+		// check for coordinate message
+		if body.Text().Coord != nil {
+			return []ExtractorHit{ExtractorHit{
+				URL: fmt.Sprintf("https://%s/?lat=%f&lon=%f&acc=%f", mapsDomain, body.Text().Coord.Lat,
```
joshblum: can this be a const in the maps package?

go/chat/commands/location.go
```diff
@@ -0,0 +1,74 @@
...
+	if !h.Match(ctx, text) {
+		return ErrInvalidCommand
+	}
+	coord, err := h.getChatUI().ChatGetCoordinate(ctx)
+	if err != nil {
+		return err
+	}
+	if _, err := h.G().ChatHelper.SendMsgByIDNonblock(ctx, convID, tlfName,
+		chat1.NewMessageBodyWithText(chat1.MessageText{
+			Body:  "_Sharing my location..._",
```
joshblum: is this hard to remove because you can't send a text with an empty body? could we modify to allow if `Coord != nil`?

shared/ios/Keybase/Info.plist
```diff
@@ -45,10 +45,6 @@
 	<string>Attaching images to chat messages</string>
 	<key>NSCalendarsUsageDescription</key>
 	<string>Keybase does not use this data! Please send feedback if you see this message.</string>
-	<key>NSLocationAlwaysAndWhenInUseUsageDescription</key>
```
buoyad: We need to leave this key in for app store review

shared/actions/chat2/index.tsx
```diff
@@ -2807,6 +2808,26 @@ const onChatMaybeMentionUpdate = (state, action: EngineGen.Chat1ChatUiChatMaybeM
   })
 }
 
+const onChatGetCoordinate = (state, action: EngineGen.Chat1ChatUiChatGetCoordinatePayload) => {
```
buoyad: get `logger` as the third param here

chrisnojima: this should go into actions/platform-specific

shared/actions/chat2/index.tsx
```diff
@@ -2807,6 +2808,26 @@ const onChatMaybeMentionUpdate = (state, action: EngineGen.Chat1ChatUiChatMaybeM
...
+  const response = action.payload.response
+  if (isMobile) {
+    navigator.geolocation.getCurrentPosition(
+      pos =>
+        response.result({accuracy: pos.coords.accuracy, lat: pos.coords.latitude, lon: pos.coords.longitude}),
+      err => logger.warn(err.message),
+      {enableHighAccuracy: true, maximumAge: 0, timeout: 30000}
+    )
+  } else {
+    // desktop doesn't really work, so just stick us in SF
```
buoyad: `doesn't really work and is disabled in the service`

shared/actions/chat2/index.tsx
```diff
@@ -2807,6 +2808,26 @@ const onChatMaybeMentionUpdate = (state, action: EngineGen.Chat1ChatUiChatMaybeM
...
+    response.result({
```
chrisnojima: its possible to make this work, are we going to skip supporting it?

go/chat/externalapikeys.go
```diff
@@ -33,6 +33,13 @@ func (r *RemoteExternalAPIKeySource) GetKey(ctx context.Context, typ chat1.Exter
 	if len(keys) != 1 {
 		return res, fmt.Errorf("wrong number of keys returned: %d", len(keys))
 	}
+	rtyp, err := keys[0].Typ()
```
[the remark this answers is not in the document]
mmaxim: In a followup PR we can do that

go/chat/unfurl/scrape_map.go
mmaxim: I'll just make an int.

go/chat/commands/location.go
mmaxim: Discussed offline

shared/actions/chat2/index.tsx
mmaxim: Not sure we need to do that, the API is the same on both platforms, they made a conscious decision to do that on the RN side it seems.

shared/actions/chat2/index.tsx
mmaxim: Might be possible, but I hit a bunch of snags so just decided against it for this PR.

mmaxim:
CI actually passed here I think

PR #17956, opened 2019-06-15
(WIP) Body 1px bigger

cecileboucheron:
@adamjspooner


adamjspooner:
I'm not sure we'll just be able to increase the font size. I used this branch for an hour this morning, and the spacing (padding/margin) on messages, channel names, inputs, etc. feel incorrectly spaced. I prefer the larger font, though.

PR #17960, opened 2019-06-16
Trustline modal component

songgao:
image

image

cc @keybase/picnicsquad

### DIFF
--- a/shared/actions/json/wallets.json
+++ b/shared/actions/json/wallets.json
@@ -425,6 +425,10 @@
       "airdropState": "Types.AirdropState"
     },
     "updateAirdropDetails": {},
-    "updatedAirdropDetails": {"details": "Types.AirdropDetails"}
+    "updatedAirdropDetails": {"details": "Types.AirdropDetails"},
+    "setTrustlineExpanded": {
+      "expanded": "boolean",
+      "trustlineAssetID": "Types.TrustlineAssetID"
+    }
   }
 }
--- a/shared/constants/types/wallets.tsx
+++ b/shared/constants/types/wallets.tsx
@@ -278,6 +278,27 @@ export type _AirdropDetails = {
 }
 export type AirdropDetails = I.RecordOf<_AirdropDetails>
 
+export type TrustlineAssetID = string
+export const makeTrustlineAssetID = (issuerAccountID: string, assetCode: string): TrustlineAssetID =>
+  `${issuerAccountID}-${assetCode}`
+
+export type _TrustlineAsset = {
+  assetCode: string
+  issuerAccountID: string
+  issuerVerifiedDomain: string
+  trustedLimit: number // 0 if not trusted
+}
+export type TrustlineAsset = I.RecordOf<_TrustlineAsset>
+
+export type _Trustline = {
+  acceptedAssets: I.List<TrustlineAssetID>
+  assetMap: I.Map<TrustlineAssetID, TrustlineAsset>
+  expandedAssets: I.Set<TrustlineAssetID>
+  popularAssets: I.List<TrustlineAssetID>
+  searchingAssetsHit: I.List<TrustlineAssetID>
+}
+export type Trustline = I.RecordOf<_Trustline>
+
 export type _State = {
   acceptedDisclaimer: boolean
   acceptingDisclaimerDelay: boolean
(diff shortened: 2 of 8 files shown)


shared/constants/types/wallets.tsx
```diff
@@ -275,6 +275,27 @@ export type _AirdropDetails = {
...
+export type TrustlineAsset = I.RecordOf<_TrustlineAsset>
+
+export type _Trustline = {
+  acceptedAssets: I.List<TrustlineAssetID>
+  assetMap: I.Map<TrustlineAssetID, TrustlineAsset>
+  expandedAssets: I.Set<TrustlineAssetID>
+  popularAssets: I.List<TrustlineAssetID>
+  searchingAssetsHit: I.List<TrustlineAssetID>
+}
+export type Trustline = I.RecordOf<_Trustline>
```
songgao: I'm changing these structures in my next PR, so might as well skip these for now and continue to the actual components.

shared/actions/json/wallets.json
```diff
@@ -425,6 +425,10 @@
       "airdropState": "Types.AirdropState"
     },
     "updateAirdropDetails": {},
-    "updatedAirdropDetails": {"details": "Types.AirdropDetails"}
+    "updatedAirdropDetails": {"details": "Types.AirdropDetails"},
+    "setTrustlineExpanded": {
```
chrisnojima: i think this kind of thing should likely be kept as local state in the modal. it doesn't need to be in the store i think. can leave it for now but in general

shared/constants/types/wallets.tsx
```diff
@@ -275,6 +275,27 @@ export type _AirdropDetails = {
...
 
+export type TrustlineAssetID = string
+export const makeTrustlineAssetID = (issuerAccountID: string, assetCode: string): TrustlineAssetID =>
+  `${issuerAccountID}-${assetCode}`
+
+export type _TrustlineAsset = {
+  assetCode: string
+  issuerAccountID: string
+  issuerVerifiedDomain: string
+  trustedLimit: number // 0 if not trusted
```
chrisnojima: what does this mean? it's just treated as a boolean

shared/wallets/trustline/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+import * as React from 'react'
+import * as Styles from '../../styles'
+import * as Kb from '../../common-adapters'
+import * as Types from '../../constants/types/wallets'
+import Asset from './asset-container'
+
+type BodyProps = {
+  totalAssetsCount: number
+  errorMessage?: string
+  acceptedAssets: I.List<Types.TrustlineAssetID>
```
chrisnojima: we typically avoid passing immutable structure to the components

shared/wallets/trustline/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+      sections={makeSections(props)}
+      renderItem={({index, item}) => <Asset firstItem={index === 0} trustlineAssetID={item} />}
+      renderSectionHeader={({section}) => sectionHeader(section)}
+    />
+  </Kb.Box2>
+)
+
+const TrustlineDesktop = Kb.HeaderOrPopup((props: Props) => {
+  const {onDone, ...rest} = props
+  const bodyProps = rest as BodyProps
```
chrisnojima: in new code we should avoid `as` casts if we can

shared/wallets/trustline/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+      const {onDone} = props
+      return <TrustlineDesktop onCancel={onDone} {...props} />
+    }
+
+export default Trustline
+
+const styles = Styles.styleSheetCreate({
+  body: {
+    ...Styles.globalStyles.flexGrow,
+    // backgroundColor: Styles.globalColors.red,
```
chrisnojima: 🔥

shared/constants/types/wallets.tsx
songgao: We have a number from the service, where user can specify a maximum amount to trust. It's not in current design -- we either trust it to maximum amount or don't trust it. But since the data is already there I thought I'd plumb it through anyway in case we want to display it somewhere in the future.

shared/wallets/trustline/index.tsx
songgao: Cool; I'll address this in the next PR https://github.com/keybase/client/pull/17992

songgao:
bump ⪮

PR #17975, opened 2019-06-18
PGP Push secret when local secret is already present

zapu:
This PR changes `PGPKeyImportEngine` to be able to only push secret key through `key/add` endpoint when public key is already present in the sigchain.

This fixes the following use cases:

**Use case A:**
1) User imports their key using `keybase pgp select`, with or without importing private key to local keyring.
2) User imports their secret key using `keybase pgp import --push-secret` with the intention for the secret key to be pushed to Keybase servers.

(previously it would have been not pushed to keybase servers, secret key would only be available locally).

**Use case B:**
1) User imports their key using `keybase pgp import` with the intention for the secret key to only be available locally.
2) User changes their mind and uses `keybase pgp import --push-secret` to also push the secret key to the servers.

(previously would fail with "This key is already active. You can't add it twice." error)

There is a new test for use case A in `TestPGPSelectThenPushSecret` and new test for B in `TestPGPImportLocalPrivateThenServer`.

### DIFF
--- a/go/engine/pgp_import_key_test.go
+++ b/go/engine/pgp_import_key_test.go
@@ -89,8 +89,8 @@ func TestPGPImportAndExport(t *testing.T) {
 	}
 }
 
-// TestPGPImportPrivImport - import the same key twice, only importing the private key
-// the second time / on the second device (CORE-10562).
+// TestPGPImportPrivImport - import the same key twice, only importing the
+// private key the second time / on the second device (CORE-10562).
 func TestPGPImportPrivImport(t *testing.T) {
 	user, dev1, dev2, cleanup := SetupTwoDevices(t, "login")
 	defer cleanup()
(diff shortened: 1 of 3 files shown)


go/engine/pgp_import_key_test.go
```diff
@@ -188,6 +188,40 @@ func TestPGPImportPrivImport(t *testing.T) {
...
+	require.NoError(t, err)
+	mctx = NewMetaContextForTest(tc).WithUIs(uis)
+	err = RunEngine2(mctx, eng)
+	require.NoError(t, err)
+
+	ss, err := mctx.ActiveDevice().SyncSecretsForce(mctx)
+	require.NoError(t, err)
+	privKey, ok := ss.FindPrivateKey(kid.String())
+	require.True(t, ok)
+	require.NotEmpty(t, privKey.Bundle)
```
zapu: i don't know if this is a good check - secret syncer here seems to only use server side keys using `key/fetch_private` but maybe i missed something.

PR #17977, opened 2019-06-18
Add HasRandomPW prefetching in login hooks

zapu:


go/libkb/user.go
```diff
@@ -917,6 +922,10 @@ func LoadHasRandomPw(mctx MetaContext, arg keybase1.LoadHasRandomPwArg) (res boo
 		return res, err
 	}
 
+	if arg.ForUID != nil && !arg.ForUID.Equal(currentUID) {
+		return res, errors.New("current user changed")
```
zapu: maybe all this is not needed anymore if the context is cancelled after logout

mmaxim: Why even do this comparison? Just use the UID in the param.

go/service/main.go
```diff
@@ -926,6 +931,11 @@ func (d *Service) OnLogout(m libkb.MetaContext) (err error) {
 		d.walletState.Reset(m)
 	}
 
+	log("cancelling random pw prefetch")
+	if d.hasRPWCancelFn != nil {
```
mmaxim: Can probably just use a logout hook on G for this.

go/libkb/user.go
zapu: the idea was that the uid might change while the https request is still in flight. but maybe this is not needed since we are cancelling the context. not sure if onLogout fires during user switching though, i need to find out

zapu: @mmaxim so after remove all that `ForUID` stuff, will it be good to go?

mmaxim: I think you can also simplify the logout hook stuff per my other comment.

zapu:
@mmaxim another take? I added a type that implements hooks and call `g.AddLoginHook` `g.AddLogoutHook` to wire it up

zapu:
@mmaxim ping

PR #17998, opened 2019-06-19
fix bug in shutting down old payment-build-requests

xgess:
this was cancelling the newest request instead of the oldest.


xgess:
@mlsteele if you get a few minutes, do you mind trying to recreate your bug on this branch? i think i might have fixed a different one from what you actually reported.

mlsteele:
The bug where "Change" gets stuck as a spinner still repros for me on this branch.

This does look like a legit fix though!

By design, `maxConcurrentBuilds` should _not_ be getting hit. There should not be more than 1 active bid. Were you seeing many bids? If so, I'd be interested in looking into whether something is forgetting to `stopBuildPaymentLocal`.

PR #18001, opened 2019-06-19
Add logging to upgradeable and file secret store

buoyad:
cc @keybase/y2ksquad

### DIFF
--- a/go/libkb/secret_store_file.go
+++ b/go/libkb/secret_store_file.go
@@ -30,8 +30,10 @@ func NewSecretStoreFile(dir string) *SecretStoreFile {
 	return &SecretStoreFile{dir: dir}
 }
 
-func (s *SecretStoreFile) RetrieveSecret(m MetaContext, username NormalizedUsername) (LKSecFullSecret, error) {
-	secret, err := s.retrieveSecretV2(username)
+func (s *SecretStoreFile) RetrieveSecret(mctx MetaContext, username NormalizedUsername) (secret LKSecFullSecret, err error) {
+	defer mctx.TraceTimed(fmt.Sprintf("SecretStoreFile.RetrieveSecret(%s)", username), func() error { return err })()
+	mctx.Debug("Retrieving secret V2 from file")
+	secret, err = s.retrieveSecretV2(username)
 	switch err.(type) {
 	case nil:
 		return secret, nil
@@ -41,16 +43,17 @@ func (s *SecretStoreFile) RetrieveSecret(m MetaContext, username NormalizedUsern
 	}
 
 	// check for v1
+	mctx.Debug("Retrieving secret V1 from file")
 	secret, err = s.retrieveSecretV1(username)
 	if err != nil {
 		return LKSecFullSecret{}, err
 	}
 
 	// upgrade to v2
-	if err := s.StoreSecret(m, username, secret); err != nil {
+	if err = s.StoreSecret(mctx, username, secret); err != nil {
 		return secret, err
 	}
-	if err := s.clearSecretV1(username); err != nil {
+	if err = s.clearSecretV1(username); err != nil {
 		return secret, err
 	}
 	return secret, nil
@@ -100,7 +103,8 @@ func (s *SecretStoreFile) retrieveSecretV2(username NormalizedUsername) (LKSecFu
 	return newLKSecFullSecretFromBytes(secret)
 }
 
-func (s *SecretStoreFile) StoreSecret(m MetaContext, username NormalizedUsername, secret LKSecFullSecret) error {
+func (s *SecretStoreFile) StoreSecret(mctx MetaContext, username NormalizedUsername, secret LKSecFullSecret) (err error) {
+	defer mctx.TraceTimed(fmt.Sprintf("SecretStoreFile.StoreSecret(%s)", username), func() error { return err })()
 	noise, err := MakeNoise()
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


go/libkb/secret_store_file.go
```diff
@@ -198,22 +202,22 @@ func (s *SecretStoreFile) StoreSecret(m MetaContext, username NormalizedUsername
...
 
 	errV1 := s.clearSecretV1(username)
 	errV2 := s.clearSecretV2(username)
 
-	if errV1 != nil {
-		return errV1
-	}
-	if errV2 != nil {
-		return errV2
+	err = CombineErrors(errV1, errV2)
```
heronhaye: :+1:

go/libkb/secret_store_linux.go
```diff
@@ -31,5 +31,5 @@ func NewSecretStoreAll(mctx MetaContext) SecretStoreAll {
 		// Use system keychain but fall back to file store if not available.
 		return SecretStoreFallbackBehaviorOnError
 	}
-	return NewSecretStoreUpgradeable(ssecretservice, sfile, shouldUpgradeOpportunistically, shouldStoreInFallback)
+	return NewSecretStoreUpgradeable(ssecretservice, sfile, "linux keyring secret store", "file-based secret store", shouldUpgradeOpportunistically, shouldStoreInFallback)
```
heronhaye:
```suggestion
	return NewSecretStoreUpgradeable(ssecretservice, sfile, "system keyring", "file-based secret store (see https://keybase.io/docs/crypto/local-key-security)", shouldUpgradeOpportunistically, shouldStoreInFallback)
```

go/libkb/secret_store_upgradeable.go
```diff
@@ -86,29 +90,29 @@ func (s *SecretStoreUpgradeable) StoreSecret(mctx MetaContext, username Normaliz
...
 		} else {
-			mctx.Debug("Failed to clear secret for %s from secretstore b: %s", username, clearBErr)
+			mctx.Debug("Failed to clear secret for %s from secondary store (%s): %s", username, s.labelB, clearBErr)
 		}
 		return nil
 	}
 
 	if fallbackBehavior == SecretStoreFallbackBehaviorNever {
-		mctx.Warning("Failed to reach system keyring (store A: %s), not falling back to store B because of fallback behavior.", err1)
+		mctx.Warning("Failed to reach system keyring (primary store (%s): %s), not falling back to store B because of fallback behavior.", s.labelA, err1)
```
heronhaye: no label B here?

go/libkb/secret_store_upgradeable.go
```diff
@@ -86,29 +90,29 @@ func (s *SecretStoreUpgradeable) StoreSecret(mctx MetaContext, username Normaliz
...
 		return err1
 	}
 
-	mctx.Warning("Failed to reach system keyring (store A: %s), falling back to file-based secret store (store B).", err1)
+	mctx.Warning("Failed to reach system keyring (primary store (%s): %s), falling back to file-based secret store (store B).", s.labelA, err1)
```
heronhaye: no label B here?

buoyad:
cc @zapu how does this look? want to add anything to the logs?

go/libkb/secret_store_file_test.go
```diff
@@ -42,6 +42,10 @@ func testSSDir(t *testing.T) (string, func()) {
 }
 
 func TestSecretStoreFileRetrieveSecret(t *testing.T) {
+	tc := SetupTest(t, "SecretStoreFile", 1)
+	defer tc.Cleanup()
+	m := NewMetaContextForTest(tc)
```
zapu: i don't think entire test structure needs to be initialized for this test. or did it not work without it?

go/libkb/secret_store_linux.go
```diff
@@ -31,5 +31,5 @@ func NewSecretStoreAll(mctx MetaContext) SecretStoreAll {
 		// Use system keychain but fall back to file store if not available.
 		return SecretStoreFallbackBehaviorOnError
 	}
-	return NewSecretStoreUpgradeable(ssecretservice, sfile, shouldUpgradeOpportunistically, shouldStoreInFallback)
+	return NewSecretStoreUpgradeable(ssecretservice, sfile, "system keyring", "file-based secret store (see https://keybase.io/docs/crypto/local-key-security)", shouldUpgradeOpportunistically, shouldStoreInFallback)
```
zapu: do we need the keybase.io url here? does user see this somewhere?

go/libkb/secret_store_file_test.go
buoyad: It didn't work because of the added `TraceTimed`s, they need G

go/libkb/secret_store_linux.go
buoyad: I'm not sure, @heronhaye suggested this

go/libkb/secret_store_file_test.go
zapu: got it, thanks

go/libkb/secret_store_linux.go
heronhaye: user sees it in warning when during fallback on signup

zapu: wouldn't this also cause long and less readable long lines?

heronhaye: Maybe. Do you have a suggestion on how to fix it? We could pass in a third "info string" to the struct that is only printed in user-facing logs. Imo it's fine as is, up to you

zapu: let's go with this and figure it out later.

I wonder if typical user even sees it, this does not plumb up to the gui, right?

Or is it more for the cli users?

heronhaye: For CLI users. Good point about GUI, idk.

PR #17982, opened 2019-06-18
react redux 7.1 (hooks)

chrisnojima:
- [x] Updates react-redux to 7.1 which enables hooks
- [x] Ported over device page to use hooks
- [x] Hooks usage requires us changing how we mock out things in stories. Before we patched how connect worked but now we just make a fake store
- [x] Example mock for common things. In order for this to work with the new style and old style we entirely merge the old PropProvider keys and the real store shape

PR #18004, opened 2019-06-19
Add Stellar education modal directly in Keybase app

nathunsmitty:
Adds a popup with information about Stellar when a user clicks on one of the various "What is Stellar?" buttons through the app. (Previously we linked to https://keybase.io/what-is-stellar.) Note that the mockups originally included a couple of videos about Stellar, but since we don't have those yet, they aren't included. 

Screenshot of the page on desktop:
Screen Shot 2019-06-19 at 2 46 10 PM

And mobile:
Simulator Screen Shot - iPhone Xʀ - 2019-06-19 at 16 25 04

Locations the "What is Stellar?" button appears:
- Wallet List on desktop
- Wallet Switcher on mobile
- Clicking on a stellar address assertion on the profile page (desktop or mobile) or tracker (desktop only)

I also added storybook entry for this new page. (Links – note you'll need to be checked out on this branch and have the appropriate storybook server running: [🖥](http://localhost:6006/?path=/story/wallets--what-is-stellar-modal) , [📱](http://localhost:7007/?selectedKind=Wallets&selectedStory=What%20is%20Stellar%3F%20Modal&full=0&addons=1&stories=1&panelRight=0)) 

I'm a little rusty on how we do routing (I mostly copied code from other components that were already routed), so I'd especially appreciate feedback on that.


shared/wallets/wallet-list/container.tsx
```diff
@@ -35,7 +35,9 @@ const mapDispatchToProps = (dispatch: Container.TypedDispatch) => ({
       RouteTreeGen.createNavigateAppend({path: [{props: {showOnCreation: true}, selected: 'linkExisting'}]})
     )
   },
-  onWhatIsStellar: () => openURL('https://keybase.io/what-is-stellar'),
+  onWhatIsStellar: () => {
```
cjb: 🐑 can remove braces around these

nathunsmitty:
Potential follow up ticket: we could probably remove the https://keybase.io/what-is-stellar URL [from here](https://github.com/keybase/keybase/blob/master/config/static.iced#L43).

nathunsmitty:
Updated this component to use react-redux's `useDispatch`, made available by #17982. Since we don't need any props from the state for `WhatIsStellarModal`, I don't think we need to mock the state at all, but maybe @chrisnojima can confirm this?

PR #18005, opened 2019-06-19
populate frequently used emojis from the service [HOTPOT-130]

joshblum:
use the service backed `topEmojis` to populate the frequently used emojis on desktop and mobile

cc @keybase/hotpotatosquad


shared/chat/conversation/messages/react-button/emoji-picker/container.tsx
```diff
@@ -26,7 +28,9 @@ type WrapperState = {
   width: number
 }
 
-const mapStateToProps = () => ({})
+const mapStateToProps = (state, ownProps: OwnProps) => ({
+  topReacjis: state.config.userReacjis ? state.config.userReacjis.topReacjis : defaultTopReacjis,
```
chrisnojima: just default the store value to this

shared/chat/conversation/messages/react-button/emoji-picker/index.tsx
```diff
@@ -7,11 +7,22 @@ import {chunk} from 'lodash-es'
 import {memoize} from '../../../../../util/memoize'
 
 // defer loading this until we need to, very expensive
-const getData = memoize(() => {
+const getData = memoize((topReacjis: Array<string>) => {
```
chrisnojima: i don't think you want to memoize these two things together. keep them as 2 different things

shared/reducers/config.tsx
```diff
@@ -1,5 +1,7 @@
+import {store} from 'emoji-mart'
```
chrisnojima: i don't think this belongs in the reducer. better as a side effect. also does emoji-mart work on mobile?

joshblum: nope will short circuit on mobile

joshblum:
@chrisnojima this is ready

shared/actions/config/index.tsx
```diff
@@ -500,6 +502,37 @@ function* criticalOutOfDateCheck() {
   }
 }
 
+// onUpdateUserReacjis hooks `userReacjis`, frequently used reactions
```
chrisnojima: actually maybe this whole thing should be in chat and not config. We get a static chat config that maybe should hold this data

chrisnojima: cc: @mmaxim

joshblum: moved `userReacjis` over to the chat store, @mmaxim said the static config isn't the right place so chained to the bootstrap action

PR #18007, opened 2019-06-19
teams: upgrade FTL storage from 10.0 to 10.1

maxtaco:
- fill in seedchecks when loading in an old object; fix broken sends on master


maxtaco:
not waiting for CI

PR #18008, opened 2019-06-19
make indexer suspendable

joshblum:
suspends the indexer everywhere we suspend the convloader

cc @keybase/hotpotatosquad

### DIFF
--- a/go/chat/server.go
+++ b/go/chat/server.go
@@ -173,6 +173,10 @@ func (h *Server) suspendInboxSource(ctx context.Context) func() {
 	return utils.SuspendComponent(ctx, h.G(), h.G().InboxSource)
 }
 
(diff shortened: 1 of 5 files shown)


go/chat/server.go
```diff
@@ -2902,6 +2911,7 @@ func (h *Server) LoadGallery(ctx context.Context, arg chat1.LoadGalleryArg) (res
 	defer func() { err = h.squashSquashableErrors(err) }()
 	defer func() { h.setResultRateLimit(ctx, &res) }()
 	defer h.suspendConvLoader(ctx)()
+	defer h.suspendSearchIndexer(ctx)()
```
mmaxim: Can this just be made one function?

joshblum: yeah good call

joshblum:
@mmaxim this is ready

PR #18014, opened 2019-06-19
libfuse/libdokan/simplefs: don't log filenames in normal debugging

strib:
(Depends on #17979 -- review that one first.)

Either obfuscate filenames when using `CDebugf`, or change those statements that do log plaintext filenames to use the vlogger at a level of vlog1.

Also no longer require mobile to use vlog 1, to avoid logging these filenames by default.

Issue: HOTPOT-62

### DIFF
--- a/go/kbfs/libdokan/folderlist.go
+++ b/go/kbfs/libdokan/folderlist.go
@@ -151,7 +151,9 @@ func (fl *FolderList) open(ctx context.Context, oc *openContext, path []string) 
 				ctx, libkb.VLog1, "FL Lookup set alias: %q -> %q",
 				name, aliasTarget)
 			if !fl.isValidAliasTarget(ctx, aliasTarget) {
-				fl.fs.log.CDebugf(ctx, "FL Refusing alias to non-valid target %q", aliasTarget)
+				fl.fs.vlog.CLogf(
+					ctx, libkb.VLog1,
+					"FL Refusing alias to non-valid target %q", aliasTarget)
 				return nil, 0, dokan.ErrObjectNameNotFound
 			}
 			fl.mu.Lock()
@@ -168,7 +170,8 @@ func (fl *FolderList) open(ctx context.Context, oc *openContext, path []string) 
 			path[0] = aliasTarget
 			continue
 
-		case idutil.NoSuchNameError, idutil.BadTLFNameError:
+		case idutil.NoSuchNameError, idutil.BadTLFNameError,
+			tlf.NoSuchUserError, idutil.NoSuchUserError:
 			return nil, 0, dokan.ErrObjectNameNotFound
 
 		default:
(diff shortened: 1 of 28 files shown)


go/kbfs/libdokan/folderlist.go
```diff
@@ -151,7 +151,9 @@ func (fl *FolderList) open(ctx context.Context, oc *openContext, path []string)
 				ctx, libkb.VLog1, "FL Lookup set alias: %q -> %q",
 				name, aliasTarget)
 			if !fl.isValidAliasTarget(ctx, aliasTarget) {
-				fl.fs.log.CDebugf(ctx, "FL Refusing alias to non-valid target %q", aliasTarget)
+				fl.fs.vlog.CLogf(
```
jakob223: should we also print a non-vdebug line without the filename here?

go/kbfs/libfuse/dir.go
```diff
@@ -723,13 +727,18 @@ func (d *Dir) Symlink(ctx context.Context, req *fuse.SymlinkRequest) (
...
+	newNameLoggingPPS := d.node.ChildName(req.NewName)
 	ctx = d.folder.fs.config.MaybeStartTrace(ctx, "Dir.Rename",
 		fmt.Sprintf("%s %s -> %s", d.node.GetBasename(),
-			req.OldName, req.NewName))
+			oldNamePPS, newNameLoggingPPS))
 	defer func() { d.folder.fs.config.MaybeFinishTrace(ctx, err) }()
 
 	d.folder.fs.vlog.CLogf(
-		ctx, libkb.VLog1, "Dir Rename %s -> %s", req.OldName, req.NewName)
+		ctx, libkb.VLog1, "Dir Rename %s -> %s", oldNamePPS, newNameLoggingPPS)
```
jakob223: could we adopt a convention of putting a `*` at the end of filenames whose obfuscators might have the wrong conflict number mapping? e.g. here and in the CR code and maybe elsewhere?

go/kbfs/libkbfs/init.go
```diff
@@ -660,11 +660,6 @@ func doInit(
 			return lg
 		}, params.StorageRoot, params.DiskCacheMode, kbCtx)
 	config.SetVLogLevel(kbCtx.GetVDebugSetting())
-	if mode == InitConstrained {
```
jakob223: we technically don't yet have this - do you want to wait to enable this until after the GUI version of this setting is in?

go/kbfs/libdokan/folderlist.go
strib: Nah I think it should be fine.  The `reportErr` function will show the line number of the return, so an error message without any extra info wouldn't help in my opinion.

go/kbfs/libfuse/dir.go
strib: Ah, I just saw that you had the same `*` convention as I suggested in response to your other comment.  Great minds, etc etc.  As before, I'm not too worried about it.  Let's revisit if it becomes a problem and `deobfuscate` isn't as helpful as I think it will be.

go/kbfs/libkbfs/init.go
strib: Meh I say we just go for it.  If that doesn't get in, we can revisit right before the release.

PR #18015, opened 2019-06-20
support /keybase/.kbfs_status in GUI and subscription fixes

songgao:
This turns out to be more than just a markdown change. The http server needs to handle RootFS separately, so I made a `httpRootFS` for that.

When making these changes, I realized that we could potentially be stuck in a state where we think we are subscribed to a TLF and never do another RPC for that particular path, when in fact we are subscribed to a different one. This is because we compare paths for different tags separately, while kbfs daemon keeps a single subscribed TLF for notifications. So instead, add a `lastSubscribedTlf` to track that, and extract all refreshTags related stuff into a separate function for better clarity.

### DIFF
--- a/go/kbfs/libfs/httprootfs.go
+++ b/go/kbfs/libfs/httprootfs.go
@@ -1,0 +1,52 @@
+package libfs
+
+import (
+	"net/http"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+	billy "gopkg.in/src-d/go-billy.v4"
+)
+
+type httpRootFileSystem struct {
+	rfs *RootFS
+}
+
+var _ http.FileSystem = httpRootFileSystem{}
+
+type fileOnly struct {
+	billy.File
+	rfs *RootFS
+}
+
+// Readdir implements the http.File interface.
+func (fo fileOnly) Readdir(count int) ([]os.FileInfo, error) {
+	return nil, errors.New("not supported")
+}
+
+// Stat implements the http.File interface.
+func (fo fileOnly) Stat() (os.FileInfo, error) {
+	return fo.rfs.Stat(fo.File.Name())
+}
+
+func (hrfs httpRootFileSystem) Open(filename string) (entry http.File, err error) {
+	hrfs.rfs.log.CDebugf(hrfs.rfs.ctx, "hfs.Open %s", filename)
+	defer func() {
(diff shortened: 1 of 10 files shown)


shared/fs/browser/rows/load-files-when-needed.tsx
```diff
@@ -89,9 +87,6 @@ class LoadFilesWhenNeeded extends React.PureComponent<Props> {
...
 
-export default namedConnect(
-  mapStateToProps,
-  mapDispatchToProps,
-  mergeProps,
-  'LoadFilesWhenNeeded'
-)(LoadFilesWhenNeeded) as any
+export default namedConnect(mapStateToProps, mapDispatchToProps, mergeProps, 'LoadFilesWhenNeeded')(
+  LoadFilesWhenNeeded
+) as any
```
chrisnojima: if you're in here you might want to fix this

songgao: oh hmm this must be my prettier. I didn't actually touch it. Will try to fix it anyway

songgao: oh wait I did touch this file. 🤦🏻‍♂️

songgao: I'm gonna rebase https://github.com/keybase/client/pull/17684 over this, and fix a Files bug there. That gets rid of this component, so I'll just leave this for now.

go/kbfs/libfs/httprootfs.go
```diff
@@ -0,0 +1,52 @@
+package libfs
```
strib: Needs copyright notice.

go/kbfs/libfs/httprootfs.go
```diff
@@ -0,0 +1,52 @@
...
+	return nil, errors.New("not supported")
+}
+
+// Stat implements the http.File interface.
+func (fo fileOnly) Stat() (os.FileInfo, error) {
+	return fo.rfs.Stat(fo.File.Name())
+}
+
+func (hrfs httpRootFileSystem) Open(filename string) (entry http.File, err error) {
+	hrfs.rfs.log.CDebugf(hrfs.rfs.ctx, "hfs.Open %s", filename)
```
strib: I think we can do without this logging here, and then you can get rid of the new `ctx` in `RootFS`.  I would probably just need to re-do it anyway for my obfuscation PR, and I don't think it adds much.  But if you really want it for some reason, we can leave it.

PR #18016, opened 2019-06-20
live location sharing

mmaxim:
Introduce live location sharing. The basic flow is that we keep location up to date by continually deleting and re-posting the unfurl with the location data in it. We do this update on an interval to make sure we are not flashing the map (happens in the car). A map for both current position, and the history of positions during the session is shown. cc @keybase/hotpotatosquad 

1. Add `LiveLocationTracker`, which handles all requests for locations and tracks the coordinates for each request. Interacts with `Unfurler` to actually generate the location updates through encoding a key into the URL we unfurl.
2. Change the "current position" posting to use the live location tracker on a 4 second interval. This avoids the case where we post an old coordinate.
3. Add more code to the mapping routines for more flexible rendering, and the ability to render a dual map for current position and history.
4. Keep the app running in the background if currently tracking location.


go/chat/unfurl/scrape_map.go
```diff
@@ -31,17 +33,32 @@ func (s *Scraper) scrapeMap(ctx context.Context, uri string) (res chat1.UnfurlRa
...
-		ImageUrl:    &mapURL,
-		Description: &desc,
+	if len(skey) > 0 {
+		key := types.LiveLocationKey(skey)
+		coords := s.G().LiveLocationTracker.GetCoordinates(ctx, key)
+		liveMapURL = new(string)
+		if *liveMapURL, err = maps.GetLiveMapURL(ctx, s.G().ExternalAPIKeySource, coords); err != nil {
+			return res, err
+		}
+		timeStr = fmt.Sprintf("Posted %s.", time.Now().Format("15:04:05 MST"))
```
joshblum: why not `humanize.Time(...`

go/chat/maps/livelocation.go
```diff
@@ -0,0 +1,403 @@
...
+
+func (l *LiveLocationTracker) LocationUpdate(ctx context.Context, coord chat1.Coordinate) {
+	defer l.Trace(ctx, func() error { return nil }, "LocationUpdate")()
+	l.Lock()
+	defer l.Unlock()
+	if l.G().IsMobileAppType() {
+		// if the app is woken up as the result of a location update, and we think we are currently
+		// backgrounded, then go ahead and mark us as background active so that we can get
+		// location updates out
+		l.G().MobileAppState.UpdateWithCheck(keybase1.MobileAppState_BACKGROUNDACTIVE,
```
joshblum: can there be a new `MobileAppState_BACKGROUNDACTIVELOCATION` state? i worry about other background syncing coming online while doing the location updates and draining battery/data

go/chat/livelocation_test.go
```diff
@@ -0,0 +1,263 @@
...
+	clock.Advance(10 * time.Second)
+	checkCoords(t, unfurler, []chat1.Coordinate{coords[2]}, timeout)
+	select {
+	case <-chatUI.clearCh:
+	case <-time.After(timeout):
+		require.Fail(t, "no clear call")
+	}
+}
+
+func TestChatSrvLiveLocation(t *testing.T) {
```
joshblum: can you add a test case with multiple trackers running? maybe be other lingering issues besides calling `close` multiple times

go/chat/maps/livelocation.go
```diff
@@ -0,0 +1,403 @@
...
+	l.restoreLocked(ctx)
+}
+
+func (l *LiveLocationTracker) Stop(ctx context.Context) chan struct{} {
+	defer l.Trace(ctx, func() error { return nil }, "Stop")()
+	l.Lock()
+	defer l.Unlock()
+	ch := make(chan struct{})
+	for _, t := range l.trackers {
+		close(t.stopCh)
```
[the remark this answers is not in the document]
mmaxim: There is one per, why would it panic?

go/chat/maps/livelocation.go
```diff
@@ -0,0 +1,403 @@
...
+	return res
+}
+
+func (l *LiveLocationTracker) StopAllTracking(ctx context.Context) {
+	defer l.Trace(ctx, func() error { return nil }, "StopAllTracking")()
+	l.Lock()
+	defer l.Unlock()
+	for _, t := range l.trackers {
+		delete(l.trackers, t.Key())
+		close(t.stopCh)
```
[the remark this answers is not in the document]
mmaxim: I think there is a misunderstanding here, each `locationTrack` has its own `stopCh` channel.

go/chat/maps/maps.go
```diff
@@ -15,15 +21,82 @@ import (
...
+	key, err := apiKeySource.GetKey(ctx, chat1.ExternalAPIKeyTyp_GOOGLEMAPS)
+	if err != nil {
+		return "", err
+	}
+	pathStr := ""
+	startStr := ""
+	centerStr := ""
+	first := coords[0]
+	last := coords[len(coords)-1]
+	if len(coords) > 1 {
```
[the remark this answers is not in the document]
mmaxim: Sure

go/chat/maps/maps.go
```diff
@@ -15,15 +21,82 @@ import (
...
 
 func GetMapURL(ctx context.Context, apiKeySource types.ExternalAPIKeySource, lat, lon float64) (string, error) {
 	key, err := apiKeySource.GetKey(ctx, chat1.ExternalAPIKeyTyp_GOOGLEMAPS)
 	if err != nil {
 		return "", err
 	}
 	return fmt.Sprintf(
-		"https://%s/maps/api/staticmap?center=%f,%f&markers=color:red%%7C%f,%f&zoom=15&size=320x200&key=%s",
-		MapsProxy, lat, lon, lat, lon, key.Googlemaps()), nil
+		"https://%s/maps/api/staticmap?center=%f,%f&markers=color:red%%7C%f,%f&size=%dx%d&scale=%d&key=%s",
```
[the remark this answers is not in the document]
mmaxim: I like the big marker on the location map.

go/chat/maps/maps.go
```diff
@@ -15,15 +21,82 @@ import (
...
+		for _, c := range coords {
+			pathStr += fmt.Sprintf("|%f,%f", c.Lat, c.Lon)
+		}
+		pathStr += "&"
+		startStr = fmt.Sprintf("markers=color:green%%7Csize:tiny%%7C%f,%f&", first.Lat, first.Lon)
+	} else {
+		centerStr = fmt.Sprintf("center=%f,%f&", last.Lat, last.Lon)
+	}
+	url := fmt.Sprintf(
+		"https://%s/maps/api/staticmap?%s%s%smarkers=color:red%%7Csize:tiny%%7C%f,%f&size=%dx%d&scale=%d&key=%s",
```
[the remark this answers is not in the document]
mmaxim: Maybe, I don't want to move it now though.

go/chat/maps/livelocation.go
mmaxim: Can do that in a subsequent PR if we think it is useful. The main reason we need this is for various components in chat that require the app to be running (not in the background) to do stuff. I'd have to track all that down and pick out only the ones we need for location. I'm not sure that is worth it right now. The biggest battery drain of this feature is keeping the GPS powered on anyway.

go/chat/unfurl/scrape_map.go
mmaxim: Can't be relative since it is part of the unfurl. I have this mostly for debugging now, I think the best play is we will need a custom display for it. I can pick that up later.

go/chat/livelocation_test.go
mmaxim: I tested this manually, so I am pretty sure it works, but I can add a quick test.

joshblum: oops i misread that close logic

go/chat/maps/livelocation.go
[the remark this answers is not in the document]
joshblum: oops misread

go/chat/maps/livelocation.go
joshblum: ok cool we can keep on eye on it

PR #18017, opened 2019-06-20
TODO items for imp tofu

zapu:
This PR adds 5 new todo items: "add phone number", "verify phone number", "verify email", "make email discoverable", and "add email".

To implement these, the following things had to change:
- TODOs now can carry additional information besides todo type. This depends on the todo type. For example, "verify phone number" item will carry the phone number that's to be verified. This data does not come from the server, but is fetched by service when home screen is being prepared. See `go/home/userdata.go`.
- The number of buttons and buton labels vary between the TODOs now. Some also have different styles than the others. Instead of adding more constants in `shared/constants/people.tsx`, `buttons: Array<TaskButton>` was added to `TodoOwnProps`. `todo/container.tsx` will defines buttons for each todo type when defining the todo connectors. Therefore, `todoTypeToDismissable` was redundant and was removed. TODO is dismissable when the connector has a dismiss button.
- `Task` component now has a `subText` (needed for `legacyEmailDiscoverability` todo), and the text itself is a `Kb.Markdown`. This was needed because some TODO items needs the text formatted, e.g. bold phone number or email.

### DIFF
--- a/shared/people/todo/container.tsx
+++ b/shared/people/todo/container.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import {Task} from '.'
+import {Task, TaskButton} from '.'
 import * as PeopleGen from '../../actions/people-gen'
 import * as Types from '../../constants/types/people'
 import * as Tabs from '../../constants/tabs'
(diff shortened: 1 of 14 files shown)


zapu:
@buoyad ready for review

shared/people/todo/container.tsx
```diff
@@ -15,40 +15,74 @@ import openURL from '../../util/open-url'
...
+  subText: string
   todoType: Types.TodoType
+  buttons: Array<TaskButton>
 }
 
 const installLinkURL = 'https://keybase.io/download'
 const onSkipTodo = (type: Types.TodoType, dispatch) => () => dispatch(PeopleGen.createSkipTodo({type}))
 const mapStateToProps = state => ({myUsername: state.config.username || ''})
 
+type TodoButtonOpts = {
```
buoyad: It'd be nice for `makeDefaultButtons` to be less dependent on the structure of `ownProps` and `dispatchProps` below. could these params be separated out?

shared/people/todo/container.tsx
```diff
@@ -140,23 +181,41 @@ const FolderConnector = connect(
...
+      },
+      {
+        label: 'Create a team repo',
+        onClick: () => dispatchProps.onConfirm(true),
+      },
+      {
+        label: 'Later',
+        mode: 'Secondary',
+        onClick: dispatchProps.onDismiss,
+        type: 'Default',
```
buoyad: :fire: (remove throughout this file)

zapu:
@joshblum can you review go code? thanks

go/home/userdata.go
```diff
@@ -0,0 +1,79 @@
...
+	for i := range home.Items {
+		v := &home.Items[i]
+		typ, err := v.Data.T()
+		if err != nil {
+			return err
+		}
+		if typ != keybase1.HomeScreenItemType_TODO {
+			continue
+		}
+		err = getUserDataForItem(mctx, v)
```
joshblum: what happens in the error case here? the `home` endpoint is wrapped in a retry perhaps this should be as well? or skip the item if this errors out?

go/home/userdata.go
```diff
@@ -0,0 +1,79 @@
...
+func getUserDataForItem(mctx libkb.MetaContext, item *keybase1.HomeScreenItem) error {
+	todoItem := item.Data.Todo()
+	// We can't do todoItem.T() here because it would return an error because
+	// variants we are getting from the server are incomplete. We are expected
+	// to fill them here in this function. Also we want to be modifying
+	// existing HomeScreenTodo objects in place, so we can't use accessor
+	// methods.
+	todoType := todoItem.T__
+	switch todoType {
+	case keybase1.HomeScreenTodoType_VERIFY_ALL_PHONE_NUMBER:
```
joshblum: what happens if there is a mismatch between the server sending a TODO but not finding any data to attach to the TODO? Will something else fail with an unexpected nil on the variant? perhaps this function should signal the caller if no data is added so the todo item can be skipped

zapu: you're right, this is not handled gracefully at all.

here is what I can do - allow this to be nil when we can't find data, and have the frontend be able to handle it if it is. Rather than e.g. showing `undefined` in the todo description instead of phone number.

joshblum: why not just hide the TODO?

zapu: Sounds good. But server only gives us two at a time so we will not have a todo item to fall back on if this happens. Is this ok?

zapu: this seems like a bad place to do this stuff to, because this goes to the cache

shared/people/todo/container.tsx
```diff
@@ -86,17 +125,28 @@ const DeviceConnector = connect(
...
-  (s, d, o: TodoOwnProps) => ({...o, ...s, ...d, showSearchBar: true})
+  (_, d, o: TodoOwnProps) => ({
+    ...o,
+    buttons: [
+      {
+        label: 'Follow later',
+        mode: 'Secondary',
+        onClick: d.onDismiss,
+      },
+    ] as Array<TaskButton>,
```
zapu: @buoyad  and follow - removed the "Browse people" button

PR #18019, opened 2019-06-20
add unread payments to newPayments for blue highlighting

xgess:
unread payments (currently the server is only sending a subset of the expected ones) are supposed to be highlighted in blue above the orange line. It looked to me like all the code was there to highlight them, but nothing was setting new payments as actually being `new` so none were highlighted.


xgess:
@keybase/picnicsquad

xgess:
@keybase/react-hackers

PR #18021, opened 2019-06-20
monitor mobile network state

joshblum:


shared/actions/platform-specific/index.native.tsx
```diff
@@ -241,20 +241,29 @@ function* persistRoute(state, action: ConfigGen.PersistRoutePayload) {
...
   const channel = Saga.eventChannel(emitter => {
-    NetInfo.addEventListener('connectionChange', ({type}) => emitter(type === 'none' ? 'offline' : 'online'))
+    NetInfo.addEventListener('connectionChange', ({type}) => emitter(type))
     return () => {}
   }, Saga.buffers.sliding(1))
 
   while (true) {
     const status = yield Saga.take(channel)
-    yield Saga.put(ConfigGen.createOsNetworkStatusChanged({online: status === 'online'}))
+    yield Saga.put(ConfigGen.createOsNetworkStatusChanged({online: status !== 'none'}))
```
chrisnojima: we should just add a parameter here for the type

shared/actions/platform-specific/index.native.tsx
```diff
@@ -241,20 +241,29 @@ function* persistRoute(state, action: ConfigGen.PersistRoutePayload) {
   )
 }
 
+const updateMobileNetState = (type: string) => {
```
chrisnojima: this should just be a chainAction of `osNetworkStatusChanged`

chrisnojima: we prefer side effects to be chained based on actions vs having to explicitly plumb through calls

go/libkb/appstate.go
```diff
@@ -85,6 +85,57 @@ func (a *MobileAppState) StateAndMtime() (keybase1.MobileAppState, *time.Time) {
...
+		ch <- a.state
+	} else {
+		a.updateChs = append(a.updateChs, ch)
+	}
+	return ch
+}
+
+// Update updates the current network state, and notifies any waiting calls
+// from NextUpdate
+func (a *MobileNetState) Update(state keybase1.MobileNetworkState) {
```
mmaxim: This guy should just have a `State` function like the app state thing that inspires it.

PR #18023, opened 2019-06-20
More Comprehensive Proxy Support

ddworken:
This PR makes the proxy support more comprehensive. Specifically, it configures three new parts of the app to use the configured proxy:

1. Raw TCP Sockets
a. We mainly use raw TCP sockets to communicate with gregor and do faster RPC. This PR creates two new functions `ProxyDial` and `ProxyDialTimeout` that replace `net.Dial` and `net.DialTimeout` respectively. 
2. `http.Get`
a. There were four places where we directly called `http.Get` rather than using the standard HTTP client. This PR defines the convenience function `ProxyHTTPGet` that is meant to be used instead of `http.Get`. 
3. Link Unfurling
a. We use [colly](http://github.com/keybase/colly) for unfurling links. This PR configures colly to use a proxy. 

In addition, this PR adds `godoc.org/golang.org/x/net/proxy` as a dependency. `proxy` natively supports socks5 proxies but does not support HTTP connect based proxies. I created `httpConnectProxy` which is a `proxy.Dialer` that supports HTTP connect based proxies (implementing an HTTP proxy dialer is relatively simple, O(100) lines of code). 

After these changes, I was able to run the Keybase app with no internet access other than through the configured proxy and it all appeared to work correctly. I also wrote basic integration tests for the HTTP connect proxy support [here](https://github.com/keybase/client/pull/18023/commits/7b08945298aeace0f8590d32703701cc62192260#diff-e21cf46919f147354a0f19968e9e6ecdR51) (note that they are not included in this PR since they require a running HTTP proxy). 

Tested working via the proxy: 
- [x] View and create KBFS files
- [x] Send and receive chat messages
- [x] Giphy
- [x] Unfurl
- [x] Attachments 
- [x] Send and receive stellar
- [x] View profiles

Note that this PR was blocked on https://github.com/keybase/go-framed-msgpack-rpc/pull/174. `go-framed-msgpack-rpc` did not support proxies so I added a `Dialable` interface that can be used to route it over arbitrary network connections.


ddworken:
@mmaxim @maxtaco Let me know if you want help setting up a test environment on your computer or would like to see it on my computer.

go/chat/attachments/preprocess.go
```diff
@@ -3,6 +3,8 @@ package attachments
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/keybase/client/go/chat/globals"
```
mmaxim: Nit: could you use `goimports` or some other such import adder so these imports end up in the same places as other files?

go/libkb/proxy.go
```diff
@@ -95,3 +98,200 @@ func MakeProxy(e *Env) func(r *http.Request) (*url.URL, error) {
...
+	}
+	dialer, err := proxy.FromURL(proxyURL, getUpstreamDialer(proxyURL, opts))
+	if err != nil {
+		return nil, err
+	}
+
+	// Currently proxy.Dialer does not support DialContext. This is being actively worked on and will probably
+	// land in the next release, but for now we are just dropping the context on the floor
+	// See: https://github.com/golang/go/issues/17759
+	return dialer.Dial(network, address)
```
mmaxim: Could we simulate context here with a goroutine that checks for canceled and timeout? That gets used a lot. Something like:
```
doneCh := make(chan net.Conn, 1)
go func() {
  doneCh <- dialer.Dial(network, address)
}()
  select {
    case <-ctx.Done():
        return nil, ctx.Err()
    case ret := <-doneCh:
        return ret, nil
    }
```

mmaxim:
One other comment

go/libkb/proxy.go
ddworken: Ah, clever! I've never seen that pattern before though it definitely works (with the caveat of leaking a goroutine until `dialer.Dial` times out). Tweaked slightly to handle errors but updated.

PR #18028, opened 2019-06-20
[Y2K-150] Improve follow TODO layout

jzila:
cc @keybase/y2ksquad 

Screen Shot 2019-06-20 at 1 28 25 PM

### DIFF
--- a/shared/people/todo/index.tsx
+++ b/shared/people/todo/index.tsx
@@ -6,7 +6,7 @@ import * as Styles from '../../styles'
 
 export type Props = {
   badged: boolean
-  icon: IconType
+  icon: Kb.IconType
   instructions: string
   confirmLabel: string
   dismissable: boolean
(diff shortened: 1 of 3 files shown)


shared/people/todo/index.tsx
```diff
@@ -16,27 +16,25 @@ export type Props = {
...
           label={props.confirmLabel}
           onClick={props.onConfirm}
           style={{marginRight: Styles.globalMargins.small}}
         />
       )}
       {props.dismissable && (
-        <Text type="BodyPrimaryLink" onClick={props.onDismiss}>
-          Later
-        </Text>
+        <Kb.Button small={true} label="Follow later" mode="Secondary" onClick={props.onDismiss} />
```
zapu: did this change every todo's dismiss label to "Follow later" ?

jzila: Hmm that's a good point. Also, WDYT about it being a button instead of a label @cecileboucheron?

jzila: I'll make a fix to the follow text at least

PR #18030, opened 2019-06-20
data: a couple enhancements to `PathPartString`

strib:
1. `PathPartString` should still obfuscate fileinfo paths
2. Don't obfuscate over conflict suffixes, since it's nice to be able to see in the logs which conflict file corresponds to which actual file.

### DIFF
--- a/go/kbfs/data/path_part_string.go
+++ b/go/kbfs/data/path_part_string.go
@@ -14,6 +14,7 @@ import (
 const (
 	prefixToSkipObfsucation = ".kbfs_"
 	prefixFileInfo          = ".kbfs_fileinfo_"
+	suffixConflictStart     = ".conflicted ("
 	tarGzSuffix             = ".tar.gz"
 
 	// extRestoreMaxLen specifies the max length of an extension, such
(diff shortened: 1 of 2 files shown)


go/kbfs/data/path_part_string.go
```diff
@@ -48,21 +52,48 @@ func NewPathPartString(
...
 
 func (pps PathPartString) String() string {
-	if strings.HasPrefix(pps.plaintext, prefixToSkipObfsucation) ||
-		pps.obfuscator == nil {
+	if pps.obfuscator == nil {
+		return pps.plaintext
+	}
+
+	var prefix string
+	if strings.HasPrefix(pps.plaintext, prefixFileInfo) {
```
jakob223: I think it's confusing to take off the prefix and store the rest up on line 69, but to seemingly discard it and then check for it again down here. Unless you think it's important to save the space the prefix would take up, I'd advocate for keeping the prefix in the `PathPartString`.

strib: Yeah checking for the prefix again was partly for saving that space, but also it's somewhat required to check for it again explicitly since we have to distinguish it from the regular `.kbfs_` case. But i will make that into a `fileInfo` bool instead, and set the obfuscator to `nil` when it's a regular `kbfs_` prefix, to save some CPU on each call to `String()`.

PR #18032, opened 2019-06-21
[Y2K-148] syncing folder dark background

jzila:
cc @keybase/y2ksquad 

Before:

Screen Shot 2019-06-20 at 5 36 39 PM

After:
Screen Shot 2019-06-20 at 5 36 11 PM

### DIFF
--- a/shared/fs/common/pie-slice.tsx
+++ b/shared/fs/common/pie-slice.tsx
@@ -5,6 +5,7 @@ import * as Styles from '../../styles/index'
 type Props = {
   degrees: number
   animated?: boolean
+  style?: Styles.StylesCrossPlatform
 }
 
 const PieSliceDefault = (props: Props) => {
@@ -12,7 +13,7 @@ const PieSliceDefault = (props: Props) => {
     ? {transform: [{rotate: props.degrees + 'deg'}]}
     : {transform: 'rotate(' + props.degrees + 'deg)'}
   return (
-    <Kb.Box style={styles.container}>
+    <Kb.Box style={Styles.collapseStyles([styles.container, ...(props.style ? [props.style] : [])])}>
       <Kb.Box style={styles.wholeGrey} />
       <Kb.Box style={Styles.collapseStyles([styles.rotateContainer, styleRotate])}>
         <Kb.Box style={styles.leftBlue} />
@@ -25,10 +26,10 @@ const PieSliceDefault = (props: Props) => {
 const PieSlice = (props: Props) => {
   return props.animated ? (
     <Kb.Animated to={{degrees: props.degrees}}>
-      {({degrees}: Props) => <PieSliceDefault degrees={degrees} />}
+      {({degrees}: Props) => <PieSliceDefault degrees={degrees} style={props.style} />}
     </Kb.Animated>
   ) : (
-    <PieSliceDefault degrees={props.degrees} />
+    <PieSliceDefault degrees={props.degrees} style={props.style} />
   )
 }
 const pieSize = Styles.isMobile ? 16 : 12
--- a/shared/fs/common/pie-slice.tsx
+++ b/shared/fs/common/pie-slice.tsx
@@ -5,20 +5,27 @@ import * as Styles from '../../styles/index'
 type Props = {
   degrees: number
   animated?: boolean
+  negative?: boolean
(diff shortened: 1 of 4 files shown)


shared/fs/common/pie-slice.tsx
```diff
@@ -76,12 +87,15 @@ const styles = Styles.styleSheetCreate({
     ...stylePieWhole,
     left: 0,
   },
-  wholeGrey: {
-    ...stylePieWhole,
+  unfilledNegative: {
+    backgroundColor: Styles.globalColors.blueDark,
```
jakob223: I think this blue might be a little too close to the filled blue when it's empty, especially on a non-blue background (there are red and green too). Cécile suggested `black_05` or similar.

shared/fs/common/pie-slice.tsx
```diff
@@ -73,10 +73,8 @@ const styles = Styles.styleSheetCreate({
   },
   leftUnfilled: {
     ...stylePieHalf,
-    // -1 is a workaround for a rendering issue where the filled part of the
```
jakob223: I still prefer the slightly rounded look to the bleeding-through look, especially since how bad the bleeding is depends on your resolution. In your screenshot it's not too bad:
image

but still I don't like it. I'd rather keep the rounded rectangles and hope that they're not noticeable when the background is correct.

jzila: The rectangles look pretty awful. Much more noticeable than the slight spillage. But I will investigate another approach.

shared/fs/common/pie-slice.tsx
jzila: Transparent black doesn't work because it doesn't cover the unfilled sections.

jakob223: Oh, I didn't realize 05 meant transparent. Maybe a black that is slightly gray? I'm ok with the current color though if there isn't an obvious alternative.

shared/fs/common/pie-slice.tsx
jakob223: I think the rectangles only look awful when there is a lot of contrast between the background color and the main circle color. If we can't avoid that, though, I agree that the spillage is preferable and probably ok.

shared/fs/common/pie-slice.tsx
jzila: I think we should keep it as is right now, and then if we ever need it on a red/green dark background, we can refactor again to explicitly mention background colors instead of positive/negative.

jzila:
@jakob223 I think this PR is a big improvement over what's there right now. Let's merge it and then fix it if we're still unhappy with it.

shared/router-v2/header/index.desktop.tsx
```diff
@@ -180,7 +180,7 @@ class Header extends React.PureComponent<Props> {
               />
             )}
             <Kb.Box2 direction="horizontal" style={styles.topRightContainer}>
-              {flags.kbfsOfflineMode && <SyncingFolders />}
+              {flags.kbfsOfflineMode && <SyncingFolders negative={true} />}
```
jakob223: instead of making this always negative, can you make it negative depending on the color of the background?

jzila: Do we even know that?

shared/router-v2/header/index.desktop.tsx
```diff
@@ -180,7 +180,16 @@ class Header extends React.PureComponent<Props> {
...
             )}
             <Kb.Box2 direction="horizontal" style={styles.topRightContainer}>
-              {flags.kbfsOfflineMode && <SyncingFolders />}
+              {flags.kbfsOfflineMode && (
+                <SyncingFolders
+                  negative={
+                    this.props.style &&
+                    this.props.style.backgroundColor &&
+                    this.props.style.backgroundColor !== Styles.globalColors.transparent &&
+                    this.props.style.backgroundColor !== Styles.globalColors.white
```
jzila: @buoyad I'm really guessing here as to how to check the background color of the header. Would this work?
Keep in mind that the header props are not typed :(

buoyad: Yes this is alright. I don't think it would work on native though.

PR #18033, opened 2019-06-21
teams: digest gregor updates for busting HTCM cache

maxtaco:
- systest and team test to prove it out


maxtaco:
@heronhaye one last PR, this one for forcing refresh in the HTCM loader

PR #18037, opened 2019-06-21
Revendor stellar packages from fork

patrickxb:


mlsteele:
does logrus get used?

patrickxb:
Actually, I think I can get rid of it...let me try.  support/log is only in one place.

patrickxb:
ok, it's gone.

mlsteele:
nice

PR #18041, opened 2019-06-21
Add sbsContacts feature flag

buoyad:
Importing/syncing contacts, searching for contacts, chatting with contacts, and SBS profiles for email/phone will be behind this.

@keybase/y2ksquad


shared/__mocks__/feature-flags.tsx
```diff
@@ -1,8 +1,9 @@
+import {FeatureFlags} from '../util/feature-flags'
```
jakob223: yay typing

PR #18045, opened 2019-06-21
Fast user switching

jakob223:
This PR:
 - adds full names to the RPC that is called to get configured accounts
 - plumbs data to the user switching component
 - implements user switching
 - allows provisioning while logged in

Known issues:
 - [x] Does not let you log in as a user that is currently logged out, unless that user's password is 'password'. Waiting on design for how that should look.
 - [x] Does not let you escape once you start provisioning a new user
 - [x] When you click on the username instead of the row, it takes you to the profile instead of logging you in.
 - [x] No mobile support. Waiting on design to tell me where to put it.
 - [ ] Some full names are not loading for some reason. Maybe whoever does the go-side review can take a guess as to why.
 - [ ] In the middle of the user switch, you see the login screen.
 - [x] if you create a new user and then go back halfway through, you're trapped in the login screen.


jakob223:
@ whoever reviews this, ignore the first commit, it's in a separate PR (which is also up for review!)

CC @keybase/y2ksquad [@Y2Ksquad sorry I've been inconsistent about tagging, will try to do from now on.]

shared/constants/config.tsx
```diff
@@ -74,6 +74,12 @@ export const urlToUsername = (url: {
   const username = usernameMatch.toLowerCase()
   return username
 }
+export const makeConfiguredAccount = I.Record<Types._ConfiguredAccount>({
+  fullname: '',
```
chrisnojima: i think we should keep user metadata in the user store and not scattered around. you can load it through this flow but the data (fullname) should be kept there

shared/login/relogin/container.tsx
```diff
@@ -56,15 +61,16 @@ type State = {
...
 
 type Props = {
-  users: Array<string>
-  onForgotPassword: () => void
-  onSignup: () => void
-  onSomeoneElse: () => void
   bannerError: string
   inputError: string
-  selectedUser: string
+  loggedInMap: I.Map<string, boolean>
```
chrisnojima: we try and keep immutable out of props for components

shared/provision/set-public-name/container.tsx
```diff
@@ -1,37 +1,35 @@
...
 import SetPublicName from '.'
-import {connect, withStateHandlers, compose, safeSubmit} from '../../util/container'
+import {
+  connect,
+  withStateHandlers,
+  compose,
+  safeSubmit,
+  TypedState,
+  TypedDispatch,
+} from '../../util/container'
```
chrisnojima: as these imports get long (or before) likely should switch to `* as Container`

shared/reducers/config.tsx
```diff
@@ -25,7 +25,9 @@ export default function(state: Types.State = initialState, action: Actions): Typ
       return state.merge({
         configuredAccounts: state.configuredAccounts,
         defaultUsername: action.payload.wasCurrentDevice // if revoking self find another name if it exists
-          ? state.configuredAccounts.find(n => n !== state.defaultUsername) || ''
+          ? I.List(state.configuredAccounts.map(account => account.username)).find(
```
chrisnojima: you should find then map instead of mapping all the stuff you're going to throw away

shared/reducers/config.tsx
```diff
@@ -211,14 +213,22 @@ export default function(state: Types.State = initialState, action: Actions): Typ
     case ConfigGen.setAccounts:
       // already have one?
       let defaultUsername = state.defaultUsername
-      if (action.payload.usernames.indexOf(defaultUsername) === -1) {
+      let currentFound = false
+      action.payload.configuredAccounts.forEach(account => {
```
chrisnojima: use `.some`

shared/router-v2/account-switcher/container.tsx
```diff
@@ -0,0 +1,64 @@
...
+import {connect, TypedState, TypedDispatch} from '../../util/container'
+import {memoize} from '../../util/memoize'
+import * as ConfigGen from '../../actions/config-gen'
+import * as ProvisionGen from '../../actions/provision-gen'
+import * as SignupGen from '../../actions/signup-gen'
+
+type OwnProps = {}
+
+const mapStateToProps = (state: TypedState) => ({
+  accountRows: state.config.configuredAccounts
```
chrisnojima: do this in mergeprops, you'll get a new thing every time and thrash

shared/router-v2/account-switcher/container.tsx
```diff
@@ -0,0 +1,64 @@
...
+  (stateProps, dispatchProps, ownProps: OwnProps): Props => ({
+    accountRows: stateProps.accountRows.toArray(),
+    fullname: stateProps.fullname,
+    onAddAccount: dispatchProps.onAddAccount,
+    onCancel: dispatchProps.onCancel,
+    onCreateAccount: dispatchProps.onCreateAccount,
+    onProfileClick: () => dispatchProps._onProfileClick(stateProps.username),
+    onSelectAccount: (username: string) => {
+      const rows = stateProps.accountRows.filter(account => account.username === username)
+      const loggedIn = rows.isEmpty() ? false : rows.get(0).hasStoredSecret
```
chrisnojima: `.first()`

shared/router-v2/tab-bar/container.desktop.tsx
```diff
@@ -1,29 +1,35 @@
...
 type OwnProps = {
   navigation: any
   selectedTab: Tabs.AppTab
 }
 
-const mapStateToProps = state => ({
+const mapStateToProps = (state: TypedState) => ({
   _badgeNumbers: state.notifications.navBadges,
   _walletsAcceptedDisclaimer: state.wallets.acceptedDisclaimer,
+  accountRows: state.config.configuredAccounts
```
chrisnojima: the fact that this is done at least 3x implies we should make this a utility in constants i think

shared/router-v2/tab-bar/container.desktop.tsx
```diff
@@ -65,29 +73,42 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
...
+    fullname: stateProps.fullname,
+    isWalletsNew: stateProps.isWalletsNew,
+    onAddAccount: dispatchProps.onAddAccount,
+    onCreateAccount: dispatchProps.onCreateAccount,
+    onHelp: dispatchProps.onHelp,
+    onProfileClick: () => dispatchProps._onProfileClick(stateProps.username),
+    onQuit: dispatchProps.onQuit,
+    onSelectAccount: (username: string) => {
+      const rows = stateProps.accountRows.filter(account => account.username === username)
+      const loggedIn = rows.isEmpty() ? false : rows.get(0).hasStoredSecret
```
chrisnojima: `.first()`

shared/router-v2/tab-bar/index.desktop.tsx
```diff
@@ -5,8 +5,9 @@ import * as Tabs from '../../constants/tabs'
 import KeyHandler from '../../util/key-handler.desktop'
 import {isDarwin} from '../../constants/platform'
 import './tab-bar.css'
-
-type Props = {
+import flags from '../../util/feature-flags'
```
chrisnojima: i'm surprised to see this featureflag here but not anywhere above. does this all work correctly if the flag is off

shared/util/feature-flags.desktop.tsx
```diff
@@ -15,6 +15,7 @@ const ff: FeatureFlags = {
   chatIndexProfilingEnabled: false,
   conflictResolution: false,
   dbCleanEnabled: false,
+  fastAccountSwitch: true,
```
chrisnojima: why have this flag and just turn it on?

shared/router-v2/tab-bar/index.desktop.tsx
jakob223: I think so.. the only other thing that `configuredAccounts` is used for is the login screen.

I did forget to flag off mobile though. Shouldn't be hard to fix.

shared/util/feature-flags.desktop.tsx
jakob223: Since release is soon and there are a lot of edge cases that felt easy to mess up.

shared/login/relogin/container.tsx
jakob223: Why is that?

shared/util/feature-flags.desktop.tsx
chrisnojima: no i mean, why is this true. this means its on for everyone right now

shared/login/relogin/container.tsx
chrisnojima: because we have a separation of concerns. we can discuss irl if you wish. also for consistencies sake

shared/reducers/config.tsx
jakob223: I'm only throwing away one row - I find it implausible that all the extra indirection from checking `.username` on all the rows I'm mapping is worth not mapping that one row.

shared/util/feature-flags.desktop.tsx
jakob223: I want people to test it - I think it is in a working state and I think I have fixed all the edge-case bugs (though Y2K-154 still should get looked at even if I haven't been able to trigger it).

jakob223:
@chrisnojima can you take another look?

shared/login/relogin/container.tsx
jakob223: sure, I gotta go but will poke you tomorrow. Already made the change - I don't feel strongly about it, just was curious.

shared/util/feature-flags.desktop.tsx
chrisnojima: That's what admin is for

go/libkb/secret_store.go
```diff
@@ -129,6 +129,39 @@ func GetConfiguredAccounts(m MetaContext, s SecretStoreAll) ([]keybase1.Configur
 			IsCurrent: username.Eq(currentUsername),
 		}
 	}
+
+	// Get the full names
+
+	uids := make([]keybase1.UID, 0, len(allUsernames))
+	uidMapper := m.G().UIDMapper
```
pzduniak: nit: you're just referencing this twice and it's saving like 10 characters in total, I avoid doing stuff like this unless I reference something tens of times

go/libkb/secret_store.go
```diff
@@ -129,6 +129,39 @@ func GetConfiguredAccounts(m MetaContext, s SecretStoreAll) ([]keybase1.Configur
 			IsCurrent: username.Eq(currentUsername),
 		}
 	}
+
+	// Get the full names
+
```
pzduniak: nit: I actively fight against empty lines after comments / func decls, while this function is written in this style I _really_ don't like it. It's OK though.

go/libkb/secret_store.go
```diff
@@ -129,6 +129,39 @@ func GetConfiguredAccounts(m MetaContext, s SecretStoreAll) ([]keybase1.Configur
...
+		}
+		uids = append(uids, uid)
+	}
+	usernamePackages, err := uidMapper.MapUIDsToUsernamePackages(m.Ctx(), m.G(),
+		uids, time.Hour*24, time.Second*10, false)
+	if err != nil {
+		if usernamePackages != nil {
+			// If data is returned, interpret the error as a warning
+			m.G().Log.CInfof(m.Ctx(),
+				"error while retreiving full names: %+v", err)
```
pzduniak: `retreiving` -> `retrieving`

go/libkb/secret_store.go
```diff
@@ -129,6 +129,39 @@ func GetConfiguredAccounts(m MetaContext, s SecretStoreAll) ([]keybase1.Configur
...
+		} else {
+			return nil, err
+		}
+	}
+	for _, uPackage := range usernamePackages {
+		account, ok := accounts[uPackage.NormalizedUsername]
+		if ok && (uPackage.FullName != nil) {
```
pzduniak: `()` is redundant, I'd write it like this

```
for _, upkg := range usernamePackages {
	if account, ok := accounts[upkg.NormalizedUsername]; ok && upkg.FullName != nil {
		accounts[upkg.NormalizedUsername].Fullname = upkg.FullName.FullName
	}
}
```

shared/util/feature-flags.desktop.tsx
jakob223: admin doesn't work on mobile right?

go/libkb/secret_store.go
jakob223: It feels ugly to me to be calling a method to get a thing multiple times... maybe I just need to get used to this whole `m.G()` pattern..

go/libkb/secret_store.go
jakob223: AIUI, comments without a line break following them ought to be for the following line, whereas comments with a line break ought to be for the following section. What would you write instead?

shared/util/feature-flags.desktop.tsx
chrisnojima: This isn't a mobile file

shared/reducers/config.tsx
chrisnojima: I'm confused why it's one row. Please explain

jakob223: I misread the code as `filter` not `find`.. you're right, fixing.

shared/util/feature-flags.desktop.tsx
jakob223: ok, I'll turn off the flag on desktop and put it in admin override.

junderw:
Looking forward to this!

go/libkb/secret_store.go
pzduniak: therefore it's a nit, G() just returns a pointer anyways

go/libkb/secret_store.go
pzduniak: I suppose this makes sense, I've _never_ seen it in the wild though!

shared/reducers/config.tsx
```diff
@@ -211,14 +211,30 @@ export default function(state: Types.State = initialState, action: Actions): Typ
...
-        defaultUsername = action.payload.defaultUsername
+      let currentFound = action.payload.configuredAccounts.some(
+        account => account.username === defaultUsername
+      )
+
+      if (!currentFound) {
+        const defaultUsernames = action.payload.configuredAccounts
+          .filter(account => account.isCurrent)
+          .map(account => account.username)
+        defaultUsername = defaultUsernames.length > 0 ? defaultUsernames[0] : ''
```
chrisnojima: 🐑 could be `defaultUsername = defaultUsernames[0] || ''`

shared/router-v2/account-switcher/container.tsx
```diff
@@ -0,0 +1,72 @@
...
+    dispatch(ConfigGen.createSetDefaultUsername({username}))
+    dispatch(RouteTreeGen.createSwitchRouteDef({loggedIn: false, path: ''}))
+  },
+  onSignOut: () => dispatch(RouteTreeGen.createNavigateAppend({path: [SettingsConstants.logOutTab]})),
+})
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps,
+  (stateProps: ReturnType<typeof mapStateToProps>, dispatchProps, ownProps: OwnProps): Props => {
```
chrisnojima: don't need the `ReturnType<>` here

go/libkb/secret_store.go
pzduniak: FWIW either of my suggestions wouldn't compile because `account` isn't used. My concern here was mostly the copying rather than setting a field directly in the struct. Not that it matters performance-wise, it just looks off.

So:
```go
if _, ok := accounts[uPackage.NormalizedUsername]; ok {
	accounts[uPackage.NormalizedUsername].Fullname = uPackage.FullName.FullName
}
```

shared/router-v2/account-switcher/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+const AccountRow = ({
+  entry,
+  onSelectAccount,
+}: {
+  entry: AccountRowItem
+  onSelectAccount: (username: string) => void
+}) => (
+  <Kb.NameWithIcon
+    clickType={onSelectAccount}
```
chrisnojima: i don't think this is valid

shared/router-v2/account-switcher/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+  ]
+}
+
+const styles = Styles.styleSheetCreate({
+  avatarSignedOut: {opacity: 0.4},
+  buttonBox: Styles.padding(
+    0,
+    Styles.globalMargins.small,
+    Styles.globalMargins.tiny,
+    Styles.globalMargins.small
```
chrisnojima: if right === left you can leave out right

shared/router-v2/account-switcher/container.tsx
jakob223: For some reason VS code wasn't inferring the types here until I added this.

shared/router-v2/account-switcher/index.tsx
jakob223: I changed `nameWithIcon` so it would be.

go/libkb/secret_store.go
jakob223: I'm pretty sure ```	accounts[uPackage.NormalizedUsername].Fullname = uPackage.FullName.FullName``` is not allowed by the compiler

pzduniak:
image
i think we need a limit on how many accounts we're displaying here

chrisnojima:
it can just scroll

jakob223:
@chrisnojima anything else?

chrisnojima:
- [ ] switch to an unlogged in user, type a bad passphrase, it'll relog you into the previous user

chrisnojima:
- [ ] kinda a weird experience. click on a user thats not logged in. you're in the relogin screen. you can know use the dropdown to switch users again. if you just select one that's logged in it'll log you in. in the new dropdown we mark the ones that are logged in or not, maybe we need to do that in the relogin screen now (cc: @malgorithms )

chrisnojima:
- [ ] what is add another account. what is create a new account

chrisnojima:
- [ ] choose create new account, click back, now you're at the relogin screen

jakob223:
- [ ] click "add another account", type in a username that doesn't exist, back to the main app with no error

jakob223:
- [ ] "Create another account" back button takes you to login screen instead of back to logged-in universe

jakob223:
>     * [ ]  kinda a weird experience. click on a user thats not logged in. you're in the relogin screen. you can know use the dropdown to switch users again. if you just select one that's logged in it'll log you in. in the new dropdown we mark the ones that are logged in or not, maybe we need to do that in the relogin screen now (cc: @malgorithms )

bump @malgorithms here

malgorithms:
> > ```
> > * [ ]  kinda a weird experience. click on a user thats not logged in. you're in the relogin screen. you can know use the dropdown to switch users again. if you just select one that's logged in it'll log you in. in the new dropdown we mark the ones that are logged in or not, maybe we need to do that in the relogin screen now (cc: @malgorithms )
> > ```
> 
> bump @malgorithms here

nojima's proposal sounds fine to me, unless there's something I'm missing

jakob223:
>     * [ ]  switch to an unlogged in user, type a bad passphrase, it'll relog you into the previous user

This is a service issue - if I send a login to the service with an incorrect password, it will send me a `keybase1.notifySession.loggedIn` message. I tried debugging but wasn't familiar enough with the service code to get anywhere. @pzduniak do you know who I might poke about that?

jakob223:
>     * [ ]  kinda a weird experience. click on a user thats not logged in. you're in the relogin screen. you can know use the dropdown to switch users again. if you just select one that's logged in it'll log you in. in the new dropdown we mark the ones that are logged in or not, maybe we need to do that in the relogin screen now (cc: @malgorithms )

Dev-designed version of this (cc @adamjspooner if we want to actual-design it?)

image

jzila:
* [ ] Switching users routes through the login screen even if the user being switched to is already logged in.

jzila:
I fixed all the conflicts, fixed storyshots, and put up a commit that enables this only on desktop admin. I'm creating tickets to capture the remaining work.

jzila:
I just tested with the admin flag off, and everything looks correct (like master).

shared/actions/config/index.tsx
```diff
@@ -212,11 +217,11 @@ function* loadDaemonAccounts(
       )
     }
 
-    const result = yield* Saga.callPromise(RPCTypes.configGetAllProvisionedUsernamesRpcPromise)
-    let usernames = result.provisionedUsernames || []
-    let defaultUsername = result.defaultUsername
-    usernames = usernames.sort()
-    const loadedAction = ConfigGen.createSetAccounts({defaultUsername, usernames})
+    const configuredAccounts: Array<RPCTypes.ConfiguredAccount> = yield* Saga.callPromise(
+      RPCTypes.loginGetConfiguredAccountsRpcPromise
```
chrisnojima: do we need to do this on login? maybe we can skip if its the first one? i want to avoid extra calls on start that we can avoid

shared/actions/login.tsx
```diff
@@ -72,6 +72,7 @@ function* login(state, action: LoginGen.LoginPayload) {
           clientType: RPCTypes.ClientType.guiMain,
           deviceName: '',
           deviceType: isMobile ? 'mobile' : 'desktop',
+          doUserSwitch: true,
```
chrisnojima: feature flag here?

shared/actions/login.tsx
```diff
@@ -81,8 +82,10 @@ function* login(state, action: LoginGen.LoginPayload) {
     logger.info('login call succeeded')
     yield Saga.put(ConfigGen.createLoggedIn({causedBySignup: false, causedByStartup: false}))
   } catch (e) {
-    // If we're canceling then ignore the error
-    if (e.desc !== cancelDesc) {
+    if (e.name === 'ALREADY_LOGGED_IN') {
```
chrisnojima: maybe better to use code

shared/actions/provision.tsx
```diff
@@ -397,6 +397,7 @@ function* startProvisioning(state) {
         clientType: RPCTypes.ClientType.guiMain,
         deviceName: '',
         deviceType: isMobile ? 'mobile' : 'desktop',
+        doUserSwitch: true, // TODO: maybe this should be false sometimes
```
chrisnojima: feature flag here?

shared/common-adapters/floating-menu/index.tsx
```diff
@@ -8,6 +8,7 @@ import * as React from 'react'
 import Overlay from '../overlay'
 import {Position} from '../relative-popup-hoc.types'
 import MenuLayout, {MenuItem, MenuItems} from './menu-layout'
+import ScrollView from '../scroll-view'
```
chrisnojima: why

shared/login/relogin/container.tsx
```diff
@@ -4,19 +4,20 @@ import * as ProvisionGen from '../../actions/provision-gen'
 import * as SignupGen from '../../actions/signup-gen'
 import HiddenString from '../../util/hidden-string'
 import Login from '.'
-import {connect, isNetworkErr} from '../../util/container'
+import {connect, TypedState, TypedDispatch, isNetworkErr} from '../../util/container'
```
chrisnojima: in general prefer `* as Container`

shared/login/relogin/container.tsx
```diff
@@ -4,19 +4,20 @@ import * as ProvisionGen from '../../actions/provision-gen'
...
+import * as ConfigTypes from '../../constants/types/config'
```
chrisnojima: prefer `Types`

shared/login/relogin/container.tsx
```diff
@@ -25,15 +26,18 @@ const mapDispatchToProps = (dispatch, ownProps: OwnProps) => ({
   onSomeoneElse: () => dispatch(ProvisionGen.createStartProvision()),
 })
 
-const mergeProps = (stateProps, dispatchProps) => {
-  const users = stateProps._users.sort().toArray()
+const mergeProps = (stateProps: ReturnType<typeof mapStateToProps>, dispatchProps) => {
```
chrisnojima: prefer to move this into the connect call inline

shared/login/relogin/index.stories.tsx
```diff
@@ -1,7 +1,10 @@
 import * as React from 'react'
 import * as Sb from '../../stories/storybook'
 import Login, {Props} from '.'
+import * as Constants from '../../constants/config'
 
+const makeAccount = (username: string) =>
+  Constants.makeConfiguredAccount({hasStoredSecret: !username.startsWith('a'), username})
```
chrisnojima: this is a little too tricky (even for a story). lets just make a username `'noSecret'` or something

shared/people/index.d.ts
```diff
@@ -8,6 +8,7 @@ export type Props = {
   followSuggestions: Array<Types.FollowSuggestion>
   getData: (markViewed?: boolean) => void
   onClickUser: (username: string) => void
+  onOpenAccountSwitcher?: (() => void) | null
```
chrisnojima: i'm fixing up a bunch of this in the ts strict branch. this basically allows the value to be a function or null or undefined. we can collapse it by removing the null . if you need it to not be a function just set it as undefined

shared/provision/paper-key/index.tsx
```diff
@@ -59,7 +59,8 @@ class PaperKey extends React.Component<
               disabled={!this.state.paperKey}
               waitingKey={Constants.waitingKey}
             />
-          </Kb.ButtonBar>
+          </Kb.ButtonBar>{' '}
```
chrisnojima: this space likely crashes mobile (did you test mobile)

shared/provision/select-other-device/index.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import * as Types from '../../constants/types/provision'
-import {ListItem2, Box2, List, Text, Icon} from '../../common-adapters'
+import {ListItem2, Box2, List, Text, Icon, Button} from '../../common-adapters'
```
chrisnojima: as we see these we should just make them `* as Kb`

shared/provision/set-public-name/index.tsx
```diff
@@ -1,6 +1,6 @@
 import * as Constants from '../../constants/provision'
 import * as React from 'react'
-import {Input, Box2, WaitingButton, Text, Icon} from '../../common-adapters'
+import {Input, Box2, WaitingButton, Text, Icon, Button} from '../../common-adapters'
```
chrisnojima: same here

shared/router-v2/account-switcher/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+import * as Styles from '../../styles'
+import HeaderHOC from '../../common-adapters/header-hoc'
+import * as ConfigTypes from '../../constants/types/config'
+import {Props as HeaderHocProps} from '../../common-adapters/header-hoc/types'
+
+export type AccountRowItem = {
+  account: ConfigTypes.ConfiguredAccount
+  fullName: string
+}
+export type RowsProps = {}
```
chrisnojima: why?

shared/router-v2/account-switcher/index.tsx
```diff
@@ -0,0 +1,116 @@
+import * as React from 'react'
+import * as Kb from '../../common-adapters'
+import * as Styles from '../../styles'
+import HeaderHOC from '../../common-adapters/header-hoc'
```
chrisnojima: this exists in Kb

shared/router-v2/account-switcher/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+const AccountsRows = (props: Props) => (
+  <Kb.Box2 direction="vertical" fullWidth={true}>
+    {props.accountRows.map(entry => (
+      <Kb.ListItem2
+        type="Small"
+        icon={
+          <Kb.Avatar
+            size={32}
+            username={entry.account.username}
+            style={!entry.account.hasStoredSecret && styles.avatarSignedOut}
```
chrisnojima: this should be s ternary, style can't be false. this is being fixed in my strict branch

shared/router-v2/tab-bar/container.desktop.tsx
```diff
@@ -82,13 +89,14 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => ({
...
+  onAddAccount: dispatchProps.onAddAccount,
+  onCreateAccount: dispatchProps.onCreateAccount,
   onHelp: dispatchProps.onHelp,
   onProfileClick: () => dispatchProps._onProfileClick(stateProps.username),
   onQuit: dispatchProps.onQuit,
   onSettings: dispatchProps.onSettings,
   onSignOut: dispatchProps.onSignOut,
-  onTabClick: (tab: Tabs.AppTab) =>
-    dispatchProps._onTabClick(tab, stateProps._peopleJustSignedUpEmail, stateProps._settingsEmailBanner),
+  onTabClick: (tab: Tabs.AppTab) => dispatchProps._onTabClick(tab, stateProps._walletsAcceptedDisclaimer),
```
chrisnojima: bad merge?

shared/signup/username/container.tsx
```diff
@@ -14,25 +14,38 @@ import {InfoIcon} from '../common'
 type OwnProps = {}
 
 const mapStateToProps = (state: Container.TypedState) => ({
-  error: state.signup.usernameError,
-  initialUsername: state.signup.username,
-  usernameTaken: state.signup.usernameTaken,
-  waiting: anyWaiting(state, Constants.waitingKey),
+  _users: state.config.configuredAccounts,
+  main: {
```
chrisnojima: do not make these nested props. you're going to render thrash

shared/signup/username/container.tsx
```diff
@@ -14,25 +14,38 @@ import {InfoIcon} from '../common'
...
+  _onBack: () => {
     dispatch(SignupGen.createRestartSignup())
     dispatch(RouteTreeGen.createNavigateUp())
   },
-  onContinue: (username: string) => dispatch(SignupGen.createCheckUsername({username})),
-  onLogin: (initUsername: string) => dispatch(ProvisionGen.createStartProvision({initUsername})),
+  _onReturnToMain: () => {
+    dispatch(RouteTreeGen.createSwitchRouteDef({loggedIn: true, path: ''}))
+  },
+  main: {
```
chrisnojima: don't do this either

shared/router-v2/tab-bar/container.desktop.tsx
jzila: Yep, I'll fix.

shared/login/relogin/container.tsx
jzila: `Types` is only when it's the same component.

jzila:
Given the extensive feedback, some of which I'd have to load too much context to apply, I'm going to let Jakob take it on tomorrow. At least he'll come to a conflict-free PR.

jakob223:
Thanks for the help @jzila, and sorry for leaving for vacation with a bunch of in-progress work :|.

Will aim to finish this up today.

shared/common-adapters/floating-menu/index.tsx
jakob223: oops this was from a previous attempt at scrolling

shared/router-v2/account-switcher/index.tsx
jakob223: I guess each thing got removed from this one-by-one and then it got refactored out, and I forgot to remove it.

jakob223:
@chrisnojima can you take another look?

shared/people/index.native.tsx
```diff
@@ -14,7 +15,7 @@ export const Header = (props: Props) => (
         custom: (
           <Kb.Avatar
             username={props.myUsername}
-            onClick={() => props.onClickUser(props.myUsername)}
+            onClick={ff.fastAccountSwitch ? props.onOpenAccountSwitcher : props.onClickUser(props.myUsername)}
```
chrisnojima: this is incorrect, its calling it immediately on render

shared/provision/paper-key/container.tsx
```diff
@@ -1,25 +1,28 @@
...
 import {RouteProps} from '../../route-tree/render-route'
-
+import * as LoginGen from '../../actions/login-gen'
 type OwnProps = RouteProps<{}, {}>
 
-const mapStateToProps = state => ({
+const mapStateToProps = (state: Container.TypedState) => ({
   error: state.provision.error.stringValue(),
   hint: `${state.provision.codePageOtherDeviceName || ''}...`,
+  loggedInAccounts: state.config.configuredAccounts
```
chrisnojima: don't ever map or filter in mapstate functions

shared/provision/set-public-name/container.tsx
```diff
@@ -1,49 +1,50 @@
...
 type OwnProps = {
   deviceName: string
   onChange: (text: string) => void
 } & RouteProps<{}, {}>
 
-const mapStateToProps = state => ({
+const mapStateToProps = (state: Container.TypedState) => ({
   _existingDevices: state.provision.existingDevices,
   error: state.provision.error.stringValue(),
+  loggedInAccounts: state.config.configuredAccounts
```
chrisnojima: same here

shared/provision/set-public-name/container.tsx
```diff
@@ -1,49 +1,50 @@
...
-    onBack: dispatchProps.onBack,
-    onChange: ownProps.onChange,
-    onSubmit,
-  }
-}
-
-export default compose(
-  withStateHandlers<any, any, any>(
+export default Container.compose(
+  Container.withStateHandlers<any, any, any>(
```
chrisnojima: ideally we'd remove this if we're in this code again. don't have to do this now as the PR is so big but in general we want to remove all recompose

chrisnojima:
The list of other logins seems styled incorrectly. A user which is logged in seems greyed out. Also we should sort logged in ones on top of non logged in ones i think

jakob223:
@chrisnojima fixed those things

go/libkb/secret_store.go
```diff
@@ -111,24 +111,55 @@ func NewSecretStore(g *GlobalContext, username NormalizedUsername) SecretStore {
...
 		accounts[username] = keybase1.ConfiguredAccount{
 			Username:  username.String(),
 			IsCurrent: username.Eq(currentUsername),
 		}
 	}
+
+	// Get the full names
+
+	uids := make([]keybase1.UID, 0, len(allUsernames))
+	for _, username := range allUsernames {
```
patrickxb: it's more efficient to do 

    uids := make([]keybase1.UID, len(allUsernames)
    for i, username := range allUsernames {
        // ...
        uids[i] = uid
    }

go/libkb/secret_store.go
```diff
@@ -111,24 +111,55 @@ func NewSecretStore(g *GlobalContext, username NormalizedUsername) SecretStore {
...
+		uid := m.G().UIDMapper.MapHardcodedUsernameToUID(username)
+		if !uid.Exists() {
+			uid = UsernameToUIDPreserveCase(username.String())
+		}
+		uids = append(uids, uid)
+	}
+	usernamePackages, err := m.G().UIDMapper.MapUIDsToUsernamePackages(m.Ctx(), m.G(),
+		uids, time.Hour*24, time.Second*10, false)
```
patrickxb: use `MapUIDsReturnMapMctx`

go/libkb/secret_store.go
```diff
@@ -111,24 +111,55 @@ func NewSecretStore(g *GlobalContext, username NormalizedUsername) SecretStore {
...
+	if err != nil {
+		if usernamePackages != nil {
+			// If data is returned, interpret the error as a warning
+			m.G().Log.CInfof(m.Ctx(),
```
heronhaye:
```suggestion
			m.Info
```

go/libkb/secret_store.go
jzila: Just for fun I ran a test, and it is indeed 7-8% faster!

jakob223:
@chrisnojima any more changes? or can I make those go changes and then merge?

chrisnojima:
I think we need to defer this PR (or feature flag this off) . Too much stuff is messed up if you actually do a switch. Aka my inbox is empty when i switch (things like that). Not necessarily related to this PR but this would let users see this

jakob223:
we're currently only flagging this on for admins, though I can turn it off for admins too if you like. My main goals here are avoiding another painful merge and getting this as a feature I can use when working on other things.

chrisnojima:
k it can go in feature flagged to admins

go/libkb/secret_store.go
jakob223: that makes an import loop

dabura667:
Quick Q for anyone involved in this that can answer:

I just tested using `-s` on command line, but it seems like I don't get desktop notifications for any account other than the account currently "active".

Is this the intended final spec?

Part of the appeal of Slack is that I get notifications for each of my "servers" I am a member of simultaneously. I would be insanely happy to see all currently "logged in and switchable" users on the device to get notifications on any platform where they are currently "logged in and switchable, but not necessarily the "current" user on the device"

Keybase is getting better and better every day. Great job guys.

go/libkb/secret_store.go
maxtaco: please don't make libkb depend on uidmap

maxtaco: ... but if you can break the dep chain with interfaces, that's fine

PR #18046, opened 2019-06-21
Move sticky headers into the scroll view.

adamjspooner:


shared/common-adapters/list.d.ts
```diff
@@ -15,6 +15,7 @@ export type Props<Item> = {
   keyboardShouldPersistTaps?: 'never' | 'always' | 'handled' // mobile only,
   windowSize?: number // Mobile only, has a non-RN default,
   onEndReached?: () => void
+  ListHeaderComponent?: React.Component | Function | React.ReactNode
```
buoyad: `React.ReactNode` is wide enough alone

adamjspooner:
Yup, will do.

shared/provision/select-other-device/index.tsx
```diff
@@ -65,15 +65,17 @@ class SelectOtherDevice extends React.Component<Props> {
         gapEnd={true}
       >
         <Box2 direction="vertical" fullWidth={true} style={styles.contentBox} gap={'medium'}>
-          <Text center={true} type={isMobile ? 'BodyBig' : 'Header'}>
-            For security reasons, you need to authorize with an existing device. Which of your existing
```
buoyad: cc @cecileboucheron should this line to appear on desktop too? it might be confusing to be presented with a list of devices with no explanation otherwise

cecileboucheron: Totally. The desktop design was updated recently to this: https://zpl.io/ank1goQ

Issue #17994, opened 2019-06-19
Error I/O


KernelPan1k:
Hello,

It's impossible to list the content or create new files/folders in a shared private folder. 

     mm@r2d2:~$ ls /keybase/private/kernelpanik,qzll/
     ls: lecture du répertoire '/keybase/private/kernelpanik,qzll/': Erreur d'entrée/sortie

my log id: bebf148eadf605f00a07de1c

Thank you^^

strib:
Hi @KernelPan1k, thanks for the report.  I'm investigating.  It looks like there might have been something weird happening with your device `portable1` on June 2nd.  Would you be able to do a `keybase log send` from that device and post the log ID here?  Thanks!

strib:
Separately, it seems like there are some corrupt or zero-byte files on the disk of the device you did send logs from.  I don't think it is causing the IO error, but it might be affecting performance on that device.  I'd recommend doing the following (in addition to sending the logs I requested above):

1. Make sure you don't have any outstanding data being uploaded from that device in any of your folders.
2. Shutdown Keybase and KBFS completely.
3. Run this command:
```sh
mv $HOME/.local/share/keybase/kbfs_journal .local/share/keybase/kbfs_journal.bkp
```
4. Restart Keybase and KBFS.

At that point your disk issues should be fixed, but you'll still be getting the IO errors on that folder you mentioned.

KernelPan1k:
Hello,

Thank you for your quick answers. I made your instructions and here is the id of the report from the other computer.

portable1 log id: d0a394bf71662cc9dce0b11c

Thank you again for everything.

strib:
Thanks!  I'm making some progress, but during a `keybase log send` we only send the most recent part of the logs, and unfortunately I need some data from older logs.  If you're willing to share more logs with me from that device, it would be great if you can do the following on that device:

```sh
tar -czf /keybase/private/strib@github,kernelpanik/client-17994.logs.tgz $HOME/.cache/keybase/keybase.kbfs.*
```

Thanks!

strib:
Actually, I have another theory -- it seems like your device `portable1` is still in the middle of an upload of some data.  It seems to be having network issues uploading the data sometimes, and whenever it's on it's been trying to upload data but sometimes failing.  You can check `/keybase/.kbfs_status` and look for `UnflushedBytes` to see if it's still trying to upload data.  If so, please try to leave it on until all the bytes are flushed, and that might fix your broken folder.

[...]

PR #18047, opened 2019-06-21
folder_branch_ops: two fixes to the latest merged revision

strib:
1. There was a copy-paste bug in `loadCachedMDChanges()` that always made the `ImmutableRootMetadata` look like it had been flushed to the server whenever the block caches were unembedded from the MD.  That in turn caused the FBO to think that the revision had been merged, when it really hadn't, which wreaked all kinds of havoc.

2. A similar thing could happen if the mdserver connection reset while a bunch of MDs were stil lstuck in the journal.  The background thread would fetch the MDs and, assuming they were all from the server rather than from the journal, would use them to set the latest merged revision.  Instead, only count revisions that go past our current known journal end.  (That shouldn't ever happen, but it's worth guarding against just in case.)

The new test catches both cases, when the fixes are reverted.

Issue: HOTPOT-193
Issue: #17994

### DIFF
--- a/go/kbfs/libkbfs/folder_block_manager_test.go
+++ b/go/kbfs/libkbfs/folder_block_manager_test.go
@@ -59,7 +59,11 @@ func testQuotaReclamation(ctx context.Context, t *testing.T, config Config,
 	require.NoError(t, err, "Couldn't sync from server: %+v", err)
 
 	// Make sure no blocks are deleted before there's a new-enough update.
-	bserverLocal, ok := config.BlockServer().(blockServerLocal)
+	bserver := config.BlockServer()
+	if jbserver, ok := bserver.(journalBlockServer); ok {
+		bserver = jbserver.BlockServer
+	}
+	bserverLocal, ok := bserver.(blockServerLocal)
 	if !ok {
 		t.Fatalf("Bad block server")
 	}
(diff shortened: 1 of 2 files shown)


go/kbfs/libkbfs/folder_block_manager_test.go
```diff
@@ -176,6 +184,43 @@ func TestQuotaReclamationUnembedded(t *testing.T) {
...
+
+	rootNode := GetRootNodeOrBust(
+		ctx, t, config, userName.String(), tlf.Private)
+	jManager, err := GetJournalManager(config)
+	require.NoError(t, err)
+	jManager.PauseBackgroundWork(ctx, rootNode.GetFolderBranch().Tlf)
+
+	ops, _ := testQuotaReclamation(ctx, t, config, userName)
+
+	t.Log("Check that the latest merged revision didn't get updated")
```
jzila: So we're just using the merged revision as a proxy check for unembedding?

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -906,17 +906,12 @@ var errNoFlushedRevisions = errors.New("No flushed MDs yet")
...
 
-// getJournalPredecessorRevision returns the revision that precedes
-// the current journal head if journaling enabled and there are
-// unflushed MD updates; otherwise it returns
-// kbfsmd.RevisionUninitialized.  If there aren't any flushed MD
-// revisions, it returns errNoFlushedRevisions.
-func (fbo *folderBranchOps) getJournalPredecessorRevision(ctx context.Context) (
-	kbfsmd.Revision, error) {
+func (fbo *folderBranchOps) getJournalRevisions(ctx context.Context) (
+	kbfsmd.Revision, kbfsmd.Revision, error) {
```
jzila: Name these please, I had to scroll down to figure out what they represent.

go/kbfs/libkbfs/folder_block_manager_test.go
strib: No, that's actually the whole point of this check.  We know they'll be unembedded because of the `SetBlockChangeEmbedMaxSizeForTesting()` call above.  But the bad merged revision was one of the things causing lots of problems for the user, so that's why I'm checking it directly here.

jzila: Got it, thanks.

PR #18049, opened 2019-06-21
cleanup spammy log lines

joshblum:


joshblum:
cc @mmaxim as we discussed

PR #18055, opened 2019-06-23
signup feedback form

pzduniak:
I don't like how hacky some stuff like `sendError` nulling is but it works OK. Decided to get rid of the long output on errors here as the UI would break.

### DIFF
--- a/shared/settings/feedback/index.tsx
+++ b/shared/settings/feedback/index.tsx
@@ -22,6 +22,8 @@ type Props = {
   onSendFeedback: (feedback: string, sendLogs: boolean, sendMaxBytes: boolean) => void
   sending: boolean
   sendError: Error | null
+  showInternalSuccessBanner: boolean // if true, disables the internal success bar
+  onFeedbackDone: (success: boolean) => void
 }
 
 type State = {
@@ -58,8 +60,9 @@ class Feedback extends React.Component<Props, State> {
       const success = !this.props.sending && !this.props.sendError
       this.setState({
         feedback: success ? '' : this.state.feedback,
-        showSuccessBanner: success,
+        showSuccessBanner: this.props.showInternalSuccessBanner && success,
       })
+      this.props.onFeedbackDone(success)
     }
   }
 
--- a/shared/settings/feedback/index.tsx
+++ b/shared/settings/feedback/index.tsx
@@ -22,7 +22,7 @@ type Props = {
   onSendFeedback: (feedback: string, sendLogs: boolean, sendMaxBytes: boolean) => void
   sending: boolean
   sendError: Error | null
-  showInternalSuccessBanner: boolean // if true, disables the internal success bar
+  showInternalSuccessBanner: boolean // if true, enables the internal success bar
   onFeedbackDone: (success: boolean) => void
 }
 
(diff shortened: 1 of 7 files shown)


pzduniak:
If I should split the actual form into a separate component in some sort of a shared directory let me know.

shared/settings/feedback/index.tsx
```diff
@@ -22,6 +22,8 @@ type Props = {
   onSendFeedback: (feedback: string, sendLogs: boolean, sendMaxBytes: boolean) => void
   sending: boolean
   sendError: Error | null
+  showInternalSuccessBanner: boolean // if true, disables the internal success bar
```
buoyad: `disables` -> `enables`

shared/signup/feedback/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+      buttons={[]}
+      banners={[
+        ...(feedbackSent
+          ? [<Kb.Banner key="feedbackSent" text="Thanks! Your feedback was sent." color="green" />]
+          : []),
+        ...(props.sendError ? errorBanner(props.sendError.message) : []),
+      ]}
+      title="Send feedback"
+    >
+      <FeedbackForm
```
buoyad: Can we wrap this with something to limit the width?

shared/signup/routes.tsx
```diff
@@ -2,6 +2,7 @@ export const newRoutes = {
   signupEnterDevicename: {getScreen: () => require('./device-name/container').default, upgraded: true},
   signupEnterEmail: {getScreen: () => require('./email/container').default, upgraded: true},
   signupEnterUsername: {getScreen: () => require('./username/container').default, upgraded: true},
+  signupSendFeedback: {getScreen: () => require('./feedback/container').default, upgraded: true},
```
buoyad: this should be a modal route

pzduniak: lmk if the current impl is ok, i had to add a new prop to signupscreen to keep the styling consistent

shared/signup/common.tsx
```diff
@@ -66,9 +74,9 @@ const Header = (props: HeaderProps) => (
     fullWidth={true}
     style={Styles.collapseStyles([styles.headerContainer, props.style])}
   >
-    {props.showInfoIcon && (
+    {(props.showInfoIcon || props.showInfoIconRow) && (
       <Kb.Box2 direction="horizontal" fullWidth={true} style={styles.infoIconContainer}>
-        <InfoIcon invisible={props.negative as boolean} />
+        <InfoIcon invisible={(props.negative as boolean) || (props.showInfoIconRow && !props.showInfoIcon)} />
```
buoyad: `|| !props.showInfoIcon` is enough right?

shared/signup/common.tsx
```diff
@@ -7,9 +7,10 @@ import openURL from '../util/open-url'
 import * as Styles from '../styles'
 
 type InfoIconProps = {
+  loggedIn: boolean
   invisible?: boolean
   onDocumentation: () => void
-  onFeedback: () => void
+  onFeedback: (loggedIn: boolean) => void
```
buoyad: :sheep: you can pass this in `mergeProps` in the connector instead of having to plumb `loggedIn` through the component.

shared/signup/common.tsx
pzduniak: `showInfoIconRow` just takes up the space, otherwise the UI jumps around on transition

PR #18056, opened 2019-06-24
fix a bunch of fs load bugs

songgao:
This fixes two fs load bugs that cause immortal spinners:

1. We were using "whether we are connected to MD server" rather than "whether kbfsDaemon is connected to us" as a guard for whether to fire load RPCs. This caused the mobile-only bug where Files folder list just didn't load because we don't have offline mode turned on for mobile yet, so that `online` field is just never populated.

2. In the refresh tags logic, I forgot to check if the actual subscription path equals. So it was only checking if the subscribed TLF is the same. This caused the immortal spinner in file preview because we already had subscribed to the same TLF, but didn't have the mimeType which needed the loadPathMetadata to retrieve.

### DIFF
--- a/shared/fs/common/use-fs-load-effect.tsx
+++ b/shared/fs/common/use-fs-load-effect.tsx
@@ -28,13 +28,15 @@ const useFsLoadEffect = ({
     [dispatch, path, isPathItem]
   )
 
-  const online = Container.useSelector(state => state.fs.kbfsDaemonStatus.online)
+  const kbfsDaemonConnected =
+    Container.useSelector(state => state.fs.kbfsDaemonStatus.rpcStatus) ===
+    Types.KbfsDaemonRpcStatus.Connected
   const load = React.useCallback(
     (refreshTag?: Types.RefreshTag) => {
-      online && wantPathMetadata && loadPathMetadata(refreshTag)
-      online && wantChildren && loadChildren(refreshTag)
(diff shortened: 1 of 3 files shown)


shared/fs/common/use-fs-load-effect.tsx
```diff
@@ -28,13 +28,15 @@ const useFsLoadEffect = ({
...
+    Container.useSelector(state => state.fs.kbfsDaemonStatus.rpcStatus) ===
+    Types.KbfsDaemonRpcStatus.Connected
   const load = React.useCallback(
     (refreshTag?: Types.RefreshTag) => {
-      online && wantPathMetadata && loadPathMetadata(refreshTag)
-      online && wantChildren && loadChildren(refreshTag)
+      kbfsDaemonConnected && wantPathMetadata && loadPathMetadata(refreshTag)
+      kbfsDaemonConnected && wantChildren && loadChildren(refreshTag)
     },
-    [online, wantChildren, wantPathMetadata, loadPathMetadata, loadChildren]
```
jakob223: did you fix this in the other PR that touches this code?

songgao: Yeah, that PR is actually based after this: https://github.com/keybase/client/pull/18057/files#diff-6ed8e652e7555af697a744d854958a92R47

PR #18057, opened 2019-06-24
only favoritesLoad when user is in the Files tab

songgao:
cc @mmaxim @jakob223

### DIFF
--- a/shared/fs/container.tsx
+++ b/shared/fs/container.tsx
@@ -89,6 +89,7 @@ const ChooseComponent = (props: ChooseComponentProps) => {
     path: props.path,
     refreshTag: Types.RefreshTag.Main,
     wantPathMetadata: true,
+    wantTlfs: true,
   })
 
   if (props.kbfsDaemonStatus.rpcStatus !== Types.KbfsDaemonRpcStatus.Connected) {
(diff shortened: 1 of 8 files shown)


shared/fs/container.tsx
```diff
@@ -89,6 +89,7 @@ const ChooseComponent = (props: ChooseComponentProps) => {
     path: props.path,
     refreshTag: Types.RefreshTag.Main,
     wantPathMetadata: true,
+    wantTlfs: true,
```
jakob223: This file is still called when we're in a browser within the Chat tab, right?

songgao: Thanks for the early review! Nothing beats waking up in the morning with a review response :-)

This hook is called in many places, but only this one `wantsTlfs`. And this file is only used in Files tab -- it's the routable component from routes/fs.tsx.

songgao: Oh just realized I might have mis-understood you. Yeah, if you go to a KBFS path from chat, it'd go through this.

jakob223: I mean, if I click on a KBFS link in chat, am I in the FS container? if not then we need to do something to re-fix the sync load config situation.

jakob223: oh, oops, race condition 😛 . Looks like we're on the same page.

jzila:
@songgao was this supposed to fix the mdserver connected race? I'm still getting these issues. I'm also getting issues when syncing a folder, including red bars and weird syncing circle inconsistencies.

songgao:
@jzila no this PR doesn't address that. I'm planning to fix that later this week. Assuming you're talking about AreWeConnectedToMDServer blackbar thing. Though I'd be happy to review if anybody is will to make a PR

PR #18060, opened 2019-06-24
pod install for RN 59.9

buoyad:
cc @chrisnojima we need to `pod install` after updating native deps. I'll get something in `docs/` for this soon. Anywhere else you think would be a good place to write this down?

- [x] Check iOS build


buoyad:
I ticketed making CI check for this Y2K-183

chrisnojima:
Yeah a 1 line ci check seems fine

mmaxim:
So no one has been running iOS on 59.9 then until this goes in I guess, hope it works.

PR #18066, opened 2019-06-24
Show the advanced send banner if the from address and destination address have any trustlines in common

ddworken:
This completes [PICNIC-100](https://keybase.atlassian.net/browse/PICNIC-100) and [PICNIC-105](https://keybase.atlassian.net/browse/PICNIC-105). 

I added a new RPC call `ShowAdvancedSendForm` that returns whether or not the advanced send form banner should be shown when sending from an address to a `(username | account ID)`. The javascript calls this RPC and shows the banner if necessary. 

Note that currently the banner does not contain a link to the advanced send page since it is not yet complete. Thus we should either wait on this PR for the advanced send page to be complete or create another ticket to track adding the link.


shared/constants/types/wallets.tsx
```diff
@@ -83,6 +83,7 @@ export type _Building = {
   secretNote: HiddenString
   sendAssetChoices: Array<StellarRPCTypes.SendAssetChoiceLocal> | null
   to: string
+  shouldShowAdvancedSendForm: boolean
```
cjb: keep alphabetized

shared/reducers/wallets.tsx
```diff
@@ -439,6 +439,10 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
       )
     case WalletsGen.clearTrustlineSearchResults:
       return state.update('trustline', trustline => trustline.set('searchingAssets', undefined))
+    case WalletsGen.showAdvancedSendFormReceived:
+      return state.merge({
+        building: state.get('building').merge({shouldShowAdvancedSendForm: action.payload.shouldShow}),
```
cjb: I can see an argument that the bool should be going into `built` (data received by the service) instead of `building` (data the frontend is preparing to send down to the service to be built), what do you think?

cjb:
Looks like there's an eslint error -- we have a pre-commit hook that runs eslint to help catch those before CI does:

https://github.com/keybase/client#development-guidelines

mlsteele:
Could use `BuildPaymentResLocal`'s `banners`. That would be fewer messages back and forth (fewer states) and utilize existing race-avoidance stuff.

cjb:
@mlsteele :+1:

ddworken:
@mlsteele I didn't use `SendBannerLocal` since it doesn't contain any way of embedding a link in the banner. Do you think it would be better to extend the `SendBannerLocal` record to add support for embedding links?

cjb:
@ddworken I think there are other cases where we show a link in a SendBanner -- it's okay to do a hacky thing where you recognize in the frontend that you're being asked to show the advanced form banner, and then append the link to the service's banner text.

mlsteele:
Yes. I hate to make more work for you, but I do think that would be better.

Like @cjb said, something like `boolean advancedForm` should work

`proofsChanged` addresses a similar special case
```
    // Set if this banner means the user's proofs have changed and that is blocking the payment.
    // `message` may be blank in this case.
    boolean proofsChanged;
```

cjb:
So the change would be:

* send a new `offerAdvancedForm`bool in the banner RPC type

* wallets/send-form/container adds that bool as well as `action: () => dispatch(WalletsGen.createShowAdvancedForm()` to the `<Banner />` component for that banner

* wallets/banner/index.tsx has an analog to this existing code for the new bool:
```js
      {props.reviewProofs && props.onAction && (
        <Text
          type="BodySmallSemiboldPrimaryLink"
          center={true}
          style={styles.secondText}
          negative={true}
          onClick={props.onAction}
        >
          Please review.
        </Text>
      )}
```

go/stellar/build.go
```diff
@@ -18,6 +19,107 @@ import (
...
+	}
+	for _, choice := range res {
+		if choice.Enabled {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func GetSendAssetChoicesLocal(mctx libkb.MetaContext, remoter remote.Remoter, arg stellar1.GetSendAssetChoicesLocalArg) (res []stellar1.SendAssetChoiceLocal, err error) {
```
ddworken: Note that this method was moved from `frontend.go` so that I could reference it from within this file.

protocol/avdl/stellar1/local.avdl
```diff
@@ -322,6 +322,9 @@ protocol local {
     // Set if this banner means the user's proofs have changed and that is blocking the payment.
     // `message` may be blank in this case.
     boolean proofsChanged;
+    // Set if this banner means there are valid trustlines for non-native assets to be sent from
+    // the from address to the to address. `message` may be blank in this case.
+    boolean offerAdvancedSendFormbool;
```
mlsteele: drop extra `bool`

go/stellar/build.go
```diff
@@ -185,6 +287,16 @@ func BuildPaymentLocal(mctx libkb.MetaContext, arg stellar1.BuildPaymentLocalArg
...
+			offerAdvancedForm, err := ShowAdvancedSendBanner(mctx, remoter, arg.From, arg.To)
+			if err != nil {
+				return res, err
+			}
+			if offerAdvancedForm {
+				res.Banners = append(res.Banners, stellar1.SendBannerLocal{
+					Level:                     "info",
+					OfferAdvancedSendFormbool: true,
+				})
+			}
```
mlsteele: Bump one level in, into the `else`. `Because %v hasn’t set up their wallet yet` should preclude `OfferAdvancedSendForm`.

go/stellar/build.go
```diff
@@ -18,6 +19,107 @@ import (
 	stellarAddress "github.com/stellar/go/address"
 )
 
+func ShowAdvancedSendBanner(mctx libkb.MetaContext, remoter remote.Remoter, from stellar1.AccountID, to string) (shouldShow bool, err error) {
```
mlsteele: Maybe `ShouldShowAdvancedSendBanner` or `ShouldOfferAdvancedSend`. `ShowAdvancedSendBanner` sounds like a function that does the showing

go/stellar/build.go
```diff
@@ -185,6 +287,16 @@ func BuildPaymentLocal(mctx libkb.MetaContext, arg stellar1.BuildPaymentLocalArg
 					})
 				}
 			}
+			offerAdvancedForm, err := ShowAdvancedSendBanner(mctx, remoter, arg.From, arg.To)
```
mlsteele: This is a good candidate for `buildPaymentCache`. This function is called roughly every keystroke when composing a send. And `ShowAdvancedSendBanner` looks to be hitting a remote every time. Could cache in `bpc` ala `LookupRecipient`. Another strategy could be to only recalculate if `From` or `To` change, though there's nothing like that right now.

shared/wallets/banner/index.tsx
```diff
@@ -41,6 +43,25 @@ const Banner = (props: Props) => (
           Please review.
         </Text>
       )}
+      {(props.offerAdvancedSendForm === AdvancedBanner.receiverBanner ||
+        props.offerAdvancedSendForm === AdvancedBanner.senderBanner) &&
+        props.onAction && (
+          <Box2 direction={'vertical'}>
+            {props.offerAdvancedSendForm === AdvancedBanner.receiverBanner &&
+              'This user accepts more assets than XLM. '}
```
cjb: I think this would crash on Android, text string outside of a Kb.Text.

ddworken: Yup you were right but it crashed for a different reason on android (nesting a box inside a text). Fixed and now tested on Android.

shared/wallets/banner/index.tsx
```diff
@@ -41,6 +43,26 @@ const Banner = (props: Props) => (
           Please review.
         </Text>
       )}
+      {alert(JSON.stringify(props))}
```
mlsteele: 🔥

shared/wallets/banner/index.tsx
```diff
@@ -41,6 +43,26 @@ const Banner = (props: Props) => (
...
+            {props.offerAdvancedSendForm === AdvancedBanner.senderBanner &&
+              'You can send more assets than XLM. '}
+            <Text
+              type="BodySmallSemiboldPrimaryLink"
+              center={true}
+              style={styles.secondText}
+              negative={true}
+              onClick={props.onAction}
+            >
+              Send other assets?
```
mlsteele: I'm mildly opposed to this question mark. It's already clear this is a link you are not required to tap. And there's no question that it will work when tapped.

go/stellar/build.go
```diff
@@ -184,6 +299,17 @@ func BuildPaymentLocal(mctx libkb.MetaContext, arg stellar1.BuildPaymentLocalArg
 						Message: "Your Keybase username will not be linked to this transaction.",
 					})
 				}
+
+				offerAdvancedForm, err := bpc.ShouldOfferAdvancedSend(mctx, arg.From, arg.To)
+				if err != nil {
+					return res, err
```
mlsteele: `BuildPaymentLocal` isn't allowed to error when given valid (including partial) inputs. But I think this would error on RPC. Instead, on error `log` and omit the banner.

shared/wallets/banner/index.tsx
ddworken: :man_shrugging:

shared/wallets/banner/index.tsx
ddworken: Sure. Do we want to put a period after it or just delete the question mark?

shared/wallets/banner/index.stories.tsx
```diff
@@ -13,6 +14,18 @@ const examples: Props[] = [
     text: 'russel has a maximum allowed balance of this asset. You may send a maximum of 880.2387456.',
   },
   {background: 'HighRisk' as const, text: 'Connection error. You are offline.'},
+  {
+    background: 'Announcements' as const,
+    offerAdvancedSendForm: AdvancedBanner.receiverBanner,
+    onAction: () => {},
```
cjb: import * as Sb from '../../stories/storybook'

onAction: Sb.action('onAction'),

ddworken:
:+1: @cjb My plan was to wait for the advanced send form to land and then just merge in master and test it with this. That way I can avoid having to do do any weird rebases/un-rebases.

PR #18067, opened 2019-06-24
Trustline transaction specialization

mlsteele:
Historical view of trustline transactions.

image

image


shared/wallets/transaction/index.tsx
```diff
@@ -127,6 +129,33 @@ const Detail = (props: DetailProps) => {
...
+      const asset = (
+        <Text type={'BodySmall'}>
+          <Text type="BodySmallBold">{assetCode}</Text>/{assetIssuer}
+        </Text>
+      )
+      if (props.trustline.remove) {
+        return (
+          <Text type="BodySmall" style={{...{wordBreak: 'break-word'}, ...textStyle}}>
+            You removed a trustline:
+            <Text type="BodySmall" style={textStyle}> </Text>
```
mlsteele: Linter cries out ```error  Empty components are self-closing  react/self-closing-comp```.
```<Text type="BodySmall" style={textStyle} />``` doesn't work since there's no space.

What should it be? There should be this space after the colon.
image

buoyad: You can use `{' '}` to insert a space

mlsteele: Thanks!

PR #18068, opened 2019-06-24
libkbfs: use `data.PathPartString` to obfuscate symlinks

strib:
(Based on #18014, which should be reviewed first.)

Wherever possible, use a `data.PathPartString` to represent a symlink, and obfuscate directly when that's not possible.

Issue: HOTPOT-63

### DIFF
--- a/go/kbfs/libkbfs/folder_branch_ops.go
+++ b/go/kbfs/libkbfs/folder_branch_ops.go
@@ -3255,7 +3255,7 @@ func (fbo *folderBranchOps) makeFakeDirEntry(
 
 func (fbo *folderBranchOps) makeFakeFileEntry(
 	ctx context.Context, dir Node, name data.PathPartString, fi os.FileInfo,
-	sympath string) (de data.DirEntry, err error) {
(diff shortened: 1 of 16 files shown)


go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -3362,10 +3363,11 @@ func (fbo *folderBranchOps) statUsingFS(
 	}
 
 	if fi.Mode()&os.ModeSymlink != 0 {
-		sympath, err = fs.Readlink(name.Plaintext())
+		sympathPlain, err := fs.Readlink(name.Plaintext())
 		if err != nil {
 			return data.DirEntry{}, false, err
```
jakob223: might this error have a plaintext name in it?

strib: Not that I can tell.  If it does, it would be a bug in `libfs.FS`, or here in `folderBranchOps` which is used (recursively) by `libfs.FS`.  It's confusing.  But no, I don't think it can.

PR #18071, opened 2019-06-24
lock value of cli we are using

mmaxim:


go/service/gregor.go
```diff
@@ -313,7 +313,10 @@ func (g *gregorHandler) GetIncomingClient() gregor1.IncomingInterface {
 }
 
 func (g *gregorHandler) GetClient() chat1.RemoteInterface {
-	if g.IsShutdown() || g.cli == nil {
+	g.connMutex.Lock()
+	cli := g.cli
+	g.connMutex.Unlock()
+	if g.IsShutdown() || cli == nil {
 		select {
 		case <-g.connectHappened:
```
joshblum: lock/refresh in here too?

mmaxim: Yeah

joshblum:
here too? https://github.com/keybase/client/blob/b199ec2c91944cd32ea9e31dead4f72c695b94d9/go/service/gregor.go#L480

joshblum:
maybe this should just hold the lock and be called in these places https://github.com/keybase/client/blob/4f2f4add4d82b77724a9fe542a0d9cb6f2aa6c3e/go/service/gregor.go#L407

PR #18072, opened 2019-06-24
signup: add checkbox for email visibility

jakob223:
Add (uncomment, thanks danny!) a checkbox to the email form during signup, and fire off an additional RPC when signup succeeds if it is checked.

Issue: Y2K-66

CC @keybase/y2ksquad

### DIFF
--- a/shared/actions/settings.tsx
+++ b/shared/actions/settings.tsx
@@ -410,7 +410,7 @@ const editEmail = (state, action: SettingsGen.EditEmailPayload, logger) => {
   if (action.payload.verify) {
     return RPCTypes.emailsSendVerificationEmailRpcPromise({email: action.payload.email})
   }
-  if (action.payload.makeSearchable != null) {
+  if (action.payload.makeSearchable !== undefined && action.payload.makeSearchable !== null) {
     return RPCTypes.emailsSetVisibilityEmailRpcPromise({
       email: action.payload.email,
       visibility: action.payload.makeSearchable
--- a/shared/actions/settings.tsx
+++ b/shared/actions/settings.tsx
@@ -410,14 +410,12 @@ const editEmail = (state, action: SettingsGen.EditEmailPayload, logger) => {
   if (action.payload.verify) {
     return RPCTypes.emailsSendVerificationEmailRpcPromise({email: action.payload.email})
   }
-  if (action.payload.toggleSearchable) {
-    const currentSettings = state.settings.email.emails.get(action.payload.email)
-    const newVisibility = currentSettings
-      ? flipVis(currentSettings.visibility)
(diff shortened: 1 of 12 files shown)


shared/actions/signup.tsx
```diff
@@ -50,6 +51,12 @@ const showErrorOrCleanupAfterSignup = (state: TypedState) =>
     ? SignupGen.createRestartSignup()
     : RouteTreeGen.createNavigateAppend({parentPath: [loginTab], path: ['signupError']})
 
+// If the email was set to be visible during signup, we need to set that with a separate RPC.
+const setEmailVisibilityAfterSignup = (state: TypedState) =>
+  noErrors(state) &&
+  state.signup.emailVisible &&
+  SettingsGen.createEditEmail({email: state.signup.email, toggleSearchable: true})
```
buoyad: I got a black bar here on signing up:
https://github.com/keybase/client/blob/8cac5949400d62f5b259b5d793c27068bd696ba9/shared/actions/settings.tsx#L414
`TypeError: Cannot read property 'get' of null`. Maybe because it's disabled on the server for the user I signed up? In any case, maybe it makes more sense to call the RPC here and not depend on the state being predictable during signup / login.

shared/signup/email/index.tsx
```diff
@@ -51,13 +49,14 @@ const EnterEmail = (props: Props) => {
...
             value={email}
           />
-          {/* TODO hook in to "add an email" settings
-          <Kb.Checkbox
-          label="Allow friends to find you by this email address"
-          checked={props.allowSearch}
-          onCheck={props.onChangeAllowSearch}
-          style={styles.checkbox}
-        /> */}
+          {
```
buoyad: Don't need brackets

shared/actions/signup.tsx
jakob223: I think I'll change that action to be `setVisibility` instead of `toggleVisibility`. Then we won't need to check `email.emails` at all.

shared/actions/settings.tsx
```diff
@@ -410,14 +410,12 @@ const editEmail = (state, action: SettingsGen.EditEmailPayload, logger) => {
   if (action.payload.verify) {
     return RPCTypes.emailsSendVerificationEmailRpcPromise({email: action.payload.email})
   }
-  if (action.payload.toggleSearchable) {
-    const currentSettings = state.settings.email.emails.get(action.payload.email)
-    const newVisibility = currentSettings
-      ? flipVis(currentSettings.visibility)
-      : ChatTypes.Keybase1.IdentityVisibility.private
+  if (action.payload.makeSearchable != null) {
```
buoyad: This may catch `action.payload.makeSearchable === undefined`

PR #18075, opened 2019-06-25
New Team Building for adding people to Teams

MarcoPolo:
Uses the team builder component we are using for chat to add people to teams.

* Supports keyboard navigation!
* Removes the old add people component and replaces all instances of it.

Mostly there were 2 necessary changes to make the team builder work for this.

1. Namespace the team-building actions. 
We need to know whether an action was designed for the chat conversation builder or for the teams team builder. A simple namespace field `type Namespace = 'teams' | 'chat2'` does the trick

2. Add a role picker to the team builder component.
This is needed because you have to specify the role that you will add them as. 

@keybase/react-hackers 

Preview:
add-to-team


shared/actions/chat2/index.tsx
```diff
@@ -2991,6 +2993,27 @@ const onMarkInboxSearchOld = state =>
...
+  Chat2Gen.createSelectConversation({
+    conversationIDKey: Constants.pendingWaitingConversationIDKey,
+    reason: 'justCreated',
+  }),
+  Chat2Gen.createCreateConversation({
+    participants: state[namespace].teamBuilding.teamBuildingFinishedTeam.toArray().map(u => u.id),
+  }),
+]
+
+export function* chatTeamBuildingSaga(): Saga.SagaGenerator<any, any> {
```
chrisnojima: i think this needs to be yielded to in chat2Saga

shared/actions/typed-routes.tsx
```diff
@@ -0,0 +1,14 @@
+// Typed aliases for certain routes so you they are guaranteed to be routed
+// correctly
+
+import * as RouteTreeGen from './route-tree-gen'
```
chrisnojima: lets not expand on this. i have tickets to make new routing actions and that'll include better typing. this will be deprecated

chrisnojima: you can keep it for now but i don't expect this file will grow

shared/team-building/container.tsx
```diff
@@ -89,45 +102,74 @@ const deriveUserFromUserIdFn = memoize(
...
     ownProps.selectedService,
   ])
 
+  const preExistingTeamMembers: I.Map<string, MemberInfo> = ownProps.teamname
+    ? state.teams.teamNameToMembers.get(ownProps.teamname)
+    : I.Map()
+
+  const disabledRoles = ownProps.teamname
+    ? getDisabledReasonsForRolePicker(state, ownProps.teamname, null)
+    : {}
```
chrisnojima: should hoist this so its not new every time

shared/team-building/role-picker-header-action.tsx
```diff
@@ -0,0 +1,36 @@
...
+    (props.rolePickerProps && props.rolePickerProps.showRolePicker) || false
+  )
+  return (
+    <FloatingRolePicker
+      open={rolePickerOpen}
+      onConfirm={props.onFinishTeamBuilding}
+      onSelectRole={props.rolePickerProps.onSelectRole}
+      selectedRole={props.rolePickerProps.selectedRole}
+      onCancel={() => setRolePickerOpen(false)}
+      confirmLabel={`Add as ${props.rolePickerProps.selectedRole}${props.plural ? 's' : ''}`}
```
chrisnojima: `pluralize` util/string

shared/team-building/team-box.tsx
```diff
@@ -62,19 +66,40 @@ const TeamInput = (props: Props) => (
...
+        {!!props.teamSoFar.length &&
+          (props.rolePickerProps ? (
+            <FloatingRolePicker
+              open={props.rolePickerProps.showRolePicker}
+              onConfirm={props.onFinishTeamBuilding}
+              onSelectRole={props.rolePickerProps.onSelectRole}
+              selectedRole={props.rolePickerProps.selectedRole}
+              onCancel={() => props.rolePickerProps.changeShowRolePicker(false)}
+              disabledRoles={props.rolePickerProps.disabledRoles}
+              confirmLabel={`Add as ${props.rolePickerProps.selectedRole}${plural ? 's' : ''}`}
```
chrisnojima: ditto

shared/tsconfig.fast.json
```diff
@@ -3,6 +3,13 @@
   "compilerOptions": {
     "isolatedModules": false,
     "tsBuildInfoFile": "./.tsOuts/.tsOutFast/cache",
-    "outDir": "./.tsOuts/.tsOutFast/emit"
+    "outDir": "./.tsOuts/.tsOutFast/emit",
```
chrisnojima: this already derives from `tsconfig.json` so its not needed

shared/team-building/container.tsx
MarcoPolo: hoist the empty obj?

shared/actions/typed-routes.tsx
MarcoPolo: Cool, I think having something like this is very useful to make sure all callers are routing correctly.

PR #18077, opened 2019-06-25
teambot eks

joshblum:
depends on https://github.com/keybase/keybase/pull/3974 for server changes. Introduces a new teambot ephemeral key which is derived from the teamEK for a given user and boxed with this user's userEK. The key will be used for isolating team content for bot consumption, allowing members to proxy content to/from the team to a bot. We use the ephemeral infrastructure to keep any ephemeral team content sent to the bot from persisting. 

cc @keybase/hotpotatosquad


joshblum:
merging over unrelated ci fail

PR #18082, opened 2019-06-25
Update lint

chrisnojima:
- [x] Updates eslint to 6
- [x] Moves eslint up into shared which helps it figure out node_modules (needed in 6)
- [x] Adopts eslint:recommended
- [x] Fixes prototype pollution (not really a problem but a best practice: https://eslint.org/docs/rules/no-prototype-builtins)
- [x] Wraps switch statements in scope when there is a var (we mostly do this but its not consistent)
- [x] There is a rule about atomic updates which i think is pretty flakey, i left it on in cases they fix it up as i thinks useful
- [x] Adds override to react-native-image-picker typing for the extra props we pass
- [x] Adds cache flag to eslint which speeds it up a ton

Reviewer! Definitely turn on the hide whitespace setting


chrisnojima:
@keybase/react-hackers

chrisnojima:
@keybase/react-hackers bump

shared/package.json
```diff
@@ -7,7 +7,7 @@
     "_node": "cross-env-shell BABEL_ENV=yarn babel-node --extensions='.tsx' --extensions='.js' ",
     "help": "yarn run _helper help",
     "lint": "yarn run lint-specific .",
-    "lint-specific": "cd .. && shared/node_modules/.bin/eslint --quiet --ext js --ext ts --ext tsx --ext d.ts",
+    "lint-specific": "eslint --cache --quiet --ext js --ext ts --ext tsx --ext d.ts --ext mobile.tsx --ext desktop.tsx",
```
nathunsmitty: Why not use a single `--ext` option like `--ext js,ts,tsx,d.ts,mobile.tsx,desktop.tsx`?

chrisnojima: didn't try it

PR #18084, opened 2019-06-25
Mobile 'Manage contacts' setting screen

buoyad:
Done:
- ~Add `expo-contacts`~ (not needed yet, will revert)
- Store whether contact import is enabled in config
- Request contact permissions - handle these cases (reviewer please test what you have patience for)
  - Granted
  - Denied (can ask again)
  - Denied (never ask again) (show a red banner)
- Flag `sbsContacts` off for mobile

Not done:
- Send contacts over to service cache (doesn't exist yet - ticketed Y2K-192)

cc @keybase/y2ksquad


shared/android/app/src/main/java/io/keybase/ossifrage/generated/BasePackageList.java
```diff
@@ -8,6 +8,7 @@
   public List<Package> getPackageList() {
     return Arrays.<Package>asList(
         new expo.modules.constants.ConstantsPackage(),
+        new expo.modules.contacts.ContactsPackage(),
```
chrisnojima: not needed right

buoyad: Yep thanks

PR #18090, opened 2019-06-25
Add conversion rate section to path payments

nathunsmitty:
Screen Shot 2019-06-25 at 5 23 41 PM

1/2 of integrating path payments into the transaction detail page.


shared/wallets/transaction/index.tsx
```diff
@@ -212,7 +212,7 @@ const Detail = (props: DetailProps) => {
       {' '}
       by converting{' '}
       <Text type={textTypeExtrabold}>
-        {props.sourceAmount} {props.sourceAsset}
+        {props.sourceAmount} {props.sourceAsset || 'XLM'}
```
nathunsmitty: This gives us transaction messages like:
Screen Shot 2019-06-26 at 11 25 09 AM

I'm not sure if we'd also want to include a local currency conversion for XLM in lumen<->non-native asset conversions.

nathunsmitty:
@keybase/picnicsquad this is ready for review

protocol/avdl/stellar1/local.avdl
```diff
@@ -134,6 +136,8 @@ protocol local {
     string sourceAmountMax;
     string sourceAmountActual;
     Asset sourceAsset;
+    double sourceToDestinationConversionRate; // The amount of the destination asset worth 1 unit of the source asset at the time of the transaction
```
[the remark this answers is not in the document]
mlsteele: Good point. I think conversion rates always require a bit of thinking. So the name doesn't have to encode the direction, and the comment is helpful. `sourceConverstionRate` sounds fine. Other options: `pathConvRate`, `pathConv`, `convRate`.

Storing amounts in strings is lossless. No matter what language you run it through, nothing about the amount will be changed. Storing amounts in floating point is susceptible to a number of problems and should be avoided. For example, some arithmetic will yield inaccurate results. Storing amounts in an integer is better, but still has pitfalls. Different languages and transport protocols may drop bits off an int64.

So keep playing it safe and use strings for representing amounts. And if any manipulations need to be done, do them in go. We have existing functions with tests like `func FormatAmount`.

Btw if the value is actually being changed as in rounding then I like to use `amount` (or `rate`) to mean the precise amount. And `amountDescription` or `amountDesc` (`rateDesc`) to indicate a human-readable value that may be slightly inaccurate. To avoid a potential bug where non-gui code would ingest a rounded value.

shared/wallets/transaction-details/index.tsx
```diff
@@ -322,6 +344,29 @@ const TransactionDetails = (props: NotLoadingProps) => {
...
+        {isPathPayment && (
+          <Kb.Box2 direction="vertical" gap="xtiny" fullWidth={true}>
+            <Kb.Text type="BodySmallSemibold">Conversion rate:</Kb.Text>
+            <Kb.Box2 direction="horizontal" gap="small" fullWidth={true}>
+              <ConvertedCurrencyLabel
+                amount={1}
+                assetCode={props.sourceAsset}
+                issuerDescription={
+                  // If we don't have a sourceAsset, the source is native Lumens
+                  props.sourceAsset === '' ? 'Stellar Lumens' : props.sourceIssuer
```
nathunsmitty: When transferring XLM to an asset or an asset to XLM, it made the most sense to me to just set the "issuer" field to be `Stellar Lumens`. Open to other suggestions, though.

shared/wallets/transaction-details/index.tsx
```diff
@@ -287,8 +296,20 @@ export const TimestampLine = (props: TimestampLineProps) => {
   )
 }
 
+const ConvertedCurrencyLabel = (props: ConvertedCurrencyLabelProps) => (
+  <Kb.Box2 direction="vertical" noShrink={true}>
+    <Kb.Text type="BodyBigExtrabold">
+      {props.amount} {props.assetCode || 'XLM'}
+    </Kb.Text>
+    <Kb.Text type="BodySmall">/{props.issuerDescription || 'Unknown issuer'}</Kb.Text>
```
nathunsmitty: cc @patrickxb on if we need this `Unknown issuer` check or if that'll always be provided server-side.

mlsteele: It's good to be resilient like this. The payment field plumbing tends to break or change over time.

shared/wallets/transaction/index.tsx
patrickxb: we don't need the local currency conversion for XLM in path payments.

shared/wallets/transaction-details/index.tsx
patrickxb: Yes, as discussed in chat, this sounds good to me.

chrisnojima:
@keybase/react-hackers

shared/constants/wallets.tsx
```diff
@@ -293,6 +293,7 @@ export const currencyResultToCurrency = (w: RPCTypes.CurrencyLocal) =>
 
 const _defaultPaymentCommon = {
   amountDescription: '',
+  assetCode: '',
```
buoyad: Which asset is this referring to? Could this field have a more specific name

nathunsmitty: The destination asset. I suppose `destinationAssetCode`, `destinationCode`, etc. could be more specific. I went with just `assetCode` because the payment object seems to consider all properties not prefixed with `source` to be referring to the destination – e.g., `issuerAccountID` and `issuerDescription` refers to the destination asset's issuer's account id and description respectively. The corresponding `source` properties are `sourceAccountID` and `sourceIssuer`.

Since payment can refer to both a direct payment or a path payment, we should probably decide on some sort of convention for names like these

buoyad: Agreed about making a convention for them. For now a comment in the avdl is fine.

go/stellar/transform.go
```diff
@@ -216,6 +223,8 @@ func transformPaymentDirect(mctx libkb.MetaContext, acctID stellar1.AccountID, p
 		}
 	}
 
+	loc.AssetCode = p.Asset.Code
```
mlsteele: might as well do `transformPaymentRelay` too. Which can be hardcoded XLM

PR #18083, opened 2019-06-25
better typing around userReacjis

joshblum:
cc @keybase/hotpotatosquad

PR #18092, opened 2019-06-25
Fix reacji error

mlsteele:
This global errored on `nil` (or whatever) when logging in for the first time since restarting.

The semicolon keeps it from trying to _call_ `{}`


joshblum:
have a fix here: https://github.com/keybase/client/pull/18083

PR #18093, opened 2019-06-25
only store known emojis as frequently used

joshblum:
prevent non-emoji reactions from appearing as blanks in the reacji bar

cc @keybase/hotpotatosquad


joshblum:
merging over flake

PR #18095, opened 2019-06-25
simplefs: add a debug command for path deobfuscation

strib:
This can show multiple paths per obfuscated path, since conflict
suffixes and symlinks make this possible.

Sample output:
```sh
$ keybase fs debug deobfuscate /keybase/private/strib/entry-voyage/physical-exit.pkg
/keybase/private/strib/tmp/trezor-bridge-2.0.13.pkg

$ keybase fs debug deobfuscate /keybase/private/strib/entry-voyage/physical-exit-85487.pkg /keybase/private/strib/entry-voyage/sign-reduce-2
/keybase/private/strib/entry-voyage/physical-exit-85487.pkg:
  /keybase/private/strib/tmp/trezor-bridge-2.0.13.pkg

/keybase/private/strib/entry-voyage/sign-reduce-2:
  /keybase/private/strib/tmp/z9

```

Issue: HOTPOT-66

### DIFF
--- a/go/kbfs/libfs/util.go
+++ b/go/kbfs/libfs/util.go
@@ -6,7 +6,11 @@ package libfs
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"path"
+	"regexp"
+	"strings"
 
 	billy "gopkg.in/src-d/go-billy.v4"
 )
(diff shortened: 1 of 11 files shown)


strib:
Would be nice to sneak this into the release, to help debug any issues that come up where I need to interact with a user about specific obfuscated filenames.  Should be pretty low risk, if anyone has time to review today, I'd appreciate it!

jakob223:
asdfglkjh github ate my most important comment :(

go/kbfs/libfs/util.go
```diff
@@ -48,3 +52,74 @@ func RecursiveDelete(
...
+	if err != nil {
+		return nil, err
+	}
+
+	elem := stripObfuscatedConflictSuffix(p[0])
+	for _, fi := range fis {
+		name := fi.Name()
+		obsName := stripObfuscatedConflictSuffix(fs.PathForLogging(name))
+		if obsName == elem {
+			if len(p) == 1 || !fi.IsDir() {
```
jakob223: I think you don't want the `|| !fi.IsDir()` here. It means that if `a` and `b` hash to the same thing, a deobfuscate that should yield `b/c` would also yield `a`.

strib: @jakob223 	Sorry I don't follow.  The check right above this is `obsName == elem`, which means this code will only be run on things that already match.  This is just putting filenames directly into the results, instead of trying to chroot into them.

strib:
@jakob223:

> I think it would be more user-friendly if the CLI tool gave an error message and a non-zero return code when it can't find a match, instead of silently printing nothing.

What should it do in the case where multiple paths are given, but only one of them has no match?  Fully error and non-0 exit code, or just print the error under the path with no matches and still return 0?

jakob223:
I'd prefer the latter, but I don't have strong feelings as long as there's an acknowledgement that something had no results.

strib:
Addressed feedback, except for the `|| !fi.IsDir()` thing, which I think is correct.

go/kbfs/libfs/util.go
jakob223: if `a` and `b` both obfuscate to `cloth-matrix`, then this check will catch `a` even when the input is `cloth-matix/coin-upset` unless I'm misunderstanding something.

strib: Ohh I see what you're saying.  `a` is a file and `b` is a directory, they're both in the same subdirectory, and we're trying to deobfuscate something that's under `b/`.  Ok you're right, I can drop that clause. One sec.

jakob223:
```$ keybase fs debug deobfuscate /keybase/private/jakob223,strib/just-treat/simple-onion                                                                                                                               
/keybase/private/jakob223,strib/file304/hello
/keybase/private/jakob223,strib/file_3017```

strib: Ok done, please check it out.

strib: And nice catch, by the way!

PR #18097, opened 2019-06-26
Trustlines: Integrate new canAddTrustline bool

cjb:

Fixes being unable to add a trustline to an account with >three digits of XLM in, because there was a string >= number compare happening.


cjb:
@buoyad Thanks -- that might be okay since I think I fixed the black bar in #18101, and the requirement's supposed to be for >0.5 XLM so 1 XLM is fine.  I'll re-test once these fixes are in.

buoyad:
@cjb IIRC it needs to be 0.5XLM in addition to the 1XLM minimum balance, this was a black bar on hitting 'Accept', not the one in #18101

PR #18106, opened 2019-06-26
fix panicky FBO error

jakob223:
If `getRootNode` would return an error and not finish, the `node` it was returning would be `nil`. This would lead to the deferred end log calling `.Obfuscator()` on the nil node, resulting in a panic instead of an error.


jakob223:
Gonna merge as is so as to avoid running CI again.

strib:
Cool thanks, I'll put up another PR to remove it.

PR #18107, opened 2019-06-26
Move canAddTrustline to WalletAccountLocal

ddworken:
See PICNIC-125


ddworken:
@cjb Should [this code](https://github.com/keybase/client/blob/180adce3166afc92234b876950d8102aec2b5fc1/shared/constants/wallets.tsx#L645-L650) be removed now that it is swapped over? Or do we need to keep that around for some reason?

cjb:
@ddworken Oops, good call, removed.

I wish we had dead code detection -- our linters don't do it 'cause they assume we're creating modules with unknown downstream users, instead of a monorepo.  There's probably a way to get webpack etc to do it, though.  I bet it'd find a lot of cruft.

ddworken:
LGTM :+1:

PR #18111, opened 2019-06-26
Ignore error from phoneNumbers feature not enabled in loadMySettings

buoyad:
cc @pzduniak

### DIFF
--- a/go/service/user.go
+++ b/go/service/user.go
@@ -6,6 +6,7 @@ package service
 import (
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/keybase/client/go/avatars"
 	"github.com/keybase/client/go/chat"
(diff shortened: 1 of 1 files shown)


go/service/user.go
```diff
@@ -154,7 +155,11 @@ func (h *UserHandler) LoadMySettings(ctx context.Context, sessionID int) (res ke
 	}
 	phoneNumbers, err := phonenumbers.GetPhoneNumbers(mctx)
 	if err != nil {
-		return res, err
+		if errStr := err.Error(); strings.Contains(errStr, "not enabled") {
```
zapu: we can also check the feature flag here.

but this is ok as well

buoyad: Unable to do that because this is a flag on the server

PR #18116, opened 2019-06-26
Fix PICNIC-98 so that when a new user clicks on the wallet tab it takes them to the wallet tab

ddworken:
Prior to this commit, when a new user clicked on the wallet tab they would see the introduction/warning modal but the highlighted tab would stay the same as the one they were previously on. This PR changes it so that it correctly navigates them to the wallet tab (though the background stays blank). `navigateUp` did not appear to work correctly to navigate away from the wallet tab so now when a new user exits out of the introduction modal they will simply be sent to the first tab (the people tab).

Note that this also changes the code not to directly navigate to the `walletOnboarding` modal since the Wallet tab will automatically bring it up if necessary. 

after

### DIFF
--- a/shared/actions/wallets.tsx
+++ b/shared/actions/wallets.tsx
@@ -913,7 +913,7 @@ const maybeNavToLinkExisting = (state, action: WalletsGen.CheckDisclaimerPayload
   })
 
 const rejectDisclaimer = (state, action: WalletsGen.RejectDisclaimerPayload) =>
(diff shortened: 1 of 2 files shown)


ddworken:
@cjb How do you feel about the behavior of always kicking them back to the people tab? If you'd like to take the ticket and try to make it correctly go back to their previous tab let me know. I'd be curious to see how you get it to work.

PR #18117, opened 2019-06-26
add a new generic status bar for slash commands

mmaxim:
Patch does the following:

1. Add a new status bar for the results of slash commands.
2. Hook up /msg and /location to it so they can plumb errors back to the user.

cc @keybase/hotpotatosquad


go/chat/commands/msg.go
```diff
@@ -26,6 +26,12 @@ func (d *Msg) Execute(ctx context.Context, uid gregor1.UID, convID chat1.Convers
 	if !d.Match(ctx, text) {
 		return ErrInvalidCommand
 	}
+	defer func() {
+		if err != nil {
+			d.getChatUI().ChatCommandStatus(ctx, convID, "Failed to send message",
```
joshblum: headline can fail if the length is too long as well, can you add a status there or a ticket to address other commands?

PR #18122, opened 2019-06-26
leveldb: wait up to 10s if leveldbs are open by another process

strib:
If the supervisor process hasn't completely killed the previous kbfs process yet, this affords us a little buffer time to wait for that process to fully die before we give up and fail to open the caches.

Issue: HOTPOT-214

### DIFF
--- a/go/kbfs/libkbfs/leveldb.go
+++ b/go/kbfs/libkbfs/leveldb.go
@@ -5,11 +5,14 @@
 package libkbfs
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"strconv"
+	"syscall"
+	"time"
 
 	"github.com/keybase/client/go/kbfs/ioutil"
 	"github.com/keybase/client/go/logger"
@@ -25,6 +28,7 @@ const (
 	diskCacheVersionFilename string = "version"
 	metered                         = true
 	unmetered                       = false
+	leveldbMaxWaitOnEagain          = 10 * time.Second
 )
 
 var leveldbOptions = &opt.Options{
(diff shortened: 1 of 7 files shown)


go/kbfs/libkbfs/leveldb.go
```diff
@@ -118,6 +123,41 @@ func openLevelDBWithOptions(stor storage.Storage, options *opt.Options) (
...
+func openLevelDBStorage(
+	ctx context.Context, p string, log logger.Logger) (storage.Storage, error) {
+	// If we can't open the storage file because another process is
+	// still alive that has it open, wait a bit for that other process
+	// to be killed and release the resource.  Count the time from the
+	// beginning of the process's initialization, so we don't wait
+	// N*timeout for every DB we try to open.
+	eagainDeadline := initialStartTime.Add(leveldbMaxWaitOnEagain)
+	logged := false
+	for time.Now().Before(eagainDeadline) {
```
jzila: Shouldn't this be using `config.Clock()`?

go/kbfs/libkbfs/leveldb.go
```diff
@@ -118,6 +123,41 @@ func openLevelDBWithOptions(stor storage.Storage, options *opt.Options) (
...
+			if !logged {
+				if log != nil {
+					log.CDebugf(ctx,
+						"Waiting %s for the db %s to become available: %+v",
+						leveldbMaxWaitOnEagain, p, err)
+				}
+				logged = true
+			}
+			// Sleep and try again.
+			time.Sleep(100 * time.Millisecond)
```
jzila: Should we be blocking the main goroutine?

go/kbfs/libkbfs/leveldb.go
strib: Ideally yes, but we don't have a `config` here, and even if we plumbed it here we wouldn't have one when `initialStartTime` is set.  I guess I could do that in a `sync.Once` within this method but since this code doesn't get tested it didn't see, worth it.

go/kbfs/libkbfs/leveldb.go
strib: I don't see a way not to.  The calling thread is what requires the caches to be initialized.  We'd have to rewrite everything to be non-blocking when opening storage layers.

strib:
I'm not sure I'm going to go forward with this, as new (to me) evidence suggested that in the cases I'm trying to fix, the old process just never dies.  So I'm going to investigate that first.

strib:
The problem I thought was happening wasn't actually happening, so this doesn't solve a real problem. I'm going to abandon it without merging.

PR #18128, opened 2019-06-27
"Share install link"

pzduniak:


shared/chat/conversation/bottom-banner/index.tsx
```diff
@@ -72,10 +74,24 @@ const BrokenTrackerBanner = ({users, onClick}: BrokenTrackerProps) =>
     </BannerBox>
   )
 
-const InviteBanner = ({users}: InviteProps) => (
-  <BannerBox color={globalColors.blue}>
-    <BannerText>Your messages to {users.join(' & ')} will unlock when they join Keybase.</BannerText>
-  </BannerBox>
-)
+const InviteBanner = ({users, inviteEnabled, onShareClick}: InviteProps) =>
+  inviteEnabled && users.length === 1 && users[0].endsWith('@phone') ? (
```
buoyad: Why limit to just one other participant?

shared/chat/conversation/bottom-banner/index.tsx
```diff
@@ -9,6 +9,8 @@ export type BrokenTrackerProps = {
 }
 
 export type InviteProps = {
+  inviteEnabled: boolean
```
buoyad: no need to pass this in, import `flags` in this file

shared/chat/conversation/bottom-banner/index.tsx
pzduniak: gonna ask design about this

malgorithms:
@pzduniak we should do emails too, so if you can ticket that, please do. Please make sure like SMS the email scenario has a follow-up from the server, even if the user doesn't trigger an email from themselves.

Also, make sure that phone number-based messaging from the desktop still works; if not we should have a separate follow-up ticket for that.

Ok, so for this issue...question, if it were all SMS (or all email), would we have the ability to let the user spawn an SMS to the group? one possibility I'm thinking is if there are more than 2 SMS users we use that same row, but we change the wording a bit and let the user spawn a group SMS. They can choose not to do it, but many will, I think, which would be an excellent way for the group to discuss Keybase quickly. Same for email, and in the rare case where there are both SMS and email, we just end up with 2 such rows.

Here's the text in zeplin now, wihch looking at it feels a bit long:

```
Your messages will unlock once they join
 Keybase and verify their phone number.
        Help them install Keybase:
              [Share install link]
```

Looking at the zeplin exports, I'd say that's way too long-winded, and likely I just wanted to say too much there. The targeted individual message could be much more concise,  like this:

```
     Last step: summon Max Krohn!
          [Send install link]
```

(if we don't have a name for them, just use "them")

In the group case:

```
Last step: summon these 4 people!
     [Send install link]
```

^ this would start a group chat. if the user doesn't want to do it (rare) then that's fine, at least we fall back on the less effective server messages.

We can do the same thing with emails. In the weird case where we have email + SMS in the same chat (super rare) then I'd say we could just put both rows. 

pzduniak:
>Please make sure like SMS the email scenario has a follow-up from the server, even if the user doesn't trigger an email from themselves

It does.
image

I'll do email support in this PR. What email template should we use?

>Last step: summon Max Krohn!
This will be blocked by Y2K-178 because it requires contacts (?). @jzila

cecileboucheron:
> In the weird case where we have email + SMS in the same chat (super rare) then I'd say we could just put both rows.
@malgorithms In that case the button could say "Copy share link" and it'd copy the link to the clipboard.

malgorithms:
hmm, I lean towards both rows. if people are doing a large group it'll be extra annoying to go and start a fresh convo with all those people elsewhere just to paste the link, and each row would be a fast shortcut for them with all the info they need. having 2 rows will be a little bit ugly, but people will have a faster experience and onboard the others better.

pzduniak:
What about desktop?

malgorithms:
maybe on desktop, a copy/paste link?

pzduniak:
@buoyad im finishing this ticket now, final review? please test it out on ios, android seems to be working ok
@cecileboucheron added tiny margin to the bottom bar, should be ok for now i think, let me know wdyt

shared/chat/conversation/bottom-banner/index.tsx
```diff
@@ -22,6 +26,7 @@ const commonBannerStyle = {
   paddingLeft: 24,
   paddingRight: 24,
   paddingTop: 8,
+  marginBottom: globalMargins.tiny,
```
buoyad: Can we make this desktop only?

shared/chat/conversation/bottom-banner/index.tsx
```diff
@@ -72,10 +77,72 @@ const BrokenTrackerBanner = ({users, onClick}: BrokenTrackerProps) =>
...
+        <BannerText>Your messages to {users.join(' & ')} will unlock when they join Keybase.</BannerText>
+      </BannerBox>
+    )
+  }
+
+  // On mobile, single recipient, a phone number
+  if (isMobile && users.length === 1 && users[0].endsWith('@phone')) {
+    return (
+      <BannerBox color={globalColors.blue}>
+        <BannerText>Last step: summon Firstname Lastman!</BannerText>
```
buoyad: `// TODO Y2K-203 get the real full name`

What should we do if there is no full name? "Last step: summon them!" ? cc @cecileboucheron

shared/chat/conversation/bottom-banner/index.tsx
```diff
@@ -72,10 +77,72 @@ const BrokenTrackerBanner = ({users, onClick}: BrokenTrackerProps) =>
...
+        <Button
+          label="Send install link"
+          onClick={() => openSMS(users[0].slice(0, -6))}
+          mode="Secondary"
+          style={{
+            marginTop: globalMargins.xxtiny,
```
buoyad: In general we prefer to use wrapping `Box2`s with `gap`s instead of setting all these margins, it's one of the uses it was made for.

shared/util/feature-flags.native.tsx
```diff
@@ -18,7 +18,7 @@ const ff: FeatureFlags = {
   outOfDateBanner: false,
   plansEnabled: false,
   proofProviders: true,
-  sbsContacts: false,
+  sbsContacts: true,
```
buoyad: :fire:

shared/chat/conversation/bottom-banner/index.tsx
```diff
@@ -1,14 +1,18 @@
...
+import {isMobile} from '../../../constants/platform'
 
 export type BrokenTrackerProps = {
   users: Array<string>
   onClick: (user: string) => void
 }
 
 export type InviteProps = {
+  openShareSheet: () => void
+  openSMS: (email: string) => void
```
buoyad: `email` -> `phoneNumber`

buoyad:
This is ready cc @keybase/y2ksquad. Added go changes include
- Make `SyncedContactList` available on `G` (by @zapu)

shared/chat/conversation/bottom-banner/index.tsx
```diff
@@ -89,54 +90,38 @@ const InviteBanner = ({users, openSMS, openShareSheet}: InviteProps) => {
...
-          }}
-        />
+        <Button label="Send install link" onClick={() => openSMS(users[0].slice(0, -6))} mode="Secondary" />
       </BannerBox>
     )
   }
 
   // Any number of recipients, on iOS / Android show the share screen
-  if (isMobile) {
+  if (!isMobile) {
```
buoyad: :fire:

pzduniak: 🧯

PR #18129, opened 2019-06-27
Increase default log send size when on wifi

joshblum:
calculated avg compression ratio from ~50 logs i had downloaded

- [x] verify on mobile

cc @keybase/hotpotatosquad


joshblum:
(ci flake unrelated)

PR #18130, opened 2019-06-27
Fix log-to-trace command

chrisnojima:
was broken on tsx conversion
also fixes some tags getting lost

@keybase/react-hackers

### DIFF
--- a/shared/desktop/yarn-helper/log-to-trace.tsx
+++ b/shared/desktop/yarn-helper/log-to-trace.tsx
@@ -12,11 +12,11 @@ const fs = require('fs')
 const moment = require('moment')
 
 // core regs
-const reg = /([^ ]+) ▶ \[DEBU (keybase|kbfs) ([^:]+):(\d+)] ([0-9a-f]+) ([^[]+)(\[tags:([^\]]+)])?/
+const reg = /([^ ]+) ▶ \[DEBU (keybase|kbfs) ([^:]+):(\d+)] ([0-9a-f]+) ([^[]+)(.*)?/
 const tagsReg = /\[tags:([^\]]+)]/
 const methodPrefixReg = /^(\+\+Chat: )?/
 const methodResultReg = / -> .*$/
-const typeAndMethodReg = /^(\W*)/
+const typeAndMethodReg = /^(gui|[+-|])/
 // gui regs
 const guiCountTypeTimeReg = /\["(Info|Warn|Action)","([^"]+)","(.*)"]/
 const actionReg = /type: ([^ ]+) (.*)/
@@ -31,7 +31,7 @@ const getSwimlane = line => {
 const convertGuiLine = line => {
   const e = guiCountTypeTimeReg.exec(line)
   if (!e) {
-    console.log('Skipping unparsed line:', line)
+    console.log('🛑 Skipping unparsed line:', line)
     return
   }
   const [, type, time, _data] = e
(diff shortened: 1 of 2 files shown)


chrisnojima:
@keybase/react-hackers bump

shared/desktop/yarn-helper/log-to-trace.tsx
```diff
@@ -154,20 +154,25 @@ const buildGood = (old, info) => {
...
+// lines = [
+// 'Line to debug',
+// ]
 let lastGuiLine = null
 const knownIDs = {}
 lines.forEach(line => {
   const info = convertLine(line)
   if (!info) return
 
+  // console.log(`DEBUG line type: '${info.type}' \n${line}\n${JSON.stringify(info, null, 2)}`)
```
nathunsmitty: Should this be here?

shared/desktop/yarn-helper/log-to-trace.tsx
```diff
@@ -12,11 +12,11 @@ const fs = require('fs')
 const moment = require('moment')
 
 // core regs
-const reg = /([^ ]+) ▶ \[DEBU (keybase|kbfs) ([^:]+):(\d+)] ([0-9a-f]+) ([^[]+)(\[tags:([^\]]+)])?/
+const reg = /([^ ]+) ▶ \[DEBU (keybase|kbfs) ([^:]+):(\d+)] ([0-9a-f]+) ([^[]+)(.*)?/
 const tagsReg = /\[tags:([^\]]+)]/
 const methodPrefixReg = /^(\+\+Chat: )?/
 const methodResultReg = / -> .*$/
-const typeAndMethodReg = /^(\W*)/
+const typeAndMethodReg = /^(gui|[+-|])/
```
nathunsmitty: The `[+-|]` captures all characters between character codes 43 (+) to 124 (|) because the `-` makes in , not just `+`, `-`, and `|` like it looks like you want to. Escaping the `-` will just match the `+`, `-`, and `|`. For easy copy/paste: `^(gui|[+\-|])`

shared/desktop/yarn-helper/log-to-trace.tsx
```diff
@@ -196,18 +204,20 @@ lines.forEach(line => {
       lastGuiLine = info
       break
     default:
-      console.log('Unknown line type:', info.type, ':', line)
+    // Kinda noisy, off for now
```
nathunsmitty: This might be a little less noisey with the above regex change.

shared/desktop/yarn-helper/log-to-trace.tsx
```diff
@@ -188,6 +193,9 @@ lines.forEach(line => {
         output.unmatched.push(info)
       }
       break
+    case '|':
```
nathunsmitty: Why do we want to capture pipes if we're ignoring them?

chrisnojima: because they're valid ones to ignore. it's to skip the default case which is a totally unexpected marker

shared/desktop/yarn-helper/log-to-trace.tsx
chrisnojima: its just a useful debugging line. this isn't a prod script so /shrug

shared/desktop/yarn-helper/log-to-trace.tsx
chrisnojima: its actually noisy cause there's a bunch of lines that aren't delimited correctly (+-| etc) so you get a lot of other chars

shared/desktop/yarn-helper/log-to-trace.tsx
chrisnojima: updated

PR #18131, opened 2019-06-27
small random perf

chrisnojima:


chrisnojima:
some small perf things i saw while profiling
@keybase/react-hackers

PR #18132, opened 2019-06-27
Get full name from contacts in conversation header

jakob223:
Parse and nicely display the phone number on mobile. Include a fake contact name.

Issue: Y2K-77


shared/chat/conversation/header-area/normal/container.tsx
```diff
@@ -41,31 +41,32 @@ const mapDispatchToProps = (
...
-  onToggleInfoPanel: dispatchProps.onToggleInfoPanel,
-  onToggleThreadSearch: dispatchProps.onToggleThreadSearch,
-  participants: stateProps._participants.toArray(),
-  pendingWaiting: stateProps.pendingWaiting,
-  smallTeam: stateProps.smallTeam,
-  teamName: stateProps.teamName,
-  unMuteConversation: dispatchProps._onUnMuteConversation,
-})
+const usePhoneHeader = (props: Props): boolean =>
+  props.participants.length === 2 && props.participants.some(participant => participant.endsWith('@phone'))
```
jzila: What about more than 2 participants?

Also: have we confirmed that `@phone` is the canonical way we are representing this?

shared/chat/conversation/header-area/normal/container.tsx
```diff
@@ -41,31 +41,32 @@ const mapDispatchToProps = (
...
+    onToggleInfoPanel: dispatchProps.onToggleInfoPanel,
+    onToggleThreadSearch: dispatchProps.onToggleThreadSearch,
+    participants: stateProps._participants.toArray(),
+    pendingWaiting: stateProps.pendingWaiting,
+    smallTeam: stateProps.smallTeam,
+    teamName: stateProps.teamName,
+    unMuteConversation: dispatchProps._onUnMuteConversation,
+  })),
+  Container.branch((props: Props) => !!props.teamName, Container.renderComponent(ChannelHeader)),
+  Container.branch(usePhoneHeader, Container.renderComponent(ChannelHeader))
```
jzila: `Container.renderComponent(PhoneHeader)`

shared/util/phone-numbers/index.tsx
```diff
@@ -39,4 +39,10 @@ export const validateNumber = (rawNumber: string, region: string) => {
   }
 }
 
+export const formatPhoneNumber = (rawNumber: string) => {
+  // TODO: support non-US numbers
```
jzila: here: 
https://github.com/keybase/client/blob/f85e66bde8535a3ec34d3aa021e1f2f3171b7bc1/shared/util/phone-numbers/index.tsx#L49-L59

shared/chat/conversation/header-area/normal/index.stories.tsx
```diff
@@ -28,6 +29,7 @@ const load = () => {
     .add('Username Header muted', () => <UsernameHeader {...defaultProps} muted={true} />)
     .add('Channel Header for small team', () => <ChannelHeader {...defaultProps} />)
     .add('Channel Header for big team', () => <ChannelHeader {...defaultProps} smallTeam={false} />)
+    .add('Phone header', () => <PhoneHeader {...defaultProps} participants={phones} />)
```
jzila: You need to also change the prop provider so that it does the same thing `container.tsx` does.

https://github.com/keybase/client/blob/f85e66bde8535a3ec34d3aa021e1f2f3171b7bc1/shared/stories/prop-providers.tsx#L38-L46

shared/util/phone-numbers/index.tsx
jakob223: The design uses `+1 (234) 555-6789` whereas this returns either `+12345556789` or `+1 234-555-6789`, I forget which.

shared/chat/conversation/header-area/normal/index.stories.tsx
jakob223: I'm confused - I did this work in `index` - the work I did in `container` doesn't need to be replicated in the provider since I'm specifically calling `PhoneHeader` here. Unless I'm misunderstanding?

shared/chat/conversation/header-area/normal/container.tsx
jakob223: I'm not convinced it's worth the effort to make this header work with a combination of phone numbers and not-phone-numbers. I'll fix it to work with multiple phone numbers though.

I think it is `@phone`. @zapu ?

shared/util/phone-numbers/index.tsx
pzduniak: Have you tried out `NATIONAL`? I only know its there. Otherwise you could hack around it by using the type as you go formatter.

shared/chat/conversation/header-area/normal/container.tsx
zapu: is there a way we can not do stuff like this and make service return `serviceName` field or something?

where does this data come from?

mmaxim: Yeah this isn't good, we can discuss tomorrow.

shared/util/phone-numbers/index.tsx
jzila: @pzduniak the current code uses `NATIONAL`.

shared/chat/conversation/header-area/normal/index.stories.tsx
jzila: You're right, my bad.

buoyad:
cc @keybase/y2ksquad this is ready. Added go changes include:
- Make `SyncedContactList` available on `G` (by @zapu)
- Find & plumb contact name through `InboxUIItem` to conversation header

go/chat/utils/utils.go
```diff
@@ -190,6 +190,55 @@ func splitAndNormalizeTLFNameCanonicalize(mctx libkb.MetaContext, name string, p
...
+			cPhoneOrEmail = nonDigits.ReplaceAllString(cPhoneOrEmail, "")
+		}
+		if cPhoneOrEmail == phoneOrEmail {
+			return &contact.ContactName
+		}
+	}
+	return nil
+}
+
+func isPhoneOrEmail(username string) bool {
```
mmaxim: @zapu is there a better way to do this?

mmaxim: I won't I see, this is kind of a heuristic?

go/chat/utils/utils.go
```diff
@@ -190,6 +190,55 @@ func splitAndNormalizeTLFNameCanonicalize(mctx libkb.MetaContext, name string, p
...
+	return withContacts
+}
+
+var nonDigits = regexp.MustCompile("[^\\d]")
+
+func findContactName(contacts []keybase1.ProcessedContact, phoneOrEmail string, isPhone bool) *string {
+	for _, contact := range contacts {
+		cPhoneOrEmail := contact.Component.ValueString()
+		if isPhone {
+			cPhoneOrEmail = nonDigits.ReplaceAllString(cPhoneOrEmail, "")
```
mmaxim: What is going on here?

go/chat/utils/utils.go
```diff
@@ -1148,13 +1197,18 @@ func PresentConversationErrorLocal(ctx context.Context, g *globals.Context, rawC
...
-	fullNames := make(map[string]string)
+	var writerNames []chat1.UIParticipant
 	for _, p := range rawConv.Info.Participants {
-		writerNames = append(writerNames, p.Username)
-		if p.Fullname != nil {
-			fullNames[p.Username] = *p.Fullname
-		}
+		participantType := chat1.UIParticipantType_USER
+		if isPhoneOrEmail(p.Username) {
+			participantType = chat1.UIParticipantType_IMPTOFU
```
mmaxim: Are you sure you want to use the name "IMPTOFU" for this? No one will no what that means in 6 months.

go/chat/utils/utils.go
zapu: it would be better to use assertion parser and look at assertion key, but this would work too, not as elegantly though.

buoyad: This is used in `PresentConversationLocal` which can't create an `AssertionContext` for the parser. The use in `AttachContactNames` could use the parser, though I thought this was a nice short circuit so we wouldn't run it on every participant in the inbox.

go/chat/utils/utils.go
buoyad: I don't think phone numbers in `ProcessedContact` are guaranteed to be all digits, while ones in assertions are. This is removing non-digits from `cPhoneOrEmail` to make the comparison more reliable.

buoyad:
Thanks @jzila + @mmaxim, addressed feedback PTAL

PR #18133, opened 2019-06-27
fix reply styling on mobile

mmaxim:
Without this, replies on mobile get sized down to not take up the full width of the thread. Since `styles.wrapper` has `flex: 1`, I'm pretty sure this is necessary, but happy to hear about any other ideas.


mmaxim:
Should mention the bad reply behavior only happens with exploding messages.

PR #18134, opened 2019-06-27
use emoji adapter in the phone selector

pzduniak:


shared/util/phone-numbers/index.tsx
```diff
@@ -1,6 +1,8 @@
+import {isElectron} from '../../constants/platform'
```
buoyad: use `isMobile`

PR #18137, opened 2019-06-27
do less in background when on mobile data

joshblum:
patch does the following:

- change default network state on desktop to `notavailable`
- cancel/prevent indexer syncs when on mobile data
- limit convload (from`Syncer` and `InboxSource`) when on mobile data
- skip merkle audits on mobile data
- loosen team audit params on mobile data 
- skip team box audits on mobile data

cc @keybase/hotpotatosquad

### DIFF
--- a/go/chat/convloader.go
+++ b/go/chat/convloader.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/keybase/client/go/chat/globals"
 	"github.com/keybase/client/go/chat/storage"
@@ -112,6 +113,7 @@ type BackgroundConvLoader struct {
 	resumeCh      chan struct{}
 	loadCh        chan *clTask
 	identNotifier types.IdentifyNotifier
+	eg            errgroup.Group
 
 	clock      clockwork.Clock
 	resumeWait time.Duration
@@ -142,6 +144,7 @@ func NewBackgroundConvLoader(g *globals.Context) *BackgroundConvLoader {
 	b.identNotifier.ResetOnGUIConnect()
 	b.newQueue()
 	go b.monitorAppState()
+	go b.monitorNetState()
 
 	return b
 }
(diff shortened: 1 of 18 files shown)


shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -204,7 +204,7 @@ function* checkRPCOwnership(_, action: ConfigGen.DaemonHandshakePayload) {
 }
 
 const initOsNetworkStatus = (state, action) =>
-  ConfigGen.createOsNetworkStatusChanged({isInit: true, online: navigator.onLine, type: 'unknown'})
+  ConfigGen.createOsNetworkStatusChanged({isInit: true, online: navigator.onLine, type: 'wifi'})
```
chrisnojima: why?

joshblum: was worried someone would get the state when not on mobile and disable something since the state wasn't `wifi`

chrisnojima: i think the background services should treat unknown that way instead of lying about being on wifi

joshblum: yeah could wrap it in a method that allows both, just seems more error prone if someone doesn't use it properly. since we have no info for desktop i think `wifi` is a more sensible default

mmaxim:
I'm a little scared about having this in the release, but I guess we have some time.

chrisnojima:
Lets wait

joshblum:
No rush 
- [ ] after release

joshblum:
@mmaxim this is ready

go/chat/convloader.go
```diff
@@ -156,27 +159,30 @@ func (b *BackgroundConvLoader) removeActiveLoadLocked(key string) {
...
-				b.Resume(ctx)
-				suspended = false
-			}
-		case keybase1.MobileAppState_BACKGROUND:
-			b.Debug(ctx, "monitorAppState: backgrounded, suspending load thread")
-			if !suspended {
-				b.Suspend(ctx)
-				suspended = true
+		select {
+		case state = <-b.G().MobileAppState.NextUpdate(&state):
```
mmaxim: I'm not sure this is how we want to do this. Often times I am on good LTE and would appreciate this to run. I think a better strategy would be controlling how many jobs get queued into this on LTE, probably in `Syncer` and `InboxSource`.

joshblum: i'm often on throttled data so the connection type is lte but it's limited by my provider, was hoping to cut down bandwidth in this case. we could also plumb through the connection type and limit on LTE and stop completely on 4g or something

mmaxim: In `Syncer` we load conversations after they are synced, and you will likely be tapping them soon anyway. My proposal was something smarter on that end, where it doesn't queue up more than like 3 convs and they all have to be imp or small teams convs.

joshblum:
in chat 👍 from @maxtaco and @heronhaye for audtior changes

maxtaco:
:+1: ok on the auditor

go/chat/inboxsource.go
```diff
@@ -724,7 +724,14 @@ func (s *HybridInboxSource) fetchRemoteInbox(ctx context.Context, uid gregor1.UI
 		return types.Inbox{}, err
 	}
 
-	bgEnqueued := 0
+	var bgEnqueued int
+	// Limit the number of jobs we enqueue when on a limited data connection in
+	// mobile.
+	maxBgEnqueued := 50
```
mmaxim: Let's lower this on all platforms to like 10 and 3

go/chat/sync.go
```diff
@@ -273,24 +273,43 @@ func (s *Syncer) getShouldUnboxSyncConvMap(ctx context.Context, convs []chat1.Co
...
 			m[conv.GetConvID().String()] = true
 		}
 	}
 	return m
 }
 
+func (s *Syncer) shouldUnboxSyncConv(conv chat1.Conversation) bool {
+	// Skips convs we don't care for.
+	switch conv.Metadata.Status {
+	case chat1.ConversationStatus_BLOCKED,
```
mmaxim: I'd drop `MUTED` from here, otherwise the snippet will be old

joshblum: ah right

PR #18139, opened 2019-06-27
Small xcode cleanup

chrisnojima:
- [x] remove keybase test scheme. 
- [x] cleanup app delegate
- [x] add go debugging env vars while debugging (shows GC details and thread details)

@keybase/react-hackers


buoyad:
> remove keybase test scheme.

There's a target "Keybase for tests" in `Podfile`, want to remove?

PR #18143, opened 2019-06-27
libkb: retire merkle strict time check, it's too dangerous

maxtaco:
- it happened on my machine that time was set back, and everything hard-failed after that
- if ever there is an NTP hiccup or a leap second, etc, this would brick all apps and could not be rolled back on the server
- just retire the check


maxtaco:
Right, some leeway might be a better idea. That makes it slightly more complicated though

maxtaco:
see #18285

PR #18145, opened 2019-06-27
Adjust the min balance to create a new trustline to 1.5

ddworken:
Per PICNIC-128.


ddworken:
And if you don't mind, I'm gonna sneak in another of commit with this PR to make the message "You only have xxxxx XLM available to send" selectable. Just noticed it wasn't selectable while testing this and I think it should be since it is a pretty normal thing to want to copy your maximum available amount to send it.

PR #18153, opened 2019-06-28
label live location as finished

mmaxim:


go/chat/unfurl/scrape_map.go
```diff
@@ -51,6 +56,9 @@ func (s *Scraper) scrapeMap(ctx context.Context, uri string) (res chat1.UnfurlRa
 		}
 		timeStr = fmt.Sprintf("Posted %s.", time.Now().Format("15:04:05 MST"))
```
joshblum: should drop before release or convert to local time

go/chat/unfurl/scrape_map.go
```diff
@@ -51,6 +56,9 @@ func (s *Scraper) scrapeMap(ctx context.Context, uri string) (res chat1.UnfurlRa
...
 		siteName = "Live Location Share"
+		if done {
+			siteName += " (finished)"
```
joshblum: maybe `Live Location Share, finished at XXX` if we get local time?

go/chat/unfurl/scrape_map.go
mmaxim: I'll get time eventually, still thinking about it.

PR #18158, opened 2019-06-28
simplefs: no error when getting sync config for non-existent TLF

strib:
The test is kind of useless but it still seemed like a nice reminder that we should handle this case.


strib:
We might not go with this PR, and just handle the existing error in the GUI.  So I'll remove the reviewers for now.

PR #18163, opened 2019-06-28
Better route types

chrisnojima:
Previously components that were only ever instantiated through the routing mechanism weren't being validated as actual react components. This PR attempts to rectify that. Originally I was going to fix up all the connectors but that ended up being too much work. I've added comments and will ticket that work out.

NOTE: I have changed every route definition to add some casting to help get the typing correct. Due to how TS works if we require() we'll get any back and the any in these objects will 'upcast' the whole object into any and we'll lose all the typing. This is why we have to use the `typeof`. Please follow this pattern (it won't be checkable until we get much better strict support everywhere)

- New debug method: `Container.connectDEBUG`
- New hook `Container.usePrevious` to capture previous props (etc)
- `compose` deprecated (types as any in TS)
- New way to describe routes (with typeof)


shared/package.json
```diff
@@ -212,6 +212,7 @@
     "@types/base64-js": "1.2.5",
     "@types/classnames": "2.2.8",
     "@types/emoji-mart": "2.8.4",
+    "@types/hoist-non-react-statics": "3.3.1",
```
chrisnojima: we had this before somehow and it went away so now we need this ourselves

shared/router-v2/routes.tsx
```diff
@@ -12,29 +12,27 @@ import {newRoutes as teamsNewRoutes, newModalRoutes as teamsNewModalRoutes} from
...
-  {route: deviceNewRoutes, tab: isMobile ? Tabs.settingsTab : Tabs.devicesTab},
-  {route: chatNewRoutes, tab: Tabs.chatTab},
-  {route: peopleNewRoutes, tab: Tabs.peopleTab},
-  {route: profileNewRoutes, tab: Tabs.peopleTab},
-  {route: fsNewRoutes, tab: Tabs.fsTab},
-  {route: settingsNewRoutes, tab: Tabs.settingsTab},
-  {route: teamsNewRoutes, tab: Tabs.teamsTab},
-  {route: walletsNewRoutes, tab: Tabs.walletsTab},
-  {route: gitNewRoutes, tab: Tabs.gitTab},
+// Need all these as clauses as TS will ignore everything if it sees a single any
```
chrisnojima: see https://github.com/microsoft/TypeScript/issues/26184

chrisnojima:
@keybase/react-hackers this is ready. please take a peek even if you don't review it

shared/chat/routes.tsx
```diff
@@ -1,51 +1,96 @@
 import {isMobile} from '../constants/platform'
 
+import ChatConversation from './conversation/container'
+import ChatEnterPaperkey from './conversation/rekey/enter-paper-key'
```
MarcoPolo: Does doing this import here negate the dynamic loading of the Component? What does the compiled js look like for this?

shared/devices/device-page/container.tsx
```diff
@@ -2,15 +2,15 @@ import * as Container from '../../util/container'
 import * as RouteTreeGen from '../../actions/route-tree-gen'
 import DevicePage from '.'
 
-type OwnProps = {}
+type OwnProps = Container.RouteProps<{}, {}>
 
 // TODO(newRouter) after committing to new router:
```
MarcoPolo: [unrelated] We should make a ticket to fix up any of the `todo.*new.*router.*` todos

shared/profile/pgp/finished.desktop.tsx
```diff
@@ -84,21 +84,17 @@ const styles = Styles.styleSheetCreate({
...
-    dispatch(ProfileGen.createFinishedWithKeyGen({shouldStoreKeyOnServer}))
-    dispatch(RouteTreeGen.createClearModals())
-  },
-})
-
 export default namedConnect(
-  mapStateToProps,
-  mapDispatchToProps,
-  (s, d, o) => ({...o, ...s, ...d}),
+  state => ({
```
MarcoPolo: 🐑 Might as well type this, no?

shared/util/container.tsx
```diff
@@ -19,14 +20,28 @@ export const networkErrorCodes = [
 
 export const isNetworkErr = (code: number) => networkErrorCodes.includes(code)
 
-export const getRouteProps = (ownProps: any, key: string) => ownProps.navigation.getParam(key)
+export function getRouteProps <T = any>(ownProps: {navigation: {getParam: (key: string) => unknown}}, key: string) { return ownProps.navigation.getParam(key) as T }
```
MarcoPolo: Does this work?

shared/util/container.tsx
```diff
@@ -19,14 +20,28 @@ export const networkErrorCodes = [
...
 
 export const useAnyWaiting = (...waitingKeys: string[]) =>
   useSelector(state => anyWaiting(state, ...waitingKeys))
 export const useAnyErrors = (...waitingKeys: string[]) => useSelector(state => anyErrors(state, waitingKeys))
+export function usePrevious<T>(value: T) {
+  const ref = React.useRef<T>()
+  React.useEffect(() => {
+    ref.current = value
+  })
+  return ref.current
```
MarcoPolo: This is kinda tricky, but interesting that it works. Here's my understanding:

You create a closure that contains the `value`. You pass that closure to useEffect so that after a render it sets the mutable `.current` field of the `ref` to the value. 

The function returns ref.current which will refer to the value set after the previous render.

MarcoPolo: Maybe a comment like the above, should be included as usage doc, because it may not be straightforward to use.

shared/wallets/sep7-confirm/container.tsx
```diff
@@ -53,11 +53,18 @@ const Connected = Container.connect(mapStateToProps, mapDispatchToProps, (stateP
...
     memoType,
     message,
-    operation,
     originDomain,
     recipient,
     summary,
   } = stateProps.sep7ConfirmInfo
+
+  const rawOp = stateProps.sep7ConfirmInfo.operation
+  const operation = rawOp === 'pay' ? ('pay' as const) : rawOp === 'tx' ? ('tx' as const) : ('' as const)
```
MarcoPolo: `const operation = rawOp === 'pay' ? 'pay' : rawOp === 'tx' ? 'tx' : ''` this seems to work in the TS Playground

shared/chat/routes.tsx
chrisnojima: in my testing it did not. since we just pull it in and use it in typeof its only visible to TS

shared/devices/device-page/container.tsx
chrisnojima: i have a ticket to change all the router2 actions / etc. it'll clean this all up

shared/profile/pgp/finished.desktop.tsx
chrisnojima: ? this is typed

shared/util/container.tsx
chrisnojima: see https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state

PR #18164, opened 2019-06-28
Remove SafeAreaViewTop in order to make the search list properly aligned

ddworken:
Tested with all the different android notches and on a non-notched android.

### DIFF
--- a/shared/wallets/send-form/participants/search.tsx
+++ b/shared/wallets/send-form/participants/search.tsx
@@ -99,7 +99,8 @@ class Search extends React.Component<SearchProps, SearchState> {
             positionFallbacks={[]}
             propagateOutsideClicks={true}
           >
-            {/* If changing layout here, make sure to test on a notched and un-notched phone */}
+            {/* If changing layout here, make sure to test on a notched and un-notched iphone and android */}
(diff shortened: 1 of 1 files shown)


nathunsmitty:
I don't think we can safely remove the safe area on iphone:
## iPhone 5s

### Safe Area
iphone 5s safe area

### No Safe Area
iphone xr no safe area



## iPhone XR

### Safe Area
iphone xr safe area

### No Safe Area
iphone xr no safe area

ddworken:
Hmm, @chrisnojima thoughts on the best approach here? Removing the safe area fixes this for android but causes issues on iOS. From looking at the CSS that the safe area applies to android, if we get rid of [the minimum height](https://github.com/keybase/client/blob/master/shared/common-adapters/safe-area-view.native.tsx#L18-L19) it works but that essentially makes it so the safe area only applies to iOS. Do you have any insight on how that CSS was supposed to work on Android phones and whether this is something that is broken just here v/s with the safe area in general? 

See [PICNIC-127](https://keybase.atlassian.net/secure/RapidBoard.jspa?rapidView=32&projectKey=PICNIC&modal=detail&selectedIssue=PICNIC-127&assignee=5cf537f86b4e420f21f5eed8) for more context.

nathunsmitty:
tested on both notched/unnotched iphone simulators, looks great!

PR #18166, opened 2019-06-28
Remove proxysettings component for the release

ddworken:
Since TCP proxy support isn't going to land in time for the release, let's remove the GUI so people don't think that the feature is fully complete.

### DIFF
--- a/shared/settings/advanced/index.tsx
+++ b/shared/settings/advanced/index.tsx
@@ -97,6 +97,7 @@ const Advanced = (props: Props) => {
             />
           </Kb.Box>
         )}
+        {/* Removed for this release. Will add back in later. */}
         {/* <ProxySettings {...props} /> */}
         <Developer {...props} />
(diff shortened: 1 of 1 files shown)


ddworken:
@keybase/picnicsquad for a review and FYI that we're backing out proxy support for the release.

shared/settings/advanced/index.tsx
```diff
@@ -97,7 +97,7 @@ const Advanced = (props: Props) => {
             />
           </Kb.Box>
         )}
-        <ProxySettings {...props} />
+        {/* <ProxySettings {...props} /> */}
```
nathunsmitty: may be good to add a TODO note here so people know why it's commented, but not a big deal since I'm assuming you'll be uncommenting this tomorrow.

ddworken: Done.

PR #18168, opened 2019-06-28
Don't modify error results after tx finishes

patrickxb:
Some results are marked as errors before the tx happens.  The code before was marking everything as successful when the multi-op tx finished.  This fixes that.

@malgorithms this should fix your test.


patrickxb:
Actually, i'm going to turn off UseMulti for now...

PR #18171, opened 2019-06-28
Use flexbox to center the progress indicator.

adamjspooner:


shared/common-adapters/progress-indicator.native.tsx
```diff
@@ -5,7 +5,7 @@ import {globalColors, collapseStyles} from '../styles'
 
 class ProgressIndicator extends Component<Props> {
   render() {
-    const size = this.props.type === 'Large' ? 'large' : 'small'
+    const size = this.props.type === 'Small' ? 'small' : 'large'
```
chrisnojima: This may effect a ton of screens

PR #18176, opened 2019-06-28
Some QOL changes for new signup+phone+email modals

buoyad:
- Get the right icons
- Add dimension styles to all the icons to eliminate layout thrashing
- Remove icons on email/phone screens in desktop signup
- Reload account settings on every mount instead of just once
- [ ] Storyshots before merge

cc @keybase/y2ksquad


buoyad:
@chrisnojima I'm going to rebase this on master to get it into the release

buoyad:
Er, actually, harder to do that than I thought

PR #18178, opened 2019-06-28
show accountID in settings when domain is empty; BodySmall "Unknown"

songgao:


songgao:
@chrisnojima probably a GH race condition? I rebased this over master head before opening the PR

PR #18180, opened 2019-06-28
kbfs_ops: towards better startup performance on mobile

strib:
[Not to be merged until after code freeze is over.]

* We were unnecessarily fully loading, including identifies, for edit history TLFs.  Now only fully load them if we need to do prefetching.
* Add delays for non-critical startup background work on mobile.
* Make sure we're only running one copy of each background startup routine.
* Not perf-related, but make sure we re-init the synced TLFs on login events.

cc: @keybase/hotpotatosquad 

Issue: HOTPOT-244

### DIFF
--- a/go/kbfs/libkbfs/kbfs_ops.go
+++ b/go/kbfs/libkbfs/kbfs_ops.go
@@ -75,7 +75,7 @@ func NewKBFSOpsStandard(appStateUpdater env.AppStateUpdater, config Config) *KBF
 		ops:                   make(map[data.FolderBranch]*folderBranchOps),
 		opsByFav:              make(map[favorites.Folder]*folderBranchOps),
 		reIdentifyControlChan: make(chan chan<- struct{}),
-		favs: NewFavorites(config),
+		favs:                  NewFavorites(config),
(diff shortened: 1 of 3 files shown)


go/kbfs/libkbfs/kbfs_ops.go
```diff
@@ -1801,26 +1855,60 @@ func (fs *KBFSOpsStandard) initTlfsForEditHistories() {
...
+	if fs.initSyncCancel != nil {
+		fs.initSyncCancel()
 	}
+	fs.initSyncCancel = cancel
+	return ctx, cancel
 }
 
 func (fs *KBFSOpsStandard) initSyncedTlfs() {
+	if fs.config.MDServer() == nil {
+		return
```
songgao: Do we need to try init'ing again later or is this process just gonna die without ever getting a non-nil `mdserver`?

strib: We don't need to retry explicitly because we now call this function in the `Push...StatusChange()` functions, one of which will fire when the mdserver connectivity changes.  So as far as I can tell, it shouldn't ever get stuck because of this return.

PR #18181, opened 2019-06-29
fix broken mobile share in Files tab and uploading row style

songgao:


shared/common-adapters/floating-menu/index.tsx
```diff
@@ -38,6 +47,7 @@ export default (props: Props) => {
       attachTo={props.attachTo}
       style={props.containerStyle}
       propagateOutsideClicks={props.propagateOutsideClicks}
+      key={itemChangeCounter.toString()}
```
songgao: No idea what's going on here, but without this forced re-render it just shows stale data.

chrisnojima: Did you merge master for my fix and run modules.  Let's not do this

songgao: I think so; this is pretty new-ish. OK I’ll try to figure out this weekend

songgao:
Just tried the latest TestFlight which has @chrisnojima’s fix to react gateway and this is actually fixed. I just didn’t do yarn modules earlier.

PR #18182, opened 2019-06-29
do isSelf in advanced send form properly

songgao:


shared/wallets/send-form/body/advanced.tsx
```diff
@@ -37,14 +37,9 @@ const SecretNoteAndPublicMemo = () => {
       <SecretNote
         secretNote={secretNote}
         onChangeSecretNote={onChangeSecretNote}
-        toSelf={false /* TODO */}
-        secretNoteError={null /* TODO PICNIC-142 */}
```
songgao: It looks like these shouldn't happen at all, now that we have the word count check inside the component -- we should never be submitting over-bloated strings to service. If it happens it'd be a bug and should probably black bar anyway. So I'm removing these TODOs for message errors for both note and memo, in the advanced form. Please let me know if something doesn't sound right!

buoyad: Hm, IIRC a problem here was `maxBytes` in plain input wasn't 100% reliable, so we could still end up with notes that were too long. Not sure what "word count check" refers to though.

songgao: Yeah that's what I meant, `maxBytes`. Sounds like we should fix that instead? Either way I've added the TODOs back. PTAL thanks!

PR #18183, opened 2019-06-29
move follower loading into the background instead of bootstrap

mmaxim:
The goal here is to get follower loading off of the critical path of app startup, for which on mobile was pretty much a guaranteed multiple networks calls because of the short cache time. We do the following:
cc @keybase/hotpotatosquad 

1. Refactor the bootstrap RPC to no longer return follower information.
2. Create a new service that loads follower info for the current logged in user in the background, and name it `TrackerLoader`. This loader works by sending two notifications to the frontend when getting a request. The first is the cached data (no matter how old), and the second is the result of a call to the network. The UI can receive the first callback, and if the second is slow, its result will fill in later instead of blocking app startup (or more commonly now, just ending up with no tracker info).
3. The UI still runs the ordering for requesting the load, since I wanted to rule out any race chance between the callback from the background, and the setting of the current logged-in user in the store.


chrisnojima:
- [ ] after release

mmaxim:
I actually think this would be good to get into the release. After those change last to bootstrap from earlier, I very often get no tracker info when I start the app, which is a big problem, even if most users don’t know what those colors mean. Yes it has been “in there forever”, but it is a bad bug.

chrisnojima:
I'd prefer not to change bootstrap code the weekend before a release

mmaxim:
It’s close to release, but my guess is we have at least until Tuesday before that release happens given how this normally plays out. This is actually a fairly safe change, it’s the same flow as before for the most part, you just get faster startup and avoid the horrible bug of having all the usernames be blue.

go/libkb/sync_trackers2.go
```diff
@@ -43,7 +43,7 @@ func (t *Tracker2Syncer) loadFromStorage(m MetaContext, uid keybase1.UID) error
 		return nil
 	}
 	cachedAt := keybase1.FromTime(tmp.Time)
-	if time.Now().Sub(cachedAt) > cacheTimeout {
+	if useExpiration && time.Now().Sub(cachedAt) > cacheTimeout {
```
patrickxb: time.Since(cachedAt)

go/libkb/syncer.go
```diff
@@ -5,15 +5,16 @@ package libkb
...
 	keybase1 "github.com/keybase/client/go/protocol/keybase1"
 	"golang.org/x/net/context"
-	"sync"
 )
 
 type Syncer interface {
 	Contextifier
 	sync.Locker
-	loadFromStorage(m MetaContext, u keybase1.UID) error
+	loadFromStorage(m MetaContext, u keybase1.UID, useExpiration bool) error
```
patrickxb: loadFromStorageUseExpiration(m, u) and loadFromStorageNoExpiration(m, u) might be clearer, but  you don't have to make that change, just a thought.

go/engine/bootstrap.go
```diff
@@ -86,38 +85,6 @@ func (e *Bootstrap) Run(m libkb.MetaContext) error {
...
 	m.Debug("Bootstrap status: uid=%s, username=%s, deviceID=%s, deviceName=%s", e.status.Uid, e.status.Username, e.status.DeviceID, e.status.DeviceName)
 
-	// get user summaries
-	ts := libkb.NewTracker2Syncer(e.G(), e.status.Uid, true)
-	if e.G().ConnectivityMonitor.IsConnected(context.Background()) == libkb.ConnectivityMonitorYes {
-		m.Debug("connected, loading self user upak for cache")
-		arg := libkb.NewLoadUserArgWithMetaContext(m).WithUID(e.status.Uid)
-		if _, _, err := e.G().GetUPAKLoader().Load(arg); err != nil {
-			m.Debug("Bootstrap: error loading upak user for cache priming: %s", err)
-		}
```
patrickxb: this isn't needed anymore?

patrickxb:
This looks fine to me.

One other thing I could see that might be a smaller change would be to just change bootstrap to do RunSyncerCached regardless on network status.  This would assume that UI at some point does another call when it needs the information that will use the network if necessary.

But I'm fine with this.  The current system is broken, my usernames have been consistently blue for a while so i don't think it can be any worse.

go/engine/bootstrap.go
mmaxim: No, I think it is a waste. The purpose was to prime the UPAK cache with yourself, but you end up just slowing down startup to do it.

mmaxim:
The UI does call bootstrap again, but one of the main causes of the lost info is a race between this and Gregor connecting. The UI right now only calls back if they were once offline (I don't think this counts initial state), so we end up losing the info. I had the dual goal of fixing the names not being colored properly, and also making this bootstrap call instant under all network conditions.

PR #18201, opened 2019-07-01
populate default account for path if it's not set

songgao:


shared/wallets/send-form/body/container.tsx
```diff
@@ -33,7 +35,13 @@ const mergeProps = (stateProps, dispatchProps) => ({
       dispatchProps._onGoAdvanced(
         stateProps._building.to,
         stateProps._building.recipientType,
-        stateProps._building.from
+        // This can happen when sending from chat. The normal payment path
+        // goes through buildPayment to get the sender AccountID so we don't
+        // have it here.
+        stateProps._building.from === Types.noAccountID
+          ? stateProps._accountMap.find(account => account.isDefault, null, Constants.unknownAccount)
+              .accountID
```
buoyad: Could this also be `noAccountID`?

songgao: It could, but that’d imply we don’t have a default account or something else is wrong, and in that case it seems better to leave it as `noAccountID` so pick asset button will be disabled.

PR #18204, opened 2019-07-01
settings: badge app with verify emails and phones

jakob223:
Issue: Y2K-189

CC @keybase/y2ksquad

### DIFF
--- a/shared/actions/settings.tsx
+++ b/shared/actions/settings.tsx
@@ -11,6 +11,8 @@ import * as RouteTreeGen from '../actions/route-tree-gen'
 import * as RPCTypes from '../constants/types/rpc-gen'
 import * as SettingsGen from '../actions/settings-gen'
 import * as WaitingGen from '../actions/waiting-gen'
+import * as Tabs from '../constants/tabs'
+import * as NotificationsGen from '../actions/notifications-gen'
 import {mapValues, trim} from 'lodash-es'
 import {delay} from 'redux-saga'
 import {isAndroidNewerThanN, pprofDir, version} from '../constants/platform'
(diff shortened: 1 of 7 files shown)


shared/actions/settings.tsx
```diff
@@ -393,6 +395,21 @@ const loadSettings = () =>
     })
   })
 
+const computeBadgesOnLoadedSettings = (state, action: SettingsGen.LoadedSettingsPayload) => {
+  const emailCount = (action.payload.emails || I.Map())
+    .map(emailRow => (emailRow.isVerified ? 0 : 1))
```
jzila: I think this would be cleaner as a filter-reduce instead of map-reduce.

chrisnojima: why not just do reduce, you don't need a map or filter first

jzila: Hah yes, good call.

shared/actions/settings.tsx
```diff
@@ -393,6 +395,23 @@ const loadSettings = () =>
...
+    (count, row) => (row.verified ? count : count + 1),
+    0
+  )
+
+  return NotificationsGen.createSetBadgeCounts({
+    counts: I.Map({
+      [Tabs.settingsTab as Tabs.Tab]: emailCount + phoneCount,
+    }) as I.Map<Tabs.Tab, number>,
+  })
+}
```
jzila: Actually one more suggestion to evaluate: can you do this inside `loadSettings` instead? You can `SettingsGen.createLoadedSettings()` and `NotificationsGen.createSetBadgeCounts()` in parallel.

shared/actions/settings.tsx
```diff
@@ -726,6 +729,10 @@ function* settingsSaga(): Saga.SagaGenerator<any, any> {
     SettingsGen.deleteAccountForever,
     deleteAccountForever
   )
+  yield* Saga.chainAction<ConfigGen.BootstrapStatusLoadedPayload>(
```
buoyad: I think it's easier to follow causation if `loadSettings` listens for this action itself

shared/actions/settings.tsx
jakob223: How do I return more than one action?

shared/actions/settings.tsx
jakob223: hmm, I prefer this way because `loadSettings` would then have to check logged-in-ness in only some cases.

chrisnojima: i dont think this is a safe assumption

jakob223: what is the assumption?

chrisnojima: That this gets called only when your logged in

jakob223: currently the function that is called in this `chainAction` is checking for logged-in-ness before returning the loadSettings action.

chrisnojima: Doesn't mean you won't get logged out. Do your closure relies on being logged in it's safer to validate it yourself vs relying on something upstream

jakob223:
- [ ] after release

PR #18206, opened 2019-07-01
add team role BOT

joshblum:
patch adds support for team member role BOT. bot members don't receive boxes for the PTK and don't have access to any of the application reader masks. the team cache version is bumped so new clients will be able to correctly process change membership links with bot members.

- [x] after release

depends on https://github.com/keybase/keybase/pull/4013
cc @keybase/hotpotatosquad


joshblum:
note: need to add support for bot members to get the public half of the signing key to verify signatures of derived teambot keys

go/libkb/chain_link_v2.go
```diff
@@ -141,12 +141,11 @@ func (t SigchainV2Type) RequiresAtLeastRole() keybase1.TeamRole {
 	if !t.IsSupportedTeamType() {
 		// Links from the future require a bare minimum.
 		// They should be checked later by a code update that busts the cache.
-		return keybase1.TeamRole_READER
+		return keybase1.TeamRole_BOT
```
mlsteele: bots sign links?

go/teams/chain.go
```diff
@@ -267,6 +262,10 @@ func (t TeamSigChainState) AssertWasRoleOrAboveAt(uv keybase1.UserVersion,
 	return mkErr("%v role point not found", role)
 }
 
+func (t TeamSigChainState) AssertWasBotAt(uv keybase1.UserVersion, scl keybase1.SigChainLocation) (err error) {
```
mlsteele: this and AssertWasReaderAt are unused

go/teams/loader.go
```diff
@@ -332,9 +332,11 @@ func (l *TeamLoader) load1(ctx context.Context, me keybase1.UserVersion, lArg ke
...
 	}
 
-	// Only public teams are allowed to be behind on secrets.
-	// This is allowed because you can load a public team you're not in.
-	if !l.hasSyncedSecrets(mctx, ret.teamShim()) && !ret.team.Chain.Public {
+	// Public teams are allowed to be behind on secrets since you can load a
+	// public team you're not in. Bot members don't have any secrets and are
+	// also exempt.
+	if !l.hasSyncedSecrets(mctx, ret.teamShim()) &&
+		!(ret.team.Chain.Public || ret.team.Chain.UserRole(me).IsBot()) {
```
mlsteele: Makes sense. But you can ditch this paragraph if you want. External callers will need to handle secretless teams because of bots anyhow.

go/teams/loader2.go
```diff
@@ -256,22 +256,33 @@ func (l *TeamLoader) verifyLink(ctx context.Context,
...
+	case keybase1.TeamRole_BOT:
+		err = l.verifyExplicitPermission(ctx, state, link, signerUV, keybase1.TeamRole_BOT)
+		if err == nil {
+			return &signer, err
+		}
+		if !ShouldSuppressLogging(ctx) {
+			l.G().Log.CDebugf(ctx, "verifyLink: %v not a %v: %v", linkType, keybase1.TeamRole_BOT, err)
+		}
+		// Fall through to a higher role check
+		fallthrough
```
mlsteele: This makes `RequiresAtLeastRole -> bot` actually mean `RequiresAtLeastRole -> reader`. Which doesn't make much sense to me. If bots can sign links this shouldn't fall through. If they can't, `RequiresAtLeastRole` shouldn't return `bot`.

go/teams/loader2.go
```diff
@@ -715,6 +726,15 @@ func (l *TeamLoader) addSecrets(mctx libkb.MetaContext,
...
 	state := team.MainChain()
+	stateWrapper := newTeamSigChainState(team)
+	role, err := stateWrapper.GetUserRole(me)
+	if err != nil {
+		role = keybase1.TeamRole_NONE
+	}
+	// Bots don't have any team secrets, so we short circuit.
+	if role.IsBot() {
+		return nil
+	}
```
mlsteele: This might make more sense in loader.go at the callsite of addSecrets. Assuming `computeSeedChecks` and `addKBFSCryptKeys` should also be skipped. Also, bots don't have secrets, could you remind what they _do_ have that enables them to communicate?

go/teams/rpc_exim.go
```diff
@@ -39,13 +39,13 @@ func (t *Team) ExportToTeamPlusApplicationKeys(ctx context.Context, idTime keyba
...
 
-	var writers []keybase1.UserVersion
-	var onlyReaders []keybase1.UserVersion
+	var writers, onlyReaders, onlyBots []keybase1.UserVersion
 
 	writers = append(writers, members.Writers...)
 	writers = append(writers, members.Admins...)
 	writers = append(writers, members.Owners...)
 	onlyReaders = append(onlyReaders, members.Readers...)
+	onlyReaders = append(onlyBots, members.Bots...)
```
mlsteele: `onlyReaders -> onlyBots`

go/libkb/chain_link_v2.go
joshblum: bots cant sign but this is used to verify links in `TeamLoader.verifyLink`

mlsteele: `verifyLink` only handles `BOT` because it's output here `RequiresAtLeastRole`

go/teams/rpc_exim.go
joshblum: oh good catch

go/teams/loader2.go
joshblum: non-bot team members derive a key from the team ephemeral key and box it for the bot's user ephemeral key: https://github.com/keybase/client/blob/master/go/ephemeral/teambot_ek.go#L22-L27. Conversations involving the bot use this derived key and then members pipe the bot's responses back into a normal team conv

go/libkb/chain_link_v2.go
joshblum: actually bots have to be able to sign leave links, can make the default for unsupported back to reader though

go/teams/loader2.go
joshblum: i thought the `fallthrough` was really acting as a goto to the implicit admin case. `verifyExplicitPermission` is (confusingly) checking `atOrAbove` for the role so failing a lower permission check would imply failing all of the subsequent cases except if you're an offchain implicit admin https://github.com/keybase/client/pull/18206/files#diff-53390555ab6dc886c9994361396f9627R300

mlsteele: Yup my bad. fallthrough looks ok

PR #18207, opened 2019-07-01
fix black bar after tapping cancel in Trustline search

songgao:


songgao:
storyshot error seems to be the one fixed in master. merging!

PR #18214, opened 2019-07-01
Path payment recipient fixes

nathunsmitty:
Two things:

1. Removes the "by converting" text in path payments
2. Corrects the issues with the recipient of a path payment not seeing the amount in the correct asset.

### DIFF
--- a/shared/wallets/transaction/index.tsx
+++ b/shared/wallets/transaction/index.tsx
@@ -540,7 +540,9 @@ export const Transaction = (props: Props) => {
                   canceled={props.status === 'canceled'}
                   pending={pending}
                   yourRole={props.yourRole}
-                  sourceAmountDescription={`${props.sourceAmount} ${props.sourceAsset || 'XLM'}`}
+                  sourceAmountDescription={
+                    props.sourceAmount ? `${props.sourceAmount} ${props.sourceAsset || 'XLM'}` : undefined
+                  }
                   amountDescription={props.amountXLM}
                 />
               )}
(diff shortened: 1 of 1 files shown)


shared/wallets/transaction/index.tsx
```diff
@@ -212,29 +216,20 @@ const Detail = (props: DetailProps) => {
...
-      </Text>
-    </Text>
-  ) : null
-
   switch (props.yourRole) {
     case 'airdrop':
-      return <Text type={textType} style={textStyle}>
-        {counterparty()}
-      </Text>
+      return (
```
nathunsmitty: I think prettier made a couple changes here

shared/wallets/transaction/index.tsx
```diff
@@ -309,24 +301,24 @@ const roleToColor = (role: Types.Role): string => {
...
+      return `+ ${amount}`
     case 'senderAndReceiver':
       return '0 XLM'
     default:
       throw new Error(`Unexpected role ${role}`)
   }
 }
 
-const AmountXLM = (props: AmountXLMProps) => {
+const Amount = (props: AmountProps) => {
```
nathunsmitty: Since we don't display just XLM amounts, it made sense to tweak the name of this component.

shared/wallets/transaction/index.tsx
```diff
@@ -309,24 +301,24 @@ const roleToColor = (role: Types.Role): string => {
...
   switch (role) {
     case 'senderOnly':
-      return `- ${amountXLM}`
+      return `- ${sourceAmount || amount}`
     case 'airdrop':
     case 'receiverOnly':
-      return `+ ${amountXLM}`
+      return `+ ${amount}`
     case 'senderAndReceiver':
       return '0 XLM'
```
nathunsmitty: I'm not sure if `0 XLM` is the best label for when you're both the sender and receiver. It leads to transactions like these:
Screen Shot 2019-07-01 at 4 02 44 PM
which is technically correct, but doesn't have anything to do with lumens.

nathunsmitty:

patrickxb:
This looks good to me

PR #18217, opened 2019-07-01
Get the new isPromoted field for AirdropDetailsLocal

patrickxb:


patrickxb:
tsx failed on commit.  can you take this branch and use it to hook up to the ui?

cjb:
yup

cjb:
@patrickxb There's a golang failure too though:
```
# github.com/keybase/client/go/systests [github.com/keybase/client/go/systests.test]
systests/stellar_client_test.go:609:12: cannot assign stellar1.AirdropDetails to res (type string) in multiple assignment
systests/stellar_client_test.go:749:5: cannot use (*stellarRetryClient)(nil) (type *stellarRetryClient) as type stellar1.LocalInterface in assignment:
	*stellarRetryClient does not implement stellar1.LocalInterface (wrong type for AirdropDetailsLocal method)
		have AirdropDetailsLocal("context".Context, int) (string, error)
		want AirdropDetailsLocal("context".Context, int) (stellar1.AirdropDetails, error)
systests/teams_test.go:263:18: cannot use newStellarRetryClient(cli) (type *stellarRetryClient) as type stellar1.LocalInterface in assignment:
	*stellarRetryClient does not implement stellar1.LocalInterface (wrong type for AirdropDetailsLocal method)
		have AirdropDetailsLocal("context".Context, int) (string, error)
		want AirdropDetailsLocal("context".Context, int) (stellar1.AirdropDetails, error)
+ ERR=2
+ echo go vet failed with error code 2
go vet failed with error code 2
+ exit 2
```

patrickxb:
i fixed that test file

cjb:

Ready for a JS review now -- this hooks the new `isPromoted` field to a JS prop of `airdropIsLive` for each of the components that want to render something about the airdrop -- the wallet list row for the airdrop, the "lucky airdropee" stars on the profile page, and the airdrop banner.

shared/wallets/wallet-list/index.tsx
```diff
@@ -82,12 +81,13 @@ const WhatIsStellar = (props: {onWhatIsStellar: () => void}) => (
...
 export type Props = {
   accountIDs: Array<AccountID>
+  airdropIsLive: boolean
   airdropSelected: boolean
   style?: Styles.StylesCrossPlatform
   loading: boolean
   inAirdrop: boolean
   onAddNew: () => void
-  onJoinAirdrop: () => void
+  onJoinAirdrop: () => void | null
```
songgao: I think you want `(()=>void)|null`. Otherwise it's `() => (void | null)` IIRC. Seems like this can also be `onJoinAirdrop?: ...`

shared/reducers/wallets.tsx
```diff
@@ -436,7 +436,8 @@ export default function(state: Types.State = initialState, action: WalletsGen.Ac
     case WalletsGen.updateAirdropBannerState:
       return state.merge({airdropShowBanner: action.payload.show})
     case WalletsGen.updatedAirdropDetails:
-      return state.merge({airdropDetails: action.payload.details})
+      const {details, isPromoted} = action.payload
+      return state.set('airdropDetails', Constants.makeAirdropDetails({details, isPromoted}))
```
songgao: How ofter does this happen? If a lot, seems we could use some comparison here, something like

```
      return state.update('airdropDetails', airdropDetails =>
        airdropDetails
          .set('isPromoted', isPromoted)
          .update('details', oldDetails => oldDetails.equals(details) ? oldDetails : details)
      )

```

If not frequent, then I'm OK with this too!

shared/wallets/wallet-list/index.tsx
cjb: Thanks, fixed.

shared/reducers/wallets.tsx
cjb: Ah, looks like just once at app start, so I'll leave it as-is.

PR #18225, opened 2019-07-02
Sep7 UI Tweaks

ddworken:
This addresses the majority of PICNIC-145:

[x] Rounded modal corners on desktop
[x] Delete back button on desktop
[x] BodyTiny → BodySmall 
[x] Disable pay button if they haven't entered an amount yet
[ ] Scrolling
[x] Left align recipient
[x] Margins around stellar address
[x] Black stellar logo
[x] Blue highlight for pending transaction

Note the change for the last one that makes it so all pending transactions are highlighted in blue.


shared/wallets/transaction/index.tsx
```diff
@@ -74,7 +74,11 @@ type CounterpartyTextProps = {
 export const CounterpartyText = (props: CounterpartyTextProps) => {
   switch (props.counterpartyType) {
     case 'airdrop':
-      return <Text style={{color: globalColors.white}} type={props.textTypeSemibold}>Stellar airdrop</Text>
+      return (
```
nathunsmitty: I ran into a similar autoformatting thing in #18214, think your merge conflicts are probably from that

nathunsmitty:
Note that we'll need to cherry-pick this in, see #releases

shared/wallets/transaction/index.tsx
ddworken: :+1: Fixed the merge conflict

ddworken:
Rounded corners: 

image

Blue highlight: 

image

Margins and disabled pay button:

image

Enabled pay button:

image

cjb:
@ddworken Thanks! Could we keep the memo/message/etc vertically aligned instead of indented?  Here's the Zeplin shot:

Screen Shot 2019-07-02 at 8 11 39 AM

PR #18228, opened 2019-07-02
refactor teamEK storage with variant

joshblum:
surgery on the `TeamEKBoxStorage` to store a new variant `TeamEphemeralKey`. This allows us to interact with regular `TeamEKs` as well as new `TeambotEKs`.

cc @keybase/hotpotatosquad


joshblum:
this will clear team EK caches since we change the storage format and key structure. the on disk version number is bumped https://github.com/keybase/client/pull/18228/files#diff-df0df33eb376c53f10d5ae88b890d476R14 and the dbcleaner will clear out old teamek boxes (deviceEKs will still be purged on time making these boxes useless)

PR #18231, opened 2019-07-02
add log tag for GetOrCreateLatestTeamEK

joshblum:
cc @keybase/hotpotatosquad

### DIFF
--- a/go/chat/server.go
+++ b/go/chat/server.go
@@ -98,8 +98,7 @@ func (h *Server) shouldSquashError(err error) bool {
 	// these are not offline errors, but we never want the JS to receive them and potentially
 	// display a black bar
 	switch terr := err.(type) {
-	case storage.AbortedError,
-		libkb.LoginRequiredError:
+	case storage.AbortedError:
 		return true
 	case TransientUnboxingError:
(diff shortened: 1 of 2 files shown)


mmaxim:
Hmm I don't know about this change, what inspired it? How does it affect the CLI?

joshblum:
ah yeah not good for the CLI. @buoyad was getting some blackbars on logout, i think there were some chat calls in progress when logout was triggered and the errors bubbled up

mmaxim:
That shouldn't happen if that RPC canceller stuff is working right. It's supposed to cancel any live RPCs on logout for chat.

joshblum:
i think there is a race between the active device being cleared and the RPC cancelling. Iwonder why `ActiveDevice.Clear()` doesn't happen as one of the last steps https://github.com/keybase/client/blob/master/go/libkb/globals.go#L329

mmaxim:
Ok if that is true, I think that is what we need to fix rather than swallowing these. I intended that squash function for stuff that is just clearly not a problem, like `context.Canceled` and company.

joshblum:
yeah agreed, have a ticket to look into the race more closely

PR #18234, opened 2019-07-02
path payment QA fixes

songgao:
cc @patrickxb / @mlsteele for Go and protocol changes


protocol/avdl/stellar1/common.avdl
```diff
@@ -21,6 +21,7 @@ protocol common {
     string desc; // 'Worthwhile token' or empty
     string infoUrl; // 'https://...' or empty
     string infoUrlText; // 'Exchange' for an exchange, 'View details' for anything else
+    boolean sendDisabled; // set to true if we can't send this as destination asset when we don't own it. For example when auth is required
```
mlsteele: This would make more sense as a field on `Balance`. It's about a user's relationship with an asset. Would that be easy?

go/stellar/format.go
```diff
@@ -142,7 +142,15 @@ func FormatAssetIssuerString(asset stellar1.Asset) string {
 	if asset.VerifiedDomain != "" {
 		return asset.VerifiedDomain
 	}
-	return "Unknown issuer"
+	iaid := asset.IssuerString()
+	iaidLen := len(iaid)
+	if iaidLen > 16 {
+		return iaid[:8] + "..." + iaid[iaidLen-8:]
```
mlsteele: I really think this shortening ought to happen in the ui. Because any time the shortened version is present it should be possible to tap, expand, and copy the full issuer

songgao: we have this in several places and I've fixed them in the UI. but in this case it's a strin gincluding the issuer description from the service, and GUI just display the string

shared/wallets/banner/index.tsx
```diff
@@ -48,7 +48,7 @@ const Banner = (props: Props) => (
         props.onAction && (
           <>
             {props.offerAdvancedSendForm === AdvancedBanner.receiverBanner &&
-              'This user accepts more assets than XLM. '}
+              'This person accepts other assets than XLM. '}
```
mlsteele: "assets other than" reads better than "other assets than" to me. All the ways are weird. Could be a personal thing

protocol/avdl/stellar1/common.avdl
songgao: yeah I think so; will do now

shared/wallets/banner/index.tsx
songgao: So "This person accepts assets other than XLM."? Will do

mlsteele: Cool. Probably want to get an ok from design too as @cecileboucheron has worked on this wording already

shared/wallets/banner/index.tsx
```diff
@@ -48,7 +48,7 @@ const Banner = (props: Props) => (
...
             {props.offerAdvancedSendForm === AdvancedBanner.senderBanner &&
               'You can send more assets than XLM. '}
```
cecileboucheron: @songgao You may want to change this one too: "You can send assets other than XLM."

protocol/avdl/stellar1/common.avdl
songgao: (talked on chat and decided to do it later)

PR #18237, opened 2019-07-03
[Y2K-221] Fix photo picker

jzila:
There was incorrect text on the video picker.

cc @keybase/y2ksquad

### DIFF
--- a/shared/actions/fs/common.native.tsx
+++ b/shared/actions/fs/common.native.tsx
@@ -11,11 +11,19 @@ import {saveAttachmentDialog, showShareActionSheetFromURL} from '../platform-spe
 import {types} from '@babel/core'
 
 const pickAndUploadToPromise = (state: TypedState, action: FsGen.PickAndUploadPayload): Promise<any> =>
-  new Promise((resolve, reject) =>
-    ImagePicker.showImagePicker(
+  new Promise((resolve, reject) => {
+    let title = 'Select a Photo'
+    let captureTitle = 'Take Photo...'
+    if (action.payload.type === Types.MobilePickType.Video) {
+      title = 'Select a Video'
+      captureTitle = 'Take Video...'
+    }
+    return ImagePicker.showImagePicker(
       {
         mediaType: action.payload.type,
         quality: 1,
+        takePhotoButtonTitle: captureTitle,
+        title: title,
         videoQuality: 'high',
       },
       response =>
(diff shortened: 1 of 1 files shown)


shared/actions/fs/common.native.tsx
```diff
@@ -11,11 +11,19 @@ import {saveAttachmentDialog, showShareActionSheetFromURL} from '../platform-spe
...
 const pickAndUploadToPromise = (state: TypedState, action: FsGen.PickAndUploadPayload): Promise<any> =>
-  new Promise((resolve, reject) =>
-    ImagePicker.showImagePicker(
+  new Promise((resolve, reject) => {
+    let title = 'Select a Photo'
+    let captureTitle = 'Take Photo...'
+    if (action.payload.type === Types.MobilePickType.Video) {
+      title = 'Select a Video'
+      captureTitle = 'Take Video...'
+    }
```
songgao: On iOS we used `Mixed`. Not sure if the title is shown there though. If shown, we should probably use a switch/case here and say something like "Select a photo or video". If not shown, I'd prefer not setting them on iOS in case they start to show up at some point in the future.

shared/actions/fs/common.native.tsx
```diff
@@ -11,11 +11,19 @@ import {saveAttachmentDialog, showShareActionSheetFromURL} from '../platform-spe
...
+    return ImagePicker.showImagePicker(
       {
         mediaType: action.payload.type,
         quality: 1,
+        takePhotoButtonTitle: captureTitle,
+        title: title,
```
songgao: Name the variables the same so these can just be in shorthand form?

```
takePhotoButtonTitle,
title,
```

shared/actions/fs/common.native.tsx
```diff
@@ -11,11 +11,19 @@ import {saveAttachmentDialog, showShareActionSheetFromURL} from '../platform-spe
 import {types} from '@babel/core'
 
 const pickAndUploadToPromise = (state: TypedState, action: FsGen.PickAndUploadPayload): Promise<any> =>
-  new Promise((resolve, reject) =>
-    ImagePicker.showImagePicker(
+  new Promise((resolve, reject) => {
+    let title = 'Select a Photo'
+    let captureTitle = 'Take Photo...'
```
songgao: I think design prefer not to capitalize every workd. So "Select a photo" and "Take photo..." here. Same below.

PR #18245, opened 2019-07-03
Disable trustline changes on desktop for mobile-only accounts.

nathunsmitty:
Passes the `thisDeviceIsLockedOut` boolean property to trustline assets, disabling them and displaying a tooltip if the account is changed to mobile only. Note that this case only happens if the trustlines popup is open and then the account is changed to mobile-only; the button in wallet settings to manage trustlines is disabled if the account is mobile-only.

Example of the tooltip:
Screen Shot 2019-07-03 at 12 12 36 PM


nathunsmitty:

PR #18248, opened 2019-07-03
RN 0.59.10

chrisnojima:


chrisnojima:
@keybase/react-hackers this updates rn to 59.10 which includes the jsc we manually added (and nothing more)

PR #18250, opened 2019-07-03
More caching in buildpayment

mlsteele:
Some more of `BuildPaymentCache`'s methods now actually have a cache. Reflecty attempt at a generic cache to avoid copypasta bugs.

### DIFF
--- a/go/stellar/bpc.go
+++ b/go/stellar/bpc.go
@@ -5,7 +5,6 @@ import (
 	"sync"
 	"time"
 
-	lru "github.com/hashicorp/golang-lru"
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/stellar1"
 	"github.com/keybase/client/go/stellar/remote"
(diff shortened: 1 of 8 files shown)


xgess:
weird CI failure

mlsteele:
```
Perhaps you forgot to surround the code with a step that provides this, such as: node
```
uh huh. silly me

go/stellar/bpc.go
```diff
@@ -69,8 +65,17 @@ func (c *buildPaymentCache) AccountSeqno(mctx libkb.MetaContext,
 }
 
 func (c *buildPaymentCache) IsAccountFunded(mctx libkb.MetaContext,
-	accountID stellar1.AccountID) (bool, error) {
-	return isAccountFunded(mctx.Ctx(), c.remoter, accountID)
+	accountID stellar1.AccountID, bid stellar1.BuildPaymentID) (res bool, err error) {
+	fill := func() (interface{}, error) {
+		return isAccountFunded(mctx.Ctx(), c.remoter, accountID)
+	}
+	if !bid.IsNil() {
```
xgess: what's the case when this is nil?

mlsteele: In theory buildPayment can be called with no bid. Though deprecated. Also some back story I thought it was like `GetOutsideCurrencyPreference` which is used by buildRequest, but it's not.

go/stellar/time_cache.go
```diff
@@ -0,0 +1,124 @@
...
+	return c.getHelper(mctx, key, into, fill)
+}
+
+func (c *TimeCache) getHelper(mctx libkb.MetaContext, key string, into interface{}, fill cacheFillFunc) error {
+	lock, err := c.lockTab.AcquireOnNameWithContextAndTimeout(mctx.Ctx(), mctx.G(), key, time.Minute)
+	if err != nil {
+		return fmt.Errorf("could not acquire cache lock for key %v", key)
+	}
+	defer lock.Release(mctx.Ctx())
+	storeResult := func(val interface{}) (ok bool) {
```
xgess: not a huge deal, but preference for this to be a top-level function that takes two parameters. not that i think it'll be re-used, just a little easier to follow this code path and an opportunity for more obvious naming.

mlsteele: So `Get(mctx, key, into, fill) error` where `fill` is optional? The thing I think is weird is this. If you _do_ pass `fill` then the `error` returned should be treated as a failure and likely returned from the caller. But if `fill` is not passed then `error` is just a cache miss.

xgess: just storeResult. storeResult doesn't call `fill`. it puts the thing in the thing if it can. just that.

mlsteele: oh that, yeah sounds good

PR #18252, opened 2019-07-03
GO <-> JSI <-> RN

MarcoPolo:
Draft of hooking up the Keybase engine directly to javascript via the JSI. Android only

### TODO
- [x] Use events to return readB64 data instead of calling callback from separate thread (this is wrong and unsafe)
- [x] Come up with a quantitative measurment

### DIFF
--- a/go/bind/keybase.go
+++ b/go/bind/keybase.go
@@ -324,6 +324,17 @@ func WriteB64(str string) (err error) {
 	return nil
 }
 
+// WriteB64FromC is a wrapper for WriteB64
+//export WriteB64FromC
+func WriteB64FromC(str *C.char) *C.char {
+	var goStr = C.GoString(str)
+	var err = WriteB64(goStr)
+	if err != nil {
+		return C.CString(fmt.Sprintf("Read error: %s", err))
+	}
+	return C.CString("")
+}
+
 // TestNum to test jsi
 //export TestNum
 func TestNum() int {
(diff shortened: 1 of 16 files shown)


MarcoPolo:
I think I've got this in a state where we can do a fair evaluation on it.

I made a ping test component which calls config.getConfig over and over and tracks how long it takes to make those requests and how long.

Some quick observations:
* In the tests it seems like the JSI is only *slightly* faster than the bridge. Not enough to matter for us.
  * The JSI may be able to be optimized further if we minimize the amount of times we string copy and use a buffer instead. Haven't bothered looking into this, but I don't imagine it would affect things by an order of magnitude.

* The UI looks smoother on the bridge version
  * Probably because of the way I'm passing the data back into RN in the JSI version. (I just did the simplest thing).
  * I think this is because in the bridge model you emit the data through the RN event emitter, which probably gives UI work higher priority.

The downsides of using JSI are:
* A new language to deal with (cpp is not friendly)
* A new way to interact with the engine. 
    * It could fail in subtle ways that will be hard to debug, while our current implementation is more "battle tested".
* Linking/Building the JSI/C++ is not trivial
* Won't work with chrome debugger.

------------
### Ping Test Results
Test Format
* Latency is calculated for how long it takes to make 100 roundtrips of getConfig.
* Lower Latency is better
* The ping number update for each round trip
* Smoother UI Updates are better

### JSI
Technically faster, but slower UI updates. (Visible skipping numbers)
Screen Shot 2019-07-09 at 6 43 51 PM

### On the bridge:
A little slower, but the numbers counted up smoothly.
Screen Shot 2019-07-09 at 6 43 59 PM

--------------

I'll try to write a more in depth summary later. But for now I think, given the few upsides and the many downsides, it's not worth switching to using the JSI.

PR #18258, opened 2019-07-03
shortenAccountID passes through bogus input

mlsteele:
TransactionDetails was crashing [here](https://github.com/keybase/client/blob/master/shared/wallets/transaction-details/index.tsx#L324) for relay payments which don't have those path-payment fields set

### DIFF
--- a/shared/constants/wallets.tsx
+++ b/shared/constants/wallets.tsx
@@ -752,7 +752,13 @@ export const getSecretKey = (state: TypedState, accountID: Types.AccountID) =>
     ? state.wallets.exportedSecretKey
     : new HiddenString('')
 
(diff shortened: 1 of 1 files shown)


songgao:
should probably just add a none falsey check here: https://github.com/keybase/client/blob/master/shared/wallets/transaction-details/index.tsx#L329 We'd want to show "Unknown issuer" in that case I assume

cjb:
> Accountid can be null?

It's a "source issuer account ID", rather than the main account one.

mmaxim:
Isn't the typing just completely wrong here then?

mlsteele:
CI looks like it's failing because it thinks go changed. Why?

PR #18266, opened 2019-07-04
Save and search through contact list

zapu:
Save resolutions of contact list to leveldb.

Expected flow is as following:

1. Something triggers contact update flow on the phone. We discussed doing this on foregrounding, but given our existing troubles with app doing too much on startup, maybe we should reconsider other options.
2. App calls `void saveContactsList(int sessionID, array<Contact> contacts);` with contacts fetched using phone OS API.
3. `saveContactsList` resolves contacts using caching contact resolver and saves `array<ProcessedContact>` to leveldb under one key (`DBSavedContacts = 0xb7`)
4. Now app can use `userSearch` RPC or `lookupSavedContactsList` to either search through saved processed contacts or get the entire list ready to be displayed. Search will either work on resolution data (username, full name), or contact info (contact name, components - emails / phone numbers, labels).
5. When app calls saveContactsList again and contacts have not changed, most of the resolutions will come from resolver cache without burning rate limit. Also following info or full names might get updated from UIDMapper calls. So it's important this is called every now and then even if user does not change their phonebook.

I also added some debug commands to help testing this, because right now only `team-building` is hooked up and I didn't do a very good job at this either. Here is how I am testing it:

Make a `contacts.json` file, e.g.:
```
[
  {
    "name": "test2",
    "components": [
      {
        "label": "email",
        "email": "test@keyba.se"
      }
    ]
  },
  {
    "name": "michal",
    "components": [
      {
        "label": "phone",
        "phoneNumber": "+15553411420"
      },
      {
        "label": "work",
        "phoneNumber": "+15553411425"
      }
    ]
  }
]
```

Invoke save RPC using `contacts save` command:
```
cat contacts.json | keybase contacts save 
```

Try cli people search:
```
» keybase people-search --contacts "Michal"
[
  {
    "score": 1,
    "contact": {
      "contactIndex": 1,
      "contactName": "michal",
      "component": {
        "label": "phone",
        "phoneNumber": "+15553411420"
      },
      "resolved": true,
      "uid": "12184a4d1f91737c7e19f310ac8caa19",
      "username": "v83eb975c",
      "fullName": "",
      "following": false,
      "assertion": "15553411420@phone",
      "displayName": "v83eb975c",
      "displayLabel": "michal",
      "rawScore": 1
    },
    "services_summary": null
  },
  {
    "score": 0.5,
    "keybase": {
      "username": "michal0d85",
      "uid": "33c05a7699e206f85c4502e454cceb19",
      "raw_score": 0.018181818181818184,
      "is_followee": false
    },
    "services_summary": {}
  },
]
```


shared/actions/team-building.tsx
```diff
@@ -115,16 +116,30 @@ const search = (state: TypedState, {payload: {namespace}}: NSAction) => {
...
-          id: username,
-          prettyName: fullname,
-          serviceMap: {keybase: username},
+  Promise.all([
+    RPCTypes.contactsLookupSavedContactsListRpcPromise(),
+    RPCTypes.userInterestingPeopleRpcPromise({maxUsers: 50}),
+  ])
+    .then(results => {
+      const contactUsernames = new Set(results[0].map(x => x.username).filter(Boolean))
+      const contacts = results[0].map(
```
zapu: @buoyad maybe we should make the service do all of this work. Initial idea was to have separate RPCs but it looks we will need some logic if we go this road.

shared/actions/team-building.tsx
```diff
@@ -17,6 +17,7 @@ const apiSearch = (
   includeServicesSummary: boolean
 ): Promise<Array<TeamBuildingTypes.User>> => {
   return RPCTypes.userSearchUserSearchRpcPromise({
+    includeContacts: service === 'keybase',
```
buoyad: Should be behind `flags.sbsContacts`

shared/actions/team-building.tsx
```diff
@@ -115,16 +116,30 @@ const search = (state: TypedState, {payload: {namespace}}: NSAction) => {
...
 const fetchUserRecs = (state: TypedState, {payload: {namespace}}: NSAction) =>
-  RPCTypes.userInterestingPeopleRpcPromise({maxUsers: 50})
-    .then((suggestions: Array<RPCTypes.InterestingPerson> | null) =>
-      (suggestions || []).map(
-        ({username, fullname}): TeamBuildingTypes.User => ({
-          id: username,
-          prettyName: fullname,
-          serviceMap: {keybase: username},
+  Promise.all([
+    RPCTypes.contactsLookupSavedContactsListRpcPromise(),
```
buoyad: Here too

shared/actions/team-building.tsx
```diff
@@ -115,16 +116,30 @@ const search = (state: TypedState, {payload: {namespace}}: NSAction) => {
...
+    RPCTypes.userInterestingPeopleRpcPromise({maxUsers: 50}),
+  ])
+    .then(results => {
```
buoyad: Maybe `[contactsRes, suggestionsRes]`

shared/actions/team-building.tsx
buoyad: Hm, ok, I think this is fine for now.

go/service/usersearch.go
```diff
@@ -0,0 +1,209 @@
...
+		return nil, err
+	}
+	return response.List, nil
+}
+
+func normalizeText(str string) string {
+	return strings.ToLower(string(norm.NFKD.Bytes([]byte(str))))
+}
+
+var splitRxx = regexp.MustCompile(`[-\s!$%^&*()_+|~=` + "`" + `{}\[\]:";'<>?,.\/]+`)
```
heronhaye: can we have some testcases with CJK characters?

go/service/usersearch.go
```diff
@@ -0,0 +1,209 @@
...
+	for _, p := range parts {
+		if p != "" {
+			nonEmptyParts = append(nonEmptyParts, p)
+		}
+	}
+	rxx, err := regexp.Compile(".*" + strings.Join(nonEmptyParts, ".*") + ".*")
+	if err != nil {
+		return nil, err
+	}
+	rxx.Longest()
```
heronhaye: cool, never seen this before

shared/actions/team-building.tsx
```diff
@@ -115,16 +116,30 @@ const search = (state: TypedState, {payload: {namespace}}: NSAction) => {
...
+      const suggestions = results[1]
+        .filter(({username}) => !contactUsernames.has(username))
+        .map(
+          ({username, fullname}): TeamBuildingTypes.User => ({
+            id: username,
+            prettyName: fullname,
+            serviceMap: {keybase: username},
+          })
+        )
+      return contacts.concat(suggestions)
```
buoyad: Designs have 5 recommendations preceding the contacts. Want to change that here? I think `maxUsers` should be `flags.sbsContacts && state.settings.contacts.importEnabled ? 5 : 50`.

go/service/usersearch.go
```diff
@@ -0,0 +1,209 @@
...
+		return res, err
+	}
+	res = compiledQuery{
+		query: query,
+		rxx:   rxx,
+	}
+	return res, nil
+}
+
+func (q *compiledQuery) scoreString(str string) (bool, float64) {
```
heronhaye: may be a phase 2 thing but we should consider fuzziness. especially people are more likely to misspell people's names (either in query or their addressbook) than usernames.

go/contacts/phonebook.go
```diff
@@ -0,0 +1,94 @@
...
+// passed as an argument. Always cache the result of last resolution, do not do
+// any result merging.
+
+type SavedContactsStore struct {
+	encryptedDB *encrypteddb.EncryptedDB
+}
+
+func NewSavedContactsStore(g *libkb.GlobalContext) *SavedContactsStore {
+	keyFn := func(ctx context.Context) ([32]byte, error) {
+		return storage.GetSecretBoxKey(ctx, g, storage.DefaultSecretUI)
```
heronhaye: no context string?

go/service/usersearch.go
heronhaye: ok, coyne says not necessary and this is fine for now - nvm

zapu: thanks, good idea though.

shared/constants/team-building.tsx
```diff
@@ -74,6 +74,13 @@ const parseRawResultToUser = (
       prettyName: result.keybase.fullName || result.keybase.username,
       serviceMap,
     }
+  } else if (service === 'keybase' && result.contact) {
```
buoyad: might as well `&& flags.sbsContacts`

PR #18268, opened 2019-07-05
dont remove tracker on location stop too early

mmaxim:


mmaxim:
cc @joshblum if you have a spare second to confirm this is ok

PR #18277, opened 2019-07-08
add runtime stats view for admins

mmaxim:
Add a box to monitor runtime stats of the app. Shows up on mobile and desktop, however desktop only gives stats about the service, while mobile is the entire app. Stats include the size of the process in memory from the OS perspective (just Darwin for now), Go heap stats, and an indicator if the chat background conv loader and indexer sync loops are running or not. 

I think it is useful to keep an eye on, and should be a nice tool in making sure perf regressions (especially of the memory variety) are obvious (since the values are fairly consistent). I plan on always having it enabled. Looks like this:

image

cc @maxtaco @malgorithms @keybase/hotpotatosquad


shared/actions/settings.tsx
```diff
@@ -775,6 +778,12 @@ function* settingsSaga(): Saga.SagaGenerator<any, any> {
   yield* Saga.chainAction<SettingsGen.LoadProxyDataPayload>(SettingsGen.loadProxyData, loadProxyData)
   yield* Saga.chainAction<SettingsGen.SaveProxyDataPayload>(SettingsGen.saveProxyData, saveProxyData)
 
+  // Runtime Stats
+  yield* Saga.chainAction<SettingsGen.ToggleRuntimeStatsPayload>(
```
chrisnojima: This should all belong in config. Not across all these various sub parts.

shared/settings/advanced/index.tsx
```diff
@@ -235,6 +236,14 @@ class Developer extends React.Component<Props, State> {
             style={styles.developerButtons}
           />
         )}
+        {flags.admin && (
+          <Kb.Button
```
chrisnojima: This means never in TestFlight?

shared/app/runtime-stats/container.tsx
```diff
@@ -0,0 +1,41 @@
...
+  hasData: false,
+  resident: '???',
+  residentSeverity: RPCTypes.StatsSeverityLevel.normal,
+  selectiveSyncActive: false,
+  virt: '???',
+}
+
+const mapStateToProps = (state: TypedState) => {
+  return state.config.runtimeStats
+    ? {
```
chrisnojima: Just pull this into a variable so your lines aren't so long

shared/app/runtime-stats/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+          styles.stat,
+          props.selectiveSyncActive ? styles.statWarning : styles.statNormal,
+        ])}
+        type="BodyTiny"
+      >{`IndexerSyncActive: ${yesNo(props.selectiveSyncActive)}`}</Kb.Text>
+    </Kb.Box2>
+  )
+}
+
+export const RuntimeStatsMobile = (props: Props) => {
```
chrisnojima: Just export one thing depending on ismobile. Don't need the container to decide this

shared/app/runtime-stats/container.tsx
```diff
@@ -0,0 +1,41 @@
...
+        hasData: true,
+        resident: state.config.runtimeStats.resident,
+        residentSeverity: state.config.runtimeStats.residentSeverity,
+        selectiveSyncActive: state.config.runtimeStats.selectiveSyncActive,
+        virt: state.config.runtimeStats.virt,
+      }
+    : blank
+}
+
+export default namedConnect(mapStateToProps, () => ({}), s => ({...s}), 'RuntimeStats')(
```
chrisnojima: Just FYI we're moving away from named connect. It's not compatible with hooks. Also FYI you could just use the redux hook for this easily as your just pulling out one var from the store (don't need to change this just FYI)

shared/actions/settings.tsx
mmaxim: Ok, this is on the settings screen so I stuck it here, but can move it.

shared/settings/advanced/index.tsx
mmaxim: Ah right, I guess that won't work. Hmm, I'm not sure what the best way to show this in the mobile app is then.

chrisnojima: We already have the hidden buttons behind the clicks

mmaxim: K I guess that works, will do that.

go/chat/search/indexer.go
```diff
@@ -510,11 +511,17 @@ func (idx *Indexer) Search(ctx context.Context, uid gregor1.UID, query, origQuer
 	return sess.run(ctx)
 }
 
+func (idx *Indexer) IsActivelySyncing() bool {
+	return idx.selectiveSyncActive
```
joshblum: locks

go/chat/search/indexer.go
```diff
@@ -510,11 +511,17 @@ func (idx *Indexer) Search(ctx context.Context, uid gregor1.UID, query, origQuer
...
+}
+
 // SelectiveSync queues up a small number of jobs on the background loader
 // periodically so our index can cover all conversations. The number of jobs
 // varies between desktop and mobile so mobile can be more conservative.
 func (idx *Indexer) SelectiveSync(ctx context.Context, uid gregor1.UID) (err error) {
 	defer idx.Trace(ctx, func() error { return err }, "SelectiveSync")()
+	idx.selectiveSyncActive = true
```
joshblum: lock while modifying

go/chat/types/interfaces.go
```diff
@@ -131,6 +131,7 @@ type Indexer interface {
 	PercentIndexed(ctx context.Context, convID chat1.ConversationID, uid gregor1.UID) (int, error)
 	SearchableConvs(ctx context.Context, uid gregor1.UID, convID *chat1.ConversationID) ([]RemoteConversation, error)
 	OnDbNuke(mctx libkb.MetaContext) error
+	IsActivelySyncing() bool
```
joshblum: Could both be `IsBackgroundActive()`

go/chat/utils/utils.go
```diff
@@ -1235,7 +1235,7 @@ func formatVideoDuration(ms int) string {
 	return fmt.Sprintf("%d:%02d", minutes, seconds)
 }
 
-func formatVideoSize(bytes int64) string {
+func PresentBytes(bytes int64) string {
```
joshblum: could just use `humanize.Bytes` throughout

go/runtimestats/stats.go
```diff
@@ -0,0 +1,137 @@
...
+		r.debug(ctx, "not starting, not enabled in env")
+		return
+	}
+	if r.started {
+		return
+	}
+	r.stopCh = make(chan struct{})
+	r.started = true
+	r.eg.Go(func() error { return r.statsLoop(r.stopCh) })
+	r.G().PushShutdownHook(func() error {
```
joshblum: should do this in the `NewRunner` function or it can get pushed on multiple times

go/chat/utils/utils.go
mmaxim: I already have this.

PR #18282, opened 2019-07-08
dont show federated address qrcodes

xgess:
when receiving a payment, stop showing the federated address qrcode because no one else in the ecosystem knows how to read those qrcodes yet.


xgess:
@keybase/picnicsquad

PR #18284, opened 2019-07-08
libkb: allow more wiggle room in server-side merkle clock jitter

maxtaco:
- it might be the case that the server has some sort of time hiccup, whether due to a leap second, or an NTP problem or whatnot. The clock might in fact go backwards, but it shouldn't go too far backwards
- this commit limits time-travel to 5 minutes between two subsequent skips, and at most 5 minutes over a whole skip sequence
- change tests accordingly
- fix tests to all be autogenerated


maxtaco:
See #18284

maxtaco:
See #18285 I mean

PR #18285, opened 2019-07-08
libkb: allow more wiggle room in server-side merkle clock jitter

maxtaco:
- it might be the case that the server has some sort of time hiccup, whether due to a leap second, or an NTP problem or whatnot. The clock might in fact go backwards, but it shouldn't go too far backwards
- this commit limits time-travel to 5 minutes between two subsequent skips, and at most 5 minutes over a whole skip sequence
- change tests accordingly
- fix tests to all be autogenerated


go/libkb/merkle_client.go
```diff
@@ -1135,14 +1139,29 @@ func (ss SkipSequence) verify(m MetaContext, thisRoot keybase1.Seqno, lastRoot k
 			return MerkleClientError{fmt.Sprintf("Skip pointer mismatch at %d->%d", thisRoot, prevRoot), merkleErrorSkipHashMismatch}
 		}
 
-		// Check that ctimes in the sequence are also strictly ordered
+		// Check that ctimes in the sequence are nearly strictly ordered; we have to make sure we can handle slight
+		// clock jitters since the servrer might need to rewind time due to leap seconds or NTP issues.
```
heronhaye:
```suggestion
		// clock jitters since the server might need to rewind time due to leap seconds or NTP issues.
```

go/libkb/merkle_client.go
```diff
@@ -1135,14 +1139,29 @@ func (ss SkipSequence) verify(m MetaContext, thisRoot keybase1.Seqno, lastRoot k
...
+		// We'll allow at most 5 minutes of "time-travel" between 2 updates, and no more than 5minutes or time travel
```
heronhaye:
```suggestion
		// We'll allow at most 5 minutes of "time-travel" between 2 updates, and no more than 5minutes of time travel
```

go/libkb/merkle_client.go
```diff
@@ -1135,14 +1139,29 @@ func (ss SkipSequence) verify(m MetaContext, thisRoot keybase1.Seqno, lastRoot k
...
 				merkleErrorOutOfOrderCtime,
 			}
 		}
+		if index == 0 {
+			latestCtime = ss[index].ctime()
+		}
+		earliestCtime = ss[index].ctime()
+	}
+
+	if latestCtime > 0 && earliestCtime > 0 && earliestCtime > latestCtime && (earliestCtime-latestCtime > maxClockDriftSeconds) {
```
heronhaye: good idea

PR #18289, opened 2019-07-08
fix windows electron 5

chrisnojima:
The flow is
1. Run heat which generates an xml file
1. Run the xml file through gui.xsl as an xslt transform to build a file wix uses. If we find a directory called 'resources' we give it a special id
1. Have special code to uninstall this due to some weird issue w/ wix (code comment: ` The heat tool output (which generates gui.wxs) won't work if there are directories with only subdirectories
        and no removable components, so we have to special-case them here and in the xsl transform (gui.xsl) `
1. To work around this above comment we just inject a dummy file into that directory

The above seems to work given my light testing and lack of expertise


chrisnojima:
@keybase/react-hackers

PR #18291, opened 2019-07-08
fix settings count w/ notifications badging

chrisnojima:
if you have no push permissions on the notifications settings is badged but isn't part of the nav count


shared/router-v2/router.native.tsx
```diff
@@ -71,16 +71,18 @@ const TabBarIcon = ({badgeNumber, focused, routeName}) => (
...
 const ConnectedTabBarIcon = connect(
-  (state: any, {routeName}: OwnProps) => ({
-    badgeNumber: getBadgeNumber(state.notifications.navBadges, routeName),
-  }),
+  (state, {routeName}: OwnProps) => {
+    const onSettings = routeName === Tabs.settingsTab
+    const badgeNumber = (onSettings ? settingsTabChildren : [routeName]).reduce(
+      (res, tab) => res + (state.notifications.navBadges.get(tab) || 0),
+      // notifications gets badged on native if there's no push, special case
+      onSettings && !state.push.hasPermissions ? 1 : 0
```
chrisnojima: pulling this into badgeState might be the better idea, but keeping this small for now

chrisnojima:

PR #18294, opened 2019-07-08
support inline actions in Kb.Banner and better API

songgao:
I'll have a another PR that uses this in user proof breaking banner.


shared/common-adapters/banner.tsx
```diff
@@ -6,75 +6,108 @@ import * as Styles from '../styles'
...
+type Segment = _Segment | string | null | false
+
+// See banner.stories.tsx for examples
+type Content =
+  // single paragraph content
+  | string
+  // single paragraph content with an array of segments
+  | (Array<Segment>)
+  // an array of paragraphes, each represented by an array of segments
+  | Array<Array<Segment>>
```
chrisnojima: Lets not do this. Let's chat

songgao:
Made the change, but I'm too lazy to go back to all places that use it to change them, so will keep them that way. The story is indeed updated though.

PR #18295, opened 2019-07-08
add a payment popup for pending transactions

xgess:
### before
pending transactions are unclickable. when it's only in this state for a second or two, it's hard to notice and not a big thing. but if it ever takes longer (or if we're having issues propagating tx state correctly), the user experience can be sub-optimal.

### after
do something really similar to how it normally works and err on the side of showing more information rather than less. i took a stab at a design that follows the same pattern of pending being grey instead of purple, and putting the word `pending` in there really obviously. and again, when everything is working normally, this will only be visible like this for a few seconds. 

Screen Shot 2019-07-11 at 12 11 10 PM


xgess:
@keybase/picnicsquad

PR #18298, opened 2019-07-09
Fix null arrays in team-building

buoyad:
cc @keybase/y2ksquad


shared/actions/team-building.tsx
```diff
@@ -123,7 +123,7 @@ const fetchUserRecs = (state: TypedState, {payload: {namespace}}: NSAction) =>
       ? RPCTypes.contactsLookupSavedContactsListRpcPromise()
       : Promise.resolve([] as RPCTypes.ProcessedContact[]),
   ])
-    .then(([suggestionRes, contactRes]) => {
+    .then(([suggestionRes = [], contactRes = []]) => {
```
jzila: TIL! Nice :)

buoyad: actually, there's a problem here. I think this only works if the array passed in is of a shorter length. Since this one contains `null` it doesn't fix the problem. Will get another one up shortly...

PR #18299, opened 2019-07-09
Better null check in team-building

buoyad:
There's a problem with the other fix, see [here](https://github.com/keybase/client/pull/18298#discussion_r301354942) @keybase/react-hackers @keybase/y2ksquad


zapu:
oops, sorry! does it return null when there are no suggestions / contacts?

buoyad:
@zapu at least for contacts, yeah. More a problem with our TS config.

PR #18301, opened 2019-07-09
add FS_GUI identify behavior and a ProofBroken banner

songgao:
With this PR, viewing TLFs with broken proofs doesn't error anymore. It adds a banner that shows up when any user that shows up in the TLF path (i.e. impl team users only) is known to be broken according to `state.users.infoMap`. However, no mechanism is in place to actually plumb that data from Files tab activities. In other words, TLFs with broken proofs just work, without any warning at all (unless we know about the breaks from other tabs such as Profile or Chat), which is kind of dangerous. I've created HOTPOT-300 to track actually getting that data through notification or some on-demand ways. But if that ends up delayed, we should probably unset the identify behavior from this PR before next release.

cc @keybase/hotpotatosquad


protocol/avdl/keybase1/simple_fs.avdl
```diff
@@ -264,59 +264,59 @@ protocol SimpleFS {
...
+  void simpleFSSetStat(Path dest, DirentType flag, union{ null, TLFIdentifyBehavior } identifyBehavior);
 
   /**
    Read (possibly partial) contents of open file,
    up to the amount specified by size.
    Repeat until zero bytes are returned or error.
    If size is zero, read an arbitrary amount.
    */
-  FileContent simpleFSRead(OpID opID, long offset, int size);
+  FileContent simpleFSRead(OpID opID, long offset, int size, union{ null, TLFIdentifyBehavior } identifyBehavior);
```
strib: Read and Write don't need this param, because it should already have been triggered by Open I think.  Seems like a good rule of thumb is that it should go along with any Path that can be a KBFS path.

You've already done the work here to plumb it through as a separate param, which is fine with me, but perhaps in the future, or if you like the idea a lot better, we could make a `KbfsPath` type in the avdl that contains both a string and the identify behavior?  That would probably be a significant code change though, since it's currently just a bare string everywhere.

go/protocol/keybase1/extras.go
```diff
@@ -1266,6 +1266,7 @@ func (b TLFIdentifyBehavior) AlwaysRunIdentify() bool {
 	switch b {
 	case TLFIdentifyBehavior_CHAT_CLI,
 		TLFIdentifyBehavior_CHAT_GUI,
+		TLFIdentifyBehavior_FS_GUI,
```
strib: Seems like you didn't add `FS_GUI` to the function below this one, `CanUseUntrackedFastPath()`.  I'm not sure exactly what that does, but did you try it to see if it helps with the broken track reporting?

go/kbfs/simplefs/simplefs.go
```diff
@@ -1114,17 +1127,26 @@ func (k *SimpleFS) doCopy(
...
 		func(ctx context.Context) (err error) {
 			return k.doCopy(ctx, arg.OpID, arg.Src, arg.Dest)
 		})
 }
 
 // SimpleFSSymlink starts making a symlink of a file or directory
-func (k *SimpleFS) SimpleFSSymlink(ctx context.Context, arg keybase1.SimpleFSSymlinkArg) (err error) {
+func (k *SimpleFS) SimpleFSSymlink(
+	ctx context.Context, arg keybase1.SimpleFSSymlinkArg) (err error) {
+	if arg.IdentifyBehavior != nil {
```
strib: Why not pass the identify behavior into `startSyncOp` instead of handling it separately in each function?  Actually, it would be even better if it was an argument to `makeContext`, which is called in both places, and others too.

protocol/avdl/keybase1/simple_fs.avdl
songgao: That's a great idea! I don't think it'd be as bad as my current plumbing. The identify behavior is optional, so we'll only need to add that in GUI. Will try to do that. Thanks!

go/protocol/keybase1/extras.go
songgao: I _think_ I tried. Will try again. I don't know what it is either

protocol/avdl/keybase1/simple_fs.avdl
songgao: Ah I see what you mean here. It's a variant and the kbfs case is just a string. Will try and see how bad it is

songgao:
@strib I'm gonna try the CHAT_GUI later. PTAL the rest of the changes; thanks!

songgao:
> Did you consider plumbing the paths all the way to makeContext so it could be used by any RPC without putting explicit identify logic in the ones that don't use ops?

Yes; the reason why I didn't do it was that makeContext doesn't have an err now, and I thought it was nice to be able to use that inline for other functions args. I guess it's not a real benefit though. I can change it when fixing CI issues.

shared/chat/conversation/bottom-banner/container.tsx
```diff
@@ -41,11 +39,7 @@ const mapStateToProps = (state, {conversationIDKey}) => {
...
 }
 
-const mapDispatchToProps = dispatch => ({
-  onClick: isMobile
-    ? (username: string) => dispatch(ProfileGen.createShowUserProfile({username}))
-    : (username: string) => dispatch(Tracker2Gen.createShowUser({asTracker: true, username})),
-})
+const mapDispatchToProps = dispatch => ({})
 
 const mergeProps = (stateProps, dispatchProps) => {
```
jakob223: Nit: you can delete the `dispatchProps` parameter. Also, while you're here, can you move the `mergeProps` to make it typed?

songgao: Good call. I think it was already moved upstream, so I just rebased.

PR #18305, opened 2019-07-09
Add generic encrypteddb.GetSecretBoxKey

zapu:
Replace code that used to call chat/storage.GetSecretBoxKey but was not related to chat. Add "generic" function to aid with encrypting local storage. Note that this is **not new crypto**, it's merely a function that was already being present in two places (`go/chat/storage/crypt.go` and `go/teams/storage/generic.go`) that's now copied into third place. It should be used by anyone who wants to encrypt leveldb data instead of making their own variant.
 
There was a mixup on how we were doing encrypted caches. `chat/storage.GetSecretBoxKey` was being used outside of chat, and once it's been used once, more occurrences followed by people finding that call site and doing it similar way. I don't have the exact timeline here and it doesn't matter.

Contact cache will not be encrypted using key derived with `ChatLocalStorage` string anymore. Added new derivation string for contacts: `EncryptionReasonContactsLocalStorage = "Keybase-Contacts-Local-Storage-1"`

Use the new function in `go/offline/rpc_cache.go` but keep `EncryptionReasonChatLocalStorage` so users will not lose cached data.

### DIFF
--- a/go/chat/storage/crypt.go
+++ b/go/chat/storage/crypt.go
@@ -1,7 +1,7 @@
 package storage
 
 import (
-	"github.com/keybase/client/go/engine"
+	"github.com/keybase/client/go/encrypteddb"
 	"github.com/keybase/client/go/libkb"
 	"golang.org/x/net/context"
 )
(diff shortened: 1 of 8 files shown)


joshblum:
what's the reasoning for copying to a third place instead of refactoring teams/chat to use this now?

zapu:
changing teams/chat to use this is still on the table. i was approaching this very carefully so if anything broke, it wouldn't be chat or teams storage

zapu:
@maxtaco can you take a look as well? it touches some team storage code that you were working in October 2018. 

also cc @mmaxim

PR #18307, opened 2019-07-09
Render some explanatory text when there's no data on attachments.

adamjspooner:
@keybase/react-hackers, this is ready for review. I'm open to suggestions if there's a better way.


chrisnojima:
when does this happen

Issue #18300, opened 2019-07-09
Pushing to a git repo fails - worktree should be provided


avioli:
```

> ls -la
drwxr-xr-x   - avioli  9 Jul 10:50 .git
.rw-r--r-- 150 avioli  9 Jul 10:11 docker-compose.yaml
.rw-r--r-- 260 avioli  9 Jul 10:09 Dockerfile
.rw-r--r-- 184 avioli  9 Jul 10:28 Home.md
.rw-r--r--  32 avioli  9 Jul  9:45 readme.md

> keybase git create --team=stamatovs wiki
Repo created! You can clone it with:
  git clone keybase://team/stamatovs/wiki
Or add it as a remote to an existing repo with:
  git remote add origin keybase://team/stamatovs/wiki

> git remote add origin keybase://team/stamatovs/wiki

> git push -u origin master
Initializing Keybase... done.
Syncing with Keybase... This operation conflicted with another operation on the server, or it took too long. Please try again.
git-remote-keybase error: (1) This operation conflicted with another operation on the server, or it took too long. Please try again.

> git push -u origin master
Initializing Keybase... done.
Syncing with Keybase... worktree should be provided
git-remote-keybase error: (1) worktree should be provided

> git push -u origin master
Initializing Keybase... done.
Syncing with Keybase... worktree should be provided
git-remote-keybase error: (1) worktree should be provided

> keybase --version
keybase version 4.1.0-20190612165948+952fee6c59
```

my log id: 63151da48f044a9354c5fd1c

strib:
Thanks for the report.  I see the issue, it's a bug in our code triggered by a slower-than-expected connection to one of our servers from your client.  I'll fix the bug on our end.

On your end, you should be able to fix it by deleting the repo with `keybase git delete`, and then recreating it.  If you don't get the "This operation conflicted" error again, you should be good to go.  Sorry for the inconvenience!

PR #18310, opened 2019-07-09
folder_branch_ops: don't assume offline mode when trying to lock

strib:
For offline mode, we put in a quick check in `SyncFromServer()` that bails out early if there's a long timeout trying to reach the mdserver.  However, this function is used for real during git operations, and if `SyncFromServer()` returns `nil`, it's assumed that a lock was taken at the server.  So if a client takes too long in pinging the mdserver and doesn't actually take the lock, the later `Put` to the mdserver will fail.  If this is the first put for initializing a repo, the repo will be left in a weird state where the initial config file was written, but none of the initial git repo files, and future actions will fail.

So instead, just don't test the timeout logic when we're trying to take a lock.

Issue: #18300

### DIFF
--- a/go/kbfs/libkbfs/folder_branch_ops.go
+++ b/go/kbfs/libkbfs/folder_branch_ops.go
@@ -7377,7 +7377,7 @@ func (fbo *folderBranchOps) SyncFromServer(ctx context.Context,
 		timedOut = true
 	default:
 	}
-	if timedOut || !mdserver.IsConnected() {
+	if lockBeforeGet == nil && (timedOut || !mdserver.IsConnected()) {
 		fbo.vlog.CLogf(
 			ctx, libkb.VLog1, "Not fetching new updates while offline")
 		return nil


MarcoPolo:
Wow that's kinda subtle. Shouldn't the SyncFromServer function return something different is it timed out?

strib:
Whoa @MarcoPolo coming in from left field for the Golang review, thanks!

> Wow that's kinda subtle. Shouldn't the SyncFromServer function return something different is it timed out?

Maybe.  The problem is that this function is used a lot in tests, and if it returns an error in the timeout case, it will completely break all the offline tests that depend on it.  Plumbing through the fact that a timeout error is acceptable in all those places would be a pretty huge undertaking.

MarcoPolo:
Yeah, I get that. It was just a comment. Maybe we can have a field on fbo called `lastSync` and it's a timestamp or nil. Then the caller would know if the sync worked or failed

strib:
The problem is that generally the caller is at least two levels removed from the fbo, and can't easily check a field like that.  Plus, if two syncs happened at once, that would be overwritten.  A specific error type or return param would be ideal, but it's just a pain to plumb it.

PR #18315, opened 2019-07-09
Better handling of stellar request message limits

nathunsmitty:
This PR fixes an issue in which a request message over 240 bytes was allowed by the frontend but blocked by the client, resulting in a black bar.

The reason for this issue is that we had separate definitions for the note length in the frontend and client code. For requests, this number was out of sync. Rather than just update the frontend number, I figured it made more sense to retrieve the number from the client on start and place it in the store. We currently do this for a few chat values using an RPC called `getStaticConfig`.

A problem with extending `getStaticConfig` though – because it retrieves not just chat values now, it doesn't really make sense to keep it defined in the `chat1` namespace. My next choice was `keybase1/config`, but because `chat1` depends on definitions in `keybase1` and `getStaticConfig` depends on chat type definitions in `chat1`, this creates a circular dependency that causes our avdl files to not compile. The only two solutions I could think of to solve this problem are:
1. Keep `getStaticConfig` defined in `chat1`. This results in the smallest diff, but doesn't really make a ton of logical sense.
2. Move `getStaticConfig` into a new namepace (e.g., `config1` or something similarly named). Pros: makes more logical sense and is more extendable, but I'm not sure if config RPCs are really worthy of a whole separate namespace.

Currently, 1 is implemented in this PR. Would be interested in getting opinions on what to do here.


go/client/wallet_api_handler.go
```diff
@@ -465,7 +465,7 @@ func (c *sendOptions) Check() error {
 	if len(c.Message) > libkb.MaxStellarPaymentNoteLength {
 		return ErrMessageTooLong
 	}
-	if len(c.MemoText) > 28 {
+	if len(c.MemoText) > libkb.MaxStellarPaymentPublicNoteLength {
```
nathunsmitty: I tried to remove as many magic `28`s, `500`s, and `240`s in the Go code and replace them with constants for maintainability.

cjb:
@nathunsmitty I'm okay with this approach, but I think there's a third option I like best -- leave chat1's getStaticConfig RPC alone and add a new one inside stellar1, called the same way but from inside actions/wallets.js:
```js
  yield* Saga.chainGenerator<ConfigGen.DaemonHandshakePayload>(
    ConfigGen.daemonHandshake,
    loadStaticConfig,
    'loadStaticConfig'
  )
```

nathunsmitty:
Alright, restructured this so that there are two `getStaticConfig` RPCs: one for chat and one for wallets.

PR #18316, opened 2019-07-09
teams: feature-flagging for hidden chain rotation on a per-team basis

maxtaco:
- for now, don't hook CLKR into this, we'll be doing that in future PRs (since tests don't currently pass)
- if an admin on prod, hit the server and ask if we can use hidden links for this team

### DIFF
--- a/go/teams/hidden/hidden.go
+++ b/go/teams/hidden/hidden.go
@@ -220,10 +220,73 @@ func generateKeyRotationSig3(mctx libkb.MetaContext, p GenerateKeyRotationParams
 	return &bun, ratchets, nil
 }
 
-func CheckFeatureGateForSupport(mctx libkb.MetaContext, teamID keybase1.TeamID, isWrite bool) (err error) {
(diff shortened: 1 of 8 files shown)


go/teams/hidden/hidden.go
```diff
@@ -220,10 +220,73 @@ func generateKeyRotationSig3(mctx libkb.MetaContext, p GenerateKeyRotationParams
...
+		return rt, nil
+	}
+	ok, err := checkFeatureGateForSupport(mctx, teamID, isWrite)
+	if err != nil {
+		return rt, err
+	}
+
+	switch {
+	case rt == keybase1.RotationType_CLKR && !ok:
+		return keybase1.RotationType_VISIBLE, nil
```
heronhaye: is this intended to return nil error?

maxtaco: Yes. On prod, we're only allowing CLKRs to go through the hidden chain if it's explicitly allowed by the server.

maxtaco:
merging over flake

PR #18322, opened 2019-07-10
Airdrop global banner for desktop

cjb:
@keybase/picnicsquad CC @keybase/react-hackers @keybase/design 

Here's the main desktop airdrop banner.  I tried having the macOS version of the banner fit on a single row as in the Zeplin design -- the reason it didn't work is that we have a macOS-only Electron bug where the pointer doesn't turn into a hand if a clickable area is inside the top 40px or so of the screen.  So I don't think that'd make for good UX.

This PR means the nav header's connected now, so it can see whether there's a live airdrop.  That header already included other connected components (the KBFS uploading status) so I think that's okay.

### DIFF
--- a/shared/constants/platform.desktop.tsx
+++ b/shared/constants/platform.desktop.tsx
@@ -11,6 +11,7 @@ export const isDarwin = platform === 'darwin'
 export const isWindows = platform === 'win32'
 export const isLinux = platform === 'linux'
 export const isAndroidNewerThanN = false
+export const isMac = isDarwin && !isIOS
 export const shortcutSymbol = isDarwin ? '⌘' : 'Ctrl-'
 
 export const defaultUseNativeFrame = isDarwin || isLinux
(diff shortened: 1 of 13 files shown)


shared/constants/platform.desktop.tsx
```diff
@@ -11,6 +11,7 @@ export const isDarwin = platform === 'darwin'
 export const isWindows = platform === 'win32'
 export const isLinux = platform === 'linux'
 export const isAndroidNewerThanN = false
+export const isMac = isDarwin && !isIOS
```
nathunsmitty: I never knew `isDarwin` was true on mobile! 🤯

cecileboucheron:
@nathunsmitty 
That was just to simulate Windows right? 
The Mac controls shouldn't be shown to the left on Windows.

cjb:
Yep! Just simulating, poorly.

PR #18329, opened 2019-07-10
Fix sign out.

adamjspooner:
@keybase/react-hackers, this is ready for review. It broke when I updated the `UpdatePassword` screen.

cc @zapu

### DIFF
--- a/shared/common-adapters/rounded-box.tsx
+++ b/shared/common-adapters/rounded-box.tsx
@@ -8,13 +8,14 @@ const Kb = {
 
 type Props = {
   children: React.ReactNode
-  side: 'bottom' | 'middle' | 'top'
+  side?: 'bottom' | 'middle' | 'top'
 }
 
 const RoundedBox = (props: Props) => (
   <Kb.Box2
     direction="vertical"
     style={Styles.collapseStyles([
+      styles.default,
       props.side === 'bottom' && styles.bottom,
       props.side === 'middle' && styles.middle,
       props.side === 'top' && styles.top,
@@ -30,6 +31,7 @@ const roundedBox: Styles.StylesCrossPlatform = {
   borderBottomWidth: 1,
   borderColor: Styles.globalColors.greyDark,
   borderLeftWidth: 1,
+  borderRadius: Styles.borderRadius,
   borderRightWidth: 1,
   borderStyle: 'solid',
   borderTopWidth: 1,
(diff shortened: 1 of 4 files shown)


shared/settings/logout/index.tsx
```diff
@@ -15,133 +15,124 @@ export type Props = {
...
-        </Kb.Box2>
-        {!this.props.checkPasswordIsCorrect ? (
-          <Kb.Box2 direction="vertical">
-            <Kb.ButtonBar align="center" direction="row" fullWidth={true}>
-              {!Styles.isMobile && (
-                <Kb.Button onClick={this.props.onCancel} label="Cancel" fullWidth={true} type="Dim" />
-              )}
+    const inputType = this.state.showTyping ? 'text' : 'password'
+    const keyboardType = this.state.showTyping && Styles.isAndroid ? 'visible-password' : 'default'
+    return this.props.hasRandomPW == null ? (
```
chrisnojima: 1. don't ever use `==`
1. just check the falseyness of it. `!this.props.hasRandomPW ? `

shared/settings/logout/index.tsx
```diff
@@ -15,133 +15,124 @@ export type Props = {
...
+          ),
+        ]}
+        footer={{
+          content: !this.props.checkPasswordIsCorrect ? (
+            <Kb.ButtonBar align="center" direction="column" fullWidth={true} style={styles.buttonBar}>
               <Kb.WaitingButton
                 fullWidth={true}
                 waitingKey={Constants.checkPasswordWaitingKey}
-                disabled={!!this.props.checkPasswordIsCorrect}
+                disabled={!!this.props.checkPasswordIsCorrect || this.state.password === ''}
```
chrisnojima: same here, lets avoid explicit checks against `''` or null or undefined etc

shared/common-adapters/rounded-box.tsx
```diff
@@ -8,13 +8,14 @@ const Kb = {
 
 type Props = {
   children: React.ReactNode
-  side: 'bottom' | 'middle' | 'top'
+  side?: 'bottom' | 'middle' | 'top'
```
chrisnojima: why make this optional

shared/settings/logout/index.tsx
adamjspooner: This was just copied over. I wasn't sure if the `==` was the original author's intent. I'll fix it.

shared/common-adapters/rounded-box.tsx
adamjspooner: You might want a fully rounded box (as I'm using on the “Do you know your password?” input).

chrisnojima: maybe just add 'all'?

PR #18337, opened 2019-07-10
Surya/y2 k 28/blindtree types

heronhaye:
These are just helpers types but mostly finalized. Rest of work is benchmarking and optimizing serverside.

Let me know if you see parts that could be smaller to reduce bandwidth.

### DIFF
--- a/go/badges/badgestate.go
+++ b/go/badges/badgestate.go
@@ -34,10 +34,10 @@ func (d dummyLocalChatState) ApplyLocalChatState(ctx context.Context, i []keybas
 // See service:Badger for the service part that owns this.
 type BadgeState struct {
 	sync.Mutex
-	libkb.Contextified
 
 	localChatState LocalChatState
 	log            logger.Logger
+	env            *libkb.Env
 	state          keybase1.BadgeState
 
 	inboxVers chat1.InboxVers
@@ -48,9 +48,20 @@ type BadgeState struct {
 }
 
 // NewBadgeState creates a new empty BadgeState.
-func NewBadgeState(g *libkb.GlobalContext) *BadgeState {
+func NewBadgeState(log logger.Logger, env *libkb.Env) *BadgeState {
+	return newBadgeState(log, env)
+}
+
+// NewBadgeState creates a new empty BadgeState in contexts
+// where notifications do not need to be handled.
+func NewBadgeStateForServer(log logger.Logger) *BadgeState {
+	return newBadgeState(log, nil)
+}
+
+func newBadgeState(log logger.Logger, env *libkb.Env) *BadgeState {
 	return &BadgeState{
-		Contextified:    libkb.NewContextified(g),
+		log:             log,
+		env:             env,
 		inboxVers:       chat1.InboxVers(0),
 		chatUnreadMap:   make(map[string]keybase1.BadgeConversationInfo),
 		walletUnreadMap: make(map[stellar1.AccountID]int),
@@ -216,7 +227,7 @@ func (b *BadgeState) UpdateWithGregor(ctx context.Context, gstate gregor.State) 
 			byt := item.Body().Bytes()
 			dec := json.NewDecoder(bytes.NewReader(byt))
 			if err := dec.Decode(&tmp); err != nil {
-				b.G().Log.CDebugf(ctx, "BadgeState got bad home.state object; error: %v; on %q", err, string(byt))
+				b.log.CDebugf(ctx, "BadgeState got bad home.state object; error: %v; on %q", err, string(byt))
 				continue
 			}
 			sentUp := false
@@ -225,16 +236,16 @@ func (b *BadgeState) UpdateWithGregor(ctx context.Context, gstate gregor.State) 
 				b.state.HomeTodoItems = countKnownBadges(hsb.BadgeCountMap)
 				sentUp = true
 			}
-			b.G().Log.Debug("incoming home.state (sentUp=%v): %+v", sentUp, tmp)
+			b.log.Debug("incoming home.state (sentUp=%v): %+v", sentUp, tmp)
 		case "tlf":
 			jsw, err := jsonw.Unmarshal(item.Body().Bytes())
 			if err != nil {
-				b.G().Log.CDebugf(ctx, "BadgeState encountered non-json 'tlf' item: %v", err)
+				b.log.CDebugf(ctx, "BadgeState encountered non-json 'tlf' item: %v", err)
 				continue
 			}
 			itemType, err := jsw.AtKey("type").GetString()
 			if err != nil {
-				b.G().Log.CDebugf(ctx, "BadgeState encountered gregor 'tlf' item without 'type': %v", err)
+				b.log.CDebugf(ctx, "BadgeState encountered gregor 'tlf' item without 'type': %v", err)
 				continue
 			}
 			if itemType != "created" {
@@ -244,7 +255,7 @@ func (b *BadgeState) UpdateWithGregor(ctx context.Context, gstate gregor.State) 
 		case "kbfs_tlf_problem_set_count", "kbfs_tlf_sbs_problem_set_count":
 			var body problemSetBody
 			if err := json.Unmarshal(item.Body().Bytes(), &body); err != nil {
-				b.G().Log.CDebugf(ctx, "BadgeState encountered non-json 'problem set' item: %v", err)
+				b.log.CDebugf(ctx, "BadgeState encountered non-json 'problem set' item: %v", err)
 				continue
 			}
 			b.state.RekeysNeeded += body.Count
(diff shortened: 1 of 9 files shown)


go/client/cmd_team_rotate_key.go
```diff
@@ -29,7 +29,7 @@ func newCmdTeamRotateKey(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli
 		},
 		Flags: []cli.Flag{
 			cli.BoolFlag{
-				Name:  "h, hidden",
+				Name:  "hidden",
```
heronhaye: panicd because h conflicts with help

go/merkle/encoder.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+func (e *Encoder) Hash(preimage BlindedPreimage) ([]byte, error) {
+	b, err := preimage.LeafContainer.Serialize()
+	if err != nil {
+		return nil, err
+	}
+	switch e.encodingType {
+	case EncodingTypeBlindedSHA256v1:
```
maxtaco: why do we need this?

go/merkle/encoder.go
```diff
@@ -0,0 +1,71 @@
...
+		h := hmac.New(sha256.New, preimage.BlindedEntropy)
+		h.Write(b)
+		return h.Sum(nil), nil
+	case EncodingTypeBlindedSHA512_256v1:
+		h := hmac.New(sha512.New512_256, preimage.BlindedEntropy)
```
maxtaco: elsewhere we've been using HMAC-SHA512 truncated to the first 32-bytes, which is different. Can we use that here to for HMAC?  For hashing below we can use 512/256

heronhaye: ok, so use sha512[:32] for the blinding hashes, but sha512/256 for the tree hashes?

go/merkle/encoder.go
heronhaye: no reason in particular, can get rid of it

go/merkle/types.go
```diff
@@ -0,0 +1,175 @@
...
+type Leaf interface {
+	Serialize() ([]byte, error)
+	Type() LeafType
+	ID() []byte
+	GetSeqno() keybase1.Seqno
+}
+
+type Chain17v1Leaf struct {
+	_struct bool `codec:",toarray"`
+	TeamID  sig3.TeamID
```
heronhaye: by the way, I think we can get rid of this...the msgpack includes the full key in the leaf's child map

heronhaye: although it's only a constant space factor in the end

go/merkle/encoder.go
maxtaco: 👍 yeah let's keep it down to as few options and features as possible

go/merkle/encoder.go
maxtaco: I think this would be more consistent with what we're doing now. Thanks

go/merkle/types.go
maxtaco: oh that's a good point. I'd be in favor of shaving it off then

go/merkle/util.go
```diff
@@ -0,0 +1,68 @@
...
+	for diff > 0 {
+		if diff&1 == 1 {
+			start += z
+			jumps = append(jumps, start)
+		}
+		diff >>= 1
+		z *= 2
+	}
+
+	for i := len(jumps)/2 - 1; i >= 0; i-- {
```
maxtaco: comment that this is an array reversal pls

go/badges/badgestate.go
```diff
@@ -299,12 +310,12 @@ func (b *BadgeState) UpdateWithGregor(ctx context.Context, gstate gregor.State)
...
 			var body []teamDeletedBody
 			if err := json.Unmarshal(item.Body().Bytes(), &body); err != nil {
-				b.G().Log.CDebugf(ctx, "BadgeState unmarshal error for team.delete item: %v", err)
+				b.log.CDebugf(ctx, "BadgeState unmarshal error for team.delete item: %v", err)
 				continue
 			}
 
 			msgID := item.Metadata().MsgID().(gregor1.MsgID)
-			username := b.G().Env.GetUsername().String()
+			username := b.env.GetUsername().String()
```
joshblum: should you check `env != nil` here? what will the server pass in?

heronhaye: it will be panic if constructed `NewBadgeStateForServer`, but server shouldn't be calling this function right? Should I make it return an error instead?

joshblum: the username check is just so the UI doesn't show a badge for the team you just deleted, i'd just set `username = ""` if `env == nil`

joshblum: 🤦‍♂️ missed `NewBadgeStateForServer` but yeah think it's safer to explicitly check

go/msgpack/msgpack.go
```diff
@@ -10,6 +10,13 @@ import (
 	"github.com/keybase/go-codec/codec"
 )
 
+func EncodeCanonical(src interface{}) (dst []byte, err error) {
+	ch := kbcrypto.CodecHandle()
+	ch.Canonical = true
```
maxtaco: Is this something we have on everywhere? How does it differ from what we are currently doing?

go/badges/badgestate.go
```diff
@@ -48,9 +48,20 @@ type BadgeState struct {
 }
 
 // NewBadgeState creates a new empty BadgeState.
-func NewBadgeState(g *libkb.GlobalContext) *BadgeState {
+func NewBadgeState(log logger.Logger, env *libkb.Env) *BadgeState {
```
maxtaco: why changes here?

go/msgpack/msgpack.go
heronhaye: We do it in some places, like for box public summaries. The difference is just that every encoding of the same go struct will end with the same bytes. Without it I was running into trouble because when I reconstructed leafs, encoded and hashed them, they weren't matching the server's commitment.

go/badges/badgestate.go
heronhaye: I needed to update protocols to get the sig3 chaintype, and that spiraled out into me having to update all of client. This function is used by both server and client and was recently changed to use `g`, so after I vendored I had to change it back so server could call it without `g`.

go/merkle/types.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+type SigID []byte
+type Chain17v1Leaf struct {
+	_struct bool `codec:",toarray"`
+
+	// do not encode teamID; it's redundant in the tree
+	TeamID sig3.TeamID `codec:"-"`
+
+	SigID  SigID
```
maxtaco: Do we really need the sigID I wonder? We never in practice use them, only the `LinkID`s

go/msgpack/msgpack.go
maxtaco: I don't understand this: "every encoding of the same go struct will end with the same bytes". What does this mean?

go/merkle/types.go
heronhaye: I don't think it's needed but it also doesn't hurt much to leave in, since it isn't in the tree and generated on demand

heronhaye:
>I don't understand this: "every encoding of the same go struct will end with the same bytes". What does this mean?
(can't inline reply for some reason)
for example, if you have `type foo struct {a int, b int}`, without the canonical encoding, the library might put `a` first in some serializations or `b` first in other ones. This messes up checking hashes.

maxtaco:
OK, I understand what you wrote above. I think in general we solve this problem by putting the struct fields in sorted order. It's also worth doing that

go/merkle/types.go
maxtaco: Why doesn't this have `ChainType` too?  Where is that data going to live, assuming we have multiple chains for each team?

Also, it would be great if we can use the same code / representation as in `hidden/ratchet.go` but is that not possible?

heronhaye: ChainType is in the type of the leaf. What's actually in the tree is this

```
type LeafContainer struct {
    _struct   bool      `codec:",toarray"`
    LeafType  LeafType  // specifies structure of leafBytes
    LeafBytes LeafBytes // msgpack deserialization implements Leaf
}
```
```
type LeafType uint16

const (
    LeafTypeChain17v1 = 1
)
```

>Also, it would be great if we can use the same code / representation as in hidden/ratchet.go but is that not possible?
Not sure exactly what you mean, do you want to use sig3.Tail here?

maxtaco: Yeah, though the fields are different in a sig3.Tail. It's too bad that off the bat, we have two totally different types of ratchets. The fields in the `sig3.Tail` are:

```type Tail struct {
	Hash      LinkID    `codec:"h"` // hash of the outer link
	Seqno     Seqno     `codec:"s"`
	ChainType ChainType `codec:"t"`
}```

heronhaye:
merging this over broken kbfs server build

PR #18338, opened 2019-07-10
show emails in account cli

heronhaye:
btw i made a commit to move upload-avatar but keep it hidden in top-level for backwards compat. if you disagree we can back it out.


go/client/commands_common.go
```diff
@@ -68,9 +68,7 @@ func GetCommands(cl *libcmdline.CommandLine, g *libkb.GlobalContext) []cli.Comma
 		NewCmdVerify(cl, g),
 		NewCmdVersion(cl, g),
 		newCmdWallet(cl, g),
-		NewCmdPhoneNumber(cl, g),
-		NewCmdEmail(cl, g),
-		newCmdUploadAvatar(cl, g),
+		newCmdUploadAvatar(cl, g, true /* hidden */),
```
zapu: looks good

PR #18346, opened 2019-07-11
Make icon inline with text.

adamjspooner:


shared/wallets/wallet/settings/index.tsx
```diff
@@ -35,6 +35,12 @@ export type SettingsProps = {
   thisDeviceIsLockedOut: boolean
 }
 
+const HoverBox = Styles.isMobile
+  ? Kb.ClickableBox
+  : Styles.styled(Kb.ClickableBox)({
```
chrisnojima: just use `className='hover-underline'` and .hover-underline-container (see style.css)

PR #18351, opened 2019-07-11
Improved error flows in adding phone numbers

pzduniak:


shared/settings/account/add-modals.tsx
```diff
@@ -214,7 +218,11 @@ export const VerifyPhone = () => {
       header={{
         hideBorder: true,
         style: styles.blueBackground,
-        title: <Kb.Text type="BodySmall">{pendingVerification || 'wut'}</Kb.Text>,
+        title: (
+          <Kb.Text type="BodySmall" negative={true}>
+            {pendingVerification || 'wut'}
```
buoyad: Oh, forgot about this `wut`. Could we make it "Unknown number" or something?

PR #18356, opened 2019-07-11
add chat button to own profile

joshblum:
image

tested new/existing self conversation flows. Did we want different button styling here?

cc @keybase/hotpotatosquad @cecileboucheron @malgorithms

### DIFF
--- a/shared/profile/edit-profile/container.tsx
+++ b/shared/profile/edit-profile/container.tsx
@@ -32,7 +32,6 @@ const mergeProps = (stateProps, dispatchProps) => ({
   location: stateProps.location,
   onCancel: dispatchProps.onCancel,
   onSubmit: dispatchProps.onSubmit,
-  title: 'Edit Profile',
 })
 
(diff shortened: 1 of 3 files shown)


cecileboucheron:
lgtm!

shared/profile/edit-profile/container.tsx
```diff
@@ -32,7 +32,6 @@ const mergeProps = (stateProps, dispatchProps) => ({
   location: stateProps.location,
   onCancel: dispatchProps.onCancel,
   onSubmit: dispatchProps.onSubmit,
-  title: 'Edit Profile',
```
buoyad: This adds the title on mobile
image
However it doesn't need to be added back to the type, this HOC adds it https://github.com/keybase/client/blob/d77e8b873b5fd1c4fc8988416128987e7d9b5944/shared/profile/edit-profile/index.tsx#L117

Maybe it would be less confusing if we required the title. Anyway, for now can we add it back?

joshblum: oops yes

PR #18357, opened 2019-07-11
Downcase search result usernames

mlsteele:
Downcase user search results so nobody has to think about malgorithms != maIgorithms

image


zapu:
I guess i missed that one when porting this code from frontend. thanks

zapu:
This breaks hackernews invites and SBS. Mind if I change it backk?

zapu:
nevermind, I think we are fine with downcasing assertions here.

PR #18361, opened 2019-07-11
prefetcher: fix two races with the sync progress counts

strib:
I noticed a few problem races under this scenario:

1. Start KBFS.
2. It starts partially-syncing the recently-edited files for a TLF.
3. Manually set that TLF to be fully synced.
4. Sometimes this can cause the byte progress counts for the syncing TLF, and the overall sync state, to become incorrect, which can either stop the sync too early, or make the progress look hung forever.

The solutions are:

1. When a block has been prefetched already but now is being fetched with the sync action, we need to re-trigger its children fetches as well, to make sure their blocks are synced properly and their bytes counted correctly.

2. Similarly, if a request's original parent prefetch has already completed (because it wasn't a deep fetch), but then the parent is re-fetched with the deep sync action, we need to make sure that the request gets attached to the new parent.  We do this by removing the old parent entry when the sync mode has changed, allowing it to be attached to the new parent request farther below in the function.

I tried for a few hours to come up with some good regression tests for these, but I failed.  The interactions between the block retrieval queue, the prefetcher, and the disk block cache are very difficult to control.

Issue: HOTPOT-277

### DIFF
--- a/go/kbfs/libkbfs/prefetcher.go
+++ b/go/kbfs/libkbfs/prefetcher.go
@@ -74,6 +74,7 @@ const (
 type prefetch struct {
 	subtreeBlockCount int
 	subtreeTriggered  bool
+	subtreeRetrigger  bool
 	req               *prefetchRequest
 	// Each refnonce for this block ID can have a different set of
 	// parents.  Track the channel for the specific instance of the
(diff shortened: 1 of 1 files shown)


jzila:
Nice catch :)

PR #18363, opened 2019-07-12
teams: enable hidden rotation in dev

maxtaco:
- fix the deadly TeamMemberAddRace to handle total ordering violations
- requires https://github.com/keybase/keybase/pull/4044


maxtaco:
Can't get CI to work, I'm pretty sure it mainly works though from my own testing. It might introduce flakes though.

PR #18366, opened 2019-07-12
Add support for lookup errors from the server

zapu:


zapu:
cc @heronhaye we are merging this before it fixes an annoying bug

heronhaye:
sorry for delay, looks good

PR #18368, opened 2019-07-12
Show error from prove engine

mlsteele:


shared/profile/generic/result/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+import * as Tracker2Types from '../../../constants/types/tracker2'
+import {SiteIcon} from '../shared'
+
+type Props = {
+  serviceIcon: Tracker2Types.SiteIconSet
+  proofUsername: string
+  onClose: () => void
+  error: boolean
+  errorText: string
+}
```
mlsteele: This file is mostly not new code. From success/index.tsx .

PR #18370, opened 2019-07-12
Give the airdrop link a right-click/longpress way to be copied to clipboard

cjb:
@keybase/picnicsquad


cjb:
@keybase/react-hackers bump

PR #18371, opened 2019-07-12
don't double collapse bannerStyle

buoyad:
cc @pzduniak you don't need to spread in `collapseStyles`. merging


pzduniak:
shoot, sorry, hopefully that didn't break anything

PR #18373, opened 2019-07-12
Fix some send form styling

nathunsmitty:
Fixes an issue with the send form on desktop where the send button was placed below the rest of the send form content as a result of the height of the body of the send form being longer than it should be.

Story: http://localhost:6006/?path=/story/wallets-sendform--send

Note that the mobile send form is unchanged.

### DIFF
--- a/shared/wallets/send-form/index.stories.tsx
+++ b/shared/wallets/send-form/index.stories.tsx
@@ -23,12 +23,12 @@ const provider = banner =>
       amountErrMsg: '',
     }),
     Banner: props => ({}),
-    ConnectedPublicMemo: props => ({onChangePublicMemo: Sb.action('onChangePublicMemo')}),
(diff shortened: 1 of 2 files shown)


shared/wallets/send-form/index.stories.tsx
```diff
@@ -23,12 +23,12 @@ const provider = banner =>
       amountErrMsg: '',
     }),
     Banner: props => ({}),
-    ConnectedPublicMemo: props => ({onChangePublicMemo: Sb.action('onChangePublicMemo')}),
+    ConnectedPublicMemo: props => ({onChangePublicMemo: Sb.action('onChangePublicMemo'), maxLength: 28}),
```
nathunsmitty: I noticed that the memo and secret notes for the send form story were no longer getting character limits, since that data was changed to be provided by redux state in #18315.

nathunsmitty:

PR #18376, opened 2019-07-12
fix a bug in processing wallet events

xgess:
when a payment status updates, gregor sends an update message to the user with the paymentID and the accountID. the accountID is relevant (although not necessary) because it's used in, for example, transaction details to show a lumen payment denominated in the correct currency. among other things, this message triggers new notifications to `keybase chat api-listen --wallet`. this codepath expects the accountID to be findable for the user receiving the message. 

when alice sends bob a payment, and then it transitions from `pending` to `completed`, gregor sends the exact same message to both of them in the conversation alice,bob. this is a problem for bob, because the accountID on the message belongs to alice. if bob is listening for wallet events, there will be an account-not-found error which gets swallowed, and it looks like the payment is left on the floor. this is what i was seeing with my smsbot. 

so here's what i did:
1. ignore the accountID that gets sent in `stellar.payment_status` and `stellar.payment_notification` messages. instead, load the payment and then see which of the accountIDs on it (sender or recipient) belongs to the running user. use that accountID instead of erroring.
2. making change 1 resulted in a lot of duplicate messages being sent to `keybase chat api-listen --wallet`, so i added a little thing to dedupe on `{txID}:{status}`.

these two changes should make building on top of this cli command WAY nicer.

### DIFF
--- a/go/client/wallet_api_listen_display.go
+++ b/go/client/wallet_api_listen_display.go
@@ -1,6 +1,8 @@
 package client
 
 import (
+	"fmt"
+
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/stellar1"
 	"golang.org/x/net/context"
@@ -32,7 +34,8 @@ func newWalletNotification(source string) *walletNotification {
 
 type walletNotificationDisplay struct {
 	*baseNotificationDisplay
-	cli stellar1.LocalClient
+	cli     stellar1.LocalClient
+	deduper map[string]bool
 }
 
 func newWalletNotificationDisplay(g *libkb.GlobalContext) *walletNotificationDisplay {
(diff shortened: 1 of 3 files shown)


go/client/wallet_api_listen_display.go
```diff
@@ -43,21 +46,33 @@ func newWalletNotificationDisplay(g *libkb.GlobalContext) *walletNotificationDis
 	return &walletNotificationDisplay{
 		baseNotificationDisplay: newBaseNotificationDisplay(g),
 		cli:                     cli,
+		deduper:                 make(map[string]bool),
```
patrickxb: is there any concern that for the life of the command this will just increase in size?

xgess: i was thinking about that. i figured the data structure is likely so small, even if you're sending/receiving a ridiculous amount of payments on a very long-lived process, this would probably not be the first thing to break. but i definitely do not know that for sure. if you think it's worth using a different data structure that only keeps the more recent 100 or so transactions, i can look into that.

patrickxb: let's just go with this.

PR #18377, opened 2019-07-12
Better SBS profiles

zapu:
- [x] Fetch contact data for phone/email profiles. (another PR)
- [x] Fix clicking on "username" trying to load profile with that username - this will be wrong for SBS, because we no longer display assertions there but service usernames or formatted phone numbers / emails.

These will be split to separate ticket:
- [ ] Add site icon near username (see designs).
- [ ] Set pictureUrl as avatar, at least for twitter if the domain is allowed.


shared/profile/user/index.tsx
```diff
@@ -36,6 +36,10 @@ export type Props = {
   suggestionKeys: Array<string> | null
   userIsYou: boolean
   username: string
+  name: string // assertion value
+  service: string // assertion key (if SBS)
+  fullName: string | null // full name from profile
```
buoyad: `profile` -> `external profile`

shared/profile/user/index.tsx
```diff
@@ -72,6 +80,40 @@ const BioLayout = p => (
...
+    }
+    return (
+      <Kb.Text type="BodySmall" style={styles.proveIt}>
+        Tell {p.fullName || 'them'} to join Keybase and verify their {verifyWhat}.
+      </Kb.Text>
+    )
+  } else {
+    const url = 'https://keybase.io/install'
+    return (
+      <React.Fragment>
```
buoyad: You can use `<>` and `</>` for these

shared/tracker2/bio/index.tsx
```diff
@@ -150,7 +151,7 @@ const Bio = (p: Props) => (
         {p.bio}
       </Kb.Text>
     )}
-    {!!p.location && (
+    {!!p.location && !p.sbsDescription && (
```
buoyad: I don't think `p.sbsDescription` needs to affect whether the location appears

shared/util/phone-numbers/index.tsx
```diff
@@ -122,3 +122,8 @@ export const formatPhoneNumber = (rawNumber: string) => {
 }
 
 export const AsYouTypeFormatter = libphonenumber.AsYouTypeFormatter
+
+export const formatPhoneNumberInternational = (rawNumber: string) => {
+  const number = phoneUtil.parse(rawNumber)
```
buoyad: This can throw

zapu: oh good catch

shared/profile/user/index.tsx
```diff
@@ -72,6 +80,40 @@ const BioLayout = p => (
...
+        <Kb.Text type="BodySmall" style={styles.proveIt}>
+          Tell {p.fullName || p.name} to join Keybase and prove their {upperFirst(p.service)}.
+        </Kb.Text>
+        <Kb.Text type="BodySmall" style={styles.proveIt}>
+          Send them this link:{' '}
```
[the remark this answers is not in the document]
zapu: oh interesting, somehow I thought the design didn't want it but it's there. I'm changing it then, thank you

zapu:
@heronhaye can you do another go one? this one is just one new go function

PR #18380, opened 2019-07-13
 add a subscription manager for KBFS

songgao:
This is the Go side of the ticket. I'll have a different PR for using that in GUI, which I think would be another 2-3 points.

I haven't done the offline status plumbing yet; can add that later.

It'd be nice to have a `keybase fs watch` CLI that spits out either plaintext or jsos. I created HOTPOT-347 to track it.

cc @keybase/hotpotatosquad


go/kbfs/libkbfs/subscription_manager_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/stretchr/testify/require"
+	"golang.org/x/net/context"
+)
+
+func delayedFinish(ctl *gomock.Controller, timeout time.Duration) (
+	finish func(), done func(args ...interface{})) {
+	ch := make(chan struct{})
+	return func() {
+			select {
+			case <-time.After(timeout):
```
strib: Should this fail the test if it triggers?  Or is there a legitimate reason this could fire?

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	ref := pathSubscriptionRef{
+		folderBranch: node.GetFolderBranch(),
+		path:         normalizedInTlfPath(path),
+	}
+	if sm.pathSubscriptions[ref] == nil {
+		return
+	}
+	for _, notifier := range sm.pathSubscriptions[ref] {
+		notifier.notify()
```
strib: So this is intentionally triggering all topics for the path for now?  For example, a `touch` of a directory will trigger a `NodeChange` event for that directory's node, but the children of the directory will not have changed.  Similarly if someone (mistakenly?) signed up for children notifications on a file, they would get them whenever the file was written to.  Probably not a big deal but maybe worth documenting somewhere?  I'm just trying to figure out why path-based subscriptions have topics at all.

go/kbfs/libkbfs/subscription_manager_test.go
```diff
@@ -0,0 +1,91 @@
...
+	finish func(), lastMockDone func(args ...interface{})) {
+	ctl := gomock.NewController(t)
+	finish, lastMockDone = delayedFinish(ctl, time.Second)
+	config = MakeTestConfigOrBust(t, "jdoe")
+	notifier = NewMockSubscriptionNotifier(ctl)
+	subscriber = config.SubscriptionManager().Subscriber(notifier)
+	return config, subscriber, notifier, finish, lastMockDone
+}
+
+func TestSubscriptionManagerSubscribePath(t *testing.T) {
```
strib: It would be great to include file writes (before the `SyncAll` call) in this test, since they use a different notification path (`LocalChange`).

go/kbfs/libkbfs/tlf_journal.go
```diff
@@ -2180,6 +2181,7 @@ func (j *tlfJournal) putBlockData(
 		j.unsquashedBytes += uint64(bufLen)
 	}
 
+	j.config.SubscriptionManagerPublisher().JournalStatusChanged()
```
strib: Wouldn't you also need to trigger a notification when data is flushed as well?  Or is that a separate path?

go/kbfs/libkbfs/tlf_journal_test.go
```diff
@@ -248,6 +254,7 @@ func setupTLFJournalTest(
 		nil, nil, NewMDCacheStandard(10), ver,
 		NewReporterSimple(clocktest.NewTestClockNow(), 10), uid, verifyingKey, ekg, nil,
 		mdserver, defaultDiskLimitMaxDelay + time.Second,
+		NewMockSubscriptionManagerPublisher(gomock.NewController(t)),
```
strib: Why doesn't this break every test that does a `putBlockData()` call?  I would have expected an `EXPECT` somewhere.

go/kbfs/simplefs/simplefs.go
```diff
@@ -2204,6 +2208,8 @@ func (k *SimpleFS) LocalChange(
 	ctx context.Context, node libkbfs.Node, _ libkbfs.WriteRange) {
 	k.subscribeLock.RLock()
 	defer k.subscribeLock.RUnlock()
+	p, ok := node.GetPathPlaintextSansTlf()
+	fmt.Printf("SONGGAO-LocalChange: %q %v\n", p, ok)
```
strib: Nuke these debug logs and the otherwise unused line above.

go/kbfs/simplefs/simplefs.go
```diff
@@ -2212,6 +2218,10 @@ func (k *SimpleFS) LocalChange(
 // BatchChanges implements the libkbfs.Observer interface for SimpleFS.
 func (k *SimpleFS) BatchChanges(
 	ctx context.Context, changes []libkbfs.NodeChange, _ []libkbfs.NodeID) {
+	for _, change := range changes {
```
strib: Same for this loop.

protocol/avdl/keybase1/simple_fs.avdl
```diff
@@ -546,4 +546,19 @@ protocol SimpleFS {
...
+  enum SubscriptionTopic {
+    FAVORITES_0,
+    JOURNAL_STATUS_1
+  }
+  enum PathSubscriptionTopic {
+    CHILDREN_0,
+    STAT_1
+  }
+
+  // returns subscription ID. notificates are delivered through notify_fs
```
strib: notificates -> notifications

go/kbfs/libkbfs/subscription_manager.go
songgao: Admittedly I didn't put much thoughts on making this exact. Since the notification just triggers a load on the GUI side, I was just going along the path to rather have more notifications than missing one. Ideally it should do exactly as the topics describe, i.e. notifying children when children changes, and notifying stat when the file (or dir) itself changes. Perhaps we can do this in a separate ticket. I'll add a comment.

The reason why I put multiple topics here is not only for optimization (i.e. the "refining" mentioned above), but also so we don't have to keep track of topics in GUI. Then when receiving notification we can just mechanically trigger a load based on the topic inside the notification.

go/kbfs/libkbfs/kbfs_ops.go
```diff
@@ -79,7 +79,7 @@ func NewKBFSOpsStandard(appStateUpdater env.AppStateUpdater, config Config) *KBF
 		ops:                   make(map[data.FolderBranch]*folderBranchOps),
 		opsByFav:              make(map[favorites.Folder]*folderBranchOps),
 		reIdentifyControlChan: make(chan chan<- struct{}),
-		favs: NewFavorites(config),
+		favs:                  NewFavorites(config),
```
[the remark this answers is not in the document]
songgao: hmm I tried 1.12.4 and it didn't change anything here. So I guess the old one just had incorrect format?

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -0,0 +1,331 @@
...
+		sm.unsubscribeNonPath(ctx, sid)
+	}
+}
+
+func (sm *subscriptionManager) Subscriber(notifier SubscriptionNotifier) Subscriber {
+	return subscriber{sm: sm, notifier: notifier}
+}
+
+func makeSubscriptionID() SubscriptionID {
+	return SubscriptionID(strconv.FormatInt(time.Now().UnixNano(), 16))
```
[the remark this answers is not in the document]
songgao: Good point; though after thinking through how it'd work with GUI and I think it's easier to just have GUI pass in the subscription ID. So I'll just do that, and have the subscription manager check if the ID is a duplicate.

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	delete(sm.nonPathSubscriptionIDToTopic, subscriptionID)
+	if sm.nonPathSubscriptions[topic] == nil {
+		return
+	}
+	if notifier, ok := sm.nonPathSubscriptions[topic][subscriptionID]; ok {
+		notifier.shutdown()
+	}
+	delete(sm.nonPathSubscriptions[topic], subscriptionID)
+}
```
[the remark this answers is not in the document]
songgao: There are very few topics here, and they very likely need to be used soon, so I figured I'd just leave it there. The path subscriptions are different as they are referenced by path.

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -0,0 +1,331 @@
...
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+	return debouncedNotify{
+		notify: func() {
+			select {
+			case ch <- struct{}{}:
+			default:
```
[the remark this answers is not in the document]
songgao: If deduplicateInterval is set, we do want to drop notifications. Though I guess when it's not set this can potentially drop as well if the other side of the channel doesn't spawn goroutines fast enough. I don't think making it buffered is a best answer though since it'd buffer one more when debounced. I'll make it a blocking send when deduplicateInterval is not set, but let me know if you think that's a bad idea!

go/kbfs/libkbfs/subscription_manager_test.go
songgao: The mock tests would fail because it never saw the expect call. But I guess failing here is better. Will do!

go/kbfs/libkbfs/tlf_journal_test.go
songgao: I think it's because the `.Finish()` method on the controller is not called anywhere: https://godoc.org/github.com/golang/mock/gomock#Controller.Finish

songgao: Actually you're right; this breaks tests because it causes those notifications to block. Would need to `EXPECT()` those to happen `AnyTimes()`

go/kbfs/libkbfs/tlf_journal.go
songgao: I added one here: https://github.com/keybase/client/pull/18380/files#diff-f2721a37acfebd98c14216cf40e39098R1101 please let me know if that's not the correct place!

go/kbfs/libkbfs/subscription_manager.go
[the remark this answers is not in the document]
strib: It would be good to put a comment like that in here, so a future memory-leak-searcher knows the reasoning here.

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -0,0 +1,366 @@
...
+// at any time.
+// Examples:
+//   "/keybase/public/karlthefog@twitter/dir
+//   "/keybase/team/keybase/dir/../file"
+type userPath string
+
+// cleanInTlfPath is clean path rooted at a TLF, and it's what we get
+// from Node.GetPathPlaintextSansTlf().
+// Examples, considering TLF /keybase/private/user1,user2:
+//   "/foo/bar" (representing /keybsae/private/user1,user2/foo/bar)
```
strib: keybsae -> keybase, here and below.

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -0,0 +1,366 @@
...
+	return cleanInTlfPath(path.Clean(p.rawInTlfPath))
+}
+
+type debouncedNotify struct {
+	notify   func()
+	shutdown func()
+}
+
+func getChSender(ch chan<- struct{}, blocking bool) func() {
+	if blocking {
```
strib: I don't see how these "blocking" changes address my issue, which was with the following scenario when a non-Inf rate limit is set:

1. A node is changed, sending on `ch`.
2. The main `for` loop in the `debounce()` goroutine is triggered, `do()` is called, and then `limiter.Wait()` is entered, which causes it to wait for the interval.  (Maybe blocking in `Wait()` requires one more change; if so just pretend I wrote that out too.)
3. The GUI gets the notification, calls into KBFS and gets the new state of the node.
4. The node is changed again, and tries to send another notification, which is dropped because `ch` isn't trying to receive at the moment.
5. The node is never changed again.

Now the GUI will be stuck displaying the old state, because it never got the last notification.  That's why I think `ch` needs to have a buffer length of 1, so if a notification comes in during the wait period, it will eventually get sent.  You absolutely don't want to drop that notification, you just want to rate limit it so that not too many come at once, right?

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -0,0 +1,366 @@
...
+
+// FavoritesChanged implements the SubscriptionManagerPublisher interface.
+func (sm *subscriptionManager) FavoritesChanged() {
+	sm.lock.RLock()
+	defer sm.lock.RUnlock()
+	if sm.nonPathSubscriptions[keybase1.SubscriptionTopic_FAVORITES] == nil {
+		return
+	}
+	for _, notifier := range sm.nonPathSubscriptions[keybase1.SubscriptionTopic_FAVORITES] {
+		notifier.notify()
```
strib: If the new "blocking" mode isn't removed and is used by something, this could block while holding the lock, which seems bad.  I'd argue just to remove "blocking" mode and buffer the channel.

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -286,6 +291,10 @@ func (sm *subscriptionManager) nodeChangeLocked(node Node) {
 		return
 	}
 	for _, notifier := range sm.pathSubscriptions[ref] {
+		// We are notify()-ing while holding a lock, but it's fine since the
+		// other side of the channel consumes it pretty fast, eitehr by
```
strib: eitehr -> either

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -286,6 +291,10 @@ func (sm *subscriptionManager) nodeChangeLocked(node Node) {
...
+		// dropping overflown ones, or by doing the actual send in a separate
```
strib: "overflown" isn't really the right one -- maybe just "deduplicated" or "rate-limited"?

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -270,6 +270,11 @@ func (sm *subscriptionManager) unsubscribeNonPath(
 		notifier.shutdown()
 		delete(sm.nonPathSubscriptions[topic], subscriptionID)
 	}
+	// We are not deleting empty topic here because there are very few topics
```
strib: "empty topic" -> "empty topics"

PR #18383, opened 2019-07-14
ts strict part 1: chat

chrisnojima:
This is part 1 of 8 PRs which turn on strict mode (https://github.com/keybase/client/pull/18390/files#diff-9d7844586bcf395c9c0871c2dc4fa6a1R16) . This won't pass CI until all 8 are merged down 

Changes you'll see
- [x] Less `yield * Call`. TS loses all types if you yield. That should change in 3.6 but thats not out yet
- [x] Empty arrays need types `const toAddTo: Array<whatever> = []`  before you add things
- [x] `.filter(Boolean)` doesn't work. Most of those were turned into reduce
- [x] `.reduce<Array<whatever>>()` type reduce so it can tell what the extra param are
- [x] `null` and `undefined` are way more explicit. Instead of `foo?: string | null` which allows 3 types i changed it mostly to `foo?: string` so it's just two. If you have ternarys like `style={props.foo ? styles.red : null}` instead just pass `undefined`
- [x]  fixed a bunch of incorrect flowToTS closures: We had a bug where it'd make `onDone?: () => void | null` vs ` onDone?: (() => void) | null` (also the rule above mostly fixes this)
- [x] `RouteProps` only accepts 1 param (vs always an empty `{}` in the second param and param 1 is optional
- [x] Moved some `mergeProps` calls into `connect` so the types could be inferred
- [x] New rpc helper to get the return type `Saga.RPCPromiseValue<typeof RPCTypes.whateverrpc>`
- [x] React function components don't let you return `false` as a value so instead it does `props.foo ? <...> : null`


shared/actions/chat2/index.tsx
```diff
@@ -92,7 +91,11 @@ function* inboxRefresh(
     const result: RPCChatTypes.UnverifiedInboxUIItems = JSON.parse(inbox)
     const items: Array<RPCChatTypes.UnverifiedInboxUIItem> = result.items || []
     // We get a subset of meta information from the cache even in the untrusted payload
-    const metas = items.map(item => Constants.unverifiedInboxUIItemToConversationMeta(item)).filter(Boolean)
+    const metas = items.reduce<Array<Types.ConversationMeta>>((arr, item) => {
```
chrisnojima: TS doesn't like filter(Boolean)

shared/actions/chat2/index.tsx
```diff
@@ -221,7 +224,7 @@ function* unboxRows(
     // We allow empty conversations now since we create them and they're empty now
     const allowEmpty = action.type === Chat2Gen.selectConversation
     const meta = Constants.inboxUIItemToConversationMeta(inboxUIItem, allowEmpty)
-    const actions = []
+    const actions: Array<Saga.PutEffect> = []
```
chrisnojima: TS wants you to define your array types before you push into them

shared/actions/chat2/index.tsx
```diff
@@ -270,7 +273,7 @@ function* unboxRows(
       default:
         logger.info(`onFailed: displaying error for convID: ${conversationIDKey} error: ${error.message}`)
         return Saga.callUntyped(function*() {
-          const state = yield* Saga.selectState()
+          const state: TypedState = yield* Saga.selectState()
```
chrisnojima: anytime you yield (until TS 3.6 you get any :( )

shared/actions/chat2/index.tsx
```diff
@@ -285,6 +288,7 @@ function* unboxRows(
   yield Saga.put(Chat2Gen.createMetaRequestingTrusted({conversationIDKeys}))
   yield RPCChatTypes.localGetInboxNonblockLocalRpcSaga({
     incomingCallMap: {
+      // @ts-ignore TODO fix
```
chrisnojima: I didn't bother to fix the incoming map typing

shared/chat/conversation/info-panel/attachments.tsx
```diff
@@ -312,20 +319,23 @@ export class DocView {
...
+          renderItem: ({item}) => _renderEmptyItem(item),
+          renderSectionHeader: () => null,
+        },
+      ]
+    const sections = formMonths(docs).reduce<Array<Section>>((l, m) => {
       l.push(this._monthToSection(m))
       return l
     }, [])
-    return sections.concat(createLoadMoreSection(onLoadMore, onRetry, status))
+    return onLoadMore ? sections.concat(createLoadMoreSection(onLoadMore, onRetry, status)) : sections
```
chrisnojima: logic slightly changed but likely what we want

shared/chat/conversation/info-panel/attachments.tsx
```diff
@@ -382,20 +392,23 @@ export class LinkView {
...
+          renderItem: ({item}) => _renderEmptyItem(item),
+          renderSectionHeader: () => null,
+        },
+      ]
+    const sections = formMonths(links).reduce<Array<Section>>((l, m) => {
       l.push(this._monthToSection(m))
       return l
     }, [])
-    return sections.concat(createLoadMoreSection(onLoadMore, onRetry, status))
+    return onLoadMore ? sections.concat(createLoadMoreSection(onLoadMore, onRetry, status)) : sections
```
chrisnojima: slight logic change

shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
```diff
@@ -132,13 +128,10 @@ class _PlatformInput extends React.Component<PlatformInputPropsInternal, State>
   _insertEmoji = (emojiColons: string) => {
     if (this._input) {
       this._input.transformText(({text, selection}) => {
-        const newText = text.slice(0, selection.start) + emojiColons + text.slice(selection.end)
-        const pos = selection.start + emojiColons.length
+        const newText = text.slice(0, selection.start || 0) + emojiColons + text.slice(selection.end || 0)
```
chrisnojima: selection can actually be null. we could enforce this check way higher but i did this for now

chrisnojima:

shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
nathunsmitty: if selection can be null, wouldn't something like `selection.start` cause a property undefined error?

shared/actions/chat2/index.tsx
```diff
@@ -719,7 +728,7 @@ const onChatSetConvSettings = (_, action: EngineGen.Chat1NotifyChatChatSetConvSe
       conv.convSettings &&
       conv.convSettings.minWriterRoleInfo &&
       conv.convSettings.minWriterRoleInfo.cannotWrite) ||
-    null
+    false
```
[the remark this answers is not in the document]
chrisnojima: the question is get entirely typed correctly in TS

shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
chrisnojima: selection isn't null. selection.start can be

PR #18386, opened 2019-07-14
ts strict part 4: common-adapters

chrisnojima:
Please see https://github.com/keybase/client/pull/18383 for an overview


chrisnojima:

shared/common-adapters/floating-menu/menu-layout/index.d.ts
```diff
@@ -7,7 +7,7 @@ export type MenuItem = {
   decoration?: React.ReactNode // on the right side. unused if `view` is given,
   disabled?: boolean
   newTag?: boolean | null
-  onClick?: (evt?: React.SyntheticEvent) => void | null
+  onClick?: ((evt?: React.SyntheticEvent) => void) | null
```
jakob223: does this need to be `null | undefined`?

shared/common-adapters/form-input.native.tsx
```diff
@@ -14,7 +14,7 @@ export type Props = {
   // eslint-disable-next-line
   inputStyle?: any
   label?: string
-  onEnterKeyDown?: () => void | null
+  onEnterKeyDown?: (() => void)| null
```
jakob223: Does this need to be `undefined | null`?

shared/common-adapters/icon.stories.tsx
```diff
@@ -23,8 +23,8 @@ const commonProps = {
 
 const load = () => {
   const sizes = {}
-  Object.keys(iconMeta).map((type: IconType) => {
-    const meta = iconMeta[type]
+  Object.keys(iconMeta).map((type: any) => {
```
jakob223: what's going on here?

shared/common-adapters/input.desktop.tsx
```diff
@@ -273,7 +273,7 @@ class Input extends React.PureComponent<Props, State> {
     const defaultRowsToShow = Math.min(2, this.props.rowsMax || 2)
     const containerStyle = this._containerStyle(underlineColor)
 
-    const commonInputStyle = {
+    const commonInputStyle: any = {
```
jakob223: why are these `any` instead of the CSS type?

shared/common-adapters/input.shared.tsx
```diff
@@ -2,6 +2,12 @@
 import {TextInfo} from './input'
 
 const checkTextInfo = ({text, selection}: TextInfo) => {
+  if (selection.end === null) {
+    throw new Error('selection.end null')
```
jakob223: why even make these nullable?

shared/common-adapters/plain-input.desktop.tsx
```diff
@@ -175,8 +175,8 @@ class PlainInput extends React.PureComponent<InternalProps> {
 
   _getMultilineProps = () => {
     const rows = this.props.rowsMin || Math.min(2, this.props.rowsMax || 2)
-    const textStyle = getTextStyle(this.props.textType)
-    const heightStyles: _StylesDesktop = {
+    const textStyle: any = getTextStyle(this.props.textType)
+    const heightStyles: any = {
```
jakob223: what happened here?

shared/common-adapters/section-list.stories.tsx
```diff
@@ -27,16 +27,16 @@ const customRowRender = ({item}) => (
   </Text>
 )
 
-const small = []
+const small: Array<unknown> = []
```
jakob223: what's `unknown`? Why are you using it here?

shared/common-adapters/standard-screen.stories.tsx
```diff
@@ -31,14 +31,14 @@ const load = () => {
     ))
     .add('Back Button', () => (
       <Wrapper>
-        <StandardScreen {...props} onClose={null} onBack={action('onBack')} />
+        <StandardScreen {...props} onClose={undefined} onBack={action('onBack')} />
```
jakob223: is there a difference between including `onClose` as `undefined` and not including it?

shared/common-adapters/toast.d.ts
```diff
@@ -4,10 +4,10 @@ import {StylesCrossPlatform} from '../styles'
 
 export type Props = {
   children: React.ReactNode
-  className?: string | null
+  className?: string
   containerStyle?: StylesCrossPlatform
   visible: boolean
-  attachTo?: (() => any | null) | null
+  attachTo?: () => any | null
```
jakob223: Is this redundant? is `any | null` different from `any`?

shared/common-adapters/avatar.d.ts
```diff
@@ -16,7 +16,7 @@ export type Props = {
   isTeam?: boolean
   loadingColor?: string
   onClick?: (e?: React.SyntheticEvent) => void
-  onEditAvatarClick?: (e?: React.SyntheticEvent) => void | null
+  onEditAvatarClick?: ((e?: React.SyntheticEvent) => void )| null
```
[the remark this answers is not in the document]
chrisnojima: we'll take a further pass to do the whole undefined instead of null thing. i realized the change was better halfway through and only did it to reconcile a cluster of issues. we'll clean it up as we move forward

shared/common-adapters/banner.tsx
```diff
@@ -21,7 +21,10 @@ type BannerParagraphProps = {
 export const BannerParagraph = (props: BannerParagraphProps) => (
   <Text type="BodySmallSemibold" style={styles.text}>
     {(Array.isArray(props.content) ? props.content : [props.content])
-      .filter(Boolean)
+      .reduce<Array<_Segment | string>>((arr, s) => {
```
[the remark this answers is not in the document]
chrisnojima: it worked but then it didn't. see https://github.com/microsoft/TypeScript/pull/31515

shared/common-adapters/button.tsx
```diff
@@ -78,13 +78,13 @@ class Button extends React.Component<Props> {
 
     containerStyle = Styles.collapseStyles([containerStyle, this.props.style])
 
-    const onClick = (!unclickable && this.props.onClick) || null
+    const onClick = (!unclickable && this.props.onClick) || undefined
```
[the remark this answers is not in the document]
chrisnojima: my goal for these prs was (when possible) to try and change it as little as possible to get the typing to be correct and avoid little bugs as i was moving stuff around. my hope is the ugliness thats left around like this will be annoying and be fixed as we revisit these files

shared/common-adapters/clickable-box.d.ts
```diff
@@ -2,28 +2,28 @@ import * as React from 'react'
 import {StylesCrossPlatform} from '../styles'
 
 export type Props = {
-  className?: string | null
+  className?: string
   children?: any
   style?: StylesCrossPlatform
-  onClick?: (event: React.SyntheticEvent) => void | null
-  onDoubleClick?: (event: React.SyntheticEvent) => void | null
+  onClick?: (event: React.MouseEvent) => void
```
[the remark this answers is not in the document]
chrisnojima: a large source of these was an error in the codemod tool we used. this exercise caught a bunch of them

shared/common-adapters/clickable-box.desktop.tsx
```diff
@@ -75,8 +70,8 @@ class ClickableBox extends React.Component<
         onMouseEnter={needMouseEnterLeaveHandlers(this.props) ? this._onMouseEnter : undefined}
         onMouseLeave={needMouseEnterLeaveHandlers(this.props) ? this._onMouseLeave : undefined}
         onMouseUp={this._onMouseUp}
-        onDoubleClick={onDoubleClick}
-        onClick={onClick}
+        onDoubleClick={onDoubleClick || undefined}
```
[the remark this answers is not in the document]
chrisnojima: often its just too wide, aka the type is `() => void | null | undefined` and this needs `() =>void | undefined` so we need to get rid of the `null`

shared/common-adapters/confirm-modal/index.tsx
```diff
@@ -71,13 +71,13 @@ class ConfirmModal extends React.PureComponent<Props> {
...
                 style={styles.button}
-                waitingKey={this.props.waitingKey}
+                waitingKey={this.props.waitingKey || null}
               />
             </ButtonBar>
           ),
           hideBorder: true,
         }}
-        onClose={this.props.onCancel}
+        onClose={this.props.onCancel || undefined}
```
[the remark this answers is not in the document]
chrisnojima: yes

shared/common-adapters/icon.stories.tsx
chrisnojima: TS wants the cast to make sense so it wants conflicting types (aka `string` => `IconType`) to make sense, and a way to do the dangerous cast (which we're doing here, cause its a story and its not that scary) is `string => any => IconType` (below). It could be `unknown` or even better IconMeta could be typed better

shared/common-adapters/input.desktop.tsx
chrisnojima: some of these types were complicated by issues that weren't fixing at this phase of this project so i punted

shared/common-adapters/input.shared.tsx
chrisnojima: its cause they are nullable. We could enforce they're not and catch it upstream but at the time it wasn't clear to me how many upstream vs downstream usages of it there were so i decided to let the type flow down so i could see. i think we likely do want to enforce it higher up. There is some code that assumes they're valid and i didn't want to test all those cases so this was simpler for now

shared/common-adapters/plain-input.desktop.tsx
chrisnojima: I didn't want to correctly type getTextStyle right now

shared/common-adapters/section-list.stories.tsx
chrisnojima: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type
in terms of when i'm fixing things up for real vs punting usually involves a hierarchy like this (most important short term to least)

1. Reducers
2. Actions
3. Connected components
4. Stories

shared/common-adapters/standard-screen.stories.tsx
chrisnojima: no it could be removed

shared/common-adapters/toast.d.ts
chrisnojima: it is

shared/common-adapters/standard-screen.stories.tsx
chrisnojima: actually i take this back. normally it would be but the `{...props}` means it could be set, and this story intended to unset it so it has to be explicitly set. aka `props.conClose={Sb.Action('onClose')}` needs to be overwritten

chrisnojima:
@jakob223 i commented on everything but am leaning towards keeping it all so we can merge this all down. this is all just phase 1 of a larger project so i hope we'll clean this up further soon. sound good?

PR #18390, opened 2019-07-14
ts strict part 8: actions-tsconfig-constants

chrisnojima:
Please see https://github.com/keybase/client/pull/18383 for an overview


chrisnojima:

shared/tsconfig.base.json
```diff
@@ -9,9 +9,15 @@
     "jsx": "preserve",
     "module": "commonjs",
     "noFallthroughCasesInSwitch": true,
+    "noImplicitAny": false,
```
buoyad: Does this override `strict`?

chrisnojima: strict is a bag of flags. we aren't anywhere close to supporting any
```Enable all strict type checking options. 
Enabling --strict enables --noImplicitAny, --noImplicitThis, --alwaysStrict, --strictBindCallApply, --strictNullChecks, --strictFunctionTypes and --strictPropertyInitialization.```

chrisnojima: the goal is to turn them all on eventually but we can't do them all at the same time and get it working across all files at the same time

PR #18400, opened 2019-07-15
Imptofu search in team-building modal

zapu:


zapu:
@heronhaye can you take the go side of this PR? Frontend still waits to be rebased on John's code after he merges, but service can be review. Thanks!

go/contacts/contacts.go
```diff
@@ -5,27 +5,72 @@ package contacts
...
+	if err != nil {
+		return "", err
+	}
+	return ret.String(), nil
 }
 
-type contactAssertionPair struct {
-	contactName    string
-	componentValue string
+func findUsernamesAndTracking(mctx libkb.MetaContext, provider ContactsProvider, uidSet map[keybase1.UID]struct{},
```
heronhaye:
```suggestion
func findUsernamesAndFollowing(mctx libkb.MetaContext, provider ContactsProvider, uidSet map[keybase1.UID]struct{},
```

go/contacts/contacts.go
```diff
@@ -169,6 +220,10 @@ func ResolveContacts(mctx libkb.MetaContext, provider ContactsProvider, contacts
 	// contact name and hold the same assertion. Will also skip same assertions
 	// within one contact (duplicated components with same value and same or
 	// different name)
+	type contactAssertionPair struct {
```
heronhaye: defining type inside a function?

heronhaye: maybe pass c.Name into AssertionFromComponent and return this type from there

go/service/usersearch.go
```diff
@@ -188,33 +197,159 @@ func contactSearch(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []ke
...
+var nonDigits = regexp.MustCompile("[^\\d]")
+
+func imptofuQueryToAssertion(typ keybase1.ImpTofuSearchType, val string) string {
+	switch typ {
+	case keybase1.ImpTofuSearchType_PHONE:
+		return fmt.Sprintf("%s@phone", nonDigits.ReplaceAllString(val, ""))
+	case keybase1.ImpTofuSearchType_EMAIL:
+		return fmt.Sprintf("[%s]@email", strings.ToLower(val))
+	default:
+		panic("invalid keybase1.ImpTofuSearchType enum value")
```
heronhaye: can we just return an error instead since the caller returns an error?

go/service/usersearch.go
```diff
@@ -188,33 +197,159 @@ func contactSearch(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []ke
...
+		email := imptofuQuery.Email()
+		queryString = string(email)
+		emails = append(emails, email)
+	case keybase1.ImpTofuSearchType_PHONE:
+		phone := keybase1.RawPhoneNumber(imptofuQuery.Phone())
+		queryString = string(phone)
+		phones = append(phones, phone)
+	}
+
+	lookupRes, err := provider.LookupAll(mctx, emails, phones, keybase1.RegionCode(""))
```
heronhaye: i guess no luck on the regioncode? i thought ios had this info available for localization?

heronhaye: although I guess it's not strictly related

go/service/usersearch.go
```diff
@@ -188,33 +197,159 @@ func contactSearch(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []ke
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(lookupRes) > 0 {
```
heronhaye: when would this be greater than 1?

go/service/usersearch.go
```diff
@@ -188,33 +197,159 @@ func contactSearch(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []ke
...
+
+			assertionValue := queryString
+			if v.Coerced != "" {
+				assertionValue = v.Coerced
+			}
+			imptofu := &keybase1.ImpTofuSearchResult{
+				CoercedQuery: assertionValue,
+				Resolved:     true,
+				Uid:          v.UID,
+				Assertion:    imptofuQueryToAssertion(imptofuType, assertionValue),
```
heronhaye: What is this used for in the resolved case?

go/service/usersearch.go
```diff
@@ -188,33 +197,159 @@ func contactSearch(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []ke
...
 	mctx := libkb.NewMetaContext(ctx, h.G()).WithLogTag("USEARCH")
 	defer mctx.TraceTimed(fmt.Sprintf("UserSearch#UserSearch(s=%q, q=%q)", arg.Service, arg.Query),
 		func() error { return err })()
 
 	if arg.Query == "" {
-		return res, nil
+		return nil, nil
+	}
+
+	if !h.disableServiceSearch {
```
heronhaye: if this is only for testing can we put it in `libkb.TestOptions`?

go/service/usersearch.go
```diff
@@ -188,33 +197,159 @@ func contactSearch(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []ke
...
-			res2 = append(res2, res...)
-			res = res2
-
-			maxRes := arg.MaxResults
-			if maxRes > 0 && len(res) > maxRes {
-				res = res[:maxRes]
+	}
+
+	if arg.ImpTofuQuery != nil {
+		imptofuRes, err := imptofuSearch(mctx, h.contactsProvider, *arg.ImpTofuQuery)
```
heronhaye: is the query e164 checked for validity before burning rate limit in the gui?

heronhaye: (and email regex checked)

go/service/usersearch.go
heronhaye: (ah I see: later you use it for deduplication?)

go/service/usersearch.go
```diff
@@ -188,33 +197,159 @@ func contactSearch(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []ke
...
+
+			if !found {
+				// Prepend *imptofuRes
+				res = append([]keybase1.APIUserSearchResult{*imptofuRes}, res...)
 			}
 		}
 	}
 
+	maxRes := arg.MaxResults
+	if maxRes > 0 && len(res) > maxRes {
```
heronhaye: what is this value usually going to be? can we save work by not doing a bunch of service queries and so on before if we're going to throw them away later?

go/service/usersearch.go
zapu: even for resolved case we want the UI to start conversation with assertion, in case the ownership changes between search and createConversation

go/service/usersearch.go
zapu: yes, by frontend

zapu: although the email checker is much more powerful on serverside - even knows valid tlds - clientside check is much simpler

protocol/avdl/keybase1/usersearch.avdl
```diff
@@ -39,14 +39,39 @@ protocol userSearch {
         string username;
     }
 
+    record ImpTofuSearchResult {
+        string coercedQuery;
+        string assertion; // for use in team / chat building
+
+        boolean resolved;
+        // If resolved:
```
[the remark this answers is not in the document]
zapu: ProcessedContact type is very similar with `boolean resolved` and `UID uid; string username; etc.`

go/contacts/contacts.go
zapu: I don't think there's a need to. This type is only used here as key to a map for deduplicating `(contactName,assertionValue)` values.

go/protocol/keybase1/extras.go
```diff
@@ -2775,6 +2781,17 @@ func (c ContactComponent) ValueString() string {
 	}
 }
 
+func (c ContactComponent) AssertionType() string {
```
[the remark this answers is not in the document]
zapu: let's not change these types now, contacts are used in frontend already in master and other PRs

go/service/usersearch.go
zapu: it's handled on the frontend entirely - frontend knows the carrier region code and can transform inputs that are ultimately sent to this RPC. one day we will do a cleanup pass to remove this stuff from service.

go/service/usersearch.go
zapu: it tries to be resilient against buggy server, but you are right that we expecting this to be either 0 or 1

go/service/usersearch.go
zapu: you're right ill just redo this function - reorder the calls to allow us skip some of them if we are over the limit of search result anyways.

i don't know what i was thinking when i've done the reverse order calls with prepending to the list, maybe there was a reason, i need to think

go/service/usersearch.go
heronhaye: hm, does that work for emails like `n@ai`? may be too strict

zapu: is this a valid email?

clientside checker looks for `@` character and `.` that's after `@` but isn't the last char in the string.

serverside has some external validator from `validator` package

go/service/usersearch.go
zapu: This is postponed till later - also we are sorting this list differently now so we most likely always have to ask for full result set from API.

shared/team-building/user-result.native.tsx
```diff
@@ -147,8 +147,8 @@ const Username = (props: {
       </>
     ) : (
       <>
-        <Kb.Text type="BodySemibold"> {props.prettyName} </Kb.Text>
-        {props.displayLabel && props.displayLabel !== props.prettyName && (
+        <Kb.Text type="BodySemibold">{props.prettyName}</Kb.Text>
+        {!!props.displayLabel && props.displayLabel !== props.prettyName && (
```
zapu: fix unnecessary whitespace and also `Text` mobile crash

zapu:
Ok @buoyad this is ready for review. Thanks!

go/protocol/keybase1/extras.go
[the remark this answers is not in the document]
buoyad: don't block this on frontend's behalf if there's a benefit, it's only used in one place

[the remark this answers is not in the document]
zapu: there is another tickket to swap all these types to something better

shared/constants/team-building.tsx
```diff
@@ -79,7 +79,14 @@ const parseRawResultToUser = (
       id: result.contact.assertion,
       label: result.contact.displayLabel,
       prettyName: result.contact.displayName,
-      serviceMap: {...serviceMap, keybase: result.contact.username},
+      serviceMap: {keybase: result.contact.username},
```
buoyad: no `...serviceMap` anymore?

shared/team-building/user-result.desktop.tsx
```diff
@@ -95,6 +96,8 @@ const Avatar = ({
 }) => {
   if (keybaseUsername) {
     return <Kb.Avatar size={AvatarSize} username={keybaseUsername} />
+  } else if (resultForService === 'keybase' || resultForService === 'contact') {
+    return <Kb.Avatar size={AvatarSize} username={'invalid username for placeholder avatar'} />
```
buoyad: No need for brackets around string literal, use double quotes

zapu: I think this is copied from user-result.native, ill change there as well

shared/constants/team-building.tsx
zapu: It was always empty for contacts, but I guess it's ok to keep it so it magically works once we bring this in on core side

shared/actions/team-building.tsx
```diff
@@ -15,9 +16,11 @@ const apiSearch = (
   query: string,
   service: TeamBuildingTypes.ServiceIdWithContact,
   maxResults: number,
-  includeServicesSummary: boolean
-) => {
-  return RPCTypes.userSearchUserSearchRpcPromise({
+  includeServicesSummary: boolean,
+  impTofuQuery: RPCTypes.ImpTofuQuery | null
```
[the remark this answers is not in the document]
zapu: this is a function argument though, do I then pass `undefined` on the call site?

shared/util/phone-numbers/index.tsx
```diff
@@ -102,10 +102,10 @@ export const areaCodeIsCanadian = (input: string): boolean => {
   return !!canadianAreaCodes[input]
 }
 
-export const validateNumber = (rawNumber: string, region?: string) => {
+export const validateNumber = (rawNumber: string, region: string | null) => {
   try {
-    const number = phoneUtil.parse(rawNumber, region)
-    const valid = phoneUtil.isValidNumberForRegion(number, region)
+    const number = phoneUtil.parse(rawNumber, region || '')
+    const valid = phoneUtil.isValidNumber(number)
```
zapu: @buoyad  this seems to validate numbers correctly now. also see tests

go/service/usersearch.go
heronhaye: Ok, can you put a comment with ticket number?

PR #18401, opened 2019-07-15
Plumb badge for unverified phones+emails through gregor

buoyad:
Remove the startup call to `loadMySettings` and use added gregor items instead. cc @keybase/y2ksquad


mmaxim:
Thanks @buoyad!

PR #18403, opened 2019-07-15
fix teambot tests from server refactor

joshblum:
depends on https://github.com/keybase/keybase/pull/4054

cc @keybase/hotpotatosquad


joshblum:
tested locally, merging over CI to support the server changes

PR #18406, opened 2019-07-15
Make text white when selected.

adamjspooner:


shared/chat/selectable-big-team-channel.tsx
```diff
@@ -90,7 +90,10 @@ class SelectableBigTeamChannel extends PureComponent<Props, State> {
               </Kb.Text>
             </Kb.Box2>
             {!!this.props.numSearchHits && (
-              <Kb.Text type="BodySmall">
+              <Kb.Text
+                type="BodySmall"
+                style={Styles.collapseStyles([this.props.isSelected && styles.selectedText])}
```
MarcoPolo: Is there a reason you're doing collapsStyles here instead of just doing `this.props.isSelected ? styles.selectedText : undefined` ?

adamjspooner: I preferred the cleanliness of not having a dangling `undefined`.

buoyad:
I think you want to change `selectable-small-team` as well to make this consistent:
highlight_color

shared/chat/top-line.tsx
```diff
@@ -24,38 +25,42 @@ class FilteredTopLine extends PureComponent<Props> {
...
-    const {participants, showBold, usernameColor} = this.props
-    const boldOverride = showBold ? globalStyles.fontBold : null
     return (
-      <Box
-        style={{
-          ...globalStyles.flexBoxRow,
-          alignItems: 'center',
-          flex: 1,
-          justifyContent: 'flex-start',
-          position: 'relative',
```
buoyad: Unsure what these styles were for but I checked desktop / ios and the inbox rows look ok.

adamjspooner: Yeah, I checked on both and couldn't see a reason for those styles either.

PR #18410, opened 2019-07-15
Fix trustline threshold

mlsteele:
An account with 1.51 XLM and one existing trust line cannot add another trust line. Previously this offered that they could.


mlsteele:
@patrickxb I only need your review on that first commit with some go. The other commit has already been reviewed by cjb and then merged in here.

PR #18414, opened 2019-07-15
use Kb.Text for text filepreview on desktop; fix ts mime

songgao:


shared/fs/filepreview/text-view.desktop.tsx
```diff
@@ -0,0 +1,49 @@
...
+import * as Styles from '../../styles'
+import * as Kb from '../../common-adapters'
+import * as SafeElectron from '../../util/safe-electron.desktop'
+import {Props} from './text-view'
+
+const TextView = (props: Props) => {
+  const {url} = props
+  const [content, setContent] = React.useState('')
+  React.useEffect(() => {
+    const req = SafeElectron.getRemote().net.request({method: 'GET', url})
```
chrisnojima: lets chat about this approach

chrisnojima:
@keybase/react-hackers this fixes text in files tab

shared/fs/filepreview/text-view.desktop.tsx
```diff
@@ -0,0 +1,57 @@
...
+      try {
+        if (req.readyState === XMLHttpRequest.DONE && req.status === 200) {
+          setContent(req.responseText)
+        }
+      } catch (e) {}
+    }
+    try {
+      req.open('GET', url)
+      req.send()
+    } catch (e) {}
```
jakob223: Do we want to show the user a "Could not load file" or something in this case?

jakob223:
(otherwise LGTM)

chrisnojima:
We can add that later if it really happens

PR #18415, opened 2019-07-16
[Y2K-60] phone and email contacts in new team building

jzila:
Not ready yet:
* [x] Feature flag it off
* [x] Handle showing multiple reasons for a single contact
* [x] Fix bug where the same contact is showing up multiple times
* [ ] Format phone numbers correctly
* [x] Undo setting the feature flag in util/feature-flags.native

### DIFF
--- a/shared/actions/team-building.tsx
+++ b/shared/actions/team-building.tsx
@@ -140,10 +140,9 @@ const fetchUserRecs = (state: TypedState, {payload: {namespace}}: NSAction) =>
       const contactUsernames = new Set(contactRes.map(x => x.username).filter(Boolean))
       const contacts = contactRes.map(
         (x): TeamBuildingTypes.User => ({
-          // TODO: this doesn't work properly, since it doesn't highlight.
-          // Refactor this to work more sanely.
           id: x.assertion,
-          prettyName: x.displayLabel,
+          label: x.displayLabel,
+          prettyName: x.displayName,
           serviceMap: {keybase: x.username},
         })
       )
--- a/shared/actions/team-building.tsx
+++ b/shared/actions/team-building.tsx
@@ -140,6 +140,8 @@ const fetchUserRecs = (state: TypedState, {payload: {namespace}}: NSAction) =>
       const contactUsernames = new Set(contactRes.map(x => x.username).filter(Boolean))
       const contacts = contactRes.map(
         (x): TeamBuildingTypes.User => ({
+          // TODO: this doesn't work properly, since it doesn't highlight.
+          // Refactor this to work more sanely.
           id: x.assertion,
           prettyName: x.displayLabel,
           serviceMap: {keybase: x.username},
--- a/shared/local-debug.native.tsx
+++ b/shared/local-debug.native.tsx
@@ -59,7 +59,6 @@ if (__DEV__) {
   config.enableActionLogging = true
   config.enableStoreLogging = true
   config.immediateStateLogging = false
-  config.forceImmediateLogging = true
   // Move this outside the if statement to get notifications working
   // with a "Profile" build on a phone.
   config.isDevApplePushToken = true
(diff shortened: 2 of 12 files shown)


jzila:
I think this is ready for review.

jzila:
cc @keybase/y2ksquad

shared/actions/team-building.tsx
```diff
@@ -135,6 +135,8 @@ const fetchUserRecs = (state: TypedState, {payload: {namespace}}: NSAction) =>
       const contactUsernames = new Set(contactRes.map(x => x.username).filter(Boolean))
       const contacts = contactRes.map(
         (x): TeamBuildingTypes.User => ({
+          // TODO: this doesn't work properly, since it doesn't highlight.
+          // Refactor this to work more sanely.
           id: x.assertion,
           prettyName: x.displayLabel,
```
zapu: no `label: x.displayLabel` here?

shared/team-building/user-result.desktop.tsx
```diff
@@ -1,29 +1,15 @@
 import * as React from 'react'
```
zapu: I think the desktop version of this component should be ready to render phone numbers / email "results" because users will be able to search for phone number or emails even if they are not in their contacts (or they are not syncing contacts at all).

jzila: Yeah that's fine, but I still wanted to split it out because the layout is now very different.

jzila: @zapu does that search work yet?

zapu: nope it's in my pr and i have some rebasing to do... will get to it tomorrow

jzila: OK I'll make that a separate PR then.

zapu: I can also do this change in my PR

jzila:
@zapu @jakob223 PTAL at this. I'll try to implement feedback first thing if there's any more.

shared/team-building/user-result.native.tsx
```diff
@@ -0,0 +1,210 @@
...
+  keybaseUsername: string | null
+  prettyName: string
+  services: [Types.ServiceIdWithContact]
+}) =>
+  props.keybaseResult ? (
+    <Kb.Box2 direction="horizontal" fullWidth={true} style={styles.services}>
+      {!!props.prettyName && (
+        <Kb.Text type="BodySmall">{props.prettyName + (props.services.length ? ' •' : '')}</Kb.Text>
+      )}
+      {props.services.map(service => (
```
jakob223: instead of the `as` below, can you `map<Types.ServiceIdWithContact>`?

shared/local-debug.native.tsx
```diff
@@ -59,6 +59,7 @@ if (__DEV__) {
   config.enableActionLogging = true
   config.enableStoreLogging = true
   config.immediateStateLogging = false
+  config.forceImmediateLogging = true
```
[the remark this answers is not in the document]
chrisnojima: this should not be checked in

shared/team-building/user-result.native.tsx
chrisnojima: i'm confused why you need either. TS should know the type

shared/local-debug.native.tsx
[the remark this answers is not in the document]
zapu: i don't think that's intentional, it fixes some issue with John's setup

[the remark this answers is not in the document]
zapu: there is a TODO in this PR description to revert this file

shared/team-building/user-result.native.tsx
```diff
@@ -66,6 +66,8 @@ const Avatar = ({
 }) => {
   if (keybaseUsername) {
     return <Kb.Avatar size={AvatarSize} username={keybaseUsername} />
+  } else if (resultForService === 'keybase' || resultForService === 'contact') {
+    return <Kb.Avatar size={AvatarSize} username={'invalid username for placeholder avatar'} />
```
zapu: we need some story for this because I ran to this same issue in my other PR for SBS profiles

jzila: Sort of...this current approach does work correctly.

PR #18418, opened 2019-07-16
Match to more fields when searching for contacts

zapu:
This fixes bad behavior where if you had a contact e.g. phone number "+123456" that resolved to `michal`, then you couldn't search for that phone number in your contact list anymore because it wouldn't try to match to that field during searching.


zapu:
cc @jzila

PR #18420, opened 2019-07-16
Display proxy settings on the login page if the user is not connected to the internet

ddworken:
Internet detection is done via the ping endpoint (the same way that `keybase ping` works). If the user is on the join-or-login page and does not have internet, then a link is displayed that when clicked pops open a modal that displays the proxy settings. Since the login page does not support modals (and that work is out of scope for this PR) this modal simply displays on top of the home page. From there users can configure a proxy and save their settings. This allows users to join/login to Keybase even if they only have internet through a proxy.

See these screenshots:

normallogin
loginwithbox
modalunexpanded
proxymodal
doublemodal


shared/login/join-or-login/container.tsx
```diff
@@ -16,24 +18,39 @@ const mapStateToProps = state => {
...
-  mapStateToProps,
-  mapDispatchToProps,
-  mergeProps
+export default compose(
+  connect(
+    mapStateToProps,
+    mapDispatchToProps,
+    mergeProps
+  ),
+  lifecycle({
```
cjb: We're trying to deprecate using compose() and lifecycle(), because they're untyped.. looks like this goes through to a class component so can you just use that component's componentDidMount instead?

You could also consider using React Hooks useEffect() to keep that component as a function instead of class.

maxtaco:
It still feels to me that the MITM warning might be inappropriate for certain contexts. It still requires the checkbox plus rooting your OS to MITM SSL. We're just providing half of that...

ddworken:
I think part of the concern is that if someone is running a proxy, they are more likely to have a corporate TLS certificate added to their certificate store. So while we are only providing half of that, P(other half | our half) might still be decently high.

shared/login/join-or-login/container.tsx
ddworken: Done, moved it over to that component's lifecycle methods.

ddworken:
Well I couldn't access the settings page in order to allow taking screenshots, so here we go for mobile screenshots:

login
initial
expanded
modal

shared/login/join-or-login/index.tsx
```diff
@@ -24,43 +27,71 @@ const Feedback = ({onFeedback}) =>
...
+
+  constructor(props: Props) {
+    super(props)
+
+    this.intervalId = null
+  }
+
+  componentDidMount() {
+    this.props.checkIsOnline()
+    this.intervalId = setInterval(this.props.checkIsOnline, 2000)
```
cjb: Ah -- I think we should use the version of setInterval that's inside common-adapters/hoc-timers, using the native one can crash if the component's not mounted.

shared/settings/proxy/container.tsx
```diff
@@ -0,0 +1,83 @@
...
+  // @ts-ignore
+)(ProxySettingsPopup)
+
+const ProxySettings = compose(
+  connect(
+    mapStateToProps,
+    mapDispatchToProps,
+    (s, d, o) => ({...o, ...s, ...d})
+  ),
+  lifecycle({
```
cjb: I think this code was already like this so it doesn't have to happen in this PR, but it looks like if this too was just `export HeaderHoc(ProxySettings)` then you could use a class or useEffect() and be able to have it be typed.

shared/login/join-or-login/index.tsx
ddworken: Done.

shared/settings/proxy/container.tsx
ddworken: Done.

PR #18424, opened 2019-07-16
quick fix: more aggressively hide the cancel button

xgess:
canceling a relay payment doesn't correctly disable (or mark as waiting, or remove completely) the cancel button. 

i kind of think the nicest behavior here would be rearchitecting how we wait in cancellation, and passing additional data from the service so the front-end understands this additional state (requested cancellation but that process is still pending) which i guess we never really handled explicitly. but that's kind of a lot of work and just doing this (removing the button more aggressively) seems to solve the issue in a fairly intuitive way. the downside is that it's the same UI for pending. but if you just clicked on the cancel button, you're not going to be confused.

### DIFF
--- a/go/stellar/transform.go
+++ b/go/stellar/transform.go
@@ -313,13 +313,18 @@ func transformPaymentRelay(mctx libkb.MetaContext, acctID stellar1.AccountID, p 
 		}
 	}
 	if p.Claim != nil {
-		loc.ShowCancel = false // if there's a claim in either direction, it's too late to cancel
 		loc.StatusSimplified = p.Claim.ToPaymentStatus()
 		loc.ToAccountID = &p.Claim.ToStellar
 		loc.ToType = stellar1.ParticipantType_STELLAR
 		loc.ToUsername = ""
 		loc.ToAccountName = ""
-		if p.Claim.ToPaymentStatus() == stellar1.PaymentStatus_CANCELED {
+		claimStatus := p.Claim.ToPaymentStatus()
+		if claimStatus != stellar1.PaymentStatus_ERROR {
+			// if there's a claim and it's not currently erroring, then hide the
+			// `cancel` button
+			loc.ShowCancel = false
+		}
+		if claimStatus == stellar1.PaymentStatus_CANCELED {
 			// canceled payment. blank out toAssertion and stow in originalToAssertion
 			// set delta to what it would have been had the payment completed
 			loc.ToAssertion = ""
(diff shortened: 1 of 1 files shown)


go/stellar/transform.go
```diff
@@ -313,6 +313,7 @@ func transformPaymentRelay(mctx libkb.MetaContext, acctID stellar1.AccountID, p
 		}
 	}
 	if p.Claim != nil {
+		loc.ShowCancel = false // if there's a claim in either direction, it's too late to cancel
```
mlsteele: wdym? If the claim failed you should be able to try again.

xgess: oh interesting. i kind of had a feeling there was a case i was missing.

xgess: @mlsteele doesn't actually matter. the claim falls off the original transaction if it fails, so either way in the UI all you see is "sending...". which isn't great since the thing failed. the good news is that, because the failed claim falls away, it retries periodically and would fix itself if we ever pushed a fix to a bug.

mlsteele: I tried it and it seems to make a difference. Idk what's "falls off the original".
1. Sent 3 XLM to torproject@twitter
2. [Disabled stellard](https://github.com/keybase/keybase/compare/miles/stellard-fake-error?expand=1#diff-a62658675e3d7357b831df38554db421R2124) 
3. Clicked Cancel. The cancel failed (expected due to disabled stellard) with a global error (unhandled rpc failure strikes again).
4. On master the Cancel button is still there.
image
5. On this branch, the Cancel button is gone.
image


The money is [still in the intermediate relay account](https://stellar.expert/explorer/testnet/account/GC5I237ZVYO5BRWXPD67ZX7RLLZT63I4SL3ISUTSIOY7GR6HLHHI67AU) so the Cancel button should be available for the clicking.

patrickxb: can we just use a spinner on cancel?  and have the spinner go away and the button reactivated when there is an error?

mlsteele: it appears to spin on master

xgess: spoke with miles offline. pushed a fix to his break. i had broken it a different way that showed different behavior. 🙃 
@patrickxb there already is a spinner actually. it's spinning for the life of the `cancel` rpc, which isn't the right timeframe. we can change that, but whether or not it's showing is driven by this code. since there's no point in having the cancel button show up if it's not cancellable anyway, this feels acceptable to me. i can ping you in chat if you're still curious about it.

PR #18426, opened 2019-07-16
Use expo-image-picker instead of RN image picker

MarcoPolo:
@keybase/react-hackers Fixes the issue of uses external storage as temp cache on android.


shared/chat/conversation/input-area/filepicker-popup/index.native.tsx
```diff
@@ -40,12 +40,6 @@ class FilePickerPopup extends React.Component<Props> {
             },
             title: 'Take Photo',
           },
-          {
```
MarcoPolo: expo doesn't let you take videos on the fly

shared/util/expo-image-picker.tsx
```diff
@@ -0,0 +1,44 @@
+import {isIOS} from '../constants/platform'
+import * as ImagePicker from 'expo-image-picker'
+import * as Permissions from 'expo-permissions'
+
+export const parseUri = (result: {uri: string}): string => {
+  return isIOS ? result.uri.replace('file://', '') : result.uri.replace('file:', '')
+}
+
+const retyAfterAskingPerm = (retryFn: null | (() => Promise<ImagePicker.ImagePickerResult>)) => (
```
MarcoPolo: This keeps the old functionality where we try to ask for permissions if we didn't have it already.

shared/package.json
```diff
@@ -126,6 +126,8 @@
     "emoji-mart": "2.9.2",
     "exif-js": "2.3.0",
     "expo-contacts": "5.0.1",
+    "expo-image-picker": "git://github.com/MarcoPolo/expo-image-picker",
```
MarcoPolo: Needed for now because https://github.com/expo/expo/pull/4717 isn't in the published version yet.

Happy to move to a keybase repo (I didn't have permissions for that), but I don't think we'll need this for long.

shared/package.json
```diff
@@ -126,6 +126,8 @@
     "emoji-mart": "2.9.2",
     "exif-js": "2.3.0",
     "expo-contacts": "5.0.1",
+    "expo-image-picker": "keybase/expo#keybase-image-picker",
```
chrisnojima: should use `git://github.com/` to be consistent

chrisnojima:
- [ ] taking a video crashes ios

PR #18427, opened 2019-07-16
dockerized service

pzduniak:
```
cd $GOPATH/src/github.com/keybase/client
docker build -f ./packaging/linux/docker/Dockerfile -t "keybase/client" .
```

Potential improvements: 

- [x] dockerignore?
- [ ] docs!


pzduniak:
@heronhaye Tagging you because it's sort of a Linux release, using all these build scripts.
@jzila How can we get this out? Do we own the `keybase` org on Docker Hub?

@xgess i guess you made something like this before?

packaging/linux/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian
+FROM debian:10
```
jzila: `debian:stretch` is what we use everywhere else

packaging/linux/docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN gpg --batch --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
+    && gpg --batch --verify /tini.asc /tini \
+    && chmod +x /tini
+
+RUN useradd --create-home --shell /bin/bash keybase
+USER keybase
+
+COPY --from=builder /binaries/amd64/usr/bin/keybase /usr/bin/keybase
+
+ENTRYPOINT ["/tini", "--"]
```
jzila: neat, TIL!

packaging/linux/docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini.asc /tini.asc
+RUN gpg --batch --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
+    && gpg --batch --verify /tini.asc /tini \
+    && chmod +x /tini
+
+RUN useradd --create-home --shell /bin/bash keybase
+USER keybase
```
heronhaye: @jzila I remember there used to be some weirdness wrt mountpoints about using the user "keybase", is it still a problem?

heronhaye:
I don't know anything about docker but the keybase-build-specific parts look fine

packaging/linux/docker/Dockerfile
jzila: Not sure what you're referring to, we use `keybase` in the test dockers.

packaging/linux/Dockerfile
pzduniak: sure thing, I thought we'd go with latest because that's what it defaulted to

packaging/linux/docker/Dockerfile
strib: The redirector does look for paths that contain "keybase/" when trying to find the right mountpoint for redirecting.  But it should only matter if there will be another mountpoint in the container that also contains "keybase".

pzduniak: Shouldn't be an issue - I'll elaborate further in another reply here.

packaging/linux/docker/Dockerfile
pzduniak: You can just pass `--init` and it does literally the same thing, but i'd rather enforce it. It won't matter for this particular Docker image but if we add other processes it might actually clean up some stuff.

pzduniak:
You unfortunately can't (or _shouldn't_) run KBFS in Docker. You can't get FUSE to work without making a container privileged / effectively giving it root privileges, which will render it useless in any "managed" environment and would limit the application of Docker to just being a packaging solution.

A while ago I filed this ticket to summarize the requirements:
https://keybase.atlassian.net/browse/CORE-11028

What we definitely should look into is spawning KBFS in such a way that it only does simplefs. I'm not familiar enough with it to make it work that way so it'd be great if someone could make that work it way.

pzduniak:
also wrt the keybase namespace:

image

xgess:
looks good to me. and i like the serverless bot example! very useful.

packaging/linux/build_binaries.sh
```diff
@@ -3,7 +3,8 @@
 set -e -u -o pipefail
 
 here="$(dirname "$BASH_SOURCE")"
-this_repo="$(git -C "$here" rev-parse --show-toplevel)"
+this_repo="$(git -C "$here" rev-parse --show-toplevel ||
+  echo -n $GOPATH/src/github.com/keybase/client)"
```
heronhaye: ?

packaging/linux/docker/client-slim/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+FROM debian:stretch-slim
+LABEL maintainer="Keybase <admin@keybase.io>"
+
+RUN apt-get update \
+    && apt-get install -y gnupg2 procps \
+    && rm -rf /var/lib/apt/lists/*
+
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini.asc /tini.asc
```
heronhaye: we don't need to similarly sign ours?

packaging/version.sh
```diff
@@ -16,7 +16,8 @@ build=""
 
 if [ ! "$mode" = "production" ] ; then
   current_date="$(date -u +%Y%m%d%H%M%S)" # UTC
-  commit_short="$(git -C "$here" log -1 --pretty=format:%h)"
+  commit_short="$(git -C "$here" log -1 --pretty=format:%h || \
+    echo -n ${SOURCE_COMMIT:0:10})"
```
heronhaye: ?

packaging/linux/docker/client-slim/Dockerfile
pzduniak: That'll require your expertise, I have no idea how to approach that.

packaging/version.sh
pzduniak: dockerignore skips `.git` so all the git commands won't work, SOURCE_COMMIT is pushed into the container using docker build args

ignore is done to reduce the build times ~3x

packaging/linux/build_binaries.sh
pzduniak: answered below

w1966106:
ok

PR #18430, opened 2019-07-16
Fix contact resolve issues

zapu:
Deduplicate components within contacts when adding unresolved entries, do a better job of skipping already resolved contacts.

### DIFF
--- a/go/contacts/contacts.go
+++ b/go/contacts/contacts.go
@@ -15,7 +15,7 @@ import (
 func formatSBSAssertion(c keybase1.ContactComponent) string {
 	switch {
 	case c.Email != nil:
-		return fmt.Sprintf("[%s]@email", *c.Email)
+		return fmt.Sprintf("[%s]@email", strings.ToLower(string(*c.Email)))
 	case c.PhoneNumber != nil:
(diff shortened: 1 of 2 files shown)


go/contacts/contacts.go
```diff
@@ -104,13 +112,14 @@ func ResolveContacts(mctx libkb.MetaContext, provider ContactsProvider, contacts
 							continue
 						}
 
+						assertion := formatSBSAssertion(component)
```
jzila: why?

go/contacts/contacts_test.go
```diff
@@ -453,3 +453,48 @@ func TestErrorsInResolution(t *testing.T) {
...
+			Components: []keybase1.ContactComponent{
+				makeEmailComponent("email", "bob+test@keyba.se"),
+			},
+		},
+		keybase1.Contact{
+			Name: "Robert B.",
+			Components: []keybase1.ContactComponent{
+				makeEmailComponent("E-Mail", "bob+test@keyba.se"),
+			},
+		},
```
jzila: I think you need another `keybase1.Contact` whose name is also `Alice` and has the phone number `+1111222`. I'm pretty sure you don't handle that case right now, and that's another level of deduping that would be handy.

PR #18433, opened 2019-07-17
SEP7: Path payments

cjb:
@keybase/picnicsquad CC @keybase/react-hackers @patrickxb 

Here's how SEP7 path payments work:

* we were already calling the `FindPaymentPath` RPC in the `calculateBuildingAdvanced` saga, reading from `state.wallets.buildingAdvanced` and writing the result to `state.wallets.builtPaymentAdvanced`.
* for SEP7, we call the same RPC in the same saga but read from `sep7ConfirmInfo` and write to `sep7ConfirmPath`.
* perform that path search from a `useEffect` in sep7-confirm when the dialog opens, show a spinner and disable the send button until it's ready
* when the Send button's clicked, we call `approvePathURI` instead of `approvePayURI`.  You can tell that it's a Path payment because `assetCode` is set.
* to actually draw the path, re-use the `<AssetPathIntermediate />` component from the advanced send form.


shared/wallets/send-form/body/advanced.tsx
```diff
@@ -63,7 +63,7 @@ const SendBodyAdvanced = (props: SendBodyAdvancedProps) => (
   <Kb.Box2 fullWidth={true} direction="vertical" style={sharedStyles.container}>
     <Kb.ScrollView style={sharedStyles.scrollView}>
       <AssetInputRecipientAdvanced />
-      <AssetPathIntermediate />
+      <AssetPathIntermediate forSEP7={false} />
```
nathunsmitty: would it make sense to make `forSEP7` default to false since that's likely the more common case?

PR #18441, opened 2019-07-17
Format phone numbers in settings

buoyad:
- Transform phone numbers in settings to store both e164 and display formats.
- Use display formats in Settings > Account and in modals

Second part will format them in chat, it's a bit more complicated there. cc @keybase/y2ksquad

### DIFF
--- a/shared/util/phone-numbers/index.tsx
+++ b/shared/util/phone-numbers/index.tsx
@@ -1,8 +1,8 @@
-import {isMobile} from '../../constants/platform'
+import { isMobile } from '../../constants/platform'
 import libphonenumber from 'google-libphonenumber'
 import countries from './country-data/countries.json'
 import supportedCodes from './sms-support/data.json'
-import {emojiIndexByChar} from '../../common-adapters/markdown/emoji-gen'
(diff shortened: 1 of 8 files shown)


shared/util/phone-numbers/index.tsx
```diff
@@ -121,4 +121,15 @@ export const formatPhoneNumber = (rawNumber: string) => {
   return `+${number.getCountryCode()} ${phoneUtil.format(number, PNF.NATIONAL)}`
 }
 
+// Return phone number in international format, e.g. +1 (800) 555 0123
+// or e.164 if parsing fails
+export const e164ToDisplay = (e164: string) => {
+  try {
+    const number = phoneUtil.parse(e164)
+    return phoneUtil.format(number, PNF.INTERNATIONAL)
```
jakob223: International format gives `+1 345-543-1232` - it would be nice if we did a special case to get the parens for US/cCA numbers with `+1 (234) 555-6789`

buoyad: Agreed. I'm working on another PR based off this one, I'll look into it in that one.

PR #18450, opened 2019-07-17
fast tsc for precommit

chrisnojima:
Use fast for pre-commit
CI will continue to run the slower one
@keybase/react-hackers cc: @mmaxim


mmaxim:
Think it might make sense to just completely disable it, but let’s try this

chrisnojima:
We cannot doable it

PR #18453, opened 2019-07-17
rpccache: key on UID too

jakob223:
When switching between users that had synced folders, users would not be able to decrypt entries others left in the RPC cache. This led to a broken KBFS after user switching in some cases.

CC @joshblum @strib @keybase/y2ksquad 

Issue: Y2K-191

### DIFF
--- a/go/offline/rpc_cache.go
+++ b/go/offline/rpc_cache.go
@@ -44,8 +44,9 @@ func NewRPCCache(g *libkb.GlobalContext) *RPCCache {
 	}
 }
 
-func hash(rpcName string, arg interface{}) ([]byte, error) {
+func hash(rpcName string, uid keybase1.UID, arg interface{}) ([]byte, error) {
 	h := sha256.New()
+	h.Write(uid.ToBytes())
 	h.Write([]byte(rpcName))
 	h.Write([]byte{0})
 	raw, err := msgpack.Encode(arg)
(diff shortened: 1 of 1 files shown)


go/offline/rpc_cache.go
```diff
@@ -80,7 +82,7 @@ func (c *RPCCache) get(mctx libkb.MetaContext, version Version, rpcName string,
 	c.Lock()
 	defer c.Unlock()
 
-	dbk, err := dbKey(rpcName, arg)
+	dbk, err := dbKey(rpcName, mctx.G().GetMyUID(), arg)
```
maxtaco: What about the case of no logged in uid?  Also, this might not give the right answer if the user is in the process of getting logged in ---- which I guess should never happen offline...

jakob223: that should be a separate namespace entirely and work fine with an empty UID -- shouldn't it?

maxtaco: I think so, just 2x-checking

jakob223: cool. Do you think this is good to merge then?

go/offline/rpc_cache.go
```diff
@@ -44,8 +44,9 @@ func NewRPCCache(g *libkb.GlobalContext) *RPCCache {
 	}
 }
 
-func hash(rpcName string, arg interface{}) ([]byte, error) {
+func hash(rpcName string, uid keybase1.UID, arg interface{}) ([]byte, error) {
 	h := sha256.New()
+	h.Write(uid.ToBytes())
```
maxtaco: there's a chance you can collide unless you separate the UID and the rpcName. Maybe it makes sense to:

```go
type hashStruct struct {
    Uid uid
    RpcName string
    Arg interface{}
}
```
and msgpack / hash that, instead.

patrickxb: `h.Write(':')` (after line 49, before line 50) would also work, right?

heronhaye: how would it collide if uid is fixed length?

jakob223: UID can be empty if nobody is logged in.

PR #18455, opened 2019-07-18
bot command UI support

mmaxim:
Support in the UI for showing bot commands after typing `!`.


mmaxim:
To try this just type ! in any Keybase team channel that I am in.

shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
```diff
@@ -186,6 +188,12 @@ class _PlatformInput extends React.Component<PlatformInputPropsInternal, State>
         onKeyDown={this._globalKeyDownPressHandler}
         onKeyPress={this._globalKeyDownPressHandler}
       >
+        {this.props.suggestBotCommandsUpdateStatus !== undefined &&
```
chrisnojima: does this not type w/o the `!== undefined` ? it should work just doing a truthy check ( `this.props.suggestBotCommandsUpdateStatus && `

shared/chat/conversation/input-area/normal/shared.tsx
```diff
@@ -19,7 +20,41 @@ export const ExplodingMeta = ({explodingModeSeconds}: {explodingModeSeconds: num
   )
 }
 
+type BotCommandUpdateStatusProps = {
+  status: RPCChatTypes.UIBotCommandsUpdateStatus
+}
+
+export class BotCommandUpdateStatus extends React.Component<BotCommandUpdateStatusProps> {
```
chrisnojima: should just be a function component, we're trying to reduce classes we don't need

shared/constants/chat2/meta.tsx
```diff
@@ -431,6 +433,15 @@ export const getCommands = (state: TypedState, id: Types.ConversationIDKey) => {
   }
 }
 
+export const getBotCommands = (state: TypedState, id: Types.ConversationIDKey) => {
+  const {botCommands} = getMeta(state, id)
+  if (botCommands.typ === RPCChatTypes.ConversationCommandGroupsTyp.custom && botCommands.custom) {
+    return botCommands.custom.commands || []
```
chrisnojima: should hoist this empty array up else your containers will thrash

shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
mmaxim: What is the best way to handle this if the value can be `0`?

mmaxim: That is why I did it like this, the truthy check fails if this thing has the value 0.

shared/constants/chat2/meta.tsx
mmaxim: Is there like a global empty array? We will never hit this case anyway, this is just to appease TS, the service will always be providing a value here.

shared/chat/conversation/input-area/normal/shared.tsx
mmaxim: Is there anything wrong with classes? I just liked having these things in the same namespace, and having props floating around on this.

shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
chrisnojima: oh i see, didn't realize thats a valid value, k

shared/chat/conversation/input-area/normal/shared.tsx
chrisnojima: there's overhead we can remove. it can all be teh same just not a class. i don't understand what you mean by 'same namespace' and having 'props floating around'

shared/constants/chat2/meta.tsx
chrisnojima: no. this is why traditionally we use immutable for these types of datastructures but you're using an array and raw rpc stuff

go/chat/commands/bot.go
```diff
@@ -0,0 +1,89 @@
...
+		}(globals.BackgroundChatCtx(ctx, b.G()))
+	}
+
+	cmds, err := b.G().BotCommandManager.ListCommands(ctx, convID)
+	if err != nil {
+		b.Debug(ctx, "Preview: failed to list commands: %s", err)
+		return
+	}
+	for _, cmd := range cmds {
+		if strings.HasPrefix(text, fmt.Sprintf("!%s", cmd.Name)) && cmd.ExtendedDescription != nil {
```
joshblum: could `text = text[1:]` so we can skip adding `!` to each cmd

joshblum: discussed argparse/exact match in chat so we don't select the wrong command

go/chat/bots/commands.go
```diff
@@ -343,11 +385,22 @@ func (b *CachingBotCommandManager) commandUpdate(ctx context.Context, job comman
...
-			UpdateType: chat1.StaleUpdateType_CONVUPDATE,
-		}}, true)
+	if err := storage.NewInbox(b.G()).IncrementLocalConvVersion(ctx, b.uid, job.convID); err != nil {
+		b.Debug(ctx, "commandUpdate: unable to IncrementLocalConvVersion, err", err)
+	}
+	conv, err := utils.GetVerifiedConv(ctx, b.G(), b.uid, job.convID, types.InboxSourceDataSourceAll)
+	if err != nil {
+		return err
+	}
+	username, err := b.getMyUsername(ctx)
```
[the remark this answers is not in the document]
mmaxim: Don't think so, UPAK should be able to handle this case hopefully.

PR #18459, opened 2019-07-18
Use correct type for --disable-cert-pinning command line flag in serv…

heronhaye:
…er spawn


heronhaye:
(CI passed in hotfix branch)

NieDzejkob:
Is it feasible to add a regression test for this?

heronhaye:
@NieDzejkob For this specific problem we could type the arrays better (`[]cli.BoolFlag` and `[]cli.StringFlag` instead of just strings, and import them from the `libcmdline` package). 

But in general to check for correct behavior across systemd, non-systemd, *BSD, Termux, etc., it is difficult. We could do some basic docker tests though.

PR #18460, opened 2019-07-18
rethink some of the wallet RPCs to be more direct

xgess:
some of the wallet RPCs make changes and then wait for other things to happen (e.g. gregor notifications or subsequent RPCs) to push the updates back into the front end. here, i'm trying to pass back the updates directly with the RPC and then handle the new data synchronously. 

i strongly suggest reviewing this PR commit-by-commit.

- change currency
- change account name
- not really any lift from inflation setting
- setting to and from mobile-only

### DIFF
--- a/go/stellar/stellarsvc/frontend.go
+++ b/go/stellar/stellarsvc/frontend.go
@@ -540,7 +540,7 @@ func (s *Server) ChangeWalletAccountNameLocal(ctx context.Context, arg stellar1.
 	return stellar.WalletAccount(mctx, s.remoter, arg.AccountID)
 }
 
-func (s *Server) SetWalletAccountAsDefaultLocal(ctx context.Context, arg stellar1.SetWalletAccountAsDefaultLocalArg) (err error) {
+func (s *Server) SetWalletAccountAsDefaultLocal(ctx context.Context, arg stellar1.SetWalletAccountAsDefaultLocalArg) (accts []stellar1.WalletAccountLocal, err error) {
 	mctx, fin, err := s.Preamble(ctx, preambleArg{
 		RPCName:       "SetWalletAccountAsDefaultLocal",
 		Err:           &err,
@@ -548,15 +548,19 @@ func (s *Server) SetWalletAccountAsDefaultLocal(ctx context.Context, arg stellar
 	})
 	defer fin()
 	if err != nil {
-		return err
+		return accts, err
 	}
 
 	if arg.AccountID.IsNil() {
 		mctx.Debug("SetWalletAccountAsDefaultLocal called with an empty account id")
-		return ErrAccountIDMissing
+		return accts, ErrAccountIDMissing
 	}
 
-	return stellar.SetAccountAsPrimary(mctx, s.walletState, arg.AccountID)
+	err = stellar.SetAccountAsPrimary(mctx, s.walletState, arg.AccountID)
+	if err != nil {
+		return accts, err
+	}
+	return stellar.AllWalletAccounts(mctx, s.remoter)
 }
 
 func (s *Server) DeleteWalletAccountLocal(ctx context.Context, arg stellar1.DeleteWalletAccountLocalArg) (err error) {
--- a/go/stellar/stellarsvc/frontend.go
+++ b/go/stellar/stellarsvc/frontend.go
@@ -517,7 +517,7 @@ func (s *Server) ValidateAccountNameLocal(ctx context.Context, arg stellar1.Vali
 	return nil
 }
 
-func (s *Server) ChangeWalletAccountNameLocal(ctx context.Context, arg stellar1.ChangeWalletAccountNameLocalArg) (err error) {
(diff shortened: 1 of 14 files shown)


shared/actions/wallets.tsx
```diff
@@ -641,7 +646,17 @@ const setAccountAsDefault = (_: TypedState, action: WalletsGen.SetAccountAsDefau
   RPCStellarTypes.localSetWalletAccountAsDefaultLocalRpcPromise(
     {accountID: action.payload.accountID},
     Constants.setAccountAsDefaultWaitingKey
-  ).then(() => WalletsGen.createDidSetAccountAsDefault({accountID: action.payload.accountID}))
+  ).then(accountsAfterUpdate =>
+    WalletsGen.createDidSetAccountAsDefault({
+      // @ts-ignore codemod-issue
```
cjb: What's the TS issue?

xgess: nice. no issue at all. removed.

cjb: Maybe the one you copied can be removed too?

xgess: nope. i tried.

protocol/avdl/stellar1/local.avdl
```diff
@@ -471,9 +471,13 @@ protocol local {
...
-  void setAccountMobileOnlyLocal(int sessionID, AccountID accountID);
-  void setAccountAllDevicesLocal(int sessionID, AccountID accountID); // opposite of mobile-only
-  boolean isAccountMobileOnlyLocal(int sessionID, AccountID accountID);
+  record AccountMobileOnly {
+    AccountID accountID;
+    boolean enabled;
+  }
+  AccountMobileOnly setAccountMobileOnlyLocal(int sessionID, AccountID accountID);
+  AccountMobileOnly setAccountAllDevicesLocal(int sessionID, AccountID accountID); // opposite of mobile-only
+  AccountMobileOnly isAccountMobileOnlyLocal(int sessionID, AccountID accountID);
```
mlsteele: I think booleans would be preferable unless you've got reason to expect more return values. The frontend knows the accountID it passed in. Reflecting it by hand seems more likely to have a bug.

go/stellar/stellarsvc/frontend.go
```diff
@@ -879,74 +879,95 @@ func (s *Server) MarkAsReadLocal(ctx context.Context, arg stellar1.MarkAsReadLoc
...
+		return res, err
 	}
 
 	if err = s.walletState.UpdateAccountEntries(mctx, "set account mobile only"); err != nil {
-		return err
+		return res, err
 	}
 
-	return nil
+	isMobileOnly, err := s.remoter.IsAccountMobileOnly(mctx.Ctx(), arg.AccountID)
```
mlsteele: Returning `true` would be the simplest and instant. Is this about races?

protocol/avdl/stellar1/local.avdl
xgess: this is because i thought it would be weird to get a `false` back from a method where you're changing something and it actually worked. i guess we also have the error object and no one _should_ make this mistake.

mlsteele: Good point, better return void

xgess: > better return void for the setters

lol. but that's what this whole chunk of work was about!

go/stellar/stellarsvc/frontend.go
xgess: just a stronger guarantee that the thing actually worked. i have no reason to suspect it could fail without erroring, but it costs only an http round trip to verify, which we were also doing before this PR. at least this time it doesn't include an additional roundtrip between js and go.

protocol/avdl/stellar1/local.avdl
mlsteele: yeah... sorry. Do you agree?

xgess: hahaha. well, given the speedup and reduced communication, i guess i think it's worth it as a followon optimization like how we're doing it. on the whole though, i'm definitely a fan of separating command and query methods. i could go either way. 🤷‍♂

xgess: discussed. i will set these back to `void`

mlsteele: https://www.youtube.com/watch?v=KSSEzWXqGKY

shared/actions/wallets.tsx
xgess: @cjb this is the one i copied from. it looks to me like an incorrect type being ignored, and this seems right to me. https://github.com/keybase/client/pull/18460/commits/a27edfd6071e8f06612d6ebc9a356fbb4473ef07

xgess:
@mlsteele k. i rewrote that mobile-only commit to do much less.

PR #18461, opened 2019-07-18
superseded phone number banner

pzduniak:


shared/settings/account/container.tsx
```diff
@@ -8,16 +8,38 @@ import AccountSettings from '.'
...
-const mapStateToProps = (state: TypedState) => ({
-  _emails: state.settings.email.emails,
-  _phones: state.settings.phoneNumbers.phones,
-  addedEmail: state.settings.email.addedEmail,
-  bootstrapDone: state.settings.email.emails !== null && state.settings.phoneNumbers.phones !== null,
-  hasPassword: !state.settings.password.randomPW,
-  waiting: anyWaiting(state, Constants.loadSettingsWaitingKey),
-})
+const mapStateToProps = (state: TypedState) => {
+  const phoneNumberRecord =
```
jakob223: can this variable name include the fact that it's superseded?

shared/settings/account/index.tsx
```diff
@@ -11,6 +11,8 @@ export type Props = {
   addedEmail: string | null
   contactKeys: I.List<string>
   hasPassword: boolean
+  supersededPhoneNumber?: ContactRowProps
```
jakob223: This doesn't really need to be a full `ContactRowProps`, does it? Can you use only the one or two fiels of that you actually need?

shared/settings/account/container.tsx
```diff
@@ -8,16 +8,38 @@ import AccountSettings from '.'
...
+
+  return {
+    _emails: state.settings.email.emails,
+    _phones: state.settings.phoneNumbers.phones,
+    _supersededPhoneNumberKey: phoneNumberRecord && phoneNumberRecord.e164,
+    addedEmail: state.settings.email.addedEmail,
+    bootstrapDone: state.settings.email.emails !== null && state.settings.phoneNumbers.phones !== null,
+    hasPassword: !state.settings.password.randomPW,
+    supersededPhoneNumber: phoneNumberRecord
+      ? {
```
jakob223: can you cut out those of these fields you don't need?

pzduniak: i just realized that after all my changes i just needed 1 field

PR #18463, opened 2019-07-18
TS: noImplicitThis flag

chrisnojima:
Turn on the no implicit this flag


chrisnojima:

PR #18465, opened 2019-07-18
Fix lint

MarcoPolo:
cc @mmaxim


mmaxim:
Probably going to conflict with @chrisnojima 's fix

PR #18467, opened 2019-07-18
Fix tab shortcuts.

adamjspooner:


shared/router-v2/tab-bar/index.desktop.tsx
```diff
@@ -194,7 +194,7 @@ const styles = Styles.styleSheetCreate({
 })
 
 const keysMap = Tabs.desktopTabOrder.reduce((map, tab, index) => {
-  map[`${Platforms.shortcutSymbol}+${index + 1}`] = tab
+  map[`${Platforms.isDarwin ? 'command' : 'ctrl'}+${index + 1}`] = tab
```
heronhaye: may want to make this a platforms constant too if it can be used in other places

PR #18472, opened 2019-07-18
remove more unused params

chrisnojima:
@keybase/react-hackers removing more unused vars

### DIFF
--- a/shared/chat/avatars.tsx
+++ b/shared/chat/avatars.tsx
@@ -57,7 +57,7 @@ const MutedIcon = (p: {
 
 class Avatars extends React.Component<AvatarProps> {
   shouldComponentUpdate(nextProps: AvatarProps) {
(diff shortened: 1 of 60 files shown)


shared/chat/avatars.tsx
```diff
@@ -57,7 +57,7 @@ const MutedIcon = (p: {
 
 class Avatars extends React.Component<AvatarProps> {
   shouldComponentUpdate(nextProps: AvatarProps) {
-    return !shallowEqual(this.props, nextProps, (obj, oth, key) => {
+    return !shallowEqual(this.props, nextProps, (_, __, key) => {
```
nathunsmitty: Is that second argument supposed to be a double underscore? Guessing it somehow distinguishes itself from the previous underscore?

chrisnojima: these are params, not object keys

PR #18481, opened 2019-07-19
final no unused params

chrisnojima:
@keybase/react-hackers last one for this flag


shared/fs/banner/system-file-manager-integration-banner/container.tsx
```diff
@@ -23,15 +23,15 @@ const mapDispatchToProps = (dispatch, ownProps: OwnProps) => ({
...
   'SystemFileManagerIntegrationBanner'
 )(Banner)
 
 export default ConnectedBanner
 
 export const asRows = isMobile
-  ? (path: Types.Path, showBanner: boolean) => []
-  : (path: Types.Path, showBanner: boolean): Array<RowTypes.HeaderRowItem> =>
+  ? () => []
+  : (_: Types.Path, showBanner: boolean): Array<RowTypes.HeaderRowItem> =>
```
jakob223: this path parameter could just go away

chrisnojima: yeah some stuff i didn't plumb through all the way and cleanup all call sites, just trivial ones. hopefully the _ will stick out and people will clean it up when they're in these files

PR #18484, opened 2019-07-19
Surya/y2 k 278/guiconfig

heronhaye:
@chrisnojima there are the values (and types) that'll be migrated over, please sanity check.

```
{
    "stellar": {
        "lastSentXLM": false
    },
    "ui": {
        "importContacts": {
            "user1": true,
            "user2": false,
            "user3": true
        },
        "routeState": "foo",
        "routeState2": "bar",
        "shownMonsterPushPrompt": true
    }
}
```


go/libkb/config.go
```diff
@@ -72,14 +72,6 @@ func (f *JSONConfigFile) GetTopLevelBool(s string) (res, isSet bool) {
 	return
 }
 
-func (f *JSONConfigFile) SetWrapperAtPath(p string, w *jsonw.Wrapper) error {
```
heronhaye: moved to go/libkb/json.go

patrickxb:
What about the existing `app-state.json` file?  Can some (or all) of this live there?  that lastSentXLM thing seems like application state to me, for example.

heronhaye:
@patrickxb I believe that @chrisnojima wants to move away from app-state.json long term.

chrisnojima:
yeah lets have core handle this keystore. app-state is a hack and should go away and be replaced w/ this. Those keys look fine, but we should just kill the old `routeState`
remove this line in this PR: https://github.com/keybase/client/blob/ade88a6044a40072ef8b7d28d657f63579662b67/shared/actions/platform-specific/index.native.tsx#L217
(you could also remove that generator and the chain call

heronhaye:
@chrisnojima there is a "clearRouteState" call in loadStartupDetails, should I remove it too?

```diff
diff --git a/shared/actions/platform-specific/index.native.tsx b/shared/actions/platform-specific/index.native.tsx
index fac123618d..f97d2486bd 100644
--- a/shared/actions/platform-specific/index.native.tsx
+++ b/shared/actions/platform-specific/index.native.tsx
@@ -212,14 +212,6 @@ const getStartupDetailsFromShare = (): Promise<
       })
     : Promise.resolve(null)
 
-function* clearRouteState() {
-  yield Saga.spawn(() =>
-    RPCTypes.configGuiSetValueRpcPromise({path: 'ui.routeState', value: {isNull: false, s: ''}}).catch(
-      () => {}
-    )
-  )
-}
-
 let _lastPersist = ''
 function* persistRoute(state, action: ConfigGen.PersistRoutePayload) {
   const path = action.payload.path
@@ -331,9 +323,6 @@ function* loadStartupDetails() {
         startupConversation = item.param && item.param.selectedConversationIDKey
         startupTab = item.routeName
       }
-
-      // immediately clear route state in case this is a bad route
-      yield clearRouteState()
     } catch (_) {
       startupConversation = null
       startupTab = null
@@ -574,7 +563,6 @@ const getE164 = (phoneNumber: string, countryCode?: string) => {
 function* platformConfigSaga(): Saga.SagaGenerator<any, any> {
   yield* Saga.chainGenerator<ConfigGen.PersistRoutePayload>(ConfigGen.persistRoute, persistRoute)
   yield* Saga.chainAction<ConfigGen.MobileAppStatePayload>(ConfigGen.mobileAppState, updateChangedFocus)
-  yield* Saga.chainGenerator<ConfigGen.LoggedOutPayload>(ConfigGen.loggedOut, clearRouteState)
   yield* Saga.chainAction<ConfigGen.OpenAppSettingsPayload>(ConfigGen.openAppSettings, openAppSettings)
   yield* Saga.chainAction<ConfigGen.CopyToClipboardPayload>(ConfigGen.copyToClipboard, copyToClipboard)
   yield* Saga.chainGenerator<ConfigGen.DaemonHandshakePayload>(
```

go/libkb/json.go
```diff
@@ -51,27 +51,37 @@ func (f *JSONFile) GetWrapper() *jsonw.Wrapper {
...
-			return nil
+			return false, nil
 		}
 
 		MobilePermissionDeniedCheck(f.G(), err, fmt.Sprintf("%s: %s", f.which, f.filename))
 
 		if os.IsPermission(err) {
 			f.G().Log.Warning("Permission denied opening %s file %s", f.which, f.filename)
-			return nil
+			return true, nil
```
jakob223: It's weird to me that this returns a `nil` error on this case.

chrisnojima:
yeah

go/libkb/json.go
heronhaye: i'm not sure either, @patrickxb?

heronhaye:
merging over TestMemberAddRace flake...

PR #18487, opened 2019-07-19
Use PaymentDetailCLILocal instead of GetPaymentDetailsLocal for chat listen --wallet notifications

nathunsmitty:
We currently use a call to `GetPaymentDetailsLocal` to retrieve and format incoming payments when running `keybase chat api-listen --wallet`. This incoming data is a different shape than the data returned by `keybase wallet api`'s `details` method, which uses `PaymentDetailCLILocal`. In the interest of consistency in our API and bots, it seemed to make sense to use `PaymentDetailCLILocal` for all API output.

### DIFF
--- a/go/client/wallet_api_listen_display.go
+++ b/go/client/wallet_api_listen_display.go
@@ -50,17 +50,14 @@ func newWalletNotificationDisplay(g *libkb.GlobalContext) *walletNotificationDis
 	}
 }
 
-func deduperKey(details stellar1.PaymentDetailsLocal) string {
-	return fmt.Sprintf("%s:%s", details.Summary.TxID, details.Summary.StatusSimplified)
+func deduperKey(details stellar1.PaymentCLILocal) string {
+	return fmt.Sprintf("%s:%s", details.TxID, details.Status)
 }
 
 func (d *walletNotificationDisplay) displayPaymentDetails(ctx context.Context, source string,
 	accountID stellar1.AccountID, paymentID stellar1.PaymentID) error {
 	notif := newWalletNotification(source)
-	details, err := d.cli.GetPaymentDetailsLocal(ctx, stellar1.GetPaymentDetailsLocalArg{
-		AccountID: accountID,
-		Id:        paymentID,
-	})
+	details, err := d.cli.PaymentDetailCLILocal(ctx, paymentID.String())
 	if err != nil {
 		errStr := err.Error()
 		notif.Error = &errStr
@@ -72,6 +69,7 @@ func (d *walletNotificationDisplay) displayPaymentDetails(ctx context.Context, s
 	}
 
 	d.deduper[deduperKey(details)] = true
+
 	notif.Notification = details
 	d.printJSON(notif)
 	return nil


go/client/wallet_api_listen_display.go
```diff
@@ -50,17 +50,14 @@ func newWalletNotificationDisplay(g *libkb.GlobalContext) *walletNotificationDis
...
 }
 
 func (d *walletNotificationDisplay) displayPaymentDetails(ctx context.Context, source string,
 	accountID stellar1.AccountID, paymentID stellar1.PaymentID) error {
 	notif := newWalletNotification(source)
-	details, err := d.cli.GetPaymentDetailsLocal(ctx, stellar1.GetPaymentDetailsLocalArg{
-		AccountID: accountID,
-		Id:        paymentID,
-	})
+	details, err := d.cli.PaymentDetailCLILocal(ctx, paymentID.String())
```
nathunsmitty: Not sure if this paymentID string conversion is the best way to do this – `PaymentDetailCLILocal` takes a transactionID of type `string`, while `GetPaymentDetailsLocal` takes an `AccountID` and `PaymentID`.

nathunsmitty:
Not sure why jenkins is failing, all tests seem to pass locally and even jenkins seems to show all checks

xgess:
how much does this change the API of, for example, a normal completed payment?

nathunsmitty:
@xgess Signature of `PaymentLocal` (current output):
```
  // Summary of a payment.
  record PaymentLocal {
    PaymentID id;
    TransactionID txID;             // For relay payments, the funding tx.
    TimeMs time;
    PaymentStatus statusSimplified;
    string statusDescription;       // "pending", "completed", "error", "claimable", "canceled"
    string statusDetail;
    boolean showCancel;             // true if cancel is an option for this payment
    string amountDescription;       // "1,323.1234567 XLM" - always amount of XLM, even if converted from OutsideCurrency
    BalanceDelta delta;             // NONE/INCREASE/DECREASE (e.g. INCREASE for "+ 1,323.1234567 XLM" amount above)
    string worth;                   // "$123.23 CLP", senders intended amount if this was made as a lumens-worth-currency payment
    string worthAtSendTime;         // "$123.23 CLP", amount in current user's display currency. Set if frontend should show an
                                    // (APPROXIMATELY $X.XX) message i.e. if this is a pure-XLM payment.

    string issuerDescription;       // For non-XLM assets: either "Unknown issuer" or domain name
                                    // ("example.com") if domain is verified by stellard.
    union{ null, AccountID } issuerAccountID;

    // What type to display the sender and recipient as.
    // Either neither or both are OWNACCOUNT.
    ParticipantType fromType; // cannot be SBS
    ParticipantType toType;

    string assetCode; // The code of the destination asset. e.g., "USD", "XLM", etc.

    AccountID fromAccountID;   // Always filled
    string    fromAccountName; // Optional. User's name for the account or Account ID if not named. Filled if fromType = OWNACCOUNT
    string    fromUsername;    // Optional. Filled if KEYBASE.

    union { null, AccountID } toAccountID; // Optional.
    string    toAccountName;               // Optional. Filled if toType = OWNACCOUNT
    string    toUsername;                  // Optional. Filled if KEYBASE.
    string    toAssertion;                 // Optional. Filled if SBS and statusSimplified != 'canceled'
    string    originalToAssertion;         // Optional. Username or SBS. Filled if statusSimplified == 'canceled'

    string note;                    // secret note
    string noteErr;

    // for path payments:
    string sourceAmountMax;
    string sourceAmountActual;
    Asset sourceAsset;
    string sourceConvRate; // The amount of the destination asset worth 1 unit of the source asset at the time of the transaction


    // for non-payment txs, multiple ops:
    boolean isAdvanced;
    string summaryAdvanced;         // if isAdvanced is true, just show this string
    array<string> operations;       // string summary of each operation in the transaction

    boolean unread;                 // true if user hasn't seen the payment yet (sourceType STELLAR only)

    string batchID;                 // if payment sent via batch, will contain the batch id
    boolean fromAirdrop;            // true if payment pertains to a keybase airdrop
    boolean isInflation;            // true if payment from known inflation source
    union { null, string } inflationSource; // name of inflation source

    union { null, PaymentTrustlineLocal } trustline; // this tx is only a trust line change.
  }

  // Details of a payment.
  record PaymentDetailsLocal {
    PaymentLocal summary;
    PaymentDetailsOnlyLocal details;
  }

  record PaymentDetailsOnlyLocal {
    string publicNote;              // public memo for the payment
    string publicNoteType;          // public memo type: "text", "id", "hash", "return"

    @lint("ignore")
    string externalTxURL;           // link to external web page with tx details
    string feeChargedDescription;   // "0.0000100 XLM" the actual fee that was charged for this tx (formatted)
    array<Asset> pathIntermediate;  // for payment paths, the additional hops it took
  }

  record PaymentTrustlineLocal {
    Asset asset;
    boolean remove;
  }
```

Signature of `PaymentCLILocal` (what this change does):
```
  record PaymentCLILocal {
    TransactionID txID; // For relay payments, the funding tx.
    TimeMs time;
    string status;
    string statusDetail;
    string amount;
    Asset asset;
    union { null, string } displayAmount;
    union { null, string } displayCurrency;

    string sourceAmountMax;
    string sourceAmountActual;
    Asset sourceAsset;

    // for non-payment txs, multiple ops:
    boolean isAdvanced;
    string summaryAdvanced;         // if isAdvanced is true, just show this string
    array<string> operations;       // string summary of each operation in the transaction

    AccountID fromStellar;
    union { null, AccountID } toStellar;
    union { null, string } fromUsername;
    union { null, string } toUsername;
    union { null, string } toAssertion;
    string note;
    string noteErr;
    boolean unread;

    // details
    string publicNote;              // public memo for the payment
    string publicNoteType;          // public memo type: "text", "id", "hash", "return"
    string feeChargedDescription;   // "0.0000100 XLM" the actual fee that was charged for this tx (formatted)
  }
```

So most of the same stuff is there, but it a different enough shape to break anything that currently uses it.

nathunsmitty:
That being said, I think it's important we only have one payment shape in our api

xgess:
the new format removes some potentially useful info (e.g. the two `worth` fields). i'm guessing since most of this is reusing the interfaces that currently exist between js and go, versioning just hasn't been necessary. but we should definitely think about a version before breaking this api. i do agree that it's unfortunate we're returning nearly identical data in two completely different formats, but we should try to reduce breaking this thing if it's going to become a real interface. this is probably more work that you want to do here. :/

go/client/wallet_api_listen_display.go
```diff
@@ -47,32 +48,53 @@ func newWalletNotificationDisplay(g *libkb.GlobalContext) *walletNotificationDis
...
-	if err != nil {
-		errStr := err.Error()
-		notif.Error = &errStr
-		d.printJSON(notif)
-		return nil
-	}
-	if _, dupeMsg := d.deduper[deduperKey(details)]; dupeMsg {
-		return nil
+
+	if d.useV1 {
```
nathunsmitty: Not sure if there's a cleaner way to do this, as there's a lot of duplication between the two paths.

go/client/wallet_api_listen_display.go
```diff
@@ -47,32 +48,53 @@ func newWalletNotificationDisplay(g *libkb.GlobalContext) *walletNotificationDis
 		baseNotificationDisplay: newBaseNotificationDisplay(g),
 		cli:                     cli,
 		deduper:                 make(map[string]bool),
+		useV1:                   useV1,
 	}
 }
 
-func deduperKey(details stellar1.PaymentDetailsLocal) string {
```
xgess: you could also leave this as deduperKey that takes an empty interface, switch on its type, and blow up if it gets something other than the two expected types.

nathunsmitty: I kind of prefer getting compiler error about the type vs a runtime error. Ideally, we could define a type union or overload `deduperKey`...

PR #18492, opened 2019-07-19
bkg teambot keygen

joshblum:
adds background key generation for ephemeral and non-ephemeral teambot keys for bot members when added to the team or when the PTK is rotated. generation is best effort, if the generation fails or the bot wins a race to access the key they will notify that they need a new key and fail with a transient error or fall back to an older key if available.

depends on https://github.com/keybase/keybase/pull/4076

cc @keybase/hotpotatosquad

### DIFF
--- a/go/teams/member_set.go
+++ b/go/teams/member_set.go
@@ -8,6 +8,14 @@ import (
 	"golang.org/x/net/context"
 )
 
+type storeMemberKind int
+
+const (
+	storeMemberKindNone = iota
+	storeMemberKindRecipient
+	storeMemberKindBotRecipient
+)
+
 type member struct {
 	version    keybase1.UserVersion
 	perUserKey keybase1.PerUserKey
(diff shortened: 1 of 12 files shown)


go/teams/member_set.go
```diff
@@ -265,7 +297,18 @@ func (m *memberSet) AddRemainingRecipients(ctx context.Context, g *libkb.GlobalC
...
+		if _, ok := m.botRecipients[uv]; ok {
+			continue
+		}
+
+		var storeMemberKind storeMemberKind
+		if _, ok := existingBots[uv]; ok {
+			storeMemberKind = storeMemberKindBotRecipient
+		} else {
+			storeMemberKind = storeMemberKindRecipient
+		}
```
mlsteele: god forbid some user be both bot and not would you rather they get non-bot here?

joshblum:
merging over unrelated flakes

PR #18498, opened 2019-07-21
Tweaks to imp tofu search

zapu:
- Do a better job of merging two lists - keybase users and contacts - also deduplicate by username if a search yields both a keybase user and a contact resolving to same user. The precedence of which is displayed depends on the "search score", so how well the query matched an entry.
- Change frontend search rows to be aware of resolved contacts and display coloured username (blue or green depending on a follow)
- Plumb matched value from contact entry to the frontend label for resolved user. So if someone's searching for phone number and we matched a contact that resolves to a user, make sure that phone number appears as label in the results.

### DIFF
--- a/go/service/usersearch.go
+++ b/go/service/usersearch.go
@@ -62,6 +62,7 @@ func doSearchRequest(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []
 	for i, row := range response.List {
 		if row.Keybase != nil {
 			response.List[i].Keybase.Username = strings.ToLower(row.Keybase.Username)
+			response.List[i].RawScore = row.Keybase.RawScore
 		}
 		if row.Service != nil {
 			response.List[i].Service.Username = strings.ToLower(row.Service.Username)
@@ -132,7 +133,7 @@ var fieldsAndScores = []struct {
 	multiplier float64
 	getter     func(*keybase1.ProcessedContact) string
 }{
-	{1.0, func(contact *keybase1.ProcessedContact) string { return contact.ContactName }},
+	{1.5, func(contact *keybase1.ProcessedContact) string { return contact.ContactName }},
 	{1.0, func(contact *keybase1.ProcessedContact) string { return contact.Component.ValueString() }},
 	{1.0, func(contact *keybase1.ProcessedContact) string { return contact.DisplayName }},
 	{0.8, func(contact *keybase1.ProcessedContact) string { return contact.DisplayLabel }},
@@ -171,19 +172,11 @@ func contactSearch(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []ke
 		found, score := matchAndScoreContact(query, c)
 		if found {
 			contact := c
-			contact.RawScore = score
 			res = append(res, keybase1.APIUserSearchResult{
-				Score:   score,
-				Contact: &contact,
+				Contact:  &contact,
+				RawScore: score,
 			})
 		}
-	}
-
-	sort.Slice(res, func(i, j int) bool {
-		return res[i].Score > res[j].Score
-	})
-	for i := range res {
-		res[i].Score = 1.0 / float64(1+i)
 	}
 
 	return res, nil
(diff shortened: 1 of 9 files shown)


zapu:
cc @jzila i didn't get to finish it yesterday, but here it is. mostly golang changes

go/service/usersearch.go
```diff
@@ -129,29 +130,34 @@ func (q *compiledQuery) scoreString(str string) (bool, float64) {
...
-	{1.0, func(contact *keybase1.ProcessedContact) string { return contact.DisplayName }},
-	{0.8, func(contact *keybase1.ProcessedContact) string { return contact.DisplayLabel }},
-	{0.7, func(contact *keybase1.ProcessedContact) string { return contact.FullName }},
-	{0.7, func(contact *keybase1.ProcessedContact) string { return contact.Username }},
+	{1.5, true, func(contact *keybase1.ProcessedContact) string { return contact.ContactName }},
+	{1.0, true, func(contact *keybase1.ProcessedContact) string { return contact.Component.ValueString() }},
+	{1.0, false, func(contact *keybase1.ProcessedContact) string { return contact.DisplayName }},
+	{0.8, false, func(contact *keybase1.ProcessedContact) string { return contact.DisplayLabel }},
+	{0.7, false, func(contact *keybase1.ProcessedContact) string { return contact.FullName }},
+	{0.7, false, func(contact *keybase1.ProcessedContact) string { return contact.Username }},
```
jzila: Where do these multiplier values come from? And why are they not all distinct?

go/service/usersearch.go
```diff
@@ -167,24 +173,23 @@ func contactSearch(mctx libkb.MetaContext, arg keybase1.UserSearchArg) (res []ke
...
+		found, score, matchedVal := matchAndScoreContact(query, c)
 		if found {
 			contact := c
-			contact.RawScore = score
+			if contact.Resolved && matchedVal != "" {
+				// If contact is resolved, make sure to plumb matched query to
+				// display label. This is not needed for unresolved contacts,
+				// which can only match on ContactName or Component Value, and
+				// both of them always appear as name and label.
+				contact.DisplayLabel = matchedVal
```
jzila: We talked about formatting returned phone numbers here, so we don't have to do a bunch of "isLabelPhoneNumber" and "isNamePhoneNumber" branching logic in the frontend. Is that still the plan?

go/service/usersearch.go
```diff
@@ -197,18 +202,51 @@ func (h *UserSearchHandler) UserSearch(ctx context.Context, arg keybase1.UserSea
...
-		}
-		if len(contactsRes) > 0 {
-			var res2 []keybase1.APIUserSearchResult
-			res2 = append(res2, contactsRes...)
-			res2 = append(res2, res...)
-			res = res2
+			mctx.Warning("Failed to do contacts search: %s", err)
+		} else {
+			res = append(res, contactsRes...)
+			// Sort first - we are going to be deduplicating on usernames,
```
jzila: What happened to deduplicating on username and display label? Keep in mind that one name can show up multiple times if you have multiple forms of match.

e.g. One contact has a phone number and an email. You searched for their contact name, so each phone number should appear as an entry, as should their email.

go/service/usersearch.go
```diff
@@ -197,18 +202,51 @@ func (h *UserSearchHandler) UserSearch(ctx context.Context, arg keybase1.UserSea
...
+			// entries with higher score have precedence.
+			sort.Slice(res, func(i, j int) bool {
+				return res[i].RawScore > res[j].RawScore
+			})
```
jzila: If you sort by your deduping metric first (`username`, or `{username, displayLabel}` depending on your answers to my previous questions), then you can get away with not building the set at all. Sort once by deduping metric, then iterate through the list, picking the highest score of any dupes, and then one final sort by score.

jzila: Alternatively, you can build a heap indexed by deduping metric, only overwriting if the score is higher, then pull each entry into a new heap, now indexed on score. Finally you can pull out `MaxResults` entries from the heap.

go/service/usersearch.go
zapu: It's still part of the grand plan. There are few issues with current design, e.g. it should not do the deduplication during contact syncing stage, but rather save everything and only process that when fetching contacts.

I wanted this PR to be a simple fix to most apparent issues.

go/service/usersearch.go
zapu: I really don't want to spend too much time on that file right now. There are more substantial issues in how contact syncing works right now that the way they are fetched here might change anyway.

go/service/usersearch.go
zapu: Arbitrarily chosen, we wanted to bump search results higher up in the list if query matches contact name or component value.

go/service/usersearch.go
zapu: I don't follow, did we used to have a feature and this removes it? I may be a bit lost of what's going on in different PRs.

This one had a specific things to fix in mind - so two result from both resolved contact and keybase search do not show up for the same user.

jzila: I'm worried that this deduplication will remove too many contacts, instead of returning multiple entries per contact as is sometimes required.

PR #18499, opened 2019-07-21
prepend @ mention to replies

mmaxim:
Patch does the following:

1. Add ability to prepend text into the input bar.
2. Prepend an @ mention in for replies.
3. Fix input bar focus when setting text from the store.


chrisnojima:
i didn't dig into it but it fails for me sometimes and its not clear why.  i'm in a team convo and replying and in some convos its ok but other it doesn't prepend

mmaxim:
Maybe send me a video? I can repro this in the case where you have a prepend from a different message in there. Is that what you hit?

PR #18501, opened 2019-07-22
swipe right to reply

mmaxim:
Use gesture library to make swipe right to reply happen.

### DIFF
--- a/shared/chat/conversation/messages/wrapper/long-pressable/index.native.tsx
+++ b/shared/chat/conversation/messages/wrapper/long-pressable/index.native.tsx
@@ -23,9 +23,7 @@ const LongPressable = (props: {children: React.ElementType; onSwipeRight: () => 
   return (
     <Swipeable ref={swipeable} renderLeftActions={_renderLeftActions} onSwipeableLeftWillOpen={_onLeftOpen}>
       <Kb.NativeTouchableHighlight key="longPressbale" {...rest}>
-        <Kb.NativeView key="longPressable" style={styles.view}>
-          {children}
-        </Kb.NativeView>
+        <Kb.NativeView style={styles.view}>{children}</Kb.NativeView>
       </Kb.NativeTouchableHighlight>
     </Swipeable>
   )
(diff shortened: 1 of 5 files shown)


shared/chat/conversation/messages/wrapper/long-pressable/index.native.tsx
```diff
@@ -1,18 +1,38 @@
...
 
 // See '.js.flow' for explanation
-const LongPressable = (props: {children: React.ElementType}) => {
+const LongPressable = (props: {children: React.ElementType; onSwipeRight: () => void}) => {
   const {children, ...rest} = props
+  const swipeable = React.useRef()
+  const _onLeftOpen = () => {
+    props.onSwipeRight()
+    // @ts-ignore the type returned by useRef doesn't seem to work, it always thinks current can be null
+    swipeable.current && swipeable.current.close()
```
chrisnojima: Current can be null

chrisnojima: Store the value in a const to fix the ts issue

chrisnojima:
I need to perf test to see if we're thrashing any

shared/chat/conversation/messages/wrapper/long-pressable/index.native.tsx
```diff
@@ -1,18 +1,38 @@
...
+    </Kb.Box2>
+  )
+}
 
 // See '.js.flow' for explanation
-const LongPressable = (props: {children: React.ElementType}) => {
+const LongPressable = (props: {children: React.ElementType; onSwipeRight: () => void}) => {
   const {children, ...rest} = props
+  const swipeable = React.useRef()
+  const _onLeftOpen = () => {
```
chrisnojima: we usually use the `_` prefix to indicate a module/class private thing. Since this is a closure variable i don't think its necessary. Tangentially TS supports `private` class member annotations so we can use that going forward in the class case.

shared/chat/conversation/messages/wrapper/long-pressable/index.native.tsx
```diff
@@ -1,18 +1,38 @@
 import * as React from 'react'
 import * as Kb from '../../../../../common-adapters/mobile.native'
 import * as Styles from '../../../../../styles'
+import Swipeable from 'react-native-gesture-handler/Swipeable'
```
chrisnojima: this should move into `common-adapters/mobile.native`

chrisnojima:
a quick glance the perf seems fine. i think the upstream components do a good job of stopping the renders. i think ux wise it might make sense to color the background the action. a lot of apps do this so you can tell there's actions and to differentiate it

PR #18504, opened 2019-07-22
fix resend sms during signup

pzduniak:
also includes an onenterkeydown in pw reset because it was frustrating to create 20 accounts to debug this without proper keydowns

### DIFF
--- a/shared/settings/password/index.tsx
+++ b/shared/settings/password/index.tsx
@@ -139,6 +139,7 @@ class UpdatePassword extends Component<Props, State> {
               keyboardType={keyboardType}
               value={this.state.passwordConfirm}
               onChangeText={password => this._handlePasswordConfirmChange(password)}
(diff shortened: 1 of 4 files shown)


shared/settings/password/index.tsx
```diff
@@ -139,6 +139,7 @@ class UpdatePassword extends Component<Props, State> {
               keyboardType={keyboardType}
               value={this.state.passwordConfirm}
               onChangeText={password => this._handlePasswordConfirmChange(password)}
+              onEnterKeyDown={() => this.props.onSave(this.state.password, this.state.passwordConfirm)}
```
buoyad: make sure this is disabled at the same time the save button is

pzduniak: bleh i thought onsave checked it, it blocked on the rpcs instead

PR #18506, opened 2019-07-22
Make the profile page work with a banner

nathunsmitty:
Tweaks some CSS styles on desktop to ensure that the airdrop banner works with the profile page. The issue with the profile page was the use of absolute positioning to achieve the effect of coloring the top of the header.

My solution to this was to make the banner part of the top header, and have the profile page adjust the size of the padding it would use if the banner is present.

Note this only affects desktop as the banner is not shown on any page but the people page on mobile.

### DIFF
--- a/shared/profile/user/index.tsx
+++ b/shared/profile/user/index.tsx
@@ -375,7 +375,8 @@ class User extends React.Component<Props, State> {
           fullWidth={true}
           fullHeight={true}
           style={Styles.collapseStyles([
-            {...styles.container, paddingTop},
+            styles.container,
+            {paddingTop},
             colorTypeToStyle(this.props.backgroundColorType),
           ])}
         >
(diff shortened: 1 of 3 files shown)


nathunsmitty:
@keybase/react-hackers

shared/profile/user/index.tsx
```diff
@@ -371,7 +374,10 @@ class User extends React.Component<Props, State> {
           direction="vertical"
           fullWidth={true}
           fullHeight={true}
-          style={Styles.collapseStyles([styles.container, colorTypeToStyle(this.props.backgroundColorType)])}
+          style={Styles.collapseStyles([
+            {...styles.container, paddingTop},
```
cjb: I think collapseStyles is supposed to avoid spreads, would go with a three elem array instead

PR #18508, opened 2019-07-22
White backgroundColor on reaction tooltip

buoyad:
On master when the reaction tooltip is scrollable what's underneath can peek through when scrolled to the bottom. Fix by adding a backgroundColor to the overlay.


buoyad:
@keybase/react-hackers bump, one liner

PR #18510, opened 2019-07-22
unused vars TS flag: Part 1

chrisnojima:
@keybase/react-hackers turns on the unused vars flag (in the last part)


shared/chat/conversation/input-area/suggestors/index.stories.tsx
```diff
@@ -37,14 +37,14 @@ class _TestArea extends React.Component<TestAreaProps> {
...
+// @ts-ignore
 const typingTests = () => {
-  // eslint-disable-next-line no-unused-vars
-  let test
+  let test: unknown
   // @ts-ignore should error (bad other prop)
   test = <TestArea {...props} somethingElse="not this" />
   // @ts-ignore should error (bad suggestor prop)
   test = <TestArea {...props} dataSources={[1, 2]} />
+  console.log(test)
```
cjb: :fire: ?

chrisnojima: this code is actually never ran its just so TS complains if the data doesn't match up. we never call this function. this stops TS further complaining that we never use test

PR #18513, opened 2019-07-22
TS: no unused vars 1

chrisnojima:
@keybase/react-hackers this turns on no unused vars


shared/actions/settings.tsx
```diff
@@ -670,12 +668,11 @@ const loadContactImportEnabled = async (
   }
   let enabled = false
   try {
-    const res = await RPCTypes.configGuiGetValueRpcPromise(
-      {
-        path: Constants.importContactsConfigKey(state.config.username),
-      },
+    const value = await RPCTypes.configGuiGetValueRpcPromise(
```
chrisnojima: if we're using await lets just use it and not mix that and promise thens

shared/tracker2/remote-proxy.desktop.tsx
```diff
@@ -20,6 +20,7 @@ const windowOpts = {hasShadow: false, height: 470, transparent: true, width: 320
 const trackerMapStateToProps = (state, ownProps) => {
   const d = Constants.getDetails(state, ownProps.username)
   return {
+    airdropIsLive: state.wallets.airdropDetails.isPromoted,
```
chrisnojima: @keybase/picnicsquad please validate this

PR #18514, opened 2019-07-22
TS: no unused vars and no implicit returns

chrisnojima:
@keybase/react-hackers


chrisnojima:
sadly no. then lint will complain about an unneeded return! we can change these rules but i think its fine for now. i think the ugliness of it will make us change it. in a lot of cases we really could change it to be of the form `() => check && Saga.put()`. i just added return undefined to do it quickly and not to have to change it. but honestly we have very few of these cases (i was surprised actually)

PR #18515, opened 2019-07-22
Always load configured accounts on daemon handshare/loggedin/loggedout

buoyad:
Fix getting to the relogin screen on mobile. cc @keybase/y2ksquad

### DIFF
--- a/shared/actions/config/index.tsx
+++ b/shared/actions/config/index.tsx
@@ -223,15 +223,12 @@ function* loadDaemonAccounts(
       )
     }
 
-    // only reload in the user-switching case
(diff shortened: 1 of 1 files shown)


shared/actions/config/index.tsx
```diff
@@ -223,15 +223,12 @@ function* loadDaemonAccounts(
       )
     }
 
-    // only reload in the user-switching case
-    const loadConfiguredAccountsAgain = state.config.loggedIn && flags.fastAccountSwitch
```
jakob223: whoops, I guess this is overly restrictive. I'll come up with a new heuristic, I guess, so as to not be calling this super duper often.

PR #18516, opened 2019-07-22
Increase targetBufferSize so that 99% of ReadB64 calls fit into the b…

patrickxb:
…uffer

### DIFF
--- a/go/bind/keybase.go
+++ b/go/bind/keybase.go
@@ -323,7 +323,7 @@ func WriteB64(str string) (err error) {
 	return nil
 }
 
-const targetBufferSize = 50 * 1024
+const targetBufferSize = 300 * 1024
 
 // bufferSize must be divisible by 3 to ensure that we don't split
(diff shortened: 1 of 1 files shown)


patrickxb:
cc @chrisnojima

patrickxb:
i'm trying this now on 4.3.0 (22) ... can't tell a difference.

chrisnojima:
We can merge and revert if it's weird

go/bind/keybase.go
```diff
@@ -342,6 +342,10 @@ func ReadB64() (res string, err error) {
 	}
 	n, err := conn.Read(buffer)
 	if n > 0 && err == nil {
+		if n == bufferSize {
+			kbCtx.Log.Info("ReadB64 %d bytes (full buffer)", n)
```
mmaxim: Is this output useful?

patrickxb: I removed it.

PR #18519, opened 2019-07-22
Fix double gapu in keybase status

heronhaye:
We were calling GAPU twice in keybase status, this is slow on machines with a lot of provisioned users.


heronhaye:
ok merging over TestChatSrvLiveLocationMultiple and TestConversationLockingDeadlock flakes...

PR #18522, opened 2019-07-22
improve phrasing in delete phone/email modal

pzduniak:


shared/settings/account/email-phone-row.tsx
```diff
@@ -197,10 +209,19 @@ const mapDispatchToProps = (dispatch: Container.TypedDispatch, ownProps: OwnProp
       dispatch(SettingsGen.createResendVerificationForPhoneNumber({phoneNumber}))
       dispatch(RouteTreeGen.createNavigateAppend({path: ['settingsVerifyPhone']}))
     },
-    onDelete: () =>
+    _onDelete: (address: string, discoverable: boolean) =>
```
jakob223: why pass these in instead of just getting them from `OwnProps`?

pzduniak: Address is in ownprops, discoverable isn't + there's no way to plumb discoverable from the parent. Instead of mixing idioms I thought I'd just use one.

PR #18523, opened 2019-07-22
Calculate whether to show the unread shortcut more often on desktop

buoyad:
@malgorithms noticed an issue: if you try to find unread team conversations by scrolling until the unread shortcut disappears, it won't work because of the long delay between scrolling and the shortcut updating. This patch moves the unread shortcut calculation out of the debounced function so it will run on every `onItemsRendered`. cc @keybase/y2ksquad

### DIFF
--- a/shared/chat/inbox/index.desktop.tsx
+++ b/shared/chat/inbox/index.desktop.tsx
@@ -136,9 +136,13 @@ class Inbox extends React.PureComponent<T.Props, State> {
     this.setState(old => (old.showFloating !== showFloating ? {showFloating} : null))
   }
 
-  _onItemsRendered = debounce(({visibleStartIndex, visibleStopIndex}) => {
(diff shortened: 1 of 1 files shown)


mmaxim:
Maybe throttle it?

chrisnojima:
agree w/ the throttle. it'll be a 1 liner and likely better perf

buoyad:
@chrisnojima / @mmaxim how can we throttle and ensure we recalculate right when we pass the unread rows?

chrisnojima:
I don't see why it's different than the existing solution

mmaxim:
@buoyad The problem with debounce for this is it doesn’t do anything until you stop scrolling, which means you can scroll way by and not have this respond. A throttle of like 50ms probably solves any scroll spam perf problem, and won’t be noticeable in lag in the UI. Agree?

buoyad:
Right, right. I was confusing the existing behavior with what throttle does. Sounds good.

PR #18527, opened 2019-07-23
[Y2K-228] import contacts row

jzila:
* [x] - "Later" button doesn't format correctly: padding doesn't seem to work
* [x] - Importing contacts doesn't refresh search results
* [x] - handle `never_ask_again`
* [ ] - Importing contacts doesn't refresh recommendations

@buoyad I might need some help with the above TODOs.

### DIFF
--- a/shared/team-building/index.tsx
+++ b/shared/team-building/index.tsx
@@ -91,9 +91,6 @@ const ContactsBanner = ({onRedoSearch}: {onRedoSearch: () => void}) => {
     }
   }, [numContactsImported, prevNumContactsImported, onRedoSearch])
 
-  if (!Flags.sbsContacts) return null
-  if (!Styles.isMobile) return null
-
   // Ensure that we know whether contacts are loaded, and if not, that we load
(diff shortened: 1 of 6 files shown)


shared/team-building/index.tsx
```diff
@@ -62,6 +64,75 @@ type Props = {
   rolePickerProps?: RolePickerProps
 }
 
+const ContactsBanner = ({onRedoSearch}: {onRedoSearch: () => void}) => {
+  const dispatch = Container.useDispatch()
+
+  const contactsImported = Container.useSelector(s => s.settings.contacts.importEnabled)
+  const arePermissionsGranted = Container.useSelector(s => s.settings.contacts.permissionStatus)
```
buoyad: We should handle the `never_ask_again` case here too

jzila:
@buoyad let's call this ready. I'm struggling with reloading the recommendations due to them being namespaced, and having to deal with races. So I'll put that in a separate ticket.

PR #18529, opened 2019-07-23
invite banner phrasing

pzduniak:
prettier messed up the syntax, i could move it to another block of code

### DIFF
--- a/shared/chat/conversation/bottom-banner/index.tsx
+++ b/shared/chat/conversation/bottom-banner/index.tsx
@@ -109,11 +109,19 @@ const InviteBanner = ({users, openSMS, openShareSheet}: InviteProps) => {
   }
 
   const hasPhoneNumber = users.some(user => user.endsWith('@phone'))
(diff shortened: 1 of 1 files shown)


shared/chat/conversation/bottom-banner/index.tsx
```diff
@@ -109,11 +109,19 @@ const InviteBanner = ({users, openSMS, openShareSheet}: InviteProps) => {
   }
 
   const hasPhoneNumber = users.some(user => user.endsWith('@phone'))
+  const hasEmailAddress = users.some(user => user.endsWith('@email'))
   return (
     <BannerBox color={Styles.globalColors.blue}>
       <BannerText>
         Your messages will unlock once they join Keybase
-        {hasPhoneNumber && ' and verify their phone number'}.
+        {hasPhoneNumber
```
buoyad: let's derive this string earlier in the function

pzduniak: k rdy

PR #18532, opened 2019-07-23
Wrap Icon in a padded Box to reposition the tooltip.

adamjspooner:


shared/chat/conversation/messages/react-button/emoji-row/index.tsx
```diff
@@ -13,6 +13,15 @@ type Props = {
   style?: Styles.StylesCrossPlatform
 }
 
+const HoverBox = Styles.isMobile
+  ? Kb.Box
+  : Styles.styled(Kb.Box)({
+      ...Styles.desktopStyles.clickable,
+      ':hover .icon': {
+        color: Styles.globalColors.blue,
```
buoyad: we have [some classes](https://github.com/keybase/client/blob/master/shared/styles/index.desktop.tsx#L122) that are supposed to do this, try `hover_color_blue`

adamjspooner: Sadly, that doesn't work in this case (`Icon` in a `Box2`).

chrisnojima: use hover-container, see style.css

adamjspooner: Thanks!

PR #18537, opened 2019-07-23
Update provisioning add device error as per design

taruti:


cecileboucheron:
Could we have the banner take up the full viewport width?

taruti:
Sure, I'll update it!

PR #18540, opened 2019-07-23
Fix undefined fcn call in select-other-device

buoyad:
Fixes crasher on selecting "reset your account" on the select-other-device page in provisioning. cc @keybase/y2ksquad

### DIFF
--- a/shared/provision/select-other-device/container.tsx
+++ b/shared/provision/select-other-device/container.tsx
@@ -1,42 +1,41 @@
 import * as ProvisionGen from '../../actions/provision-gen'
 import * as LoginGen from '../../actions/login-gen'
+import * as RouteTreeGen from '../../actions/route-tree-gen'
 import SelectOtherDevice from '.'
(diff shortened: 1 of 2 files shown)


shared/provision/select-other-device/container.tsx
```diff
@@ -1,42 +1,41 @@
...
   configuredAccounts: state.config.configuredAccounts,
   devices: state.provision.devices,
 })
-const mapDispatchToProps = (dispatch: TypedDispatch, ownProps: OwnProps) => ({
+const mapDispatchToProps = (dispatch: Container.TypedDispatch, _: OwnProps) => ({
   onLogIn: (username: string) => dispatch(LoginGen.createLogin({password: new HiddenString(''), username})),
   onResetAccount: () => {
     dispatch(LoginGen.createLaunchAccountResetWebPage())
-    dispatch(ownProps.navigateUp())
+    dispatch(RouteTreeGen.createNavigateUp())
```
buoyad: This is the fix. The rest is `* as Container` and removing `compose`.

buoyad:
@keybase/react-hackers one liner

PR #18544, opened 2019-07-23
Server-provided cache freshness for imptofu

zapu:


go/contacts/cache.go
```diff
@@ -75,67 +75,63 @@ func newLookupResultCache() (ret lookupResultCache) {
...
-	return minimumFreshness
-}
-
-func (c cachedLookupResult) getFreshness() time.Duration {
-	if c.Resolved {
-		return contactCacheFreshness
-	}
-	return unresolvedContactCacheFreshness
+func (c cachedLookupResult) getLongStaleDate() time.Time {
+	return c.ExpiresAt.Add(minimumFreshness)
```
heronhaye: so for resolved contacts it's 30 days + 45 days?

go/contacts/cache.go
```diff
@@ -157,33 +153,42 @@ func (c *CachedContactsProvider) LookupAll(mctx libkb.MetaContext, emails []keyb
 		conCache = newLookupResultCache()
 	} else {
 		mctx.Debug("Fetched cache, current cache size: %d", len(conCache.Lookups))
+		conCache.Version.Major = cacheCurrentMajorVersion
```
heronhaye: ?

also, where is the part where it throws away result if it's old major version?

go/contacts/cache_test.go
```diff
@@ -184,13 +185,25 @@ func TestLookupCacheExpiration(t *testing.T) {
...
+
+	provider.queryCount = 0
+
+	// Push us over resolved contact cache expiration time.
+	clock.Advance(10*24*time.Hour + time.Hour) // see *MockContactsProvider::LookupAll for correct value
+
+	res4, err := ResolveContacts(libkb.NewMetaContextForTest(tc), cacheProvider, contactList, keybase1.RegionCode(""))
+	require.NoError(t, err)
+	require.Equal(t, res1, res4)
+
```
heronhaye: test for going past the minimumFreshness?

go/contacts/cache.go
```diff
@@ -157,33 +153,42 @@ func (c *CachedContactsProvider) LookupAll(mctx libkb.MetaContext, emails []keyb
...
+		conCache.Version.Minor = cacheCurrentMinorVersion
 	}
 
 	var remainingEmails []keybase1.EmailAddress
 	var remainingNumbers []keybase1.RawPhoneNumber
 
 	mctx.Debug("Populating results from cache")
 
+	// Map of keys of new unresolved cache entries, to set ExpireAt value after
```
heronhaye:
```suggestion
	// List of keys of new unresolved cache entries, to set ExpireAt value after
```

heronhaye: Also it's not just unresolved entries, stale resolved entries are looked up too right?

go/contacts/cache.go
heronhaye: (in particular, don't we need to do that in this pr because we're changing the cache type?)

zapu: I thought it didn't have to for this one, it will load expireAt to 0 for all entries and they will be refreshed (or deleted) next time LookupAll is performed. So it might be a minor inconvenience for people but for now we are only testing internally. Let me know if you think that's not the case.

go/contacts/cache.go
zapu: yes, thanks

go/contacts/cache.go
heronhaye: good point, it'd've been fine then

PR #18554, opened 2019-07-24
reply swipe touchup

mmaxim:
I believe this gets swiping right to get back to the inbox on par with the release version. The prop `failOffsetX` is part of `PanGestureHandler` (which is what `Swipeable` uses), and is used to early out the underlying `UIGestureRecognizer` if the translation on the X axis is outside some range. Giving 0 seems to block any swipe right from triggering this recognizer, which makes it so the touches don't get cancelled. 

Also increase padding on the reply arrow.


mmaxim:
cc @mlsteele

PR #18555, opened 2019-07-24
keybase:// URL scheme for paramproofs and other app links

cjb:
@keybase/react-hackers, CC
@keybase/picnicsquad
@xgess (for paramproofs)
@songgao @jzila (for the Files links)
@max @mmaxim (for chat and general interest in the URL structure)

Here's code to support the following custom URL scheme links:

[keybase://profile/new-proof/keybase-test.demo.discourse.org](keybase://profile/new-proof/keybase-test.demo.discourse.org)
[keybase://public/chris](keybase://public/chris)
[keybase://team/keybasefriends](keybase:/team/keybasefriends)
[keybase://chat/max](keybase://chat/max)
[keybase://chat/max,chris](keybase://chat/max,chris)
[keybase://chat/keybasefriends#general](keybase://chat/keybasefriends#general)

And examples of failing links that show an error:

[keybase://profile/bad-path](keybase://profile/bad-path)
[keybase://team/nonexistentteam](keybase://team/nonexistentteam)
[keybase://chat/foo/bar/baz](keybase://chat/foo/bar/baz)
[keybase://chat/keybasefriends#nonexistentchannel](keybase://chat/keybasefriends#nonexistentchannel)

I went with simple string splitting in JS rather than using a real URL parser, seemed easier to convince ourselves that it's secure and correct.  Very open to feedback on the URL layout or the handling.  The link handler's moved from `actions/config.tsx` to `actions/deeplinks.tsx` now.

There's a strategy here that links should only function as GET requests would in REST -- they can take you to a screen, but not prefill or confirm anything once you're there.  I didn't add support for opening a prefilled join-a-team screen from a link because it violates that strategy, but maybe that's actually a reasonable thing to do.

The component that was previously `shared/wallets/sep7-confirm/error.tsx` is now at `shared/deeplinks/error.tsx` and handles both SEP7 link errors and app link errors.


cjb:
@mmaxim I'd like to get `keybase://.*` and `web+stellar:.*` links showing up as links in Chat, and when you click on one I'd intercept the call out to the system browser, and just handle it internally instead of invoking the browser.  So I think the first step is to get them marked up as links in Chat messages.  I started tracking that down towards https://github.com/keybase/xurls, am I on the right track?

mmaxim:
@cjb, yes that sounds right, you probably just need to add `keybase://` as a known URL scheme to that fork, it that might just do it for you.

Also, as for the chat links, I think it would be cool to link to a specific message. Something like:
keybase://chat/max/567

You could get that link from within the UI.

mlsteele:
Proof creation would benefit from filling the username. One less thing for the user to do.
`keybase://profile/new-proof/buster@keybase-test.demo.discourse.org`
`keybase://profile/new-proof/keybase-test.demo.discourse.org/scruggs`
As a justification for filling, navigating to the site is similar to filling an option selector. And `web+stellar` already fills.

xgess:
> Proof creation would benefit from filling the username. One less thing for the user to do.
> `keybase://profile/new-proof/buster@keybase-test.demo.discourse.org`
> `keybase://profile/new-proof/keybase-test.demo.discourse.org/scruggs`
> As a justification for filling, navigating to the site is similar to filling an option selector. And `web+stellar` already fills.

+1. i was actually trying to see if it would work already because it looks like the startProof event has an empty username field. but `keybase://profile/new-proof/keybase-test.demo.discourse.org?username=xgess` threw a warning with no UI feedback.

xgess:
it occurred to me that these team links look really similar to our git links. so, like:
`keybase://team/keybasefriends/list-members` could take you to the members of the team or do a git repo called `list-members`. :/ 
the git links are pretty established. `keybase://team-info/teamname`? or `team/info/teamname`. all around unfortunate. wish we could go back and make the git links like `kbgit://` or something more scoped. 🤷‍♂

cjb:
I think I'd be okay with going with:

keybase://teams/keybasefriend
keybase://teams/keybasefriends/members
keybase://teams/keybasefriends/join

Even though `team` meaning KBFS and `teams` meaning Team is confusing.

cjb:
@xgess @mlsteele Just pushed a comment to make:

keybase://profile/new-proof/keybase-test.demo.discourse.org/prefilledusername

work 🙏

xgess:
> keybase://profile/new-proof/keybase-test.demo.discourse.org/prefilledusername

looks good! was trying to break it with existing proofs and proofs of some of the weirder types (e.g. bitcoin). first pass and nothing unexpected. very nice. do we want to offer feedback when it fails? it's a little weird if you click on a link somewhere to add a keybase proof, and then it's just normal keybase with nothing going on. maybe not a blackbar, but some kind of notice that the link was malformed?

cjb:
@xgess Pushed a change to show the error screen when the service in a link doesn't exist! Ready for review now.

shared/actions/deeplinks.tsx
```diff
@@ -0,0 +1,114 @@
...
+      }
+      break
+    // Fall-through
+    case 'private':
+    case 'public':
+    case 'team':
+      return [
+        RouteTreeGen.createSwitchTab({tab: Tabs.fsTab}),
+        RouteTreeGen.createNavigateAppend({
+          path: [{props: {path: '/keybase/' + parts.join('/')}, selected: 'main'}],
```
songgao: Could just do `/keybase/${action.payload.link}` here instead of joining?

shared/actions/deeplinks.tsx
```diff
@@ -0,0 +1,114 @@
...
+  } else {
+    // Normal deeplink
+    const url = new URL(action.payload.link)
+    const username = Constants.urlToUsername(url)
+    if (username) {
+      return [
+        RouteTreeGen.createNavigateAppend({path: [Tabs.peopleTab]}),
+        ProfileGen.createShowUserProfile({username}),
+      ]
+    }
```
songgao: This is for https links on mobile right? I wonder if it'd make sense to support stuff like https://links.keybase.io/chat https://links.keybase.io/team/keybase/file here as well since we already have the code handling those in the app.

shared/actions/deeplinks.tsx
cjb: Thanks, done.

shared/actions/deeplinks.tsx
cjb: Not sure how I feel about this.  An advantage of the protocol handler is that you don't have to tell Keybase's web server every time you're performing a Keybase interaction, and tell it what the interaction was specifically.  It seems like using the web redirect introduces some more server trust and server knowledge, and I'm not sure what we would gain in return over just using keybase://.

cjb: Specifically, note that https://links.keybase/io/team/someteam/somefile would normally be encrypted metadata, not plaintext.  If someone shared a keybase:// link to it in a private place, it would stay private, but if someone shared a https://links.keybase.io/ link to it in a private place, it becomes disclosed to us (and e.g. a corporate TLS proxy if they are using one).

xgess:
`keybase://profile/garbage/keybase-test.demo.discourse.org/ness7` <-- that link doesn't throw an error

shared/actions/deeplinks.tsx
```diff
@@ -0,0 +1,113 @@
...
+      }
+    default:
+      return [
+        DeeplinksGen.createSetKeybaseLinkError({error}),
+        RouteTreeGen.createNavigateAppend({
+          path: [{props: {errorSource: 'app'}, selected: 'keybaseLinkError'}],
+        }),
+      ]
+  }
+  return undefined
```
xgess: maybe this should set the default error. then in this branching, we can just break whenever we see something that doesn't match.

shared/actions/deeplinks.tsx
```diff
@@ -0,0 +1,113 @@
...
+            DeeplinksGen.createSetKeybaseLinkError({error}),
+            RouteTreeGen.createNavigateAppend({
+              path: [{props: {errorSource: 'app'}, selected: 'keybaseLinkError'}],
+            }),
+          ]
+        }
+        const [teamname, channelname] = teamChat
+        return [
+          RouteTreeGen.createSwitchTab({tab: Tabs.chatTab}),
+          ChatGen.createPreviewConversation({channelname, reason: 'appLink', teamname}),
```
xgess: this blackbars for an unknown team `keybase://chat/unknownteam#general`

shared/actions/deeplinks.tsx
```diff
@@ -0,0 +1,113 @@
...
+        ]
+      } else {
+        return [
+          RouteTreeGen.createSwitchTab({tab: Tabs.chatTab}),
+          ChatGen.createPreviewConversation({participants: parts[1].split(','), reason: 'appLink'}),
```
xgess: `keybase://chat/idontexist` navigates to chat with no errors or feedback. maybe we're ok with that, especially as a first pass. looks like it could be tricky to handle those errors here.

shared/actions/deeplinks.tsx
cjb: Makes sense!  Pushed a change, how's this?

shared/actions/deeplinks.tsx
cjb: Thanks, pushed a fix.

packaging/linux/keybase.desktop
```diff
@@ -6,4 +6,4 @@ Terminal=false
 Type=Application
 Categories=Network;
 StartupNotify=false
-MimeType=x-scheme-handler/web+stellar;
+MimeType=x-scheme-handler/web+stellar;x-scheme-handler/keybase;
```
heronhaye: what is the "web+" for in stellar://? do we need it?
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler
>For security reasons, registerProtocolHandler() has restrictions on which schemes may be registered. A custom scheme may be registered as long as the scheme's name begins with "web+", is at least five characters long (including the "web+" prefix), and has only lower-case ASCII letters in its name. For example, "web+burger", as shown in the Example below.

shared/actions/deeplinks.tsx
songgao: Got it; good point! I recall briefly discussing this with Max in in 1:1. No definitely answer came out of that, but one idea we liked was to encrypt the path with a PTK, and client would just decrypt that to get the real path. We'd need to add a KID, which shortens the available space in the URL, but should still work for most paths.

Either way, for chat and profile it'd be an easier story. Perhaps we can support those as a first step? (Not suggesting we should do it in this PR; just thinking out loud about ideas.)

packaging/linux/keybase.desktop
cjb: @heronhaye that's a web browser API -- we're registering a scheme handler outside a web browser, so it can be whichever scheme we like.

`web+stellar:` was chosen by the Stellar folks, I guess they wanted it to be registerable by web pages too?  We don't want `keybase://` to be registerable by web pages, so I think we're good as-is.

xgess: @heronhaye those are sep7 links https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0007.md

heronhaye: oh interesting. sounds good then. i think you have to add a case in run_keybase, it only accepts web+ links right now

cjb: @heronhaye Thanks!  How's 8315da4ba9?  I noticed we had an unescaped `+`, too.

shared/actions/deeplinks.tsx
cjb: @songgao I guess I feel like the only reason we would bother with all this work is because `keybase://` didn't exist before.  But now it does, so why not just use it?

packaging/linux/keybase.desktop
heronhaye: nice catch. seems ^(web\+|keybase).* needs to be in single quotes too (`'^(web\+|keybase).*'`). so tricky...

cjb: @heronhaye single quotes seems to break it for me

shared/actions/deeplinks.tsx
songgao: I'm not saying we should use https instead of keybase://. I agree we should use it on desktop. The reason why I brought it up was (I thought) we can't use `keybase://` on mobile, but only https://. Is that not true?

songgao: (and I'm realizing I didn't mention mobile in my first comment. Sorry!)

cjb: @songgao Not true!  `keybase://` works just the same on mobile after this PR :)

songgao: nice! 🎉

packaging/linux/keybase.desktop
heronhaye: my bad, I was testing on zsh instead of bash. lgtm

shared/desktop/app/node.desktop.tsx
```diff
@@ -73,8 +73,11 @@ const focusSelfOnAnotherInstanceLaunching = (_, commandLine) => {
 
   // The new instance might be due to a URL schema handler launch.
   logger.info('Launched with URL', commandLine)
-  if (commandLine.length > 1 && commandLine[1] && commandLine[1].startsWith('web+stellar:')) {
-    sendToMainWindow('dispatchAction', {payload: {link: commandLine[1]}, type: ConfigGen.link})
+  if (commandLine.length > 1 && commandLine[1]) {
+    const link = commandLine[1]
+    if (link.startsWith('web+stellar:') || link.startsWith('keybase://')) {
+      sendToMainWindow('dispatchAction', {payload: {link}, type: DeeplinksGen.link})
+    }
```
eli-schwartz: If I understand correctly, this is only checking argv1, i.e. the calling mechanism must be e.g. `./Keybase keybase://chat/username`. (And I guess this was true for the web+stellar:// links before this PR.)

This means you cannot use electron's advertised alternative invocation method: `/usr/lib/electron/electron path-to-app keybase://chat/username` (because it is in argv2 in this case).
As such, the Arch Linux package I maintain which uses a generic electron binary does not register these command-line options.

Is there something that can be done about this?

packaging/linux/run_keybase
```diff
@@ -187,7 +187,7 @@ check_for_url_scheme_launch() {
   # by checking argv.
 
   # If Keybase is already running, then pass it the link and exit.
-  if [[ $# -eq 1 && "$1" =~ ^web+.* ]]; then
+  if [[ $# -eq 1 && "$1" =~ ^(web\+|keybase).* ]]; then
```
taigrr: @eli-schwartz I believe this section is relevant in #18925 as well

shared/desktop/app/node.desktop.tsx
```diff
@@ -156,11 +159,11 @@ const createMainWindow = () => {
     if (startupURL) {
       // Mac calls open-url for a launch URL before redux is up, so we
       // stash a startupURL to be dispatched when we're ready for it.
-      sendToMainWindow('dispatchAction', {payload: {link: startupURL}, type: ConfigGen.link})
+      sendToMainWindow('dispatchAction', {payload: {link: startupURL}, type: DeeplinksGen.link})
       startupURL = null
     } else if (!isDarwin && process.argv.length > 1 && process.argv[1].startsWith('web+stellar:')) {
```
taigrr: @eli-schwartz And here! (referring to #18925 )

packaging/linux/run_keybase
heronhaye: I'm not sure if this is relevant to the Arch packages because they don't ship this script.

PR #18557, opened 2019-07-24
Fix BulkLookupContacts

zapu:


zapu:
no worries, i should have paid more attention, i thought your change only renamed variables, i missed the bug

PR #18559, opened 2019-07-24
non-restricted bot role

joshblum:
non restricted bots have access to the PTK and are treated close to the existing `READER` role

cc @keybase/hotpotatosquad


go/protocol/keybase1/extras.go
```diff
@@ -2324,10 +2330,18 @@ func (r TeamRole) IsReaderOrAbove() bool {
...
 
+func (r TeamRole) IsBotOrAbove() bool {
+	return r.IsOrAbove(TeamRole_BOT)
+}
+
 func (r TeamRole) IsRestrictedBotOrAbove() bool {
 	return r.IsOrAbove(TeamRole_RESTRICTEDBOT)
 }
 
+func (r TeamRole) IsBot() bool {
```
mlsteele: This name could be easily misunderstood as `r == BOT || r == RESTRICTEDBOT`. Might be less trouble to use `==` at the callsite.

(Or names `IsBotOfAnySort` `IsExactlyBot`)

go/protocol/keybase1/extras.go
```diff
@@ -2337,9 +2351,23 @@ func (r TeamRole) IsOrAbove(min TeamRole) bool {
...
+		default:
+			return false
+		}
 	default:
-		return int(r) >= int(min) || min == TeamRole_RESTRICTEDBOT
+		return (int(r) >= int(min) ||
+			min == TeamRole_BOT ||
+			min == TeamRole_RESTRICTEDBOT)
 	}
 }
```
mlsteele: This method is getting tricky. How about using very-private int values like this? https://goplay.space/#-7jmf09gPbX

go/teams/rpc_exim.go
```diff
@@ -45,6 +45,7 @@ func (t *Team) ExportToTeamPlusApplicationKeys(ctx context.Context, idTime keyba
 	writers = append(writers, members.Admins...)
 	writers = append(writers, members.Owners...)
 	onlyReaders = append(onlyReaders, members.Readers...)
+	onlyReaders = append(onlyReaders, members.Bots...)
```
mlsteele: onlyReaders -> onlyBots
which would then be unused

go/protocol/keybase1/extras.go
joshblum: Oh the only place we use it is where we want `BotLike` so i'll convert to that

go/protocol/keybase1/extras.go
joshblum: yeah much better

go/teams/rpc_exim.go
joshblum: this would exclude `Bots` from being able to read kbfs (only one that uses this afaik) which they should be able to

mlsteele: oh ok

go/libkb/chain_link_v2.go
```diff
@@ -157,7 +157,7 @@ func (t SigchainV2Type) RequiresAtLeastRole() keybase1.TeamRole {
 }
 
 func (t SigchainV2Type) TeamAllowStubWithAdminFlag(isAdmin bool) bool {
-	role := keybase1.TeamRole_READER
+	role := keybase1.TeamRole_BOT
```
mlsteele: There's something pre-existingly weird about these methods. I think `TeamAllowStub` is unused. If you'd like you can merge them and I think `role` goes away.

go/protocol/keybase1/extras.go
```diff
@@ -2324,25 +2330,48 @@ func (r TeamRole) IsReaderOrAbove() bool {
...
+	case TeamRole_BOT:
+		return 2
+	case TeamRole_READER,
+		TeamRole_WRITER,
+		TeamRole_ADMIN,
+		TeamRole_OWNER:
+		return int(r) + 2
 	default:
-		return int(r) >= int(min) || min == TeamRole_RESTRICTEDBOT
+		panic("unhandled teamRole")
```
mlsteele: woah there how about 0?

joshblum: oh i thought it would be better to crash since this is a sensitive check

mlsteele: Previously it didn't panic and just returned the wrong result. An error return path would be one way to go. I think the panic would do more harm than good imagining some ancillary system that using the comparison on untrusted input. Like some chat doohickey where someone sends a system message "Josh joined the team as a role 10".

joshblum: Heh yeah chat already uses this in a few places, good call. Will switch back to 0

PR #18561, opened 2019-07-24
libkb: don't use `poll=10` when loading upaks for chat, etc

maxtaco:
I'm trying to reduce the DB load of that polling query, and it bubbles all the way back to the client. It would be nice to disable server polling in more places, but a lot of tests depend upon it. Sigh.

### DIFF
--- a/go/engine/merkle_audit.go
+++ b/go/engine/merkle_audit.go
@@ -27,7 +27,7 @@ var (
 var MerkleAuditSettings = BackgroundTaskSettings{
 	Start:        5 * time.Minute,
 	StartStagger: 1 * time.Hour,
-	Interval:     6 * time.Hour,
+	Interval:     24 * time.Hour,
 	Limit:        1 * time.Minute,
 }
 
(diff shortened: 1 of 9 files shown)


go/engine/merkle_audit.go
```diff
@@ -27,7 +27,7 @@ var (
 var MerkleAuditSettings = BackgroundTaskSettings{
 	Start:        5 * time.Minute,
 	StartStagger: 1 * time.Hour,
-	Interval:     6 * time.Hour,
+	Interval:     24 * time.Hour,
```
heronhaye: Just wondering, do you know how this works on mobile? Is it 24 hours real time or amount of time in foreground?

maxtaco: Good Q, I don't know... @pzduniak ?

maxtaco: Seems to only run on wifi

heronhaye: I think josh did that for the box auditor too a while back.

pzduniak: >do you know how this works on mobile?
there should be an early return if it detects its not running on desktop, unless someone changed it

heronhaye: ok I asked miles, he says timer is real time

PR #18566, opened 2019-07-24
[Y2K-334] redo recommendations once contacts are loaded

jzila:
cc @keybase/y2ksquad

### DIFF
--- a/shared/team-building/index.tsx
+++ b/shared/team-building/index.tsx
@@ -86,7 +86,7 @@ const ContactsBanner = ({onRedoSearch, onRedoRecs}: {onRedoSearch: () => void; o
   const onLater = React.useCallback(() => dispatch(SettingsGen.createImportContactsLater()), [dispatch])
   // Redo search if # of imported contacts changes
   React.useEffect(() => {
-    if (prevNumContactsImported !== numContactsImported) {
+    if (prevNumContactsImported !== undefined && prevNumContactsImported !== numContactsImported) {
       onRedoSearch()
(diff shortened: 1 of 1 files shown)


shared/team-building/index.tsx
```diff
@@ -86,10 +86,11 @@ const ContactsBanner = ({onRedoSearch}: {onRedoSearch: () => void}) => {
   const onLater = React.useCallback(() => dispatch(SettingsGen.createImportContactsLater()), [dispatch])
   // Redo search if # of imported contacts changes
   React.useEffect(() => {
-    if (prevNumContactsImported !== undefined && prevNumContactsImported !== numContactsImported) {
+    if (prevNumContactsImported !== numContactsImported) {
```
buoyad: We should keep `prevNumContactsImported !== undefined &&`, it avoids a false positive on mount.

jzila: Doesn't it always start as undefined until we import the contacts? I'll put it back in and test, I guess.

buoyad: On mount `numContactsImported === null` and `prevNumContactsImported === undefined`
https://github.com/keybase/client/blob/1227eeb259d67a5bd35ceef507be69370ca60d6b/shared/constants/settings.tsx#L90-L95

jakob223: 🤯

jzila: @buoyad got it, thanks.

PR #18568, opened 2019-07-25
libkb: raise timeout on UPAK loads from 10minutes to 1 hour

maxtaco:
- this should have a big impact on server load
- this should have a small impact on mobile performance, but a positive one
- it's slightly worse for security, since it'll take longer to realize that a key has been revoked (up to an hour)


maxtaco:
@mmaxim you expressed second-thoughts to me in chat. should we kill this PR?

mmaxim:
Yeah maybe not necessary given how I think we will end up dealing with this in chat.

maxtaco:
OK, not going to do this.

PR #18570, opened 2019-07-25
use the new FS subscription stuff in GUI

songgao:
This gets rid of the old refreshTag stuff and uses the new subscription stuff for various Fs data. Subscribe and unsubscribe happen along with component mounting state through hooks, so we won't trigger unnecessary load when user's not on the corresponding screens. This should also be a lot simpler to maintain and hopefully significantly less buggy.


shared/actions/fs/index.tsx
```diff
@@ -1147,6 +1025,54 @@ const setTlfsAsUnloadedWhenKbfsDaemonDisconnects = state =>
 const setDebugLevel = (_, action: FsGen.SetDebugLevelPayload) =>
   RPCTypes.SimpleFSSimpleFSSetDebugLevelRpcPromise({level: action.payload.level})
 
+const subscriptionDeduplicateIntervalSecond = 1
+
+const subscribePath = (_, action: FsGen.SubscribePathPayload) =>
```
chrisnojima: lets add `TypedState` to all these so nothing is left untyped, even if unused

shared/actions/fs/index.tsx
```diff
@@ -1147,6 +1025,54 @@ const setTlfsAsUnloadedWhenKbfsDaemonDisconnects = state =>
...
+
+const onPathChange = (_, action: EngineGen.Keybase1NotifyFSFSSubscriptionNotifyPathPayload) => {
+  const {path, topic} = action.payload.params
+  switch (topic) {
+    case RPCTypes.PathSubscriptionTopic.children:
+      return FsGen.createFolderListLoad({path: Types.stringToPath(path)})
+    case RPCTypes.PathSubscriptionTopic.stat:
+      return FsGen.createLoadPathMetadata({path: Types.stringToPath(path)})
+    default:
+      Flow.ifFlowComplainsAboutThisFunctionYouHaventHandledAllCasesInASwitch(topic)
```
chrisnojima: i believe we no longer need this construct. due to the 'not all paths return a value' if you remove the default it should be fine and if you commented out one handler it'd complain

shared/fs/common/hooks.tsx
```diff
@@ -0,0 +1,66 @@
...
+import uuidv1 from 'uuid/v1'
+
+const isPathItem = (path: Types.Path) => Types.getPathLevel(path) > 2 || Constants.hasSpecialFileElement(path)
+
+const useDispatchWhenConnected = () => {
+  const kbfsDaemonConnected =
+    Container.useSelector(state => state.fs.kbfsDaemonStatus.rpcStatus) ===
+    Types.KbfsDaemonRpcStatus.Connected
+  const dispatch = Container.useDispatch()
+  return kbfsDaemonConnected ? dispatch : () => {}
```
chrisnojima: you may want to make this empty call a const and hoist it so it doesn't change which will trash the hooks that use it as a dep

chrisnojima: did you put some logging in and verify this all works as intended and is only firing when you expect it to?

shared/fs/common/path-item-action/header-container.tsx
```diff
@@ -11,20 +10,13 @@ type OwnProps = {
 type StateProps = {
   _pathItems: Types.PathItems
 }
-type DispatchProps = {
-  loadFolderList: () => void
-  loadPathMetadata: () => void
-}
+type DispatchProps = {}
```
chrisnojima: we shouldn't type the return values of mapState or mapDispatch. TS should derive it fine

shared/fs/common/hooks.tsx
songgao: I played around and tried a few things and they worked. Will inspect the log a bit to make sure we are not doing more than necessary

songgao: Actually found a bug! I was loading all Tlfs instead of synced ones. Fixed in latest commit with a load-only component since we can't branch hooks.

shared/fs/footer/upload.desktop.tsx
```diff
@@ -37,37 +36,34 @@ const realCSS = `
...
-          {!!(timeLeft && timeLeft.length) && (
-            <Kb.Text key="left" type="BodySmall" style={styles.stylesText}>{`${timeLeft} left`}</Kb.Text>
-          )}
-        </Kb.Box2>
-      </CSSTransition>
-    </React.Fragment>
-  )
-}
+const Upload = ({showing, files, fileName, totalSyncingBytes, timeLeft, debugToggleShow}: UploadProps) => (
+  <React.Fragment>
```
chrisnojima: use shorthand `<>`

go/kbfs/libkbfs/subscription_manager.go
```diff
@@ -37,6 +38,14 @@ func getCleanInTlfPath(p *parsedPath) cleanInTlfPath {
 	return cleanInTlfPath(path.Clean(p.rawInTlfPath))
 }
 
+func getParentPath(p cleanInTlfPath) (parent cleanInTlfPath, ok bool) {
```
strib: Wouldn't `path.Split()` do this same thing, and you can just check for a `""` parent instead of using a bool?

songgao: Thanks Strib! I was thinking finding last `/` would be slightly less wasteful than split-then-join. An explicit bool is just to make it ... explicit. I'm gonna merge this for now since once we have actual topic checks these will all likely change.

strib: @songgao 	`path.Split` is different from `strings.Split` -- it does exactly what you're done here, except it doesn't return a bool.  https://golang.org/src/path/path.go?s=3743:3785#L137

songgao: Ooh! I missed that. Thanks

songgao: And clever use of the `-1` returned value!

songgao: Actually `path.Split` is different. It includes the `/` at the end of parent dir. I guess we could do a `path.Clean`.

PR #18571, opened 2019-07-25
inbox swipe actions

mmaxim:
I had fun working on the swipe-to-reply feature, so I thought it would be cool to add something I thought would be useful, which is swipe left on inbox items to reveal actions. This patch adds it for "small team" rows (anything in time sorted list), and reveals "Mute" and "Hide" which are the same options available from long press. Here is a little demo of it in action (feel free to suggest different colors for the buttons):

demo


cecileboucheron:
Nice.
I don't know that we should use orange for the Mute/Unmute. Apart from that, lgtm!

adamjspooner:
Let's play with the colors later.

shared/chat/inbox/row/small-team/swipe-conv-actions/index.native.tsx
```diff
@@ -0,0 +1,113 @@
+import * as React from 'react'
+import * as Kb from '../../../../../common-adapters/mobile.native'
+import * as Styles from '../../../../../styles'
+import {Props} from '.'
+
+let curSwipeRef: React.MutableRefObject<Kb.Swipeable> | null = null
```
chrisnojima: i think we likely prefer `RefObject` as it captures current can be null, `MutableRefObject` doesn't

chrisnojima: why do we need to keep track of `curSwipeRef`?

shared/chat/inbox/row/small-team/swipe-conv-actions/index.native.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const onMute = (onMuteConversation: () => void, ref: React.MutableRefObject<Kb.Swipeable>) => {
+  ref.current && ref.current.close()
+  curSwipeRef = null
+  onMuteConversation()
+}
+
+const onHide = (onHideConversation: () => void, ref: React.MutableRefObject<Kb.Swipeable>) => {
```
chrisnojima: these two helpers can be one, just pass onHideConversation or onMuteConversation as param 1

shared/chat/inbox/row/small-team/swipe-conv-actions/index.desktop.tsx
```diff
@@ -0,0 +1,5 @@
+import * as React from 'react'
+const SwipeConvActions = props => {
+  return <div>{props.children}</div>
```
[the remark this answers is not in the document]
mmaxim: Maybe, I'll try.

shared/chat/inbox/row/small-team/swipe-conv-actions/index.native.tsx
mmaxim: I just used that type since it is what `useRef` returns, I can try changing it.

mmaxim: In order to close an open swipe on a row if we swipe another row.

shared/chat/inbox/row/small-team/swipe-conv-actions/index.native.tsx
mmaxim: Sure

shared/chat/inbox/row/small-team/swipe-conv-actions/index.native.tsx
```diff
@@ -0,0 +1,107 @@
+import * as React from 'react'
+import * as Kb from '../../../../../common-adapters/mobile.native'
+import * as Styles from '../../../../../styles'
+import {Props} from '.'
+
+let curSwipeRef: React.RefObject<Kb.Swipeable> | null = null
```
chrisnojima: i'd change this to 
```
let curSwipeRef = React.createRef<Kb.Swipeable>()
```

then when you want to null it just `curSwipRef.current = null`

chrisnojima: then `curSwipRef` itself is never null

shared/chat/inbox/row/small-team/swipe-conv-actions/index.native.tsx
chrisnojima: it actually has multiple types. you can merge types in TS: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L939

PR #18572, opened 2019-07-25
Make keybase:// links clickable inside Chat, handled directly in-app

cjb:

This commit pairs up with keybase/xurls#1 to make `keybase://` links clickable inside the app, and handled directly in-app without going through the web browser.  (If you want to test it, you'll need a service with that xurls change too.)

The link decoration component wasn't connected, but I needed to use dispatch to perform the link handling, so I went with creating a new function/hooks component for Keybase links only.


shared/common-adapters/markdown/service-decoration.tsx
```diff
@@ -11,6 +14,29 @@ import MaybeMention from '../../chat/conversation/maybe-mention'
...
+
+const KeybaseLink = (props: KeybaseLinkProps) => {
+  const dispatch = Container.useDispatch()
+  const onClick = React.useCallback(() => dispatch(DeeplinksGen.createLink({link: props.link})), [dispatch])
+
+  return (
+    <Text
+      className="hover-underline"
+      type="BodyPrimaryLink"
+      style={Styles.collapseStyles([props.styles.wrapStyle, linkStyle, props.styleOverride.link])}
```
nathunsmitty: Since we're only reading `wrapStyle` from `styles`, would it make more sense to just accept `wrapStyle` as a prop? Or would it make sense to define the `StyleOverride` values in `styles`? It seems a little weird to have two props that do the same thing here.

cjb: @nathunsmitty Thanks, perhaps like this? b9a70bb1a9

shared/common-adapters/markdown/service-decoration.tsx
```diff
@@ -11,6 +14,29 @@ import MaybeMention from '../../chat/conversation/maybe-mention'
 import Text, {StylesTextCrossPlatform} from '../text'
 import {StyleOverride} from '.'
 
+type KeybaseLinkProps = {
+  link: string
+  linkStyle: StylesTextCrossPlatform | undefined
```
nathunsmitty: maybe make these optional with a `?`? or do we want do be explicit when we're not passing styles?

PR #18575, opened 2019-07-25
Persist staticConfig for chat and wallet on logout

nathunsmitty:
Fixes an issue where constant data such as the maximum length of a payment message would be reset on logout.

props to @patrickxb for noticing this bug

### DIFF
--- a/shared/reducers/chat2.tsx
+++ b/shared/reducers/chat2.tsx
@@ -411,7 +411,9 @@ const rootReducer = (
 ): Types.State => {
   switch (action.type) {
     case Chat2Gen.resetStore:
-      return initialState
+      return initialState.merge({
+        staticConfig: state.staticConfig,
+      })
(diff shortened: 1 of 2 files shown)


nathunsmitty:
@keybase/react-hackers 👀 ?

chrisnojima:
why isn't this fetched on bootstrap again

nathunsmitty:
@chrisnojima the constants in `staticConfig` don't change for users – it's data like what commands can a user select in chat or how the max length of stellar payments, basically just mappings of constants in Go to the frontend. It seemed better to just fetch those once on startup and use them throughout the life of the app as opposed to refetching the same values on each new login.

chrisnojima:
so nothing could ever change given a client version (zero server interaction)

nathunsmitty:
@chrisnojima exactly

PR #18576, opened 2019-07-25
limit background loader jobs on sync on mobile

mmaxim:
* Limit background loading on mobile to max number of convs, and only the first page. If in the foreground, we really limit the loading to just the top convs.
* Sort the convs based on time, and also deprioritize big team convs.

### DIFF
--- a/go/chat/sync.go
+++ b/go/chat/sync.go
@@ -288,6 +288,10 @@ func (s *Syncer) getShouldUnboxSyncConvMap(ctx context.Context, convs []chat1.Co
 }
 
 func (s *Syncer) shouldUnboxSyncConv(conv chat1.Conversation) bool {
+	// only chat on mobile
+	if s.G().IsMobileAppType() && conv.GetTopicType() != chat1.TopicType_CHAT {
+		return false
+	}
 	// Skips convs we don't care for.
 	switch conv.Metadata.Status {
 	case chat1.ConversationStatus_BLOCKED,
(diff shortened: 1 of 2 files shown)


go/chat/sync.go
```diff
@@ -458,11 +465,34 @@ func (s *Syncer) sync(ctx context.Context, cli chat1.RemoteInterface, uid gregor
...
-			maxConvs = 3
+		pageBack := 3
+		num := 50
+		netState := s.G().MobileNetState.State()
+		state := s.G().MobileAppState.State()
+		if s.G().IsMobileAppType() {
+			maxConvs = s.maxConvLoads
+			num = 30
+			pageBack = 0
+			if netState.IsLimited() || state == keybase1.MobileAppState_FOREGROUND {
```
joshblum: how do we know we just came into the foreground here?

go/chat/sync.go
```diff
@@ -491,14 +521,18 @@ func (s *Syncer) sync(ctx context.Context, cli chat1.RemoteInterface, uid gregor
 				}
 				queuedConvs++
 			} else {
-				// If we are a limited data connection only queue selected
-				// convs up to maxConvs
-				if state.IsLimited() && (queuedConvs >= maxConvs || !s.shouldUnboxSyncConv(conv)) {
+				// Only do this for chat convs
+				if conv.GetTopicType() != chat1.TopicType_CHAT {
```
joshblum: why not move this into `shouldUnboxSyncConv`?

mmaxim: Need KBFS convs...

go/chat/sync.go
mmaxim: Just the foreground check is good enough

go/chat/sync.go
mmaxim: Maybe we can ignore them on mobile.

PR #18577, opened 2019-07-25
resolve both sides of name check before throwing an error

mmaxim:
Attempt to solve SBS assertion resolution race.


mmaxim:
cc @patrickxb this probably would have saved you, won't help you now though.

PR #18580, opened 2019-07-25
Refresh airdrop details/state on new login

cjb:
@keybase/picnicsquad 

In this flow:

* app starts up logged out
* you sign up as a new user

The airdropDetails and airdropState RPCs will both fail during startup (they require a logged in user), but they won't be retried after the login happens.  The result is that a new user who should be seeing the airdrop banner doesn't see it until they restart their app.  This PR fixes it by hooking those RPCs up to ConfigGen.loggedIn.

Also, since those RPCs require login, there's no point making them if we know we aren't logged in.

### DIFF
--- a/shared/actions/wallets.tsx
+++ b/shared/actions/wallets.tsx
@@ -1122,10 +1122,11 @@ type AirdropDetailsJSONType = {
 } | null
 
 const updateAirdropDetails = (
-  _: TypedState,
+  state: TypedState,
   __: WalletsGen.UpdateAirdropStatePayload | ConfigGen.DaemonHandshakeDonePayload,
   logger: Saga.SagaLogger
 ) =>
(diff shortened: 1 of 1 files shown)


shared/actions/wallets.tsx
```diff
@@ -1122,10 +1122,11 @@ type AirdropDetailsJSONType = {
 } | null
 
 const updateAirdropDetails = (
-  _: TypedState,
+  state: TypedState,
   __: WalletsGen.UpdateAirdropStatePayload | ConfigGen.DaemonHandshakeDonePayload,
```
nathunsmitty: nit: `__` could become `_`

cjb: Thanks!  I wonder if there's an eslint rule for validating uses of underscore placeholders :)

PR #18581, opened 2019-07-25
team.bot_settings link

joshblum:
add team.bot_settings link
depends on https://github.com/keybase/keybase/pull/4101 and https://github.com/keybase/proofs/pull/146

- [x]  check an old client loading a team with the new link. Should error out so we'll have to have a few releases with reader support before we have these links in the wild

cc @keybase/hotpotatosquad


protocol/avdl/keybase1/teams.avdl
```diff
@@ -646,6 +646,9 @@ protocol teams {
     boolean open;
     TeamRole openTeamJoinAs;
 
+    // restricted-bot configurations
+    map<UserVersion, TeamBotSettings> bots;
```
mlsteele: Where `B` is this commit and `A` is the previous commit. Remember to handle a client running `A` that loads a team made after `B`. And then upgrades to `B` without doing db nuke. Might need to bust cache or backfill.

go/libkb/chain_link_v2.go
```diff
@@ -57,6 +57,7 @@ const (
 	// Note that 45 is skipped, since it's retired; used to be LegacyTLFUpgrade
 	SigchainV2TypeTeamSettings     SigchainV2Type = 46
 	SigchainV2TypeTeamKBFSSettings SigchainV2Type = 47
+	SigchainV2TypeTeamBotSettings  SigchainV2Type = 48
```
mlsteele: `IsSupportedTeamType`

maybe `TeamAllowStub`. I'm surprised that other settingsish types aren't in there.

go/teams/chain.go
```diff
@@ -1716,6 +1728,23 @@ func (t *teamSigchainPlayer) addInnerLink(mctx libkb.MetaContext,
 		if err != nil {
 			return res, err
 		}
+	case libkb.LinkTypeTeamBotSettings:
+		if err = enforce(LinkRules{
+			Admin:               TristateRequire,
```
mlsteele: Admin should be optional. It's an optional reason that the poster is an admin that only exists because of implicit admins. Though it's probably present in all links, that's just laziness.

go/teams/chain.go
```diff
@@ -2248,6 +2277,48 @@ func (t *teamSigchainPlayer) parseTeamSettings(settings *SCTeamSettings, newStat
 	return nil
 }
 
+func (t *teamSigchainPlayer) parseTeamBotSettings(bots *[]SCTeamBot, newState *TeamSigChainState) error {
```
mlsteele: dereferencing `bots` at the callsite would be a bit more normal. Btw afaik the slice struct (not the data) is tiny

go/teams/chain.go
```diff
@@ -2248,6 +2277,48 @@ func (t *teamSigchainPlayer) parseTeamSettings(settings *SCTeamSettings, newStat
...
+		newState.inner.Bots[bot.Bot.ToUserVersion()] = keybase1.TeamBotSettings{
+			Cmds:     bot.Cmds,
+			Mentions: bot.Mentions,
+			Joins:    bot.Joins,
+			Triggers: triggers,
+			Convs:    convs,
+		}
+	}
+	return nil
+}
```
mlsteele: This checker looks correct. However, it might be easier to update this feature (bot settings) in the future if it wasn't picky here. And errored on reading bot settings rather than loading the team. Up to you

protocol/avdl/keybase1/teams.avdl
joshblum: TODO: check an old client loading a team with the new link. Should error out so we'll have to have a few releases with reader support before we have these links in the wild

joshblum:
old clients error on the new link type:

```
$ kb script teambot
Running `go install github.com/keybase/client/go/keybase`... Done.
▶ WARNING not exportable error: unsupported team link type: 48 (team.bot_settings) (teams.UnsupportedLinkTypeError)
▶ ERROR unsupported team link type: 48 (team.bot_settings)
```

joshblum:
merging over kbfs flake

PR #18583, opened 2019-07-25
fix bug in parsing invalid source account of tx operations in sep7 links

xgess:
sep7 links for transactions (`web+stellar:tx?...`), if there is a source account in an operation, apparently it cannot be empty. this change should error when that happens. 

i'm having some trouble testing this. i added a little tool for building these xdr objects, but I can't even seem to create one that is invalid in this way.


patrickxb:
I think this check should be in stellarnet.  stellar_uri.go 

I believe that the `tx` test in stellar_uri_test.go is an empty source account tx.  Line 102.  If it is, it should now be an invalid URI.

I, too, was unable to create a tx with an empty source account, which is how I figured out that sep7 changed underneath us after developing support for it.

xgess:
pushed down to https://github.com/keybase/stellarnet/pull/69

PR #18584, opened 2019-07-25
teams: box auditor integration when load says we need rotation

maxtaco:
- (draft, needs tests)


maxtaco:
This was actually pretty simple, if you think it's doing the right thing w/r/t the box auditor

maxtaco:
Hmm, still have to do a little more plumbing if we get back the `NeedRotateError`

maxtaco:
@heronhaye this is now ready for review

go/teams/loader.go
```diff
@@ -909,9 +907,15 @@ func (l *TeamLoader) load2InnerLockedRetry(ctx context.Context, arg load2ArgT) (
 	}
 
 	if hd := hiddenPackage.ChainData(); hd != nil {
+		hd.NeedRotate = needHiddenRotate
 		load2res.hidden = hd
 	}
 
+	if !arg.fromBoxAudit && needHiddenRotate {
+		mctx.Info("Scheduling box audit since we needed a rotation upon load (and we're not being directly called from the box audit)")
+		go BoxAuditAfterBackoff(mctx, arg.teamID)
```
maxtaco: @heronhaye what happens if a bunch of these pile up. Will the BoxAuditor internally discard the later ones?

go/libkb/constants.go
```diff
@@ -712,6 +712,8 @@ const MinEphemeralKeyLifetime = MaxEphemeralContentLifetime + EphemeralKeyGenInt
 
 const MaxTeamMembersForPairwiseMAC = 100
 
+const TeamBackoffBeforeAuditOnNeedRotate = time.Minute
```
heronhaye: even with jitter it feels like a lot of people will race at same time in a medium sized team, is that ok?

go/teams/loader.go
```diff
@@ -909,9 +907,15 @@ func (l *TeamLoader) load2InnerLockedRetry(ctx context.Context, arg load2ArgT) (
 	}
 
 	if hd := hiddenPackage.ChainData(); hd != nil {
+		hd.NeedRotate = needHiddenRotate
 		load2res.hidden = hd
 	}
 
+	if !arg.fromBoxAudit && needHiddenRotate {
+		mctx.Info("Scheduling box audit since we needed a rotation upon load (and we're not being directly called from the box audit)")
```
heronhaye: Make this a `Debug`?

Also, as discussed this will recurse up the parent chain in the worst case causing quadratic audits. So say A.B.C.D.E, and all teams need hidden rotations. Box auditing E will load A, B, C, and D and then box audit them since this flag isn't passed in through `MapTeamAncestors`. Then the box audit for D will cause a box audit for A, B, C, and so on.

heronhaye: also box audits succeed without checking for readers. if that's not ok we could pass in an option to make that not ok. I'm guessing only writers will attempt to rotate team hidden though?

go/teams/rotate_hidden_test.go
```diff
@@ -184,3 +187,88 @@ func TestRotateHiddenImplicitAdmin(t *testing.T) {
...
+			break
+		}
+		g.Log.Debug("Didn't get an update; waiting %s more", wait)
+		time.Sleep(wait)
+		wait = wait * 2
+	}
+	require.True(t, found, "whether condition was satisfied after polling ended")
+}
+
+func TestHiddenNeedRotate(t *testing.T) {
```
heronhaye: worth it to add a  test to make sure it's queued properly in box auditor? can use "RetryNextBoxAudit" to consume one off the queue.

go/teams/box_audit.go
```diff
@@ -1149,3 +1166,43 @@ func randomKnownTeamID(mctx libkb.MetaContext) (teamID *keybase1.TeamID, err err
...
+	a.delayedSlotsMutex.Lock()
+	defer a.delayedSlotsMutex.Unlock()
+	delete(a.delayedSlots, teamID)
+}
+
+func (a *BoxAuditor) ScheduleDelayedBoxAuditTeam(mctx libkb.MetaContext, teamID keybase1.TeamID) {
+	defer mctx.Trace(fmt.Sprintf("BoxAuditor#ScheduleDelayedBoxAuditTeam(%s)", teamID), func() error { return nil })()
+
+	if !a.getDelayedSlotForTeam(teamID) {
+		mctx.Debug("not scheduling delayed audit, since one is already in progress")
```
heronhaye: only one team id can be in the queue at a time, but I agree this is needed to prevent several box audits for the same team happening.

go/teams/rotate_hidden_test.go
maxtaco: So you're saying after the test is over, there will only be my team stuck in the box auditor?

maxtaco: Couldn't get this to work!

maxtaco:
Actually passed CI twice in a row!

heronhaye:
>I think a random wait from 30s to 90s should give us good protection against all team members rushing off to rekey. Do you think I should turn it up? Or make it dynamic?
Dynamic based on the number of writers might be good. Maybe it's fine for now. But I think if there's a team with over a thousand writers (maybe a lot of them are online because they're at work), the box audit itself will be very slow so network conditions will make it so they're racing.

I saw the changes to the merkleroots in the teamchain, looks good. Did you get the box audit to work in the test?

What is the clockwork bug in CI btw?

maxtaco:
Oh, I see what you mean.  OK, I'll put in a follow on ticket for the racing, etc.

Yes, the boxaudit was clean in the test after that change!

There's just some bug in clockwork. If one goroutine is sleeping and the other is advancing, it just sometimes doesn't work, I don't know why. I saw it fail locally too. I've bashed my head against this before.  

You know what it could be, it could the the sleep and the advance need to get there in the right order, or else it's a deadlock.

PR #18585, opened 2019-07-25
fix logout delay

jakob223:
- reduce timeout on pre-logout call
 - add a spinner while logging out

Issue: Y2K-315

@zapu are you good to do both sides review or should I tag react hackers?
cc @keybase/y2ksquad

### DIFF
--- a/go/engine/login_test.go
+++ b/go/engine/login_test.go
@@ -97,6 +97,24 @@ func TestLoginAndSwitchWithoutLogout(t *testing.T) {
 	}
 }
 
+func TestLogoutWhileOffline(t *testing.T) {
+	// Create an account
+	tc := SetupEngineTest(t, "login")
+	defer tc.Cleanup()
+	CreateAndSignupFakeUser(tc, "first")
+
+	// set server uri to nonexistent ip so api calls will fail
+	prev := os.Getenv("KEYBASE_SERVER_URI")
+	os.Setenv("KEYBASE_SERVER_URI", "http://127.0.0.127:3333")
+	defer os.Setenv("KEYBASE_SERVER_URI", prev)
+	tc.G.ConfigureAPI()
+
+	// Try to logout while offline
+	ctx, _ := context.WithTimeout(context.Background(), time.Second)
+	err := tc.G.Logout(ctx)
+	require.NoError(t, err)
+}
+
 func TestLoginUsernameWhitespace(t *testing.T) {
 	tc := SetupEngineTest(t, "login")
 	defer tc.Cleanup()
--- a/go/engine/login_test.go
+++ b/go/engine/login_test.go
@@ -97,24 +97,6 @@ func TestLoginAndSwitchWithoutLogout(t *testing.T) {
 	}
 }
 
-func TestLogoutWhileOffline(t *testing.T) {
-	// Create an account
-	tc := SetupEngineTest(t, "login")
-	defer tc.Cleanup()
-	CreateAndSignupFakeUser(tc, "first")
-
-	// set server uri to nonexistent ip so api calls will fail
-	prev := os.Getenv("KEYBASE_SERVER_URI")
-	os.Setenv("KEYBASE_SERVER_URI", "http://127.0.0.127:3333")
-	defer os.Setenv("KEYBASE_SERVER_URI", prev)
-	tc.G.ConfigureAPI()
-
-	// Try to logout while offline
-	ctx, _ := context.WithTimeout(context.Background(), time.Second)
-	err := tc.G.Logout(ctx)
-	require.NoError(t, err)
-}
-
 func TestLoginUsernameWhitespace(t *testing.T) {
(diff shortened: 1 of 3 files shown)


jzila:
cc @patrickxb

go/libkb/user.go
```diff
@@ -941,7 +941,10 @@ func CanLogout(mctx MetaContext) (res keybase1.CanLogoutRes) {
 		return res
 	}
 
-	if err := CheckCurrentUIDDeviceID(mctx); err != nil {
+	// since this is on the critical path to logout,
+	// use a short timeout in case we're offline.
+	m, _ := mctx.WithTimeout(600 * time.Millisecond)
+	if err := CheckCurrentUIDDeviceID(m); err != nil {
```
zapu: I'm worried that if we are online, but 600ms is not enough for the API call to finish, we might not be able to success early on `case DeviceNotFoundError, ...:` and keep failing on `LoadHasRandomPw` if user really doesn't have a device or doesn't exist anymore. So on slow connection you will be in permanent state of timeouting on `CheckCurrentUIDDeviceID` and then failing on `LoadHasRandomPw`.

`libkb/api.go` already uses `ConnectivityMonitor.IsConnected` to reduce timeout to 5 seconds if we are not connected to gregor (so when we are potentially offline). Maybe 5 seconds is not bad here? Especially with the GUI spinner.

zapu: Sorry I didn't bring this up yesterday, I only realized this now. Let's also see what Patrick thinks.

go/engine/login_test.go
```diff
@@ -97,6 +97,24 @@ func TestLoginAndSwitchWithoutLogout(t *testing.T) {
...
+	CreateAndSignupFakeUser(tc, "first")
+
+	// set server uri to nonexistent ip so api calls will fail
+	prev := os.Getenv("KEYBASE_SERVER_URI")
+	os.Setenv("KEYBASE_SERVER_URI", "http://127.0.0.127:3333")
+	defer os.Setenv("KEYBASE_SERVER_URI", prev)
+	tc.G.ConfigureAPI()
+
+	// Try to logout while offline
+	ctx, _ := context.WithTimeout(context.Background(), time.Second)
```
maxtaco: This test is sure to flake on a slow CI. You really can't depend on the actually clock in CI.

go/libkb/user.go
maxtaco: I agree I don't think this is the right startegy. 600msec is not enough. I think a spinner is the right strategy.

shared/settings/logout/index.tsx
```diff
@@ -28,10 +29,16 @@ const HoverBox = Styles.isMobile
...
 class LogOut extends React.Component<Props, State> {
   state = {
+    loggingOut: false,
     password: '',
     showTyping: false,
   }
 
+  logOut = () => {
+    this.props.onLogout()
+    this.setState({loggingOut: true})
```
zapu: I'd love to take GUI review but I cannot judge whether we should do this instead of having this in redux state and setting it in reducers. cc @keybase/react-hackers please review

shared/settings/logout/index.tsx
```diff
@@ -103,35 +105,39 @@ class LogOut extends React.Component<Props, State> {
         }}
         onClose={this.props.onCancel}
       >
-        <Kb.Box2 direction="vertical" fullHeight={true} style={styles.container}>
-          {Styles.isMobile && (
-            <Kb.Text style={styles.headerText} type="Header">
-              Do you know your password?
+        {this.state.loggingOut ? (
```
buoyad: The entire modal swiching to this is jarring - could we replace the "Just sign out" text with the spinner?

patrickxb:
To refresh my memory, why is there a need for a network call to logout?

Don't you always know if the user is PW or NOPW?

heronhaye:
>Don't you always know if the user is PW or NOPW?
I think what we need is a ticket to move the value from leveldb to config.json, and make sure it's set properly after signup and KEX

maxtaco:
You can't always know, since while you were offline, the user could have changed this on another device.

maxtaco:
What about putting a flag that's always in the gregor state here?  Then, as long as you've recently synced with gregor, then you have the right answer.

patrickxb:
But you can't go from PW to NOPW, right?

So if the device thinks you are NOPW and you are offline, just say you can't logout.

maxtaco:
@patrickxb I don't think that's what we want, since you might be in an emergency situation where you have to.

patrickxb:
So if you are NOPW and offline, how are you supposed to logout?  Doesn't that mean your account is dead?

shared/settings/logout/index.tsx
jakob223: Seems reasonable.

jakob223:
Okay, I've reduced the scope of this PR to adding the spinner.
@buoyad I have implemented your suggested GUI changes.

PR #18587, opened 2019-07-25
Disable asset picking buttons while payment is built

nathunsmitty:
Disables changing the source or destination asset while a payment path is being constructed. Previously, a user could get the UI into a weird state where the built path was not the same as their currently selected assets if they selected different assets while the path was loading.

Another option I considered was to reset the built payment if it came back and the assets differed from the UI state, but I ultimately decided it'd be better UX to just not give the users the option to change the assets until their current path has loaded.


nathunsmitty:
@keybase/react-hackers

PR #18589, opened 2019-07-25
add recent TLFs (up to 10) to Files tab root

songgao:
image

image

### DIFF
--- a/shared/fs/browser/root.tsx
+++ b/shared/fs/browser/root.tsx
@@ -1,0 +1,107 @@
+import * as I from 'immutable'
+import * as React from 'react'
+import * as Types from '../../constants/types/fs'
+import * as Kb from '../../common-adapters'
+import * as Container from '../../util/container'
+import TlfType from './rows/tlf-type-container'
+import Tlf from './rows/tlf-container'
+import {WrapRow} from './rows/rows'
+
+type Props = {}
+
+type SectionListItem = {
+  name: string
+  tlfType: Types.TlfType
+}
+
+const rootRows = [
+  {
+    name: Types.TlfType.Private,
+    tlfType: Types.TlfType.Private,
+  },
+  {
+    name: Types.TlfType.Public,
+    tlfType: Types.TlfType.Public,
+  },
+  {
+    name: Types.TlfType.Team,
+    tlfType: Types.TlfType.Team,
+  },
+]
+
+const renderItem = ({item, section}) =>
+  section.key === 'section-top' ? (
+    <WrapRow>
+      <TlfType name={item.name} routePath={I.List()} />
+    </WrapRow>
+  ) : (
+    <WrapRow>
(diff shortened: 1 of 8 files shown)


shared/fs/browser/root.tsx
```diff
@@ -0,0 +1,107 @@
...
+  tlfMtime: number
+  tlfType: Types.TlfType
+}> =>
+  React.useMemo(
+    () =>
+      tlfs
+        .valueSeq()
+        .toArray()
+        .sort((tlf1, tlf2) => tlf2.tlfMtime - tlf1.tlfMtime)
+        .slice(0, n)
```
jzila: I'm worried about perf here for large numbers of TLFs. Can we retrieve this top10 list from Go, and pull only those 10 TLFs from the state? Alternatively, you could implement this as a heap operation instead of sort-then-slice, but that might not be worth the extra code, unless you made/included a heap library.

songgao: Thanks for the fast response! I'm not too worried about performance here. It's memo-ed, and won't recalculate unless the corresponding tlfList (i.e. private, public, or team) changes. We make sure of reusing the same immutable object in the reducer when nothing changes, so this really isn't called that much.

I like the idea of moving to Go. Though I think Saga is also a proper place to produce this list. This way we don't need to do a separate RPC that basically does the same thing as favoritesLoad. WDYT?

chrisnojima: We want go to do more of this work

jzila: @songgao the TLF list will change every time there's an edit of any file. And IIRC the RPC is cheap compared to the difference in JS/Go efficiency, especially as the list sizes grow. Anyway feel free to merge without the change, just preferably add a TODO to move to Go.

songgao: Hmm we already have sorting implemented in JS, and look at the mtime there. So it's not like GUI doesn't have to worry about that field otherwise.

Or perhaps the ordering should also be from Go so JS doesn't have to sort?

jzila: Yeah exactly. And if we use Go, we can use a heap sort that runs in `k*log(n)` for a slice of size `k` instead of doing `n*log(n)` on the whole TLF list every time we need to get it.

songgao: If we do the sorting in Go, it seems better to just do that n*log(n) for each tlfList since we need that ordering in GUI anyway, plus a separate list for recent TLFs. So the API could return something like `{tlfs, private, public, team, recentList}`, where `tlfs` is a map; `private`, `public`, `team`, and `recentList` are all arrays of TLF paths, sorted by mtime. Potentially we can also move A-Z into the same API as well.

jzila: Sure, that sounds fine to me, as long as we're only doing that sort once.

songgao: Cool, thanks for the thoughts! I'll make a ticket and reference this discussion.

PR #18590, opened 2019-07-25
[Y2K-361] phone number should be searchable by default

jzila:
Also refactor the state management so toggles always operate on what the user has seen, instead of allowing background races.

cc @keybase/y2ksquad

### DIFF
--- a/shared/actions/json/settings.json
+++ b/shared/actions/json/settings.json
@@ -48,7 +48,7 @@
     "editPhone": {
       "phone": "string",
       "delete?": "boolean",
-      "toggleSearchable?": "boolean",
+      "setSearchable?": "boolean",
       "verify?": "boolean"
     },
     "sentVerificationEmail": {
--- a/shared/actions/json/settings.json
+++ b/shared/actions/json/settings.json
@@ -49,7 +49,6 @@
       "phone": "string",
       "delete?": "boolean",
       "setSearchable?": "boolean",
-      "verify?": "boolean"
     },
     "sentVerificationEmail": {
       "email": "string"
--- a/shared/constants/settings.tsx
+++ b/shared/constants/settings.tsx
@@ -45,7 +45,7 @@ export const makeEmailRow = I.Record<Types._EmailRow>({
 export const makePhoneRow = I.Record<Types._PhoneRow>({
   displayNumber: '',
   e164: '',
-  searchable: true,
+  searchable: false,
   superseded: false,
   verified: false,
 })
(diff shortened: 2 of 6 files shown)


shared/constants/settings.tsx
```diff
@@ -45,7 +45,7 @@ export const makeEmailRow = I.Record<Types._EmailRow>({
 export const makePhoneRow = I.Record<Types._PhoneRow>({
   displayNumber: '',
   e164: '',
-  searchable: false,
+  searchable: true,
```
buoyad: I think this should stay `false`

shared/signup/phone-number/index.tsx
```diff
@@ -15,10 +15,9 @@ export type Props = {
 const EnterPhoneNumber = (props: Props) => {
   const [phoneNumber, onChangePhoneNumber] = React.useState('')
   const [valid, onChangeValidity] = React.useState(false)
-  // const [allowSearch, onChangeAllowSearch] = React.useState(false)
+  const [allowSearch, onChangeAllowSearch] = React.useState(true)
```
buoyad: Did we reverse the earlier decision to disable toggling this during signup?

shared/actions/json/settings.json
```diff
@@ -48,7 +48,7 @@
     "editPhone": {
       "phone": "string",
       "delete?": "boolean",
-      "toggleSearchable?": "boolean",
+      "setSearchable?": "boolean",
       "verify?": "boolean"
```
buoyad: should we get rid of `verify`? Looks like the saga doesn't use it anymore.

shared/signup/phone-number/index.tsx
jzila: I wasn't aware of that decision. I any case, the default during signup should be to allow search. If we don't show the checkbox, doesn't the `makePhoneRow.searchable` above need to stay true?

buoyad: The previous code was passing `true` to the callback in lieu of showing the checkbox

jzila: Ok I'll restore it

shared/actions/json/settings.json
jzila: How does verification work then?

jzila:
@buoyad please take another look

shared/actions/json/settings.json
buoyad: It uses `resendVerificationForPhoneNumber`

buoyad: And `addPhoneNumber` automatically sends a verification text

PR #18596, opened 2019-07-26
Contact deduplication during search and tests

zapu:
This PR changes how we do contact filtering. We used to filter and throw away contact entries during contact syncing (during `SaveContactList` RPC) by keeping only one processed contact (so one component from one contact) if multiple components resolve to the same user, preferring e-mail components. This would filter out other, non resolved components, from a contact that had one component resolved, as well as other contacts that had components resolving to the same user.

This was good for presentation for the people recommendation list, but not good for searching.

Contacts will no longer be filtered during `SaveContactList` RPC, apart from removing exact duplicates (so components being identical including values and labels). Logic that was previously filtering out contacts to make a good contact list was ported over to new RPC `GetContactsForUserRecommendations` that's now used by people recommendation list on the frontend. This logic was also improved to also skip `CurrentUID` contact, because people sometimes have their own phone number / email in contact list.

There is lesser amount of filtering being done in user search when contacts are included. Firstly, we don't include contact results resolving for user that's also ben returned from textsearchd search. We will also only present one contact resolving to same user if multiple contacts come up in contacts search. But because we do this during search and not during sync, the one contact being returned is the better match.

This fixes all bugs where one would have some contacts being deduplicated out and not only they wouldn't appear in recommendations, but also would not be searchable at all.

This PR also introduces testing framework for contact syncing and usersearch. See `service/contacts_test.go` and `service/usersearch_test.go`. Everything search related and contact resolution related is mocked out, so this is only for testing how search results or recommendations are combined and presented.


go/contacts/contacts.go
```diff
@@ -225,47 +121,105 @@ func ResolveContacts(mctx libkb.MetaContext, provider ContactsProvider, contacts
...
+	emails := make([]keybase1.EmailAddress, 0, len(emailSet))
+	for phone := range phoneSet {
+		phones = append(phones, phone)
+	}
+	for email := range emailSet {
+		emails = append(emails, email)
+	}
+	providerRes, err := provider.LookupAll(mctx, emails, phones, regionCode)
+	if err != nil {
+		return res, err
```
heronhaye: we want to quick return error here?

zapu: oh that's interesting. I wonder what we should do when someone tries to sync contacts while being offline

PR #18599, opened 2019-07-26
Add section headers to search recommendations + alphabet index on mobile

buoyad:
Known issues:
- Desktop suggestions no longer scroll if you keyboard nav to an item off screen (Y2K-364)
- Alphabet index can move vertically when keyboard appears
- [x] 0-9 section of index doesn't match design
- [ ] Contacts are not sorted within sections

I could only test this with real data on an iPhone X, if someone has a smaller-screened phone with lots of contacts that would be a useful test case.

fancy-list

cc @keybase/y2ksquad

### DIFF
--- a/shared/team-building/container.tsx
+++ b/shared/team-building/container.tsx
@@ -2,7 +2,7 @@ import logger from '../logger'
 import * as React from 'react'
 import * as I from 'immutable'
 import {debounce, trim} from 'lodash-es'
-import TeamBuilding, {RolePickerProps} from '.'
+import TeamBuilding, {RolePickerProps, SearchRecSection} from '.'
 import RolePickerHeaderAction from './role-picker-header-action'
 import * as WaitingConstants from '../constants/waiting'
 import * as ChatConstants from '../constants/chat2'
@@ -60,6 +60,7 @@ const deriveSearchResults = memoize(
   ) =>
     searchResults &&
     searchResults.map(info => ({
+      contact: !!info.contact,
       displayLabel: info.label || '',
       followingState: followStateHelperWithId(myUsername, followingState, info.serviceMap.keybase),
       inTeam: teamSoFar.some(u => u.id === info.id),
@@ -298,6 +299,59 @@ const deriveRolePickerArrowKeyFns = memoize(
   })
 )
 
+const alphabet = 'abcdefghijklmnopqrstuvwxyz'
+const isAlpha = (letter: string) => alphabet.includes(letter)
+const letterToAlphaIndex = (letter: string) => alphabet.indexOf(letter)
+
+// Returns array with 28 entries
+// 0 - "Recommendations" section
+// 1-26 - a-z sections
+// 27 - 0-9 section
+const sortAndSplitRecommendations = memoize(
+  (results: Unpacked<typeof deriveSearchResults>): Array<SearchRecSection> | null => {
+    if (!results) return results
+
+    const sections: Array<SearchRecSection> = [
+      {
+        data: [],
+        label: 'Recommendations',
+        shortcut: false,
+      },
+    ]
+    results.forEach(rec => {
+      if (!rec.contact) {
+        sections[0].data.push(rec)
+        return
+      }
+      const letter = (rec.prettyName || rec.displayLabel || '')[0].toLowerCase()
+      if (isAlpha(letter)) {
+        // offset 1 to skip recommendations
+        const index = letterToAlphaIndex(letter) + 1
+        if (!sections[index]) {
+          sections[index] = {
+            data: [],
+            label: letter,
+            shortcut: true,
+          }
+        }
+        sections[index].data.push(rec)
+      } else if (!letter) {
+        // skip
+      } else {
+        if (!sections[27]) {
+          sections[27] = {
+            data: [],
+            label: '0-9',
+            shortcut: true,
+          }
+        }
+        sections[27].data.push(rec)
+      }
+    })
+    return sections
+  }
+)
+
 const mergeProps = (
   stateProps: ReturnType<typeof mapStateToProps>,
   dispatchProps: ReturnType<typeof mapDispatchToProps>,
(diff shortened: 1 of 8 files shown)


shared/team-building/container.tsx
```diff
@@ -282,6 +283,59 @@ const deriveRolePickerArrowKeyFns = memoize(
   })
 )
 
+const alphabet = 'abcdefghijklmnopqrstuvwxyz'
+const isAlpha = (letter: string) => alphabet.includes(letter)
```
jzila:
```suggestion
const alphaset = new Set(alphabet)
const isAlpha = (letter: string) => alphaset.has(letter)
```

shared/team-building/container.tsx
```diff
@@ -282,6 +283,59 @@ const deriveRolePickerArrowKeyFns = memoize(
...
+const letterToAlphaIndex = (letter: string) => alphabet.indexOf(letter)
```
jzila:
```suggestion
const letterToAlphaIndex = (letter: string) => letter.charCodeAt(0) - alphabet[0].charCodeAt(0)
```

shared/team-building/container.tsx
```diff
@@ -282,6 +283,59 @@ const deriveRolePickerArrowKeyFns = memoize(
...
+
+// Returns array with 28 entries
+// 0 - "Recommendations" section
+// 1-26 - a-z sections
+// 27 - 0-9 section
+const sortAndSplitRecommendations = memoize(
+  (results: Unpacked<typeof deriveSearchResults>): Array<SearchRecSection> | null => {
+    if (!results) return results
```
jzila: I think this would be clearer if it explicitly returned null.
```suggestion
    if (!results) return null
```

shared/team-building/container.tsx
```diff
@@ -282,6 +283,59 @@ const deriveRolePickerArrowKeyFns = memoize(
...
+        label: 'Recommendations',
+        shortcut: false,
+      },
+    ]
+    results.forEach(rec => {
+      if (!rec.contact) {
+        sections[0].data.push(rec)
+        return
+      }
+      const letter = (rec.prettyName || rec.displayLabel || '')[0].toLowerCase()
```
jzila: `''[0]` is `undefined`, which will fail:
```
   console.log(''[0].toLowerCase());
TypeError: Cannot read property 'toLowerCase' of undefined
    at evalmachine.<anonymous>:1:19
    at Script.runInContext (vm.js:133:20)
    at Object.runInContext (vm.js:311:6)
    at evaluate (/run_dir/repl.js:133:14)
    at ReadStream.<anonymous> (/run_dir/repl.js:116:5)
    at ReadStream.emit (events.js:198:13)
    at addChunk (_stream_readable.js:288:12)
    at readableAddChunk (_stream_readable.js:269:11)
    at ReadStream.Readable.push (_stream_readable.js:224:10)
    at lazyFs.read (internal/fs/streams.js:181:12)
```

shared/team-building/index.tsx
```diff
@@ -132,15 +138,178 @@ const ContactsBanner = ({onRedoSearch, onRedoRecs}: {onRedoSearch: () => void; o
...
   componentDidMount = () => {
     this.props.fetchUserRecs()
   }
 
+  _alphabetIndex = () => {
+    return (
+      <Kb.Box2 direction="vertical" centerChildren={true} style={styles.alphabetIndex}>
+        {this.props.recommendations &&
+          this.props.recommendations.map(section =>
+            section.label && section.label.length <= 3 ? (
```
jzila: Why the `section.label.length` check? And how can it be empty? Above you do a `.filter(Boolean)` so empty ones shouldn't appear in the list at all.

shared/team-building/index.tsx
```diff
@@ -132,15 +138,178 @@ const ContactsBanner = ({onRedoSearch, onRedoRecs}: {onRedoSearch: () => void; o
...
+              <React.Fragment key={section.label}>
+                <Kb.Text type="BodyTiny" onClick={() => this._onScrollToSection(section.label)}>
```
jzila: Is this going to cause unnecessary rerenders?

shared/team-building/index.tsx
```diff
@@ -132,15 +138,178 @@ const ContactsBanner = ({onRedoSearch, onRedoRecs}: {onRedoSearch: () => void; o
...
+    }
+  }
+
+  _getRecLayout = (
+    sections: Array<SearchRecSection>,
+    indexInList: number
+  ): {index: number; length: number; offset: number} => {
+    let numSections = 0
+    let numData = 0
+    let length: 40 | 64 = 64
```
jzila: What's the meaning behind these magic numbers?

shared/team-building/index.tsx
```diff
@@ -132,15 +138,178 @@ const ContactsBanner = ({onRedoSearch, onRedoRecs}: {onRedoSearch: () => void; o
...
+            Enter a {serviceIdToLabel(this.props.selectedService)} username above.
+          </Kb.Text>
+          <Kb.Text center={true} type="BodySmall">
+            Start a Keybase chat with anyone on {serviceIdToLabel(this.props.selectedService)}, even if they
+            don’t have a Keybase account.
+          </Kb.Text>
+        </Kb.Box2>
+      )
+    }
+    if (this.props.showRecs && this.props.recommendations) {
```
jzila: What about for search? No section list for search results?

shared/team-building/container.tsx
buoyad: Oof, thanks

shared/team-building/index.tsx
buoyad: I updated the code to check `=== 1` and do the last section "0-9" separately. I'll remove the empty check

shared/team-building/index.tsx
buoyad: Yes, I could `React.memo` this class method out, but maybe we can leave that for a second pass.

shared/team-building/index.tsx
buoyad: Ok, made some changes, let me know if they still don't make sense.

shared/team-building/index.tsx
buoyad: No, I didn't see anything relevant in designs for sections on search results.

buoyad:
Thanks @jzila, addressed feedback PTAL

PR #18602, opened 2019-07-26
Improve generation calculation for device eks

joshblum:


heronhaye:
also the debug might be confusing if theres no error

joshblum:
ah yeah i'll kill that log, yeah -1 is the default generation returned

PR #18612, opened 2019-07-26
Better errors for IMP TOFU rate limits in the GUI

jakob223:
cc @keybase/y2ksquad 

Issue: Y2K-279


heronhaye:
Ok looks fine, just close out the ticket after. We'll make other ones better as they come up.

PR #18615, opened 2019-07-27
fix chat header overflow on mobile

songgao:
This is not exactly the same as the new design, but I thought it'd be a good temporary solution for long conversation names on mobile. At least this makes sure the "i" icon is usable.

I didn't end up doing the popup thing as I mentioned in the squad meeting, since the usernames are already clickable.

More specifically, for 1:1 convo, or big team, this keeps original font size, but turns overflown part into `...`. For small team that has more than 2 participants, use BodyTiny, and use `lineClamp=2`.

Some screenshots:

image
image
image
image
image
image

cc @cecileboucheron @adamjspooner @keybase/hotpotatosquad


adamjspooner:
I dunno. I think it looks pretty good.

PR #18618, opened 2019-07-29
phone-app / app link handling

pzduniak:
1) I had to add a new intent filter to test it on Android -> we will need to validate it! will be handled by the same ticket that implements phone-app
2) testing in the iOS Simulator was pretty much impossible, the logic did get triggered like every 10th or 20th time so I suppose it's a simulator thing (?)

### DIFF
--- a/shared/actions/config/index.tsx
+++ b/shared/actions/config/index.tsx
@@ -3,11 +3,11 @@ import {log} from '../../native/log/logui'
 import * as ConfigGen from '../config-gen'
 import * as GregorGen from '../gregor-gen'
 import * as Flow from '../../util/flow'
+import * as SettingsGen from '../settings-gen'
 import * as ChatGen from '../chat2-gen'
 import * as EngineGen from '../engine-gen-gen'
 import * as DevicesGen from '../devices-gen'
 import * as ProfileGen from '../profile-gen'
-import * as SettingsGen from '../settings-gen'
 import * as FsGen from '../fs-gen'
 import * as RPCTypes from '../../constants/types/rpc-gen'
 import * as Constants from '../../constants/config'
@@ -326,24 +326,8 @@ function* maybeDoneWithLogoutHandshake(state) {
 }
 
 let routeToInitialScreenOnce = false
-let loadMySettingsOnce = false
 
 const routeToInitialScreen2 = (state: Container.TypedState) => {
-  if (
-    state.config.loggedIn &&
-    state.config.startupLink &&
-    state.config.startupLink.endsWith('/phone-app') &&
-    !state.settings.phoneNumbers.phones
-  ) {
-    if (!loadMySettingsOnce) {
-      loadMySettingsOnce = true
-      return [SettingsGen.createLoadSettings()]
-    }
-
-    // pending loadMySettings finishing
-    return
-  }
-
   // bail if we don't have a navigator and loaded
   if (!Router2._getNavigator()) {
     return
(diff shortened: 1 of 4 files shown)


shared/actions/config/index.tsx
```diff
@@ -280,6 +281,14 @@ const switchRouteDef = (
           ? [RouteTreeGen.createNavigateAppend({path: ['signupEnterPhoneNumber']})]
           : []),
       ]
+    } else if (action.type === ConfigGen.loggedIn) {
+      if (
+        state.config.startupLink &&
+        state.config.startupLink.endsWith('/phone-app') &&
+        !state.settings.phoneNumbers.phones
+      ) {
+        return [SettingsGen.createLoadSettings()]
```
chrisnojima: this should only happen once and with the logic that already does this kind of stuff
https://github.com/keybase/client/blob/master/shared/actions/config/index.tsx#L399

shared/actions/config/index.tsx
```diff
@@ -327,6 +336,16 @@ function* maybeDoneWithLogoutHandshake(state) {
...
 
 const routeToInitialScreen2 = (state: Container.TypedState) => {
+  if (
+    state.config.loggedIn &&
+    state.config.startupLink &&
+    state.config.startupLink.endsWith('/phone-app') &&
+    !state.settings.phoneNumbers.phones
+  ) {
+    // pending loadMySettings finishing
+    return
```
chrisnojima: lets not do this. just go there and show a spinner

shared/actions/config/index.tsx
```diff
@@ -553,8 +584,10 @@ function* configSaga(): Saga.SagaGenerator<any, any> {
   // MUST go above routeToInitialScreen2 so we set the nav correctly
   yield* Saga.chainAction<ConfigGen.SetNavigatorPayload>(ConfigGen.setNavigator, setNavigator)
   // Go to the correct starting screen
-  yield* Saga.chainAction<ConfigGen.DaemonHandshakeDonePayload | ConfigGen.SetNavigatorPayload>(
-    [ConfigGen.daemonHandshakeDone, ConfigGen.setNavigator],
+  yield* Saga.chainAction<
+    ConfigGen.DaemonHandshakeDonePayload | ConfigGen.SetNavigatorPayload | SettingsGen.LoadedSettingsPayload
+  >(
+    [ConfigGen.daemonHandshakeDone, ConfigGen.setNavigator, SettingsGen.loadedSettings],
```
chrisnojima: don't change this

shared/android/app/src/main/AndroidManifest.xml
```diff
@@ -57,6 +57,13 @@
           <category android:name="android.intent.category.BROWSABLE" />
           <data android:scheme="keybase" />
         </intent-filter>
+        <intent-filter>
+          <action android:name="android.intent.action.VIEW" />
+          <category android:name="android.intent.category.DEFAULT" />
+          <category android:name="android.intent.category.BROWSABLE" />
+          <data android:scheme="http" android:host="keybase.io" android:pathPrefix="/" />
```
chrisnojima: we need to discuss

shared/constants/config.tsx
```diff
@@ -65,9 +65,9 @@ export const urlToUsername = (url: {
   }
 
   const pathname = url.pathname
-  // Adapted username regexp (see libkb/checkers.go) with a leading / and an
-  // optional trailing /.
-  const match = pathname.match(/^\/((?:[a-zA-Z0-9][a-zA-Z0-9_]?)+)\/?$/)
+  // Adapted username regexp (see libkb/checkers.go) with a leading /, an
+  // optional trailing / and a dash for custom links.
+  const match = pathname.match(/^\/((?:[a-zA-Z0-9][a-zA-Z0-9_-]?)+)\/?$/)
```
chrisnojima: i don't understand the meaning of this change

pzduniak: we'll have /phone-app that'll trigger the "add phone number" flow and /app for other uses (for example email)

shared/android/app/src/main/AndroidManifest.xml
pzduniak: I have literally no idea how we should approach this thing - we mostly just need a regexp so that this only happens on usernames _and_ I believe it should work exactly the same way as on iOS (which I haven't looked into, all I know is that it opens the app there). I can add the link verification logic on the server if we go forward with this change.

shared/actions/config/index.tsx
pzduniak: im not sure if i understand this, could you please clarify?

shared/actions/config/index.tsx
pzduniak: again im not sure what i should do here instead

shared/android/app/src/main/AndroidManifest.xml
pzduniak: I found that some people are using https://github.com/airbnb/DeepLinkDispatch for more complicated matching. I limited it to the 2 special routes instead.

pzduniak:
@chrisnojima ready for final review

PR #18622, opened 2019-07-29
format phone number in delete confirm

jakob223:
Issue: Y2K-327

cc @keybase/y2ksquad

### DIFF
--- a/shared/settings/account/confirm-delete.tsx
+++ b/shared/settings/account/confirm-delete.tsx
@@ -22,8 +22,10 @@ const getIcon = (props: Props) => {
 }
 const getPrompt = (props: Props) => {
   return props.type === 'email'
-    ? `Delete email ${props.address}`
-    : `Delete number ${PhoneUtil.e164ToDisplay(props.address)}`
+    ? `Delete email
+${props.address}`
+    : `Delete number
+${PhoneUtil.e164ToDisplay(props.address)}`
 }
 const ConfirmDeleteAddress = (props: Props) => (
   <Kb.ConfirmModal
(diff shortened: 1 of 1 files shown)


shared/settings/account/confirm-delete.tsx
```diff
@@ -19,11 +19,14 @@ const getIcon = (props: Props) => {
   }
   return Styles.isMobile ? 'icon-phone-number-remove-64' : 'icon-phone-number-remove-48'
 }
-
+const getPrompt = (props: Props) => {
+  const addr = props.type === 'email' ? props.address : PhoneUtil.e164ToDisplay(props.address)
```
chrisnojima: if you're going to split on the types you might as well just build 2 strings
```
return props.type === 'email'
  ? `Delete email
 ${props.address}`
  : `Delete number
${PhoneUtil.e164ToDisplay(props.address)}`
```

shared/settings/account/confirm-delete.tsx
```diff
@@ -20,8 +21,9 @@ const getIcon = (props: Props) => {
   return Styles.isMobile ? 'icon-phone-number-remove-64' : 'icon-phone-number-remove-48'
 }
 const getPrompt = (props: Props) => {
-  const addr = props.type === 'email' ? props.address : PhoneUtil.e164ToDisplay(props.address)
-  return `Delete ${props.type === 'email' ? 'address' : 'number'} \n${addr}?`
+  return props.type === 'email'
+    ? `Delete email ${props.address}`
```
chrisnojima: no newline?

jakob223: Oh, I'm ambivalent about it - I didn't notice that the newlines in your suggestion are in the strings. I feel like it might look odd with the newline with a short email address, but if you feel strongly I can add them.

chrisnojima: i don't care, it should match the design. just pointing out it mutated from the original

jakob223: (I see - fixed to match design)

PR #18623, opened 2019-07-29
Define Chat API return types in AVDL

nathunsmitty:
Currently, most of our return types from `keybase chat api` are defined directly in Go. This PR moves those definitions out into `protocol/avdl/chat1/local.avdl` and generates them. This will allow us to not only generate the types in Go, but also TypeScript, Python, and other languages.

This is a refactor PR. There should be no changes to external output, except for the `omitempty` note below.

A couple of notes: 
- By default, the chat API Go struct fields (and thus API output) are defined in camel case by the AVDL compiler. Previous versions of the API had these fields snake cased. Thus there are a lot of `@jsonkey`s in the AVDL. (This actually wasn't too bad to add – ask me about my vim macro!)
- A few variable names like `UID` and `ID` had to be changed to `uid` and `id` respectively due to `bin/flow.js` lint errors.
- If there was a name conflict in the protocol, I generally appended `Res` to the end of the added struct, to denote that it's a result value. I'm sure the naming could be improved, though, so I'm open to ideas.
- Currently, our AVDL compiler doesn't add `omitempty` to fields unless they're strings. This has the result of lots of extra fields with values like `''`, `null`, or `0` in the output of API calls. This doesn't break anything, but I'm not the biggest fan of the extra noise. I have a PR to parse an `@optional` decorator in the compiler which will fix this (keybase/node-avdl-compiler#28)

I tested this PR against the tests in https://github.com/keybase/keybase-bot, which I figured were probably the most comprehensive set of Keybase chat api tests we have. Additional useful tools were [jq](https://stedolan.github.io/jq/), [icdiff](https://www.jefftk.com/icdiff), and lots of unix streams/pipes.


nathunsmitty:
@keybase/picnicsquad this is ready for review

protocol/avdl/chat1/local.avdl
```diff
@@ -1137,4 +1137,267 @@ protocol local {
     array<RateLimit> rateLimits;
   }
   ClearBotCommandsLocalRes clearBotCommandsLocal();
+
+
+  record RateLimitRes {
```
xgess: this file is pretty big. any chance these it's not too hard to move all of these new types to a new file called something like `cli.avdl`? they're only used in the cli, right?

nathunsmitty: done! 👍

ddworken:
Out of curiosity, what is the ultimate plan for the python bot library when used with these definitions? It seems that the python library does not support avro code generation and instead references schemas as dicts:

https://avro.apache.org/docs/current/gettingstartedpython.html

So this really loses the benefits of python type hints that we might get if we did have python code generation. From a quick look at the avsc file type, it does look like it wouldn't be tooooo hard to write a python code generator for it. But just curious whether that is planned to be included in this work.

nathunsmitty:
@ddworken I was planning on writing a Python code generator. We use some custom extensions like variants to AVDL so a stock AVDL tool wouldn't work without some modification. Since we already construct an AVDL AST, I figured extending that to Python wouldn't be too difficult. (I like @xgess' current types that use dataclasses and [mashumaro](https://github.com/Fatal1ty/mashumaro), so it'll probably compile to that) I was gonna go a similar thing for TypeScript, too, since our TS compiler is very RPC specific.

ddworken:
Awesome! Excited to see it :tada:

protocol/Makefile
```diff
@@ -51,11 +51,11 @@ js/rpc.js: build-stamp | config
...
-	$(AVDLC) -b -l go -t -o ../../go-keybase-chat-bot/kbchat/types/keybase avdl/keybase1/*.avdl
-	$(AVDLC) -b -l go -t -o ../../go-keybase-chat-bot/kbchat/types/gregor avdl/gregor1/*.avdl
-	$(AVDLC) -b -l go -t -o ../../go-keybase-chat-bot/kbchat/types/chat avdl/chat1/*.avdl
-	$(AVDLC) -b -l go -t -o ../../go-keybase-chat-bot/kbchat/types/wallet avdl/stellar1/*.avdl
-	(cd ../../go-keybase-chat-bot/kbchat/types && go fmt ./...)
+	$(AVDLC) -b -l go -t -o ../../go-keybase-chat-bot/kbchat/types/keybase1 avdl/keybase1/*.avdl
+	$(AVDLC) -b -l go -t -o ../../go-keybase-chat-bot/kbchat/types/gregor1 avdl/gregor1/*.avdl
+	$(AVDLC) -b -l go -t -o ../../go-keybase-chat-bot/kbchat/types/chat1 avdl/chat1/*.avdl
+	$(AVDLC) -b -l go -t -o ../../go-keybase-chat-bot/kbchat/types/stellar1 avdl/stellar1/*.avdl
+	(cd ../../go-keybase-chat-bot/kbchat/types && goimports -w .)
```
nathunsmitty: This Makefile is what I used to compile the types in go-keybase-chat-bot#14. I had to make a couple of name tweaks to match the package names. The compiler also spits out imports relative to `github.com/keybase/client`, so I opted to use `goimports` to fix that issue until I update the compiler to support relative imports.

xgess:
this seems good to me. will there also be a make task in the bot libraries to compile and pull in new types?

nathunsmitty:
@xgess hmm, I hadn't thought of that! I was thinking it'd go the other way, update the types here, then run a make task in `protocol` that updates all of the types for each bot library.

I suppose it's kind of a tricky problem because we're spreading our definitions across repos. A potential alternative I've been thinking about is bringing the bot libs into the client monorepo in a `bots/` dir or something. open to any and all suggestions 🤷‍♂

PR #18624, opened 2019-07-29
team building: no contacts in teams tab

jakob223:
This removes contacts from the team builder when trying to add people to a team.

Issue: Y2K-363
CC @keybase/y2ksquad

### DIFF
--- a/shared/actions/team-building.tsx
+++ b/shared/actions/team-building.tsx
@@ -142,7 +142,7 @@ const search = (state: TypedState, {payload: {namespace}}: NSAction) => {
 
   const query = teamBuildingSearchQuery
   let impTofuQuery: RPCTypes.ImpTofuQuery | null = null
-  if (flags.sbsContacts && teamBuildingSelectedService === 'keybase') {
+  if (flags.sbsContacts && teamBuildingSelectedService === 'keybase' && namespace == 'chat2') {
     const userRegion = state.settings.contacts.userCountryCode
     impTofuQuery = makeImpTofuQuery(query, userRegion)
   }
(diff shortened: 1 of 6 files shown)


shared/actions/team-building.tsx
```diff
@@ -142,7 +142,7 @@ const search = (state: TypedState, {payload: {namespace}}: NSAction) => {
 
   const query = teamBuildingSearchQuery
   let impTofuQuery: RPCTypes.ImpTofuQuery | null = null
-  if (flags.sbsContacts && teamBuildingSelectedService === 'keybase') {
+  if (flags.sbsContacts && teamBuildingSelectedService === 'keybase' && namespace == 'chat2') {
```
chrisnojima: this file should be namespace agnostic, no chat2 specific stuff should be in here

jakob223: What do you think the best way to only get contacts on New Chat searches is? It seems like it would be bad to split this into two files, but I need to have some logic somewhere in here that acts different in one case vs the other.

shared/team-building/container.tsx
```diff
@@ -199,9 +201,10 @@ const mapDispatchToProps = (dispatch: Container.TypedDispatch, {namespace, teamn
     dispatch(SettingsGen.createRequestContactPermissions({thenToggleImportOn: true})),
   _search: (query: string, service: ServiceIdWithContact, limit?: number) => {
     const func = service === 'keybase' ? debouncedSearchKeybase : debouncedSearch
-    return func(dispatch, namespace, query, service, limit)
+    return func(dispatch, namespace, query, service, namespace === 'chat2', limit)
```
chrisnojima: this should also be a prop and not some built in thing

jakob223: As in, an OwnProp for the team builder? Why is this different than the code below that looks at `ownProps.namespace === 'teams'`?

chrisnojima: ok lets leave this as is and just ticket cleaning this up as it shouldn't be below either

PR #18627, opened 2019-07-29
user search ordering fixes

heronhaye:


go/service/usersearch.go
```diff
@@ -327,13 +327,14 @@ func (h *UserSearchHandler) UserSearch(ctx context.Context, arg keybase1.UserSea
 		if err != nil {
 			mctx.Warning("Failed to do contacts search: %s", err)
 		} else {
-			res = append(res, contactsRes...)
 			// Sort first - we are going to be deduplicating on usernames,
 			// entries with higher score have precedence.
-			sort.Slice(res, func(i, j int) bool {
```
zapu: sorting everything was intentional, do we want to start displaying contacts at the end of the list even if they are more relevant to the query?

heronhaye: hm. ok, how about we rewrite the scores of all the followees to be at least what the previous one was? then sort as normal

heronhaye: /shrug let's do this until we get a better design

zapu: sounds good, I misunderstood at first. thanks for explaining

zapu: it would probably be ideal to fix textsearchd but lets not worry about this right now

PR #18629, opened 2019-07-29
Better handling of sending non-native assets without an exchange

nathunsmitty:
out

Also updates the transaction details page to hide the conversion rate and payment path for these "trivial path" cases, since seeing a conversion rate of 1 = 1 isn't super helpful.

Note that removing the "by converting..." message in chat is unchanged; we need to plumb through some additional data in that RPC unless we want to remove that text completely, so that'd probably best be handled in a separate PR.

@keybase/react-hackers 
@keybase/picnicsquad


shared/wallets/send-form/calculate-advanced-button.tsx
```diff
@@ -42,7 +44,7 @@ const CalculateAdvancedButton = (props: CalculateAdvancedButtonProps) => {
     ) : (
       <Kb.Button
         type="Wallet"
-        label="Calculate"
+        label={hasTrivialPath ? 'Confirm details' : 'Calculate'}
```
nathunsmitty: Trivial path sends use the same path payment RPC as all other path payments, so we first need to calculate the path. However, since we don't need to display a path, exchange rate info, etc. I decided to fake needing to calculate it with a confirm button instead.

I like this approach because otherwise there'd be no confirmation step in the send flow. Open to any other ideas on how to do this, though.

PR #18636, opened 2019-07-29
Jenkinsfile: linting new code

strib:


strib:
I'm getting lots of unrelated flakes so haven't passed CI yet, but this seems to work. Earlier versions of this PR inserted some bad code into `libkb` and it was found and caused a failure.

@joshblum one weird side effect with the `-new-from-rev` flag is that, if there's new commits in `master` that haven't been rebased/merged into this PR yet, those commits are actually included in the diff that is checked.  So you could see weird false negatives where it finds a bug in code that's actually been changed or deleted in master.  But hopefully that's not a big deal, since once this is in, no more buggy code should get into master, I hope?

PR #18637, opened 2019-07-29
track online status through subscription manager and don't load when offline

songgao:


songgao:
thanks both! Will wait until after release

PR #18639, opened 2019-07-30
pre-commit: update for golangci-lint hook

strib:


strib:
Merged this to just take away that hook for now.

PR #18649, opened 2019-07-30
Add some padding to the “Out of date” banner.

adamjspooner:


shared/common-adapters/banner.tsx
```diff
@@ -188,6 +195,11 @@ const styles = Styles.styleSheetCreate({
       textDecorationLine: 'underline',
     },
   }),
+  wideTextContainer: {
```
jakob223: this isn't used anywhere, is it?

adamjspooner: Good catch!

PR #18653, opened 2019-07-30
Simplify conv localizer code for contacts

zapu:
No need to parse the assertion there. Every processed contact will have assertion field that's generated during syncing and it has to match chat assertions because they are used to create these SBS chat convos.

Pass participant list as argument and modify participants in place rather than returning new list.

Fix `contactsFetched` not being set to `true` in the loop, causing contacts to be reloaded for every participant in the list.

Another change - the subject of the ticket - is not to localize participant name if more than one contact matches an assertion. Because we are not saving which contact exactly was used to make the conversation, we can't risk picking the wrong one and confusing the user.

### DIFF
--- a/go/chat/utils/utils.go
+++ b/go/chat/utils/utils.go
@@ -213,8 +213,8 @@ func AttachContactNames(mctx libkb.MetaContext, participants []chat1.Conversatio
 				contactsFetched = true
 			}
 			participant.ContactName = findContactName(contacts, participant.Username)
+			participants[i] = participant
 		}
-		participants[i] = participant
 	}
 }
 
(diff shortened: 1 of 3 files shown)


zapu:
cc @buoyad

go/chat/utils/utils.go
```diff
@@ -190,47 +190,48 @@ func splitAndNormalizeTLFNameCanonicalize(mctx libkb.MetaContext, name string, p
...
-		withContacts = append(withContacts, participant)
+		participants[i] = participant
 	}
-	return withContacts
 }
 
-func findContactName(contacts []keybase1.ProcessedContact, phoneOrEmail string, isPhone bool) *string {
+func findContactName(contacts []keybase1.ProcessedContact, assertion string) *string {
+	var result *string
 	for _, contact := range contacts {
```
heronhaye: i assume the speed of this isn't an issue but maybe add a comment that this is technically n*m

zapu: at some point we might want to save `chatHeaders map[assertion]contact` when syncing contacts to simplify this function

heronhaye: might have to be map[assertion][]contact?

PR #18659, opened 2019-07-30
format phone numbers in search

jzila:
cc @keybase/y2ksquad

Issue: Y2K-388

[w=1 is your friend](https://github.com/keybase/client/pull/18659/files?w=1)


zapu:
This might not work for cases where we append contact component label to `DisplayLabel` - like "(phone)" or "(work)" when contact has more than one component.

https://github.com/keybase/client/blob/master/go/contacts/contacts.go#L193

for simplicity we can remove this for now

jzila:
@zapu I know but it's better than nothing for now.

PR #18660, opened 2019-07-30
make lag radar off by default. connect to flag. make togglable. reduce frames

chrisnojima:
@keybase/react-hackers turns off lag radar by default. you can click to turn on or add a feature flag (or override) to have it on by default
cc: @strib

### DIFF
--- a/shared/app/runtime-stats/index.tsx
+++ b/shared/app/runtime-stats/index.tsx
@@ -4,6 +4,7 @@ import * as Styles from '../../styles'
 import {isIPhoneX} from '../../constants/platform'
 import * as RPCTypes from '../../constants/types/rpc-gen'
 import lagRadar from 'lag-radar'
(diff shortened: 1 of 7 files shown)


shared/app/runtime-stats/index.tsx
```diff
@@ -74,26 +75,42 @@ const dbTypeString = (s: RPCTypes.DbType) => {
...
+      radarNode = node
+      makeRadar(showRadar)
+    },
+    [showRadar]
+  )
+  const toggleRadar = () => {
+    const show = !showRadar
+    setShowRadar(show)
+    makeRadar(show)
+  }
```
songgao: Any reason why this wasn't memo'ed? i.e. React.useCallback here too

chrisnojima: i don't think we should memo everything. this takes 3ms to render every 5 seconds

PR #18662, opened 2019-07-31
share into kbfs from ios

jzila:


jzila:
@mmaxim FYI it didn't work. If the browsing itself doesn't OOM-kill the extension, then actually sending the image does. The baseline RAM usage is just too high. Sometimes it starts at 106MB, where the limit is 120MB.

PR #18663, opened 2019-07-31
SEP6 Deposit/Withdraw: GUI integration

cjb:
r? @patrickxb @keybase/picnicsquad 

* When we're receiving an asset in GetAccountAssetsLocal, save the new `show{Deposit,Withdraw}Button` and `{deposit,withdraw}ButtonText` fields into Redux
* When we're displaying an asset on the main wallet page, add the Deposit and Withdraw links if they're present for the asset
* When we're clicking on one of those links, call `localAsset{Deposit,Withdraw}LocalRpc` with the asset and the current accountID as arguments
* When we get a result from that RPC, open the `externalURL` field in the RPC result in a web browser if it exists

Screen Shot 2019-07-31 at 2 21 46 AM


patrickxb:
I believe design asked for `|` between the links.  Is that right @adamjspooner?

patrickxb:
So the spec says that just calling the `deposit` or `withdraw` endpoints on the anchor could succeed.  I don't think this is possible in the minimal sep6 that we and the anchors are implementing, but the protocol has this:

  record AssetActionResultLocal {
    union { null, string } externalUrl;
    union { null, string } messageFromAnchor;
  }
  AssetActionResultLocal assetDepositLocal(AccountID accountID, Asset asset);
  AssetActionResultLocal assetWithdrawLocal(AccountID accountID, Asset asset);

One of externalUrl and messageFromAnchor will exist.  If it is messageFromAnchor, can we just display it to the user with an "Ok" button?

Again, I don't think this will ever happen, but just in case.

patrickxb:
If assetDepositLocal or assetWithdrawLocal return an error, is it shown to the user in a dialog window?

adamjspooner:
That's correct (regarding pipes).

cjb:
@patrickxb Added the pipe separators, showing deposit/withdraw errors in a dialog on a red background, showing `messageFromVendor`s in a dialog on a green background, think that should be everything.

PR #18665, opened 2019-07-31
Fix highlighting, scrolling, and selecting in "new chat" section list

zapu:
To highlight selected row properly in "new chat" section list, we need to do some calculations to be able to go from "selected index" to "section, index within section". For scrolling (on desktop), we need to be able to go from "selected index" to "index in flat list", where flat list contains rows, sections, and placeholders.

Pressing "enter" key also did not work correctly because to determine which item was highlighted it was using the list from before transforming it into sections. So I made it keep an additional list of flattened sections, that keeps the order within sections but all items are in one list for easy access.

### DIFF
--- a/shared/team-building/index.tsx
+++ b/shared/team-building/index.tsx
@@ -268,18 +268,18 @@ class TeamBuilding extends React.PureComponent<Props, {}> {
     (
       highlightedIndex: number | null,
       sections: SearchRecSection[] | null
-    ): [SearchRecSection | null, number] => {
+    ): {index: number; section: SearchRecSection} | null => {
       if (highlightedIndex !== null && sections !== null) {
         let index = highlightedIndex
-        for (const s of sections) {
-          if (index >= s.data.length) {
-            index -= s.data.length
+        for (const section of sections) {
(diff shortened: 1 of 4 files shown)


shared/team-building/index.tsx
```diff
@@ -281,9 +304,10 @@ class TeamBuilding extends React.PureComponent<Props, {}> {
         >
           <Kb.SectionList
             ref={this.sectionListRef}
+            selectedIndex={this.props.highlightedIndex || 0}
```
zapu: This prop does not exist in react-native `SectionList`. It's only relevant to desktop section list where entries can be selected. I don't know how to pass this prop properly.

zapu: I added `type DesktopProps = { selectedIndex: number | undefined }` in `shared/common-adapters/section-list.d.ts` and changed exported Props type to be `React.ComponentProps<typeof SectionList> & DesktopProps`. It still resolves to `any`, but the intent is there.

zapu:
ready to review cc @keybase/react-hackers

PR #18667, opened 2019-07-31
fix desktop router

chrisnojima:
@keybase/react-hackers fixes some layout issues in electron 6 backported


shared/fs/browser/index.tsx
```diff
@@ -51,19 +51,17 @@ const Browser = (props: Props) =>
   props.path === Constants.defaultPath ? (
     <Root />
   ) : (
-    <Kb.BoxGrow>
```
chrisnojima: handled in the router now

PR #18676, opened 2019-07-31
Sort in contactSearch

zapu:
Add sorting in `usersearch.go` `contactSearch` so if any matches are deduplicated with usernames, the best match will remain.

NOTE: needs a test, but I want to merge this ASAP because it's a flake in one of the tests.


zapu:
@heronhaye can you take a look? it doesn't change much in grand scheme of things but one test was flaking because we were not storting contact results anymore. So right now we have to sort twice - first contact results and then all results. It was the easiest fix.

PR #18677, opened 2019-07-31
fix phone popup on settings page

jakob223:
Issue: Y2K-393

cc @keybase/y2ksquad

### DIFF
--- a/shared/actions/config/index.tsx
+++ b/shared/actions/config/index.tsx
@@ -420,21 +420,17 @@ const routeToInitialScreen = (state: Container.TypedState) => {
   }
 }
 
-let maybeLoadAppLinkOnce = true
+let maybeLoadAppLinkOnce = false
 const maybeLoadAppLink = (state: Container.TypedState) => {
   const phones = state.settings.phoneNumbers.phones
   if (!phones || phones.size > 0) {
     return
   }
 
-  if (
-    !maybeLoadAppLinkOnce ||
-    !state.config.startupLink ||
-    !state.config.startupLink.endsWith('/phone-app')
-  ) {
+  if (maybeLoadAppLinkOnce || !state.config.startupLink || !state.config.startupLink.endsWith('/phone-app')) {
     return
   }
-  maybeLoadAppLinkOnce = false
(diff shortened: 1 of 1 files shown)


shared/actions/config/index.tsx
```diff
@@ -420,12 +420,22 @@ const routeToInitialScreen = (state: Container.TypedState) => {
...
     return
   }
 
+  if (
+    !maybeLoadAppLinkOnce ||
+    !state.config.startupLink ||
+    !state.config.startupLink.endsWith('/phone-app')
+  ) {
+    return
+  }
```
heronhaye: is this racy?

jakob223: What is racing what?

heronhaye: idk if this is possible in react but if maybeLoadAppLink runs twice at the same time, they might both pass the !maybeLoadAppLinkOnce check

jakob223: hmm. I think (1) there's no reason this action should get triggered twice anywhere near one another and (2) if it did then we'd just route to the same screen twice and it would be fine.

shared/actions/config/index.tsx
```diff
@@ -420,12 +420,22 @@ const routeToInitialScreen = (state: Container.TypedState) => {
   }
 }
 
+let maybeLoadAppLinkOnce = true
```
pzduniak: i think this should be false?

shared/actions/config/index.tsx
```diff
@@ -420,12 +420,22 @@ const routeToInitialScreen = (state: Container.TypedState) => {
...
 const maybeLoadAppLink = (state: Container.TypedState) => {
   const phones = state.settings.phoneNumbers.phones
   if (!phones || phones.size > 0) {
     return
   }
 
+  if (
+    !maybeLoadAppLinkOnce ||
```
pzduniak: and this should be true, just for consistency with the initial routing. my snippet was wrong

jakob223:
Want to make this popup go away so merging, will open up another PR if there are more changes.

PR #18680, opened 2019-07-31
Support unsigned sep7 URLs with a warning

ddworken:
Completes PICNIC-285. This allows unsigned sep7 links if the origin_domain and signature fields are both not set. See the doc page with example sep7 links for an example valid sep7 link that does not have a signature. 

image
image
paste (1)


shared/wallets/sep7-confirm/index.tsx
```diff
@@ -93,35 +94,53 @@ const InfoRow = (props: InfoRowProps) => (
...
-        <Kb.Box2
-          direction="horizontal"
-          style={{backgroundColor: Styles.globalColors.transparent, marginLeft: Styles.globalMargins.xtiny}}
-        >
-          <Kb.Icon sizeType="Small" style={styles.verifiedIcon} type="iconfont-success" />
-        </Kb.Box2>
+        {props.signed && (
+          <Kb.Box2
+            direction="horizontal"
+            style={{backgroundColor: Styles.globalColors.transparent, marginLeft: Styles.globalMargins.xtiny}}
```
cjb: add to `styles`

ddworken: Done

PR #18684, opened 2019-08-01
phonenumbers: delete phone number from cache when user deletes it

jakob223:
cc @patrickxb @keybase/y2ksquad 

Issue: Y2K-391

### DIFF
--- a/go/contacts/phonebook.go
+++ b/go/contacts/phonebook.go
@@ -93,15 +93,16 @@ func (s *SavedContactsStore) RetrieveContacts(mctx libkb.MetaContext) (ret []key
 	return cache.Contacts, nil
 }
 
-func (s *SavedContactsStore) ClearPhoneNumber(mctx libkb.MetaContext,
-	phoneNumber keybase1.PhoneNumber) {
+func (s *SavedContactsStore) UnresolveContactsWithComponent(mctx libkb.MetaContext,
+	phoneNumber *keybase1.PhoneNumber, email *keybase1.EmailAddress) {
 	contactList, err := s.RetrieveContacts(mctx)
 	if err != nil {
 		mctx.Warning("Failed to get cached contact list: %x", err)
+		return
 	}
 	for i, con := range contactList {
-		if con.Component.PhoneNumber != nil &&
-			*con.Component.PhoneNumber == keybase1.RawPhoneNumber(phoneNumber) {
+		if (phoneNumber != nil && con.Component.PhoneNumber != nil && *con.Component.PhoneNumber == keybase1.RawPhoneNumber(*phoneNumber)) ||
+			(email != nil && con.Component.Email != nil && *con.Component.Email == *email) {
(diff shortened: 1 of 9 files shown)


heronhaye:
sorry i'll leave one comment: add a test please

go/phonenumbers/user.go
```diff
@@ -72,6 +73,11 @@ func DeletePhoneNumber(mctx libkb.MetaContext, phoneNumber keybase1.PhoneNumber)
 	}
 
 	_, err := mctx.G().API.Delete(mctx, arg)
+
+	if err != nil {
```
patrickxb: if err == nil?

go/contacts/phonebook.go
```diff
@@ -92,3 +92,21 @@ func (s *SavedContactsStore) RetrieveContacts(mctx libkb.MetaContext) (ret []key
...
 }
+
+func (s *SavedContactsStore) ClearPhoneNumber(mctx libkb.MetaContext,
+	phoneNumber keybase1.PhoneNumber) {
+	contactList, err := s.RetrieveContacts(mctx)
+	if err != nil {
+		mctx.Warning("Failed to get cached contact list: %x", err)
+	}
+	for i, con := range contactList {
+		if *con.Component.PhoneNumber == keybase1.RawPhoneNumber(phoneNumber) {
```
zapu: `con.Component.PhoneNumber` can be nil if it's an email component

zapu:
@heronhaye can you take a look? this is still pretty rough but might be good enough for now

go/emails/user.go
```diff
@@ -77,7 +91,13 @@ func SetVisibilityEmail(mctx libkb.MetaContext, email keybase1.EmailAddress, vis
 	}
 
 	_, err := mctx.G().API.PostJSON(mctx, arg)
-	return err
+	if err != nil {
+		return err
+	}
+	if visibility == keybase1.IdentityVisibility_PRIVATE {
```
heronhaye: good catch, also need to do it below for SetVisibilityAllEmail

heronhaye: (same for phone)

go/contacts/phonebook.go
```diff
@@ -92,3 +92,40 @@ func (s *SavedContactsStore) RetrieveContacts(mctx libkb.MetaContext) (ret []key
...
+		mctx.Warning("Failed to get cached contact list: %x", err)
+		return
+	}
+	for i, con := range contactList {
+		var unresolve bool
+		switch {
+		case phoneNumber != nil && con.Component.PhoneNumber != nil:
+			unresolve = *con.Component.PhoneNumber == keybase1.RawPhoneNumber(*phoneNumber)
+		case email != nil && con.Component.Email != nil:
+			unresolve = *con.Component.Email == *email
```
heronhaye: could quickly change this to use `strings.ToLower`

PR #18686, opened 2019-08-01
proto teamIDOfConv returns pointer

mlsteele:


mlsteele:
merging while CI is down

PR #18695, opened 2019-08-01
Use withdraw type if it is in the Asset

patrickxb:


patrickxb:
got a bunch of lint errors in ci for stuff i didn't touch...ignoring.

PR #18698, opened 2019-08-01
settings/notifications: redesign, empty screens

jakob223:
Issue: Y2K-259

https://zpl.io/VQBzLLn
https://zpl.io/VDB8vvJ though the copy still needs to be updated in the server

### DIFF
--- a/shared/settings/notifications/container.tsx
+++ b/shared/settings/notifications/container.tsx
@@ -34,7 +34,7 @@ export default Container.connect(
     _groups: state.settings.notifications.groups,
     allowEdit: state.settings.notifications.allowEdit,
     mobileHasPermissions: state.push.hasPermissions,
-    showEmailSection: !state.settings.email.emails,
+    showEmailSection: !!state.settings.email.emails && state.settings.email.emails.size > 0,
     sound: state.config.notifySound,
     waitingForResponse: state.settings.waitingForResponse,
   }),
--- a/shared/settings/notifications/container.tsx
+++ b/shared/settings/notifications/container.tsx
@@ -1,11 +1,11 @@
 import * as React from 'react'
 import {Reloadable} from '../../common-adapters'
 import * as SettingsGen from '../../actions/settings-gen'
-import {refreshNotificationsWaitingKey} from '../../constants/settings'
 import * as Container from '../../util/container'
 import Notifications, {Props} from '.'
 import * as RouteTreeGen from '../../actions/route-tree-gen'
 import * as ConfigGen from '../../actions/config-gen'
+import * as Constants from '../../constants/settings'
 
 type OwnProps = {}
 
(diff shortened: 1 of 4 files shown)


shared/settings/notifications/container.tsx
```diff
@@ -34,11 +34,13 @@ export default Container.connect(
     _groups: state.settings.notifications.groups,
     allowEdit: state.settings.notifications.allowEdit,
     mobileHasPermissions: state.push.hasPermissions,
+    showEmailSection: !state.settings.email.emails,
```
songgao: Should this check `state.settings.email.emails.size` too?

shared/settings/notifications/render.tsx
```diff
@@ -49,56 +56,89 @@ const Group = (props: {
...
+  <Group
+    allowEdit={props.allowEdit}
+    groupName="email"
+    label="All email notifications will be sent to your primary email address."
+    onToggle={props.onToggle}
+    onToggleUnsubscribeAll={() => props.onToggleUnsubscribeAll('email')}
+    title="Email notifications"
+    unsub="mail"
+    settings={props.groups.email && props.groups.email.settings}
+    unsubscribedFromAll={props.groups.email && props.groups.email.unsubscribedFromAll}
```
songgao: Does this need a few `!!`'s?

shared/settings/notifications/render.tsx
```diff
@@ -49,56 +56,89 @@ const Group = (props: {
...
           label={`Unsubscribe me from all ${props.unsub} notifications`}
         />
-      </Box>
+      </Kb.Box>
     )}
-  </Box>
+    <Kb.Divider style={styles.divider} />
+  </Kb.Box2>
+)
+const EmailSection = (props: Props) => (
```
songgao: I used to do a lot of these too, but I think if you use it as a function instead of component, it's slightly cheaper. Not a big deal though and I'm OK as-is!

shared/settings/notifications/container.tsx
jakob223: Oops, also I removed an `!` in debugging and forgot to put it back.

shared/settings/notifications/render.tsx
jakob223: I don't think so. It passes TS and the component just checks if settings is falsey.

jakob223: oh oops you meant unsub. That should be fine too, the component includes a `|| null`.

jakob223: Eh, I'd rather move it to where these are. You're right.

shared/settings/notifications/render.tsx
jakob223: Hmm, how much cheaper? This definitely feels like a legibility win to me.

chrisnojima: prefer components to functions, don't optimize early for no reason

jakob223:
- [ ] after release

PR #18699, opened 2019-08-01
add another stage to loading threads so we can skip the verifying keys step

mmaxim:
Patch does the following:

1. Adds the new `UIThreadLoader` which now fully implements `GetThreadNonblock`. 
2. Add a new stage to `UIThreadLoader` which runs `Boxer.ValidSenderKey` after we send up the full thread to the UI. The point of this is to get messages on the screen faster. Also, do some fancy footwork with caching in order to get messages into the cache only after all stages of unboxing are complete.
3. Add an optional `UnboxMode` to chat contexts that allows the caller to control if `Boxer` runs `ValidSenderKey` (and possibly other things) or not. In addition, in `UnboxModeQuick`, we do not cache any results on disk.
4. Add `Boxer.ResolveSkippedUnboxed` in order to run any checks that were skipped in any unbox mode other than `UnboxModeFull`. 
5. Make `TransformSupersedes` handle replies as well.
6. Add a whole status bar UI for handling loading status.


mmaxim:
(I'm going to merge over the gocritic errors)

go/chat/boxer.go
```diff
@@ -903,7 +948,7 @@ func (b *Boxer) unboxV2orV3orV4(ctx context.Context, boxed chat1.MessageBoxed,
 		BodyHash:              bodyHashSigned,
 		HeaderHash:            headerHash,
 		HeaderSignature:       nil,
-		VerificationKey:       &boxed.VerifyKey,
+		VerificationKey:       &senderKeyToValidate,
```
joshblum: how did this work before?

go/chat/uithreadloader.go
```diff
@@ -0,0 +1,518 @@
...
+				displayed = true
+			}
+		case <-ctx.Done():
+			t.Debug(ctx, "setUIStatus: context canceled")
+		}
+		if displayed {
+			typ, _ := status.Typ()
+			t.Debug(ctx, "setUIStatus: displaying: %v", typ)
+		}
+		resCh <- displayed
```
joshblum: maybe better in a `defer` in case a short circuit is added later

PR #18702, opened 2019-08-01
notify GUI about breaks when FS_GUI identify behavior is used

songgao:


go/kbfs/simplefs/simplefs.go
```diff
@@ -2905,21 +2905,39 @@ func (k *SimpleFS) SimpleFSGetStats(ctx context.Context) (
 
 // SimpleFSSubscribePath implements the SimpleFSInterface.
 func (k *SimpleFS) SimpleFSSubscribePath(
-	ctx context.Context, arg keybase1.SimpleFSSubscribePathArg) error {
+	ctx context.Context, arg keybase1.SimpleFSSubscribePathArg) (err error) {
+	if arg.IdentifyBehavior != nil {
```
strib: Would still be nice to put this in `makeContext()` as we talked about before, but I can't remember why we didn't do that in a previous PR (or why the identify behavior didn't end up in the path type itself -- though that wouldn't have helped us here).  Maybe that PR isn't merged yet?

Also, this isn't a new issue but we should be using `makeContext()` in these functions as well, to make sure we get a proper debug tag.

songgao: > or why the identify behavior didn't end up in the path type itself

It did end up in keybase1.Path, but since this is a KBFS path only API, I just went with plain string for the path. I think the reason for makeContext() was iti'd require adding an error return parameter which would make makeContext() used inline impossible. Probably not a huge deal though. I'll try to put that in.

strib: It's fine that this is plain string I guess, but I don't think the identity behavior stuff is in `keybase1.Path` yet: https://github.com/keybase/client/blob/c167df74baae895b3726d0b82a7c47bf0df92600/protocol/avdl/keybase1/simple_fs.avdl#L62-L63

Maybe I'm looking in the wrong place though...

songgao: Oh sorry; it's not merged yet. I was gonna merge these two together. https://github.com/keybase/client/pull/18301

strib: Ah that's why I was confused.  Thanks.  Would still be nice to have a helper function for this that could be called in these three functions.

songgao:
finally

PR #18703, opened 2019-08-01
folder_branch_ops: cache broken proof identify result for a bit

strib:
When something like simplefs sets an identify behavior that just warns on broken proofs, we should cache that identify result for a while, to avoid overloading the service with identify calls on every libkbfs call.

Issue: HOTPOT-487

### DIFF
--- a/go/kbfs/libkbfs/folder_branch_ops.go
+++ b/go/kbfs/libkbfs/folder_branch_ops.go
@@ -87,7 +87,7 @@ const (
 	markAndSweepPeriod = 1 * time.Hour
 	// A hard-coded reason used to derive the log obfuscator secret.
 	obfuscatorDerivationString = "Keybase-Derived-KBFS-Log-Obfuscator-1"
-	// How long to we skip identifies after seeing one with a broken
+	// How long do we skip identifies after seeing one with a broken
 	// proof warning?
 	cacheBrokenProofIdentifiesDuration = 5 * time.Minute
 )
@@ -313,9 +313,10 @@ type folderBranchOps struct {
 	nodeCache NodeCache
 
 	// Whether we've identified this TLF or not.
-	identifyLock sync.Mutex
-	identifyDone bool
-	identifyTime time.Time
+	identifyLock            sync.Mutex
+	identifyDone            bool
+	identifyTime            time.Time
+	identifyDoneWithWarning bool
 
 	// The current status summary for this folder
 	status *folderBranchStatusKeeper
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -87,6 +87,9 @@ const (
 	markAndSweepPeriod = 1 * time.Hour
 	// A hard-coded reason used to derive the log obfuscator secret.
 	obfuscatorDerivationString = "Keybase-Derived-KBFS-Log-Obfuscator-1"
+	// How long to we skip identifies after seeing one with a broken
```
songgao: s/to/do?

strib:
@songgao changed the logic as discussed, please take another sanity check pass!  Thanks!

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -2166,7 +2170,9 @@ func (fbo *folderBranchOps) identifyOnce(
 	defer fbo.identifyLock.Unlock()
 
 	ei := tlfhandle.GetExtendedIdentify(ctx)
-	if fbo.identifyDone && !ei.Behavior.AlwaysRunIdentify() {
+	if (fbo.identifyDone && !ei.Behavior.AlwaysRunIdentify()) ||
+		(fbo.identifyDoneWithWarning &&
+			ei.Behavior.WarningInsteadOfErrorOnBrokenTracks()) {
```
songgao: So even if `ei.Behavior.AlwaysRunIdentify()` is true, we could still skip. Though I guess AlwaysRunIdentify and WarningInsteadOfErrorOnBrokenTracks are mutually exclusive.

strib: Good point, I'll put it all in a big `AlwaysRunIdentify` statement.

go/kbfs/libkbfs/folder_branch_ops.go
```diff
@@ -2186,8 +2195,31 @@ func (fbo *folderBranchOps) identifyOnce(
...
+		// still allows access to the files.  Identifies are
+		// expensive, so in this case we skip future identifies that
+		// also have this behavior for a short time period.
 		fbo.log.CDebugf(ctx,
-			"Identify finished with no error but broken proof warnings")
+			"Identify finished with no error but broken proof warnings; "+
+				"caching result for %d", cacheBrokenProofIdentifiesDuration)
+		fbo.identifyDoneWithWarning = true
+		timer := time.NewTimer(cacheBrokenProofIdentifiesDuration)
+		fbo.goTracked(func() {
```
songgao: Since now checks are moved inside the !alwaysrun block, it's possible we reach here even when identifyDoneWithWarning==true, then we'd have multiple timers. Looks like the timer needs to be move up so we can stop it before starting a new one?

strib: You're talking about a case where both `AlwaysRunIdentify == true` and `WarningInsteadOfErrorOnBrokenTracks == true`?  No matter what we can't stop the timer from the previous call into this function unless we save it a way to cancel it somewhere, which doesn't seem worth it for this case that can't even happen.

strib: Or actually I can just check whether it's already set in that case, and avoid making a new timer.

PR #18709, opened 2019-08-02
increase fs http token to 32 bytes and base64 it

songgao:


strib:
(I hit rebuild on this PR FYI, seemed like an infrastructure flake)

PR #18712, opened 2019-08-02
Fix the selected tab changing while in 'What is Stellar?' from airdrop

cjb:
@keybase/picnicsquad 

The Airdrop option in the wallet account switcher isn't a real account, so we fake its selection in the account list by looking to see if the current route is the Airdrop page. That works up until you go to the "What is Stellar?" modal from that Airdrop page, at which point it's not the current route anymore.

Here's a fix for that, though seems like we'll also want to change this up to make it less brittle.

### DIFF
--- a/shared/constants/wallets.tsx
+++ b/shared/constants/wallets.tsx
@@ -705,8 +705,14 @@ export const getAccountIDs = (state: TypedState) => state.wallets.accountMap.key
 export const getAccounts = (state: TypedState) => state.wallets.accountMap.valueSeq().toList()
 
 export const getAirdropSelected = () => {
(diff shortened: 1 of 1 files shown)


nathunsmitty:
@cjb do you have any ideas on how to make this less brittle in the future? From my experience working with wallet routing it seems like we derive a lot more state through the routes than in other parts of the app. It'd be nice if we could start a frontend discussion or wiki page about what the recommended way to handle state/routing in cases like this are so we have a set of defined best practices

cjb:
@nathunsmitty I was reluctant to put `airdrop` into `state.wallets.selectedAccount` since it's not a real AccountID, but maybe that's just actually the right idea given how ugly the alternative's getting.

PR #18713, opened 2019-08-02
email todo verification state

pzduniak:
depends on https://github.com/keybase/keybase/pull/4136

somewhat ready for a review, will get changed once @jzila finishes his changes to the todo in the release branch


shared/actions/settings.tsx
```diff
@@ -416,9 +418,14 @@ const editEmail = (_, action: SettingsGen.EditEmailPayload, logger: Saga.SagaLog
     return RPCTypes.emailsSetPrimaryEmailRpcPromise({email: action.payload.email})
   }
   if (action.payload.verify) {
-    return RPCTypes.emailsSendVerificationEmailRpcPromise({email: action.payload.email}).then(() =>
-      SettingsGen.createSentVerificationEmail({email: action.payload.email})
-    )
+    return RPCTypes.emailsSendVerificationEmailRpcPromise({email: action.payload.email}).then(() => [
+      SettingsGen.createSentVerificationEmail({email: action.payload.email}),
+      PeopleGen.createGetPeopleData({
```
zapu: I don't think that's the right way to do it. It won't update other clients that may be observing people tab.

With the last fix to propagate todo updates using gregor state notifications, we should be fine if the server sends the update notification. But I'm not 100% familiar with how this works right now, I'll find out.

protocol/avdl/keybase1/home.avdl
```diff
@@ -89,6 +94,15 @@ protocol home {
     default: void;
   }
 
+  variant HomeScreenTodoExt switch (HomeScreenTodoType t) {
```
zapu: @heronhaye another look if this is necessary to extend existing todo? I'm not a big fan of having to do all this code on both server and client becasue of avdl limitation

heronhaye: what's wrong with the data you already added?

pzduniak: We need to somehow communicate to the client when was the last time we sent the verification email. I'd imagine another way to approach this would be to communicate this through gregor state?

zapu: I think there are two separate issues here:
1) returning extra data from `/home` handler. We can't change existing `HomeScreenTodo` variant without making it incompatible with client going out today. So one way is adding separate field entirely, which is not ideal, especially considering it will be used just for one todo type.
2) Refreshing people tab when a button is clicked. I think because this additional field does not increase todo version, clients beside the one that makes that action will not know to refresh that data. But I'm not sure if that's exactly how it works.

go/home/home.go
```diff
@@ -100,7 +100,7 @@ func (h *Home) getToCache(ctx context.Context, markedViewed bool, numPeopleWante
 	return nil
 }
 
-func (h *Home) Get(ctx context.Context, markViewed bool, numPeopleWanted int) (ret keybase1.HomeScreen, err error) {
+func (h *Home) Get(ctx context.Context, markViewed bool, numPeopleWanted int, bustCache bool) (ret keybase1.HomeScreen, err error) {
```
maxtaco: Can you explain why this needs to change? What about the previous system was insufficient for this use case?

go/home/home.go
```diff
@@ -111,23 +111,26 @@ func (h *Home) Get(ctx context.Context, markViewed bool, numPeopleWanted int) (r
 	h.Lock()
 	defer h.Unlock()
 
-	useCache, people := h.peopleCache.isValid(ctx, h.G(), numPeopleWanted)
+	var useCache bool
```
zapu: I got it to work without changes here. See comment on server PR

jzila:
FYI my changes are in.

shared/constants/people.tsx
```diff
@@ -151,8 +160,11 @@ export const reduceRPCItemToPeopleItem = (
     if (!todo) {
       return list
     }
+    const todoExt: RPCTypes.HomeScreenTodoExt | null =
```
zapu: won't it infer this type correctly?

pzduniak: yeah, probably

shared/constants/types/people.tsx
```diff
@@ -12,7 +12,11 @@ export const stringToItemID: (id: string) => ItemID = id => id
 export type TodoTypeEnum = RPCTypes.HomeScreenTodoType
 export type TodoType = keyof typeof RPCTypes.HomeScreenTodoType
 
-export type _TodoMetaEmail = {type: 'email'; email: string}
+export type _TodoMetaEmail = {
+  type: 'email'
+  email: string
+  lastVerifyEmailDate?: number
```
buoyad: `// in seconds`

shared/people/todo/container.tsx
```diff
@@ -269,14 +269,19 @@ const VerifyAllEmailConnector = connect(
     },
   }),
   (s, d, o: TodoOwnProps) => {
-    const meta = o.metadata
+    const meta = o.metadata && o.metadata.type === 'email' ? o.metadata : undefined
+
+    // Has the user received a verification email less than 30 minutes ago?
+    const hasRecentVerifyEmail =
+      meta && meta.lastVerifyEmailDate && new Date().getTime() / 1000 - meta.lastVerifyEmailDate < 30 * 60
```
buoyad: `new Date().getTime()` -> `Date.now()`

shared/settings/account/email-phone-row.tsx
```diff
@@ -40,13 +42,24 @@ const _EmailPhoneRow = (props: Kb.PropsWithOverlay<Props>) => {
...
+    subtitle = 'Check your inbox'
+  } else {
+    if (props.type === 'email' && props.primary) {
+      subtitle = addSpacer(subtitle, 'Primary')
+    }
+    if (!props.searchable) {
+      subtitle = addSpacer(subtitle, 'Not searchable')
+    }
+    if (hasRecentVerifyEmail) {
+      subtitle = addSpacer(subtitle, 'Check your inbox')
```
buoyad: Maybe this should go first on desktop so "Check your inbox" has some context next to "UNVERIFIED"
image

shared/constants/types/people.tsx
buoyad: also, since this is getting `0` as a default value in `makeTodoMetaEmail`, might as well get rid of the `?` here

PR #18719, opened 2019-08-02
disable buttons when too many phones/emails

jakob223:
Issue: Y2k-368


shared/settings/account/index.tsx
```diff
@@ -10,15 +10,17 @@ import flags from '../../util/feature-flags'
 export type Props = {
   addedEmail: string | null
   contactKeys: I.List<string>
-  hasPassword: boolean
-  supersededPhoneNumber?: string
+  hasPassword: Boolean
```
buoyad: lowercase `boolean`

PR #18723, opened 2019-08-02
[Y2K-411] Clear the email error when the field is changed

jzila:
cc @keybase/y2ksquad

### DIFF
--- a/shared/settings/account/add-modals.tsx
+++ b/shared/settings/account/add-modals.tsx
@@ -31,6 +31,13 @@ export const Email = () => {
       dispatch(RouteTreeGen.createClearModals())
     }
   }, [addEmailInProgress, addedEmail, dispatch])
+  // clean on edit
+  React.useEffect(() => {
+    if (email !== addEmailInProgress) {
+      onAddEmailInProgress('')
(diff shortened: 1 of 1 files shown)


shared/settings/account/add-modals.tsx
```diff
@@ -31,6 +31,13 @@ export const Email = () => {
       dispatch(RouteTreeGen.createClearModals())
     }
   }, [addEmailInProgress, addedEmail, dispatch])
+  // clean on edit
+  React.useEffect(() => {
+    if (email !== addEmailInProgress) {
+      onAddEmailInProgress('')
```
jakob223: I don't understand what you're doing with `onAddEmailInProgress('')`

shared/settings/account/add-modals.tsx
```diff
@@ -31,6 +31,13 @@ export const Email = () => {
...
+      dispatch(SettingsGen.createClearAddingEmail())
```
jakob223: Can we call this only when there's an error, instead of every time?

shared/settings/account/add-modals.tsx
jzila: Just resetting it, although it isn't strictly necessary.

PR #18725, opened 2019-08-02
Fix contact list rendering issues on ios / android

zapu:
cc @keybase/react-hackers @jzila

### DIFF
--- a/shared/team-building/__test__/team-building.test.tsx
+++ b/shared/team-building/__test__/team-building.test.tsx
@@ -33,7 +33,10 @@ describe('team building list', () => {
       },
     ]
 
-    // our example list is as follows:
+    // Our example list is as follows:
+    // Team building list does not render section footers, but they still occupy
+    // an index that getItemLayout has to be aware of.
+    //
     //  0  [ test 1 ]
     //  1  row 0
     //  2  row 1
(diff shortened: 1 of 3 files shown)


shared/team-building/__test__/team-building.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    //  6  row 0
+    //  7  row 1
+    //  8  (footer)
+    //  9  [ test 3 ]
+    // 10  row 0
+    // 11  row 1
+    // 12  row 2
+    // 13  (footer)
+
+    const sectionLength = SectionDivider.height
```
jzila: Please name this `sectionDividerHeight`.

shared/team-building/index.tsx
```diff
@@ -238,9 +239,12 @@ class TeamBuilding extends React.PureComponent<Props, {}> {
     sections: Array<SearchRecSection>,
     indexInList: number
   ): {index: number; length: number; offset: number} => {
+    const sectionLength = Kb.SectionDivider.height
+    const rowLength = userResultHeight
```
jzila: Let's use `height` instead of length. I'm realizing it'd be a lot easier to understand the code with that simple change.

zapu: sounds good. react-native return type here is `{ index, offset, length }` but I agree that "length" can be misleading in lists growing vertically

shared/team-building/index.tsx
```diff
@@ -250,17 +254,23 @@ class TeamBuilding extends React.PureComponent<Props, {}> {
         break
       }
       numSections++
-      const indexInSection = indexInList - currSectionHeaderIdx
+      const indexInSection = indexInList - currSectionHeaderIdx - 1
+      if (indexInSection === s.data.length) {
+        // it's the section footer (we don't render footers so 0px).
```
buoyad: 😮

PR #18726, opened 2019-08-02
explicitly ask for location perm on android

mmaxim:
Should get in for the release.


shared/actions/platform-specific/index.native.tsx
```diff
@@ -47,6 +47,22 @@ const requestPermissionsToWrite = (): Promise<void> => {
   return Promise.resolve()
 }
 
+const requestLocationPermission = (): Promise<void> => {
```
chrisnojima: why not make this async too

PR #18733, opened 2019-08-02
Move asset links to expanded view

cecileboucheron:
I moved the links to the expanded view and made them buttons. I kept the Exchange one to the right but I would gladly ditch it if Patrick agrees (I think it clutters the view).

image

It's a bit tight on iPhone SE though. It should be okay on bigger devices.
image

@keybase/react-hackers @patrickxb

### DIFF
--- a/shared/wallets/asset/index.tsx
+++ b/shared/wallets/asset/index.tsx
@@ -89,7 +89,7 @@ export default class Asset extends React.Component<Props, State> {
           </Kb.Box2>
         </Kb.ClickableBox>
         {this.state.expanded && (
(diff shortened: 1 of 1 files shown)


patrickxb:
If we're going to do this, the "Exchange" link under "0 USD" should go away and just have the button in the expanded view.

patrickxb:
Ok, let's do this and remove the exchange link.

patrickxb:
One more question:  now that they are buttons, is it easy to put a spinner on Deposit, Withdraw?  Not needed for Exchange.

chrisnojima:
lets make another pr to fix this up

PR #18744, opened 2019-08-02
kbnm: de-linting

strib:
cc: @keybase/hotpotatosquad

Issue: HOTPOT-479


go/kbnm/main_test.go
```diff
@@ -27,7 +27,7 @@ func TestProcess(t *testing.T) {
 	out := json.NewEncoder(&outBuf)
 
 	// Check invalid JSON:
-	io.WriteString(&inBuf, "invalid json, hi\n")
+	_, _ = io.WriteString(&inBuf, "invalid json, hi\n")
```
joshblum: can't hurt to check the error here and below

PR #18745, opened 2019-08-02
kbtest: de-linting

strib:
cc: @keybase/hotpotatosquad

Issue: HOTPOT-479


go/kbtest/kbtest.go
```diff
@@ -430,7 +430,7 @@ func RunTrackWithOptions(tc libkb.TestContext, fu *FakeUser, username string, op
 // considered `possible`, but not `valid` by libphonenumber.
 func GenerateTestPhoneNumber() string {
 	ret := make([]byte, 7)
-	rand.Read(ret)
+	_, _ = rand.Read(ret)
```
zapu: ouch, are these required now?

might as well panic on error here

strib: Yes, error checking or explicit non-checking is now required, as it should be!  Sure, i can change this one to a panic.

PR #18747, opened 2019-08-02
gregor: de-linting

strib:
cc: @keybase/hotpotatosquad

Issue: HOTPOT-479


go/gregor/client/client.go
```diff
@@ -165,7 +165,7 @@ func (c *Client) Stop() chan struct{} {
 	ch := make(chan struct{})
 	close(c.stopCh)
 	go func() {
-		c.eg.Wait()
+		_ = c.eg.Wait()
```
joshblum: log this

PR #18749, opened 2019-08-02
lru: de-linting

strib:
cc: @keybase/hotpotatosquad

Issue: HOTPOT-479


go/lru/disk_lru.go
```diff
@@ -324,7 +327,10 @@ func (d *DiskLRU) Put(ctx context.Context, lctx libkb.LRUContext, key string, va
 	defer func() {
 		// Commit the index
 		if err == nil && index != nil && index.IsDirty() {
-			d.writeIndex(ctx, lctx, index, true)
+			err := d.writeIndex(ctx, lctx, index, true)
```
joshblum: we shouldn't have a new `err` variable in this case but return the error since we are force writing the index

PR #18755, opened 2019-08-02
pvl: de-linting

strib:
cc: @keybase/hotpotatosquad

Issue: HOTPOT-479


go/pvl/interp_test.go
```diff
@@ -2096,7 +2096,7 @@ func runPvlTest(t *testing.T, unit *interpUnitTest) {
 			fail("proof should have failed")
 		}
 
-		err = nil
+		// err is definitely nil here.
```
mlsteele: since this is a test I don't care. But this makes the code fragile since it's easy to unwittingly change this invariant by changing distant code above

go/pvl/interp.go
```diff
@@ -195,11 +195,12 @@ func checkProofInner(m metaContext, pvlS string, service keybase1.ProofType, inf
...
 	}
 
-	if len(errs) == 0 {
+	switch len(errs) {
+	case 0:
 		return nil
-	} else if len(errs) == 1 {
+	case 1:
 		return errs[0]
-	} else {
```
mlsteele: Switch is clearer. Fine to merge anyway. Hope disabling this rule works out

strib: Yeah, the new code is the switch, which I agree is clearer.  Did you mean you think the if-else chain is clearer?  I was actually glad this one was caught by the linter.

go/pvl/interp_test.go
strib: True, but in general if `err` is ever `nil` the function should be returning (or the test failiing) anyway, right?  So we wouldn't even get down here.

go/pvl/interp.go
mlsteele: oh, my bad. I thought the switch was old even though generally I can read diffs :D
I meant the switch is clearer. Maybe my understanding of the linter rule is backwards, it's been adding switches the whole time?

go/pvl/interp_test.go
mlsteele: Assuming you meant if `err` is ever not-nil. Sometimes errors are logged and discarded. If `err` is used then it can end up non-nil even though the function should keep chugging. Generally, not here.

go/pvl/interp.go
strib: Yep, adding switches.  Some people don't like that partly because it negates the built-in scoping of some if-else statements like:

```go
if err != nil {
	return err
} else if otherErr, ok := err.(OtherErr); ok {
  // do something with otherErr
} else {
  // do something else
}
```

go/pvl/interp_test.go
strib: Yeah true, though I'd argue that `err` should be set to `nil` only in the place where it's explicitly being ignored, like:

```go
if err != nil {
   log.CDebugf(ctx, "This is the error: %+v", err)
  err = nil
}
```

That seems most foolproof and lintable.

PR #18756, opened 2019-08-02
runtimestats: de-linting

strib:
cc: @keybase/hotpotatosquad

HOTPOT-479


go/runtimestats/stats.go
```diff
@@ -66,7 +65,7 @@ func (r *Runner) Stop(ctx context.Context) chan struct{} {
 		close(r.stopCh)
 		r.started = false
 		go func() {
-			r.eg.Wait()
+			_ = r.eg.Wait()
```
joshblum: log error?

PR #18762, opened 2019-08-03
[Y2K-413] fix verification todo

jzila:


jzila:
So currently this sets the "Waiting" on click, but it doesn't correctly refresh or go away once the email is actually verified. Not sure where that hook is, but at least this moves us closer to what it should be.

jzila:
cc @keybase/y2ksquad

PR #18765, opened 2019-08-04
fix types and modernize some devices

chrisnojima:
@keybase/react-hackers this removes all implicit any types from devices and switches the paper key screen to hooks
also adds the hoist non react statics so we can have better acting hocs (if any)


shared/devices/container.tsx
```diff
@@ -106,14 +115,4 @@ const NamedConnected = Container.namedConnect(
 const SafeSub = Container.safeSubmitPerMount(['onBack'])
 const Connected = NamedConnected(SafeSub(ReloadableDevices))
 
-if (!Container.isMobile) {
-  // @ts-ignore fix this
-  Connected.navigationOptions = {
```
chrisnojima: connect actually hoists this correctly so we can apply it to the wrapped class instead of adding it to the connect after the fact

chrisnojima:
@keybase/react-hackers bump

PR #18766, opened 2019-08-04
Change assertion colors for profile pages

zapu:
There is an unfortunate change where it has to do `ReportLastTrack` first because `DisplayKey` and other proofs depend on whether user is tracked or not. This visibly changes the text order during console `keybase id` command, but does not seem to have any other side effects.

Also changed UI text colors to `dark` variants.


zapu:
cc @keybase/react-hackers

chrisnojima:
- [ ] after release

zapu:
@maxtaco another look?

CI fail is golint in chat package, unrelated to this PR.

PR #18768, opened 2019-08-04
impove avatar loading

mmaxim:
Patch does the following:

1. Refactor attachment HTTP server so we can use the server in more packages in the service. Create `manager.Srv` in `kbhttp` with the factored out code.
2. Adds a new generic token in front of `manager.Srv`.
3. Add protocol for keeping JS current with the address and token of HTTP server.
4. Add `avatars.Srv` for serving avatars via the HTTP server.
5. Refactor avatars JS to strip out all old code that invoked RPCs to get URLs, and replace with simpler code that just gets the URLs _a priori_.


shared/actions/config/index.tsx
```diff
@@ -102,6 +110,10 @@ function* loadDaemonBootstrapStatus(
     yield Saga.put(loadedAction)
     // request follower info in the background
     yield RPCTypes.configRequestFollowerInfoRpcPromise({uid: s.uid})
+    // set HTTP srv info
+    yield Saga.put(
+      ConfigGen.createUpdateHTTPSrvInfo({address: s.httpSrvInfo.address, token: s.httpSrvInfo.token})
```
chrisnojima: You don't need to make this. Just handle it in the reducer

shared/actions/teams.tsx
```diff
@@ -1089,19 +1089,7 @@ function* setPublicity(state: TypedState, action: TeamsGen.SetPublicityPayload)
...
-      return [
-        ConfigGen.createLoadTeamAvatars({teamnames: [name]}),
-        ConfigGen.createLoadAvatars({usernames: [name]}),
-      ]
-    case RPCTypes.AvatarUpdateType.user:
-      return [ConfigGen.createLoadAvatars({usernames: [name]})]
-    case RPCTypes.AvatarUpdateType.team:
-      return [ConfigGen.createLoadTeamAvatars({teamnames: [name]})]
-  }
+  return [ConfigGen.createAvatarRefreshIncrement({name: action.payload.params.name})]
```
chrisnojima: Don't need this action either

shared/actions/config/index.tsx
mmaxim: I did it like this because this action also gets fired as a result of a notification from the service. With this, there is just one place in the reducer where these values are updated, instead of the bootstrap handler doing it, and then a handler for the notification. Seemed cleaner this way.

shared/actions/teams.tsx
mmaxim: This one I agree with, will change.

mmaxim:
cc @songgao

PR #18770, opened 2019-08-05
Add support for sep10 authorization in sep6

patrickxb:


go/stellar/stellarsvc/anchor.go
```diff
@@ -238,3 +286,113 @@ func (a *anchorInteractor) domainMatches(url string) bool {
 	}
 	return urlTLD == assetTLD
 }
+
+func (a *anchorInteractor) getAuthToken(mctx libkb.MetaContext) error {
+	u, err := url.ParseRequestURI(a.asset.AuthEndpoint)
+	if err != nil {
+		return err
+	}
+	if !a.domainMatches(u.Host) {
```
xgess: doesn't look like this verifies `https`. can we do that also?

patrickxb: done

PR #18778, opened 2019-08-05
email can be null

chrisnojima:
@keybase/react-hackers @pzduniak


shared/reducers/settings.tsx
```diff
@@ -98,12 +98,16 @@ function reducer(state: Types.State = initialState, action: Actions): Types.Stat
...
-            email.merge({
-              isVerified: true,
-            })
-          )
+        .updateIn(
+          ['email', 'emails'],
+          emails =>
+            emails
+              ? emails.update(action.payload.params.emailAddress, email =>
+                  email.merge({
```
buoyad: Would be good to check this one too.

PR #18780, opened 2019-08-05
Close profile airdrop popup on select

nathunsmitty:
Fixes an issue @xgess noted where the popup remained even after a selection.


nathunsmitty:
@keybase/picnicsquad

PR #18782, opened 2019-08-05
golangci-lint: disable `ifElseChain`

strib:
By popular vote.

### DIFF
--- a/go/.golangci.yml
+++ b/go/.golangci.yml
@@ -1,5 +1,10 @@
+linters-settings:
+  gocritic:
+    disabled-checks:
+      - ifElseChain
+
 linters:
-        enable:
-                - gofmt
-                - gocritic
-                - unconvert
+  enable:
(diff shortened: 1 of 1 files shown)


heronhaye:
It's very minor but I'm not a fan of singleCaseSwitch either. For cases where I'm switching on an enum but it only has one value right now. https://github.com/keybase/client/pull/18751/files#diff-b5f24cfde42ac5f61c23f2e532140669L24

Don't really mind either way though

strib:
@heronhaye sure, though in cases like that it's easy enough to keep the linter happy by adding a default case with a log statement or something, which is arguably what you should be doing anyway for code clarity, to communicate that the single-case switch is intentional.

strib:
(And actually, a comment in the `default` case seems fine too.)

heronhaye:
good point. I could have moved the return statement into a default in that code as well

PR #18785, opened 2019-08-05
Plumb through error messages on trivial path payments

nathunsmitty:
Plumb through amount error messages so users know why they can't send payments. The red text is what was added. This same type of error message is shown for non-trivial path payments.

Screen Shot 2019-08-05 at 11 51 39 AM


nathunsmitty:
@keybase/picnicsquad

PR #18791, opened 2019-08-05
a few Fs fixes

songgao:


songgao:
test passed locally. I think it's the flake that we have a ticket for. Merging.

PR #18793, opened 2019-08-05
Use Wallet disclaimer text from server

cjb:
@keybase/picnicsquad CC @keybase/react-hackers @patrickxb @malgorithms @xgess 

Here's a PR which makes the onboarding screen (the intro panel and then the disclaimer text) use Markdown that comes from the server config, similarly to how the airdrop details page works -- it reuses the same details structure as that page.

To keep the PR safe, it'll just show the existing disclaimer text if the server didn't return anything.  So the only visible change occurs once the server is sending down a new disclaimer field -- @xgess has a keybase/keybase xgess/airdrop-disclaimer-text PR in testing which makes that change.

The disclaimer is customizable per-airdrop so that it can talk about the current airdrop.  If there's no airdrop active, it falls back to the standard text.


patrickxb:
LGTM, will defer to @xgess for final approval

cjb:
Merging over Go CI fail in TestUIThreadLoaderCache.

PR #18797, opened 2019-08-06
Don't spam search while scrolling results

buoyad:
- Fix `onEndReached` spam by reducing `onEndReachedThreshold` & throttling the call
- Add `keyboardShouldPersistTaps="handled"` to recs / results lists so you don't have to tap twice to select a row.
- Make search result ordering deterministic (by @zapu)

cc @keybase/y2ksquad

### DIFF
--- a/shared/team-building/container.tsx
+++ b/shared/team-building/container.tsx
@@ -1,7 +1,7 @@
 import logger from '../logger'
 import * as React from 'react'
 import * as I from 'immutable'
-import {debounce, trim, throttle} from 'lodash-es'
+import {debounce, trim} from 'lodash-es'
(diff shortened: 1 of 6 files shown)


shared/team-building/container.tsx
```diff
@@ -252,13 +252,12 @@ const deriveOnEnterKeyDown = memoizeShallow(
   }
 )
 
-const deriveOnSearchForMore = memoizeShallow(
-  ({search, searchResults, searchString, selectedService}) => () => {
+const deriveOnSearchForMore = ({search, searchResults, searchString, selectedService}) =>
+  throttle(() => {
```
zapu: i think remove throttle from here if there's one in index

zapu:
@heronhaye can you take a look at go code?

PR #18806, opened 2019-08-06
dark mode part 2

chrisnojima:
- [x] storybooks have a dark/light button. if you shift+click it'll automatically swap so you can quickly go through various screens to validate


chrisnojima:
@keybase/react-hackers @cecileboucheron @adamjspooner

PR #18809, opened 2019-08-06
Display "You can send assets other than XLM" banner in the send form when started via chat messages

ddworken:
This banner did not display since the from field was not getting properly routed to the component. In order to route this information the wallet icon in chat now calls the loadaccounts RPC in order to load all of the accounts into the redux store (if they aren't already in the redux store). It then passes the default account ID to the send form which passes it to the go service when determining whether or not to show the banner.

Since the RPC is only called when the wallet icon is fist mounted AND the list of accounts in redux is empty, this is not expected to have significant performance implications. I had initially implemented this as doing the RPC call once the wallet icon was clicked, but this led to a race condition where the user could quickly click on the wallet icon and then the send button and open the send form before the load accounts RPC returned. This race condition is still technically possible but much less likely with this setup.


shared/chat/conversation/input-area/normal/wallets-icon/index.tsx
```diff
@@ -4,40 +4,54 @@ import * as Styles from '../../../../../styles'
...
-      onHidden={props.toggleShowingMenu}
-      position="top right"
-      visible={props.showingMenu}
-    />
-  </Kb.Box2>
-)
+class WalletsIcon extends React.Component<WalletsIconProps & Kb.OverlayParentProps, {}> {
+  componentDidMount() {
+    // Load the wallet data at mount so we can determine the account ID of the default wallet
+    this.props.loadWalletsData()
```
nathunsmitty: I think it makes more sense to call this RPC on wallet icon click and use a loading indicator + waiting key to handle the race condition issue. It feels a little weird to me to be prefetching data like this on a mount if we don't have to.

ddworken: Where would you display the loading indicator for that operation? I guess instead of the contents of the popup? Just trying to imagine what UI we're going for here. 

image

ddworken:
Updated per @mlsteele's feedback with a go side fix. It turns out the RPC call take in a `FromPrimaryAccount` argument which this code was ignoring. So it was a much easier fix than I expected.

PR #18813, opened 2019-08-06
fix TestSubscriptionManagerSubscribePath flake maybe

songgao:
I wasn't able to repro on my machine (with `repeat 100`), but I think what happened was that the unsubscribe happened before the notification actually got sent out from the fbo. So this PR inserts a few waits before the unsubscribe calls.

### DIFF
--- a/go/kbfs/libkbfs/subscription_manager_test.go
+++ b/go/kbfs/libkbfs/subscription_manager_test.go
@@ -17,8 +17,8 @@ import (
 	"golang.org/x/net/context"
 )
 
-func delayedFinish(t *testing.T, ctl *gomock.Controller, timeout time.Duration) (
-	finish func(), done func(args ...interface{})) {
+func waitForCall(t *testing.T, timeout time.Duration) (
+	waiter func(), done func(args ...interface{})) {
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/subscription_manager_test.go
```diff
@@ -78,39 +82,50 @@ func TestSubscriptionManagerSubscribePath(t *testing.T) {
 	notifier.EXPECT().OnPathChange(sid1, "/keybase/private/jdoe",
 		keybase1.PathSubscriptionTopic_CHILDREN)
 	notifier.EXPECT().OnPathChange(sid2, "/keybase/private/jdoe",
-		keybase1.PathSubscriptionTopic_STAT)
+		keybase1.PathSubscriptionTopic_STAT).Do(done1)
 	_, _, err = config.KBFSOps().CreateDir(
 		ctx, rootNode, rootNode.ChildName("dir1"))
 	require.NoError(t, err)
 
+	// wait for done1 to happen before unsubscribing to avoid races
```
strib: `t.Log()` these comments with capitalization and punctuation please.

PR #18815, opened 2019-08-07
immer git

chrisnojima:
- [x] change git to immer
- [x] use chainaction2
- [x] use some hooks


shared/.eslintrc.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
     'jsx-quotes': 'off',
     'no-duplicate-imports': 'off',
     'no-empty': 'off',
+    'no-useless-return': 'off',
```
chrisnojima: conflicts w/ our TS strict rules

shared/actions/git.tsx
```diff
@@ -3,74 +3,84 @@ import * as GregorGen from './gregor-gen'
...
 
-const load = (state: TypedState) =>
-  state.config.loggedIn &&
-  RPCTypes.gitGetAllGitMetadataRpcPromise(undefined, Constants.loadingWaitingKey)
-    .then((results: Array<RPCTypes.GitRepoResult> | null) =>
-      // @ts-ignore codemod-issue
-      GitGen.createLoaded(Constants.parseRepos(results || []))
-    )
-    .catch(() => {})
+const load = async (state: TypedState) => {
```
chrisnojima: promise chainActions should mostly be async await

shared/constants/entities.tsx
```diff
@@ -1,5 +1,4 @@
 import * as I from 'immutable'
-import * as Git from './git'
```
chrisnojima: we should mostly deprecate the entity store

shared/git/container.tsx
```diff
@@ -1,134 +1,94 @@
...
-  {} & {
-    _loadGit: (() => void)
-  } & Exclude<
-      React.ElementConfig<typeof Git>,
-      {
-        expandedSet: I.Set<string>
-        onToggleExpand: ((arg0: string) => void)
-      }
-    > */
+const getRepos = memoize((git: Map<string, Types.GitInfo>) =>
```
chrisnojima: simplify this a bunch

chrisnojima:
@keybase/react-hackers bump

shared/git/container.tsx
```diff
@@ -1,132 +1,94 @@
...
+  sortBy([...git.values()], ['teamname', 'name']).reduce<{personals: Array<string>; teams: Array<string>}>(
```
MarcoPolo: 🐑 Why did you get rid of partition?
```js
const [teams, personals] = partition(sortBy([...git.values()], ['teamname', 'name']), x => !!x.teamname)
return {teams, personals}
```

chrisnojima: It still partitions

MarcoPolo: oh, sorry I missed that you had still call map(x => x.id). this is fine nvm

PR #18822, opened 2019-08-07
Recover password in-app partial integration

pzduniak:
missing:
 - mobile wasn't tested at all
 - reset does not have a success screen
 - lots of blinking when you restart the flow
 - paper key input works, but you can't change the password (white screen)

before merge make sure to disable the new flag!


buoyad:
Start service with `--features env_autoreset` to test

shared/actions/recover-password.tsx
```diff
@@ -0,0 +1,173 @@
...
+      })
+    )
+    const action:
+      | RecoverPasswordGen.SubmitPaperKeyPayload
+      | RecoverPasswordGen.AbortPaperKeyPayload = yield Saga.take([
+      RecoverPasswordGen.submitPaperKey,
+      RecoverPasswordGen.abortPaperKey,
+    ])
+
+    if (action.payload && typeof action.payload.paperKey === 'string') {
```
buoyad: Check `action.type` here too

shared/actions/recover-password.tsx
```diff
@@ -0,0 +1,173 @@
...
+  )
+  yield Saga.put(RouteTreeGen.createNavigateUp())
+}
+
+function* recoverPasswordSaga() {
+  yield* Saga.chainGenerator<RecoverPasswordGen.StartRecoverPasswordPayload>(
+    RecoverPasswordGen.startRecoverPassword,
+    startRecoverPassword
+  )
+  yield Saga.chainGenerator<RecoverPasswordGen.RestartRecoveryPayload>(
```
buoyad: This can be a `chainAction`

shared/util/feature-flags.native.tsx
```diff
@@ -21,6 +21,7 @@ const ff: FeatureFlags = {
   outOfDateBanner: false,
   plansEnabled: false,
   proofProviders: true,
+  resetPipeline: true,
```
buoyad: We probably either want to leave these off or enable it by default in the service

shared/login/recover-password/device-selector/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+const DeviceSelector = (props: Props) => {
+  const items = [
+    ...props.devices,
+    {
+      name: 'account-reset',
+      type: 'account-reset',
+    },
+  ]
+  return (
+    <SignupScreen banners={[]} buttons={[]} onBack={props.onBack} title="Recover password">
```
buoyad: `banners` and `buttons` are optional. In designs the body of this screen has a white background, currently it's blue

shared/login/recover-password/error/container.tsx
```diff
@@ -0,0 +1,36 @@
...
+    error: state.recoverPassword.paperKeyError.stringValue(),
+  }),
+  dispatch => ({
+    onBack: () => dispatch(RouteTreeGen.createNavigateUp()),
+  }),
+  (s, d, o: OwnProps) => ({...o, ...s, ...d})
+)(Error)
+
+// @ts-ignore fix this
+ConnectedError.navigationOptions = {
```
buoyad: I believe the vogue is to set this on `Error` itself

shared/login/recover-password/device-selector/container.tsx
```diff
@@ -0,0 +1,47 @@
...
+    onResetAccount: d.onResetAccount,
+    onSelect: (name: string) => {
+      const device = s.devices.find(device => device.name === name)
+      d._onSelect(device ? device.id : '')
+    },
+  })
+)(DeviceSelector)
+
+// @ts-ignore fix this
+ConnectedDeviceSelector.navigationOptions = {
```
buoyad: same here

shared/login/recover-password/device-selector/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <Kb.Box2 direction="vertical" alignItems="center" fullWidth={true} fullHeight={true}>
+        <Kb.Text type="Body">Which do you have handy?</Kb.Text>
+        <Kb.Box2
+          direction="vertical"
+          fullWidth={true}
+          fullHeight={true}
+          style={styles.contentBox}
+          gap={'medium'}
```
buoyad: double quotes

shared/login/recover-password/explain-device/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export type Props = {
+  deviceName: string
+  deviceType: string
+  onBack: () => void
+  onComplete: () => void
+}
+
+const ExplainDevice = (props: Props) => {
+  const isMobile = props.deviceType === 'mobile'
```
buoyad: This screen always thinks I'm a desktop even when I choose a mobile.  Also can we rename this so we don't collide with the more common `isMobile`.

shared/login/recover-password/paper-key/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  onSubmit: (paperKey: string) => void
+}
+
+const PaperKey = (props: Props) => {
+  const [paperKey, setPaperKey] = React.useState('')
+  const onSubmit = React.useCallback(() => props.onSubmit(paperKey), [paperKey])
+
+  return (
+    <SignupScreen
+      banners={[]}
```
buoyad: :fire:

shared/login/recover-password/prompt-reset/container.tsx
```diff
@@ -0,0 +1,34 @@
...
+import * as Kb from '../../../common-adapters'
+import * as Styles from '../../../styles'
+import * as Container from '../../../util/container'
+import * as RecoverPasswordGen from '../../../actions/recover-password-gen'
+import DeviceSelector from '.'
+import {InfoIcon} from '../../../signup/common'
+
+type OwnProps = {}
+
+const ConnectedPromptReset = Container.connect(
```
buoyad: Curious why you went with `connect` rather than hooks for these?

shared/actions/recover-password.tsx
```diff
@@ -0,0 +1,173 @@
...
+import * as Constants from '../constants/provision'
+import * as Container from '../util/container'
+import {RPCError} from '../util/errors'
+
+const chooseDevice = (params, response) => {
+  return Saga.callUntyped(function*() {
+    const devices = (params.devices || []).map(d => Constants.rpcDeviceToDevice(d))
+
+    yield Saga.put(
+      RecoverPasswordGen.createShowDeviceListPage({
```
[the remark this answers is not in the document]
pzduniak: `displayDeviceSelect` for consistency with other stuff?

shared/login/recover-password/prompt-reset/container.tsx
pzduniak: Other than this being old code, I find it somewhat hard to figure out where to use hooks for redux state. Would it mean that I wouldn't have a container.tsx? How would the stories work then?

shared/login/recover-password/prompt-reset/container.tsx
```diff
@@ -0,0 +1,34 @@
...
+  _ => ({}),
+  dispatch => ({
+    onContinue: (action: boolean) => dispatch(RecoverPasswordGen.createSubmitResetPrompt({action})),
+  }),
+  (s, d, o: OwnProps) => ({...o, ...s, ...d})
+)(DeviceSelector)
+
+// @ts-ignore fix this
```
chrisnojima: please don't duplicate this ts-ignore. this is working elsewhere. just don't put it on the container, put it on the wrapped class

pzduniak: base pr is oooold and it was my first larger fe change, im nearly done with getting this up to date

shared/login/recover-password/prompt-reset/container.tsx
buoyad: Yeah, you wouldn't have `container.tsx` for these then. Stories will still work but i believe you need to add a mock store. Device stories are a good example.
https://github.com/keybase/client/blob/e3d61efb87746b074c1f0b834236750597451f6b/shared/devices/device-page/index.stories.tsx#L61

shared/actions/recover-password.tsx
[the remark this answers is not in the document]
buoyad: Ah ok. My real issue is that these actions say `display*` but their receivers don't trigger the nav. For example:
https://github.com/keybase/client/blob/e3d61efb87746b074c1f0b834236750597451f6b/shared/actions/provision.tsx#L495-L507
I think you can keep the names the same, but then there should be `chainAction`s in this file that do the nav action, rather than dispatching multiple actions at the call site to get it done.

shared/login/recover-password/device-selector/index.tsx
buoyad: use the `noBackground` prop here.

shared/login/recover-password/paper-key/container.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }),
+  dispatch => ({
+    onBack: () => dispatch(RecoverPasswordGen.createAbortPaperKey()),
+    onSubmit: (paperKey: string) => dispatch(RecoverPasswordGen.createSubmitPaperKey({paperKey})),
+  }),
+  (s, d, o: OwnProps) => ({...o, ...s, ...d})
+)(PaperKey)
+
+// @ts-ignore fix this
+ConnectedPaperKey.navigationOptions = {
```
buoyad: Move to `index` on `PaperKey`

shared/login/recover-password/prompt-reset/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import {ButtonType} from '../../../common-adapters/button'
+
+export type Props = {
+  onContinue: (choice: boolean) => void
+}
+
+const PromptReset = (props: Props) => {
+  return (
+    <SignupScreen
+      banners={[]}
```
buoyad: :fire:

shared/package.json
```diff
@@ -291,7 +291,7 @@
   },
   "optionalDependencies": {
     "regedit": "2.2.7",
-    "webfonts-generator": "0.4.0"
+    "webfonts-generator": "^0.4.0"
```
buoyad: revert

shared/util/feature-flags.desktop.tsx
```diff
@@ -25,6 +25,7 @@ const ff: FeatureFlags = {
   outOfDateBanner: false,
   plansEnabled: false,
   proofProviders: true,
+  resetPipeline: true,
```
buoyad: don't forget to flip before merge

shared/login/recover-password/paper-key/index.tsx
buoyad: :fire::fire::fire:

PR #18823, opened 2019-08-07
chat: de-linting

strib:
Issue: HOTPOT-492


go/chat/commands/source.go
```diff
@@ -200,43 +200,3 @@ func (s *Source) PreviewBuiltinCommand(ctx context.Context, uid gregor1.UID, con
 		cmd.Preview(ctx, uid, convID, tlfName, text)
 	}
 }
-
```
mmaxim: I'd like to keep this here

strib: Ok done

strib:
Thanks for looking!  On your last two points, does "not a huge fan" mean you want to make a case for disabling those rules across all of the Keybase projects?  I don't know that there's much we can do about the 3rd point though, as it's just part of the general error checking rule.  Basically, we would either need to change the function being `go`'d to not return an error, or wrap it in something that discards the error.

mmaxim:
I'd be up for taking out #2 for sure.

strib:
Ok.  Maybe on Monday we can take another poll.  I don't want to make the decision based just on someone who has a strong opinion.  I personally think it makes the code easier to read to not nest the `if` unnexessarily, so I don't think it's an obvious call.

PR #18824, opened 2019-08-07
avatar cleanup

chrisnojima:
Cleanup the avatar rendering a little bit post the url changes

### DIFF
--- a/shared/chat/conversation/messages/message-popup/exploding/index.tsx
+++ b/shared/chat/conversation/messages/message-popup/exploding/index.tsx
@@ -88,7 +88,7 @@ class ExplodingPopupHeader extends React.Component<PropsWithTimer<Props>, State>
           <Box2 direction="horizontal">
             <Text type="BodySmall">by</Text>
             <Box2 direction="horizontal" gap="xtiny" gapStart={true} style={styles.user}>
-              <Avatar username={author} size={16} clickToProfile="tracker" />
+              <Avatar username={author} size={16} />
               <ConnectedUsernames
                 onUsernameClicked="profile"
                 colorFollowing={true}
--- a/shared/common-adapters/avatar.tsx
+++ b/shared/common-adapters/avatar.tsx
@@ -86,42 +86,23 @@ const followIconHelper = (size: number, followsYou: boolean, following: boolean)
   }
 }
 
-const mapStateToProps = (state: Container.TypedState, ownProps: OwnProps) => {
-  return {
+const ConnectedAvatar = Container.connect(
+  (state, ownProps: OwnProps) => ({
     _counter: state.config.avatarRefreshCounter.get(ownProps.username || ownProps.teamname || '', 0),
     _following: ownProps.showFollowingStatus ? state.config.following.has(ownProps.username || '') : false,
     _followsYou: ownProps.showFollowingStatus ? state.config.followers.has(ownProps.username || '') : false,
     _httpSrvAddress: state.config.httpSrvAddress,
     _httpSrvToken: state.config.httpSrvToken,
-  }
-}
-
-const mapDispatchToProps = (dispatch: Container.TypedDispatch, ownProps: OwnProps) => {
-  return {
+  }),
+  (dispatch, ownProps: OwnProps) => ({
     _goToProfile: (username: string) => dispatch(ProfileGen.createShowUserProfile({username})),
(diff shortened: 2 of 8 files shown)


shared/chat/conversation/messages/message-popup/exploding/index.tsx
```diff
@@ -88,7 +88,7 @@ class ExplodingPopupHeader extends React.Component<PropsWithTimer<Props>, State>
           <Box2 direction="horizontal">
             <Text type="BodySmall">by</Text>
             <Box2 direction="horizontal" gap="xtiny" gapStart={true} style={styles.user}>
-              <Avatar username={author} size={16} clickToProfile="tracker" />
```
chrisnojima: don't ever spawn the tracker

shared/common-adapters/avatar.tsx
```diff
@@ -89,55 +84,31 @@ const followIconHelper = (size: number, followsYou: boolean, following: boolean)
...
-      ? ownProps.style
-      : Styles.collapseStyles([
-          ownProps.style,
-          onClick && Styles.platformStyles({isElectron: Styles.desktopStyles.clickable}),
-        ])
-
-    const name = isTeam ? ownProps.teamname : ownProps.username
-    const urlMap = avatarSizes.reduce((m, size: number) => {
+    const onClick =
+      dispatchProps.onClick || (username ? () => dispatchProps._goToProfile(username) : undefined)
```
chrisnojima: this makes more things clickable, which seems fine to me (cc: @cecileboucheron )

chrisnojima:
@keybase/react-hackers

shared/common-adapters/avatar.tsx
```diff
@@ -89,55 +84,31 @@ const followIconHelper = (size: number, followsYou: boolean, following: boolean)
...
-const mapDispatchToProps = (dispatch: Container.TypedDispatch, ownProps: OwnProps) => {
-  return {
-    _goToProfile: (username: string, desktopDest: 'profile' | 'tracker') =>
-      Styles.isMobile || desktopDest === 'profile'
-        ? dispatch(ProfileGen.createShowUserProfile({username}))
-        : dispatch(Tracker2Gen.createShowUser({asTracker: true, username})),
+  }),
+  (dispatch, ownProps: OwnProps) => ({
+    _goToProfile: (username: string) => dispatch(ProfileGen.createShowUserProfile({username})),
     onClick: ownProps.onEditAvatarClick ? ownProps.onEditAvatarClick : ownProps.onClick,
```
jakob223: why even do this in dispatch props?

shared/styles/index.desktop.tsx
```diff
@@ -150,7 +150,7 @@ export const collapseStyles = (styles: ReadonlyArray<T.CollapsibleStyle>): Objec
     (a: Array<T.CollapsibleStyle>, e: T.CollapsibleStyle) => a.concat(e),
     []
   ) as Array<Object | null | false>
-  const style = flattenedStyles.reduce<Object>((o, e) => (e ? {...o, ...e} : o), {})
+  const style = flattenedStyles.reduce<Object>((o, e) => Object.assign(o, e), {})
```
jakob223: Isn't `(o, e) => Object.assign(o, e)` just `Object.assign`

shared/common-adapters/avatar.tsx
chrisnojima: mergeProps evaluates username, so it should delegate it. also we likely want to avoid ownProps in mapState and mapDispatch if possible

shared/styles/index.desktop.tsx
chrisnojima: its very dangerous to pass along functions from one place to another w/o constraint. for example here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Syntax
Object.assign  takes arbitrary params and merges them, so you can do
`Object.assign(a,b,d,e,f)` etc
what does reduce pass (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#Syntax)
`accumulator, currentValue[, index[, array]])[, initialValue]`, meaning if we just passed Object.assign to reduce it'd pass alllllllll those params in, which isn't what we want.

This pattern happens elsewhere. It's better to just be explicit about it

PR #18826, opened 2019-08-07
Put spinners on SEP6 deposit/withdraw buttons

cjb:

### DIFF
--- a/shared/actions/wallets.tsx
+++ b/shared/actions/wallets.tsx
@@ -1502,28 +1502,34 @@ const handleSEP6Error = (err: RPCError) => [
 ]
 
 const assetDeposit = (_: TypedState, action: WalletsGen.AssetDepositPayload) =>
-  RPCStellarTypes.localAssetDepositLocalRpcPromise({
-    accountID: action.payload.accountID,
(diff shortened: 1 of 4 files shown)


shared/actions/wallets.tsx
```diff
@@ -1502,28 +1502,34 @@ const handleSEP6Error = (err: RPCError) => [
...
+    {
+      accountID: action.payload.accountID,
+      asset: assetDescriptionOrNativeToRpcAsset(
+        Constants.makeAssetDescription({
+          code: action.payload.code,
+          issuerAccountID: action.payload.issuerAccountID,
+        })
+      ),
+    },
+    Constants.assetDepositWaitingKey(action.payload.issuerAccountID)
```
patrickxb: is this key supposed to be unique?  if so, we should add the action.payload.code as well.

cjb: @patrickxb Thanks, good catch, made this change!

Yes, it's supposed to uniquely identify one visible button on the screen to make it spin. So, I think if we'd had two assets, both expanded and showing the buttons, both with the same issuerAccountID but different asset codes, then clicking Deposit on one would have caused both to spin at the same time.

PR #18831, opened 2019-08-08
message drafts

mmaxim:
Add support for storing unsent messages in the service, as well as more adornment in the UI when these drafts exist. The flow is the following:

1. `UpdateUnsentText` is called currently when the user is typing, so inside save drafts in `Inbox` storage.
2. Attach drafts to inbox items we send up from `GetInboxNonblock`
3. In the UI we display the draft in the snippet for small team rows, and as a pencil icon for big team rows. We don't show the draft state when the row is selected.
4. Sub in draft text in the input area in the case where it is blank and the draft exists.


shared/constants/types/chat2/index.tsx
```diff
@@ -139,6 +139,7 @@ export type _State = {
   explodingModes: I.Map<Common.ConversationIDKey, number> // seconds to exploding message expiration,
   quote: QuoteInfo | null // last quoted message,
   selectedConversation: Common.ConversationIDKey // the selected conversation, if any,
+  previousSelectedConversation: Common.ConversationIDKey // the selected conversation, if any,
```
buoyad: change comment here

shared/constants/types/chat2/meta.tsx
```diff
@@ -19,6 +19,7 @@ export type _ConversationMeta = {
   conversationIDKey: Common.ConversationIDKey // should be the key for this meta EXCEPT for pendingConversationIDKey, in that case its the resolved conversation we're previewing,
   description: string
   descriptionDecorated: string
+  draft: string | null
```
buoyad: The rest of this type uses empty strings in place of `null`. could you do that too?

PR #18832, opened 2019-08-08
2 chainAction2 settings and cleanup

chrisnojima:


shared/actions/json/settings.json
```diff
@@ -174,9 +174,6 @@
     "processorProfile": {
       "durationSeconds": "number"
     },
-    "waitingForResponse": {
```
chrisnojima: using the waiting store instead which also lets us remove a bunch of chainGenerators

chrisnojima:
@keybase/react-hackers 
cc: @keybase/y2ksquad

shared/actions/settings.tsx
```diff
@@ -17,251 +16,230 @@ import {isAndroidNewerThanN, pprofDir, version} from '../constants/platform'
...
+
+    return [
+      RouteTreeGen.createNavigateUp(),
+      ...(action.payload.thenSignOut ? [ConfigGen.createLogout()] : []),
+    ]
   } catch (error) {
-    yield Saga.put(SettingsGen.createOnUpdatePasswordError({error}))
-  } finally {
-    yield Saga.put(SettingsGen.createWaitingForResponse({waiting: false}))
+    return SettingsGen.createOnUpdatePasswordError({error})
```
jakob223: Why is returning better than `Saga.put`?

shared/actions/settings.tsx
```diff
@@ -347,22 +317,23 @@ function* refreshNotifications() {
...
     SettingsGen.createNotificationsRefreshed({
-      // @ts-ignore
       notifications: I.Map(notifications),
     })
   )
 }
 
-const dbNuke = () => RPCTypes.ctlDbNukeRpcPromise()
+const dbNuke = async () => {
+  await RPCTypes.ctlDbNukeRpcPromise(undefined, Constants.settingsWaitingKey)
```
jakob223: Why `undefined` and not `{}`?

shared/actions/settings.tsx
chrisnojima: this isn't a generator anymore. Saga.put is only for sagas to dispatch

shared/actions/settings.tsx
chrisnojima: you don't need to return anything from these functions. if you do, then you need to cover all paths (best `return false`). in this case it doesn't return anything. previously it returned whatever `RPCTypes.ctlDbNukeRpcPromise()` returned which is a little dangerous. we essentially never want to plumb that through

PR #18840, opened 2019-08-08
fix paperkey

chrisnojima:
@keybase/react-hackers @zapu

### DIFF
--- a/shared/provision/routes.tsx
+++ b/shared/provision/routes.tsx
@@ -2,7 +2,7 @@ import CodePage from './code-page/container'
 import ErrorView from './error/container'
 import ForgotUsername from './forgot-username/container'
 import GpgSign from './gpg-sign/container'
-import Paperkey from './paper-key'
+import Paperkey from './paper-key/container'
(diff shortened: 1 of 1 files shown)


shared/provision/routes.tsx
```diff
@@ -14,7 +14,7 @@ export const newRoutes = {
   error: {getScreen: (): typeof ErrorView => require('./error/container').default},
   forgotUsername: {getScreen: (): typeof ForgotUsername => require('./forgot-username/container').default},
   gpgSign: {getScreen: (): typeof GpgSign => require('./gpg-sign/container').default},
-  paperkey: {getScreen: (): typeof Paperkey => require('./paper-key').default},
+  paperkey: {getScreen: (): typeof Paperkey => require('./paper-key/container').default},
```
buoyad: we should get around to prefixing these with `provision`

chrisnojima: yeah, it'l lbe better when all the routes are actually typed w/ ownProps, i'll make a ticket to continue that work

PR #18842, opened 2019-08-08
Check the Host header on incoming kbhttp requests

cjb:
@keybase/picnicsquad CC @songgao @jzila


go/kbfs/libhttpserver/server.go
```diff
@@ -139,6 +139,17 @@ func (s *Server) getHTTPFileSystem(ctx context.Context, requestPath string) (
 //     /team/keybase/file.txt?token=1234567890abcdef1234567890abcdef
 func (s *Server) serve(w http.ResponseWriter, req *http.Request) {
 	s.logger.Debug("Incoming request from %q: %s", req.UserAgent(), req.URL)
+	addr, err := s.server.Addr()
+	if err != nil {
+		s.logger.Debug("serve: failed to get HTTP server address: %s", err)
+		s.handleBadRequest(w)
```
songgao: This should probably be `StatusInternalServerError` like in chat? (instead of BadRequest)

cjb:
@songgao Makes sense; how's this?

mmaxim:
The code in the chat part of this should be somewhere else, I'll move it.

PR #18848, opened 2019-08-08
osxfuse: upgrade to 3.10.0 and fix Catalina redirector issue

strib:
Major changes:
* Rebuilt kbfuse from osxfuse 3.10.0, with Catalina support.
* When starting the redirector, it first tries to mount at `/keybase` and symlink there from `/Volumes/Keybase`.  If that fails (because `/keybase` can't be made), it does the reverse, and ignores any failures creating the `/keybase` symlink.
  * This requires an update to the helper process, which will require a password prompt on upgrades.
* Changed the title of the Keybase dmg to "Keybase App", so the mounted dmg won't conflict with `/Volumes/Keybase`.


osx/KBKit/KBKit/Component/KBRedirector.m
```diff
@@ -46,7 +46,7 @@ - (void)install:(KBCompletion)completion {
   DDLogDebug(@"Starting redirector: %@", params);
   [self.helperTool.helper sendRequest:@"startRedirector" params:@[params] completion:^(NSError *err, id value) {
     if (!err) {
-      completion(nil, value);
+      completion(nil);
```
songgao: Hmm what is this value used for and how did it not break stuff?

strib: That was a bug that I both introduced and fixed today.  I was coding on Linux so had to commit to test compiling on macOS.  I'll squash down those commits all together so it's not so obvious how much I suck.

strib:
@cjb would love to get a security-focused review of the `KBHelper.m` changes in this PR, though I'll probably just merge it when it passes CI so that we can start testing internally, and address any problems in a future PR.

PR #18849, opened 2019-08-08
consolidate join/leave messages

mmaxim:
Goal of the patch is to consolidate adjacent join/leave messages into a single message. Works using the new `UIThreadLoader` module to modify the thread on the way to the UI.

1. Add grouping functionality in `UIThreadLoader.
2. Improve logic for figuring out how to manage the multiple forms of state these messages can be in during `LoadNonblock`.
3. Modify the UI join/leave messages to show multiple people (except for your own join message which we do not consolidate)


shared/constants/chat2/index.tsx
```diff
@@ -171,6 +171,34 @@ export const getMessage = (
   id: Types.ConversationIDKey,
   ordinal: Types.Ordinal
 ): Types.Message | null => state.chat2.messageMap.getIn([id, ordinal])
+export const getDecoratedMessage = (
```
chrisnojima: likely better to just make this a type guard instead of this message or null function
https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

PR #18850, opened 2019-08-09
Icon clean up

cecileboucheron:
@keybase/react-hackers


chrisnojima:
did you run all these through imageoptim, seems likely not

cecileboucheron:
@chrisnojima ran image optims

cecileboucheron:
@chrisnojima I actually had to bring some icons back, like the search filters or the email-invite as they are still used. Ready for review again.

PR #18851, opened 2019-08-09
possible emoji crash

chrisnojima:
i saw a crash log where this was undefined but couldn't figure out how to repro. maybe @buoyad has an idea. i tried injecting nonsense topReactji and things like that but it didn't repro


chrisnojima:
Somehow c is undefined

PR #18859, opened 2019-08-09
Delete superseded phone number on add

buoyad:
On adding a phone number, check if we currently have a superseded item for that phone number. If it exists, delete it. This makes the yellow banner on account setting disappear. 

Also did some cleanup while looking around this code.
cc @keybase/y2ksquad

### DIFF
--- a/shared/actions/settings.tsx
+++ b/shared/actions/settings.tsx
@@ -349,16 +349,6 @@ const deleteAccountForever = async (state: TypedState) => {
   return ConfigGen.createSetDeletedSelf({deletedUsername: username})
 }
 
-const comparePhoneRows = (
-  row1: ChatTypes.Keybase1.UserPhoneNumber,
-  row2: ChatTypes.Keybase1.UserPhoneNumber
-) => {
-  if (row1.superseded !== row2.superseded) {
-    return row1.superseded ? -1 : 1
-  }
-  return row1.phoneNumber.localeCompare(row2.phoneNumber)
-}
-
 const loadSettings = async (
   state: TypedState,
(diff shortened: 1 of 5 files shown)


shared/actions/settings.tsx
```diff
@@ -373,12 +363,16 @@ const loadSettings = async (
       (settings.emails || []).map(row => [row.email, Constants.makeEmailRow(row)])
     )
     const phoneMap: I.Map<string, Types.PhoneRow> = I.Map(
-      // Sort the superseded numbers first, so that if a number exxists in both
-      // superseded and non-superseded form, the non-superseded version ends up
-      // in the map.
-      (settings.phoneNumbers || [])
-        .sort(comparePhoneRows)
-        .map(row => [row.phoneNumber, Constants.toPhoneRow(row)])
+      (settings.phoneNumbers || []).reduce(
```
buoyad: cc @jakob223 I changed this to a reduce rather than sorting

shared/settings/account/index.tsx
```diff
@@ -143,7 +143,7 @@ const AccountSettings = (props: Props) => (
         <Kb.Banner color="yellow" onClose={props.onClearSupersededPhoneNumber}>
           <Kb.BannerParagraph
             bannerColor="yellow"
-            content={`Your unverified phone number ${
+            content={`Your phone number ${
```
buoyad: cc @heronhaye / @zapu I noticed this wording was wrong after having a verified number become superseded. Is that a bug? It makes sense to me that that'd work.

buoyad:
@jakob223 does the whole PR look good or just what I cc'd you on?

jakob223:
Whole PR

PR #18862, opened 2019-08-09
dark mode control native storybook

chrisnojima:
@keybase/react-hackers adds dark mode controls to storybook on native


shared/storybook/index.native.js
```diff
@@ -34,23 +35,85 @@ const loadStories = () => {
...
+      const c = savedCallback.current
+      c && c()
+    }
+    if (delay !== null) {
+      let id = setInterval(tick, delay)
+      return () => clearInterval(id)
+    } else return undefined
+  }, [delay])
+}
+// keep modes in the module so its kept between stories
```
MarcoPolo: You mean so if you go to another store and come back, this is saved?

chrisnojima: yes, when you switch stories it remounts the whole thing. this way if you switch it keeps the mode. same things happens on desktop

PR #18863, opened 2019-08-09
Y2K-383 normalize emails

zapu:
This doesn't do much. We are already normalizing e-mails manually in both processing contacts and user search. I was initially going for a rework of how some types work and introduce a `NormalizedEmail` type so it's obvious what's going on. Ideally I'd like to be consistent with stuff like `go/externals/proof_service_*.go` `NormalizeUsername` functions. But I ran into some issues with mixed casing names w.r.t display names and SBS handling and didn't get to this in time.

### DIFF
--- a/go/systests/teams_implicit_test.go
+++ b/go/systests/teams_implicit_test.go
@@ -10,6 +10,8 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/davecgh/go-spew/spew"
+	"github.com/keybase/client/go/emails"
+	"github.com/keybase/client/go/kbtest"
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/keybase1"
 	"github.com/keybase/client/go/teams"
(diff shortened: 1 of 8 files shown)


go/teams/list.go
```diff
@@ -244,7 +244,7 @@ func ListTeamsVerified(ctx context.Context, g *libkb.GlobalContext,
 		if team.IsImplicit() {
 			displayName, err := team.ImplicitTeamDisplayNameString(ctx)
 			if err != nil {
-				m.Debug("| Failed to get ImplicitTeamDisplayNameString() for team %q: %v", team.ID, err)
+				m.Warning("| Failed to get ImplicitTeamDisplayNameString() for team %q: %v", team.ID, err)
```
heronhaye: sure this should be a warning?

go/teams/implicit.go
```diff
@@ -263,6 +280,13 @@ func LookupOrCreateImplicitTeam(ctx context.Context, g *libkb.GlobalContext, dis
 		return res, teamName, impTeamName, err
 	}
 
+	if err := assertIsDisplayNameNormalized(lookupName); err != nil {
+		// Do not allow display names with mixed letter case - while it's legal
+		// to create them, it will not be possible to load them because API
+		// server always downcases during normalization.
```
heronhaye: Does it have to be lowercase for looking up, or just creating?

zapu: for both, api endpoints will just refuse to find that team, at least from my testing. maybe there's something else at play but I've exhausted time I had to debug this :(

this at least stops you from creating "broken" imp teams.

heronhaye: well this works for example `keybase chat send "ZAPU.NET@https" test`, or do you just mean for unresolved impteams?

zapu: this resolves to a username so it's fine. but something like `ZAPU.NET.PL@https` would make a new implicit team which would sort of work in chat but cause different errors along the way - i think including during sbs resolution handler.

go/systests/teams_implicit_test.go
```diff
@@ -512,3 +514,40 @@ func TestResolveSBSConsolidatedTeamWithConflict(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, teamid2, lookupTeamID)
 }
+
+func TestCreateAndResolveEmailImpTeam(t *testing.T) {
```
zapu: @heronhaye added another test to make sure mixed-case email SBS is covered

Assuming bob's mailserver is case-sensitive, this will make maintenanced not being able to send email to him because it will see "bob+someemail@noemail.keybase.io" instead of "BOB+someemail@noemail.keybase.io". But otherwise everything else works - adding that email, verifying it, making it public, and being let into an sbs team.

heronhaye: Looks good

go/teams/list.go
zapu: it's only used by the cli command if i'm not mistaken, and the result is empty displayName field in returned JSON, which was confusing when I was first debugging these issues. I can change to a `Debug` though, I don't feel strongly about this one.

go/teams/implicit.go
zapu: @maxtaco can you take a look at this change? it prevents creating an implicit team that client will not be able to load by name. I don't know what the real fix is here yet

maxtaco: I'm fine with this but I don't understand why we don't just downcase all the time.

zapu:
I'll merge this and a made a new ticket to figure out hackernews and where to downcase it.

It already works correctly in GUI because search always gets downcased assertions.

PR #18865, opened 2019-08-09
Contact token client support

heronhaye:
- [x] try this out while watching ratelimits to make sure tokens work

### DIFF
--- a/go/contacts/cache.go
+++ b/go/contacts/cache.go
@@ -59,6 +59,7 @@ type cachedLookupResult struct {
 
 type lookupResultCache struct {
 	Lookups map[ContactLookupKey]cachedLookupResult
+	Token   Token
 	Version struct {
 		Major int
 		Minor int
@@ -157,6 +158,11 @@ func (s *ContactCacheStore) ClearCache(mctx libkb.MetaContext) error {
 	return s.encryptedDB.Delete(mctx.Ctx(), cacheKey)
 }
 
+func (c *CachedContactsProvider) LookupAllWithToken(mctx libkb.MetaContext, emails []keybase1.EmailAddress,
+	numbers []keybase1.RawPhoneNumber, userRegion keybase1.RegionCode, _ Token) (res ContactLookupResults, err error) {
+	return c.LookupAll(mctx, emails, numbers, userRegion)
+}
+
 func (c *CachedContactsProvider) LookupAll(mctx libkb.MetaContext, emails []keybase1.EmailAddress,
 	numbers []keybase1.RawPhoneNumber, userRegion keybase1.RegionCode) (res ContactLookupResults, err error) {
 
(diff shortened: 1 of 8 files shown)


heronhaye:
I'll rebuild but chat test has flaked twice in a row

heronhaye:
Ping

go/service/contacts.go
```diff
@@ -20,11 +20,18 @@ type bulkLookupContactsProvider struct{}
 
 var _ contacts.ContactsProvider = (*bulkLookupContactsProvider)(nil)
 
+func (c *bulkLookupContactsProvider) LookupAllWithToken(mctx libkb.MetaContext, emails []keybase1.EmailAddress,
+	numbers []keybase1.RawPhoneNumber, userRegion keybase1.RegionCode, token contacts.Token) (contacts.ContactLookupResults, error) {
+	defer mctx.TraceTimed(fmt.Sprintf("bulkLookupContactsProvider#LookupAllWithToken(len=%d)", len(emails)+len(numbers)),
+		func() error { return nil })()
+	return contacts.BulkLookupContacts(mctx, emails, numbers, userRegion, token)
```
zapu: I think the tokens are implementation-specific enough that we can just handle them inside of `contacts.BulkLookupContacts` and not "leak" outside of that package. What do you think?

zapu: hm, that can't be exactly that... maybe that's necessary after all, too bad

go/contacts/cache.go
```diff
@@ -217,8 +223,10 @@ func (c *CachedContactsProvider) LookupAll(mctx libkb.MetaContext, emails []keyb
 	mctx.Debug("After checking cache, %d emails and %d numbers left to be looked up", len(remainingEmails), len(remainingNumbers))
 
 	if len(remainingEmails)+len(remainingNumbers) > 0 {
-		apiRes, err := c.Provider.LookupAll(mctx, remainingEmails, remainingNumbers, userRegion)
+		apiRes, err := c.Provider.LookupAllWithToken(mctx, remainingEmails, remainingNumbers, userRegion, conCache.Token)
 		if err == nil {
+			conCache.Token = apiRes.Token
```
zapu: looks like the adhoc tofu searches for just one phonenumber/email will always overwrite your big tokens for contact lists

heronhaye: I don't think so, the server sends back a token for everything in the token as well as what was requested

heronhaye: sorry I could be wrong about that...but if the server did do that, it'd be ok?

go/service/contacts.go
heronhaye: yeah i tried to make it not leak but found it hard with them both satisfying the same interface

heronhaye:
merging over lint errors that I didn't cause

PR #18866, opened 2019-08-09
teams: vendor in new teamchain test vectors

maxtaco:


maxtaco:
FYI @heronhaye, just vendoring in the test cases you took a look at earlier

maxtaco:
CI was taking close to an hour, going to merge without it.

PR #18870, opened 2019-08-09
SBS profile tweaks

jakob223:
Design: https://zpl.io/bP9OOeD
Issue: Y2K-321


MarcoPolo:
@malgorithms maybe https://keybase.io/install should redirect to https://keybase.io/download for older (and currently published) clients.

malgorithms:
> @malgorithms maybe https://keybase.io/install should redirect to https://keybase.io/download for older (and currently published) clients.

done and committed and deploying....

PR #18871, opened 2019-08-09
useSafeNavigation hook

buoyad:


shared/settings/account/add-modals.tsx
```diff
@@ -38,7 +39,7 @@ export const Email = () => {
     }
   }, [addEmailInProgress, dispatch, email, emailError])
 
-  const onClose = React.useCallback(() => dispatch(RouteTreeGen.createNavigateUp()), [dispatch])
+  const onClose = React.useCallback(() => dispatch(nav.safeNavigateUpPayload()), [dispatch, nav])
```
jbialobr: `nav` is recreated at each time the render is called - passing it to the dependencies array causes that `React.useCallback` returns always a new instance of the callback. The same goes for `dispatch`
I did a proposal for how it could be done: https://github.com/jbialobr/client/commit/102760d75f2aa938fe0bd3e538cf8ee073aa7c5b Feel free to cherry-pick the commit or any of its parts.

buoyad: Ah, thanks for pointing that out. However, `dispatch` seems to be equal across renders. I'll fix `nav` being recreated with `useMemo`.

PR #18874, opened 2019-08-09
Less printing by default

MarcoPolo:
I've been using this locally. Maybe we should reconsider all the default logging we have in dev mode? This is mostly just for discussion.

### DIFF
--- a/shared/local-debug.native.tsx
+++ b/shared/local-debug.native.tsx
@@ -56,7 +56,7 @@ let config = {
 
 // Developer settings
 if (__DEV__) {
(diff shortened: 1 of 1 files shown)


MarcoPolo:
@keybase/react-hackers Bringing this up again. Right now the default makes using logcat on android usesless without this patch.

buoyad:
Agreed with @chrisnojima, I use those the most frequently. The rest seem fine to make opt-in.

MarcoPolo:
Closing in favor of https://github.com/keybase/client/pull/20117 because it is better.

PR #18881, opened 2019-08-12
Make tsc-strict pass again

chrisnojima:
- [x] tsc scrict passes again
- [x] added all 
- [x] move override modules to base 
- [x] just explicitly marked some things as any so they're not 'implicit' anymore. we'll clean this stuff up slowly


shared/util/phone-numbers/index.tsx
```diff
@@ -55,7 +58,7 @@ countries.forEach(curr => {
 export const countryData: {[key: string]: CountryData} = countryDataRaw
 export const codeToCountry: {[key: string]: string} = codeToCountryRaw
 
-const canadianAreaCodes = {
+const canadianAreaCodes: {[kkey: string]: boolean} = {
```
buoyad: `kkey` -> `key`

PR #18895, opened 2019-08-12
Tweaks on Add button

cecileboucheron:
@keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/info-panel/add-people.tsx
+++ b/shared/chat/conversation/info-panel/add-people.tsx
@@ -35,7 +35,7 @@ const _AddPeople = (props: Props) => {
     )
   } else {
     directAction = props.onAddPeople
-    directLabel = 'Add members to team'
+    directLabel = 'Add people to team'
   }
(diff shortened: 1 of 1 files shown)


shared/chat/conversation/info-panel/add-people.tsx
```diff
@@ -92,7 +92,11 @@ const AddPeople = compose(
 )(_AddPeople) as any
 
 const styles = Styles.styleSheetCreate({
-  addButtonContainer: {marginLeft: Styles.globalMargins.small, marginRight: Styles.globalMargins.small},
+  addButtonContainer: {
+    marginBottom: Styles.globalMargins.small,
```
buoyad: cc @mmaxim do any height calculations need to change because of this?

cecileboucheron: @mmaxim ^?

mmaxim: Don't think so.

PR #18896, opened 2019-08-12
Hide header on wallet onboarding

nathunsmitty:


shared/wallets/onboarding/container.tsx
```diff
@@ -41,8 +41,15 @@ const ConnectedOnboarding = Container.connect(
 // A wrapper to harmonize the type of OwnProps between the
 // routed case and <Onboarding /> case.
 type RoutedOnboardingProps = Container.RouteProps<OwnProps>
-export const RoutedOnboarding = (ownProps: RoutedOnboardingProps) => (
-  <ConnectedOnboarding nextScreen={Container.getRouteProps(ownProps, 'nextScreen', 'openWallet')} />
-)
+export const RoutedOnboarding = (ownProps: RoutedOnboardingProps) => {
+  React.useEffect(() => {
+    // @ts-ignore
```
chrisnojima: use Container.getRouteProps

shared/wallets/onboarding/container.tsx
```diff
@@ -41,8 +41,15 @@ const ConnectedOnboarding = Container.connect(
...
+    if (!ownProps.navigation.getParam('hideHeader')) {
+      // @ts-ignore
+      ownProps.navigation.setParams({hideHeader})
```
chrisnojima: i don't understand what this is for. you're getting the value, then just settings it again?

nathunsmitty: I found without the above check, setting the params seems to cause an infinite render loop – I guess setting params triggers a rerender? I was a bit surprised by this because it wasn't mentioned in the [React Navigation example](https://reactnavigation.org/docs/en/redux-integration.html#setparams-from-your-screen) for setting params from a component.

shared/route-tree/render-route.tsx
```diff
@@ -3,7 +3,7 @@ export type RouteProps<T extends {[key: string]: any}> = {
   navigation: {
     getParam<K extends keyof T>(key: K): T[K] | undefined
     pop: () => void
-      isFirstRouteInParent: () => boolean
+    isFirstRouteInParent: () => boolean
```
nathunsmitty: Prettier fix

shared/wallets/routes.tsx
```diff
@@ -64,29 +72,58 @@ class WalletsSubNav extends React.PureComponent<NavigationViewProps<any>> {
...
+      headerMode: navigation.state.index === 0 ? 'none' : undefined,
+      headerRightActions: require('./nav-header/container').HeaderRightActions,
+      headerTitle: require('./nav-header/container').HeaderTitle,
+      title: 'Wallet',
+    }
+  }
+
+  componentDidMount() {
+    if (this.props.acceptedDisclaimer) {
+      this.props.navigation.navigate('wallet')
```
nathunsmitty: I'm not sure if it'd make more sense to dispatch one of our navigation actions here instead of using `navigate` directly?

shared/actions/wallets.tsx
```diff
@@ -1027,6 +1027,8 @@ const checkDisclaimer = (_: TypedState, action: WalletsGen.CheckDisclaimerPayloa
           } else {
             actions.push(RouteTreeGen.createNavigateAppend({path: [SettingsConstants.walletsTab]}))
           }
+        } else {
+          actions.push(RouteTreeGen.createNavigateAppend({path: [SettingsConstants.walletsTab, 'wallet']}))
```
nathunsmitty: I found that we need to explicitly navigate to the wallets page now after the disclaimer is accepted as a result of our routing changes.

shared/wallets/routes.tsx
chrisnojima: i think this is fine. its a router thing natively so it can use the native stuff

shared/actions/wallets.tsx
chrisnojima: i think if you do a componentDidUpdate call in _OnboardingOrWallets and navigate like didMount did, it'd work?

shared/wallets/routes.tsx
```diff
@@ -64,29 +72,67 @@ class WalletsSubNav extends React.PureComponent<NavigationViewProps<any>> {
...
+    }
+  }
+
+  componentDidUpdate() {
+    if (this.props.acceptedDisclaimer) {
```
chrisnojima: i'd only do this if the prop changes, this might get called a bunch

PR #18898, opened 2019-08-12
replace HOCTimers with hooks (part 1)

songgao:
There are still a few large ones (e.g. global errors, send indicator, fs upload counter, etc.) left; wondering if we should change them into hooks or just have simple wrappers for now.


shared/'
```diff
@@ -0,0 +1,127 @@
...
+const styles = Styles.styleSheetCreate({
+  progressIndicator: {
+    height: 22,
+    width: 22,
+  },
+  promptContainer: {
+    paddingLeft: Styles.globalMargins.medium,
+    paddingRight: Styles.globalMargins.medium,
+  },
+  toastText: {color: Styles.globalColors.white},
```
chrisnojima: ??????

songgao: Assuming you meant this file rather than this particular line, it's removed now.

songgao:
@chrisnojima bump

PR #18899, opened 2019-08-12
[Y2K-440] Welcome banner should not be closeable

jzila:
cc @keybase/y2ksquad

### DIFF
--- a/shared/people/index.shared.tsx
+++ b/shared/people/index.shared.tsx
@@ -2,6 +2,8 @@ import * as React from 'react'
 import * as Types from '../constants/types/people'
 import * as Kb from '../common-adapters'
 import * as Styles from '../styles'
+import * as Container from '../util/container'
+import * as SignupGen from '../actions/signup-gen'
 import Todo from './todo/container'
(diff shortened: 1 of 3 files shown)


shared/people/index.shared.tsx
```diff
@@ -57,23 +59,21 @@ export const itemToComponent: (item: Types.PeopleScreenItem, props: Props) => Re
   return null
 }
 
-const EmailVerificationBanner = ({email, clearJustSignedUpEmail}) => {
+const EmailVerificationBanner = ({email}) => {
+  const dispatch = Container.useDispatch()
+  React.useEffect(() => () =>
+    // Only have a cleanup function
+    dispatch(SignupGen.createClearJustSignedUpEmail())
```
buoyad: This cleanup function fires on every render, if you want it to run on unmount add the second param to `useEffect`.

https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

jzila:
@buoyad done

PR #18912, opened 2019-08-13
Don't close a modal if a click that ended outside the modal started inside it

cjb:

When we're showing a modal, we stop onClicks from inside the modal bubbling up to the "cover" opaque background.  Clicking on that background closes the modal.

But if a click starts inside the modal and ends on that background element, the onClick for the background fires.  So if you click and drag to select some text inside a modal but end outside of it, the modal closes.

Since onClick's not going to work, I went with using onMouseDown and onMouseUp and only treating it as a modal close if the click both started and ended inside the background element.

Am open to better ways of doing it, I was surprised I couldn't find anything more obvious to do.

### DIFF
--- a/shared/common-adapters/popup-dialog.desktop.tsx
+++ b/shared/common-adapters/popup-dialog.desktop.tsx
@@ -5,7 +5,7 @@ import {EscapeHandler} from '../util/key-event-handler.desktop'
 import * as Styles from '../styles'
 import {Props} from './popup-dialog'
 
-function stopBubbling(ev) {
+function stopBubbling(ev: React.MouseEvent) {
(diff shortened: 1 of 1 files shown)


chrisnojima:
Doesn't onclick have the location?

cjb:
@chrisnojima It has the end location, not the start location. We want to avoid firing the onClose depending on the start location.

chrisnojima:
relative-popup-hoc also needs to be fixed

chrisnojima:
here is my alternative implementation: https://github.com/keybase/client/pull/18930

chrisnojima:
alternative merged

PR #18917, opened 2019-08-13
filter conversations on way out for notifications

mmaxim:
Don't let conversations that will not show up in the UI out of notifications.


go/chat/convsource.go
```diff
@@ -867,20 +867,24 @@ func (s *HybridConversationSource) notifyExpunge(ctx context.Context, uid gregor
 	convID chat1.ConversationID, mergeRes storage.MergeResult) {
 	if mergeRes.Expunged != nil {
 		var inboxItem *chat1.InboxUIItem
+		var conv *chat1.ConversationLocal
+		var err error
```
joshblum: rm

mmaxim:
Actually this doesn't really work... I will need to re-tool.

PR #18920, opened 2019-08-13
add a wallet merge cli command

xgess:
pulls in new stellarnet command from here https://github.com/keybase/stellarnet/pull/74

looks like this:
```
keybase wallet merge GA6WYO...XTME
```
* optionally takes `--to`. this parameter can be any other account or another user, and it defaults to your primary account. 
* the from side can also be a secret key. if it's an address, then it needs to be in your wallet. but you can use this as a command to merge anything into anything. :)

notes:
* currently blocking the merge of your primary account. this seems like it would be more likely a mistake than anything else, but if you wanted to do it, just go into another account and do it from there. 
* if you want to merge away an account with assets, the target account also needs to have trustlines to those assets. i thought about automating this if it's going into one of your other accounts, but i didn't. it would either mean multiple transactions or a much more complicated one (multiple signers on the operations). 
* i added a single test here that exercises the simplest happy path (more is done over in stellarnet)
* i didn't add it to the `keybase wallet api`.

### DIFF
--- a/go/stellar/accountmerge.go
+++ b/go/stellar/accountmerge.go
@@ -9,32 +9,46 @@ import (
 	"github.com/keybase/stellarnet"
 )
 
-func AccountMerge(mctx libkb.MetaContext, walletState *WalletState, arg stellar1.AccountMergeCLILocalArg) (res stellarnet.SignResult, from stellar1.AccountID, err error) {
+func AccountMerge(mctx libkb.MetaContext, walletState *WalletState, arg stellar1.AccountMergeCLILocalArg) (res stellarnet.SignResult, err error) {
 	baseFee := walletState.BaseFee(mctx)
 	sp, unlock := NewClaimSeqnoProvider(mctx, walletState)
 	defer unlock()
 	tb, err := getTimeboundsForSending(mctx, walletState)
 	if err != nil {
-		return res, from, err
+		return res, err
 	}
 
-	bundleEntry, accountBundle, err := LookupSender(mctx, arg.From)
-	if err != nil {
-		return res, from, fmt.Errorf("account merge error looking up <from>: %v", err)
+	if arg.FromSecretKey == nil {
+		bundleEntry, accountBundle, err := LookupSender(mctx, arg.FromAccountID)
+		if err != nil {
+			return res, fmt.Errorf("account merge error looking up <from>: %v", err)
+		}
+		if len(accountBundle.Signers) == 0 {
+			return res, fmt.Errorf("secret key not found for %s", arg.FromAccountID)
+		}
+		if len(accountBundle.Signers) > 1 {
+			return res, fmt.Errorf("do not know how to handle multiple secret keys found for %s", arg.FromAccountID)
+		}
+		arg.FromSecretKey = &accountBundle.Signers[0]
+		arg.FromAccountID = bundleEntry.AccountID
+		mctx.Debug("account merge <from> lookup complete: %s", arg.FromAccountID)
 	}
-	fromSeed := stellarnet.SeedStr(accountBundle.Signers[0].SecureNoLogString())
-	mctx.Debug("account merge <from> lookup complete: %s -> %s", arg.From, bundleEntry.AccountID)
+	fromSeed := stellarnet.SeedStr(arg.FromSecretKey.SecureNoLogString())
+	mctx.Debug("account merge <from> seed lookup complete")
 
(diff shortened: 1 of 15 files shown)


go/stellar/stellar.go
```diff
@@ -1590,6 +1590,21 @@ func SetAccountAsPrimary(m libkb.MetaContext, walletState *WalletState, accountI
 	return walletState.UpdateAccountEntriesWithBundle(m, "set account as primary", &nextBundle)
 }
 
+func DeleteAccountIfEmpty(mctx libkb.MetaContext, remoter remote.Remoter, accountID stellar1.AccountID) error {
+	funded, err := isAccountFunded(mctx.Ctx(), remoter, accountID)
```
mlsteele: This uses caching when it shoudn't. `remoter` is a `WalletState`. This is where wallet state being sneaked into remoter is confusing. (I could have it wrong)

go/stellar/accountmerge.go
```diff
@@ -0,0 +1,40 @@
...
+	tb, err := getTimeboundsForSending(mctx, walletState)
+	if err != nil {
+		return res, from, err
+	}
+
+	bundleEntry, accountBundle, err := LookupSender(mctx, arg.From)
+	if err != nil {
+		return res, from, fmt.Errorf("account merge error looking up <from>: %v", err)
+	}
+	fromSeed := stellarnet.SeedStr(accountBundle.Signers[0].SecureNoLogString())
```
mlsteele: Check Signers array length. Might it be zero for mobile-only accounts? Even if not it'd be nice future proofing

go/stellar/accountmerge.go
```diff
@@ -0,0 +1,40 @@
...
+	mctx.Debug("account merge <from> lookup complete: %s -> %s", arg.From, bundleEntry.AccountID)
+
+	recipient, err := LookupRecipient(mctx, stellarcommon.RecipientInput(arg.To), true /* isCLI */)
+	if err != nil {
+		return res, from, fmt.Errorf("account merge error looking up <to>: %v", err)
+	}
+	toAddr := recipient.AccountID
+	mctx.Debug("account merge <to> lookup complete: %s -> %s", arg.To, toAddr)
+
+	signedTx, err := stellarnet.AccountMergeTransaction(fromSeed, *toAddr, sp, tb, baseFee)
```
mlsteele: toAddr can be nil in case of a username that doesn't have a wallet

go/stellar/stellar.go
xgess: hah. so, i noticed that, and there's a wallet state refresh right before this is called. but yes, i agree. i'd rather just not cache. do you think it's better to pass in the wallet state (instead of remoter) and refresh it in here? or is there something cleaner?

go/stellar/accountmerge.go
xgess: agreed

go/stellar/accountmerge.go
xgess: oh. interesting. will fix.

go/stellar/stellar.go
mlsteele: Ah nice.

1. I think ideally `WalletState` wouldn't implement `Remoter`. But that's probably too much trouble. 
2. `Server` has `.remoter` which is actually WalletState and `.walletState` which is the same (weird). Could make `.remoter` an actual remoter and change existing uses of `.remoter` over to `.walletState`.
3. Or add `.remoterActual` and pass that into DeleteAccountIfEmtpy.
4. Or add a boolean to isAccountFunded to punch through caches.

Whatever you want to do including leaving at as is if the refresh works is cool with me. But if you leave just the refresh how about moving the Refresh call down into `DeleteAccountIfEmpty`? So that a potential future caller of `DeleteAccountIfEmpty` doesn't get a cache surprise 💸 .

xgess: pretty sure this is fixed in a nice way now.

go/stellar/stellarsvc/service.go
```diff
@@ -220,6 +220,19 @@ func (s *Server) SendCLILocal(ctx context.Context, arg stellar1.SendCLILocalArg)
 	}, nil
 }
 
+func getPrimaryAccount(mctx libkb.MetaContext, remoter remote.Remoter) (acct stellar1.WalletAccountLocal, err error) {
```
patrickxb: there is `stellar.GetOwnPrimaryAccountID()`...would that work instead of this?

xgess: lol. i looked all over for this. i assumed it had to exist. yes. much better. thank you.

PR #18926, opened 2019-08-13
[Y2K-416] Only show the verify email todo after signup

jzila:
cc @keybase/y2ksquad


jzila:
@pzduniak PTAL

pzduniak:
Have you discussed this with design? I think it's OK as-is with the stateful content. PR itself is :+1:.

jzila:
@pzduniak the ticket has these requirements:
> At the end of the sign up, the people tab should show a green welcome/verify email banner. 
> The verify to-do item should only appear after you leave the tab and come back

PR #18930, opened 2019-08-14
dialog click proposal

chrisnojima:


chrisnojima:
@cjb so should i merge this instead?

Issue #18925, opened 2019-08-13
keybase:// url scheme does not support electron path-to-app invocation


eli-schwartz:
The feature introduced in #18555 does not function in my Arch Linux package for keybase-gui, because keybase is not invoked via a renamed electron binary running the default app in resources/app/ -- instead, it is invoked via an electron binary in /usr/bin/electron /path/to/app

```
$ /usr/bin/electron
Electron 6.0.1 - Build cross platform desktop apps with JavaScript, HTML, and CSS
Usage: electron [options] [path]

A path to an Electron app may be specified. It must be one of the following:
  - index.js file.
  - Folder containing a package.json file.
  - Folder containing an index.js file.
  - .html/.htm file.
  - http://, https://, or file:// URL.
```

(and in the GUI)
> To run a local app, execute the following on the command line:
> `/usr/lib/electron/electron path-to-app`

This is generally useful functionality, and in my case, it's used to ensure that all electron apps use the same runtime, with the system ffmpeg/icu/http_parser/libevent/fontconfig/etc/more/things, and built using the distro-approved clang and relevant CFLAGS.

However, keybase seems to be looking for command-line options exclusively in argv1 (which is, in the case of `electron path-to-app`, going to be "path-to-app") and not seeing the keybase:// url at all. It would be nice to support this use case.

eli-schwartz:
For context, this is probably caused by the parsing done here:
https://github.com/keybase/client/blob/8283633e9626ae34a29792b0f6695e507702de8a/shared/desktop/app/node.desktop.tsx#L75-L81
https://github.com/keybase/client/blob/8283633e9626ae34a29792b0f6695e507702de8a/shared/desktop/app/node.desktop.tsx#L164-L166

heronhaye:
Merged. Do you need us to do anything else or will you cherry pick for this release?

eli-schwartz:
I'm fine with cherry-picking it. Thanks!

PR #18932, opened 2019-08-14
allow argv2 for deeplinks

heronhaye:
Addresses https://github.com/keybase/client/issues/18925.

@eli-schwartz, @taigrr, please check out this PR.

### DIFF
--- a/shared/desktop/app/node.desktop.tsx
+++ b/shared/desktop/app/node.desktop.tsx
@@ -77,7 +77,7 @@ const focusSelfOnAnotherInstanceLaunching = (_, commandLine) => {
     // Allow both argv1 and argv2 to be the link to support "/usr/lib/electron/electron path-to-app"-style
     // invocations (used in the Arch community packages).
     for (let link of commandLine.slice(1, 3)) {
-      if (link.startsWith('web+stellar:') || link.startsWith('keybase://')) {
+      if (isRelevantDeepLink(link)) {
         sendToMainWindow('dispatchAction', {payload: {link}, type: DeeplinksGen.link})
         return
       }
@@ -146,6 +146,10 @@ const handleCrashes = () => {
   }
 }
 
+const isRelevantDeepLink = x => {
+  return x.startsWith('web+stellar:') || x.startsWith('keybase://')
+}
+
 const createMainWindow = () => {
   mainWindow = MainWindow()
   tellMainWindowAboutMenubar()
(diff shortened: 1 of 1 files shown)


eli-schwartz:
I rebuilt my distro package for keybase/kbfs/keybase-gui 4.3.1 with commit c8f65d0e2fd868327c8f275fc32177e6d1cb089f applied. Then I tried to start my keybase-gui wrapper script with a keybase:// link, and it worked as expected. So this PR looks good to me. Thanks!

heronhaye:
@eli-schwartz I noticed there's another codepath here. Technically, this should work even if Keybase isn't running already. But I was unable to figure out how to pass in the argument to the electron binary from run_keybase (or in your case keybase-gui) when starting the GUI via systemd. So this flow doesn't work, but it will if you manually run `electron path-to-app keybase://xxx`. 

In any case, I added the keybase:// handler and argv2 handling to that flow as well. If you manage to figure out how to implement it, let us know!

shared/desktop/app/node.desktop.tsx
```diff
@@ -161,9 +169,17 @@ const createMainWindow = () => {
       // stash a startupURL to be dispatched when we're ready for it.
       sendToMainWindow('dispatchAction', {payload: {link: startupURL}, type: DeeplinksGen.link})
       startupURL = null
-    } else if (!isDarwin && process.argv.length > 1 && process.argv[1].startsWith('web+stellar:')) {
+    } else if (!isDarwin) {
       // Windows and Linux instead store a launch URL in argv.
-      sendToMainWindow('dispatchAction', {payload: {link: process.argv[1]}, type: DeeplinksGen.link})
+      let link: string | null = null
```
cjb: I think `let link = ''` is fine, it'll be falsey if empty

eli-schwartz:
> I noticed there's another codepath here. Technically, this should work even if Keybase isn't running already. But I was unable to figure out how to pass in the argument to the electron binary from run_keybase (or in your case keybase-gui) when starting the GUI via systemd. So this flow doesn't work when clicking links on a website, but it will if you manually run `electron path-to-app keybase://xxx`, regardless if the GUI is already up or not.

I tested with the debian prebuilt binary, version 4.3.0-20190812143918+f8de90a10f -- and keybase://chat/username only worked when invoking 
```
/opt/keybase/Keybase &
/opt/keybase/Keybase keybase://chat/username
```

So I assumed it wasn't implemented for initial startup at all. If it is implemented for that, then, well, I can just let arguments mean "invoke directly instead of with systemd".

heronhaye:
There was a bug in the prebuilt binary where it was only accepting web+stellar links but not keybase links - that is fixed in the second commit I added.

eli-schwartz:
Oh, so if you were fixing it for the prebuilt binary too then that makes sense. :)

And indeed, I updated my distro build to use the combined patch on master, and it works for the first start. And nuking /opt/keybase/resources/app and cp'ing /usr/share/keybase-app into /opt/keybase/resources/app allows the (now frankensteined) debian prebuilt binary to accept links on startup too.

Thanks for clarifying.

heronhaye:
Great, sounds good. Just to be clear, that won't work for links you click in your browser like keybase://public/eschwartz, right? Since the registered handler is the desktop file `/usr/share/applications/keybase.desktop`, which I presume runs `keybase-gui %u`?

eli-schwartz:
It does work in Firefox, since the keybase-gui script will, if it sees arguments, simply execute `electron /usr/share/keybase/ "$@"`.

PR #18942, opened 2019-08-14
add "Saved" toast on mobile for attachment saves

songgao:


shared/constants/chat2/message.tsx
```diff
@@ -1266,3 +1266,16 @@ export const messageExplodeDescriptions: Types.MessageExplodeDescription[] = [
   {seconds: 86400 * 7, text: '7 days'},
   {seconds: 0, text: 'Never explode (turn off)'},
 ].reverse()
+
+export const messageAttachmentTransferStateToProgressLabel = (
+  transferState: Types.MessageAttachmentTransferState
+): string =>
+  transferState === 'downloading'
```
chrisnojima: better as a switch

PR #18943, opened 2019-08-14
Fix up prettier-check, prettier-write-all, prettierignore

cjb:
@keybase/react-hackers 

This PR's best reviewed as separate commits -- 9885547 for infrastructure and 4f11900 as the output of prettier-write-all.

Our prettier setup had some bitrot:

* there's no CI integration, so we're depending on everyone running pre-commit, and sometimes they don't.
* there's a prettier-check helper, but that doesn't check for unformatted files, it's a correctness check.  Added prettier-check for checking formatting and prettier-debug-check for the correctness check.
* the .prettierignore file is pre-TS, needed updates
* prettier-write-all was failing because the output buffer from git ls-files was too large -- there are 10000s of filenames under ios/Pods that weren't there before.  Changed to use a larger buffer.
* finally, run prettier-write-all over everything to catch the misformattings that were committed before the CI integration added by this PR.


chrisnojima:
👍

PR #18946, opened 2019-08-14
misc teambot fixes

joshblum:
patch does the following:

chat: cc @mikem
- expands test coverage for teambots including bots sending
- fixes key finding bugs for teambots

teams: cc @mlsteele 
- correctly posts bot_settings link for change_membership
- clears in locally stored secrets if team role is changed to restricted bot
- boxes keys if role is changed from restricted bot

depends on https://github.com/keybase/keybase/pull/4185

### DIFF
--- a/go/teams/service_helper.go
+++ b/go/teams/service_helper.go
@@ -1575,9 +1575,9 @@ func GetKBFSTeamSettings(ctx context.Context, g *libkb.GlobalContext, isPublic b
 
 func CanUserPerform(ctx context.Context, g *libkb.GlobalContext, teamname string) (ret keybase1.TeamOperation, err error) {
 	team, err := Load(ctx, g, keybase1.LoadTeamArg{
-		Name:    teamname,
-		StaleOK: true,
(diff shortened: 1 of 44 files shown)


go/teams/service_helper.go
```diff
@@ -648,22 +648,35 @@ func AddEmailsBulk(ctx context.Context, g *libkb.GlobalContext, teamname, emails
 
 func EditMember(ctx context.Context, g *libkb.GlobalContext, teamname, username string,
 	role keybase1.TeamRole, botSettings *keybase1.TeamBotSettings) error {
+	team, err := Load(ctx, g, keybase1.LoadTeamArg{
+		Name:        teamname,
+		ForceRepoll: true,
```
mlsteele: This adds one extra RPC to EditMember. It would be possible to avoid by for example passing `*Team` down into helpers after it's loaded.

joshblum: will address in the next PR and drop the extra loads with new RPCs for impteam changes

PR #18951, opened 2019-08-15
immer: waiting

chrisnojima:
@keybase/react-hackers uses immer for the waiting store


shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -353,6 +353,8 @@ function* initializeUseNativeFrame() {
   )
 }
 
+export const requestLocationPermission = () => Promise.resolve()
```
chrisnojima: fixes a missing export in desktop

shared/chat/conversation/info-panel/container.tsx
```diff
@@ -78,8 +78,10 @@ const mapStateToProps = (state: Container.TypedState, ownProps: OwnProps) => {
     selectedConversationIDKey: conversationIDKey,
     selectedTab,
     smallTeam: meta.teamType !== 'big',
-    spinnerForHide:
-      state.waiting.counts.get(Constants.waitingKeyConvStatusChange(ownProps.conversationIDKey), 0) > 0,
+    spinnerForHide: Container.anyWaiting(
```
chrisnojima: use waiting helper

shared/native/logger.desktop.tsx
```diff
@@ -1,11 +1,11 @@
...
   console.error('not supported on desktop!')
 }
-const dump: NativeLogDump = () => {
+export const dump: NativeLogDump = () => {
   const p: Promise<Array<string>> = Promise.reject(new Error('Not supported on desktop!'))
   return p
 }
 
-export {log, dump}
+export const flush = () => {}
```
chrisnojima: fix missing export

shared/tracker2/remote-proxy.desktop.tsx
```diff
@@ -35,7 +35,7 @@ const trackerMapStateToProps = (state, ownProps) => {
     registeredForAirdrop: d.registeredForAirdrop,
     state: d.state,
     teamShowcase: d.teamShowcase,
-    waiting: state.waiting.counts.get(Constants.waitingKey) || 0,
+    waiting: Container.anyWaiting(state, Constants.waitingKey),
```
chrisnojima: use helper

shared/reducers/waiting.tsx
```diff
@@ -1,71 +1,81 @@
...
 }
 
-export default reducer
+const getKeys = (k: string | Array<string>) => {
+  if (typeof k === 'string') return [k]
+  return k
+}
+
+export default (state: Types.State = initialState, action: Waiting.Actions): Types.State =>
+  Container.produce(state, (draftState: Container.Draft<Types.State>) => {
```
buoyad: Can we skip the type annotations on this function? Testing locally TS knows what it is without it.

buoyad:
I also noticed this in the immer docs. Should we turn it off in prod?
https://github.com/immerjs/immer#auto-freezing

chrisnojima:
? ' By default, it is turned on during local development and turned off in production

shared/reducers/waiting.tsx
chrisnojima: i think we should leave it. while building this out w/o it you can easily invalidate this and then it gets very confusing. this way you'll get an error that's easy to debug

chrisnojima:
https://github.com/immerjs/immer/blob/a16dedad9ce68ff996105b1b7393b0e0b52a45db/src/immer.js#L22

buoyad:
Ah, k cool. Was confused by "it is recommended to disable this option in production"

PR #18953, opened 2019-08-15
common-adapters (A-F) dark mode support

hyperobject:
This PR updates the `common-adapters` components so they can handle theme changes. Most of these changes are just using a closure in calls to `Styles.styleSheetCreate` instead of an object, but there were a few cases where I needed to move other style declarations into a `styleSheetCreate` call.

@keybase/react-hackers

### DIFF
--- a/shared/common-adapters/button.css
+++ b/shared/common-adapters/button.css
@@ -4,23 +4,38 @@
 .button__border {
   border: 1px solid rgba(0, 0, 0, 0.2);
 }
+.darkMode .button__border {
+  border: 1px solid rgba(255, 255, 255, 0.2);
+}
 .button__border_blue:hover {
   border: 1px solid rgba(77, 142, 255, 0.3);
+}
+.darkMode .button__border_blue:hover {
+  border: 1px solid rgba(178, 113, 0, 0.3);
 }
 .button__border_red:hover {
   border: 1px solid rgba(255, 77, 97, 0.3);
 }
 .darkMode .button__border_red:hover {
-  border: 1px solid rgba(255, 255, 255, 0.3);
+  border: 1px solid rgba(0, 178, 158, 0.3);
 }
 .button__border_green:hover {
   border: 1px solid rgba(55, 189, 153, 0.3);
 }
+.darkMode .button__border_green:hover {
+  border: 1px solid rgba(200, 66, 102, 0.3);
+}
 .button__border_purple:hover {
   border: 1px solid rgba(112, 78, 186, 0.3);
 }
+.darkMode .button__border_purple:hover {
+  border: 1px solid rgba(143, 177, 69, 0.3);
+}
 .button__border_black:hover {
   border: 1px solid rgba(0, 0, 0, 0.3);
+}
+.darkMode .button__border_black:hover {
+  border: 1px solid rgba(255, 255, 255, 0.3);
 }
 
 .button__underlay {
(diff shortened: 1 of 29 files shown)


shared/common-adapters/button.css
```diff
@@ -4,24 +4,39 @@
...
   border: 1px solid rgba(0, 0, 0, 0.2);
 }
+.darkMode .button__border {
+  border: 1px solid rgba(255, 255, 255, 0.2);
+}
 .button__border_blue:hover {
   border: 1px solid rgba(77, 142, 255, 0.3);
 }
+.darkMode .button__border_blue:hover {
+  border: 1px solid rgba(178, 113, 0, 0.3);
```
jakob223: It feels a little silly to me to be specifying inverse colors here when I assume we're going to be going back and replacing them with different shades of the original color. But it's like 5 lines so /shrug

shared/common-adapters/choice-list.desktop.tsx
```diff
@@ -11,17 +12,17 @@ const ChoiceList = ({options}: Props) => {
         // TODO is this okay?
         const iconType: IconType = op.icon as IconType
         return (
-          <Box style={styleEntry} key={idx} className="cl-entry" onClick={() => op.onClick()}>
-            <Box style={styleIconContainer} className="cl-icon-container">
+          <Box style={Styles.collapseStyles([Styles.globalStyles.flexBoxRow, Styles.desktopStyles.clickable, styles.entry])} key={idx} className="cl-entry" onClick={() => op.onClick()}>
```
jakob223: Can you make these into `Box2`s (which are flexbox automatically, among other things) while you're here?

shared/common-adapters/choice-list.desktop.tsx
```diff
@@ -50,41 +54,36 @@ const rawCSS = `
...
-  ...globalStyles.flexBoxColumn,
-  alignItems: 'flex-start',
-  flex: 1,
-  justifyContent: 'center',
-  marginLeft: globalMargins.small,
-  textAlign: 'left',
-}
+const styles = Styles.styleSheetCreate(() => ({
+  entry: {
+    padding: `${Styles.globalMargins.tiny}px ${Styles.globalMargins.small}px`,
```
jakob223: we have a helper `Styles.padding(up, right?, bottom?, left?)` for this

jakob223: (so that you don't need to do string interpolation)

shared/common-adapters/choice-list.desktop.tsx
```diff
@@ -37,7 +38,10 @@ const rawCSS = `
     background-color: transparent;
   }
   .cl-entry:hover {
-    background-color: ${globalColors.blueLighter2};
```
chrisnojima: let's actually move this to a real css import instead of this inline css thing. Also i think we should hide the existence of darkColors from all code

jakob223:
Oh, oops, sorry nojima I had started reviewing by the time you self-requested so I didn't see it :|

shared/common-adapters/choice-list.desktop.tsx
chrisnojima: let's not do this. we're going to touch hundreds of files and we can't stop and refactor things, especially the common-adapters where it could break a ton of components

chrisnojima:
multiple reviewers is fine

shared/common-adapters/button.css
hyperobject: I went back and forth on that myself, but I figured that since we'll have to put this dark mode css logic in place anyway, I might as well put it in while I was in the file.

shared/common-adapters/choice-list.desktop.tsx
hyperobject: Done ✔️

shared/desktop/renderer/style.css
```diff
@@ -4,6 +4,7 @@
 @import '../../common-adapters/icon.css';
 @import '../../common-adapters/text.css';
 @import '../../common-adapters/button.css';
+@import '../../common-adapters/choice-list.css';
```
chrisnojima: so the original reason these were here was because in some cases (like button.tsx) we had a unified implementation so we couldn't just `import './button.css'` in the file (React native packager didn't like that). But, I've since added the capability to do that safely in a cross platform way so
tldr;
we should remove these imports from here and instead just import them in the `.tsx` files instead

hyperobject:
mobile looks good to me!

chrisnojima:
:shipit:

PR #18957, opened 2019-08-15
teams: fixes around hidden rotates and subteam renames

maxtaco:
- there was an unlikely test flake in TeamAPI systest that triggered a failure to upload a team
- introduce new tests that repro this and related problems 100% of the time
- fix, which is that rename sigs weren't ratcheting
- also, we have a case where >1 ratchets need to be uploaded, so fix the server to allow that

### DIFF
--- a/go/teams/loader2.go
+++ b/go/teams/loader2.go
@@ -999,7 +999,6 @@ func (l *TeamLoader) computeSeedChecks(ctx context.Context, state *keybase1.Team
 	defer mctx.Trace(fmt.Sprintf("TeamLoader#computeSeedChecks(%s)", state.ID()), func() error { return err })()
 
 	latestChainGen := keybase1.PerTeamKeyGeneration(len(state.PerTeamKeySeedsUnverified))
-	mctx.Debug("FUCK %d %+v", latestChainGen, state.PerTeamKeySeedsUnverified)
 	err = computeSeedChecks(
 		ctx,
 		state.ID(),
--- a/go/teams/loader2.go
+++ b/go/teams/loader2.go
@@ -594,6 +594,8 @@ func (l *TeamLoader) checkParentChildOperations(ctx context.Context,
 		forceFullReload:                       false,
 		forceRepoll:                           false,
 		staleOK:                               true, // stale is fine, as long as get those seqnos.
+		skipSeedCheck:                         true,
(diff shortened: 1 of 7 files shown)


maxtaco:
NEeds https://github.com/keybase/keybase/pull/4187 to pass CI

go/teams/loader2.go
```diff
@@ -997,7 +999,8 @@ func (l *TeamLoader) computeSeedChecks(ctx context.Context, state *keybase1.Team
 	defer mctx.Trace(fmt.Sprintf("TeamLoader#computeSeedChecks(%s)", state.ID()), func() error { return err })()
 
 	latestChainGen := keybase1.PerTeamKeyGeneration(len(state.PerTeamKeySeedsUnverified))
-	return computeSeedChecks(
+	mctx.Debug("FUCK %d %+v", latestChainGen, state.PerTeamKeySeedsUnverified)
```
maxtaco: whoops

go/teams/rename.go
```diff
@@ -135,6 +138,15 @@ func generateRenameSubteamSigForParentChain(m libkb.MetaContext, me libkb.UserFo
 		Entropy: entropy,
 	}
 
+	ratchet, err := parentTeam.makeHiddenRatchet(m)
```
heronhaye: why do we need to update parent hidden ratchets for rename subteam?

maxtaco: Whenever we post anything to a sigchain, the server requires that the "ratchets" are up-to-date. So in my test, I rotate the parent chain, then try to rename a child, and the server is checking that we know about the rotation (via the ratchet) before it accepts the new rename link in the parent chain.

maxtaco:
Thanks! There's this mechanism in `team_total_order` to make sure that the updates to a team, across the two chains, are fully ordered. And it caught the fact we weren't properly pointing to a hidden rotation.

heronhaye:
do you know why it was flaky as opposed to always working or always failing?

maxtaco:
It would fail in the original test only 1% of the time since there was a race between the rename and rekey that the rename almost always won.  But when the rekey won, the rename would fail.

maxtaco:
Ci is borked, because master won't update. I'm just going to merge this

PR #18958, opened 2019-08-15
center modal header when text

jakob223:
Issue: Y2k-491
I think I broke this when I centered it on mobile..

cc @keybase/y2ksquad

### DIFF
--- a/shared/common-adapters/modal/index.tsx
+++ b/shared/common-adapters/modal/index.tsx
@@ -69,7 +69,6 @@ Modal.defaultProps = {
   mode: 'Default',
 }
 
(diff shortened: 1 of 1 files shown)


shared/common-adapters/modal/index.tsx
```diff
@@ -88,7 +87,7 @@ const Header = (props: HeaderProps) => (
       </Kb.Box2>
       <Kb.Box style={styles.headerCenter}>
         {typeof props.title === 'string' ? (
-          <Kb.Text type="Header" lineClamp={1}>
+          <Kb.Text type="Header" lineClamp={1} style={styles.centerHeader}>
```
buoyad: Use `center={true}`

PR #18961, opened 2019-08-15
System Message on SBS resolution

jakob223:
The frontend part of this is ~done, but the message isn't actually being injected; rather than spending a bunch of time debugging a bunch of chat code I'm not familiar with, @mmaxim is there anything obviously wrong there?

Also, if you could do a preliminary review so that any large scale changes can happen before I go back and clean stuff up, that would be great 🙇 .


jakob223:
@mmaxim bump

jakob223:
Oh, I realized you're OOO this week, this can wait.

mmaxim:
K you should request a review from me on the right side, I might ignore anything else.

jakob223:
bump @mmaxim

go/teams/chatmessaging.go
```diff
@@ -96,6 +96,37 @@ func SendChatInviteWelcomeMessage(ctx context.Context, g *libkb.GlobalContext, t
...
+	proverName, err := g.GetUPAKLoader().LookupUsername(ctx, prover)
+	if err != nil {
+		g.Log.CDebugf(ctx, "SendChatSBSResolutionMessage: failed to lookup invitee username: %s", err)
+		return false
+	}
+
+	subBody := chat1.NewMessageSystemWithSbsresolve(chat1.
+		MessageSystemSbsResolve{
+		Prover:    proverName.String(),
+		Assertion: assertion,
```
zapu: might want to parse the assertion here and pass serviceId / serviceUsername to the frontend

go/teams/handler.go
```diff
@@ -443,7 +444,13 @@ func handleSBSSingle(ctx context.Context, g *libkb.GlobalContext, teamID keybase
 		}
 
 		// Send chat welcome message
-		if !team.IsImplicit() {
+		if team.IsImplicit() {
+			assertion := fmt.Sprintf("%s@%s", string(invite.Name), ityp)
```
zapu: this is no good. what about email?

jakob223: fixed!

go/teams/handler.go
```diff
@@ -385,10 +385,11 @@ func handleSBSSingle(ctx context.Context, g *libkb.GlobalContext, teamID keybase
 		if err != nil {
 			return err
 		}
+		var ityp string
 		switch category {
 		case keybase1.TeamInviteCategory_SBS:
 			//  resolve assertion in link (with uid in invite msg)
-			ityp, err := invite.Type.String()
+			ityp, err = invite.Type.String()
```
zapu: `ityp` will not be populated for EMAIL/PHONE, maybe we can move this code to outside of `switch` (before) ?

zapu: (it's used in `SendChatSBSResolutionMessage` call)

shared/chat/conversation/messages/system-sbs-resolve/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const SBSProvedNotice = (props: Props) => {
+  const {timestamp, prover, assertionUsername, assertionService} = props.message
+  const isYou = props.you === props.message.prover
+  return (
+    <Kb.Box2 direction="vertical" fullWidth={true}>
+      <Kb.Text type="BodyTiny">{formatTimeForChat(timestamp)}</Kb.Text>
+      <Kb.Text type="BodySmall">
+        {isYou ? 'You' : <Kb.ConnectedUsernames {...connectedUsernamesProps} usernames={[prover]} />} proved{' '}
+        {formatAssertion(assertionUsername, assertionService, isYou)}, so now {isYou ? 'you' : 'they'} can see
```
zapu: lets make the text similar to the `(m MessageSystem) String()`. "proved their phone number xxx" doesn't match our phone/email vocabulary. User is "proving" social services, but "verifying" phone numbers and emails.

go/teams/handler.go
```diff
@@ -445,6 +448,22 @@ func handleSBSSingle(ctx context.Context, g *libkb.GlobalContext, teamID keybase
...
+			}
+			uid := gregor1.UID(g.ActiveDevice.UID().ToBytes())
+			conv, err := g.ChatHelper.NewConversation(ctx, uid, iteamName, nil,
+				chat1.TopicType_CHAT, chat1.ConversationMembersType_IMPTEAMNATIVE, keybase1.TLFVisibility_PRIVATE)
+			if err != nil {
+				return err
+			}
+			convID = conv.GetConvID()
+		}
+
```
zapu: let's factor this out to some `struct ImpteamConversationRef { convID, impteamName }` that lives in another file (`go/teams/chatmessaging.go` is probably cool for now), this function does too many things already.

so we can have this almost 1-liner here:
```
// Save conversation reference now before we resolve the team invite,
// because if there are already implicit teams with same display name as our
// "target" name, we won't be able to find the conversation afterwards.
impteamConvRef := GetImpteamConversationRef(ctx, g, team)
```
and then pass it to `SendChatSBSResolutionMessage`.

jakob223:
merging over docker errors that are happening on master too

PR #18964, opened 2019-08-15
Better alphabet index in search recs

buoyad:
- Factor out into `AlphabetIndex`
- Handle touch events on the parent containing the letters; calculate where to scroll
  - Allows swiping up and down to choose section
- Animate scroll

cc @keybase/y2ksquad


buoyad:
@jzila hm, ok

- [ ] Test android

buoyad:
Ok @jzila it should work on android now, PTAL

PR #18965, opened 2019-08-16
let service handle GUI log writes and rotations

songgao:


go/status/log_send.go
```diff
@@ -59,6 +59,7 @@ type LogSendContext struct {
 	kbfsLog          string
 	svcLog           string
 	ekLog            string
+	guiLog           string
```
joshblum: why do we need both `guiLog` and `desktopLog`?

go/libkb/constants.go
```diff
@@ -576,10 +576,11 @@ const (
 const (
 	ServiceLogFileName = "keybase.service.log"
 	EKLogFileName      = "keybase.ek.log"
+	GUILogFileName     = "keybase.gui.log"
 	KBFSLogFileName    = kbconst.KBFSLogFileName
 	GitLogFileName     = "keybase.git.log"
 	UpdaterLogFileName = "keybase.updater.log"
-	DesktopLogFileName = "Keybase.app.log"
+	DesktopLogFileName = "Keybase.app.log" // deprecated; use GUILogFileName
```
joshblum: why not use keep the same filename?

songgao: I was afraid I'd mess it up. But I guess it's not a huge deal to lose some logs here .. will update to use the same name.

songgao: actually I do want to lowercase that. So to avoid confusing maybe it's best to keep the `.gui` part?

songgao: ended up keeping `Keybase.app.log`

songgao:
@joshblum Thanks! ready for another look

go/bind/keybase.go
```diff
@@ -285,7 +285,11 @@ func LogSend(statusJSON string, feedback string, sendLogs, sendMaxBytes bool, ui
 	logSendContext.InstallID = env.GetInstallID()
 	logSendContext.StatusJSON = statusJSON
 	logSendContext.Feedback = feedback
-	logSendContext.Logs.Desktop = uiLogPath
+	if len(uiLogPath) == 0 {
```
joshblum: this should be `!=` right? but i think we should just drop this param entirely and let the service run the show

songgao: oops yes. cool will do!

songgao:
OK ready for another look @joshblum thanks! There was actually an issue with Android; fixed by using the passed-in path like the ekLog.

songgao:
@chrisnojima Yo! This still need a review on JS side. Mind taking a look when you have time? Thanks!

PR #18972, opened 2019-08-16
attachment zoom changes

chrisnojima:
- [x] double tap to reset
- [x] constrain zoom
- [x] naive constrain pan (not perfect, but something for now)
- [x] swipe left/right to change go to next attachments

### DIFF
--- a/shared/chat/conversation/attachment-fullscreen/container.tsx
+++ b/shared/chat/conversation/attachment-fullscreen/container.tsx
@@ -21,33 +21,20 @@ const mapStateToProps = (state: Container.TypedState) => {
 }
 
 const mapDispatchToProps = (dispatch: Container.TypedDispatch) => ({
-  _onDownloadAttachment: (message: Types.MessageAttachment) => {
-    dispatch(
-      Chat2Gen.createAttachmentDownload({
-        message,
-      })
-    )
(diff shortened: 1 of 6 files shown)


shared/chat/conversation/attachment-fullscreen/container.tsx
```diff
@@ -63,14 +49,15 @@ const Connected = Container.connect(
     )
     return {
       autoPlay: stateProps.autoPlay,
-      hotkeys: ['left', 'right'],
```
chrisnojima: we shouldn't have our connectors make these super specific props for these hocs (esp. desktop only and not in the actual prop def)

shared/local-debug.native.tsx
```diff
@@ -6,7 +6,7 @@ import {noop} from 'lodash-es'
 const nativeBridge = NativeModules.KeybaseEngine || {test: 'fallback'}
 
 // Uncomment this to disable yellowboxes
-// console.disableYellowBox = true
```
mmaxim: These changes should get 🔥

chrisnojima:
i'm going to merge this. after approval i removed the changes to local-debug and tweaked the maxZoom value (we can do a better job of this later). also protected passing a null image on swipe

PR #18977, opened 2019-08-16
Add ability to start a conversation directly with a phone number

ddworken:
Routes an RPC up from the go client to be used for querying phone numbers. Creates a new tab to search for phone numbers and uses the RPC to do an exact match search.


shared/signup/phone-number/phone-input.tsx
```diff
@@ -220,6 +220,7 @@ type Props = {
   onChangeNumber: (phoneNumber: string) => void
   onChangeValidity: (valid: boolean) => void
   onEnterKeyDown?: () => void
+  result?: React.ReactNode
```
ddworken: The design calls for having the user appear aligned with the phone number input box. Rather than using a ref, this was suggested in #frontend.

protocol/bin/enabled-calls.json
```diff
@@ -118,6 +118,7 @@
   "keybase.1.config.setProxyData": {"promise": true},
   "keybase.1.config.getProxyData": {"promise": true},
   "keybase.1.contacts.getContactsForUserRecommendations": {"promise": true},
+  "keybase.1.contacts.lookupContactList": {"promise": true},
```
nathunsmitty: I think @zapu said we want to use `userSearch.userSearch` here

ddworken: Yup, working on migrating it over now. Didn't get to it on friday because I was working on the ssh ca stuff.

PR #18979, opened 2019-08-16
fix style overriding in markdown-memo

xgess:
the bug here was that style overriding wasn't actually working on mobile in the wallet transaction markdown memo (when it's an airdrop payment, the memo on the payment is supposed to be purple). it was purple on desktop but not mobile.

### DIFF
--- a/shared/wallets/common/markdown-memo.tsx
+++ b/shared/wallets/common/markdown-memo.tsx
@@ -4,7 +4,7 @@ import * as Styles from '../../styles'
 import {StyleOverride} from '../../common-adapters/markdown'
 import {isMobile} from '../../constants/platform'
 
-const styleOverride = Styles.styleSheetCreate({
+const styleOverride: StyleOverride = {
   del: {
     color: Styles.globalColors.black,
   },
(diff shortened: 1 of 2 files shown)


xgess:
@keybase/react-hackers ^

xgess:
> It should work to just pass it and not destructure at all

i can't seem to get this to work. are you sure / how were you thinking?

shared/wallets/common/markdown-memo.tsx
```diff
@@ -40,7 +40,7 @@ const MarkdownMemo = (props: Props) =>
       {!props.hideDivider && <Kb.Divider vertical={true} style={styles.quoteMarker} />}
       <Kb.Markdown
         style={styles.memo}
-        styleOverride={Styles.collapseStyles([isMobile ? styleOverride : undefined, props.styleOverride])}
```
chrisnojima: oh i didn't realize you need to mix on mobile. i'd do
```
styleOverride={{
...(isMobile ? styleOverride : {}),
...props.styleOverride
}}
```

xgess: done

PR #18980, opened 2019-08-16
Add a 4s delay before "Resend SMS" can be clicked

buoyad:
- Add a 4s period where "Resend SMS" is disabled when
  - Verify dialog opens
  - "Resend SMS" is hit

cc @keybase/y2ksquad

### DIFF
--- a/shared/signup/phone-number/verify.tsx
+++ b/shared/signup/phone-number/verify.tsx
@@ -71,6 +71,18 @@ type BodyProps = {
   resendWaiting: boolean
 }
 export const VerifyBody = (props: BodyProps) => {
+  // build in a delay to resend
+  // disable first, then enable after 2s
+  const [resendDisabled, setResendDisabled] = React.useState(true)
+  const enableResend = Kb.useTimeout(() => setResendDisabled(false), 4000)
+  React.useEffect(() => enableResend(), [enableResend])
+  const {onResend: _onResend} = props
+  const onResend = React.useCallback(() => {
+    _onResend()
+    setResendDisabled(true)
+    enableResend()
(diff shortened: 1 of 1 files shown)


shared/signup/phone-number/verify.tsx
```diff
@@ -71,6 +71,18 @@ type BodyProps = {
...
+  // disable first, then enable after 2s
+  const [resendDisabled, setResendDisabled] = React.useState(true)
+  const enableResend = Kb.useTimeout(() => setResendDisabled(false), 4000)
+  React.useEffect(() => enableResend(), [enableResend])
+  const {onResend: _onResend} = props
+  const onResend = React.useCallback(() => {
+    _onResend()
+    setResendDisabled(true)
+    enableResend()
+  }, [_onResend, enableResend])
```
jzila: this should have `setResendDisabled` in it too, right?

buoyad: Looks like it's not needed. The linter is ambivalent if it's included but doesn't add it - I'll leave it out here.

> (The identity of the setCount function is guaranteed to be stable so it’s safe to omit.)

https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often

buoyad: Looks like it's not needed. The linter is ambivalent if it's included but doesn't add it - I'll leave it out here.

> (The identity of the setCount function is guaranteed to be stable so it’s safe to omit.)

https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often

PR #18982, opened 2019-08-16
service: de-linting

strib:
Issue: HOTPOT-494


strib:
Merging over unrelated CI failure.

PR #18983, opened 2019-08-17
teams: de-linting

strib:
Issue: HOTPOT-495

### DIFF
--- a/go/teams/errors.go
+++ b/go/teams/errors.go
@@ -230,9 +230,13 @@ func IsTeamReadError(err error) bool {
 		switch keybase1.StatusCode(e.Code) {
 		case keybase1.StatusCode_SCTeamReadError:
 			return true
+		default:
+			// Nothing to do for other codes.
 		}
+		return false
+	default:
+		return false
 	}
-	return false
 }
 
 func FixupTeamGetError(ctx context.Context, g *libkb.GlobalContext, e error, teamDescriptor string, publicTeam bool) error {
--- a/go/teams/errors.go
+++ b/go/teams/errors.go
@@ -225,18 +225,8 @@ func NewExplicitTeamOperationError(m string) error {
 }
 
 func IsTeamReadError(err error) bool {
-	switch e := err.(type) {
-	case libkb.AppStatusError:
-		switch keybase1.StatusCode(e.Code) {
-		case keybase1.StatusCode_SCTeamReadError:
-			return true
-		default:
-			// Nothing to do for other codes.
-		}
-		return false
-	default:
-		return false
-	}
+	aerr, ok := err.(libkb.AppStatusError)
+	return ok && keybase1.StatusCode(aerr.Code) == keybase1.StatusCode_SCTeamReadError
 }
 
 func FixupTeamGetError(ctx context.Context, g *libkb.GlobalContext, e error, teamDescriptor string, publicTeam bool) error {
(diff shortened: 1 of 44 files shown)


go/teams/errors.go
```diff
@@ -230,9 +230,13 @@ func IsTeamReadError(err error) bool {
 		switch keybase1.StatusCode(e.Code) {
 		case keybase1.StatusCode_SCTeamReadError:
 			return true
+		default:
```
joshblum: think this would be more readable as:
```
aerr, ok := err.(libkb.AppStatusError)
return ok && aerr.Code == keybase1.StatusCode_SCTeamReadError
```

go/teams/handler.go
```diff
@@ -258,8 +258,14 @@ func handleChangeSingle(ctx context.Context, g *libkb.GlobalContext, row keybase
 	defer mctx.Trace(fmt.Sprintf("team.handleChangeSingle(%+v, %+v)", row, change),
 		func() error { return err })()
 
-	HintLatestSeqno(mctx, row.Id, row.LatestSeqno)
-	HintLatestHiddenSeqno(mctx, row.Id, row.LatestHiddenSeqno)
+	err = HintLatestSeqno(mctx, row.Id, row.LatestSeqno)
+	if err != nil {
+		return err
```
heronhaye: Unsure if a failure here should exit the function without continuing

go/teams/service_helper.go
```diff
@@ -871,9 +868,7 @@ func Delete(ctx context.Context, g *libkb.GlobalContext, ui keybase1.TeamsUiInte
 			return err
 		}
 
-		TombstoneTeam(libkb.NewMetaContext(ctx, g), t.ID)
-
-		return nil
+		return TombstoneTeam(libkb.NewMetaContext(ctx, g), t.ID)
```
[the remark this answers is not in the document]
heronhaye: Though it'd be nice to fail in devel so we know if something is wrong with that function

go/teams/handler.go
strib: @heronhaye You tell me!  I have no clue what this is.  Should I just be logging and ignoring it?

heronhaye: I think Max or Miles would know better than me

strib: @maxtaco @mlsteele thoughts on what to do about an error from `HintLatestSeqno`?  Ignore/log or return it?

mlsteele: Ignore would be good

go/teams/handler.go
```diff
@@ -260,11 +260,11 @@ func handleChangeSingle(ctx context.Context, g *libkb.GlobalContext, row keybase
 
 	err = HintLatestSeqno(mctx, row.Id, row.LatestSeqno)
 	if err != nil {
-		return err
+		mctx.Debug("team.handleChangeSingle HintLatestSeqno error: %+v", err)
 	}
 	err = HintLatestHiddenSeqno(mctx, row.Id, row.LatestHiddenSeqno)
 	if err != nil {
-		return err
+		mctx.Debug("team.handleChangeSingle HintLatestHiddenSeqno error: %+v", err)
```
mlsteele: I should've mentioned those methods already logs its own errors. So this here is double logging. If those methods' returned errors aren't used they can be dropped if you like.

PR #18988, opened 2019-08-19
Do not validateImpTeamname in IMPTEAMNATIVE

zapu:
This fixes not being able to unbox chat messages if chat started as an SBS conversation, which was then resolved, but later the SBS-ed in user revoked the SBS proof to start convo. Chat messages from before SBS-resolution would be unboxable because chat subsystem would not be able to verify conversation display name. These messages would still hold display name with SBS assertion, but since this assertion no longer resolves to that user, it will not match team's current display name anymore.

### DIFF
--- a/go/chat/sbs_test.go
+++ b/go/chat/sbs_test.go
@@ -59,7 +59,7 @@ func (p *ProveRooterUI) ContinueChecking(_ context.Context, _ int) (bool, error)
 	return true, nil
 }
 
-func proveRooter(t *testing.T, g *libkb.GlobalContext, fu *kbtest.FakeUser) {
+func proveRooter(t *testing.T, g *libkb.GlobalContext, fu *kbtest.FakeUser) (sigID keybase1.SigID) {
 	arg := keybase1.StartProofArg{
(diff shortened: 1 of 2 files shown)


go/chat/sbs_test.go
```diff
@@ -204,76 +220,123 @@ func runChatSBSScenario(t *testing.T, testCase sbsTestCase) {
...
+					messages = tvres.Thread.Messages
+				} else {
+					tc := ctc.world.Tcs[user.Username]
+					ctx := ctc.as(t, user).startCtx
+					// Nuke DB so we don't just pull cached messages.
+					_, err = tc.G.LocalDb.Nuke()
+					require.NoError(t, err)
+					_, err := tc.G.LocalChatDb.Nuke()
+					require.NoError(t, err)
+					tv, err := tc.Context().ConvSource.Pull(
```
zapu: I wasn't able to completely bypass all caches and simulate fresh load of this thread. Even without the patch to `chat/teams.go` this part of the test would still occasionally successfully load the thread.

zapu:
cc @jakob223

PR #18989, opened 2019-08-19
linux packaging: deeplinks on startup

heronhaye:


packaging/linux/systemd/keybase.gui.service
```diff
@@ -19,7 +19,7 @@ EnvironmentFile=-%t/keybase/keybase.gui.env
 EnvironmentFile=-%h/.config/keybase/keybase.autogen.env
 EnvironmentFile=-%h/.config/keybase/keybase.env
 
-ExecStart=/opt/keybase/Keybase
+ExecStart=/usr/bin/env bash -c "/opt/keybase/Keybase $KEYBASE_STARTUP_URL"
```
eli-schwartz:
```suggestion
ExecStart=/bin/sh -c '/opt/keybase/Keybase "$KEYBASE_STARTUP_URL"'
```

No need for bash-specific features, and therefore no chance that this will fail to work on BSD systems where bash is not available in /bin/bash, but only in /usr/local/bin/bash -- hence no need for env either.

/bin/sh is POSIX and exists everywhere (though may represent any of a number of shells).

Single quotes + quoting because it's exceedingly unlikely anyone tries to use KEYBASE_STARTUP_URL='lol this has spaces', but why not make sure?

heronhaye:
Closing this out, going with another strategy due to the flaws you pointed out Eli

Issue #18940, opened 2019-08-14
Centos 7: broken upgrade


scaprile:
After yum update, Keybase no longer starts
dmesg shows:
[   92.744956] traps: Keybase[3628] trap int3 ip:562ca0f15537 sp:7ffffdb2fa00 error:0
(system start)
[ 1605.443419] traps: Keybase[5464] trap int3 ip:55b2917f2537 sp:7ffdb66774e0 error:0
(manual start from GUI)
syslog on manual start from GUI:
Aug 14 19:08:28 Workstation keybase.desktop: Unmounting /run/user/500/keybase/kbfs...
Aug 14 19:08:28 Workstation keybase.desktop: Unmounting and shutting down kbfsfuse...
Aug 14 19:08:28 Workstation keybase.desktop: Shutting down keybase service...
Aug 14 19:08:28 Workstation keybase.desktop: Launching keybase service...
Aug 14 19:08:28 Workstation keybase.desktop: Starting the redirector...
Aug 14 19:08:28 Workstation keybase.desktop: Starting KBFS...
Aug 14 19:08:28 Workstation keybase.desktop: Launching Keybase GUI...
Aug 14 19:08:28 Workstation keybase.desktop: run_keybase: Success!
Aug 14 19:08:28 Workstation kernel: traps: Keybase[5868] trap int3 ip:55a10c40d537 sp:7ffdefad2cd0 error:0
Aug 14 19:08:28 Workstation abrt-hook-ccpp: Process 5868 (Keybase) of user 500 killed by SIGTRAP - dumping core
Aug 14 19:08:32 Workstation abrt-server: Package 'keybase' isn't signed with proper key

my log id: 4647d90ab80dd83ec50d661c

After setting OpenGPGCheck = no in file /etc/abrt/abrt-action-save-package-data.conf, I uploaded a new log, id: 0ff2bdd22a8d27019e22601c

heronhaye:
So it's working now? You shouldn't need to skip the GPG check, you can import [our code signing key](https://keybase.io/docs/server_security/our_code_signing_key).

scaprile:
No, it is not working,
I'm literate enough to follow debugging instructions, if needed. My knowledge of the operating system is limited, though.

heronhaye:
Does `keybase id` work in the CLI after `run_keybase`? If so, I guess it's only the GUI that isn't working. What happens if you run `/opt/keybase/Keybase` manually?

scaprile:
Works like a charm
[...]

PR #18996, opened 2019-08-19
setuid chrome sandbox for rpm

heronhaye:
Ref: https://github.com/keybase/client/issues/18940#event-2560275531


heronhaye:
Sorry for the noise @strib. This works now.

PR #18997, opened 2019-08-19
Fix auxiliary config filenames when using --use-root-config-file

heronhaye:


go/libkb/env.go
```diff
@@ -701,6 +711,17 @@ func (e *Env) GetUpdaterConfigFilename() string {
...
 func (e *Env) GetGUIConfigFilename() string {
 	return e.GetString(
+		func() string {
+			if e.GetUseRootConfigFile() {
+				dir, err := e.GetRootConfigDirectory()
+				if err != nil {
+					return ""
+				}
+				return filepath.Join(dir, GUIConfigFile)
+
```
strib: Extra blank line.

PR #18998, opened 2019-08-19
added type to manage node position in the merkle tree

AMarcedone:
cc @maxtaco


go/merkletreev2/tree.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Hash is a byte-array, used to represent a full collision-resistant hash.
+type Hash []byte
+
+// Key is a byte-array, and it is the type of the keys in the KeyValuePairs that
+// the tree can store.
+type Key []byte
+
+// SeqNo is an integer used to differentiate different versions of a merkle tree.
+type SeqNo uint32
```
maxtaco: why only 32 bits here?

go/merkletreev2/config.go
```diff
@@ -0,0 +1,53 @@
...
+
+import "fmt"
+
+// Config defines the shape of the MerkleTree.
+type Config struct {
+	// A hasher is used to compute hashes in this configuration
+	hasher Hasher
+
+	// The number of children per node. Must be a power of two.
+	childrenPerNode uint
```
heronhaye: This can't be more than 256 right?

go/merkletreev2/position.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return siblings, father, nil
+}
+
+// getDeepestPositionForKey converts the key into the position the key would be
+// stored at if the tree was full with only one key per leaf.
+func (t *Tree) getDeepestPositionForKey(k Key) (p Position) {
+	// Here we assume that len(k)*8 is a multiple of t.cfg.bitsPerIndex
+	(*big.Int)(&p).SetBytes(k)
+	(*big.Int)(&p).SetBit((*big.Int)(&p), len(k)*8+1, 1)
```
heronhaye: should this be `len(k)*8`?  like on line 35

go/merkletreev2/position.go
```diff
@@ -0,0 +1,95 @@
...
+	return p
+}
+
+func (t *Tree) getSiblingPositionsOnPathToKey(k Key) ([][]Position, error) {
+	var err error
```
heronhaye: rm

heronhaye: Also don't really need to return an error for this function

go/merkletreev2/position_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"math/big"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPositionEncoding(t *testing.T) {
```
heronhaye: These unit test functions should use something like https://github.com/golang/go/wiki/TableDrivenTests

go/merkletreev2/position_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	_, err := NewConfig(nil, 5, 8, nil)
+	require.Error(t, err)
```
heronhaye: If you want to assert an error you should assert the type of the error too.

go/merkletreev2/position_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	var c Config
```
heronhaye: don't need this, just `c, err :=` on next line

go/merkletreev2/tree.go
AMarcedone: I didn't think we would need more, but it's free so I'll change that.

go/merkletreev2/config.go
```diff
@@ -0,0 +1,53 @@
+package merkletreev2
```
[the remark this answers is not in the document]
AMarcedone: @max any suggestions? merkle-tree-v2?

go/merkletreev2/position.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// Position represents the position of a node in the tree. When converted to
+// bytes, a Position can be interpreted as a 1 followed (from left to right) by
+// a sequence of log2(Config.m)-bit symbols, where each such symbol identifies
+// which child to descend to in a path from the root to a node. The sequence is
+// padded with 0s on the left to the nearest byte. For example, in a binary tree
+// the root has position 0x01 (i.e. 0b00000001), and the second child of the
+// first child of the root has position 0x05 (0b00000101).
+type Position big.Int
```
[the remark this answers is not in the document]
AMarcedone: It is a single file that has the casts, so I'd rather be more kosher and not expose all the other methods. :)

go/merkletreev2/config.go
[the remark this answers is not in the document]
maxtaco: how about no `v`?

go/merkletreev2/config.go
```diff
@@ -0,0 +1,53 @@
...
+	hasher Hasher
+
+	// The number of children per node. Must be a power of two.
+	childrenPerNode uint
+
+	// The maximum number of KeyValuePairs in a leaf node before we split
+	valuesPerLeaf uint
+
+	// The number of bits necessary to represent a ChildIndex, i.e. log2(m)
+	bitsPerIndex uint8
```
maxtaco: Is this derived from the other values? If so, probably want at least a test to make sure it's always in alignment.

go/merkletreev2/config.go
```diff
@@ -0,0 +1,53 @@
...
+	if !isPowerOfTwo(childrenPerNode) {
+		return Config{}, fmt.Errorf("childrenPerNode needs to be a power of two")
+	}
+	return Config{hasher: h, childrenPerNode: childrenPerNode, valuesPerLeaf: valuesPerLeaf, bitsPerIndex: log2(childrenPerNode), valueConstructor: valueConstructor}, nil
+}
+
+// Hasher is an interface for hashing MerkleTree data structures into their
+// cryptographic hashes.
+type Hasher interface {
+	Hash([]byte) Hash
```
maxtaco: shouldn't the signature then be `Hash(interface{}) Hash` ?

go/merkletreev2/position.go
```diff
@@ -0,0 +1,95 @@
...
+	q.SetBit(q, len(k)*8, 1)
+	q.Rsh(q, uint(q.BitLen()-(*big.Int)(p).BitLen()))
+	return (*big.Int)(p).Cmp(q) == 0
+}
+
+func (p *Position) equals(q *Position) bool {
+	return (*big.Int)(p).Cmp((*big.Int)(q)) == 0
+}
+
+func (t *Tree) getFather(p *Position) (f Position, err error) {
```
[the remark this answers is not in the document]
maxtaco: @heronhaye also would want this to be `getParent`?

go/merkletreev2/position.go
```diff
@@ -0,0 +1,95 @@
...
+		return false
+	}
+	q := new(big.Int)
+	q.SetBytes([]byte(k))
+	q.SetBit(q, len(k)*8, 1)
+	q.Rsh(q, uint(q.BitLen()-(*big.Int)(p).BitLen()))
+	return (*big.Int)(p).Cmp(q) == 0
+}
+
+func (p *Position) equals(q *Position) bool {
```
maxtaco: Can `p` and `q` be non-pointers?

go/merkletreev2/position.go
AMarcedone: It should be ok like this

go/merkletreev2/config.go
AMarcedone: I copied it from previous code (I think the point is that the encoding needs to happen manually before the hasher can do its job. I can change if you want.

go/merkletreev2/position.go
AMarcedone: is it just in this equal method (so it is not confusing that we are checking equality of the pointed objects and not the address)? Or is there some other reasons?

AMarcedone: In general they can, but then we will be copying these objects, and all the rest of the functions have pointer arguments. Do you want me to change that? @maxtaco

maxtaco: I usually make these sorts of things call-by-value and don't worry about the copy costs.

heronhaye:
Tests look great. Only other comment is the off by one issue we discussed + tests for that

go/merkletree2/position_test.go
```diff
@@ -0,0 +1,197 @@
...
+	tree2bits := &Tree{cfg: config2bits}
+	config3bits, _ := NewConfig(nil, 3, 1, 3, nil)
+	tree3bits := &Tree{cfg: config3bits}
+
+	encodingTests := []struct {
+		t   *Tree
+		bin string
+		p   Position
+	}{
+		{tree1bit, "1", *tree1bit.getRootPosition()},
```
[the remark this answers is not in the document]
AMarcedone: I can't test that there, as the empty string is not a valid encoding of any position.

PR #19001, opened 2019-08-19
common-adapters (G-Q) dark mode support

hyperobject:
This PR updates the common adapters (G-Q) so they support our dark mode infrastructure. Most of this work is changing calls to `Styles.styleSheetCreate` so they use closures instead of passing a static object. This PR also contains some refactoring, primarily moving imports and styles from `desktop/renderer/style.css` to their own css files and importing them in relevant components.
@keybase/react-hackers 
@keybase/hotpotatosquad

### DIFF
--- a/shared/common-adapters/icon.desktop.tsx
+++ b/shared/common-adapters/icon.desktop.tsx
@@ -202,10 +202,10 @@ export function castPlatformStyles(styles: any) {
   return Shared.castPlatformStyles(styles)
 }
 
-const styles = Styles.styleSheetCreate({
+const styles = Styles.styleSheetCreate(() => ({
   // Needed because otherwise the containing box doesn't calculate the size of
   // the inner span (incl padding) properly
   flex: {display: 'flex'},
-})
+}))
 
 export default Icon
--- a/shared/common-adapters/icon.desktop.tsx
+++ b/shared/common-adapters/icon.desktop.tsx
@@ -8,8 +8,6 @@ import {resolveImageAsURL} from '../desktop/app/resolve-root.desktop'
 import {invert} from 'lodash-es'
 import {Props, IconType} from './icon'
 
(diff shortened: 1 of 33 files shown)


shared/common-adapters/icon.desktop.tsx
```diff
@@ -110,6 +108,8 @@ class Icon extends Component<Props, void> {
           hoverColor: 'inherit',
         }
       } else {
+        // invert the colors here so it reflects the colors in current theme
+        const invertedColors = invert(Styles.globalColors)
```
jakob223: how expensive an operation is `invert`? If a page has hundreds of icons on it on mobile will this be a problem? At the very least maybe we should invert only the colors used here?

shared/common-adapters/new-input.tsx
```diff
@@ -66,7 +60,7 @@ class ReflessNewInput extends React.Component<Props & RefProps, State> {
         {!!this.props.icon && (
           <Box style={styles.icon}>
             <Icon
-              color={globalColors.black_20}
+              color={Styles.globalColors.black_20} // not sure how to make this dynamic
```
jakob223: @chrisnojima did you have a plan for icon inverting?

shared/common-adapters/icon.desktop.tsx
hyperobject: it takes an object K -> V and returns another object with V -> K, so I can't imagine it's super cheap/performant. There should almost certainly be a better solution than this, but I'm not sure what that would look like.

chrisnojima: It's fine

shared/common-adapters/new-input.tsx
chrisnojima: this will automatically get inverted since its in a render

PR #19003, opened 2019-08-19
Fix failing tests on master

nathunsmitty:
Adds a regex to match the new teamID output field in some team api responses introduced by #18807.


zapu:
:+1: thanks

PR #19004, opened 2019-08-19
Fix notif settings loading and layout

buoyad:
cc @keybase/y2ksquad

### DIFF
--- a/shared/settings/notifications/render.tsx
+++ b/shared/settings/notifications/render.tsx
@@ -15,14 +15,21 @@ const Group = (props: {
   unsub?: string
   unsubscribedFromAll: boolean
 }) => (
-  <Kb.Box2 direction="vertical">
+  <Kb.Box2 direction="vertical" fullWidth={true}>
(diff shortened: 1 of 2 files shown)


shared/settings/notifications/render.tsx
```diff
@@ -94,19 +100,24 @@ const Notifications = (props: Props) =>
...
-      ) : Styles.isMobile /* TODO: display something if the user needs to enable push? */ ? null : (
-        <Kb.Box2 direction="vertical">
-          <Kb.Text type="Header">Phone notifications</Kb.Text>
-          <Kb.Text type="BodySmall">Install the Keybase app on your phone.</Kb.Text>
+        <>
+          <PhoneSection {...props} />
           <Kb.Divider style={styles.divider} />
-        </Kb.Box2>
+        </>
+      ) : (
```
jzila: I think this set of parens isn't necessary.

buoyad: If I remove them prettier adds them back

PR #19005, opened 2019-08-19
Phone Search Interface

ddworken:
Note that currently if a phone number does not resolve, it will simply display a loading symbol forever. A later PR is going to fix that. What currently works is:

1. Start a conversation with a user from a resolved phone number
2. Start a conversation with an unresolved phone number

cc: @keybase/react-hackers 

Desktop:

Mobile: 
Screenshot_20190819-160954
Screenshot_20190819-161005
Screenshot_20190819-161017
Screenshot_20190819-161035


go/service/usersearch.go
```diff
@@ -344,6 +344,7 @@ func imptofuSearch(mctx libkb.MetaContext, provider contacts.ContactsProvider, i
 					// Ignore full-name here, force queryString as `prettyName`
 					// part in order for it to be displayed in the list.
 					imptofu.PrettyName = queryString
+					imptofu.Fullname = uname.Fullname
```
ddworken: For imptofu queries, the pretty name is set to the query string. This is done so that on the main contacts search page the subtext is the query (eg a phone number or email). But for the phone search page we want to display their full name as the subtext.

shared/team-building/index.tsx
```diff
@@ -495,12 +499,22 @@ class TeamBuilding extends React.PureComponent<Props, {}> {
             onRedoRecs={props.fetchUserRecs}
           />
         )}
-        {this._searchInput()}
-        {this._listBody()}
-        {props.waitingForCreate && (
-          <Kb.Box2 direction="vertical" style={styles.waiting} alignItems="center">
-            <Kb.ProgressIndicator type="Small" white={true} style={styles.waitingProgress} />
-          </Kb.Box2>
+        {props.selectedService === 'phone' ? (
```
nathunsmitty: If you wanna help make a future conflict a little less painful, I made this logic into a switch statement starting at 454. My code:
```
    let content
    switch (props.selectedService) {
      case 'email':
        content = <EmailInput />
        break
      // case 'phone':
      //   break
      default:
        content = (
          <>
            {this._searchInput()}
            {this._listBody()}
            {props.waitingForCreate && (
              <Kb.Box2 direction="vertical" style={styles.waiting} alignItems="center">
                <Kb.ProgressIndicator type="Small" white={true} style={styles.waitingProgress} />
              </Kb.Box2>
            )}
          </>
        )
    }
```

shared/team-building/phone-search.tsx
```diff
@@ -0,0 +1,109 @@
...
+      >
+        <PhoneInput
+          // Supply a key to force reset the PhoneInput state after a user is added
+          key={phoneInputKey}
+          autoFocus={true}
+          onChangeNumber={setPhoneNumber}
+          onChangeValidity={(val: boolean, num: string) => {
+            setValidity(val)
+            if (val) {
+              // TODO: Is this okay to reuse the 'keybase' service name? Or should we add a 'phone' one to iced?
```
ddworken: @zapu Thoughts on this? It appears as though this service name is passed back to the backend where the `phone` service name does not exist. Instead I'm reusing the `keybase` service name which is currently used in the contacts page when searching for phone numbers (and is known/recognized by the backend). Would it be better to add a backend change to add support for the `phone` service name?

shared/team-building/container.tsx
```diff
@@ -136,6 +136,7 @@ const emptyObj = {}
 
 const mapStateToProps = (state: Container.TypedState, ownProps: OwnProps) => {
   const teamBuildingState = state[ownProps.namespace].teamBuilding
+  const teamBuildingSearchResults = teamBuildingState.teamBuildingSearchResults.toJS()
   const userResults = teamBuildingState.teamBuildingSearchResults.getIn([
```
nathunsmitty: It looks like `userResults` is very similar to `teamBuildingSearchResults`, can we use that instead?

ddworken: Hmm, it looks like we could but I'm not sure whether it is a cleaner design. Using it would require having phone-search track the current phone number in redux so that this container can access the current phone number and pull out the result? Does that seem cleaner? To me it seems like that adds a bit more indirection than is necessary.

shared/team-building/shared.tsx
```diff
@@ -38,6 +38,11 @@ const services: {[K in ServiceIdWithContact]: {color: string; icon: IconType; la
     icon: 'iconfont-identity-pgp',
     label: 'PGP',
   },
+  phone: {
+    color: '#4C8EFF',
+    icon: 'iconfont-phone-contact',
```
ddworken: @cecileboucheron Do you know the name of the correct icon to use per the design? I don't see the matching icon in `shared/images/iconfont/`.

shared/team-building/index.tsx
```diff
@@ -75,13 +77,15 @@ export type Props = ContactProps & {
...
   recommendations: Array<SearchRecSection> | null
+  search: (query: string, service: ServiceIdWithContact) => void
   searchResults: Array<SearchResult> | null
   searchString: string
   selectedService: ServiceIdWithContact
   serviceResultCount: {[K in ServiceIdWithContact]?: number | null}
   showRecs: boolean
   showResults: boolean
   showServiceResultCount: boolean
+  teamBuildingSearchResults: any
```
nathunsmitty: 🔥

shared/team-building/phone-search.tsx
```diff
@@ -0,0 +1,118 @@
...
+              props.search(num, 'keybase')
+            }
+          }}
+          result={
+            // Pass a component into PhoneInput so it is displayed inline with the number input box
+            validity &&
+            !!user && (
+              <Kb.ClickableBox onClick={_onContinue} style={{margin: Styles.globalMargins.tiny}}>
+                <Kb.Box2 direction="horizontal" gap="tiny" fullWidth={true}>
+                  <Kb.Avatar size={48} username={user.username} showFollowingStatus={true} />
```
nathunsmitty: I think <NameWithIcon> is what you want here

shared/team-building/index.tsx
ddworken: @nathunsmitty Can you link me to your code where you did this? Without switch expressions I don't see how what you're doing works. Are you abstracting it into a separate method?

shared/team-building/phone-search.tsx
ddworken: Done.

shared/signup/phone-number/phone-input.tsx
```diff
@@ -218,8 +218,10 @@ type Props = {
   autoFocus?: boolean
   defaultCountry?: string
   onChangeNumber: (phoneNumber: string) => void
-  onChangeValidity: (valid: boolean) => void
+  onChangeValidity: (valid: boolean, phoneNumber: string) => void
   onEnterKeyDown?: () => void
+  // A ReactNode to display beneath and aligned to the number input box
+  result?: React.ReactNode
```
buoyad: Hm, I don't think it makes sense for `PhoneInput` to add this. Can `PhoneSearch` render the result below directly?

shared/signup/phone-number/phone-input.tsx
```diff
@@ -218,8 +218,10 @@ type Props = {
   autoFocus?: boolean
   defaultCountry?: string
   onChangeNumber: (phoneNumber: string) => void
-  onChangeValidity: (valid: boolean) => void
+  onChangeValidity: (valid: boolean, phoneNumber: string) => void
```
buoyad: Might make sense to include `valid` in `onChangeNumber` and get rid of this callback. maybe add a TODO?

shared/team-building/container.tsx
```diff
@@ -136,6 +136,7 @@ const emptyObj = {}
 
 const mapStateToProps = (state: Container.TypedState, ownProps: OwnProps) => {
   const teamBuildingState = state[ownProps.namespace].teamBuilding
+  const teamBuildingSearchResults = teamBuildingState.teamBuildingSearchResults.toJS()
```
buoyad: call `toJS` in `mergeProps` instead. The output of `mapStateToProps` is compared across state updates to decide whether to render, calling `toJS` here creates a new object each time so this'll render more than it should.

shared/signup/phone-number/phone-input.tsx
ddworken: The issue is that the [design](https://app.zeplin.io/project/56b25ffc64fc91773a73aaf8/screen/5d2775a1dadd1f77b1768116) calls for having the person aligned with the number input box rather than directly underneath it. I confirmed with Cecile that she does want it aligned in that way and when I asked in #frontend about the best way to implement it, this is what Song suggested. How would you implement it?

shared/signup/phone-number/phone-input.tsx
ddworken: Refactored to remove the callback.

shared/team-building/container.tsx
ddworken: Done

shared/signup/phone-number/phone-input.tsx
buoyad: I guess I'd do it this way. @cecileboucheron do you think we could align it on mobile the same way as on desktop? The user result area moves as a result of changing the country code as it is, which seems weird.

shared/settings/account/add-modals.tsx
```diff
@@ -166,8 +166,10 @@ export const Phone = () => {
         style={styles.body}
       >
         <EnterPhoneNumberBody
-          onChangeNumber={onChangeNumber}
-          onChangeValidity={onChangeValidity}
+          onChangeNumber={(phoneNumber: string, validity: boolean) => {
```
buoyad: Best to do
```
const onChangeNumberCb = React.useCallback((phoneNumber: string, validity: boolean) => {
  onChangeNumber(phoneNumber)
  onChangeValidity(validity)
}, [])
```
above and pass that in here. Same for the rest of these.

shared/signup/phone-number/phone-input.tsx
ddworken: Additionally, @cecileboucheron the desktop design didn't show where the person should be aligned? I decided it seemed to make more sense to have the person left aligned on the desktop since it seemed a little squished otherwise but curious to get your opinion.

buoyad: Desktop for reference:

shared/settings/account/add-modals.tsx
ddworken: Done.

shared/signup/phone-number/phone-input.tsx
cecileboucheron: Sure, all left-aligned sounds good to me.

shared/team-building/phone-search.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <Kb.Box style={styles.spaceFillingBox} />
+      <Kb.Box2 direction="horizontal" style={styles.buttonContainer} fullWidth={true}>
+        <Kb.Button style={styles.button} onClick={_onContinue} label="Continue" disabled={!validity} />
+      </Kb.Box2>
+    </>
+  )
+}
+
+const styles = Styles.styleSheetCreate(() => ({
+  button: {marginBottom: Styles.globalMargins.tiny, width: '80%'},
```
buoyad: we typically use a padding on the container to add spacing on the sides. In this case, it makes sense to wrap `PhoneSearch`'s body in something with the padding and background colors added. Then you can `width: '100%'` inside.

ddworken: Done.

shared/team-building/phone-search.tsx
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+  if (state === 'none' && validity && props.teamBuildingSearchResults[phoneNumber] === undefined) {
+    state = 'loading'
+  }
+  if (state === 'none' && validity && props.teamBuildingSearchResults[phoneNumber] !== undefined) {
+    state = 'notfound'
+  }
+
+  let _onContinue = () => {
```
buoyad: `React.useCallback`

shared/team-building/phone-search.tsx
```diff
@@ -0,0 +1,143 @@
...
+        <Kb.Box2 direction="vertical" gap="tiny">
+          <PhoneInput
+            // Supply a key to force reset the PhoneInput state after a user is added
+            key={phoneInputKey}
+            autoFocus={true}
+            onChangeNumber={onChangeNumberCb}
+          />
+          {state === 'resolved' && !!user && (
+            <Kb.Box2 direction="horizontal" style={styles.resultContainer}>
+              <Kb.NameWithIcon
```
buoyad: This could use a `width: 100%` `containerStyle`

shared/team-building/phone-search.tsx
```diff
@@ -0,0 +1,143 @@
...
+    height: '100%',
+    paddingBottom: Styles.globalMargins.tiny,
+    paddingLeft: Styles.globalMargins.small,
+    paddingRight: Styles.globalMargins.small,
+    paddingTop: Styles.globalMargins.tiny,
+    width: '100%',
+  },
+  justifyCenter: {justifyContent: 'center'},
+  loading: {alignSelf: 'center'},
+  placeholderIcon: {borderRadius: 16, height: '32px'},
```
buoyad: height should be a `number`

PR #19007, opened 2019-08-19
systests: de-lint

strib:
Issue: HOTPOT-496

### DIFF
--- a/go/systests/user_test.go
+++ b/go/systests/user_test.go
@@ -49,14 +49,6 @@ type signupSecretUI struct {
 	parent *signupUI
 }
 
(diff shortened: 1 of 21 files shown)


go/systests/user_test.go
```diff
@@ -201,7 +192,7 @@ func randomUser(prefix string) *signupInfo {
 
 func randomDevice() string {
 	b := make([]byte, 5)
-	rand.Read(b)
+	_, _ = rand.Read(b)
```
mlsteele: might as well fail or panic if Read errors

PR #19014, opened 2019-08-20
immer notifications

chrisnojima:
- [x] notifications to immer
- [x] remove unused common-adapters/tab-bar


shared/menubar/remote-proxy.desktop.tsx
```diff
@@ -136,7 +136,7 @@ const mapStateToProps = (state: Container.TypedState) => ({
   showingDiskSpaceBanner: state.fs.overallSyncStatus.showingBanner,
   userInfo: state.users.infoMap,
   username: state.config.username,
-  widgetBadge: state.notifications.get('widgetBadge') || 'regular',
+  widgetBadge: state.notifications.widgetBadge || 'regular',
```
songgao: Is the `|| 'regular'` part necessary? It's populated from the initial state right?

chrisnojima: i agree it can be removed, updated

PR #19021, opened 2019-08-20
journal_manager: rename journal dirs with broken info state

strib:
If a journal's info file becomes corrupt in some way (most likely because it is empty due to an inopportune client crash), just move it out of the way so that the TLF can create a new journal for itself.

Issue: HOTPOT-529

### DIFF
--- a/go/kbfs/libkbfs/journal_manager.go
+++ b/go/kbfs/libkbfs/journal_manager.go
@@ -8,7 +8,9 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


go/kbfs/libkbfs/journal_manager.go
```diff
@@ -26,6 +28,13 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const (
+	tlfJournalBrokenFmt = "%s-%d.broken"
+)
+
+var tlfJournalBrokenRegexp = regexp.MustCompile(
+	`[/\\]([[:alnum:]]+)-([[:digit:]]+)\.broken$`)
```
jakob223: whoa i didn't know you could do :alnum: and :digit:, cool

PR #19024, opened 2019-08-20
New team convo email search

nathunsmitty:
~~Needs some `console.log`s to be cleaned up, a story, and to be tested on mobile; but the main functionality is there.~~

This is ready for review:

desktop

Storybook link:  [🖥](http://localhost:6006/?path=/story/team-building--email-address)


shared/team-building/index.tsx
```diff
@@ -432,6 +439,27 @@ class TeamBuilding extends React.PureComponent<Props, {}> {
...
   render = () => {
     const props = this.props
+
+    let content
+    switch (props.selectedService) {
+      case 'email':
+        content = <EmailInput namespace={props.namespace} />
+        break
+      // case 'phone':
+      //   break
```
nathunsmitty: @ddworken

shared/team-building/service-tab-bar.tsx
```diff
@@ -33,76 +33,82 @@ const HoverIcon = Styles.styled<typeof Kb.Icon, ExtraProps>(Kb.Icon)(props => ({
...
 
-const ServiceIconDesktop = (props: IconProps) => (
-  <Kb.ClickableBox onClick={props.onClick} style={styles.clickableServiceIcon}>
-    <Kb.Box2 direction="horizontal" centerChildren={true} style={styles.serviceIconContainer}>
-      <Kb.WithTooltip text={serviceIdToLabel(props.service)}>
-        <HoverIcon
-          isActive={props.isActive}
-          service={props.service}
+const ServiceIconDesktop = (props: IconProps) => {
+  const showCount = props.showCount && props.service !== 'email'
```
nathunsmitty: Wish the diff was better, I basically just added this line and used `showCount` instead of `props.shouwCount`

shared/actions/team-building.tsx
```diff
@@ -191,6 +191,18 @@ const fetchUserRecs = async (
   }
 }
 
+async function searchEmailAddress(state: TypedState, {payload: {namespace}}: SearchOrRecAction) {
```
nathunsmitty: Searching an email address is a sort of special case of searching a service, so I thought it made sense to create a separate action out of it so we don't have to deal with extra results or searching the query against all services.

shared/team-building/email-input/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+import * as TeamBuildingGen from '../../actions/team-building-gen'
+import {AllowedNamespace} from 'constants/types/team-building'
+import {requestIdleCallback} from '../../util/idle-callback'
+
+type EmailInputProps = {
+  namespace: AllowedNamespace
+}
+
+function checkValidEmail(str: string): boolean {
+  const emailRegex = /^(\S+@\S+\.\S+)$/
```
nathunsmitty: Stole this regex from `libkb/checkers.go`. Couldn't find anything client side that did email validation, and I didn't want to send an RPC on each key to validate the email...

shared/team-building/email-input/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const [emailString, setEmailString] = React.useState('')
+  const dispatch = Container.useDispatch()
+  const user = Container.useSelector(state => {
+    return state[namespace].teamBuilding.teamBuildingEmailResult
+  })
+  const isSearching = Container.useSelector(
+    state => state[namespace].teamBuilding.teamBuildingEmailIsSearching
+  )
+
+  const debouncedSearch = debounce(
```
nathunsmitty: We debounce searches to services in `shared/team-building/container.tsx`, so it seemed like a good idea to do that here too.

shared/team-building/email-input/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+      if (isNewInputValid) {
+        debouncedSearch(text)
+      }
+    },
+    [isEmailValid, debouncedSearch]
+  )
+
+  const onSubmit = React.useCallback(() => {
+    if (!user) {
+      throw new Error('User not found, cannot submit')
```
nathunsmitty: We should never get into this state, but is there something more graceful we could do here than throw an error?

shared/team-building/email-input/index.tsx
buoyad: This should get `React.useCallback` too

shared/reducers/team-building.tsx
```diff
@@ -60,6 +60,20 @@ export default function(
         teamBuildingSelectedService: service,
       })
     }
+    case TeamBuildingGen.searchEmailAddress: {
+      const {query} = action.payload
+      return state.merge({
+        teamBuildingEmailIsSearching: true,
+        teamBuildingEmailSearchQuery: query.trim(),
```
buoyad: clear `teamBuildingEmailResult`

shared/team-building/email-input/index.tsx
buoyad: I don't think this should throw. Currently you can get in this state via the `onEnterKeyDown={onSubmit}` below. I don't think it'd be bad if this fails silently? You could do something like `realOnSubmit = user ? onSubmit : null` and use that in the `onEnterKeyDown` and to drive the enabled-ness of the button.

shared/team-building/email-input/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+        </Kb.Box2>
+      </Kb.Box2>
+    </Kb.Box2>
+  )
+}
+
+const styles = Styles.styleSheetCreate(() => ({
+  background: {
+    backgroundColor: Styles.globalColors.blueGrey,
+    flexBasis: 0,
```
buoyad: I think `flex: 1` does these three

shared/ios/Keybase/Storybook.m
```diff
@@ -24,7 +24,7 @@ + (BOOL)requiresMainQueueSetup
 - (NSDictionary *)constantsToExport
 {
   // Set this to true to enable storybook mode
-  return @{@"isStorybook": @false};
+  return @{@"isStorybook": false};
```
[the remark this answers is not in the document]
nathunsmitty: nope, fixed

shared/team-building/email-input/index.tsx
nathunsmitty: How about just returning instead of throwing the error here?

shared/reducers/team-building.tsx
```diff
@@ -60,6 +60,20 @@ export default function(
...
+      })
+    }
+    case TeamBuildingGen.searchEmailAddressResultLoaded: {
+      const {user} = action.payload
```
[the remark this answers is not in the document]
nathunsmitty: Don't I need the query to actually do api search? From `actions/team-building.tsx`:

```tsx
  const query = state[namespace].teamBuilding.teamBuildingEmailSearchQuery
  const impTofuQuery = makeImpTofuQuery(query, null)

  const users = await apiSearch(query, 'keybase', 1, true, impTofuQuery, false)
```

shared/team-building/email-input/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      }
+      if (isNewInputValid) {
+        debouncedSearch(text)
+      }
+    },
+    [isEmailValid, debouncedSearch]
+  )
+
+  const onSubmit = React.useCallback(() => {
+    if (user === null || !canSubmit) {
```
nathunsmitty: I have to recheck if `user === null` here or else tsc complains. Not the end of the world, but any suggestions on how I could maybe improve this?

shared/team-building/email-input/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+          onEnterKeyDown={onSubmit}
+          textContentType="emailAddress"
+          value={emailString}
+        />
+        {isSearching && (
+          <Kb.Box2 direction="horizontal" fullWidth={true}>
+            <Kb.ProgressIndicator type="Small" />
+          </Kb.Box2>
+        )}
+        {user !== null && canSubmit && emailHasKeybaseAccount && (
```
nathunsmitty: same here

shared/reducers/team-building.tsx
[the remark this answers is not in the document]
buoyad: `teamBuildingEmailSearchQuery` in the store might not match `action.payload.query`, we should ignore these results if they don't match.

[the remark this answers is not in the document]
nathunsmitty: Ahh, makes sense. I'll add that check

PR #19025, opened 2019-08-21
fix panic that surya ran into

songgao:


go/service/identify.go
```diff
@@ -357,16 +357,18 @@ func (h *IdentifyHandler) resolveIdentifyImplicitTeamDoIdentifies(ctx context.Co
...
+				if idRes.TrackBreaks != nil && idErr == nil {
+					trackBreaksLock.Lock()
+					defer trackBreaksLock.Unlock()
+					if res.TrackBreaks == nil {
+						res.TrackBreaks = make(map[keybase1.UserVersion]keybase1.IdentifyTrackBreaks)
+					}
+					res.TrackBreaks[idRes.Upk.ToUserVersion()] = *idRes.TrackBreaks
+					brokenUsernames = append(brokenUsernames, idRes.Upk.GetName())
+				} else {
+					okUsernames = append(okUsernames, idRes.Upk.GetName())
```
heronhaye: Looks ok, but does this need to be protected by a lock too?

songgao: yes, good catch!

PR #19027, opened 2019-08-21
chat1.TeamMember

joshblum:
mirrors `keybase1.TeamMember` but with a `gregor1.UID`, needed to bring the `PIN` stuff in to maintain status quo of server vendor.


mmaxim:
Sorry about pin

joshblum:
all good

PR #19028, opened 2019-08-21
service: don't block mobile init on online login

maxtaco:


maxtaco:
cc @mmaxim / @malgorithms  / @chrisnojima

PR #19029, opened 2019-08-21
deprecate app-state

chrisnojima:
The goal of this PR is to simplify and unify how we deal with app state settings. Currently we have a kinda weird one-off app-state.desktop which persists data to its own custom json file. This causes us to have a bunch of custom code that shuttles data back and forth between the ipcs to manage this data. Instead of that we do this:

1. On boot the main node thread reads the gui config.json file. It needs to know some things before the renderer even exists.
2. Renderer starts up and reads this file through the rpc mechanism
3. Changes to prefs (in settings, things like sounds on notifications, native frame, etc) use redux and as a side effect we use the rpc to write it down (like our normal flows)
4. If the electron window moves, the main thread dispatches an action to the renderer (our usual flow) to update, which as a side effect uses the rpc to write it down


chrisnojima:
- [ ] after release

shared/actions/json/config.json
```diff
@@ -138,6 +126,13 @@
       "_description": "remote electron window wants props sent",
       "component": "string",
       "param": "string"
+    },
+    "updateWindowState": {
+      "_description": "main electron window wants to store its state",
+      "windowState": "Types.WindowState",
+    },
+    "setUseNativeFrame": {
```
chrisnojima: moved from settings

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -35,9 +31,7 @@ export async function saveAttachmentToCameraRoll() {
 }
 
 const showMainWindow = () => {
-  const mw = getMainWindow()
-  mw && mw.show()
-  showDockIcon()
+  SafeElectron.getApp().emit('KBkeybase', '', {type: 'showMainWindow'})
```
chrisnojima: we just want it shown. main thread handles dock management

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -68,28 +62,6 @@ export const getContentTypeFromURL = (
   req.end()
 }
 
-const writeElectronSettingsOpenAtLogin = (
```
chrisnojima: 🔥

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -130,24 +102,6 @@ function* initializeInputMonitor(): Iterable<any> {
   }
 }
 
-// get this value from electron and update our store version
-function* initializeAppSettingsState(): Iterable<any> {
-  const getAppState = () =>
-    new Promise(resolve => {
-      SafeElectron.getApp().once(
```
chrisnojima: 🔥

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -397,17 +422,19 @@ export function* platformConfigSaga(): Saga.SagaGenerator<any, any> {
   yield* Saga.chainAction2(ConfigGen.updateNow, updateNow)
   yield* Saga.chainAction2(ConfigGen.checkForUpdate, checkForUpdate)
   yield* Saga.chainAction2(ConfigGen.daemonHandshakeWait, sendKBServiceCheck)
-  yield* Saga.chainAction2(SettingsGen.onChangeUseNativeFrame, setUseNativeFrame)
+  yield* Saga.chainAction2(ConfigGen.setUseNativeFrame, saveUseNativeFrame)
```
chrisnojima: we write to the rpc when this changes

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -397,17 +422,19 @@ export function* platformConfigSaga(): Saga.SagaGenerator<any, any> {
...
   yield* Saga.chainAction2(ConfigGen.loggedIn, initOsNetworkStatus)
+  yield* Saga.chainAction2(ConfigGen.updateWindowState, saveWindowState)
 
   if (isWindows) {
     yield* Saga.chainGenerator<ConfigGen.DaemonHandshakePayload>(ConfigGen.daemonHandshake, checkRPCOwnership)
   }
 
   yield Saga.spawn(initializeUseNativeFrame)
+  yield Saga.spawn(initializeNotifySound)
```
chrisnojima: we init from the rpc on start

shared/app/app-state.d.ts
```diff
@@ -1,21 +0,0 @@
-export type State = {
```
chrisnojima: killing this file is essentially the point of this whole thing

shared/app/app-state.d.ts
```diff
@@ -1,21 +0,0 @@
...
-  x: number | null
-  y: number | null
-  width: number
-  height: number
-  windowHidden: boolean
-  isMaximized: boolean | null
-  isFullScreen: boolean | null
-  displayBounds: any | null
-  tab: string | null
```
chrisnojima: never actually used

shared/desktop/app/dock-icon.desktop.tsx
```diff
@@ -1,22 +0,0 @@
-import * as SafeElectron from '../../util/safe-electron.desktop'
-
-const changeDock = (show: boolean) => {
```
chrisnojima: having a standalone file for this tiny function that's only imported from one place hurts more than it helps

shared/desktop/app/menu-bar.desktop.tsx
```diff
@@ -6,12 +6,19 @@ import {resolveImage, resolveRootAsURL} from './resolve-root.desktop'
 import {showDevTools, skipSecondaryDevtools} from '../../local-debug.desktop'
 import logger from '../../logger'
 
-const htmlFile = resolveRootAsURL('dist', `menubar${__DEV__ ? '.dev' : ''}.html`)
+const htmlFile = resolveRootAsURL('dist', `menubar${__DEV__ ? '.dev' : ''}.html?param=`)
```
chrisnojima: so the url scheme matches what we do elsewhere. the main thread finds things from url now

shared/desktop/app/menu-helper.desktop.tsx
```diff
@@ -1,16 +1,14 @@
...
 const windowQuit = () => {
-  SafeElectron.getApp().emit('close-windows')
-  hideDockIcon()
+  Electron.app.emit('KBkeybase', '', {type: 'closeWindows'})
 }
 
-export default function makeMenu(window: any) {
-  const editMenu = {
+export default function makeMenu(window: Electron.BrowserWindow) {
+  const editMenu = new Electron.MenuItem({
```
chrisnojima: the types that ship with electron only allow partial menu definitions if using the constructor. else you have to pass every value in, and some types are wider than they should be (see `role`), so we just use the constructor version

shared/desktop/app/menu-helper.desktop.tsx
```diff
@@ -21,58 +19,58 @@ export default function makeMenu(window: any) {
...
-      {label: 'Reset zoom ', role: 'resetZoom'},
-      {label: 'Minimize', role: 'minimize'},
-      {accelerator: 'CmdOrCtrl+W', label: 'Close', role: 'close'},
-      {type: 'separator'},
-      {label: 'Bring All to Front', role: 'front'},
-    ].concat(
-      __DEV__
+    submenu: Electron.Menu.buildFromTemplate([
+      new Electron.MenuItem({accelerator: 'CmdOrCtrl+=', label: 'Zoom In', role: 'zoomin'}),
+      new Electron.MenuItem({label: 'Zoom Out', role: 'zoomout'}),
```
chrisnojima: note the case error previously! `zoomout !== zoomOut`

shared/desktop/app/node.desktop.tsx
```diff
@@ -211,9 +244,68 @@ const plumbEvents = () => {
           startWinService()
         }
         break
-      case 'requestShowDockIcon':
+      case 'closeWindows': {
```
chrisnojima: vs before, the main renderer would make the other renderers, but really we want the node thread to do this work.

shared/desktop/app/node.desktop.tsx
```diff
@@ -211,9 +244,68 @@ const plumbEvents = () => {
...
+
+        break
+      }
+      case 'closeRenderer': {
+        const w = findRemoteComponent(action.payload.windowComponent, action.payload.windowParam)
+        w && w.close()
         break
+      }
+      case 'rendererNewProps': {
+        const w = findRemoteComponent(action.payload.windowComponent, action.payload.windowParam)
```
chrisnojima: shuttling the props between the windows is now the role of the main node thread

shared/desktop/app/window.desktop.tsx
```diff
@@ -1,67 +0,0 @@
-import {showDockIcon, hideDockIcon} from './dock-icon.desktop'
-import menuHelper from './menu-helper.desktop'
-import * as SafeElectron from '../../util/safe-electron.desktop'
-
-export default class Window {
```
chrisnojima: this is only used by main window and i've always found it confusing

shared/desktop/remote/component-loader.desktop.tsx
```diff
@@ -7,15 +7,14 @@ import ReactDOM from 'react-dom'
...
 import {disable as disableDragDrop} from '../../util/drag-drop'
-import {setupContextMenu} from '../app/menu-helper.desktop'
 import ErrorBoundary from '../../common-adapters/error-boundary'
 import {initDesktopStyles} from '../../styles/index.desktop'
 
 disableDragDrop()
 
 module.hot && module.hot.accept()
 
-type RemoteComponents = 'unlock-folders' | 'menubar' | 'pinentry' | 'tracker' | 'tracker2'
```
chrisnojima: we killed the old tracker awhile ago

shared/desktop/remote/sync-avatar-props.desktop.tsx
```diff
@@ -2,14 +2,12 @@
...
 import * as I from 'immutable'
 import * as React from 'react'
-import * as SafeElectron from '../../util/safe-electron.desktop'
 import * as Container from '../../util/container'
 import {memoize} from '../../util/memoize'
 
 type OwnProps = {
   usernames: I.Set<string>
   setUsernames: (arg0: I.Set<string>) => void
-  remoteWindow: SafeElectron.BrowserWindowType | null
```
chrisnojima: all these remote window hocs used to actually have a direct instance of the browerwindow. instead we just use the `windowComponent` and `windowParam` as a unique identifier and delegate to the node process

shared/desktop/remote/sync-browser-window.desktop.tsx
```diff
@@ -13,119 +10,35 @@ type Props = {
...
-        return
-      }
-      const webContents = this._remoteWindow.webContents
-      if (showDevTools && !skipSecondaryDevtools) {
-        webContents.openDevTools({mode: 'detach'})
+  const RemoteWindowComponent = (props: Props) => {
+    React.useEffect(() => {
+      const {windowOpts, windowComponent, windowParam, windowPositionBottomRight} = props
+
+      SafeElectron.getApp().emit('KBkeybase', '', {
```
chrisnojima: we ask the node thread to make and destroy this window when this component is mounted/unmounted (vs doing it ourselves before)

shared/desktop/renderer/main2.desktop.tsx
```diff
@@ -12,7 +12,6 @@ import configureStore from '../../store/configure-store'
 import * as SafeElectron from '../../util/safe-electron.desktop'
 import {makeEngine} from '../../engine'
 import {disable as disableDragDrop} from '../../util/drag-drop'
-import {setupContextMenu} from '../app/menu-helper.desktop'
```
chrisnojima: the menuHelper default export actually does this itself

shared/pinentry/index.desktop.tsx
```diff
@@ -82,10 +82,9 @@ class Pinentry extends React.Component<Props, State> {
           ...globalStyles.flexBoxColumn,
           backgroundColor: globalColors.white,
           paddingBottom: globalMargins.medium,
-          paddingTop: globalMargins.medium,
```
chrisnojima: this is a random fix that fixes an issue @mlsteele found. i was already in here so...

shared/pinentry/index.desktop.tsx
```diff
@@ -82,10 +82,9 @@ class Pinentry extends React.Component<Props, State> {
...
         }}
       >
-        <Header icon={true} title="" onClose={this.props.onCancel} />
+        <Header icon={false} title="" onClose={this.props.onCancel} windowDragging={true} />
```
chrisnojima: icon off as it overlaps w/ the system close stuff

chrisnojima:
- [ ] after release
- [ ] test linux
- [ ] test windows
- [ ] mobile

chrisnojima:
@keybase/react-hackers this'll go in after the release. we can review it after. will need to test linux/windows 
cc: @heronhaye

shared/desktop/app/main-window.desktop.tsx
```diff
@@ -32,75 +35,166 @@ export default function() {
...
+
+let useNativeFrame = defaultUseNativeFrame
+
+const loadWindowState = () => {
+  const filename = dataRoot + 'gui_config.json'
 
-  const openedAtLogin = app.getLoginItemSettings().wasOpenedAtLogin
+  let openAtLogin = true
+  try {
+    const s = fs.readFileSync(filename, {encoding: 'utf8'})
```
heronhaye: how will this work with sandboxing?
also, does this pr migrate your old setting?

PR #19031, opened 2019-08-21
team-building tab bar desktop

mlsteele:
@keybase/picnicsquad @keybase/react-hackers 

image

### DIFF
--- a/shared/team-building/container.tsx
+++ b/shared/team-building/container.tsx
@@ -638,7 +638,7 @@ class StateWrapperForTeamBuilding extends React.Component<RealOwnProps, LocalSta
         highlightedIndex={this.state.highlightedIndex}
         changeShowRolePicker={this.changeShowRolePicker}
         showRolePicker={this.state.showRolePicker}
-        showServiceResultCount={this.state.searchString !== ''}
+        showServiceResultCount={false}
       />
     )
   }
--- a/shared/team-building/service-tab-bar.desktop.tsx
+++ b/shared/team-building/service-tab-bar.desktop.tsx
@@ -220,7 +220,7 @@ const styles = Styles.styleSheetCreate(() => ({
     marginRight: Styles.globalMargins.xtiny,
   },
   serviceIconBox: {
-    marginTop: Styles.globalMargins.xsmall, // xxx
+    marginTop: 14,
   },
   serviceIconContainer: {
     flex: 1,
@@ -234,7 +234,6 @@ const styles = Styles.styleSheetCreate(() => ({
     justifyContent: 'flex-start',
   },
   tabBarContainer: {
-    marginTop: Styles.globalMargins.xtiny,
     minHeight: 30,
   },
 }))
--- a/shared/team-building/service-tab-bar.desktop.tsx
+++ b/shared/team-building/service-tab-bar.desktop.tsx
@@ -72,20 +72,23 @@ const MoreNetworksButton = Kb.OverlayParentHOC(
     }>
   ) => (
     <>
-      <Kb.Box2
-        direction="vertical"
-        style={styles.moreNetworks1}
-        fullHeight={true}
-        centerChildren={true}
-        ref={props.setAttachmentRef}
-      >
-        <Kb.WithTooltip text="More networks" containerStyle={styles.moreNetworks2}>
-          <Kb.ClickableBox onClick={props.toggleShowingMenu} style={styles.moreNetworks3}>
-            <Kb.Text type="BodyBigExtrabold" style={styles.moreText}>
-              •••
-            </Kb.Text>
-          </Kb.ClickableBox>
-        </Kb.WithTooltip>
+      <Kb.Box2 direction="vertical" fullHeight={true} fullWidth={true} style={styles.moreNetworks0}>
+        <Kb.Box2
+          direction="vertical"
+          style={styles.moreNetworks1}
+          fullHeight={true}
+          centerChildren={true}
+          ref={props.setAttachmentRef}
+        >
+          <Kb.WithTooltip text="More networks" containerStyle={styles.moreNetworks2}>
+            <Kb.ClickableBox onClick={props.toggleShowingMenu} style={styles.moreNetworks3}>
+              <Kb.Text type="BodyBigExtrabold" style={styles.moreText}>
+                •••
+              </Kb.Text>
+            </Kb.ClickableBox>
+          </Kb.WithTooltip>
+        </Kb.Box2>
+        <Kb.Box2 direction="horizontal" fullWidth={true} style={styles.inactiveTabBar} />
       </Kb.Box2>
       <Kb.FloatingMenu
         attachTo={props.getAttachmentRef}
@@ -104,11 +107,7 @@ const MoreNetworksButton = Kb.OverlayParentHOC(
 
 const MoreNetworkItem = (props: {service: ServiceIdWithContact}) => (
   <Kb.Box2 direction="horizontal" fullHeight={true} alignItems="center">
-    <Kb.Icon
-      style={{marginRight: Styles.globalMargins.tiny}}
-      fontSize={16}
-      type={serviceIdToIconFont(props.service)}
-    />
+    <Kb.Icon style={styles.moreNetworkItemIcon} fontSize={16} type={serviceIdToIconFont(props.service)} />
     <Kb.Text type="Body">{serviceIdToLongLabel(props.service)}</Kb.Text>
   </Kb.Box2>
 )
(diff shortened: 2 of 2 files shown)


shared/team-building/container.tsx
```diff
@@ -638,7 +638,7 @@ class StateWrapperForTeamBuilding extends React.Component<RealOwnProps, LocalSta
         highlightedIndex={this.state.highlightedIndex}
         changeShowRolePicker={this.changeShowRolePicker}
         showRolePicker={this.state.showRolePicker}
-        showServiceResultCount={this.state.searchString !== ''}
+        showServiceResultCount={false}
```
nathunsmitty: Should we rip out all of the result count code if we never want to show counts anymore? 

I think it'd make a lot of sense to remove it in this PR if so, but I'm okay if you want to punt that to a separate PR and just leave a TODO here.

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,76 +4,155 @@ import * as Styles from '../styles'
...
+        }))}
+        onHidden={props.toggleShowingMenu}
+        visible={props.showingMenu}
+      />
+    </>
+  )
+)
+
+const MoreNetworkItem = (props: {service: ServiceIdWithContact}) => (
+  <Kb.Box2 direction="horizontal" fullHeight={true} alignItems={'center'}>
```
nathunsmitty:
```suggestion
  <Kb.Box2 direction="horizontal" fullHeight={true} alignItems="center">
```

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,76 +4,155 @@ import * as Styles from '../styles'
...
+          service={service}
+          label={serviceIdToLongLabel(service)}
+          labelPresence={1}
+          onClick={() => props.onChangeService(service)}
+          count={undefToNull(props.serviceResultCount[service])}
+          showCount={props.showServiceResultCount}
+          isActive={props.selectedService === service}
+        />
+      ))}
+      {!!(moreServices.length > 0) && (
```
nathunsmitty: Why do we need to double negate this expression?

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,76 +4,155 @@ import * as Styles from '../styles'
...
   serviceIdToLongLabel,
   inactiveServiceAccentColor,
 } from './shared'
 import * as Constants from '../constants/team-building'
 import {ServiceIdWithContact} from '../constants/types/team-building'
 import {Props, IconProps} from './service-tab-bar'
 
-type ExtraProps = {service: ServiceIdWithContact; isActive: boolean}
-const HoverIcon = Styles.styled<typeof Kb.Icon, ExtraProps>(Kb.Icon)(props => ({
-  '&:hover': {
```
nathunsmitty: Is tracking hover via state instead of using CSS pseudo-selectors the preferred way of doing hover styles now?

shared/team-building/container.tsx
mlsteele: I'm in favor of punting because there's a chance we'll want to resurrect result counts while iterating on the design. Just created PICNIC-368 to remove it later.

shared/team-building/service-tab-bar.desktop.tsx
mlsteele: Before this patch there was just an icon. Now there's an icon and label. Hovering over the icon _or label_ should cause the icon to light up. It wasn't clear to me how to use `&:hover` for that.

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,76 +4,156 @@ import * as Styles from '../styles'
...
-        onClick={() => props.onChangeService(service)}
-        count={undefToNull(props.serviceResultCount[service])}
-        showCount={props.showServiceResultCount}
-        isActive={props.selectedService === service}
-      />
-    ))}
-  </Kb.Box2>
-)
+export const ServiceTabBar = (props: Props) => {
+  const [lastSelectedUnlockedService, setState] = React.useState<ServiceIdWithContact | null>(null)
```
nathunsmitty:
```suggestion
  const [lastSelectedUnlockedService, setSelectedUnlockedService] = React.useState<ServiceIdWithContact | null>(null)
```

think this is a little more descriptive as to what state you're setting 🙂

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,76 +4,156 @@ import * as Styles from '../styles'
...
+  const nLocked = 3 // Services always out front on the left. Add one to get the number out front.
+  const lockedServices = Constants.services.slice(0, nLocked)
+  let frontServices = lockedServices
+  if (Constants.services.indexOf(props.selectedService) < nLocked) {
+    // Selected service is locked
+    if (lastSelectedUnlockedService === null) {
+      frontServices = lockedServices.concat(Constants.services.slice(nLocked, nLocked + 1))
```
nathunsmitty: I think
```suggestion
      frontServices = Constants.services.slice(0, nLocked+1)
```

is a little clearer, both in terms of length and what's going on.

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,76 +4,156 @@ import * as Styles from '../styles'
...
+    } else {
+      frontServices = lockedServices.concat([lastSelectedUnlockedService])
+    }
+  } else {
+    frontServices = lockedServices.concat([props.selectedService])
+    if (lastSelectedUnlockedService !== props.selectedService) {
+      setState(props.selectedService)
```
nathunsmitty: We don't really ever want to setState or do other side effects in the render part of a component. What would you think handling this in the `onClick` callback of `ServiceIcon`?

Example:
```tsx
const onServiceIconClick = React.useCallback(service => {
  if (!lockedServices.includes(service)) {
    setSelectedUnlockedService(service)
  }
  props.onChangeService(service)
}, ...)
```

I realize the typing for this probably isn't correct, but hopefully that's enough of a starting point. Feel free to ping me too!

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,76 +4,156 @@ import * as Styles from '../styles'
...
+    }
+  }
+  const moreServices = difference(Constants.services, frontServices)
+  return (
+    <Kb.Box2 direction="horizontal" fullWidth={true} style={styles.tabBarContainer}>
+      {frontServices.slice(0, 4).map(service => (
```
nathunsmitty: I thought `frontServices` was guaranteed to have only 4 elements from the above code?

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,76 +4,156 @@ import * as Styles from '../styles'
...
+        />
+      ))}
+      {moreServices.length > 0 && (
+        <MoreNetworksButton services={moreServices} onChangeService={props.onChangeService} />
+      )}
+    </Kb.Box2>
+  )
+}
 
 const styles = Styles.styleSheetCreate({
```
nathunsmitty: let's make this a closure, since I hear that's what we're doing

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -85,24 +165,65 @@ const styles = Styles.styleSheetCreate({
...
+  moreText: {
+    color: Styles.globalColors.black_50,
+  },
   pendingIcon: {height: 10, width: 10},
-  resultCount: {},
   serviceIcon: {
     marginRight: Styles.globalMargins.xtiny,
   },
+  serviceIconBox: {
+    marginTop: 14,
```
nathunsmitty: I'm always a little hesitant to make margins that aren't part of `globalMargins`. Does this need to be 12 or can we round it to be `xsmall` or `small`?

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,76 +4,156 @@ import * as Styles from '../styles'
...
+        visible={props.showingMenu}
+      />
+    </>
+  )
+)
+
+const MoreNetworkItem = (props: {service: ServiceIdWithContact}) => (
+  <Kb.Box2 direction="horizontal" fullHeight={true} alignItems="center">
+    <Kb.Icon
+      style={{marginRight: Styles.globalMargins.tiny}}
```
nathunsmitty: Let's put this in `styles`

shared/team-building/service-tab-bar.desktop.tsx
mlsteele: What's the issue with setState in render? Loops?

That looks nice. One thing about it, that wouldn't actually come up in the current setup, is that it would miss changes to props.selectedService that happened due to some other trigger.

What do you think about something like this?
```
React.useEffect(
  () = setSelectedUnlockedService(props.selectedService),
  [props.selectedService])
```

nathunsmitty: From the [React docs](https://reactjs.org/docs/react-component.html#render):

> The render() function should be pure, meaning that it does not modify component state, it returns the same result each time it’s invoked, and it does not directly interact with the browser.

The main worry is infinite loops, since `setState` triggers a re-render meaning you're borked if the same code path evaluates twice.

shared/team-building/service-tab-bar.desktop.tsx
mlsteele: 14 matches the design exactly

shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -4,106 +4,238 @@ import * as Styles from '../styles'
...
+    frontServices = lockedServices.concat([props.selectedService])
+    if (lastSelectedUnlockedService !== props.selectedService) {
+      lastSelectedUnlockedServiceChanged = true
+    }
+  }
+  React.useEffect(() => {
+    if (lastSelectedUnlockedServiceChanged) {
+      setLastSelectedUnlockedDevice(props.selectedService)
+    }
+  }, [lastSelectedUnlockedServiceChanged, props.selectedService])
```
mlsteele: Here's a shot with `useEffect`. I just had to see if it would work. If you still prefer putting the setState in onClick lmk

nathunsmitty: I think I'd prefer setting the state in onClick. I believe the difference is:

### useEffect
1. User selects icon
2. `onChangeService` fires, dispatches action
3. Redux state updated
4. New selected service passed down from store to `ServiceTabBar`, triggering re-render
5. `lastSelectedUnlockedServiceChanged` set to true
6. `setLastSelectedUnlockedService` queued and triggered after render completes, triggering a re-render
7.  `lastSelectedUnlockedServiceChanged` set to false, no re-render

### onClick
1. User selects icon
2. `setLastSelectedUnlockedService` queued
2. `onChangeService` fires, dispatches action
3. Redux state updated
4. New selected service passed down from store to `ServiceTabBar`
5. State of `ServiceTabBar` updated
6. I believe 4 and 5 will be batched, triggering only one re-render?


I could totally be wrong though. I suppose an alternative to either of these approaches to would be to just put the last selected unlocked service in redux and just treat it as another prop here. This has the advantage of persisting state through loads of the new group. It should also trigger only one re-render.

mlsteele: This line all by itself causes render to happen twice every time a service is changed.
```React.useEffect(() => { }, [props.selectedService])```

mlsteele: You're right that calling setState in useEffect world caused it to render 4x instead of 2x times when clicking something that changed the state.

mlsteele: This line also causes render to happen twice all by itself. Must be something about hooks
```const [state, setState] = React.useState(null)```

PR #19033, opened 2019-08-21
cr_chains: handle renames correctly for deleted entries

strib:
When sending out edit notifications using the list of ops in a MD update, the `crChains` instance does a trick where it combines separate delete+create ops back into rename operations, so that they can be faithfully reported as renames in the edit notifications chat stream.

However, this wasn't working correctly for entries that were renamed, and then deleted, within the same update.  In that case, there was no original create op still left in the ops list, and so the rename op wasn't getting added back into the resulting list.

This causes problems for the edit history.  Say there is a directory with modified files inside it.  If the directory is renamed to something else, and then the whole directory structure is deleted, the edit history processor won't know to ignore those old modifications of those files when they were in the old directory name, because it never saw the rename operation and so thinks those old files are still there.

This led to a panic in the case where someone created a directory, added files to it, then renamed it and created a file with the same old name as the directory.  The old modified files were still in the recent history list, but now it looks like they are somehow subentries under a file entry, which is impossible. The `folderBranchOps` partial sync code wasn't expecting this and panicked.

So the fix for the future is to make sure that the rename op always shows up properly in the edit notification stream.

Also: fix the panic itself by not allowing lookups in non-dir nodes.

Issue: HOTPOT-616

### DIFF
--- a/go/kbfs/kbfsedits/tlf_history_test.go
+++ b/go/kbfs/kbfsedits/tlf_history_test.go
@@ -27,10 +27,10 @@ func checkTlfHistory(t *testing.T, th *TlfHistory, expected writersByRevision,
 		require.Equal(t, e.writerName, history[i].writerName)
 		require.Len(t, history[i].notifications, len(e.notifications))
 		for j, n := range e.notifications {
-			require.Equal(t, n, history[i].notifications[j])
+			require.Equal(t, n, history[i].notifications[j], i)
 		}
 		if len(e.notifications) < maxEditsPerWriter {
-			require.Contains(t, writersWhoNeedMore, e.writerName)
+			require.Contains(t, writersWhoNeedMore, e.writerName, i)
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


go/kbfs/kbfsedits/tlf_history_test.go
```diff
@@ -440,6 +440,63 @@ func TestTlfHistoryRenameParentSimple(t *testing.T) {
...
+
+	// Alice renames x to a.
+	_ = nn.makeWithType(
+		"/k/p/a,b/a", NotificationRename, aliceUID, &NotificationParams{
+			OldFilename: "/k/p/a,b/x",
+		}, time.Time{}, EntryTypeFile)
+	aliceMessages = append(aliceMessages, nn.encode(t))
+	aliceCreateA.Filename = "/k/p/a,b/a"
+
+	// Alice creates modifies file "a".
```
jakob223: same as above

PR #19034, opened 2019-08-21
Quick reply chat notifications [Android]

MarcoPolo:
@keybase/react-hackers @jzila @joshblum 

This is the first pass so it's not as nice as it could be, but I think it's a good enough start.

Screen Shot 2019-08-21 at 2 39 00 PM
Screen Shot 2019-08-21 at 2 39 06 PM (goes away after 1s)


go/bind/notifications.go
```diff
@@ -47,6 +48,19 @@ type ChatNotification struct {
...
+func HandlePostTextReply(strConvID, tlfName string, body string, replyTo int) (err error) {
+	outboxID, err := storage.NewOutboxID()
+	if err != nil {
+		return err
+	}
+	convID, err := chat1.MakeConvID(strConvID)
+	if err != nil {
+		return err
+	}
+	kbCtx.ChatHelper.SendTextByIDNonblock(context.Background(), convID, tlfName, body, &outboxID, nil)
```
joshblum: don't you need `&chat1.MessageID(replyTo)` here?

MarcoPolo: That's actually old code, I originally did a reply to, but realized that isn't how anyone else does it and it felt weird. So I just post the message without making it a reply

joshblum: oh why not drop the `replyTo` param then?

shared/android/app/src/main/java/io/keybase/ossifrage/ChatBroadcastReceiver.java
```diff
@@ -0,0 +1,87 @@
...
+import android.support.v4.app.NotificationManagerCompat;
+
+import keybase.Keybase;
+
+public class ChatBroadcastReceiver extends BroadcastReceiver {
+  public static String KEY_TEXT_REPLY = "key_text_reply";
+
+  @RequiresApi(api = Build.VERSION_CODES.KITKAT_WATCH)
+  private String getMessageText(Intent intent) {
+    Bundle remoteInput = RemoteInput.getResultsFromIntent(intent);
```
jzila: As Nojima asked previously, is there any chance this can be intercepted by a different app?

MarcoPolo: No, It uses an explicit intent (specifes what class to use for the intent) So it should stay within the application.
https://developer.android.com/training/articles/security-tips#use-intents 
> For security purposes, explicit intents are preferred.

go/bind/notifications.go
```diff
@@ -47,6 +48,19 @@ type ChatNotification struct {
...
+func HandlePostTextReply(strConvID, tlfName string, body string) (err error) {
+	outboxID, err := storage.NewOutboxID()
+	if err != nil {
+		return err
+	}
+	convID, err := chat1.MakeConvID(strConvID)
+	if err != nil {
+		return err
+	}
+	kbCtx.ChatHelper.SendTextByIDNonblock(context.Background(), convID, tlfName, body, &outboxID, nil)
```
mmaxim: You can pass `nil` for `outboxID` here to make this simpler.

go/bind/notifications.go
```diff
@@ -34,11 +35,11 @@ type Message struct {
 }
 
 type ChatNotification struct {
-	// Ordered from oldest to newest
 	Message   *Message
 	ConvID    string
 	TeamName  string
 	TopicName string
+	TlFName   string
```
mmaxim: Does this field name pass `go vet`?

MarcoPolo: Apparently! Want me to change it?

mmaxim: It looks like it was probably a typo, so I think we should.

MarcoPolo:
Oh yeah I see it now

go/bind/notifications.go
MarcoPolo: Oh, I see the typo now

PR #19042, opened 2019-08-21
update linux builds to go1.12.9

heronhaye:


joshblum:
can you update the docker CI image as well?

PR #19043, opened 2019-08-21
client: de-lint

strib:
Issue: HOTPOT-638

### DIFF
--- a/go/client/cmd_ctl.go
+++ b/go/client/cmd_ctl.go
@@ -24,9 +24,7 @@ func NewCmdCtl(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
 		NewCmdCtlAppExit(cl, g),
 	}
 
-	for _, cmd := range platformSpecificCtlCommands(cl, g) {
-		commands = append(commands, cmd)
-	}
+	commands = append(commands, platformSpecificCtlCommands(cl, g)...)
 
 	return cli.Command{
 		Name:        "ctl",
@@ -36,7 +34,7 @@ func NewCmdCtl(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
 }
 
 // availableComponents specify which components can be included or excluded
-var availableCtlComponents = []string{
+var availableCtlComponents = []string{ //nolint
 	install.ComponentNameApp.String(),
 	install.ComponentNameService.String(),
 	install.ComponentNameKBFS.String(),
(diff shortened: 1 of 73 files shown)


go/client/cmd_ctl.go
```diff
@@ -36,15 +34,15 @@ func NewCmdCtl(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
 }
 
 // availableComponents specify which components can be included or excluded
-var availableCtlComponents = []string{
+var availableCtlComponents = []string{ //nolint
```
joshblum: just curious: why the `nolint` for these?

go/client/rpc.go
```diff
@@ -24,7 +24,7 @@ func getSocket(g *libkb.GlobalContext, clearError bool) (xp rpc.Transporter, err
 	var isNew bool
 	_, xp, isNew, err = g.GetSocket(clearError)
 	if err == nil && isNew {
-		introduceMyself(g, xp)
+		_ = introduceMyself(g, xp)
```
joshblum: why not return this error?

go/client/ui.go
```diff
@@ -770,34 +770,34 @@ func (l LoginUI) DisplayPaperKeyPhrase(_ context.Context, arg keybase1.DisplayPa
...
-	l.parent.Printf("Your paper key is\n\n")
-	l.parent.Printf("\t")
-	l.parent.OutputDesc(OutputDescriptorPrimaryPaperKey, arg.Phrase)
-	l.parent.Printf("\n\n")
-	l.parent.Printf("Write it down....now!\n\n")
+	_, _ = l.parent.Printf("\n")
+	_, _ = l.parent.Printf("===============================\n")
+	_, _ = l.parent.Printf("IMPORTANT: PAPER KEY GENERATION\n")
+	_, _ = l.parent.Printf("===============================\n\n")
+	_, _ = l.parent.Printf("During Keybase's alpha, everyone gets a paper key. This is a private key.\n")
```
joshblum:
```suggestion
	_, _ = l.parent.Printf("This is a private key.\n")
```

go/client/cmd_ctl.go
strib: They are only used when the build tag is `darwin`, and so seem unused when linting on Linux.  I guess we could move them into a darwin-only file but this was faster to do at the time.

go/client/rpc.go
strib: Good question, I don't remember having a reason.  I'll change it.

strib:
Oops I merged this during code freeze by accident.  I've reverted it and will open a new PR.

PR #19045, opened 2019-08-21
engine: de-lint

strib:
Issue: HOTPOT-639

### DIFF
--- a/go/engine/bootstrap_test.go
+++ b/go/engine/bootstrap_test.go
@@ -7,6 +7,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/keybase/client/go/libkb"
+	"github.com/stretchr/testify/require"
 )
 
 func TestBootstrap(t *testing.T) {
@@ -19,9 +20,8 @@ func TestBootstrap(t *testing.T) {
 
 	// do a upak load to make sure it is cached
 	arg := libkb.NewLoadUserByUIDArg(context.TODO(), tc.G, u1.UID())
-	if _, _, err := tc.G.GetUPAKLoader().Load(arg); err != nil {
-		t.Fatal(err)
-	}
+	_, _, err := tc.G.GetUPAKLoader().Load(arg)
+	require.NoError(t, err)
 
 	// get the status values
 	uid := tc.G.Env.GetUID()
@@ -37,9 +37,8 @@ func TestBootstrap(t *testing.T) {
 	prev := os.Getenv("KEYBASE_SERVER_URI")
 	os.Setenv("KEYBASE_SERVER_URI", "http://127.0.0.127:3333")
 	defer os.Setenv("KEYBASE_SERVER_URI", prev)
-	if err := tc.G.ConfigureAPI(); err != nil {
-		t.Fatal(err)
-	}
+	err = tc.G.ConfigureAPI()
+	require.NoError(t, err)
 	tc.G.ConnectivityMonitor = OfflineConnectivityMonitor{}
 
 	eng := NewLoginOffline(tc.G)
--- a/go/engine/bootstrap_test.go
+++ b/go/engine/bootstrap_test.go
@@ -19,7 +19,9 @@ func TestBootstrap(t *testing.T) {
 
 	// do a upak load to make sure it is cached
 	arg := libkb.NewLoadUserByUIDArg(context.TODO(), tc.G, u1.UID())
-	tc.G.GetUPAKLoader().Load(arg)
+	if _, _, err := tc.G.GetUPAKLoader().Load(arg); err != nil {
+		t.Fatal(err)
+	}
 
 	// get the status values
 	uid := tc.G.Env.GetUID()
@@ -35,7 +37,9 @@ func TestBootstrap(t *testing.T) {
 	prev := os.Getenv("KEYBASE_SERVER_URI")
 	os.Setenv("KEYBASE_SERVER_URI", "http://127.0.0.127:3333")
 	defer os.Setenv("KEYBASE_SERVER_URI", prev)
-	tc.G.ConfigureAPI()
+	if err := tc.G.ConfigureAPI(); err != nil {
+		t.Fatal(err)
+	}
 	tc.G.ConnectivityMonitor = OfflineConnectivityMonitor{}
 
 	eng := NewLoginOffline(tc.G)
--- a/go/engine/login_provision.go
+++ b/go/engine/login_provision.go
@@ -342,7 +342,7 @@ func (e *loginProvision) paper(m libkb.MetaContext, device *libkb.Device, keys *
 	m.ActiveDevice().CacheProvisioningKey(m, keys)
 
 	e.saveToSecretStore(m)
-	return
+	return nil
 }
 
 var paperKeyNotFound = libkb.NotFoundError{
@@ -424,7 +424,7 @@ func (e *loginProvision) pgpProvision(m libkb.MetaContext) (err error) {
 	}
 
 	e.saveToSecretStore(m)
-	return
+	return nil
 }
 
 // makeDeviceKeysWithSigner creates device keys given a signing
--- a/go/engine/login_provision.go
+++ b/go/engine/login_provision.go
@@ -144,15 +144,16 @@ func (e *loginProvision) Run(m libkb.MetaContext) error {
 	return nil
 }
 
-func (e *loginProvision) saveToSecretStore(m libkb.MetaContext) error {
-	return e.saveToSecretStoreWithLKS(m, e.lks)
+func (e *loginProvision) saveToSecretStore(m libkb.MetaContext) {
+	e.saveToSecretStoreWithLKS(m, e.lks)
 }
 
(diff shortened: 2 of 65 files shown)


go/engine/login_provision.go
```diff
@@ -338,8 +343,7 @@ func (e *loginProvision) paper(m libkb.MetaContext, device *libkb.Device, keys *
 	// the paper key on the global and not thread-local active device.
 	m.ActiveDevice().CacheProvisioningKey(m, keys)
 
-	e.saveToSecretStore(m)
```
patrickxb: This could have been intentional...I can't remember exactly, but there were some situations where we didn't want to block provisioning if there was an error saving to secret store.

patrickxb: The fact that it's like this everywhere makes me think that is the case...I would be a little worried to make this change.

go/engine/login_provision.go
```diff
@@ -420,8 +424,7 @@ func (e *loginProvision) pgpProvision(m libkb.MetaContext) (err error) {
 		return err
 	}
 
-	e.saveToSecretStore(m)
-	return nil
```
patrickxb: see above

go/engine/login_provision.go
```diff
@@ -964,7 +967,9 @@ func (e *loginProvision) tryGPG(m libkb.MetaContext) (err error) {
 		}
 		return err
 	}
-	e.saveToSecretStore(m)
```
patrickxb: see above

go/engine/login_provision.go
```diff
@@ -1158,7 +1160,9 @@ func (e *loginProvision) makeEldestDevice(m libkb.MetaContext) error {
 	if err = e.makeDeviceKeys(m, args); err != nil {
 		return err
 	}
-	e.saveToSecretStore(m)
```
patrickxb: see above

go/engine/login_provisioned_device.go
```diff
@@ -266,7 +266,10 @@ func (e *LoginProvisionedDevice) run(m libkb.MetaContext) (err error) {
 		return err
 	}
 
-	e.runBug3964Repairman(m)
+	err = e.runBug3964Repairman(m)
+	if err != nil {
+		return err
```
patrickxb: i don't think this is fatal either...just log the error and keep going.

go/engine/paperprovision.go
```diff
@@ -54,7 +54,10 @@ func (e *PaperProvisionEngine) Run(m libkb.MetaContext) (err error) {
 	defer m.Trace("PaperProvisionEngine#Run", func() error { return err })()
 
 	// clear out any existing session:
-	e.G().Logout(m.Ctx())
+	err = e.G().Logout(m.Ctx())
+	if err != nil {
+		return err
```
patrickxb: again, i don't think this is fatal.

go/engine/login_provision.go
```diff
@@ -983,11 +988,9 @@ func (e *loginProvision) chooseGPGKeyAndMethod(m libkb.MetaContext) (*libkb.GpgP
 	// find any local private gpg keys that are in user's key family
 	matches, err := e.matchingGPGKeys(m)
 	if err != nil {
-		if _, ok := err.(libkb.NoSecretKeyError); ok {
```
patrickxb: why was this removed?

patrickxb: i think we should keep this and figure out why we aren't telling the user to do this...

patrickxb:
I urge great caution here and not just mechanical `if err != nil { return err }`.

maxtaco:
I'll proxy my review to @patrickxb since he got a head start on it.

go/engine/bootstrap_test.go
```diff
@@ -19,7 +19,9 @@ func TestBootstrap(t *testing.T) {
 
 	// do a upak load to make sure it is cached
 	arg := libkb.NewLoadUserByUIDArg(context.TODO(), tc.G, u1.UID())
-	tc.G.GetUPAKLoader().Load(arg)
+	if _, _, err := tc.G.GetUPAKLoader().Load(arg); err != nil {
```
joshblum: prefer `require.NoError(t, err)` throughout for these over consistency within a particular test function. Better readability and you get a stack trace if the test fails.

go/engine/login_provision.go
strib: @patrick it's just an empty if block though, keeping it doesn't do anything -- what code do you want me to put inside it?  A log statement?  I'm happy to file a bug for someone else to look at the reasoning behind it, but I'm not going to do any deep debugging as part of this PR.

go/engine/bootstrap_test.go
strib: @joshblum ok i'll change them all to require.  My usual reasoning here to prefer consistency is so that at some future date, the entire file can be moved over in a single git commit, which makes the git history more cohesive.  But I don't feel strongly about it, so will change all the `t.Fatal`s I added to `require.NoError`s.

go/engine/login_provision.go
strib: Cool, to prevent future misunderstandings I'll change that function to not return an error at all.  Maybe it was only doing so it log it in the `defer`, so I'll do that another way.

strib:
@patrick @joshblum thanks for the reviews, I've addressed the specific review comments.

> I'd prefer to undo any logic changes for missing error check before this goes in. If you'd like, you could log the errors. Or just ignore them. I added comments on some of them, but I think maybe we should just not change any logic flow.
>
> A lot of this code has been used by users millions of times and getting clean lint status isn't worth breaking it.

I'm happy to revert specifically-reviewed places where errors should be dropped, but I disagree that we shouldn't change all of them by default.  Over time the code will become unmaintainable if we let stuff like this stand.  The default should be to return the errors, and fix problems as they arise (and I'm happy to be involved in the debugging/fixing process).  Now is as good a time as any to rip the band-aid off.  I'll wait until after code freeze is over to merge this of course, so we get lots of time with it before the following release.

Please check out the current PR and let me know which other errors might be unsafe to return (and of course I'll fix any tests that broke as well).

joshblum:
@patrickxb I understand your concern for changes to this code but overall agree with @strib for trying to work through which errors were intentionally vs accidentally dropped. 

>  I view any bugs it does cause/expose as a good learning opportunity to refresh our knowledge of how this code works, and it will only improve our long term code, at a low risk of short term pain.

I agree with the sentiment above and am also happy to help fix any issues that fall out of these changes.

zapu:
Proposal: let's get low hanging fruits in, when error handling cannot be questioned (so e.g. test code), and make tickets for everything that we are not sure about. I looked through this PR and even though I'm familiar with the code, it's hard to make a decision about logic changes like that without getting a better view - just the call site around the change alone is not enough to judge.

go/engine/login_provision.go
zapu: this looks like a "TODO" comment left by someone that never evolved into actual code that would "tell the user they need to get a gpg key" onto this device.

patrickxb:
My request is something along the lines of what @zapu is suggesting, and I believe @chrisnojima has suggested elsewhere:

1. make a PR that satisfies the linter without any logic changes
2. if there are things that look potentially broken, make tickets to investigate
3. put the tickets in triage.

i think this will be a lot cleaner than a large PR that is delinting and changing legacy logic.

patrickxb:
As far as the comment that logic changes have already happened in other packages as part of this lint project, I didn't review any of those and would have suggested the same thing.  The ticket is to satisfy a linter, not change logic.

strib:
> As far as the comment that logic changes have already happened in other packages as part of this lint project, I didn't review any of those and would have suggested the same thing. 

Fair enough.  But given that all of those other packges are already done, and done successfully as far as we know, unless there is a specific reason not to do the same for this specific package, that seems like more of an argument to go ahead with this one for consistency's sake.  I definitely understand your reluctance, and any future "I told you so"s will be well warranted.

> The ticket is to satisfy a linter, not change logic.

I disagree.  The only point of satisfying the linter is to make sure the code is readable and maintainable, and if we are just papering over possible bugs, we shouldn't have the linter at all.  No one is going to go through and check all of the call sites one by one, -- you and zapu have already made that pretty clear.  Even if we make investigation tickets for them, they will never be high priority enough to put into a sprint, and even if somehow they make it into a sprint, there will be nothing better for someone to do then to just change it and see what happens, like we're doing here.  I think the only reasonable way to get this done is to do it all at once, like we have done for all the other packages, and see what happens.

By changing the code to return errors by default, we are purposely forcing a situation where the code is either readable, or we find/fix/document the reason it can't be more readable.  And as I sad above, this is a great time to do that.

That said, I won't merge without at least one green checkmark.

strib:
Seems we're at an impasse here.  Anyone else have further thoughts on this?  @maxtaco we might need your sagely guidance on this one -- should we make this package confirm with every other package and repo we have (see my arguments above), or should I unwind part of this PR to play it safe for this package only and leave unvetted errors specifically unchecked?

strib:
> I'm in favor of just getting this in, I don't think any of the individual changes are too scary.

Thanks!

PR #19046, opened 2019-08-22
fix moveOrCopy and share to Android app

songgao:


shared/actions/config/index.tsx
```diff
@@ -384,9 +383,13 @@ const routeToInitialScreen = (state: Container.TypedState) => {
 
     // A share
     if (state.config.startupSharePath) {
+      // TODO/FIXME: this seems unused. [ShareDataIntent] in
+      // actions/platform-specific/push.native.tsx happens at cold-start too,
+      // so maybe we can just use that. Though that happens before this
+      // (routeToInitialScreen), so in the end it just routes to the saved tab
+      // which gets rid of the destination-picker modal.
```
songgao: Any thought on this? Cold-start share still doesn't work and I think this is the reason.

go/kbfs/simplefs/simplefs.go
```diff
@@ -250,7 +250,7 @@ func populateIdentifyBehaviorIfNeeded(ctx context.Context, path1 *keybase1.Path,
 	if ib1 != nil && ib2 == nil {
 		return tlfhandle.MakeExtendedIdentify(ctx, *ib1)
 	}
-	if ib1 == ib2 {
+	if *ib1 == *ib2 {
```
songgao: bug from my identify plumbing changes 🙀

shared/actions/config/index.tsx
MarcoPolo: Cold start might not work currently, but I have a fix here: https://github.com/keybase/client/pull/19047

If it works on a non-cold start my fix should fix it

PR #19047, opened 2019-08-22
Fix Android not using Intent on startup

MarcoPolo:
@keybase/react-hackers @jzila 

This tries to take a notification action with Saga.take (or timesout). Also consolidates some of this logic with how iOS works.

I thought I could get away without this, but turns out we need it.

### DIFF
--- a/shared/actions/platform-specific/push.native.tsx
+++ b/shared/actions/platform-specific/push.native.tsx
@@ -348,8 +348,8 @@ function* _checkPermissions(action: ConfigGen.MobileAppStatePayload | null) {
 
 function* getStartupDetailsFromInitialPush() {
   const {push, pushTimeout}: {push: PushGen.NotificationPayload; pushTimeout: boolean} = yield Saga.race({
-    push: isAndroid ? Saga.take(PushGen.notification) : Saga.callPromise(getInitialPushiOS),
-    pushTimeout: Saga.delay(50),
+    push: Saga.callPromise(isAndroid ? getInitialPushAndroid : getInitialPushiOS),
+    pushTimeout: Saga.delay(10),
   })
   if (pushTimeout || !push) {
     return null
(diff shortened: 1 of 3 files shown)


buoyad:
Having trouble getting a notification to appear on iOS, testing master...

MarcoPolo:
@jzila I changed this significantly. Trying to listen for the event on startup is racy, so I was getting to fail randomly.

Now it registers the intial intent under the Engine Native Module, and the JS side can grab the intial intent whenever it wants.

See these for a bit more explanation:
https://github.com/keybase/client/pull/19047/files#diff-0644c201f67a2838f14ee08d35143737R228
and: https://github.com/keybase/client/pull/19047/files#diff-270c1a26c4443ac1cfc4c6a7b223b13fR52

shared/actions/platform-specific/push.native.tsx
```diff
@@ -347,43 +350,45 @@ function* _checkPermissions(action: ConfigGen.MobileAppStatePayload | null) {
...
 }
 
-const getStartupDetailsFromInitialPush = (): Promise<
-  | null
-  | {
-      startupFollowUser: string
+function* getStartupDetailsFromInitialPush() {
+  const {push, pushTimeout}: {push: PushGen.NotificationPayload; pushTimeout: boolean} = yield Saga.race({
+    push: Saga.callPromise(isAndroid ? getInitialPushAndroid : getInitialPushiOS),
+    pushTimeout: Saga.delay(10),
```
jzila: Is there a chance that this will delay startup? I think we're trying to prevent any significantly blocking activity there.

shared/actions/platform-specific/push.native.tsx
```diff
@@ -347,43 +350,45 @@ function* _checkPermissions(action: ConfigGen.MobileAppStatePayload | null) {
   }
 }
 
-const getStartupDetailsFromInitialPush = (): Promise<
-  | null
-  | {
-      startupFollowUser: string
+function* getStartupDetailsFromInitialPush() {
```
jzila: Does this need to be a generator? You could do the same thing with `setTimeout` calling `reject` on a promise...

shared/android/app/src/main/java/io/keybase/ossifrage/MainActivity.java
```diff
@@ -230,15 +231,19 @@ public void emit() {
       }
 
       // Closure like class so we can keep our emit logic together
-      class Emit implements Runnable {
+      class Emit {
```
jzila: Now that the rest of the logic is in `IntentEmitter`, do we still need this closure? [you don't have to address in this PR]

shared/actions/platform-specific/push.native.tsx
MarcoPolo: It used to delay startup. Now it can only delay startup at most 10ms. This is how you timeout a promise with redux

PR #19048, opened 2019-08-22
restore original behavior of WithUser

mmaxim:
cc @strib


maxtaco:
Oh, this is what was breaking the app yesterday!  @mmaxim thanks for finding it.  It resulted in a UI exception that was not easy to repro. We have a ticket in this week's sprint to look into it.

CC @nathunsmitty  / @mlsteele / @chrisnojima  who were affected.

maxtaco:
Also cc @heronhaye it wasn't related to GAPU as you said.

PR #19051, opened 2019-08-22
libkb: quick fix to full self cacher (as per @mlsteele)

maxtaco:


go/libkb/full_self_cacher.go
```diff
@@ -166,7 +167,10 @@ func (m *CachedFullSelf) WithUser(arg LoadUserArg, f func(u *User) error) (err e
 	var u *User
 
 	if m.me != nil && m.isSelfLoad(arg) {
-		_ = m.maybeClearCache(ctx, &arg)
+		if arg.uid.IsNil() {
+			arg.uid = m.me.GetUID()
+		}
+		m.maybeClearCache(ctx, &arg)
```
mlsteele: Use `arg.self` instead of `arg.uid.IsNil()` to determine this is a self load. So that bad args with neither continue to error out instead of getting self.

maxtaco: Well... we're already checking for self above.  So how's this instead?

mlsteele: My bad, extreme tunnel vision. Either way is good

PR #19052, opened 2019-08-22
fix ListItem2; use 'fade' instead of 'grow' everywhere in Files tab

songgao:


jakob223:
Hmm, I made `grow` specifically for this use case. If we don't want this anymore then maybe we should refactor list item 2 to not have this anymore.

PR #19061, opened 2019-08-22
Keybase client cli on Netbsd (using procfs on NetBSD)

atomicules:
These are really relatively minor changes (just adding `netbsd` build headers to files) that get me a working Keybase cli client on NetBSD. As such I see little harm in merging them in.

I'm relying on NetBSD's linux emulation and the fact that that gives me a profs (`/proc`). With this the cli seems to work perfectly - i.e. it launches the server to the background. I can read and send chats. It's great 😄 

This was inspired by [dressupgeekout's post](http://dressupgeekout.blogspot.com/2018/10/porting-keybase-to-netbsd.html) which I had initially followed. But then I wanted to build a more recent version and so ended up with these changes.

- Used [dressupgeekout's commit](https://github.com/dressupgeekout/keybase-client/commit/bccaaf3096a575589bb6ee8194008c34fe1f97c8) as a basis
- Added "bsd" (openbsd, freebsd and netbsd) headers to a couple of files after bsd building was inadvertently broken in:
    - https://github.com/keybase/client/pull/16949
    - https://github.com/keybase/client/commit/38900b9cfacfb03c83619191379e6d84d3dd61ec
- Removed the `go-ps/process_netbsd.go` file from dressupgeekout's approach and instead use the `go-ps/process_unix.go` for netbsd since I can take advantage of procfs on NetBSD.

Ultimately I would like to get a proper `go-ps/process_netbsd.go` approach working on NetBSD, but at the moment although it builds fine I can't get the tests to pass so that's why I've just gone for the procfs approach with this PR.

The only downsides I see with this PR is that some NetBSD users might be able to have this build and then not work because they don't have a procfs.


heronhaye:
Thanks for the PR. 

Since the go-ps change is in another repo, can you make a PR on https://github.com/keybase/go-ps with the relevant changes?

After that is merged I can vendor it in. 

It would probably be a good idea to add building this to CI so we don't break it in the future.  cc @jzila, shouldn't be too bad because of go's cross compilation support?

One last thing, I'll also need make a new PR and cherrypick these commits for CI to run.

atomicules:
Thanks for the reminder about the PR for go-ps. I had pushed that but just forgot to PR. It's here: https://github.com/keybase/go-ps/pull/10

jzila:
We can add x-compilation to CI. Go ahead and make a ticket.

atomicules:
See comment here: https://github.com/keybase/go-ps/pull/10#issuecomment-525020341

PR #19063, opened 2019-08-22
Add initial dark mode styles.

adamjspooner:
@keybase/react-hackers, this is ready for review. It isn't complete, but it's getting close. I'd like to get this in (behind the feature flag) for others to start using and providing feedback. It doesn't include mobile. No images have been updated, and there are bound to be screens I haven't accounted for yet.


shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
```diff
@@ -363,7 +363,7 @@ const styles = Styles.styleSheetCreate(() => ({
     common: {
       ...Styles.globalStyles.flexBoxColumn,
       alignSelf: 'stretch',
-      backgroundColor: Styles.globalColors.black,
+      backgroundColor: Styles.isDarkMode ? Styles.globalColors.white : Styles.globalColors.black,
```
chrisnojima: This is the kind of thing we want to avoid

adamjspooner: I'm not sure how to deal with these special cases. Maybe create a new color in `globalColors` instead?

PR #19077, opened 2019-08-23
Remove "user not found" logic from phone number search

nathunsmitty:


ddworken:
There are some unused styles you should probably also remove. LGTM otherwise.

PR #19078, opened 2019-08-23
Delay SearchFilter focusOnMount

buoyad:
Fixes a bug where the keyboard gets stuck open even after this input isn't visible anymore. cc @keybase/picnicsquad


buoyad:
@mlsteele `setState` will trigger a rerender, whereas we can stash `_mounted` like this without causing updates.

mlsteele:
the purpose of which is performance?

buoyad:
This component's rendering isn't a function of `_mounted`, so putting it in state is incorrect aiui

mlsteele:
🆗

PR #19079, opened 2019-08-23
🐇

mlsteele:
Some rabbits. Not all rabbits. After this merges please call out missing rabbits.

### DIFF
--- a/shared/chat/conversation/bottom-banner/index.tsx
+++ b/shared/chat/conversation/bottom-banner/index.tsx
@@ -3,6 +3,7 @@ import {Box2, Button, Text} from '../../../common-adapters'
 import {assertionToDisplay} from '../../../common-adapters/usernames'
 import * as Styles from '../../../styles'
 import {isMobile} from '../../../constants/platform'
+import Flags from '../../../util/feature-flags'
 
 export type InviteProps = {
   openShareSheet: () => void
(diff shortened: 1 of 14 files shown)


shared/team-building/service-tab-bar.desktop.tsx
```diff
@@ -237,6 +244,9 @@ const styles = Styles.styleSheetCreate(() => ({
   tabBarContainer: {
     minHeight: 30,
   },
+  wonderland: {
+    color: Styles.globalColors.white,
```
buoyad: ?

mlsteele: The rabbit appears transparent with the default transparent text color. This makes it full opacity.

shared/chat/conversation/bottom-banner/index.tsx
```diff
@@ -39,7 +40,11 @@ const InviteBanner = ({users, openSMS, openShareSheet, usernameToContactName}: I
     return (
       <BannerBox color={Styles.globalColors.blue} gap="xtiny">
         <BannerText>Last step: summon {theirName}!</BannerText>
-        <Button label="Send install link" onClick={() => openSMS(users[0].slice(0, -6))} mode="Secondary" />
+        <Button
+          label={Flags.wonderland ? '🐇 Send install link' : 'Send install link'}
```
xgess: there are three of these buttons in this file. you've only got the one that's mobile and one-on-one. are we doing this for desktop also? assuming no, then probably just need one more. down a few lines.

mlsteele: good catch. Afaik the 🐇 task includes 🐇 on desktop. Will get in future PR

PR #19080, opened 2019-08-23
New convo tab bar tweaks

nathunsmitty:
- Color labels the same color as the icons and accents
- Add some top margin to labels on mobile
- Make email icon color same as Keybase and Phone number

### DIFF
--- a/shared/team-building/service-tab-bar.native.tsx
+++ b/shared/team-building/service-tab-bar.native.tsx
@@ -49,10 +49,7 @@ const ServiceIcon = (props: IconProps) => {
           direction="vertical"
           style={Styles.collapseStyles([
             styles.labelContainer,
-            {
-              marginTop: Styles.globalMargins.xtiny,
-              overflow: 'hidden',
-            },
+            {height: labelHeight * props.labelPresence, marginTop: Styles.globalMargins.xtiny},
           ])}
         >
           <Kb.Box2 direction="vertical" style={{height: labelHeight, width: 74}}>
--- a/shared/team-building/service-tab-bar.native.tsx
+++ b/shared/team-building/service-tab-bar.native.tsx
@@ -47,12 +47,13 @@ const ServiceIcon = (props: IconProps) => {
         <Kb.Icon fontSize={18} type={serviceIdToIconFont(props.service)} color={color} />
         <Kb.Box2
           direction="vertical"
-          style={{
-            height: labelHeight * props.labelPresence,
-            marginTop: Styles.globalMargins.xtiny,
-            opacity: props.labelPresence,
-            overflow: 'hidden',
-          }}
+          style={Styles.collapseStyles([
+            styles.labelContainer,
+            {
+              marginTop: Styles.globalMargins.xtiny,
+              overflow: 'hidden',
+            },
+          ])}
         >
           <Kb.Box2 direction="vertical" style={{height: labelHeight, width: 74}}>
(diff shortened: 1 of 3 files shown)


shared/team-building/service-tab-bar.native.tsx
```diff
@@ -43,21 +44,16 @@ const ServiceIcon = (props: IconProps) => {
...
-            overflow: 'hidden',
-          }}
+          style={Styles.collapseStyles([
+            styles.labelContainer,
+            {height: labelHeight * props.labelPresence, marginTop: Styles.globalMargins.xtiny},
+          ])}
         >
           <Kb.Box2 direction="vertical" style={{height: labelHeight, width: 74}}>
-            <Kb.Text type="BodyTiny" center={true} lineClamp={2}>
+            <Kb.Text type="BodyTiny" center={true} lineClamp={2} style={{color}}>
```
nathunsmitty: This causes a type error because color could be any color. Not sure how best to handle this since really we do want to allow any color here: the brand colors of Facebook, Reddit, etc. aren't on our list of valid text colors for mobile.

Full error text:
```
team-building/service-tab-bar.native.tsx:56:66 - error TS2322: Type '{ color: string; }' is not assignable to type '(_StylesCrossPlatform & { color?: "rgba(0, 0, 0, 0.85)" | "rgba(0, 0, 0, 0.20)" | "rgb(204,204,204)" | "rgba(0, 0, 0, 0.35)" | "rgba(0, 0, 0, 0.50)" | "rgb(127,127,127)" | "rgb(38,38,38)" | "#3663EA" | "#A8CCFF" | "rgba(71, 31, 17, 0.75)" | "#FFFFFF" | "#189e7a" | "#ff6f21" | "#6d3fd1" | "#eb253b" | "rgba(0, 0, 0, 0)" | "rgba(255, 255, 255, 0.40)" | "rgba(255, 255, 255, 0.75)" | undefined; }) | (readonly (false | void | _StylesCrossPlatform | null)[] & { color?: "rgba(0, 0, 0, 0.85)" | "rgba(0, 0, 0, 0.20)" | "rgb(204,204,204)" | "rgba(0, 0, 0, 0.35)" | "rgba(0, 0, 0, 0.50)" | "rgb(127,127,127)" | "rgb(38,38,38)" | "#3663EA" | "#A8CCFF" | "rgba(71, 31, 17, 0.75)" | "#FFFFFF" | "#189e7a" | "#ff6f21" | "#6d3fd1" | "#eb253b" | "rgba(0, 0, 0, 0)" | "rgba(255, 255, 255, 0.40)" | "rgba(255, 255, 255, 0.75)" | undefined; }) | undefined'.
  Type '{ color: string; }' is not assignable to type 'undefined'.

56             <Kb.Text type="BodyTiny" center={true} lineClamp={2} style={{color}}>
                                                                    ~~~~~

  common-adapters/text.d.ts:62:3
    62   style?: StylesTextCrossPlatform
         ~~~~~
    The expected type comes from property 'style' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<Text> & Readonly<Props> & Readonly<{ children?: ReactNode; }>'
```

mlsteele: Maybe it would work to declare the type of colors [here](https://github.com/keybase/client/blob/3dfaead/shared/team-building/shared.tsx#L6) to be a union of specific choices rather than `string`.

PR #19082, opened 2019-08-23
add dismiss button to invite banner

xgess:
on mobile, add a dismiss button to the new invite install link banner. here is a screenshot when the new conversation is just one new person with a phone number:

Screen Shot 2019-08-23 at 2 36 10 PM

and here it is with multiple people:
Screen Shot 2019-08-23 at 2 38 07 PM

no dismiss button on desktop because it would look weird without a bigger redesign.

### DIFF
--- a/shared/chat/conversation/bottom-banner/container.tsx
+++ b/shared/chat/conversation/bottom-banner/container.tsx
@@ -50,7 +50,9 @@ const mapStateToProps = (state: Container.TypedState, {conversationIDKey}: OwnPr
   const _following = state.config.following
   const _meta = Constants.getMeta(state, conversationIDKey)
   const _users = state.users
+  const _dismissed = state.chat2.dismissedInviteBannersMap.get(conversationIDKey, false)
   return {
+    _dismissed,
     _following,
     _meta,
     _users,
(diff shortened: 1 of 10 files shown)


xgess:
@keybase/picnicsquad  @keybase/react-hackers

shared/chat/conversation/bottom-banner/container.tsx
```diff
@@ -16,19 +17,22 @@ type OwnProps = {
...
+  onDismiss: () => void
   usernameToContactName: {[username: string]: string}
 }
 
 class BannerContainer extends React.PureComponent<Props> {
   render() {
     switch (this.props.type) {
       case 'invite':
-        return (
+        return this.props.dismissed ? null : (
```
nathunsmitty: You could write to utilize short circuiting:

```suggestion
        return !this.props.dismissed && (
```

Different people have different opinions on the readability of this, but just wanted to point it out in case you didn't know. I think it's fine either way.

xgess: i did it this way to avoid the double negative, but meh. sure. your way seems more idiomatic.

PR #19083, opened 2019-08-23
Trim emails on signup and settings

buoyad:
cc @keybase/y2ksquad


shared/settings/account/add-modals.tsx
```diff
@@ -17,7 +17,7 @@ export const Email = () => {
   const [email, onChangeEmail] = React.useState('')
   const [searchable, onChangeSearchable] = React.useState(true)
   const [addEmailInProgress, onAddEmailInProgress] = React.useState('')
-  const disabled = !email
+  const disabled = !email.trim()
```
chrisnojima: You should just trim this above and just refer to the trimmed  one

buoyad: Fair enough

PR #19085, opened 2019-08-23
Add highlighting to the phone input component

ddworken:
Makes it so the phone input component is highlighted when in focus. Being in focus is determined by whether the input box for typing in a number is in focus. Because of how the phone input component is implemented with `fakeInput` I implemented this directly instead of relying on the code in NewInput. 

Note that this also means the phone input component always has highlighting which will apply to the settings page and the signup screen. @cecileboucheron can you confirm that is intended? Or do you only want this change for the new conversation mode? 

Note that this completes PICNIC-386 since the two other changes were already fixed in master.

### DIFF
--- a/shared/signup/phone-number/phone-input.tsx
+++ b/shared/signup/phone-number/phone-input.tsx
@@ -226,11 +226,13 @@ type State = {
   country: string
   prefix: string
   formatted: string
+  focused: boolean
 }
 
 class _PhoneInput extends React.Component<Kb.PropsWithOverlay<Props>, State> {
   state = {
     country: this.props.defaultCountry || defaultCountry,
+    focused: false,
     formatted: '',
     prefix: getCallingCode(this.props.defaultCountry || defaultCountry).slice(1),
   }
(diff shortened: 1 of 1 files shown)


ddworken:
Confirmed with Cecile in Slack.

cecileboucheron:
@ddworken You mean in Keybase :)

shared/signup/phone-number/phone-input.tsx
```diff
@@ -376,7 +378,10 @@ class _PhoneInput extends React.Component<Kb.PropsWithOverlay<Props>, State> {
 
   render() {
     return (
-      <Kb.Box2 direction={isMobile ? 'vertical' : 'horizontal'} style={styles.container}>
+      <Kb.Box2
+        direction={isMobile ? 'vertical' : 'horizontal'}
+        style={Styles.collapseStyles([styles.container, !isMobile && this.state.focused && styles.highlight])}
```
nathunsmitty: may be worth extracting `!isMobile && this.state.focused` into a separate boolean; I could see it being used in more places if this component grows. Also it makes the logic a little cleaner imo.

ddworken: Discussed in chat and leaving as is.

PR #19088, opened 2019-08-23
Handle both chat notifications the same way [Android]

MarcoPolo:
@jzila @joshblum @keybase/react-hackers 

We have `chat.newmessage` and `chat.newmessageSilent_2`. The silent version carries more data (can contain the encrypted message payload).

Before we were handling this in two different ways so `chat.newmessage` would keep the old style android notifications. This changes it so that it's consistent and keeps the new style on both flows.

This also:
 * preserves the server rendered notification display message for when we don't have a plaintext to display.
* Handle the case where we don't have a boxed message so that we always call the notifier. (this way native code doesn't have to catch an exception and try to figure out what notification to show).
* Parses the username from the given server message (when we don't have a msg to unbox) to keep the notification UI consistent. This is pretty hacky right now, but will change when there is server support for it.
* Preloads the message from the notification by unboxing it in both cases

### DIFF
--- a/go/bind/notifications.go
+++ b/go/bind/notifications.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/keybase/client/go/chat"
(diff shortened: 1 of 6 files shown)


go/bind/notifications.go
```diff
@@ -122,33 +115,44 @@ func HandleBackgroundNotification(strConvID, body string, intMembersType int, di
...
 			}
-			chatNotification.Message.Plaintext = emoji.Sprintf("Reacted to your message with %v", reaction)
-		default:
-			kbCtx.Log.CDebugf(ctx, "unboxNotification: Unknown message type: %v", msgUnboxed.GetMessageType())
-			return errors.New("invalid message type for plaintext")
 		}
+	} else {
+		kbCtx.Log.CDebugf(ctx, "unboxNotification: failed to unbox: %s", err)
+		// Guess the username? We need this for android
+		chatNotification.Message.From.KeybaseUsername = strings.Split(serverMessageBody, " ")[0]
```
joshblum: sanity check split length here

shared/android/app/src/main/java/io/keybase/ossifrage/KeybasePushNotificationListenerService.java
```diff
@@ -288,3 +238,70 @@ public void add(MessagingStyle.Message m) {
...
+            unixTime = ((new Date()).getTime() / 1000);
+        } else if (type.equals("chat.newmessageSilent_2"))  {
+            messageId = Integer.parseInt(bundle.getString("d", ""));
+            convID = bundle.getString("c");
+            unixTime = Long.parseLong(bundle.getString("x", "0"));
+        } else {
+            throw new Error("Tried to parse notification of unhandled type: " + type );
+        }
+
+        String pushIdTmp = "";
```
joshblum: this should happen in the silent parsing branch, will always fail for newmessage

MarcoPolo:
Needs to handle the message is not valid case too: https://github.com/keybase/client/pull/19112#pullrequestreview-279317690

Issue #18945, opened 2019-08-14
Documentation on security implications and use of phone numbers in Keybase


nealmcb:
I was surprised to see the message **"Add your phone number so your friends can find you."** on my Keybase Android app today. But I see no way to search by phone number. And the only options are "**Add number**" (by default) and "_Later_".

I searched high and low for documentation on the intended use of this, what Keybase would do with it, how it could be used by me and my friends, what the privacy policy was, etc etc.
I found nothing in the [blog](https://keybase.io/blog), and only a smattering of mentions here in issues. That's very distressing for a privacy-focused app.

The most interesting hit was [Stop nagging about email verification and allow disabling unwanted features · Issue \#18903 · keybase/client](https://github.com/keybase/client/issues/18903) with its link to [Why Apple’s Notification Bubbles Are so Stressful \- OneZero](https://onezero.medium.com/why-apples-notification-bubbles-are-so-stressful-65a544e51f10). And I agree. Having that red highlite of "1" on the bottom left of my app is very distracting even though I'm colorblind so red is not very bright.

I also found [a bunch of PRs](https://github.com/keybase/client/issues?utf8=✓&q="phone+number") but no design document or the like.

So what's up? Why would this feature be helpful? What are the security issues? And how do I get that "1" to go away without suggesting I'm interested in seeing this request again, until I know what's in it for me?

junderw:
you can search by phone number by just typing in the number in user search

jamesdwilson:
> you can search by phone number by just typing in the number in user search

but what if i don't WANT to be discoverable by phone, email or other methods?
Like @nealmcb says, I want the simple feature to say "don't ask me again about this"

heronhaye:
@nealmcb We'll have a doc out soon. 
You can get rid of the badge by clicking "later", the app won't ask you again unless you go to the settings page yourself to add a phone number.

heronhaye:
We wrote some docs at https://keybase.io/docs/chat/phones_and_emails. Let us know if there's anything else you want to know that's missing, we'll add it. Thanks!

nealmcb:
I think the information from that handy document, and tips on privacy implications, should be part of the client documentation. There are no tips in the client on how to search for phone numbers, and it is counter-intuitive at least for me that you'd just search "people" e.g. in the Keybase filter.
The "Import phone contacts" option in Settings should link to the document or provide related info in a popup or the like.
[...]

PR #19092, opened 2019-08-23
Later -> Skip in the gui

heronhaye:


heronhaye:
Ref: https://github.com/keybase/client/issues/18945

PR #19096, opened 2019-08-23
keybase_service_base: assure the current device can migrate

strib:
Before taking the migration lock on the server, assure ourselves that the current device has read access to the TLF and thus can perform the migration.  Right now it's possible for a new device, that hasn't been keyed for the TLF yet, to get the migration request, take the lock, and then fail to perform the migration.  This locks out any future rekeys that would make the TLF readable, since there's nothing that automatically releases the lock when the migration rekey fails to happen.

Issue: HOTPOT-552

### DIFF
--- a/go/kbfs/libkbfs/folder_branch_ops.go
+++ b/go/kbfs/libkbfs/folder_branch_ops.go
@@ -8393,6 +8393,17 @@ func (fbo *folderBranchOps) getMDForMigrationLocked(
 	return md, nil
 }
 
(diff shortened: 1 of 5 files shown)


strib:
> LGTM, although it sounds like a good idea to have some auto-unlock mechanism too...

Yeah, right now I'm manually unlocking things once in a while.  It would be nice to have, but it wouldn't help much in these cases because it's a race, and the same device could just take the lock again right away.

PR #19108, opened 2019-08-24
Fix contact search results and service map rendering

zapu:
This should fix a non-deterministic search result bug that @malgorithms has ran into, where if you have results with the same score coming from contact search, scrolling down the list to fetch new results would yield a list with these contact new results not being at the end.

Also fix service icons in user result rows to always show up in specific ordering.


zapu:
cc @keybase/react-hackers

PR #19110, opened 2019-08-24
New chat status: use Banner, new small prop 

cecileboucheron:
@keybase/react-hackers


mmaxim:
Does this maintain the same height?

cecileboucheron:
It's slightly bigger, I find the current design too thin. I made a new "small" version of Banner that uses BodyTiny and is half the usual height.

PR #19112, opened 2019-08-25
just bail out of here if the message is not valid

mmaxim:
Fixes a crasher handling pushes.


MarcoPolo:
Made those changes here as well https://github.com/keybase/client/pull/19088/commits/2e0db18318638a587f2ba8bb9d6ab8868fc5eccf

PR #19117, opened 2019-08-26
Fix white rabbit on Linux, change "Keybase & Contacts" on desktop

zapu:
cc @keybase/react-hackers @keybase/picnicsquad

Made a separate `ContinueButton` component so we can use unicode emoji characters on mobile but Kb.Emoji objects on desktop. This is to keep things consistent across platforms - some Linux distributions do not include emoji fonts, and if they do, the quality varies. Using Kb.Emoji ensures that same rabbit emoji will be displayed everywhere. But Kb.Emoji inside of Kb.Button does not seem to render on mobile.


shared/team-building/email-input/index.tsx
```diff
@@ -87,12 +87,16 @@ const EmailInput = ({namespace}: EmailInputProps) => {
...
-            onClick={onSubmit}
-            disabled={!canSubmit}
-          />
+          <Kb.Button fullWidth={true} onClick={onSubmit} disabled={!canSubmit}>
+            <Kb.Text type="BodyBig" style={styles.continueButton}>
+              Continue
+            </Kb.Text>
+            {Flags.wonderland && (
+              <Kb.Text type="BodyBig" style={styles.rabbitEmoji}>
+                <Kb.Emoji size={16} emojiName="rabbit2" />
```
zapu: These don't seem to display on mobile...

zapu:
cc @keybase/react-hackers ready for review, it's not needed for mobile release but it would be great if we could get it in for PC. Look into squad-picnic Keybase channel to see what the issue is about

PR #19124, opened 2019-08-26
Use Kb.Emoji in New Chat button on desktop

zapu:


zapu:
cc @keybase/picnicsquad @keybase/react-hackers

PR #19131, opened 2019-08-26
Added initial merkle tree implementation and proof verification

AMarcedone:
cc @heronhaye (for when you get back)


go/merkletree2/position.go
```diff
@@ -96,18 +96,72 @@ func (t *Tree) getDeepestPositionForKey(k Key) (*Position, error) {
...
+
+// getLevel returns the level of p. The root is at level 0, and each node has
+// level 1 higher than its parent.
+func (t *TreeConfig) getLevel(p *Position) int {
+	return ((*big.Int)(p).BitLen() - 1) / int(t.bitsPerIndex)
+}
+
+// getParentAtLevel returns nil if p is at a level lower than `level`. The root
+// is at level 0, and each node has level 1 higher than its parent.
+func (t *TreeConfig) getParentAtLevel(p *Position, level uint) *Position {
```
heronhaye: sanity tests for these new functions

go/merkletree2/position.go
```diff
@@ -96,18 +96,72 @@ func (t *Tree) getDeepestPositionForKey(k Key) (*Position, error) {
...
+	var f big.Int
+	f.Rsh((*big.Int)(p), uint(shift))
+
+	return (*Position)(&f)
+}
+
+// positionToChildIndexes returns the list of childIndexes to navigate from the
+// root to p (in reverse order).
+func (t *TreeConfig) positionToChildIndexes(p *Position) (path []ChildIndex) {
+	path = make([]ChildIndex, ((*big.Int)(p).BitLen()-1)/int(t.bitsPerIndex))
```
heronhaye: `((*big.Int)(p).BitLen()-1)/int(t.bitsPerIndex)` this seems like a common expr, maybe it could be factored out

go/merkletree2/proof_verifier.go
```diff
@@ -0,0 +1,99 @@
+package merkletree2
+
+import (
+	"fmt"
+
+	"golang.org/x/net/context"
+)
+
+type MerkleProofVerifier struct {
+	TreeConfig
```
heronhaye: don't embed the struct here

go/merkletree2/tree.go
```diff
@@ -1,38 +1,405 @@
...
 // Tree is the MerkleTree class; it needs an engine and a configuration
 // to run
 type Tree struct {
 	sync.RWMutex
-	//eng StorageEngine not yet implemented/defined
-	cfg Config
+	TreeConfig
+
+	eng StorageEngine
+	log logger.Logger
```
heronhaye: you could embed the logger

go/merkletree2/tree.go
```diff
@@ -1,38 +1,405 @@
...
-	Construct() interface{}
+// NewTree makes a new tree
+func NewTree(c TreeConfig, e StorageEngine, l logger.Logger) *Tree {
+	return &Tree{TreeConfig: c, eng: e, log: l}
 }
 
 // Hash is a byte-array, used to represent a full collision-resistant hash.
 type Hash []byte
 
+// Equal compares two hashes byte by byte
```
heronhaye: don't really need these comments

go/merkletree2/tree.go
```diff
@@ -1,38 +1,405 @@
...
+}
+
+// Cmp compares two keys lexicographically as byte slices
+func (k Key) Cmp(k2 Key) int {
+	return bytes.Compare(k, k2)
+}
+
 // Seqno is an integer used to differentiate different versions of a merkle tree.
-type Seqno uint
+type Seqno int
```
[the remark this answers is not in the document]
heronhaye: we should pick a type and use it throughout...I don't have too strong of an opinion, maybe uint64?

go/merkletree2/tree.go
```diff
@@ -1,38 +1,405 @@
...
+	h Hash
+}
+
+type RootMetadataNode struct {
+	// TODO Add skip pointers, timestamp, version.....
+	Seqno        Seqno
+	BareRootHash Hash
+}
+
+func (t *Tree) makeNextRootMNode(curr *RootMetadataNode, newRootHash Hash) RootMetadataNode {
```
heronhaye: spell metadata out

go/merkletree2/tree.go
```diff
@@ -1,38 +1,405 @@
...
+}
+
+// BuildNewTreeVersionFromAllKeys builds a new tree version, taking a batch
+// input. The sortedKeyValuePairs must be sorted (lexicographically) by Key.
+//
+// NOTE: The input to this function should contain at least all the keys which
+// were inserted in previous versions of the tree, otherwise this procedure will
+// put the tree into an inconsistent state. This function does not check the
+// condition is true for efficiency reasons.
+func (t *Tree) BuildNewTreeVersionFromAllKeys(
```
heronhaye: can just call it `Build`

go/merkletree2/tree.go
```diff
@@ -1,38 +1,405 @@
...
+	t.Lock()
+	defer t.Unlock()
+
+	var latestSeqNo Seqno
+	var rootMNode RootMetadataNode
+	if latestSeqNo, rootMNode, err = t.eng.LookupLatestRoot(ctx, tr); err != nil {
+		return err
+	}
+	if latestSeqNo == 0 {
+		t.log.CWarningf(ctx, "No root found. Starting a new merkle tree.")
```
heronhaye: maybe more of an Info than a Warning

go/merkletree2/tree.go
```diff
@@ -1,38 +1,405 @@
...
+	}
+
+	var ms MasterSecret
+	if t.useBlindedValueHashes {
```
heronhaye: this kind of stuff indicates the package is too special cased

go/merkletree2/tree.go
```diff
@@ -1,38 +1,405 @@
...
+		return nil, ctx.Err()
+	default:
+	}
+
+	if len(sortedKVPairs) <= t.maxValuesPerLeaf {
+		ret, err = t.makeAndStoreLeaf(ctx, tr, s, ms, p, sortedKVPairs)
+		return ret, err
+	}
+
+	var node Node
```
heronhaye: do node := Node{Type: ... INodes: ...}
or better yet, make two constructors like NewEmptyInternalNode and NewLeafNode

go/merkletree2/proof_verifier.go
```diff
@@ -0,0 +1,99 @@
...
+	if len(proof.OtherPairsInLeaf)+1 > m.maxValuesPerLeaf {
+		return NewProofVerificationFailedError(fmt.Sprintf("Too many keys in leaf: %v", len(proof.OtherPairsInLeaf)+1))
+	}
+
+	// Reconstruct the leaf node
+	var leaf Node
+	leaf.Type = NodeTypeLeaf
+	leaf.LeafHashes = make([]KeyHashPair, len(proof.OtherPairsInLeaf)+1)
+	// LeafHashes is obtained by adding kvp into OtherPairsInLeaf while maintaining sorted order
+	for i, j, k := 0, 0, 0; i < len(proof.OtherPairsInLeaf)+1; i++ {
```
heronhaye: wouldn't this be easier if the proof included what index the KV pair should go into? or even included the entire list including the kvpair instead of "Other"s?

heronhaye: it is kind of confusing now with all these indices (though `k` is being used as a boolean it seems)

go/merkletree2/proof_verifier.go
```diff
@@ -0,0 +1,99 @@
...
+	if len(sibH)%(m.childrenPerNode-1) != 0 {
+		return NewProofVerificationFailedError(fmt.Sprintf("Invalid number of SiblingHashes %v", len(sibH)))
+	}
+	keyAsPos, err := m.getDeepestPositionForKey(kvp.Key)
+	if err != nil {
+		return NewProofVerificationFailedErrorFromCause(err)
+	}
+	leafPosition := m.getParentAtLevel(keyAsPos, uint(len(sibH)/(m.childrenPerNode-1)))
+
+	for _, childIndex := range m.positionToChildIndexes(leafPosition) {
```
heronhaye: maybe `positionToChildIndexes` should be called `positionToChildIndexPath` or something?

heronhaye: add a comment above this for loop about what its doing

go/merkletree2/proof_verifier.go
```diff
@@ -0,0 +1,99 @@
...
+		var node Node
+		node.Type = NodeTypeINode
+		node.INodes = make([]Hash, m.childrenPerNode)
+		copy(node.INodes, sibH[:int(childIndex)])
+		node.INodes[int(childIndex)] = nodeHash
+		copy(node.INodes[int(childIndex)+1:], sibH[int(childIndex):m.childrenPerNode-1])
+
+		sibH = sibH[m.childrenPerNode-1:]
```
heronhaye: prefer not to mutate sibH and add in the start indices into the sibH slices

go/merkletree2/proof_verifier.go
```diff
@@ -0,0 +1,99 @@
...
+
+type MerkleProofVerifier struct {
+	TreeConfig
+}
+
+func NewMerkleProofVerifier(t TreeConfig) MerkleProofVerifier {
+	return MerkleProofVerifier{TreeConfig: t}
+}
+
+func (m *MerkleProofVerifier) VerifyInclusionProof(ctx context.Context, kvp KeyValuePair, proof MerkleInclusionProof, expRootHash Hash) error {
```
heronhaye: In general I think this MerkleInclusionProof struct makes this function a lot harder than it could be. For example you could have sibling hashes be a list of left, right pairs of lists, so you don't need to do any index arithmetic.
And OtherPairsInLeaves could just be the entire set of KV pairs in the leaf so you don't need to manually insert it.

I think the space increase of the proof would be minimal and this function might even get a bit faster without all the copying.

go/merkletree2/proof_verifier.go
AMarcedone: It could, but we need to check the order anyways to prevent the server from putting two different values for the same key at different indexes in the tree. So it wouldn't save much...

heronhaye: but it'd just be a simple linear pass

go/merkletree2/proof_verifier.go
AMarcedone: I agree the algorithm is not the most readable and there is some bit fiddling, but I'd rather have a smaller proof than a simpler to read algorithm. Do not feel strongly though. @maxtaco ?

go/merkletree2/config.go
```diff
@@ -19,28 +30,34 @@ type Config struct {
...
+
 // Hasher is an interface for hashing MerkleTree data structures into their
-// cryptographic hashes.
+// cryptographic hashes. It also manages blinding secrets.
 type Hasher interface {
-	Hash([]byte) Hash
+	EncodeAndHashGeneric(interface{}) (Hash, error)
+	HashKeyValuePairWithMasterSecret(KeyValuePair, Seqno, MasterSecret) (Hash, error)
+	HashKeyValuePairWithKeySpecificSecret(KeyValuePair, KeySpecificSecret) (Hash, error)
+	GenerateMasterSecret(Seqno) (MasterSecret, error)
```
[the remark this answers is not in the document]
AMarcedone: I was trying to keep it more general, in case we want to derive a MasterSecret in a pseudorandom way (for example by hashing a fixed secret with the seqno). Can take it out though.

go/merkletree2/tree.go
AMarcedone: my linter complains if I don't comment. Not sure if the one in CI does though

go/merkletree2/config.go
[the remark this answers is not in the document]
AMarcedone: Yes.

go/merkletree2/config.go
```diff
@@ -19,28 +30,34 @@ type Config struct {
...
+	ComputeKeySpecificSecret(MasterSecret, Seqno, Key) KeySpecificSecret
```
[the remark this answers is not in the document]
AMarcedone: It doesn't. Was trying to keep it a bit more general. I'll take it out.

go/merkletree2/interfaces.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// LookupKVPair returns the KeyValuePair (and its Hash) with the highest Seqno s1 <=
+	// s which was stored at position p (similarly to LookupNode).
+	LookupKVPair(c context.Context, t Transaction, s Seqno, k Key) (kvp KeyValuePair, s1 Seqno, err error)
+
+	// LookupKeyHashPairsUnderPosition returns all KeyValuePairs which were
+	// stored at a position p' which is a descendent of p and at the maximum
+	// Seqno s' <= s (similarly to LookupNode). For each such pair, it returns
+	// the Seqno at which it was stored (in the same order).
+	LookupKeyValuePairsUnderPosition(ctx context.Context, tr Transaction, s Seqno, p Position) ([]KeyValuePair, []Seqno, error)
```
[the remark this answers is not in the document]
AMarcedone: I was doing it this way to avoid unnecessary copying and defining extra types just for the purpose of this method. Want me to change it?

go/merkletree2/tree.go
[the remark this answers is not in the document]
AMarcedone: Let's do int64. The reason for the change is that if we do arithmetic over this stuff it is easy to get silent underflows.  `len([]byte)` returns an int rather than uint for this reason.

go/merkletree2/interfaces.go
```diff
@@ -0,0 +1,58 @@
+package merkletree2
+
+import "golang.org/x/net/context"
+
+// StorageEngine specifies how to store and lookup merkle tree nodes, roots and
+// KeyValuePairs. You can use a DB like Dynamo or SQL to do this.
+type StorageEngine interface {
+	// TODO update transaction management together with the blind architect
+	NewTransaction(context.Context) (Transaction, error)
```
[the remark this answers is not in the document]
AMarcedone: The todo was intended to convey that these stuff is not in scope for this PR. I'll talk to max about it. We can fix it with the next PR, when I actually implement the sql layer.

go/merkletree2/proof_verifier.go
```diff
@@ -0,0 +1,101 @@
...
+		return NewProofVerificationFailedError(fmt.Errorf("Too many keys in leaf: %v", len(proof.OtherPairsInLeaf)+1))
+	}
+
+	// Reconstruct the leaf node
+	leaf := NewNodeLeaf()
+	leaf.LeafHashes = make([]KeyHashPair, len(proof.OtherPairsInLeaf)+1)
+
+	// LeafHashes is obtained by adding kvp into OtherPairsInLeaf while maintaining sorted order
+	for i, j, k := 0, 0, 0; i < len(proof.OtherPairsInLeaf)+1; i++ {
+		if (j < len(proof.OtherPairsInLeaf) && k < 1 && proof.OtherPairsInLeaf[j].Key.Cmp(kvp.Key) > 0) || j >= len(proof.OtherPairsInLeaf) {
```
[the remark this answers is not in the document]
maxtaco: Why `k < 1` ?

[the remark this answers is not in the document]
maxtaco: Oh, shouldn't `k` be a `bool` called `didInsert` or something? That might make this easier to understand

go/merkletree2/config.go
```diff
@@ -20,27 +41,45 @@ type Config struct {
...
-// will store, and a valueConstructor (so that typed values can be pulled out
-// of the Merkle Tree).
-func NewConfig(h Hasher, logChildrenPerNode uint8, valuesPerLeaf int, keysByteLength int, valueConstructor ValueConstructor) (Config, error) {
+// NewConfig makes a new config object. It takes a a Hasher, logChildrenPerNode
+// which is the base 2 logarithm of the number of children per interior node,
+// maxValuesPerLeaf the maximum number of entries in a leaf before the leaf is
+// split into multiple nodes (at a lower level in the tree), keyByteLength the
+// length of the Keys which the tree will store, and a valueConstructor (so that
+// typed values can be pulled out of the Merkle Tree).
+func NewConfig(h Encoder, useBlindedValueHashes bool, logChildrenPerNode uint8, maxValuesPerLeaf int, keysByteLength int) (Config, error) {
```
maxtaco: Want to assert that `logChildrenPerNode > 0` since we divide by `numChildren-1` in a lot of places, which would crash.

go/merkletree2/proof_verifier.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	leafPosition := m.cfg.getParentAtLevel(keyAsPos, uint(len(sibH)/(m.cfg.childrenPerNode-1)))
+
+	// recompute the hash of the root node by recreating all the internal nodes
+	// on the path from the leaf to the root.
+	i := 0
+	for _, childIndex := range m.cfg.positionToChildIndexPath(leafPosition) {
+		node := NewNodeInternal()
+		node.INodes = make([]Hash, m.cfg.childrenPerNode)
+		copy(node.INodes, sibH[:int(childIndex)])
```
maxtaco: This seems like a bug, seems like it's copying way too much of `sibH` into place here, no? It seems like `i` should be part of the slice parameters above.

maxtaco: If this is a bug, then test coverage should increase to cover this. I'm guessing you're not testing trees that are deep enough?

maxtaco: Thinking about how you can clean this up, I think I would do this.  I'd have one slice which is all the sibling hashes. And a subslice which is just *this level's sibling pointers*.

go/merkletree2/tree_test.go
```diff
@@ -0,0 +1,417 @@
...
+
+			_, _, rootHash2, err := tree.GetLatestRoot(context.TODO(), nil)
+			require.NoError(t, err)
+			require.True(t, rootHash2Exp.Equal(rootHash2))
+
+			require.False(t, rootHash1.Equal(rootHash2))
+		})
+	}
+
+}
```
maxtaco: As noted in person, need more tests for bigger deeper trees. Can use random keys.

maxtaco: For instance, make a test of a binary and 4-way tree that each have 10000 values in them.

go/merkletree2/interfaces.go
```diff
@@ -0,0 +1,63 @@
...
+	// StoreKVPairs stores the []KeyValuePair in the tree.
+	StoreKVPairs(context.Context, Transaction, Seqno, []KeyValuePair) error
+
+	StoreNode(context.Context, Transaction, Seqno, Position, Hash) error
+
+	StoreRootMetadata(context.Context, Transaction, RootMetadata) error
+
+	// LookupLatestRoot returns the latest root metadata and sequence number in
+	// the tree. If no root is found, then a NoLatestRootFound error is returned.
+	LookupLatestRoot(context.Context, Transaction) (Seqno, RootMetadata, error)
```
maxtaco: I don't think `Transactions` should be in the `Lookup` signatures.  Also, I recommend that this interface has a function: `NewTreeBuilder` and an `type TreeBuilder interface` that should live only as long as one creation of a tree.  It can get the in-memory cache of the prior seqno's KV pairs from the `StorageEngine`.

go/merkletree2/proof_verifier.go
AMarcedone: Yeah, good catch! (originally I was chopping the pieces off the slice, but I am using an index because Surya suggested it would make it more readable. I only updated one of the two lines and my existing tests were passing). I'll add another test.

go/merkletree2/position.go
```diff
@@ -86,8 +93,8 @@ func (t *Tree) getAllSiblings(p *Position) (siblings []Position, parent *Positio
 
 // getDeepestPositionForKey converts the key into the position the key would be
 // stored at if the tree was full with only one key per leaf.
-func (t *Tree) getDeepestPositionForKey(k Key) (*Position, error) {
-	if len(k) != t.cfg.keysByteLength {
+func (t *Config) getDeepestPositionForKey(k Key) (*Position, error) {
```
maxtaco: As commented in chat, if you do this: `k = k[0:8]`, you'll get a 2x speedup in generating inclusion proofs.  That relieves almost all the memory pressure caused by bigints, since a bigint with 8 bytes is just a uint64. I think it's worth implementing this optimization for real, it's the lowest hanging fruit to speeding up running this tree.

PR #19136, opened 2019-08-26
[WIP Y2K-510] Push Prompt as modal

jzila:
Tested on Android, needs testing on iOS.

cc @keybase/y2ksquad


jzila:
Ready for another look.

PR #19160, opened 2019-08-27
netbsd support in the client

heronhaye:
@atomicules

### DIFF
--- a/go/vendor/github.com/keybase/go-ps/process_openbsd.go
+++ b/go/vendor/github.com/keybase/go-ps/process_openbsd.go
@@ -1,4 +1,4 @@
-// +build openbsd,amd64
+// +build openbsd
 
 package ps
 
--- a/go/vendor/vendor.json
+++ b/go/vendor/vendor.json
@@ -668,10 +668,10 @@
 			"revisionTime": "2018-10-16T18:57:45Z"
 		},
 		{
-			"checksumSHA1": "07FctQiwHCBvhiXp2mTLc8e0ll8=",
+			"checksumSHA1": "iC6oxp3NRhZJW5+OmHWH56j1dv0=",
 			"path": "github.com/keybase/go-ps",
-			"revision": "9c11ca40a47944f49e0784b77a66636cb098445a",
-			"revisionTime": "2016-08-09T20:56:21Z"
+			"revision": "91aafc93ba19d1988cff338c1929d35b6c6f5b50",
+			"revisionTime": "2019-08-27T17:51:25Z"
 		},
 		{
 			"checksumSHA1": "KNf4CdZexyFjpEugzBv3b7YJs4U=",


atomicules:
Anything you need from me on this?

atomicules:
Tested master and it builds and works fine (so for me this was upgrading from 4.4.0 to 4.5.0):

```
~/Code/Go/...go/keybase * master$ go111 install -tags production
~/Code/Go/...go/keybase * master$ keybase version
Client:  4.5.0
Service: 4.4.0
~/Code/Go/...go/keybase * master$ keybase chat ls
▶ WARNING Restarting after upgrade; service is running v4.4.0, while v4.5.0 is available
▶ INFO Restarting service...
▶ INFO Starting background server with pid=25672
[1]   [REDACTED]
[2]   [REDACTED]
[3]   [REDACTED]
[4]   [REDACTED]
```

atomicules:
> Anything else we can do?

I can't think of anything. I'm not personally concerned with a GUI on BSDs. As long as I can keep the command line client working I'm happy, thank you.

atomicules:
Fwiw I also cross compiled that latest master build to NetBSD arm and tried that on an NetBSD arm box and it worked there to; I didn't build on the arm box due in part to laziness - would need to also build golang there.

I did see one odd thing where it died, but it ran ok when I tried again:

```
ip-172-31-20-142$ ./keybase version
Client:  4.5.0
Service: 4.4.0
ip-172-31-20-142$ ./keybase chat ls
▶ WARNING Restarting after upgrade; service is running v4.4.0, while v4.5.0 is available
▶ WARNING Error in Stopping 2642 via RPC: EOF; trying fallback (kill via pidfile)
▶ WARNING Kill via pidfile failed: os: process already finished

▶ ERROR os: process already finished
ip-172-31-20-142$ ./keybase chat ls
▶ INFO Starting background server with pid=13602
[1]  [REDACTED]
[2]  [REDACTED]
[3]  [REDACTED]
[4]  [REDACTED]
```

You can see how lazy I was being as I couldn't even be bothered to put the executable in my `PATH`.

heronhaye:
Hmm, weird. May be due to the upgrade, if it happens again let us know. Glad you got it working!

heronhaye:
Oh, what about KBFS? Does fuse work on bsd? (although it is useful even without fuse via the git remote helper and `keybase fs` commands)

atomicules:
Hmm... I have not looked at KBFS at all yet. I will take a look and report back. At the moment:

```
$ keybase fs ps
▶ ERROR Keybase services aren't running - KBFS client not found.
```

heronhaye:
It is a different daemon, you have to build `go/kbfs/kbfsfuse` and run that in the background. I tried crosscompiling but it seems there are some differences in syscall methods that'd have to be special cased, but nothing unsurmountable.

There are some docs here: https://keybase.io/docs/kbfs, https://keybase.io/docs/kbfs/understanding_kbfs.

atomicules:
@heronhaye Just to follow-up on the kbfs stuff. Someone [started on a branch for openbsd](https://github.com/bazil/fuse/pull/104) for bazil/fuse. OpenBSD and NetBSD are often pretty close so I will dig through that and see if it's possible to finish off and get working, but won't be a quick thing.

heronhaye:
@atomicules Ah, thanks for looking into it. Let me know if you need any help getting that working on Keybase's side.

Issue #19133, opened 2019-08-26
Private folder requires reset


aaranmcguire:
Hi,

Please would you be able to reset my private folder. Account: `aaranmcguire`.

```
keybase sign -m "Mon 26 Aug 2019 15:11:23 EDT: Please reset folder /keybase/private/aaranmcguire because all the devices that have access have been revoked."

BEGIN KEYBASE SALTPACK SIGNED MESSAGE. kXR7VktZdyH7rvq v5weRa0zk9T2KUy exsTIhKY583EKBc w2Q0TqqKC1AHsIu c3Zz2OVRCPVU4ab 7KafxvDJkifbKtM kreAsuVdjS8dION i6nAYiiV4TdMpsj ZDGqN3FLoCaP5li hQKtaCoVe51guLn btryhbzqR9LiYNr 2RAjseqCamki6yx O9bvwEtqAQVRC8I AHMLUgVMPr27TIF NteqgBu2a7LLY3I VjdvgERChZ3Hoow KpK70A1DTaynzTI FI1NKP0Ss2D3ic4 AgfQZ8OHMfRVTqT JM4d19uCYr0Uvy4 pNyluCodiGQTtnz fMh0qS014jVuRxR JTTFkdqvWronJ71 612n4UDnXIpihsY RbnVaJ9IyAsUY. END KEYBASE SALTPACK SIGNED MESSAGE.
```

strib:
@aaranmcguire sorry for the delay, but here's a build with a command that will hopefully help you:

https://s3.amazonaws.com/prerelease.keybase.io/darwin-test/Keybase-4.3.2-20190829234136%2Bd8773bc86e.dmg

It is based on the current macOS release, but has one new command you can run from a terminal (cherry-picked from #19164).  Once you install it (open the dmg and drag it into your Applications folder and restart Keybase), you should be able to run this command:

```sh
keybase fs reset --id 7a1e8147fcecae97dff6fc913b31e716 /keybase/private/aaranmcguire
```

Please let me know if that helps!  (Might require a Keybase app restart on your devices.)

aaranmcguire:
Great, that worked. Thank you very much!

PR #19164, opened 2019-08-27
kbfs_ops: allow a user to reset to a particular TLF ID

strib:
Thee have been cases where a user reset their folder long ago, but it got written into their team sigchain still, even while they continued to use the post-reset TLF ID as their folder.  Then later, they upgraded their software or something and they can no longer access the new folder, because they get the old ID from the sigchain.

To save the new data, we need to force the client to sign that TLF ID into the team sigchain, like we do in resets.  This lets us accept a TLF ID from the command line, so experts can guide someone in this situation through to overwriting the old TLF ID with the new one.

Issue: HOTPOT-685
Issue: #19133

### DIFF
--- a/go/client/cmd_simplefs_reset.go
+++ b/go/client/cmd_simplefs_reset.go
@@ -16,7 +16,8 @@ import (
 // CmdSimpleFSReset is the 'fs reset' command.
 type CmdSimpleFSReset struct {
 	libkb.Contextified
-	path keybase1.Path
+	path  keybase1.Path
+	tlfID string
 }
 
 // NewCmdSimpleFSReset creates a new cli.Command.
@@ -30,6 +31,12 @@ func NewCmdSimpleFSReset(
 			cl.ChooseCommand(&CmdSimpleFSReset{
 				Contextified: libkb.NewContextified(g)}, "reset", c)
 			cl.SetNoStandalone()
+		},
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  "id",
+				Usage: "the ID to which to reset the folder (experts only!)",
+			},
 		},
 	}
 }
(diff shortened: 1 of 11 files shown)


go/client/cmd_simplefs_reset.go
```diff
@@ -31,6 +32,12 @@ func NewCmdSimpleFSReset(
 				Contextified: libkb.NewContextified(g)}, "reset", c)
 			cl.SetNoStandalone()
 		},
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  "id",
+				Usage: "the ID to which to reset the folder (experts only!)",
```
jakob223: Maybe something even more explicit like "this can break this folder if you don't know exactly what you're doing"? or do you think this is discouraging enough?

go/kbfs/libkbfs/kbfs_ops_test.go
```diff
@@ -4283,6 +4283,21 @@ func TestKBFSOpsReset(t *testing.T) {
 	children, err = kbfsOps.GetDirChildren(ctx, rootNode)
 	require.NoError(t, err)
 	require.Len(t, children, 1)
+
+	t.Logf("Reset it back")
+	md.override = true
+	err = kbfsOps.Reset(ctx, h, &oldID)
+	require.NoError(t, err)
+	require.Equal(t, oldID, oldID)
```
jakob223: you're requiring that two of the same thing are equal. I think you meant to check that `Equal(t, oldID, h.TlfID())`?

go/client/cmd_simplefs_reset.go
strib: I think we try to keep the help length to under 80 chars, so I didn't want to do too much.  But also, most people aren't going to be able to figure out what to put in there -- it will have to parse as a real TLF ID, and also the TLF it corresponds to has to match the given handle, so it seems pretty safe to me as is.  The command itself is basically self-discouraging.

go/kbfs/libkbfs/kbfs_ops_test.go
strib: Great catch, thanks!

strib:
Merging over known systests flake.

PR #19165, opened 2019-08-27
Support for immutable Device -> Icon mapping

buoyad:
The idea is that your first computer will get `background-1`, the second `background-2`, etc. and your 11th will get `background-1` again. Same for mobile.

image
image

These show the icons your **new** device will get:

image

Signup enter-device-name always shows `background-1`. cc @keybase/y2ksquad


shared/constants/devices.tsx
```diff
@@ -59,3 +60,81 @@ export const getDeviceCounts = (state: Container.TypedState) =>
...
+        {backup: [], desktop: [], mobile: []}
+      )
+    return {
+      map: new Map([
+        ...sorted.backup.map(idxMapper),
+        ...sorted.desktop.map(idxMapper),
+        ...sorted.mobile.map(idxMapper),
+      ]),
+      next: {
+        desktop: sorted.desktop.length,
```
jbialobr: I think that `(idx % numBackgrounds) + 1` arithmetic is missing here and below.

buoyad:
@chrisnojima bump, any more comments about this?

PR #19176, opened 2019-08-28
dont load git on git changes from gregor

chrisnojima:
doesn't seem to be a real reason why this is needed
@keybase/react-hackers


chrisnojima:
when you go to the screen it loads again. you can see the spinners. if i go to git / chat, then make a repo on my phone, then click back to git it successfully shows

PR #19183, opened 2019-08-28
update gobuild script

chrisnojima:
Back when we had split kbfs/client repos in order to build mobile we had to copy the source into a tmp dir and reconcile the vendoring. 
Since we merged repos we minimally edited this flow, but it still did the copying
This removes that and simplifies it so it just builds inside your dir. 
This speeds up the build for me and hopefully for you too

### DIFF
--- a/shared/react-native/gobuild.sh
+++ b/shared/react-native/gobuild.sh
@@ -45,8 +45,6 @@ echo "Using GOPATH: $GOPATH"
 # already in $PATH (like the old GOPATH), so put $GOPATH/bin first.
 PATH="$GOPATH/bin:$PATH"
 
-# if we don't set this gomobile init get confused
-# GOMOBILE="$GOPATH/pkg/gomobile"
 # need to whitelist some flags we use
 export CGO_CFLAGS_ALLOW="-fmodules|-fblocks"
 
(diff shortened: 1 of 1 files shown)


shared/react-native/gobuild.sh
```diff
@@ -40,66 +38,37 @@ GOPATH0=${GOPATH_ARRAY[0]}
...
-rsync -pr --ignore-times "$vendor_path/" "$GOPATH/src/"
 package="github.com/keybase/client/go/bind"
 tags=${TAGS:-"prerelease production"}
 ldflags="-X github.com/keybase/client/go/libkb.PrereleaseBuild=$keybase_build -s -w"
 
 gomobileinit ()
 {
   echo "Build gomobile..."
+  mkdir -p "$GOPATH/src/golang.org/x"
+  rsync -pr --ignore-times "$client_dir/go/vendor/golang.org/x" "$GOPATH/src/golang.org"
```
MarcoPolo: 🐑 `rsync -avz` `-a` is  superset of `-pr` and `z` adds compression `v` is for verbose. I think we can probably remove `--ignore-times` too, which should speed this up.

shared/react-native/gobuild.sh
```diff
@@ -40,66 +38,37 @@ GOPATH0=${GOPATH_ARRAY[0]}
...
   go install golang.org/x/mobile/cmd/{gomobile,gobind}
-  # iOS doesn't need gomobile init.
```
MarcoPolo: ? This isn't true anymore?

chrisnojima: It's benign

PR #19189, opened 2019-08-29
location sharing UI

mmaxim:
New location sharing UI and input bar on mobile.

**Go**
1.) Add a new map server on the local HTTP server for the purposes of rendering a location preview on the frontend.
2.) Change how `GetCurrentPosition` works to always post the last known position first, and possibly update if we get a newer coordinate within the 4 second window. 

**JS**
1.) Add new input bar styles for exploding mode and the new "more menu".
2.) Add some common slash commands to more menu.
3.) Add new location interface.


mmaxim:
cc @adamjspooner

mmaxim:
Added the new display stuff as well.

shared/chat/conversation/input-area/normal/location-popup.tsx
```diff
@@ -0,0 +1,142 @@
...
+          })
+        )
+      },
+      () => {},
+      {enableHighAccuracy: isIOS, maximumAge: 0, timeout: 30000}
+    )
+    return () => {
+      navigator.geolocation.clearWatch(watchID)
+    }
+  }, [])
```
songgao: This along with the `location` above can be extracted into a `useCurrentLocation` hook so it's clearer on the first look what's going on.

shared/chat/conversation/messages/wrapper/unfurl/map/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+import UnfurlImage from '../image'
+
+type Props = {
+  conversationIDKey: Types.ConversationIDKey
+  coord: Types.Coordinate
+  imageHeight: number
+  imageWidth: number
+  imageURL: string
+  isAuthor: boolean
+  liveLocationEndTime?: number | null
```
songgao: Why not just `liveLocationEndTime?: number`

shared/constants/types/chat2/index.tsx
```diff
@@ -115,6 +115,12 @@ export type UserReacjis = {
   skinTone: number
 }
 
+export type Coordinate = {
+  accuracy: number
+  lat: number
+  lon: number
+}
```
songgao: Make it into a Record? This hasn't been converted to immer yet.

shared/constants/types/chat2/index.tsx
```diff
@@ -167,6 +173,7 @@ export type _State = {
   createConversationError: string | null
   threadLoadStatus: I.Map<Common.ConversationIDKey, RPCChatTypes.UIChatThreadStatus>
   dismissedInviteBannersMap: I.Map<Common.ConversationIDKey, boolean>
+  lastCoord: Coordinate | null
```
songgao: Maybe just use a zero accuracy/lat/lon as empty instead of `null`? Can have this const and compare against it to test if it's populated.

```
const emptyCoordinate = makeCoordinate()
```

shared/chat/conversation/messages/attachment/image/image-render.native.tsx
```diff
@@ -86,3 +86,8 @@ export function imgMaxWidthRaw() {
   const {width: maxWidth} = Kb.NativeDimensions.get('window')
   return maxWidth
 }
+
+export function imgMaxHeightRaw() {
+  const {height: maxHeight} = Kb.NativeDimensions.get('window')
```
chrisnojima: this already exist as Styles.dimensionHeight etc

shared/chat/conversation/messages/wrapper/unfurl/map/index.tsx
mmaxim: Just matching an autogen'd type.

shared/common-adapters/index.tsx
```diff
@@ -55,6 +55,7 @@ export {default as LabeledInput} from './labeled-input'
 export {default as List} from './list'
 export {default as List2} from './list2'
 export {default as LoadingLine} from './loading-line'
+export {default as LocationMap} from './location-map'
```
chrisnojima: why is this common?

shared/constants/types/chat2/index.tsx
mmaxim: What do I get from doing that? I'm just feeding new versions of these into the store on every update.

shared/chat/conversation/messages/wrapper/unfurl/map/index.tsx
chrisnojima: i think the local types should be what we want them to be. i think its simpler if they're `foo?: string` so they're either a value or undefined and not value/undefined/null. when converting it from the rpc you can just do `|| undefined`

shared/common-adapters/index.tsx
mmaxim: I use it in two places, and thought it had potential to be used in other spots as well. Seems similar to QR scanner, which is in there.

shared/constants/types/chat2/index.tsx
songgao: We've had bugs before when using a mix of immutable types and regular JS objects, so just wanted to keep it consistent. Also with immutable you can actually check equality in the reducer to use the same object reference when the new location is the same. Probably won't make much difference though.

PR #19190, opened 2019-08-29
Update to typescript 3.6.3

chrisnojima:
Updates TS


shared/actions/chat2/index.tsx
```diff
@@ -112,18 +112,16 @@ function* inboxRefresh(
     )
   }
 
-  yield* Saga.callRPCs(
```
chrisnojima: i simplified / removed some utils from Saga. This was to help us with a flow issue but isn't necessary

shared/actions/chat2/index.tsx
```diff
@@ -2487,7 +2485,7 @@ function* mobileMessageAttachmentShare(
   }
   const fileName = yield* downloadAttachment('', message)
   try {
-    yield* Saga.callPromise(showShareActionSheetFromFile, fileName, message.fileType)
+    yield showShareActionSheetFromFile(fileName)
```
chrisnojima: note: TS caught the extra param here now

shared/actions/fs/index.tsx
```diff
@@ -696,7 +696,7 @@ function* loadPathMetadata(_: TypedState, action: FsGen.LoadPathMetadataPayload)
     let pathItem = makeEntry(dirent)
     if (pathItem.type === Types.PathType.File) {
       const mimeType = yield* _loadMimeType(path)
-      pathItem = pathItem.set('mimeType', mimeType)
+      pathItem = pathItem.set('mimeType', mimeType || null)
```
chrisnojima: correctly infers this can be undefined, which pathItem doesn't want

shared/actions/people.tsx
```diff
@@ -44,7 +44,13 @@ const getPeopleData = async (state: Container.TypedState, action: PeopleGen.GetP
       .reduce(Constants.reduceRPCItemToPeopleItem, I.List())
 
     if (debugTodo) {
-      const allTodos: Array<RPCTypes.HomeScreenTodoType> = Object.values(RPCTypes.HomeScreenTodoType)
```
chrisnojima: values is actually a mix of the numbers and the strings, so we need to be more explicit

chrisnojima:
https://github.com/microsoft/TypeScript/issues/33161

chrisnojima:
Ts should have the fix in 363 https://github.com/microsoft/TypeScript/pull/33168

chrisnojima:
@keybase/react-hackers this updates TS to 3.6 which will allow us to better type generators (etc). this is blocked on a perf regression w/ incremental builds which has been fixed but not deployed but our code can be reviewed in the meantime

shared/actions/wallets.tsx
```diff
@@ -13,7 +13,7 @@ import * as RouteTreeGen from './route-tree-gen'
 import * as Flow from '../util/flow'
 import * as Router2Constants from '../constants/router2'
 import HiddenString from '../util/hidden-string'
-import logger from '../logger'
+import _logger from '../logger'
```
jakob223: why?

shared/chat/conversation/list-area/normal/index.stories.tsx
```diff
@@ -24,7 +24,7 @@ const makeMoreOrdinals = (
   num = __STORYSHOT__ ? 10 : 100
 ): I.List<Types.Ordinal> => {
   if (direction === 'prepend') {
-    const oldStart = ordinals.size ? Types.ordinalToNumber(ordinals.first()) : firstOrdinal
+    const oldStart = ordinals.size ? Types.ordinalToNumber(ordinals.first() as Types.Ordinal) : firstOrdinal
```
jakob223: is this working around a bug or ignoring the fact that `ordinals` could be empty?

shared/actions/wallets.tsx
chrisnojima: this is because i have eslint warning us on variable shadowing. honestly most cases of _logger should likely go away and use the better saga logger instead

shared/chat/conversation/list-area/normal/index.stories.tsx
chrisnojima: TS doesn't understand that checking size means first() is non null

chrisnojima:
@keybase/react-hackers bump. previously approved but did a small merge master / updated to 3.6.3

PR #19195, opened 2019-08-29
paper key input screen per mocks

pzduniak:
@keybase/react-hackers 
@keybase/y2ksquad


shared/provision/paper-key/index.tsx
```diff
@@ -1,24 +1,45 @@
...
+            disabled: !this.state.paperKey,
+            label: 'Continue',
+            onClick: this._onSubmit,
+            type: 'Success',
+            waiting: props.waiting,
+          },
+        ]}
+        noBackground={true}
+        onBack={this.props.onBack}
+        title="Enter your paper key"
```
buoyad: Let's try to fit this without ellipsizing
image

shared/chat/conversation/rekey/enter-paper-key.tsx
```diff
@@ -2,13 +2,15 @@ import EnterPaperkey from '../../../provision/paper-key'
...
+import * as Constants from '../../../constants/provision'
 
 type OwnProps = {}
 
 export default Container.connect(
-  () => ({
+  state => ({
     error: '',
     hint: '',
+    waiting: Container.anyWaiting(state, Constants.waitingKey),
```
buoyad: Is this rekey dialog connected to the provision waiting key? If not can you change this?

shared/provision/paper-key/container.tsx
```diff
@@ -30,8 +34,11 @@ export default Container.connect(
       error: stateProps.error,
       hint: stateProps.hint,
       onBack:
-        loggedInAccounts.length > 0 ? () => dispatchProps._onLogIn(loggedInAccounts[0] || '') : undefined,
+        loggedInAccounts.length > 0
+          ? () => dispatchProps._onLogIn(loggedInAccounts[0] || '')
+          : dispatchProps._onBack,
```
buoyad: Have you tested this case?

pzduniak: Which one? The user switching case wasn't changed, I'd rather not mess around with it as it's out of scope of the rework.

shared/chat/conversation/rekey/enter-paper-key.tsx
pzduniak: woops, shouldnt have copypasted this, fixed

shared/provision/paper-key/container.tsx
buoyad: The case where you're logging in on a fresh install with a paper key and hit "back" on that screen. I think it should take you back to join-or-login

pzduniak: Yeah, it works.

PR #19202, opened 2019-08-29
Don't emit initial intents.

MarcoPolo:
This clears the intent data so we don't end up using stale intent data.
@keybase/react-hackers


MarcoPolo:
I purposely didn't clean up the code by removing unused parameters to keep this change as small as possible.

MarcoPolo:
cc @jzila

MarcoPolo:
@jzila I'm sorry I didn't realize sharing was in a semi-working state. The cold start story is still broken, but no reason to make it worse.

I changed this. I'll update the PR text

PR #19204, opened 2019-08-29
include padding when autoresizing multiline inputs

hyperobject:
Fixes an issue where the attachment caption text box would shrink to one line after typing in it due to its `minHeight` being less than `rowsMin` + padding. 

@keybase/react-hackers 
@keybase/hotpotatosquad

### DIFF
--- a/shared/chat/conversation/attachment-get-titles/index.tsx
+++ b/shared/chat/conversation/attachment-get-titles/index.tsx
@@ -115,6 +115,7 @@ class GetTitles extends React.Component<Props, State> {
                 placeholder={titleHint}
                 multiline={true}
                 rowsMin={2}
+                padding="tiny"
                 value={info.title}
                 onEnterKeyDown={this._onNext}
                 onChangeText={this._updateTitle}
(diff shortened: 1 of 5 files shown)


shared/common-adapters/plain-input.d.ts
```diff
@@ -72,6 +72,7 @@ export type Props = {
   onBlur?: () => void
   onChangeText?: (text: string) => void
   onFocus?: () => void
+  padding?: keyof typeof globalMargins
```
chrisnojima: Mobile will also have to support this

shared/chat/conversation/attachment-get-titles/index.tsx
```diff
@@ -204,10 +205,10 @@ const styles = Styles.styleSheetCreate({
       borderWidth: 1,
       marginBottom: Styles.globalMargins.tiny,
       // RN wasn't obeying `padding: Styles.globalMargins.tiny`.
-      paddingBottom: Styles.globalMargins.tiny,
-      paddingLeft: Styles.globalMargins.tiny,
-      paddingRight: Styles.globalMargins.tiny,
-      paddingTop: Styles.globalMargins.tiny,
+      // paddingBottom: Styles.globalMargins.tiny,
```
chrisnojima: :fire:

shared/chat/conversation/attachment-get-titles/index.tsx
```diff
@@ -115,6 +115,7 @@ class GetTitles extends React.Component<Props, State> {
                 placeholder={titleHint}
                 multiline={true}
                 rowsMin={2}
+                padding="tiny"
```
chrisnojima: when we hoist up a style into a prop we need to blacklist it. see https://github.com/keybase/client/blob/master/shared/common-adapters/icon.d.ts#L8

shared/common-adapters/plain-input.d.ts
hyperobject: Is there something I need to fix to make sure mobile supports this?

hyperobject: I _think_ mobile should support this now. lmk if I messed something up with the .native file

hyperobject: tested on ios and android, everything seems to work fine

PR #19206, opened 2019-08-29
path info popup for kbfs paths in chat and action popup in fs tab

songgao:
image

image

depends on songgao/HOTPOT-603


shared/fs/common/last-modified-line.tsx
```diff
@@ -0,0 +1,65 @@
...
+    !!props.lastModifiedTimestamp &&
+    (props.mode === 'row' ? '' : 'Last modified ') +
+      formatTimeForFS(props.lastModifiedTimestamp, props.mode !== 'row')
+  const by = !!props.lastWriter && (
+    <>
+      &nbsp;by&nbsp;
+      <Username mode={props.mode} lastWriter={props.lastWriter} />
+    </>
+  )
+  const getText = (children: React.ReactNode) => (
```
MarcoPolo: There's so much branching here it's basically two different things, and I think this would be hard to maintain.

I think you should remove this inner closure and just do a normal:
```
if (props.mode === 'row') {
 ...
} else {
...
}
```

shared/fs/common/path-info.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+const usePathInfo = (path: Types.Path, knownPathInfo: Types.PathInfo): Types.PathInfo => {
+  const pathInfo = Container.useSelector(state => state.fs.pathInfos.get(path, Constants.emptyPathInfo))
+  const dispatch = Container.useDispatch()
+  const alreadyKnown = knownPathInfo !== Constants.emptyPathInfo
+  React.useEffect(() => {
+    if (alreadyKnown) {
+      dispatch(FsGen.createLoadedPathInfo({path, pathInfo: knownPathInfo}))
+    } else {
```
MarcoPolo:
```suggestion
    } else if (pathInfo === Constants.emptyPathInfo) {
```

shared/fs/common/path-info.tsx
```diff
@@ -0,0 +1,85 @@
...
+          </Kb.Text>
+        </>
+      ) : null}
+    </Kb.Box2>
+  )
+}
+
+export default PathInfo_
+
+const styles = Styles.styleSheetCreate({
```
MarcoPolo: Use the new closure syntax to support darkmode
```suggestion
const styles = Styles.styleSheetCreate(() => ({ ... }))
```

shared/fs/common/path-item-info.tsx
```diff
@@ -1,58 +1,122 @@
...
-    <Kb.Text type="BodySmall">{lastWriter}</Kb.Text>
-  ) : (
-    <Kb.ConnectedUsernames
-      type="BodySmallSecondaryLink"
-      usernames={[lastWriter]}
-      inline={true}
-      onUsernameClicked="profile"
-      underline={true}
-    />
+const getChildrenNumbers = (
```
MarcoPolo: A comment would be nice. I can't tell what this does based on the signature. Maybe rename it? `getNumberOfFilesAndFolders`

shared/fs/common/path-item-info.tsx
```diff
@@ -1,58 +1,122 @@
...
+  path: Types.Path
+): {folders: number; files: number} => {
+  const pathItem = pathItems.get(path, Constants.unknownPathItem)
+  return pathItem.type === Types.PathType.Folder
+    ? pathItem.children.reduce(
+        ({folders, files}, p) => {
+          const isFolder =
+            pathItems.get(Types.pathConcat(path, p), Constants.unknownPathItem).type === Types.PathType.Folder
+          return {
+            files: files + (isFolder ? 0 : 1),
```
MarcoPolo: this isn't technically correct, is it? If the item is unknownPathItem (or anything else that isn't a file or a folder) then this is wrong

shared/fs/common/path-item-info.tsx
```diff
@@ -1,58 +1,122 @@
...
+    </Kb.Text>
   )
+}
+
+const getTlfInfoLineOrLastModifiedLine = (path: Types.Path) => {
+  switch (Types.getPathLevel(path)) {
+    case 0:
+    case 1:
+    case 2:
+      return null
```
MarcoPolo: why?

shared/fs/common/path-item-info.tsx
```diff
@@ -1,58 +1,122 @@
...
+    : {files: 0, folders: 0}
+}
+
+const FilesAndFoldersCount = (props: Props) => {
+  useFsChildren(props.path)
+  const pathItems = Container.useSelector(state => state.fs.pathItems)
+  const {files, folders} = getChildrenNumbers(pathItems, props.path)
+  return (
+    <Kb.Text type="BodySmall">
+      {folders ? `${folders} Folder${folders > 1 ? 's' : ''}${files ? ', ' : ''}` : undefined}
```
MarcoPolo: https://github.com/keybase/client/blob/master/shared/util/string.tsx#L2

shared/reducers/fs.tsx
```diff
@@ -634,6 +638,7 @@ export default function(state: Types.State = initialState, action: FsGen.Actions
     case FsGen.unsubscribe:
     case FsGen.pollJournalStatus:
     case FsGen.refreshMountDirsAfter10s:
+    case FsGen.loadPathInfo:
```
MarcoPolo: Unrelated, but consider using       `ifTSCComplainsAboutThisFunctionYouHaventHandledAllCasesInASwitch(action)`

protocol/avdl/keybase1/common.avdl
```diff
@@ -217,6 +217,13 @@ protocol Common {
     GUI_HELPER_4
   }
 
+  @lint("ignore")
```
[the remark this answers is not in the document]
songgao: This one's actually fine; will remove.

go/service/kbfs_mount.go
```diff
@@ -53,3 +57,29 @@ func (h *KBFSMountHandler) SetCurrentMountDir(_ context.Context, drive string) (
...
+	}
+	if backslash {
+		return strings.ReplaceAll(afterMount, "/", `\`)
+	}
+	return afterMount
+}
+
+func getKBFSDeeplinkPath(afterKeybase string) string {
+	if len(afterKeybase) == 0 {
+		return ""
```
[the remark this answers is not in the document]
songgao: I don't think so. This being empty means the path is`/keybase`.  It's a valid KBFS path but if we get a deeplink path for it it'd be `keybase://` which seems too general.

go/chat/utils/kbfs_path_utils.go
```diff
@@ -157,19 +134,16 @@ func ParseKBFSPaths(ctx context.Context, body string) (paths []chat1.KBFSPath) {
...
 	for _, match := range outerMatches {
 		if match.isKBFSPath() {
-			var platformAfterMountPath string
-			if libkb.RuntimeGroup() == keybase1.RuntimeGroup_WINDOWSLIKE {
-				platformAfterMountPath = match.afterMountPath(true)
-			} else {
-				platformAfterMountPath = match.afterMountPath(false)
+			kbfsPathInfo, err := libkb.GetKBFSPathInfo(match.standardPath())
+			if err != nil {
+				continue
```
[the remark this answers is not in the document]
songgao: Hmm not sure it's a good idea; this runs over chat messages which could contain non-valid paths, and we should just skip. If we log with content it'd leak sensitive data. For debugging, it should be easy to repro anyway if user can tell us the path that fails.

shared/fs/common/path-item-info.tsx
songgao: Because we don't want to display that for /keybase and /keybase/private -- there's nothing to display there

songgao:
@MarcoPolo ready for another look; thanks!

shared/fs/common/path-item-info.tsx
MarcoPolo: Leave a comment :)

PR #19208, opened 2019-08-29
Finesse sign up.

adamjspooner:


shared/settings/account/add-modals.tsx
```diff
@@ -88,8 +89,12 @@ export const Email = () => {
           searchable={searchable}
           onChangeSearchable={onChangeSearchable}
           onContinue={onContinue}
-          icon={
-            <Kb.Icon type={Styles.isMobile ? 'icon-email-add-96' : 'icon-email-add-64'} style={styles.icon} />
+          iconType={
+            Styles.isMobile
```
buoyad: `Styles.isMobile && Platform.isLargeScreen ? "big" : "small"`

adamjspooner: Is that why the dimension styles were on them?

buoyad: Yes

buoyad: (I'm fine with removing them, rather do it in `Icon` than add more plumbing to make it work with your changes here)

adamjspooner: Works for me.

PR #19210, opened 2019-08-29
folder_branch_ops: possibly fast-forward when clearing conflicts

strib:
Includes a test that fails on master.

Issue: HOTPOT-720

### DIFF
--- a/go/kbfs/libkbfs/kbfs_cr_test.go
+++ b/go/kbfs/libkbfs/kbfs_cr_test.go
@@ -2025,3 +2025,101 @@ func TestForceStuckConflict(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, children, 1)
 }
+
+// Tests that if clearing a CR conflict can fast-forward if needed.
+func TestBasicCRFailureClearAndFastForward(t *testing.T) {
+	// simulate two users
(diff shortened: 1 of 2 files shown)


go/kbfs/libkbfs/kbfs_cr_test.go
```diff
@@ -2025,3 +2025,95 @@ func TestForceStuckConflict(t *testing.T) {
...
+		mtime = mtime.Add(1 * time.Minute)
+		err = kbfsOps1.SetMtime(ctx, fileB1, &mtime)
+		err = kbfsOps1.SyncAll(ctx, fileB1.GetFolderBranch())
+		require.NoError(t, err)
+	}
+
+	t.Log("Ensure only conflict files are in the conflict view")
+	children, err := kbfsOps2.GetDirChildren(ctx, rootNode2)
+	require.NoError(t, err)
+	require.Len(t, children, 1+(maxConflictResolutionAttempts+1))
```
MarcoPolo: Why is this +2? (same q for below too)

strib: The `ForceStuckConflictForTesting` call makes `maxConflictResolutionAttempts+1` new files in the root directory while it is forcing the conflict, and the directory `a` was made earlier by the test.

PR #19221, opened 2019-08-30
Allow deeplinking to a specific Chat message

cjb:
@keybase/react-hackers CC @keybase/picnicsquad @keybase/design 

A present for @mmaxim :)  Adds a "Copy a link to this message" option to the message dropdown, and URL handler for going to that conversation and highlighting the message.

Perhaps this should be behind a feature flag while we try it out?  Since it exposes the deeplinking feature to users, and solidifies the URL format (because people will start keeping and sharing the links).

I plumbed this through `selectConversation`, because we won't know the conversationID for the linked message until we get to the bottom of `selectConversation`.


malgorithms:
cool feature!

shared/constants/chat2/meta.tsx
```diff
@@ -414,6 +414,16 @@ export const getParticipantSuggestions = (state: TypedState, id: Types.Conversat
   return _getParticipantSuggestionsMemoized(participants, teamType)
 }
 
+export const getConversationLabel = (conv: Types.ConversationMeta, includeChannelName: boolean): string => {
```
jakob223: maybe call the second arg `alwaysIncludeChannelName`?

cjb:
@mikem How about this, 3fc3e21d7e?  I see the sequencing happening properly in the console, though that doesn't guarantee there's no race.

cjb:
@mmaxim I've revived this PR against up to date master now, want to give it a try? I didn't have to alter the code to get it to do what inbox search seems to, which was surprising because I remember it not working properly under some circumstances last time around..

shared/actions/chat2/index.tsx
```diff
@@ -1151,6 +1151,22 @@ function* loadMoreMessages(
       )
     }
 
+    if (
```
mmaxim, line 1154: Why is this not in the navigateToThread case above?

shared/actions/json/chat2.json
```diff
@@ -38,6 +37,7 @@
       "teamname?": "string",
       "channelname?": "string",
       "conversationIDKey?": "Types.ConversationIDKey", // we only use this when we click on channel mentions. we could maybe change that plumbing but keeping it for now
+      "highlightMessageID?": "number | null",
```
chrisnojima, line 40: this stuff should be number or undefind, not null

cjb:
@mmaxim @chrisnojima Thanks! Got it working with a single thread load, ready for another look.

mmaxim:
Did you try on mobile? Unfortunately we do a pretty bad job in centering on mobile, and I wonder if when using this feature it will be extra tough since you won't know what the message is you are jumping to.

cjb:
@mmaxim yep, seems to work perfectly for me on mobile, tested by taking a link to a very old message in a convo that hadn't been opened on that device, and it loaded and highlighted the msg in the center of the convo

cjb:
@chrisnojima did you want to take another look? I think this is ready to go now

PR #19222, opened 2019-08-31
immer config

chrisnojima:


shared/util/set.tsx
```diff
@@ -0,0 +1,3 @@
+export function intersect<T>(a: Set<T>, b: Set<T>): Set<T> {
+  return new Set([...a].filter(x => b.has(x)))
```
songgao:
```suggestion
  return new Set(a.size > b.size ? [...b].filter(x => a.has(x)) : [...a].filter(x => b.has(x)))
```

shared/menubar/remote-proxy.desktop.tsx
```diff
@@ -175,7 +176,7 @@ export default Container.namedConnect(
         ? SafeElectron.getRemote().BrowserWindow.fromId(stateProps._externalRemoteWindowID)
         : null,
       fileRows: {_tlfUpdates: stateProps._tlfUpdates, _uploads: stateProps._uploads},
-      following: stateProps._following,
+      // following: stateProps._following,
```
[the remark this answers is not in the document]
chrisnojima: removed

shared/util/set.tsx
chrisnojima: ok but honestly i'm sure we could do a set of a million items and not have any difference here

PR #19229, opened 2019-09-02
Reworked fresh install screen

pzduniak:
image
image

@keybase/y2ksquad 
@keybase/react-hackers


shared/login/join-or-login/index.tsx
```diff
@@ -1,82 +1,77 @@
...
+            onClick={props.onLogin}
+            style={styles.button}
+            fullWidth={true}
+          />
+          {!props.isOnline && (
+            <Kb.Button
+              label="Configure a proxy"
+              mode="Secondary"
+              onClick={props.showProxySettings}
+              style={styles.button}
```
buoyad: `fullWidth={true}`

shared/login/join-or-login/index.tsx
```diff
@@ -1,82 +1,77 @@
...
+            fullWidth={true}
+          />
+          <Kb.Button
+            label="Log in"
+            mode="Secondary"
+            onClick={props.onLogin}
+            style={styles.button}
+            fullWidth={true}
+          />
+          {!props.isOnline && (
```
buoyad: Not sure this is doing what's intended. It stays on for me even when online.

shared/login/join-or-login/index.tsx
```diff
@@ -1,82 +1,77 @@
...
-      Send us feedback! Run{' '}
-      <Kb.Text type="TerminalInline" selectable={true}>
-        keybase log send
-      </Kb.Text>{' '}
-      from the terminal.
-    </Kb.Text>
-  )
-
 const Intro = (props: Props) => {
   Kb.useInterval(() => props.checkIsOnline, 2000)
```
buoyad: Oh, this is wrong. `() => props.checkIsOnline()`

buoyad: We should add a `useEffect` call to this too, otherwise "Configure a proxy" shows for 2s until we check the first time.

pzduniak: Found a better way to handle it, during the first 2s the prop is undefined rather than false.

pzduniak:
Ready for another pass.

PR #16265, opened 2019-02-26
Songgao/kbp cors custom err pages

songgao:

PR #19231, opened 2019-09-03
kbp: add cors support

songgao:
@strrib whenever you have time I'd love a review from you. Thanks! I deployed this and it seems to work fine.

Depends on https://github.com/keybase/client/pull/16265

### DIFF
--- a/go/kbfs/kbpagesconfig/cmd_per_path.go
+++ b/go/kbfs/kbpagesconfig/cmd_per_path.go
@@ -172,11 +172,21 @@ var perPathSetPermissionCmd = cli.Command{
 	},
 }
 
+// This isn't supported by kbpagesd at this time.
+const customPagesEnabled = false
+
+func getPerPathSetCmdUsageText() string {
+	customPage := ""
+	if customPagesEnabled {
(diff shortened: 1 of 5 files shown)


go/kbfs/kbpagesconfig/cmd_per_path.go
```diff
@@ -172,11 +172,21 @@ var perPathSetPermissionCmd = cli.Command{
 	},
 }
 
+// This isn't supported by kbpagesd at this time.
```
strib: Can you explain why in this comment?  I don't get why this is being disabled in this PR.

songgao: `kbpagesd` just ignores that field for now. I got caught up in thinking if it should be a redirect (i.e. URL address changes to the specified page), or just serve the content over the same address. The latter seems better but the relative paths referenced inside the page may break. Then there's also the browser caching headers -- with redirect we can use the last-modified thing but if we serve the content directly we'd probably need to generate [eTag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) ourselves. So I just sat on this and decided to only do the CORS header for this PR.

PR #19235, opened 2019-09-03
Use SearchFilter in chat inbox

mlsteele:
@keybase/react-hackers 

## macOS

image

image

## iOS

image
image


mmaxim:
I think it would be a mistake to drop the Command K hint for the box. Was that on purpose?

mmaxim:
Oh wait it is still there, I just missed it.

PR #19239, opened 2019-09-03
line height and line clamp tweaks in team builder 

jakob223:
cc @keybase/y2ksquad 

Issue: Y2K-359 reported by @zapu 
Before: 
Simulator Screen Shot - iPhone 5s - 2019-08-30 at 12 05 31

After:
Simulator Screen Shot - iPhone 5s - 2019-09-03 at 13 29 35

### DIFF
--- a/shared/team-building/user-result.native.tsx
+++ b/shared/team-building/user-result.native.tsx
@@ -145,7 +145,7 @@ const Username = (props: {
       </>
     ) : (
       <>
-        <Kb.Text type="BodySemibold" lineClamp={2}>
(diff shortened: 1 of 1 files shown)


shared/team-building/user-result.native.tsx
```diff
@@ -207,6 +214,7 @@ const styles = Styles.styleSheetCreate(() => ({
   rowContainer: {
     ...Styles.padding(Styles.globalMargins.tiny, Styles.globalMargins.xsmall),
     height: userResultHeight,
+    overflow: 'visible',
```
chrisnojima: still needed?

jakob223: nope.

PR #19240, opened 2019-09-03
do not configure logging for AllowRoot commands

heronhaye:


go/libkb/globals.go
```diff
@@ -391,7 +391,7 @@ func (g *GlobalContext) LogoutUsernameWithSecretKill(mctx MetaContext, username
 	return nil
 }
 
-func (g *GlobalContext) ConfigureLogging() error {
+func (g *GlobalContext) ConfigureLogging(usage *Usage) error {
```
strib: Might be worth a quick function comment to say when `usage` can be `nil`.

heronhaye: ok, I will do this in the next PR to not re trigger CI - thanks for looking!

heronhaye:
I'm merging over this TestEphemeralTeambotEK flake, 4th time today

PR #19244, opened 2019-09-03
Tweaks to team building team-in-progress

buoyad:
- Focus search input after someone is added
- Make non-kb "usernames" all black, they were grey before
- Make scrollbar align with team-in-progress container
- Don't show service logos - show avatar placeholders for all services
- Re-add '@service' suffix to SBS people
- Show tooltip when hovering entire container, not just (x)
- Make (x) black on hover

Punted:
- Format phone numbers (Y2K-588)

cc @keybase/picnicsquad @keybase/y2ksquad


shared/common-adapters/name-with-icon/index.tsx
```diff
@@ -116,7 +116,7 @@ class NameWithIcon extends React.Component<NameWithIconProps> {
         usernames={[username]}
         colorBroken={this.props.colorBroken}
         colorFollowing={this.props.colorFollowing}
-        colorYou={this.props.notFollowingColorOverride}
+        colorYou={this.props.notFollowingColorOverride || true}
```
chrisnojima: ????

buoyad: `Usernames` expects a string color to override your own username or `true` to set it to black. The ticket requested we change it in team building but it seemed fine to change it for `NameWithIcon`

chrisnojima: this makes it always true?

buoyad: This makes it so your own username in `NameWithIcon` is black by default

chrisnojima: Why have the parameter then

PR #19246, opened 2019-09-03
use dot only if inbetween things

chrisnojima:


chrisnojima:
@keybase/react-hackers bump, tiny pr

shared/fs/common/tlf-info-line.tsx
```diff
@@ -87,18 +80,34 @@ const getText = (props: Props) => {
...
+  const text = getText(props)
+  return (
+    <Kb.Box2
+      direction="horizontal"
+      fullWidth={true}
+      centerChildren={props.mode === 'default'}
+      alignItems="center"
+    >
+      {prefix}
+      {prefix && (reset || text) ? dot : null}
```
buoyad: I still see some of these
image

buoyad: LMK if I can help debug

chrisnojima: Yeah cause I was never able to repro in the first place

songgao: to reproduce maybe just set this to 0? https://github.com/keybase/client/blob/938410b53d87b0d651718e9e63f4936e09b1d3c4/shared/actions/fs/index.tsx#L108

shared/fs/common/tlf-info-line.tsx
```diff
@@ -87,18 +80,34 @@ const getText = (props: Props) => {
...
-)
+const TlfInfoLine = (props: Props) => {
+  const prefix = <PrefixText {...props} />
+  const dot = (
+    <Kb.Text
+      type="BodySmall"
+      style={props.mode === 'default' ? styles.textDefault : styles.textRow}
+      lineClamp={props.mode === 'row' && Styles.isMobile ? 1 : undefined}
+    >
+      •&nbsp;
```
songgao: Does `&nbsp;` work on mobile?

shared/fs/common/tlf-info-line.tsx
chrisnojima: ok so the issue is the siblings i'm checking always exist, even though inside they render to 0. they don't really serve a purpose being actual components so i just made them functions so we get either null or some other component and this logic works

PR #19247, opened 2019-09-03
Provisioning - password input screen rework

pzduniak:
@keybase/y2ksquad 
@keybase/react-hackers


shared/provision/password/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+          <Kb.LabeledInput
+            autoFocus={true}
+            placeholder="Password"
+            onEnterKeyDown={_onSubmit}
+            onChangeText={setPassword}
+            value={password}
+            textType="BodySemibold"
+            type="password"
+          />
+          <Kb.Text
```
buoyad: Impossible to tap this on a 5S-sized screen
image

shared/provision/password/container.tsx
```diff
@@ -1,76 +1,21 @@
...
-      />
-    )
-  }
-}
+type OwnProps = {}
 
 const mapStateToProps = state => ({
-  error: state.provision.error.stringValue(),
-  waitingForResponse: WaitingConstants.anyWaiting(state, Constants.waitingKey),
+  error: state.provision.error.stringValue() || '',
```
buoyad: shouldn't need `|| ''`

chrisnojima:
This should be fixed now. Not deferred

buoyad:
@chrisnojima I mention that because the problem might be in `UserCard`, not exactly in scope of this ticket

chrisnojima:
It doesn't have to be in this pr but it shouldn't be ticketed and deferred

PR #19249, opened 2019-09-03
join fixups

mmaxim:


shared/chat/conversation/messages/system-joined/container.tsx
```diff
@@ -12,7 +12,7 @@ const mapStateToProps = (state, {message}) => ({
   _meta: Constants.getMeta(state, message.conversationIDKey),
   author: message.author,
   authorIsYou: state.config.username === message.author,
-  joiners: message.joiners,
+  joiners: !message.joiners.length && !message.leavers.length ? [message.author] : message.joiners,
```
joshblum: why is this necessary? why doesn't the service do this?

mmaxim: The problem is if you are looking at the thread, it doesn't go through the whole consolidation path.

mmaxim:
Right now the join/leave/join is like "mikem, mikem joined. mikem left". Is that anymore clear?

joshblum:
no the status quo isn't great either

shared/chat/conversation/messages/system-joined/container.tsx
joshblum: oh so this was preexisting?

mmaxim:
I guess we can not consolidate in that case

go/chat/uithreadloader.go
```diff
@@ -76,11 +77,25 @@ func (t *UIThreadLoader) groupThreadView(ctx context.Context, uid gregor1.UID, t
...
+			if err != nil {
+				return false
+			}
+			if !(mtyp == chat1.MessageType_JOIN || mtyp == chat1.MessageType_LEAVE) {
+				return false
+			}
+			if msg.Valid().ClientHeader.Sender.Eq(uid) {
+				return false
+			}
+			for _, g := range grouped {
```
joshblum: might be better to have a global join/leave map so we don't have to traverse the loop repeatedly

mmaxim: This is the current set of consolidated messages. It won't be long.

go/chat/uithreadloader.go
```diff
@@ -59,14 +59,16 @@ func (t *UIThreadLoader) groupGeneric(ctx context.Context, uid gregor1.UID, msgs
...
-				grouped = append(grouped, msg)
-				continue
-			}
+		if msg.IsValid() && matches(msg, grouped) {
+			grouped = append(grouped, msg)
+			continue
 		}
 		addGrouped()
-		res = append(res, msg)
+		if msg.IsValid() && matches(msg, grouped) {
```
joshblum: worthwhile to leave a comment as to why this is necessary

PR #19254, opened 2019-09-04
wrap incoming engine calls in batch

chrisnojima:
@keybase/react-hackers this wraps dispatches from the engine in react/redux's batch. I believe this will slightly reduce renders when we have a ton of incoming rpcs. I actually was hosing the app by settings breakpoints and toggling wifi on/off. SetState would overflow due to incoming rpcs and react would die. This stop that from happening also

### DIFF
--- a/shared/engine/index-impl.tsx
+++ b/shared/engine/index-impl.tsx
@@ -12,7 +12,8 @@ import engineSaga from './saga'
 import {throttle} from 'lodash-es'
 import {CustomResponseIncomingCallMapType, IncomingCallMapType} from '.'
 import {SessionID, SessionIDKey, WaitingHandlerType, MethodKey} from './types'
-import {TypedState, Dispatch} from '../util/container'
+import {TypedActions, TypedState, Dispatch} from '../util/container'
+import {batch} from 'react-redux'
 
 type WaitingKey = string | Array<string>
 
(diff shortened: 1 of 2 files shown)


chrisnojima:
Note i think we could gain a lot by doing this w/ redux-saga but we need an api thats in 1.x

shared/engine/index-impl.tsx
```diff
@@ -35,8 +36,31 @@ class Engine {
   _hasConnected: boolean = isMobile // mobile is always connected
   // App tells us when the sagas are done loading so we can start emitting events
   _sagasAreReady: boolean = false
+
+  static _realDispatch: Dispatch
```
buoyad: Why use `static` for these?

chrisnojima: its static in engine so other things can just use it and not have an instance (like session etc). Before this PR dispatch was static, now these helpers of dispatch are also

shared/store/configure-store.tsx
```diff
@@ -76,6 +76,8 @@ const errorCatching = () => next => action => {
     if (lastError.message === error.message) {
       return
     }
+    // eslint-disable-next-line
+    debugger
```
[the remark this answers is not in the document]
chrisnojima: tnx

shared/engine/index-impl.tsx
buoyad: Ah k, afaict they're not used outside this class (session uses `deprecatedGetDispatch`). Slightly confusing.

PR #19266, opened 2019-09-04
location followup

mmaxim:
image

image

cc @cecileboucheron @adamjspooner @malgorithms


mmaxim:
Difference is live location button subtitles and the lack of green start pin.

mmaxim:
Added a `subLabel` prop to `Button` too, cc @buoyad

mmaxim:
All the shots changes are from that extra box.

PR #19267, opened 2019-09-04
input monitor as state machine

chrisnojima:
This updates the input monitor to be a state machine.
While debugging a ton of stuff i was getting my input monitor into some racy states. this should make the state of the thing more explicit and easier to understand
Also moves the callback out of the constructor

### DIFF
--- a/shared/actions/platform-specific/input-monitor.desktop.tsx
+++ b/shared/actions/platform-specific/input-monitor.desktop.tsx
@@ -2,68 +2,99 @@ const timeToConsiderActiveForAwhile = 300000
 const timeToConsiderInactive = 60000
 
 type NotifyActiveFunction = (isActive: boolean) => void
-// 5 minutes after being active, consider us inactive after
-// an additional minute of no input
-// for the purpose of marking chat messages read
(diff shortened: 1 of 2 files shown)


shared/actions/platform-specific/input-monitor.desktop.tsx
```diff
@@ -2,68 +2,118 @@ const timeToConsiderActiveForAwhile = 300000
...
-    this.notifyActive = notifyActive
-    // go into listening mode again
-    window.addEventListener('focus', this.goListening)
-    window.addEventListener('blur', this.goInactive)
-    this.goListening()
+  constructor() {
+    window.addEventListener('focus', this.onFocus)
+    window.addEventListener('blur', this.onBlur)
+    this.state = 'appInactive'
+    this.transition('focus')
```
songgao: Could it race? If user opens Keybase and make it background immediately, it seems it should not transition to focus here.

chrisnojima: i think its fine to start in this state. we have the mouse/keyboard listeners and when it detects something it goes into active

PR #19269, opened 2019-09-04
recover username rework

pzduniak:
@keybase/y2ksquad 
@keybase/react-hackers


shared/actions/provision.tsx
```diff
@@ -506,17 +506,31 @@ const showFinalErrorPage = (state: Container.TypedState, action: ProvisionGen.Sh
...
+      {email: action.payload.email},
+      Constants.forgotUsernameWaitingKey
+    )
+      .then(() => ProvisionGen.createForgotUsernameResult({result: 'success'}))
+      .catch(error =>
+        ProvisionGen.createForgotUsernameResult({
+          result: Constants.decodeForgotUsernameError(error),
+        })
+      )
+  } else {
```
jakob223: worth it to check here for a phone and error if there isn't one?

shared/provision/forgot-username/index.tsx
```diff
@@ -1,93 +1,142 @@
...
+  onSubmit: (email: string, phone: string) => void
+  waiting: boolean
 }
 
-type State = {
-  email: string
-}
+const ForgotUsername = (props: Props) => {
+  const [emailSelected, _setEmailSelected] = React.useState(true)
+  const [phoneSelected, _setPhoneSelected] = React.useState(false)
```
jakob223: You don't really need two state variables here, do you?

delete `   const [phoneSelected, _setPhoneSelected] = React.useState(false) `, the lines that call `_setPhoneSelected`, and just set `phoneSelected` to `!emailSelected`.

shared/provision/forgot-username/index.tsx
```diff
@@ -1,93 +1,142 @@
...
+      _setEmailSelected(!newState)
+      _setPhoneSelected(newState)
+    },
+    [_setEmailSelected, _setPhoneSelected]
+  )
 
-class ForgotUsername extends React.Component<Props, State> {
-  state = {email: ''}
+  const [email, setEmail] = React.useState('')
+  const [phoneNumber, setPhoneNumber] = React.useState<string | null>(null)
```
jakob223: why null and not just empty string? If someone has entered some phone number and then deleted it again, we don't want to be willing to send this.

shared/actions/provision.tsx
pzduniak: It's protected by the component's logic, I'll explain this below.

shared/provision/forgot-username/index.tsx
pzduniak: Good call!

shared/provision/forgot-username/index.tsx
pzduniak: onChange from the phone number input returns (text, valid) in the callback. If valid is false, setPhoneNumber(null) is executed, therefore the only time when "phoneNumber" is a string, it must be a valid number.

If you think that this is too complicated I can make it an empty string instead.

jakob223: Maybe add a comment to that effect somewhere, but sounds fine to me.

PR #19273, opened 2019-09-04
TS styles

chrisnojima:
This uses styles from the `@types of RN/react`. In a lot of places a raw string type is now an explicit list of strings, therefore the objects we use need to be const so as to not get widened
practically speaking that means

old:
```
Styles.styleSheetCreate({...}) // etc
```

new: (closure for dark mode)
```
Styles.styleSheetCreate(() => ({...} as const ))
```

https://github.com/keybase/client/pull/19273/files?w=1 is your friend


shared/tracker2/bio/index.tsx
```diff
@@ -180,42 +180,45 @@ const Bio = (p: Props) => (
...
-  floatingContainer: Styles.platformStyles({
-    common: {
-      backgroundColor: Styles.globalColors.purple,
-    },
-    isElectron: {
-      maxWidth: 200,
-    },
-  }),
-  fullName: Styles.platformStyles({
-    isElectron: {wordBreak: 'break-any'},
```
chrisnojima: incorrect!

shared/styles/css.d.ts
```diff
@@ -1,516 +1,35 @@
-export type Color = null | string
-type Dimension = string | number
+import {CSSProperties} from 'react'
```
chrisnojima: the 'meat' of the changes in this PR

chrisnojima: pull in styles from react and RN

shared/styles/css.d.ts
```diff
@@ -1,516 +1,35 @@
...
-  textDecorationColor?: Color
-  writingDirection?: 'auto' | 'ltr' | 'rtl'
-  tintColor?: Color
-  overlayColor?: string
+type _StylesCrossPlatform = {
+  [k in keyof _StylesDesktop]: k extends keyof _StylesCrossPlatformOverride // use override
+    ? _StylesCrossPlatformOverride[k] // or if its shared between desktop and mobile choose one which extends the other
+    : k extends keyof _StylesMobile
+    ? _StylesMobile[k] extends _StylesDesktop[k]
+      ? _StylesMobile[k]
```
chrisnojima: we prefer mobile to desktop as its more constrained. the css styles from react mostly are of the form
` 'left' | 'right' | undefined | string ` that last `| string` makes that type just string which is very annoying

shared/styles/index.types.d.ts
```diff
@@ -1,103 +0,0 @@
-// Styles from our designers
```
chrisnojima: not used

shared/styles/status-bar.d.ts
```diff
@@ -1,3 +0,0 @@
-import {EmitterSubscription} from 'react-native'
```
chrisnojima: not used

shared/people/follow-notification/index.tsx
```diff
@@ -101,32 +101,35 @@ export const MultiFollowNotification = (props: Props) => {
...
+        },
+        isMobile: {
+          ...Styles.globalStyles.flexBoxRow,
+          flexWrap: 'wrap',
+          height: 32,
+          overflow: 'hidden',
+          width: '100%',
+        },
+      }),
+      usernames: Styles.platformStyles({isElectron: {whiteSpace: 'normal'} as const}),
```
buoyad: Impressively the only storyshot change in here 😎

PR #19275, opened 2019-09-04
Handle multiple UIs in JSON api

joshblum:
fixes an issue where the JSON api would use the first chatUI what was registered regardless of what the caller tried to regegister. `ChatAPIUI` and `ChatCLIUI` (both used within the API) now have a random sessionID parameter which delegates ui rpc calls via the sessionID


joshblum:
merging over known flake

PR #19278, opened 2019-09-04
gpg always no-auto-check-trustdb

mlsteele:


mlsteele:
ignoring seemingly unrelated ci

PR #19279, opened 2019-09-04
Make enter key work for 'Continue' on new convo->phone number

cjb:

### DIFF
--- a/shared/team-building/phone-search.tsx
+++ b/shared/team-building/phone-search.tsx
@@ -59,6 +59,9 @@ const PhoneSearch = (props: PhoneSearchProps) => {
   }
 
   let _onContinue = React.useCallback(() => {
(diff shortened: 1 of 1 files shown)


shared/team-building/phone-search.tsx
```diff
@@ -86,6 +86,7 @@ const PhoneSearch = (props: PhoneSearchProps) => {
             key={phoneInputKey}
             autoFocus={true}
             onChangeNumber={onChangeNumberCb}
+            onEnterKeyDown={_onContinue}
```
buoyad: The continue button is disabled on `!validity`, maybe this callback can bail in that case

cjb: @buoyad Great catch, thanks!

PR #19282, opened 2019-09-04
Format phone numbers in user bubbles on teambuilding

cjb:

### DIFF
--- a/shared/team-building/team-box.tsx
+++ b/shared/team-building/team-box.tsx
@@ -25,21 +25,21 @@ type Props = {
 }
 
 const formatNameForUserBubble = (u: SelectedUser) => {
-  let technicalName: string
+  let displayName: string
   switch (u.service) {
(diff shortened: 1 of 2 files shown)


shared/team-building/team-box.tsx
```diff
@@ -28,10 +29,12 @@ const formatNameForUserBubble = (u: SelectedUser) => {
   switch (u.service) {
     case 'keybase':
     case 'contact': // do not display "michal@keyba.se on email" or similar
-    case 'phone':
     case 'email':
       technicalName = u.username
       break
+    case 'phone':
+      technicalName = formatPhoneNumber(u.username)
+      break
```
mlsteele: The variable name `technicalName` reads like it's making a statement but I'm not sure what. Could be that it goes to the service, but it doesn't seem to. I suggest renaming it to something with less import like `name` or `displayName`

PR #19292, opened 2019-09-05
account reset overlay storybook

pzduniak:
Text is a placeholder pending DESIGN-2616, icon is pending #18822 (though this PR should go first). The idea is that this will be slightly customized and all the plumbing will be done in Y2K-160.

@keybase/react-hackers 
@keybase/y2ksquad


shared/login/reset-modal/container.tsx
```diff
@@ -0,0 +1,17 @@
+import * as Container from '../../util/container'
```
buoyad: IMO let's delete this until it's integrated

pzduniak: it's there to make it testable, otherwise you pretty much cant see the whole thing in action

pzduniak: BUT this whole thing might be changed, there's been changes in what we want to put in here

buoyad: It's usually the integration's job to add the container and in-app entry point for this at once

PR #19293, opened 2019-09-05
libfuse: return ENOTSUP for hard links

strib:
I'm not sure how we never noticed this before -- I guess I though bazil did this for us, but it returns EIO if we don't implemented the `NodeLinker` interface.

Credit to @mateon1 for uncovering this issue and bringing it to my attention!

Issue: HOTPOT-781

### DIFF
--- a/go/kbfs/libfuse/folderlist.go
+++ b/go/kbfs/libfuse/folderlist.go
@@ -280,6 +280,22 @@ func (fl *FolderList) Remove(ctx context.Context, req *fuse.RemoveRequest) (err 
 	}
 }
 
+var _ fs.NodeLinker = (*FolderList)(nil)
+
+// Symlink implements the fs.NodeLinker interface for FolderList.
+func (fl *FolderList) Symlink(
+	_ context.Context, _ *fuse.SymlinkRequest) (fs.Node, error) {
+	return nil, fuse.ENOTSUP
+}
+
+var _ fs.NodeLinker = (*FolderList)(nil)
+
+// Link implements the fs.NodeLinker interface for FolderList.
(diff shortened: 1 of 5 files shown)


go/kbfs/libfuse/folderlist.go
```diff
@@ -280,6 +280,22 @@ func (fl *FolderList) Remove(ctx context.Context, req *fuse.RemoveRequest) (err
 	}
 }
 
+var _ fs.NodeLinker = (*FolderList)(nil)
+
+// Symlink implements the fs.NodeLinker interface for FolderList.
```
songgao: s/NodeLinker/NodeSymlinker

go/kbfs/libfuse/fs.go
```diff
@@ -572,6 +572,22 @@ func (r *Root) ReadDirAll(ctx context.Context) (res []fuse.Dirent, err error) {
 	return res, nil
 }
 
+var _ fs.NodeLinker = (*Root)(nil)
+
+// Symlink implements the fs.NodeLinker interface for Root.
```
songgao: same ere

strib:
Thanks for the review!

> also maybe add some symlink tests (in addition to link) as well since it's added too?

I think the symlinks under `/keybase` and `/keybase/private` aren't really worth testing for -- unlike hard links, there isn't any expectation that they must return a certain error type.  Especially in read-only directories, I think any error is fine, I just kind of threw that in as a cleanup.  Even for hard links in read-only directories, other errors are fine (on macOS actually, link attempts return `EPERM` because it checks for write access first).

PR #19295, opened 2019-09-05
fix flake on EphemeralBackgroundPurger

joshblum:
remove background draining/extra conv loads after fix in https://github.com/keybase/client/pull/19255

```
$ go test -run TestBackgroundPurge  -count 5
PASS
ok      github.com/keybase/client/go/chat       25.006s
```

cc @mmaxim


mmaxim:
sick

PR #19296, opened 2019-09-05
Use team-building apiSearch to search emails too

buoyad:
cc @keybase/y2ksquad

### DIFF
--- a/shared/team-building/email-input/index.tsx
+++ b/shared/team-building/email-input/index.tsx
@@ -1,36 +1,37 @@
 import * as React from 'react'
-import {debounce} from 'lodash-es'
 import * as Container from '../../util/container'
 import * as Kb from '../../common-adapters'
 import * as Styles from '../../styles'
 import * as TeamBuildingGen from '../../actions/team-building-gen'
+import * as Constants from '../../constants/team-building'
+import * as Types from '../../constants/types/team-building'
 import {AllowedNamespace} from '../../constants/types/team-building'
 import {validateEmailAddress} from '../../util/email-address'
 import ContinueButton from '../continue-button'
 
 type EmailInputProps = {
   namespace: AllowedNamespace
+  search: (query: string, service: Types.ServiceIdWithContact) => void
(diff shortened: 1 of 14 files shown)


shared/team-building/email-input/index.tsx
```diff
@@ -1,36 +1,37 @@
...
+  if (
+    teamBuildingSearchResults &&
+    teamBuildingSearchResults[emailString] &&
+    teamBuildingSearchResults[emailString].keybase &&
+    teamBuildingSearchResults[emailString].keybase[0]
+  ) {
+    user = teamBuildingSearchResults[emailString].keybase[0]
+  }
+  const canSubmit = !!user && !waiting && isEmailValid
+  const emailHasKeybaseAccount = user && user.serviceMap.keybase !== '' && emailString === user.username
```
MarcoPolo: > `emailString === user.username`
In this case the user.username is email, is that right?

shared/team-building/email-input/index.tsx
```diff
@@ -1,36 +1,37 @@
...
 import * as TeamBuildingGen from '../../actions/team-building-gen'
+import * as Constants from '../../constants/team-building'
+import * as Types from '../../constants/types/team-building'
 import {AllowedNamespace} from '../../constants/types/team-building'
 import {validateEmailAddress} from '../../util/email-address'
 import ContinueButton from '../continue-button'
 
 type EmailInputProps = {
   namespace: AllowedNamespace
+  search: (query: string, service: Types.ServiceIdWithContact) => void
```
MarcoPolo: 🐑 
```suggestion
  search: (query: string, service: 'email') => void
```

A small guarantee that this code will ever only search emails

shared/team-building/email-input/index.tsx
buoyad: Yeah it is. This check already existed but I think it can be removed, since we already look up by `emailString` to get user.

shared/team-building/email-input/index.tsx
buoyad: SGTM, I'll add it to the part 3 PR. Thanks for looking at these!

PR #19299, opened 2019-09-05
Remove searchResultCounts

buoyad:
And dedup some UserMatchMention code. 

This has the side effect that when entering a query other tabs are not preloaded anymore - so we trigger a load on switching tabs with this. On the other hand, that's a perf price we no longer have to pay. cc @keybase/picnicsquad

### DIFF
--- a/shared/team-building/container.tsx
+++ b/shared/team-building/container.tsx
@@ -352,6 +352,19 @@ const deriveRolePickerArrowKeyFns = memoize(
   })
 )
 
+const deriveOnChangeService = memoize(
+  (
+    onChangeService: OwnProps['onChangeService'],
+    search: (text: string, service: Types.ServiceIdWithContact) => void,
+    searchString: string
+  ) => (service: Types.ServiceIdWithContact) => {
+    onChangeService(service)
+    if (!Types.isContactServiceId(service)) {
+      search(searchString, service)
+    }
+  }
+)
+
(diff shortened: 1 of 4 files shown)


shared/team-building/container.tsx
```diff
@@ -566,7 +570,14 @@ const mergeProps = (
         ],
         title,
       }
-    : {}
+    : emptyObj
+
+  const onChangeService = (service: Types.ServiceIdWithContact) => {
+    ownProps.onChangeService(service)
+    if (!Types.isContactServiceId(service)) {
+      dispatchProps._search(ownProps.searchString, service)
```
MarcoPolo: What happens when it is contact service id?

shared/team-building/container.tsx
```diff
@@ -566,7 +570,14 @@ const mergeProps = (
         ],
         title,
       }
-    : {}
+    : emptyObj
+
+  const onChangeService = (service: Types.ServiceIdWithContact) => {
```
MarcoPolo: Everywhere else here we are being careful to memoize functions to avoid new functions everytime any prop changes. Maybe we should do that here too?

shared/team-building/container.tsx
buoyad: The email/phone tabs don't share `searchString` with the rest of the tabs, so they'll start out empty and call `search` when they see a valid phone or email.

PR #19300, opened 2019-09-05
SearchFilter size prop

mlsteele:


shared/common-adapters/search-filter.tsx
```diff
@@ -25,7 +25,7 @@ const Kb = {
 type Props = {
   icon?: IconType | null
   focusOnMount?: boolean
-  fullWidth?: boolean
+  size: 'small' | 'full-width' // only affects desktop (https://zpl.io/aMW5AG3)
```
mlsteele: @songgao what do you think? Because of Box2's props.fullWidth, I expected SearchFilter's `props.fullWidth` to make it grow to fill the width. But instead it means something else. This way it's less reminiscent of Box2.

shared/chat/conversation-filter-input.tsx
```diff
@@ -66,6 +66,7 @@ class ConversationFilterInput extends React.PureComponent<Props> {
     const searchInput = (
       <Kb.SearchFilter
         ref={this._setRef}
+        size={'small'}
```
songgao: `size="small"`

PR #19302, opened 2019-09-05
New Convo - Don't cutoff profile pictures with 5+ users in convo

thebearjew:
1. Reducing `paddingLeft` on the user bar to align the leftmost profile picture with the search bar
2. Shows horizontal scrollbar when scrolling on desktop. Also, on macOS when System Preferences > General > Show Scrollbars: always is enabled, no scrollbar is shown under the user bubbles. In this case, the scrollbar is rendered below both.

### Current Padding
Screenshot 2019-09-06 10 30 00

### New Padding
Screenshot 2019-09-06 10 28 59


thebearjew:
@keybase/react-hackers

PR #19304, opened 2019-09-05
de-dupe conv loader jobs

mmaxim:
Goal is to de-dupe jobs into the conv loader, while giving special treatment to jobs that have a special post load hook.

### DIFF
--- a/go/chat/convloader.go
+++ b/go/chat/convloader.go
@@ -34,15 +34,17 @@ type clTask struct {
 
 type jobQueue struct {
 	sync.Mutex
-	queue   *list.List
-	waitChs []chan struct{}
-	maxSize int
+	queue    *list.List
+	waitChs  []chan struct{}
+	queueMap map[string]bool
+	maxSize  int
 }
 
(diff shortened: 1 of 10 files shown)


go/chat/convloader.go
```diff
@@ -453,28 +470,31 @@ func (b *BackgroundConvLoader) load(ictx context.Context, task clTask, uid grego
...
-		chat1.GetThreadReason_BACKGROUNDCONVLOAD, query, pagination)
-	if err != nil {
-		b.Debug(ctx, "load: ConvSource.Pull error: %s (%T)", err, err)
-		if b.retriableError(err) && task.attempt+1 < bgLoaderMaxAttempts {
-			b.Debug(ctx, "transient error, retrying")
-			task.attempt++
-			task.lastAttemptAt = time.Now()
-			return &task
+	var tv chat1.ThreadView
+	if pagination.Num > 0 {
```
joshblum: I would prefer this was an explicit option to not run `Pull`

PR #19306, opened 2019-09-06
move download logic to Go

songgao:
* [ ] download path encoding (https://github.com/keybase/client/pull/19368)
* [x] test and figure out the download situation on Android


go/kbfs/simplefs/downloads.go
```diff
@@ -0,0 +1,353 @@
...
+		return keybase1.DownloadInfo{}, err
+	}
+	return d.info, nil
+}
+
+func (m *downloadManager) configureDownload(cacheDirOverride string, downloadDirOverride string) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	if len(cacheDirOverride) > 0 {
+		m.cacheDir = cacheDirOverride
```
strib: I wonder if we should try `mkdir`ing and/or `stat`ing these new directories, to make sure they are valid (would require returning an error from this function).  Not sure what the right semantics are.

go/kbfs/simplefs/simplefs.go
```diff
@@ -535,34 +539,28 @@ func (k *SimpleFS) setResult(opid keybase1.OpID, val interface{}) {
...
-		w.progress.EndEstimate = keybase1.ToTime(k.config.Clock().Now())
-	}
-	k.lock.Unlock()
-	if ok {
-		w.done <- err
-		close(w.done)
-	}
+func (k *SimpleFS) doneOp(ctx context.Context, opid keybase1.OpID, w *inprogress, err error) {
+	w.done <- err
+	close(w.done)
```
strib: Nothing is setting the `EndEstimate` anymore for non-download async ops -- maybe not strictly required, but seems nice to have.

go/kbfs/simplefs/downloads.go
```diff
@@ -0,0 +1,353 @@
...
+	d, err := m.getDownload(downloadID)
+	if err != nil {
+		return err
+	}
+	return m.k.SimpleFSCancel(ctx, d.opid)
+}
+
+func (m *downloadManager) dismissDownload(
+	ctx context.Context, downloadID string) {
+	m.cancelDownload(ctx, downloadID)
```
strib: Not checking the error here.

protocol/avdl/keybase1/simple_fs.avdl
```diff
@@ -571,4 +572,35 @@ protocol SimpleFS {
...
+    string error;
+    boolean done;
+    boolean canceled;
+  }
+  record DownloadStatus {
+    array<string> regularDownloadIDs; // the ones showing up in the GUI footer
+    array<DownloadState> states; // all downloads including non-regular ones
+  }
+  // returns download ID. isRegularDownload causes moving to the Downloads
+  // folder. Save on mobile and share to other apps use isRegularDownload=false.
```
strib: So "regular" means something that shows up in the footer?  Why don't "save on mobile" count as "regular"?  Seems like the name could be a bit more descriptive.

go/kbfs/simplefs/downloads.go
```diff
@@ -0,0 +1,353 @@
...
+					d.state.Progress = float64(
+						resp.BytesWritten) / float64(resp.BytesTotal)
+					return d
+				}); err != nil {
+					done(err)
+					return
+				}
+			case errNoResult:
+				// This is from simpleFS. Likely download has finished, but
+				// wait for the done ch.
```
strib: What does "wait for the done ch" mean?  In this code, we never see the done channel.

go/kbfs/simplefs/downloads.go
```diff
@@ -0,0 +1,353 @@
...
+)
+
+type download struct {
+	info  keybase1.DownloadInfo
+	state keybase1.DownloadState
+	opid  keybase1.OpID
+}
+
+// downloadManager manages downloads.
+type downloadManager struct {
```
strib: Any chance we can get some tests for this type please?

go/kbfs/simplefs/downloads.go
```diff
@@ -0,0 +1,353 @@
...
+			opid: opid,
+			state: keybase1.DownloadState{
+				DownloadID: downloadID,
+			},
+		}
+	}()
+
+	bgCtx, cancelBtCtx := m.makeContext()
+	done := func(err error) {
+		_ = m.updateDownload(downloadID, func(d download) download {
```
strib: I'm not quite sure -- could this function be called twice (once by the monitor and once by the waiter) and race to update `d.state`?  Not sure if that would cause problems or not, but might trigger a data race test failure if this is ever tested.

go/kbfs/simplefs/downloads.go
```diff
@@ -0,0 +1,353 @@
...
+			} else {
+				d.state.EndEstimate = keybase1.ToTime(time.Now())
+				d.state.Progress = 1
+				d.state.Done = true
+			}
+			return d
+		})
+		cancelBtCtx()
+	}
+	go m.monitorDownload(bgCtx, opid, downloadID, done)
```
strib: A little nervous about these goroutines leaking, especially in a test scenario, but I guess we have other stuff like this in SimpleFS already so it's probably not much worse.

protocol/avdl/keybase1/simple_fs.avdl
```diff
@@ -571,4 +572,35 @@ protocol SimpleFS {
...
+  string simpleFSStartDownload(KBFSPath path, boolean isRegularDownload);
+  DownloadInfo simpleFSGetDownloadInfo(string downloadID);
+  DownloadStatus simpleFSGetDownloadStatus();
+  void simpleFSCancelDownload(string downloadID);
+  // Remove a download from the download manager. If the download is still
+  // running, this cancels it as well.
+  void simpleFSDismissDownload(string downloadID);
```
strib: Seems like this is a required call to free memory in the download manager, right?  If so, that should probably be mentioned in the comment.

go/kbfs/libkbfs/test_common.go
```diff
@@ -49,7 +49,7 @@ const (
 // TODO: Move more common code here.
 func newConfigForTest(modeType InitModeType, loggerFn func(module string) logger.Logger) *ConfigLocal {
 	mode := modeTest{NewInitModeFromType(modeType)}
-	config := NewConfigLocal(mode, loggerFn, "", DiskCacheModeOff, &env.KBFSContext{})
+	config := NewConfigLocal(mode, loggerFn, "", DiskCacheModeOff, env.NewContext())
```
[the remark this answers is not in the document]
songgao: It just needs core Env. Perhaps we can have a stubbed version of global context that just serves that?

go/kbfs/simplefs/downloads.go
```diff
@@ -0,0 +1,353 @@
...
+	dlCtxIDKey dlCtxTagKey = iota
+)
+
+type download struct {
+	info  keybase1.DownloadInfo
+	state keybase1.DownloadState
+	opid  keybase1.OpID
+}
+
+// downloadManager manages downloads.
```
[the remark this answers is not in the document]
songgao: No, no curl style. I guess download is just the GUI perspective. How about something like `kbfsContentRetriever` or `kbfsContentExporter`?

go/kbfs/libkbfs/interfaces.go
```diff
@@ -2040,6 +2040,7 @@ type SubscriptionManager interface {
 // SubscriptionManagerPublisher associates with one SubscriptionManager, and is
 // used to publish changes to subscribers mangaged by it.
 type SubscriptionManagerPublisher interface {
+	DownloadStatusChanged()
```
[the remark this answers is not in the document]
songgao: Good call! I'll try; doesn't seem too complicated to change.

go/kbfs/simplefs/downloads.go
```diff
@@ -0,0 +1,353 @@
...
+	downloadDir string
+	downloads   map[string]download // download ID -> download
+}
+
+func newDownloadManager(simpleFS *SimpleFS) *downloadManager {
+	return &downloadManager{
+		k:           simpleFS,
+		publisher:   simpleFS.config.SubscriptionManagerPublisher(),
+		cacheDir:    simpleFS.config.KbEnv().GetCacheDir(),
+		downloadDir: filepath.Join(simpleFS.config.KbEnv().GetHome(), "Downloads"),
```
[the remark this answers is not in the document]
songgao: It's what frontend [uses](https://github.com/keybase/client/blob/fd0c1a1ead4a4acb2674a8d6c585249ab8ce2da0/shared/util/file.desktop.tsx#L31) today on desktop. It's different on Android, which is why ConfigureDownload exists.

go/kbfs/simplefs/downloads.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	_, filename = path.Split(path.Clean(kbfsPath.Path))
+	downloadPath = filepath.Join(parentDir, downloadID+path.Ext(kbfsPath.Path))
+	return downloadPath, filename, nil
+}
+
+func (m *downloadManager) moveToDownloadFolder(
+	ctx context.Context, srcPath string, filename string) (localPath string, err error) {
+	if libkb.GetPlatformString() == "ios" {
+		return "", errors.New("MoveToDownloadFolder is not supported on iOS")
```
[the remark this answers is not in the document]
songgao: There's no download on iOS; just saving to the photos library and sharing to other apps, both of which are handled in JS after the download (to the cache dir) finishes.

go/kbfs/simplefs/simplefs.go
songgao: Good point; I'll add it back

protocol/avdl/keybase1/simple_fs.avdl
songgao: > So "regular" means something that shows up in the footer?

Yes that's correct, plus that it's moved to the Downloads folder after it's done. It feels to be purely a GUI term,. I'll clarify in the comment.

go/kbfs/simplefs/downloads.go
songgao: it's possible the download entry is already removed by now; this just makes sure it's canceled in case caller didn't call cancel explicitly. I'll add a comment.

go/kbfs/simplefs/downloads.go
songgao: I think I meant to say `ctx.Done()`, which gets canceled through `done()`. Will fix; sorry!

go/kbfs/simplefs/downloads.go
songgao: They won't really race they way `-race` detects since `updateDownload` locks. Did you mean we'd get in correct result if they update in unfortunate order? It doesn't seem possible to me as all conditions are checked at the beginning, but please let me know if I missed something!

go/kbfs/simplefs/downloads.go
songgao: Yeah we already have these in similar manner in simplefs. But perhaps we can clean them up if idle (idle as in no GetDownloadStatus calls) for more than e.g. 1min? Currently GUI should query the status every second (the frequency notification happens).

songgao:
@strib thanks for the detailed review!

go/kbfs/libkbfs/test_common.go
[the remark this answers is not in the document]
strib: Sure, if you can figure out exactly which `g.Configure*` call is required, it would be good to call only that one.

go/kbfs/simplefs/downloads.go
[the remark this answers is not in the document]
strib: I think I'm ok with having "download" in this name under the simplefs package, but I just want there to be a big comment here explaining what it means.

go/kbfs/simplefs/downloads.go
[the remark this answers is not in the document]
strib: Would be nice to have that explained in a comment here.

go/kbfs/simplefs/downloads.go
strib: I meant that the linter is going to yell at you.  You have to at least do `_ =` here.  Though maybe `cancelDownload` shouldn't throw an error if it's already canceled, if that's an expected condition?

go/kbfs/simplefs/downloads.go
strib: Eh, I wouldn't change it for now, just something to keep in mind.

go/kbfs/simplefs/downloads.go
strib: Ah ok, I missed that `updateDownload()` locks.  Thanks.

go/kbfs/simplefs/downloads.go
songgao: Actually we already make the dir in getDownloadPath(), so we should be fine.

go/kbfs/simplefs/downloads.go
songgao: it actually doesn't error if it's already canceled, but errors if a download is not found (i.e. dismissed). I'm gonna clarify in comment.

shared/actions/fs/index.tsx
```diff
@@ -1008,13 +920,46 @@ const loadPathInfo = async (_: TypedState, action: FsGen.LoadPathInfoPayload) =>
   })
 }
 
+const loadDownloadInfo = (_: TypedState, action: FsGen.LoadDownloadInfoPayload) =>
```
chrisnojima: lets use async await for all new actions

songgao:
@strib mind dipping your thumb here again?

go/kbfs/simplefs/downloads.go
songgao: Got it! I think we're fine here; if there's an error writing into that dir we'd eventually get an error for the download, which is probably what we want anyway as this way we wouldn't bother user with download related problems until they want to download something.

strib: Ok cool.  Though if we ever let the user set their own configurable directory, we might want to reconsider that.  Ok for now though!

PR #19309, opened 2019-09-06
More async await in actions

chrisnojima:


chrisnojima:
@keybase/react-hackers bump

shared/actions/wallets.tsx
```diff
@@ -1643,22 +1652,25 @@ function* loadStaticConfig(state: TypedState, action: ConfigGen.DaemonHandshakeP
...
+    const res: Saga.RPCPromiseType<
+      typeof RPCStellarTypes.localGetStaticConfigLocalRpcPromise
+    > = yield RPCStellarTypes.localGetStaticConfigLocalRpcPromise()
+    yield Saga.put(
+      WalletsGen.createStaticConfigLoaded({
+        staticConfig: I.Record(res)(),
+      })
+    )
+  } finally {
+    yield Saga.put(
```
songgao: Did you mean to move this into `finally`?

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -234,11 +231,16 @@ const onOutOfDate = (
...
+const prepareLogSend = async (
+  _: Container.TypedState,
+  action: EngineGen.Keybase1LogsendPrepareLogsendPayload
+) => {
   const response = action.payload.response
-  dumpLogs().then(() => {
+  try {
+    await dumpLogs()
+  } finally {
     response && response.result()
```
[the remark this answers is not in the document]
chrisnojima: result is responding to the rpc. it says we're done dumping the logs. we don't return an action cause there's no action side effect

shared/actions/wallets.tsx
chrisnojima: yes, these calls where we call handshakeWait true/false should always be balanced. else if it throws you'll be stuck forever

PR #19310, opened 2019-09-06
Enlarge the SearchFilter's 'X' hitbox

cjb:

Screen Shot 2019-09-06 at 10 28 58 AM

### DIFF
--- a/shared/common-adapters/search-filter.tsx
+++ b/shared/common-adapters/search-filter.tsx
@@ -224,13 +224,14 @@ class SearchFilter extends React.PureComponent<Props, State> {
     }
     if (Styles.isMobile) {
       return (
-        <Kb.Icon
-          type="iconfont-remove"
-          sizeType={this._iconSizeType()}
-          onClick={this.props.mobileCancelButton ? this._clear : this._cancel}
-          color={this._iconColor()}
-          style={styles.removeIconNonFullWidth}
-        />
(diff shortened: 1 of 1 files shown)


shared/common-adapters/search-filter.tsx
```diff
@@ -224,13 +224,14 @@ class SearchFilter extends React.PureComponent<Props, State> {
...
     if (Styles.isMobile) {
       return (
-        <Kb.Icon
-          type="iconfont-remove"
-          sizeType={this._iconSizeType()}
-          onClick={this.props.mobileCancelButton ? this._clear : this._cancel}
-          color={this._iconColor()}
-          style={styles.removeIconNonFullWidth}
-        />
+        <Kb.ClickableBox onClick={this.props.mobileCancelButton ? this._clear : this._cancel}>
```
chrisnojima: i believe you can just do this with padding, but this is ok also

cjb: Didn't seem to work, but maybe I'm doing something wrong, adding padding to the icon still gives:

Screen Shot 2019-09-06 at 11 53 33 AM

chrisnojima: Did add padding to the original or on top of this version. Likely only works with an onclick

cjb: to original

cjb:
@chrisnojima Thumb up or down on this one?

PR #19311, opened 2019-09-06
Clear tscache on postinstall

buoyad:
also remove pegjs

### DIFF
--- a/shared/desktop/yarn-helper/index.tsx
+++ b/shared/desktop/yarn-helper/index.tsx
@@ -111,21 +111,10 @@ const decorateInfo = info => {
   return temp
 }
 
-const warnFail = path => () => console.warn(`Error cleaning tscache ${path}, tsc may be inaccurate.`)
+const warnFail = err => console.warn(`Error cleaning tscache ${err}, tsc may be inaccurate.`)
 const clearTSCache = () => {
-  const tsOutDir = path.resolve(__dirname, '..', '..', '.tsOuts')
-  const reg = path.resolve(tsOutDir, '.tsOut')
-  const slow = path.resolve(tsOutDir, '.tsOutSlow')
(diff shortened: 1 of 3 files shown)


shared/package.json
```diff
@@ -272,12 +272,12 @@
     "msgpack": "git://github.com/keybase/nullModule",
     "net": "git://github.com/keybase/nullModule",
     "null-loader": "0.1.1",
-    "pegjs": "0.10.0",
     "prettier": "1.17.1",
     "randexp": "0.5.3",
     "react-hot-loader": "4.11.0",
     "react-test-renderer": "16.8.4",
     "redux-logger": "3.0.6",
+    "rimraf": "3.0.0",
```
chrisnojima: just use 2.2.8 so we don't have any new code

buoyad: Making it 2.6.3 which we also already have makes fewer lockfile changes, i'll use that one

shared/desktop/yarn-helper/index.tsx
```diff
@@ -109,6 +111,23 @@ const decorateInfo = info => {
   return temp
 }
 
+const warnFail = path => () => console.warn(`Error cleaning tscache ${path}, tsc may be inaccurate.`)
+const clearTSCache = () => {
+  const tsOutDir = path.resolve(__dirname, '..', '..', '.tsOuts')
```
chrisnojima: you can readDirSync and just blow away everything other than README.md

buoyad: `rimraf` supports globs, so we could have it do `.tsOuts/.tsOut*`, wdyt?

chrisnojima: even better

buoyad: Done, thanks

PR #19317, opened 2019-09-06
Update AVDL compiler to 1.4.2

nathunsmitty:
This is the one with the ts and python support. Go test it by cloning the bot libs and running make!

### DIFF
--- a/protocol/package.json
+++ b/protocol/package.json
@@ -17,7 +17,7 @@
   },
   "homepage": "https://github.com/keybase/protocol",
   "dependencies": {
-    "avdl-compiler": "1.4.1",
+    "avdl-compiler": "1.4.2",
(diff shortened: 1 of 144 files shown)


nathunsmitty:
Also removes a dead make task

protocol/package.json
```diff
@@ -17,7 +17,7 @@
   },
   "homepage": "https://github.com/keybase/protocol",
   "dependencies": {
-    "avdl-compiler": "1.4.1",
+    "avdl-compiler": "1.4.2",
```
mlsteele: i'd expect this to change all the protocol output. Did you `make clean` then `make` in `protocol`?

nathunsmitty:
moot by  #19625

PR #19318, opened 2019-09-06
some chat dark colors

chrisnojima:
This updates some chat views and fixes some issues w/ the storybook

- [x] fix auto dark flip in storybook
- [x] fix popup handling in stories
- [x] fix the 'multiColors' aka in master `blackOrWhite` was using this.black in `colors` and `this.white` in `darkColors` but this.white in darkColors is black, so its really blackOrBlack, which isn't what we want. Also some colors didn't match the names which shouldn't be the case. If the color is right the name is wrong

### DIFF
--- a/shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
+++ b/shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
@@ -375,7 +375,7 @@ const styles = Styles.styleSheetCreate(
         },
       }),
       cancelEditingText: {
-        color: Styles.globalColors.blackOrBlack,
+        color: Styles.globalColors.whiteOrBlack,
       },
       container: {
         ...Styles.globalStyles.flexBoxColumn,
--- a/shared/common-adapters/relative-popup-hoc.desktop.tsx
+++ b/shared/common-adapters/relative-popup-hoc.desktop.tsx
@@ -347,13 +347,17 @@ function ModalPositionRelative<PP>(
     )
 
     componentDidMount() {
-      document.body && document.body.addEventListener('click', this._handleClick)
-      document.body && document.body.addEventListener('scroll', this._handleScroll, true)
+      const node = document.body
+      if (!__STORYBOOK__ && node) {
+        node.addEventListener('click', this._handleClick, false)
+      }
     }
 
     componentWillUnmount() {
-      document.body && document.body.removeEventListener('click', this._handleClick)
-      document.body && document.body.removeEventListener('scroll', this._handleScroll, true)
+      const node = document.body
+      if (!__STORYBOOK__ && node) {
+        node.removeEventListener('click', this._handleClick, false)
+      }
     }
 
     _setRef = r => {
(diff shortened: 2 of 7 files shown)


shared/chat/conversation/input-area/normal/platform-input.desktop.tsx
```diff
@@ -375,7 +375,7 @@ const styles = Styles.styleSheetCreate(
         },
       }),
       cancelEditingText: {
-        color: Styles.globalColors.blackOrBlack,
+        color: Styles.globalColors.whiteOrBlack,
```
chrisnojima: fixes this problem on master

shared/common-adapters/relative-popup-hoc.desktop.tsx
```diff
@@ -347,13 +347,17 @@ function ModalPositionRelative<PP>(
     )
 
     componentDidMount() {
-      document.body && document.body.addEventListener('click', this._handleClick)
-      document.body && document.body.addEventListener('scroll', this._handleScroll, true)
+      const node = document.body
+      if (!__STORYBOOK__ && node) {
```
chrisnojima: don't do this at all in stories, messes up clicks in popups
also the scroll thing did nothing

shared/stories/index.desktop.tsx
```diff
@@ -56,13 +38,13 @@ const RootWrapper = ({children}) => {
     _autoSwap = autoSwap
   }, [darkMode, autoSwap])
 
-  useInterval(
+  Kb.useInterval(
```
chrisnojima: using the common one now

shared/stories/index.desktop.tsx
```diff
@@ -56,13 +38,13 @@ const RootWrapper = ({children}) => {
...
     () => {
       const next = !darkMode
-      setDarkMode(next)
```
chrisnojima: the order here is important, else using the auto flip only half works

shared/styles/colors.tsx
```diff
@@ -132,14 +132,14 @@ export const colors = {
 export const darkColors: {[P in keyof typeof colors]: string | undefined} = {
   black: 'rgba(255, 255, 255, 0.85)',
   get blackOrBlack() {
-    return this.black
+    return colors.black
```
chrisnojima: here the intention is `black in light mode and black in dark mode`. in `colors` its safe to refer to `this.whatever` as the name is accurate. but in darkMode `this.whatever` likely *doesn't* refer to the right color

shared/styles/colors.tsx
```diff
@@ -191,7 +191,7 @@ export const darkColors: {[P in keyof typeof colors]: string | undefined} = {
   greenLight: '#B7EED9',
   greenLighter: '#E8FAF6',
   get greenOrGreenLighter() {
-    return this.green
```
chrisnojima: this is a color change: cc: @adamjspooner

shared/styles/colors.tsx
```diff
@@ -233,18 +233,18 @@ export const darkColors: {[P in keyof typeof colors]: string | undefined} = {
   transparent_on_white: '#191919',
   white: '#191919',
   get whiteOrBlack() {
-    return this.black
+    return colors.black
   },
   get whiteOrGreenDark() {
-    return this.white
+    return colors.greenDark
```
chrisnojima: this is a color change: cc: @adamjspooner

PR #19321, opened 2019-09-06
lint: remove new dead code, and non-Linux dead code

strib:
This removes all dead code that was either introduced recently (since the incremental builds don't catch it), or was hidden behind a build tag (e.g., non-Linux) that I didn't run locally.


strib:
> not really familiar with any of this code though
@heronhaye sorry, you were added based on github suggestions, but thanks for looking!

PR #19322, opened 2019-09-06
prefetcher_test: fix a test flake, and a data race

strib:
* Don't close done ch until prefetcher is shut down, otherwise there could be a data race.
* Compare contents of file blocks, not instances. In these tests, there's a rare race where a file block in the cache is constructed from an encrypted disk-block-cache version, rather than being the specific instance created by the test and returned by the fake block getter.  That means that the hash pointer inside each block is different, and thus the generic equality test will fail. Get around that by just checking the (randomly-generated) contents for each direct file block.  It's not perfect, but seems good enough.

Issue: HOTPOT-782

### DIFF
--- a/go/kbfs/libkbfs/prefetcher_test.go
+++ b/go/kbfs/libkbfs/prefetcher_test.go
@@ -1847,6 +1847,8 @@ func TestPrefetcherReschedules(t *testing.T) {
 		context.Background(), individualTestTimeout)
 	defer cancel()
 	kmd := makeKMD()
+	prefetchDoneCh := make(chan struct{})
+	defer close(prefetchDoneCh)
 	prefetchSyncCh := make(chan struct{})
 	defer shutdownPrefetcherTest(t, q, prefetchSyncCh)
 	q.TogglePrefetcher(true, prefetchSyncCh, nil)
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/prefetcher_test.go
```diff
@@ -1847,6 +1857,8 @@ func TestPrefetcherReschedules(t *testing.T) {
 		context.Background(), individualTestTimeout)
 	defer cancel()
 	kmd := makeKMD()
+	prefetchDoneCh := make(chan struct{})
```
jzila: It's a bit confusing that this is declared all the way up here, when it isn't used until the `TogglePrefetcher` much lower. Or did you intend to use it in the `TogglePrefetcher` on line 1864?

strib: @jzila it's actually used in the `defer` right below this line, which is the important thing.  Because it needs to be closed _after_ the deferred `shutdownPrefetcherTest` call a few lines below that.  If it's shutdown before, there's a potential for a data race (see the linked HOTPOT issue) where the prefetcher tries to write to that channel after it's been closed by the test.  I'll add a comment there explaining it.

jzila: Got it, thanks.

PR #19324, opened 2019-09-07
teams: fix bug in loader w/ hidden holes

maxtaco:
- we used to make a restriction on loading that PTK gens had to be laid down monotonically increasing, but we hit a case in prod where that wasn't so
  - the test case included tests this, it involves an interaction between FTL and the slow loader
- fix what we should be testing for, that we don't get conflicting PTKs for the same geneation
- introduce a new test case for the new error condition


maxtaco:
@heronhaye  / @AMarcedone this should fix our broken team

maxtaco:
CI red is a bad test in master

PR #19330, opened 2019-09-09
Refactor how phone number and email tabs do search

zapu:
Use user search RPC with `email` or `phone` service (which are now legal in this RPC, and the typeing reflects that) instead of the `imptofuQuery` object.


zapu:
Ready for review cc @keybase/react-hackers @keybase/y2ksquad

PR #19331, opened 2019-09-09
Surya/nil/swap paper test flake

heronhaye:


heronhaye:
(test failure is flake)

patrickxb:
i'm guessing that this test failed at some point because the two words that were swapped were the same?

PR #19340, opened 2019-09-09
ios dark mode detection

chrisnojima:
Beginnings of mobile plumbing


shared/util/feature-flags.native.tsx
```diff
@@ -10,7 +10,7 @@ const ff: FeatureFlags = {
   airdrop: true,
   chatIndexProfilingEnabled: false,
   conflictResolution: false,
-  darkMode: false || __STORYBOOK__,
+  darkMode: true,
```
chrisnojima: lets play with this

PR #19348, opened 2019-09-09
add pin/unpin to json api

joshblum:


go/chat/server.go
```diff
@@ -2727,10 +2727,29 @@ func (h *Server) PinMessage(ctx context.Context, arg chat1.PinMessageArg) (res c
 	ctx = globals.ChatCtx(ctx, h.G(), keybase1.TLFIdentifyBehavior_CHAT_GUI, &identBreaks, h.identNotifier)
 	defer h.Trace(ctx, func() error { return err }, "PinMessage")()
 	defer func() { h.setResultRateLimit(ctx, &res) }()
-	_, err = utils.AssertLoggedInUID(ctx, h.G())
+	uid, err := utils.AssertLoggedInUID(ctx, h.G())
```
mmaxim: Maybe we should just do this check on the server.

joshblum: can't since the id of the msg we're pinning is in the body

joshblum: i saw this more of a nicety of trying to prevent accidental pins than a strict enforcement

mmaxim: Could include on supersedes, but sounds good.

PR #19352, opened 2019-09-09
Use Kb.Button for inbox expand.

adamjspooner:


shared/chat/inbox/row/teams-divider/index.tsx
```diff
@@ -25,16 +25,14 @@ class TeamsDivider extends React.PureComponent<Props> {
         gapEnd={true}
       >
         {this.props.showButton && (
-          <Kb.ClickableBox onClick={this.props.toggle} style={styles.containerToggleButton}>
```
buoyad: Can we remove any unused styles from here?

adamjspooner: Oh, right. Good call.

nizam87872:
Ok

PR #19353, opened 2019-09-09
New Convo: Update Start Button to use Tooltip

thebearjew:
### Design
image

### Overview
Changes `Kb.WithTooltip` component to also take React nodes as `text` so that the return icon can be rendered next to text.

Not entirely feeling this approach, so here are some things I thought about changing:

1. Rename `text` prop to something like `text`/`content`/`body`
2. Or add an additional prop in addition to `text`, something like `textComponent`

### Screenshots

image

@keybase/react-hackers

### DIFF
--- a/shared/common-adapters/with-tooltip.d.ts
+++ b/shared/common-adapters/with-tooltip.d.ts
@@ -4,7 +4,7 @@ import {Position} from './relative-popup-hoc.types'
 
 export type Props = {
   disabled?: boolean
-  text: string
+  text: string | React.ReactNode
   multiline?: boolean
   containerStyle?: StylesCrossPlatform
   children: React.ReactNode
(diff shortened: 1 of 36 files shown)


shared/common-adapters/with-tooltip.d.ts
```diff
@@ -4,7 +4,7 @@ import {Position} from './relative-popup-hoc.types'
 
 export type Props = {
   disabled?: boolean
-  text: string
+  text: string | React.ReactNode
```
thebearjew: Primary point of discussion I think

chrisnojima: just call it `tooltip`

thebearjew:
Renamed prop, @chrisnojima

PR #19354, opened 2019-09-09
Show native android logs in android log dumps (+ quick reply error handling)

MarcoPolo:
(but only when we use the NativeLogger interface)

This is kind of hacked in. I basically copy the output of what `console.log` was doing, that way when we read the logs from logcat we get the native logs as well (since there is no difference between a log line from java and one from js)

This is in effort to debug why quick reply fails sometimes

@jzila @keybase/react-hackers

### DIFF
--- a/shared/android/app/src/main/java/io/keybase/ossifrage/modules/NativeLogger.java
+++ b/shared/android/app/src/main/java/io/keybase/ossifrage/modules/NativeLogger.java
@@ -1,7 +1,6 @@
 package io.keybase.ossifrage.modules;
 
 import android.util.Log;
-import android.util.JsonWriter;
 
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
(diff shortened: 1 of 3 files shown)


shared/android/app/src/main/java/io/keybase/ossifrage/modules/NativeLogger.java
```diff
@@ -14,69 +13,46 @@
...
-            rawLog(ERROR_TAG, "Exception in dumpLine: " + Log.getStackTraceString(e));
-            return toLog;
-        }
-    }
-
-    private static String dumpLine(String toLog, Throwable tr) {
-      return dumpLine(toLog + ": " + Log.getStackTraceString(tr));
+    private static String formatLine(String tagPrefix, String toLog) {
+        // Copies the Style JS outputs in native/logger.native.tsx
+        return tagPrefix + NAME + ": [" + System.currentTimeMillis() + ",\"" + toLog + "\"]";
```
jzila: Any reason this doesn't use `String.format`?

MarcoPolo: Yeah I'm pretty sure String.format is like 10x slower than a simple string concat.

PR #19359, opened 2019-09-09
Node preload

chrisnojima:
a subset of this WIP: 
https://github.com/keybase/client/pull/18893

### DIFF
--- a/shared/desktop/renderer/preload-main.shared.desktop.tsx
+++ b/shared/desktop/renderer/preload-main.shared.desktop.tsx
@@ -1,0 +1,8 @@
+// import Electron from 'electron'
+import punycode from 'punycode'
+
+const target = typeof window === 'undefined' ? global : window
+
+target.KB = {
+  punycode, // used by a dep
+}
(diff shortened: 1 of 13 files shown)


shared/desktop/renderer/preload-main.shared.desktop.tsx
```diff
@@ -0,0 +1,8 @@
+// import Electron from 'electron'
```
songgao: 👾

shared/util/file.native.tsx
```diff
@@ -1,46 +1,36 @@
 import RNFetchBlob from 'rn-fetch-blob'
 import {findAvailableFilename} from './file.shared'
+import {Encoding} from './file'
 
-import {StatResult, WriteStream, Encoding} from './file'
-
-export function tmpDir(): string {
+function tmpDir() {
```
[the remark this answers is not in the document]
chrisnojima: it's rare we need to explicitly say the return types. 1. here they're not used externally and 2. if we did need them then it should be derivable

shared/desktop/renderer/preload-main.shared.desktop.tsx
chrisnojima: i think i won't bother merging this until the next pr is done (which includes this anyways)

PR #19365, opened 2019-09-09
Mark message as read if you quick reply

MarcoPolo:
@keybase/react-hackers @mmaxim

### DIFF
--- a/go/bind/notifications.go
+++ b/go/bind/notifications.go
@@ -48,7 +48,9 @@ type ChatNotification struct {
 	BadgeCount          int
 }
 
-func HandlePostTextReply(strConvID, tlfName string, body string) (err error) {
+func HandlePostTextReply(strConvID, tlfName string, intMessageID int, body string) (err error) {
+	ctx := context.Background()
+	defer kbCtx.CTraceTimed(ctx, fmt.Sprintf("HandlePostTextReply()"), func() error { return flattenError(err) })()
 	outboxID, err := storage.NewOutboxID()
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


go/bind/notifications.go
```diff
@@ -58,6 +60,17 @@ func HandlePostTextReply(strConvID, tlfName string, body string) (err error) {
 		return err
 	}
 	_, err = kbCtx.ChatHelper.SendTextByIDNonblock(context.Background(), convID, tlfName, body, &outboxID, nil)
+
+	kbCtx.Log.CDebugf(ctx, "Marking as read from QuickReply: convID: %s", strConvID)
+	gc := globals.NewContext(kbCtx, kbChatCtx)
+	uid, err := utils.AssertLoggedInUID(ctx, gc)
```
mmaxim: Should early out here, and then just return `nil` at the bottom.

go/bind/notifications.go
```diff
@@ -58,7 +60,22 @@ func HandlePostTextReply(strConvID, tlfName string, body string) (err error) {
...
+	if err != nil {
+		return err
+	}
+
+	msgID := chat1.MessageID(intMessageID)
+	if err = kbChatCtx.InboxSource.MarkAsRead(context.Background(), convID, uid, msgID); err != nil {
+		kbCtx.Log.CDebugf(ctx, "Failed to mark as read from QuickReply: convID: %s. Err: %s", strConvID, err)
+		// We don't want to fail this method call just because we couldn't mark it as aread
+		err = nil
```
mmaxim: Don't need this anymore

PR #19386, opened 2019-09-10
various tweaks and fixes for kbfs path popup and friends

songgao:
image

image

image

cc @cecileboucheron for the styles

### DIFF
--- a/shared/common-adapters/copy-text.tsx
+++ b/shared/common-adapters/copy-text.tsx
@@ -70,7 +70,7 @@ const CopyText = (props: Props) => {
       </Toast>
       <Text
         lineClamp={lineClamp}
-        type="Body"
+        type="BodyTiny"
         selectable={true}
         center={true}
         style={styles.text}
@@ -91,11 +91,7 @@ const CopyText = (props: Props) => {
           onClick={copy}
           labelContainerStyle={styles.buttonLabelContainer}
         >
-          <Icon
-            type="iconfont-clipboard"
-            color={Styles.globalColors.white}
-            fontSize={Styles.isMobile ? 20 : 16}
-          />
+          <Icon type="iconfont-clipboard" color={Styles.globalColors.white} sizeType="Small" />
         </Button>
       )}
     </Box2>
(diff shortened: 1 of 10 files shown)


cecileboucheron:
Lgtm!

shared/common-adapters/copy-text.tsx
```diff
@@ -112,17 +108,17 @@ const styles = Styles.styleSheetCreate(
           height: undefined,
           marginLeft: 'auto',
           minWidth: undefined,
-          paddingLeft: 17,
```
jakob223: xsmall is 12, which is significantly smaller than 17. Intentional?

shared/common-adapters/copy-text.tsx
```diff
@@ -142,35 +138,34 @@ const styles = Styles.styleSheetCreate(
           overflow: 'hidden',
         },
         isMobile: {
-          minHeight: 40,
```
jakob223: Intentionally removed? if so, delete instead of commenting out, and make sure it doesn't have any side effects.

shared/common-adapters/copy-text.tsx
```diff
@@ -142,35 +138,34 @@ const styles = Styles.styleSheetCreate(
...
-          marginTop: Styles.globalMargins.xsmall / 2,
+          marginBottom: Styles.globalMargins.xtiny,
+          marginLeft: Styles.globalMargins.tiny,
+          marginRight: Styles.globalMargins.tiny,
+          marginTop: Styles.globalMargins.xtiny,
           minWidth: 0,
           textAlign: 'left',
         },
         isAndroid: {
-          position: 'relative',
```
jakob223: Same deal as above commented out code

shared/fs/common/index.tsx
```diff
@@ -19,4 +19,5 @@ export {
   useFsJournalStatus,
   useFsOnlineStatus,
   useFsPathInfo,
+  useFsSoftError,
```
jakob223: unused?

shared/common-adapters/copy-text.tsx
songgao: Yeah all these are intentional; Cecile wanted to shrink this component a bit globally

shared/fs/common/index.tsx
songgao: I was thinking to use it to drive the Oops screen selection as well but forgot; will add it!

PR #19426, opened 2019-09-10
performance improvements to the position functions and tree lookup

AMarcedone:
@maxtaco I tried implementing my solution and here are the results.
We can also try with storing the path length with the key if you think it is worth it.

Also cc @heronhaye 

Here are the profiling results
[memprofile.pdf](https://github.com/keybase/client/files/3596941/memprofile.pdf)
[profile.pdf](https://github.com/keybase/client/files/3596942/profile.pdf)

### DIFF
--- a/go/merkletree2/position.go
+++ b/go/merkletree2/position.go
@@ -29,6 +29,18 @@ func (p *Position) getBytes() []byte {
 	return (*big.Int)(p).Bytes()
 }
 
+// Set updates p to the value of q
+func (p *Position) Set(q *Position) {
+	(*big.Int)(p).Set((*big.Int)(q))
+}
+
+// Clone returns a pointer to a deep copy of a position
+func (p *Position) Clone() *Position {
+	var q Position
+	q.Set(p)
+	return &q
+}
+
 func (p *Position) isOnPathToKey(k Key) bool {
 	// If the Key is shorter than current prefix
 	if len(k)*8 < (*big.Int)(p).BitLen()-1 {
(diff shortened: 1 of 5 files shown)


go/merkletree2/tree.go
```diff
@@ -18,17 +18,31 @@ type Tree struct {
 
 	cfg Config
 	eng StorageEngine
+
+	// step is an optimization parameter for GetKeyValuePairWithProof that
+	// controls how many path positions at a time the tree requests from the
+	// storage engine. Lower values result in more storage engine request, but
```
maxtaco: request -> requests

go/merkletree2/tree.go
```diff
@@ -18,17 +18,31 @@ type Tree struct {
...
+	// less of the (somewhat expensive) bit fiddling operations.  Values higher
+	// than 63 are not recommended as the bit operations (in the best case)
+	// cannot be done using a signle 64 bit word and become more expensive. This
```
maxtaco: signle -> single

go/merkletree2/tree.go
```diff
@@ -336,15 +350,31 @@ func (t *Tree) GetKeyValuePairWithProof(ctx context.Context, tr Transaction, s S
...
-	if err != nil {
-		return KeyValuePair{}, MerkleInclusionProof{}, err
-	}
-	// Assumes nodes are returned in the same order in which they are being requested
-	siblingPosHashPairs, err := t.eng.LookupNodes(ctx, tr, s, siblingPositions)
-	if err != nil {
-		return KeyValuePair{}, MerkleInclusionProof{}, err
+	var siblingPosHashPairs []PositionHashPair
+	needMore := true
+	for curr := 1; needMore && curr <= t.cfg.keysByteLength*8/int(t.cfg.bitsPerIndex); curr += t.step + 1 {
```
maxtaco: can you make a function on `cfg` to do this division? would clear up the code

go/merkletree2/position.go
```diff
@@ -103,25 +127,47 @@ func (t *Config) getDeepestPositionForKey(k Key) (*Position, error) {
...
+func (t *Config) getDeepestPositionAtLevelAndSiblingsOnPathToKey(k Key, lastLevel int, firstLevel int) (sibs []Position) {
+
+	maxLevel := t.keysByteLength * 8 / int(t.bitsPerIndex)
+	if lastLevel > maxLevel {
+		lastLevel = maxLevel
+	}
+
+	// first, shrink the key for efficiency
+	var bytesNecessary int
+	if lastLevel*int(t.bitsPerIndex)%8 == 0 {
```
maxtaco:
```go
bytesNecessary = lastLevel * int(t.bitsPerIndex) 
if lastLevel*int(t.bitsPerIndex)%8 != 0 {
    bytesNecessary++
}
```

AMarcedone: Is this for readability? or are there other reasons for the change? Also the first line should be 
` bytesNecessary = lastLevel * int(t.bitsPerIndex) / 8`  right?

maxtaco: readability

AMarcedone:
Thanks for the quick review :)

PR #19427, opened 2019-09-10
Rework Docker images according to new specs

pzduniak:
forgot to switch the branch before starting work :/


packaging/linux/docker/build.sh
```diff
@@ -5,7 +5,7 @@ build_dir="$1"
 here="$(dirname "$BASH_SOURCE")"
 client_dir="$(git -C "$here" rev-parse --show-toplevel)"
 source_commit="$(git -C "$here" log -1 --pretty=format:%h)"
-tag="$(cat "$build_dir/VERSION" | tr "+" "-")"
+tag="$(echo $1 | tr "+" "-")"
```
heronhaye: maybe check this value before to make sure its not empty

packaging/linux/docker/standard/entrypoint.sh
```diff
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
+set -e
```
heronhaye: euo pipefail?

packaging/linux/docker/slim/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+set -e
```
heronhaye: euo pipefail?

packaging/linux/docker/slim/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+# If a command was passed and KEYBASE_SERVICE is not set, simply run the command
+if [ "$#" -gt 1 ] -a [[ -z "${KEYBASE_SERVICE}" ]]; then
+    exec "$@"
+    exit 0
+fi
+
+# If we're called with any args, don't pollute stdout and stderr and write the
+# logs into /var/log/keybase
+if [ "$#" -gt 1 ]; then
+    bash -c "keybase ${KEYBASE_SERVICE_ARGS:-"-debug"} service &> /var/log/keybase/service.log" &
```
heronhaye: why does it shell out?
also if we redirect logs like that I don't think log sends will work - is that ok?

packaging/linux/docker/build.sh
pzduniak: `set -u` already checks it

packaging/linux/docker/standard/entrypoint.sh
pzduniak: k ill do full set and i might revert it if it breaks

packaging/linux/docker/slim/entrypoint.sh
pzduniak: i _think_ it should work (you'll get 2 log files) and i can change it in the future after we do some basic tests

i'll drop the shellout, it should probably work as-is without bash -c

PR #19433, opened 2019-09-10
node more 2

chrisnojima:
- [x] Move raw imports to preload
- [x] ipc back to ipc again
- [x] validate global logger story


shared/actions/chat2/index.tsx
```diff
@@ -28,6 +28,7 @@ import * as TeamsConstants from '../../constants/teams'
 import logger from '../../logger'
 import {isMobile, isIOS} from '../../constants/platform'
 import {NotifyPopup} from '../../native/notifications'
+import {Buffer} from 'buffer/'
```
chrisnojima: note this is tricky. we need the `/` at the end so it doesn't pick it up as the node module and instead uses our polyfill that we need for mobile

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -234,11 +217,15 @@ const onOutOfDate = (
   return ConfigGen.createUpdateInfo({critical: true, isOutOfDate: true, message: upgradeMsg})
 }
 
-const prepareLogSend = (_: Container.TypedState, action: EngineGen.Keybase1LogsendPrepareLogsendPayload) => {
+const prepareLogSend = async (
```
chrisnojima: this gets called when someone does logsend from the cli

shared/actions/settings.tsx
```diff
@@ -533,8 +533,9 @@ const sendFeedback = async (state: TypedState, action: SettingsGen.SendFeedbackP
   const {feedback, sendLogs, sendMaxBytes} = action.payload
   try {
     if (sendLogs) {
-      const lines = await logger.dump()
-      await writeLogLinesToFile(lines)
+      const fromRender = await logger.dump()
```
chrisnojima: node side logs were missing!

PR #19435, opened 2019-09-10
use convIDs for api listen filters

mmaxim:
cc @malgorithms


go/client/chat_api_listen_display.go
```diff
@@ -37,22 +36,12 @@ func newMsgNotification(source string) *chat1.MsgNotification {
...
-			// to normalize display name without using resolve RPC.
-			conv, _, err := d.svc.findConversation(ctx, "", v)
-			if err != nil {
-				return fmt.Errorf("Unable to find chat channel for %q: %s", v.Name, err.Error())
-			}
-			v.Name = conv.Info.TlfName
-		} else {
-			v.Name = strings.ToLower(v.Name)
+	for _, v := range channelFilters {
+		conv, _, err := d.svc.findConversation(ctx, "", v)
```
zapu: i thought it would have to find all conversations in the team somehow to be able to support the "wildcard" for "listen to all conversations im in for the team" from line 138

mmaxim: Is anything using that functionality right now?

zapu: Hard to tell. Would the same arguments start only following `#general` now with this change?

mmaxim: I fixed it.

PR #19442, opened 2019-09-10
client: new `fs uploads` command

strib:
Lists all paths that are currently uploading, along with the number of bytes remaining and the estimated time remaining.  By default it filters out dot files, but that can be switched off with `-a`.

Example:
```sh
$ keybase fs uploads -a
/keybase/private/strib/tmp
/keybase/private/strib/tmp/.100mb.5
/keybase/private/strib/tmp/100mb.5

100.43 MB left to upload
Estimated time remaining: 13m12s
$ keybase fs uploads
/keybase/private/strib/tmp
/keybase/private/strib/tmp/100mb.5

97.25 MB left to upload
Estimated time remaining: 4m57s
$
```

Also, in simplefs, correctly filter syncing paths by base name.

Issue: HOTPOT-784

### DIFF
--- a/go/client/cmd_simplefs_uploads.go
+++ b/go/client/cmd_simplefs_uploads.go
@@ -1,0 +1,95 @@
+// Copyright 2019 Keybase, Inc. All rights reserved. Use of
+// this source code is governed by the included BSD license.
+
+package client
+
+import (
+	"context"
+	"sort"
+	"time"
+
+	"github.com/keybase/cli"
+	"github.com/keybase/client/go/libcmdline"
+	"github.com/keybase/client/go/libkb"
+	keybase1 "github.com/keybase/client/go/protocol/keybase1"
+)
+
+// CmdSimpleFSUploads is the 'fs ps' command.
+type CmdSimpleFSUploads struct {
+	libkb.Contextified
+	filter keybase1.ListFilter
+}
+
+// NewCmdSimpleFSUploads creates a new cli.Command.
+func NewCmdSimpleFSUploads(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	return cli.Command{
+		Name:  "uploads",
+		Usage: "list current uploads",
+		Action: func(c *cli.Context) {
+			cl.ChooseCommand(&CmdSimpleFSUploads{
+				Contextified: libkb.NewContextified(g)}, "uploads", c)
+			cl.SetNoStandalone()
+		},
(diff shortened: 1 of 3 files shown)


go/client/cmd_simplefs_uploads.go
```diff
@@ -0,0 +1,95 @@
...
+// CmdSimpleFSUploads is the 'fs ps' command.
+type CmdSimpleFSUploads struct {
+	libkb.Contextified
+	filter keybase1.ListFilter
+}
+
+// NewCmdSimpleFSUploads creates a new cli.Command.
+func NewCmdSimpleFSUploads(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	return cli.Command{
+		Name:  "uploads",
```
jakob223: feels a little bit odd to me as a name. Maybe `show-uploads` or something, unless you plan more subcommands? Or leave it as is if you disagree, I don't feel strongly.

go/client/cmd_simplefs_uploads.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	status, err := cli.SimpleFSSyncStatus(context.TODO(), c.filter)
+	if err != nil {
+		return err
+	}
+
+	ui := c.G().UI.GetTerminalUI()
+	if len(status.SyncingPaths) == 0 && status.TotalSyncingBytes == 0 {
+		ui.Printf("There are currently no uploads\n")
```
jakob223: s/no uploads/no uploads in progress/

go/kbfs/test/edit_history_test.go
```diff
@@ -602,3 +602,26 @@ func TestEditHistoryRenameDirAndReuseNameForLink(t *testing.T) {
 		),
 	)
 }
+
+func TestEditHistoryRenamedThenDeleted(t *testing.T) {
```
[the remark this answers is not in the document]
strib: Oops I just pushed this to the wrong branch by accident!  Will remove.

go/client/cmd_simplefs_uploads.go
strib: Yeah I see your point, but I think it's consistent with other things like `quota` and `history` -- we don't currently have any fs-level `show-` commands.  Also I imagine we eventually might need a way to cancel uploads, so maybe this will turn into `uploads show` and `uploads cancel`?  Not sure.  But I think I'll keep it as is for now and see how things evolve.

PR #19497, opened 2019-09-11
fs dark mode

hyperobject:
style object literals to stylesheetcreate calls (with new `as const` stuff). very large pr, sorry!

@keybase/react-hackers 
@keybase/hotpotatosquad


hyperobject:
@jakob223 
> Why do we need to add the `as const` to all these?

it's a change @chrisnojima requested to make style types easier/better!

PR #19509, opened 2019-09-11
Change the new convo 'Continue' button label to 'Add' once there's at least one participant

cjb:


shared/team-building/index.stories.tsx
```diff
@@ -763,6 +763,7 @@ const load = () => {
 
   Sb.storiesOf('Team-Building/Phone Search', module).add('Empty Phone Search', () => (
     <PhoneSearch
+      continueLabel="Continue"
```
thebearjew: @cjb Shouldn't this warrant a storyshot update?

cjb: @thebearjew It just means that the text didn't change from the previous storyshot which also had Continue as the label :)

PR #19551, opened 2019-09-12
Remove 'other assets' from choose-asset 'change' button

cjb:

Looks like we always want to use the path payments banner for these custom assets, so I removed all the display code for them from the choose-asset widget.  Thought about leaving the display code in and just having the container return `[]`, but then decided that getting back old code if you made a mistake is what version control is for and pulled it all out.

(Since there's now only one section competing for screen space, there's no need to have the expander logic anymore, we can just fill the dialog with the displayChoices.)


patrickxb:
I tried it, it works.

PR #19563, opened 2019-09-12
team-building wrap service bar labels with a purpose

mlsteele:
image
(ignore the icon colors, that's another thing)


shared/team-building/service-tab-bar.native.tsx
```diff
@@ -59,15 +59,19 @@ const ServiceIcon = (props: IconProps) => {
...
-              lineClamp={2}
-              // @ts-ignore: we need to allow any color here for various services
-              style={{color}}
-            >
-              {props.label}
-            </Kb.Text>
+            <Kb.Box2 direction="vertical">
+              {props.label.map((label, i) => (
+                <Kb.Text
+                  key={i}
```
thebearjew: Just a note: 
Keys need to be unique identifiers, usually using an array index as a key is not the best since when an item is added or removed, then all keys will update and re-render.
This really isn't an issue here since we're using two pieces of text.
Up to you if you want to change it

PR #19564, opened 2019-09-12
enable ephemeral keys in oneshot mode

joshblum:
enables ephemeral key generation in oneshot mode. caveats:

- if multiple services are using the same paperkey each service will fight over access to the deviceEphemeral key generation and no one will have access. error messaging indicates you are in a cloning situation and suggests using separate paperkeys
-  on logout/shutdown all ephemeral keys are forcibly purged


joshblum:
merging over unrelated flake

PR #19565, opened 2019-09-12
dark mode wallets

songgao:


shared/styles/shared.tsx
```diff
@@ -17,16 +17,25 @@ export const globalMargins = {
...
+  Announcements: 'blue',
+  Documentation: 'blueDarker',
+  HighRisk: 'red',
+  Information: 'yellow',
+  Normal: 'white',
+  Success: 'green',
+  Terminal: 'blueDarker2',
 }
 
+export const backgroundModeToColor = new Proxy(
```
chrisnojima: lets not use a proxy. why not just make backgroundModeToColor a closure and keep it simple

chrisnojima: or just make those getters instead in that const

shared/wallets/wallet-list/wallet-row/index.tsx
```diff
@@ -14,22 +14,23 @@ export type Props = {
...
 }
 
-const rightColumnStyle = Styles.platformStyles({
-  isElectron: {
-    overflow: 'hidden',
-    textOverflow: 'ellipsis',
-    whiteSpace: 'nowrap',
-  },
-})
+const rightColumnStyle = () =>
```
chrisnojima: nothing dynamic in here so it doesn't need to be a closure, but ok

PR #19590, opened 2019-09-13
GUI side of federation public memo support

cjb:
@patrickxb r?

I think this should work, want to give it some testing?  Here's a screenshot:

Screen Shot 2019-09-12 at 9 47 51 PM

For the public memo override case, the input field is disabled, memo text becomes foreground dark grey, and there's the "This memo was provided by the recipient and cannot be changed." text underneath.  Perhaps that explanation is unnecessary and people would work it out from the color and inability to edit it.

There doesn't seem to be any Go code on your branch to set publicMemoOverride to anything.  So I tried to hook it up, and it works for the testcase in the screenshot.  It doesn't do anything about the memoType field, though, and I'm not confident it's the right fix.  Maybe you could push the correct Go side change to this branch, then we'll merge this branch down onto yours after testing it all together?

I see there's a bug where changing display currency makes the memo disappear.  I'll look into it.  It's probably not caused by this branch because the "You can send assets.." banner disappears too, and that doesn't seem right.  Also to check: is it correct for the path banner to be shown?  You can do path payments to a federated address?


patrickxb:
path payments to a federated address are supported.

PR #19624, opened 2019-09-13
user-block when you chat-block an impteam 1on1 convo

xgess:
chat-blocking and user-blocking are confusingly very different things, and we have at least one log of someone who thought they blocked but they only chat-blocked. 

so, when a user chat-blocks (or reports) a 1-on-1 implicit conversation, also go ahead and user-block that person. this should only execute if it's two people in a non-team conversation, one of whom is the caller. 

i didn't implement unblocking. i figured this was unlikely to be a heavily demanded thing, it adds a bunch of complexity, and you can always manually unblock the person if you go their profile.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,6 +47,7 @@ services:
             - GREGOR_TLF_SERVER=fmprpc+tls://kbweb.local:13010
             - GREGOR_TEAM_SERVER=fmprpc+tls://kbweb.local:7890
             - GREGOR_MACCHECKER_SERVER=fmprpc://kbweb.local:13037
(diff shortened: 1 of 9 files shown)


xgess:
@keybase/picnicsquad

strib:
@xgess in the ticket PICNIC-367 that I filed, I had blocked the user from their profile.  Does this PR still fix that case?  (From the description it seems like it's fixing something different.)

xgess:
@strib the description of the ticket is fixed by https://github.com/keybase/keybase/pull/4337. this fixes the logdump that was linked. while not exactly a bug, this does fix something that's pretty unintuitive.

strib:
Ah makes sense, I missed that one.  Thanks!

PR #19625, opened 2019-09-13
avdl compiler update

malgorithms:
does it make sense that lodash is added to yarn.lock?


nathunsmitty:
Commenting on Lodash: that got added as a avdl compiler dependency, we use a couple different casing functions and uniq in python and typescript

PR #19633, opened 2019-09-13
update protocol

buoyad:


chrisnojima:
does this change anything other than the comment?

buoyad:
@chrisnojima just the comment

PR #19634, opened 2019-09-13
add FE scaffolding for autoreset

buoyad:
cc @keybase/y2ksquad


shared/util/container.tsx
```diff
@@ -55,6 +55,8 @@ export type Route = {
 }
 export type RouteMap = {[K in string]: Route}
 
+export const assertNever = (_: never) => undefined
```
buoyad: We can transition things that were using `util/flow` to use this

PR #19637, opened 2019-09-13
protocol: update node-avdl-compiler comments for 1.4.2

strib:
Looks like the version used by CI changed, so all CI runs fail on this diff now.


strib:
Merging over unrelated test flake.

PR #19639, opened 2019-09-13
client: accept "keybase://" paths in KBFS CLI commands

strib:
Suggested by jakob223.

Issue: HOTPOT-792

### DIFF
--- a/go/client/cmd_simplefs.go
+++ b/go/client/cmd_simplefs.go
@@ -56,12 +56,16 @@ func NewCmdSimpleFS(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Comm
 	}
 }
 
-const mountDir = "/keybase"
+const (
+	mountDir   = "/keybase"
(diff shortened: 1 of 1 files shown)


songgao:
`keybase://` paths can have url encoded stuff (e.g. `%20`) and we'd need to call `url.PathUnescape` on it.

Also have you considered moving some of https://github.com/keybase/client/blob/master/go/chat/utils/kbfs_path_utils.go out into `libkb` and use that? It'd be nice to reuse the same code for parsing these links in all places.

strib:
> keybase:// paths can have url encoded stuff (e.g. %20) and we'd need to call url.PathUnescape on it.

Done, thanks!

> Also have you considered moving some of /go/chat/utils/kbfs_path_utils.go@master out into libkb and use that? It'd be nice to reuse the same code for parsing these links in all places.

I have not considered that.  It seems to be doing way more work than needed for the CLI here.

PR #19659, opened 2019-09-14
chat immer 1

chrisnojima:
This moves the root of chat to immer but not any of the inner objects


chrisnojima:
- [ ] after release

PR #19679, opened 2019-09-16
libkb: strip leading whitespace for two sig ancient payloads before checking for '{'

maxtaco:
- fixes `keybase id nougad`
- See https://github.com/keybase/keybase-issues/issues/3543


maxtaco:
(only need one review whoever grabs it first)

maulana770:
i am not received reward

PR #19685, opened 2019-09-16
Handle provision cancelling

zapu:
This one fixes canceling logic in `ProvisioningManager` by canceling pending RPC by calling `error` on their `response` object, remembering that it's canceled, and swallowing "Input canceled" error on the call sites. 

There is a new event `cancelProvision` that is dispatched from `componentWillUnmount` in `code-page` component which ensures that pending provisioning is properly canceled. This is safe to do even if we do not intend to cancel, e.g. provisioning succeeded or there was another error beforehand. `cancelProvision` calls `ProvisioningManager.maybeCancelProvision` which we consider OK to call if `ProvisioningManager` is already `done()`.

There are some misc fixes around error modal, but the real fix will come in Y2K-655.

cc @keybase/react-hackers @keybase/y2ksquad

### DIFF
--- a/shared/actions/provision.tsx
+++ b/shared/actions/provision.tsx
@@ -541,7 +541,12 @@ const showPaperkeyPage = (state: Container.TypedState) =>
 
 const showFinalErrorPage = (state: Container.TypedState, action: ProvisionGen.ShowFinalErrorPagePayload) => {
   const parentPath = action.payload.fromDeviceAdd ? devicesRoot : ['login']
-  return RouteTreeGen.createNavigateAppend({path: [...parentPath, 'error'], replace: false})
+  const replace = !action.payload.fromDeviceAdd
+  const path = ['error']
+  return [
+    ...(action.payload.fromDeviceAdd ? [RouteTreeGen.createClearModals()] : []),
+    RouteTreeGen.createNavigateAppend({path: [...parentPath, ...path], replace}),
+  ]
 }
 
 const showUsernameEmailPage = () => RouteTreeGen.createNavigateAppend({path: ['username']})
--- a/shared/actions/provision.tsx
+++ b/shared/actions/provision.tsx
@@ -371,34 +371,6 @@ class ProvisioningManager {
 
     this._done = true
     this._canceled = true
-
-    // TODO fix
-    // let root = state.routeTree.routeState && state.routeTree.routeState.selected
-    // let onDevicesTab = root === devicesRoot[0]
-    // let onLoginTab = root === Tabs.loginTab
-    // const path = Router2Constants.getFullRoute().map(p => p.routeName)
-    // onDevicesTab = path.includes(devicesRoot[0])
-    // onLoginTab = path.includes('login')
-    // const doingDeviceAdd = this._addingANewDevice && onDevicesTab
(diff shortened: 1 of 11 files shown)


zapu:
For easier testing:
```
diff --git a/go/engine/kex2_provisioner.go b/go/engine/kex2_provisioner.go
index 298b7a9aa8..be8384d9cd 100644
--- a/go/engine/kex2_provisioner.go
+++ b/go/engine/kex2_provisioner.go
@@ -111,7 +111,8 @@ func (e *Kex2Provisioner) Run(m libkb.MetaContext) error {
                DeviceID:      deviceID,
                Secret:        e.secret,
                SecretChannel: e.secretCh,
-               Timeout:       60 * time.Minute,
+               //Timeout:       60 * time.Minute,
+               Timeout: 10 * time.Second,
        }
        parg := kex2.ProvisionerArg{
                KexBaseArg:   karg,

diff --git a/go/engine/kex2_provisionee.go b/go/engine/kex2_provisionee.go
index 488063de3c..cd13dd78d6 100644
--- a/go/engine/kex2_provisionee.go
+++ b/go/engine/kex2_provisionee.go
@@ -134,7 +134,8 @@ func (e *Kex2Provisionee) Run(m libkb.MetaContext) error {
                DeviceID:      e.device.ID,
                Secret:        e.secret,
                SecretChannel: e.secretCh,
-               Timeout:       60 * time.Minute,
+               //Timeout:       60 * time.Minute,
+               Timeout: 6 * time.Second,
        }
        parg := kex2.ProvisioneeArg{
                KexBaseArg:  karg,

```

shared/actions/provision.tsx
```diff
@@ -349,33 +352,25 @@ class ProvisioningManager {
...
+    }
+
+    if (this._stashedResponse && this._stashedResponseKey) {
+      logger.info('ProvisioningManager - canceling ongoing stashed response')
+      Constants.cancelOnCallback(null, this._stashedResponse)
+      this._stashedResponse = null
+      this._stashedResponseKey = null
+    }
+
+    this._done = true
```
buoyad: I think you can skip setting _done and let the `manager.done` call below set it. Then this can be `canceled` while we wait for the RPC to return, and `done` after.

shared/actions/provision.tsx
```diff
@@ -409,9 +404,21 @@ function* startProvisioning(state: Container.TypedState) {
...
+      )
+      return
+    }
+
+    if (Constants.errorCausedByUsCanceling(finalError) && manager.isCanceled()) {
+      // After cancelling the RPC we are going to get "input canceled" error.
+      return
+    }
+
+    manager.done('provision call done w/ error' + finalError ? finalError.message : ' unknown error')
```
zapu: below i.e. here? if I move this back to right before the `if (Constants.errorCausedByUsCanceling(finalError) && manager.isCanceled()) {` block for example?

buoyad: Oh i see, else we risk never setting it to done. I guess it should go first in the catch?

zapu: probably even before the singleton check, yes

zapu:
@buoyad done

zapu:
or should I say: `done()`

PR #19689, opened 2019-09-16
dark mode people/ and profile/

songgao:


shared/wallets/wallet-list/wallet-row/index.tsx
```diff
@@ -76,9 +76,9 @@ const styles = Styles.styleSheetCreate(
 
 const HoverBox = Styles.isMobile
   ? Kb.Box2
-  : Styles.styled(Kb.Box2)({
+  : Styles.styled(Kb.Box2)(() => ({
       ':hover': {backgroundColor: Styles.globalColors.blueGreyDark},
```
chrisnojima: we should ticket to remove a bunch of these. 99% of all hover things can just use css instead

songgao: OK; made https://keybase.atlassian.net/browse/HOTPOT-835

PR #19691, opened 2019-09-16
Remove old password screen impl

buoyad:
cc @keybase/y2ksquad


buoyad:
@keybase/react-hackers bump

PR #19697, opened 2019-09-16
Update delete modal.

adamjspooner:
@keybase/react-hackers, this is ready for review. I'll add it to mobile when we update the password check.


shared/settings/delete-confirm/index.tsx
```diff
@@ -0,0 +1,114 @@
+import * as React from 'react'
+import * as Styles from '../../styles'
+import * as Kb from '../../common-adapters'
+
+type CheckboxesProps = {
+  box1: boolean
```
chrisnojima: see this feedback: https://github.com/keybase/client/pull/19791#pullrequestreview-290642473

adamjspooner: @chrisnojima, mind taking another look?

PR #19706, opened 2019-09-16
rate limit gregor push state calls

mmaxim:


go/service/gregor.go
```diff
@@ -1298,6 +1318,7 @@ func (g *gregorHandler) Reset() error {
 	g.Shutdown()
 	g.setFirstConnect(true)
 	g.shutdownGregorClient(context.TODO())
+	g.pushStateCh <- nil // clear any pending push state calls
```
joshblum: close on shutdown?

go/service/gregor.go
```diff
@@ -521,21 +525,37 @@ func (g *gregorHandler) iterateOverFirehoseHandlers(f func(h libkb.GregorFirehos
 	g.firehoseHandlers = freshHandlers
 }
 
+func (g *gregorHandler) pushStateThread() {
+	var lastPush func()
+	var lastTime time.Time
+	for push := range g.pushStateCh {
+		lastPush = push
+		if time.Since(lastTime) > time.Second {
```
joshblum: if this is false we just lose the update, seems wrong

mmaxim: I believe this is handled in my latest version

go/service/gregor.go
mmaxim: I'm not sure how to get it going again given current construction.

mmaxim: Nvm, made a change

go/service/gregor.go
```diff
@@ -524,20 +524,20 @@ func (g *gregorHandler) iterateOverFirehoseHandlers(f func(h libkb.GregorFirehos
 }
 
 func (g *gregorHandler) pushStateThread(shutdownCh chan struct{}) {
```
joshblum: maybe rename to `debounceNewData`

go/service/gregor.go
```diff
@@ -549,22 +549,29 @@ func (g *gregorHandler) pushStateThread(shutdownCh chan struct{}) {
...
+	// Only send this state update on reception of new data, not a reconnect since we will
+	// be sending that on a different code path altogether (see OnConnect).
+	if g.badger != nil && r != keybase1.PushReason_RECONNECTED {
+		g.badger.PushState(context.Background(), s)
+	}
+}
+
+func (g *gregorHandler) pushState(r keybase1.PushReason) {
+	switch r {
+	case keybase1.PushReason_RECONNECTED, keybase1.PushReason_NONE:
```
joshblum: when does `PushReason_NONE` happen? does it ever? why not debounce?

go/service/gregor.go
```diff
@@ -521,7 +523,33 @@ func (g *gregorHandler) iterateOverFirehoseHandlers(f func(h libkb.GregorFirehos
...
+	for {
+		select {
+		case <-g.pushStateCh:
+			shouldSend = true
+			if time.Since(lastTime) > dur {
+				trigger()
+			}
+		case <-time.After(dur):
+			trigger()
+		case <-shutdownCh:
```
joshblum: this never gets reinitialized on reconnect

go/service/gregor.go
mmaxim: Unused, and I have `NEW_DATA` above as the reason, didn't want to muddle it.

go/service/gregor.go
mmaxim: It does in `Connect`

joshblum: oops missed this, lgtm!

PR #19709, opened 2019-09-16
find dom node 2

chrisnojima:
This attempts to skip findDomNode when dealing with html elements and will console.log if we pass a component. Longer term we can eliminate this (there is a plan to change how the floating box works so that'll likely entirely clean this up).
I also made `Box2` and `Button` function components w/ `forwardRef` so we can reach into them and get the divs


chrisnojima:
@keybase/react-hackers bump open 8 days

PR #19710, opened 2019-09-16
Autoreset storybook: do you know your password / enter password

buoyad:
cc @keybase/y2ksquad 

- [ ] Storyshots before merge

### DIFF
--- a/shared/common-adapters/labeled-input.tsx
+++ b/shared/common-adapters/labeled-input.tsx
@@ -94,6 +94,7 @@ const ReflessLabeledInput = (props: Props & RefProps) => {
         ref={props.forwardedRef}
         style={Styles.collapseStyles([
           styles.input,
+          props.style,
           {height: computedHeight, maxHeight: computedHeight},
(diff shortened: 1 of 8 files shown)


shared/common-adapters/labeled-input.tsx
```diff
@@ -94,6 +94,7 @@ const ReflessLabeledInput = (props: Props & RefProps) => {
         ref={props.forwardedRef}
         style={Styles.collapseStyles([
           styles.input,
+          props.style,
```
buoyad: This was not being plumbed through erroneously, I added it and removed the instances of this prop we had before

shared/login/relogin/index.desktop.tsx
```diff
@@ -160,19 +159,6 @@ const styles = Styles.styleSheetCreate(() => ({
     justifyContent: 'flex-end',
   },
   other: {color: Styles.globalColors.black},
-  passwordInput: {
```
jakob223: amaze

PR #19724, opened 2019-09-17
autoresets: screens 4/5/6

jakob223:
cc @keybase/y2ksquad 

@buoyad sorry, I did stories slightly differently than you so we will have to do a merge but shouldn't be too painful.

image
image
image


shared/login/reset/waiting.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </Kb.Box2>
+    </Kb.Box2>
+  </SignupScreen>
+)
+export default Waiting
+
+const styles = Styles.styleSheetCreate(() => ({
+  mainText: {
+    ...Styles.padding(0, Styles.globalMargins.xsmall),
+    textAlign: 'center',
```
buoyad: there's a `center` prop on Text for this

PR #19725, opened 2019-09-17
makeReducer

chrisnojima:
Based off of redux-start kit's version but for our types: https://github.com/reduxjs/redux-starter-kit/blob/master/src/createReducer.ts

### DIFF
--- a/shared/desktop/yarn-helper/action-creator-creator.tsx
+++ b/shared/desktop/yarn-helper/action-creator-creator.tsx
@@ -185,7 +185,8 @@ function makeTypedActions(created) {
 
   export type TypedActions = ${created.map(c => `${cleanName(c)}.Actions`).join(' | ')}
 
-  export type TypedActionsMap = {${typeMap.join(',\n')}
+  export type TypedActionsMap = {${typeMap.join(',\n')},
+    'common:resetStore': {type: 'common:resetStore', payload: {}}
   }
 `
 }
--- a/shared/reducers/git.tsx
+++ b/shared/reducers/git.tsx
@@ -7,41 +7,31 @@ const initialState: Types.State = {
   isNew: new Set(),
 }
 
-export default (state: Types.State = initialState, action: GitGen.Actions): Types.State =>
-  Container.produce(state, (draftState: Container.Draft<Types.State>) => {
-    switch (action.type) {
-      case GitGen.resetStore:
-        return initialState
-      case GitGen.loaded:
-        draftState.idToInfo = action.payload.repos
-        return
-      case GitGen.setError:
-        draftState.error = action.payload.error
-        return
-      case GitGen.badgeAppForGit:
-        // We show our badges until we clear with the clearBadges call. If there are no badges we likely cleared it from the nav ourselves
-        if (action.payload.ids.size === 0) {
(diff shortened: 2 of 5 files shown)


MarcoPolo:
What's the benefit of this style?

MarcoPolo:
What are the tradeoffs?

shared/desktop/yarn-helper/action-creator-creator.tsx
```diff
@@ -185,7 +185,8 @@ function makeTypedActions(created) {
 
   export type TypedActions = ${created.map(c => `${cleanName(c)}.Actions`).join(' | ')}
 
-  export type TypedActionsMap = {${typeMap.join(',\n')}
+  export type TypedActionsMap = {${typeMap.join(',\n')},
+    'common:resetStore': {type: 'common:resetStore', payload: {}}
```
cjb: What's special about common:resetStore that makes this required?

shared/reducers/git.tsx
```diff
@@ -7,41 +7,31 @@ const initialState: Types.State = {
...
-      case GitGen.repoDeleted:
-      case GitGen.setTeamRepoSettings:
-        return
+const clearErrors = (draftState: Container.Draft<Types.State>) => {
+  draftState.error = undefined
+}
+
+export default Container.makeReducer<GitGen.Actions, Types.State>(initialState, {
+  [GitGen.resetStore]: () => initialState,
+  [GitGen.loaded]: (draftState, action) => {
```
cjb: I guess you can ditch the braces here since it's a single statement? That's pleasingly compact.

shared/util/make-reducer.tsx
```diff
@@ -0,0 +1,22 @@
...
+function makeReducer<A, S>(
+  initialState: S,
+  map: {
+    [type in GetTypes<A>]?: type extends keyof TypedActionsMap
+      ? ((state: Draft<S>, action: TypedActionsMap[type]) => void | S)
+      : never
+  }
+) {
+  return (state: S = initialState, action: TypedActions): S =>
+    // @ts-ignore
```
cjb: Does this ignore affect typing of the reducers that use it?  Why's it here?

shared/reducers/git.tsx
```diff
@@ -7,41 +7,31 @@ const initialState: Types.State = {
...
+    if (action.payload.ids.size !== 0) {
+      draftState.isNew = action.payload.ids
     }
-  })
+  },
+  [GitGen.clearBadges]: draftState => {
+    draftState.isNew = initialState.isNew
+  },
+  // Clear errors
+  [GitGen.createPersonalRepo]: clearErrors,
```
cjb: The switch statement did at least make it easy to apply one transform to many cases without repetition.  Maybe there's a way to do that here too?

shared/util/make-reducer.tsx
chrisnojima: We say what the type is explicitly so it doesn't matter. See the original implementation for context

shared/reducers/git.tsx
chrisnojima: In practice it doesn't happen much

PR #19740, opened 2019-09-17
Tweaks on wallet modal and provisioning texts

cecileboucheron:
@keybase/react-hackers


cecileboucheron:
@MarcoPolo Can I get another look? It's all good now on iOS.

PR #19741, opened 2019-09-17
fix ash background transparency

hyperobject:
ashbox should have white background, not fastblank

@keybase/react-hackers


hyperobject:
> Fast blank needs to be black in dark mode on ios

this should be handled by the dark mode logic?

PR #19748, opened 2019-09-17
Fix LogoutEngine return val when user switching fails

zapu:
At some point we might consider this error critical - probably when User Switching feature is more mature or completed. But for now it doesn't make sense to error out here, especially that it black bars in GUI, but does so after the logout. It's too late at that point anyway - the user is logged out, secrets are killed.


maxtaco:
:+1: thanks for finding this @zapu; cc @strib i think this was a linting bug

PR #19749, opened 2019-09-17
Fix android status bar color. It is dark mode

MarcoPolo:
Fixes color of status bar icons on android


chrisnojima:
Please pick into release

PR #19750, opened 2019-09-17
dark mode teams

songgao:


shared/teams/delete-team/index.tsx
```diff
@@ -12,7 +12,7 @@ export type Props = {
 const Header = (props: Props) => (
   <>
     <Kb.Avatar teamname={props.teamname} size={64} />
-    <Kb.Icon type="icon-team-delete-28" style={{marginRight: -60, marginTop: -20}} />
+    <Kb.Icon type="icon-team-delete-28" style={{marginRight: -60, marginTop: -20, zIndex: 1}} />
```
songgao: I noticed this in storybook, so just did the simplest thing following `really-leave` modal here https://github.com/keybase/client/blob/17cc9dda1052079051a51083d7ed2317009b6edb/shared/teams/really-leave-team/index.tsx#L35

shared/teams/team/members-tab/member-row/index.tsx
```diff
@@ -103,19 +103,19 @@ export const TeamMemberRow = (props: Props) => {
...
                 <Text
                   type="BodySmall"
                   style={{
-                    ...globalStyles.fontBold,
-                    backgroundColor: globalColors.red,
-                    color: globalColors.white,
-                    marginRight: globalMargins.xtiny,
-                    paddingLeft: globalMargins.xtiny,
-                    paddingRight: globalMargins.xtiny,
+                    ...Styles.globalStyles.fontBold,
```
buoyad: I think we should move these objects into `styles`

shared/teams/team/members-tab/member-row/index.tsx
```diff
@@ -73,28 +73,28 @@ export const TeamMemberRow = (props: Props) => {
   }
 
   return (
-    <Box style={active ? stylesContainer : stylesContainerReset}>
+    <Box style={Styles.collapseStyles([styles.container, active && styles.containerReset])}>
```
buoyad:
```suggestion
    <Box style={active ? styles.container : styles.containerReset}>
```

chrisnojima: note these are not equivalent. can only be if containerReset is a super set in the style but not in general

buoyad: It is the ternary in master, the style on the rows is weird with `collapseStyles` (they're blue)

chrisnojima: k. i think the ternary is better, but it could also be `[!active && styles.container, active && styles.containerReset]` etc if mixing a bunch of concepts

songgao: Note that `containerReset` isn't a superset of `container` here anymore. I changed it so it only includes the difference now.

shared/teams/team/members-tab/member-row/index.tsx
songgao: moved this one but just realized there are more; please hold off review for a sec;

songgao: ok ready again!

shared/teams/team/members-tab/member-row/index.tsx
buoyad: It should be `!active && styles.containerReset`

songgao: oops; good catch!!

PR #19755, opened 2019-09-18
immer chat 4

chrisnojima:
- [x] unsent text
- [x] unfurl url map

### DIFF
--- a/shared/chat/conversation/messages/wrapper/unfurl/prompt-list/container.tsx
+++ b/shared/chat/conversation/messages/wrapper/unfurl/prompt-list/container.tsx
@@ -3,7 +3,6 @@ import * as Chat2Gen from '../../../../../../actions/chat2-gen'
 import * as Types from '../../../../../../constants/types/chat2'
 import * as RPCChatTypes from '../../../../../../constants/types/rpc-chat-gen'
 import {namedConnect} from '../../../../../../util/container'
-import {get} from 'lodash-es'
 import UnfurlPromptList from '.'
 
 type OwnProps = {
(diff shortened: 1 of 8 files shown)


shared/chat/conversation/messages/wrapper/unfurl/prompt-list/container.tsx
```diff
@@ -11,74 +10,56 @@ type OwnProps = {
...
+      promptDomains = pm.get(messageID)
+    }
+    return {
+      messageID,
+      promptDomains: promptDomains || noPrompts,
+    }
   },
-})
-
-const makeRes = (actionType: RPCChatTypes.UnfurlPromptAction, domain?: string) => {
```
chrisnojima: this wasn't typed correctly. the `onetime`/`accept` fields are only acceptable on those `actionType`s

shared/reducers/chat2.tsx
```diff
@@ -499,11 +499,18 @@ export default (_state: Types.State = initialState, action: Actions): Types.Stat
...
+        const mmap = new Map(unfurlPromptMap.get(conversationIDKey) || [])
+        const prompts = new Set(mmap.get(messageID) || [])
+        if (show) {
+          prompts.add(domain)
+        } else {
+          prompts.delete(domain)
+        }
+        mmap.set(messageID, prompts)
+        unfurlPromptMap.set(conversationIDKey, mmap)
+        draftState.unfurlPromptMap = unfurlPromptMap
```
buoyad: These are tricky. I wonder if some helper like `updateMap` could help. Maybe like
```
draftState.unfurlPromptMap = updateMap(draftState.unfurlPromptMap, (map) => {
  const mmap = updateMap(unfurlPromptMap.get(conversationIDKey) || [], (mmap) =>  {
    const prompts = new Set(mmap.get(messageID) || [])
    if (show) {
      prompts.add(domain)
    } else {
      prompts.delete(domain)
    }
    return mmap.set(messageID, prompts)
  })
  return map.set(conversationIDKey, mmap)
})
```
Where `updateMap` calls `new Map` on the first param. It's not much simpler code-wise but I think mentally it's easier to reach for `updateMap` every time you change a map in a reducer. Anyway, just an idea.

chrisnojima: i think its hard to get this to type correctly. hopefully a bunch of this goes away with 
`?.` and Map support in the next immer

chrisnojima: its also likely simpler to just not have all these nested things at all and just flatten it out. just make the keys `convoid + : + messageid`

buoyad: Yeah agreed we should try to avoid nesting these

PR #19759, opened 2019-09-18
onSwipeLeft is optional

chrisnojima:


chrisnojima:
i agree. this shouldn't work this way at all

PR #19762, opened 2019-09-18
Check TLFName/TopicName when formatting conversation name [HOTPOT-840]

joshblum:
don't show `#` if `TopicName`/`TlfName` are empty


mmaxim:
How does this happen?

joshblum:
not sure, will look at the log. i suspect notify

joshblum:
(ci flake unrelated)

PR #19767, opened 2019-09-18
don't show two errors when one will do

xgess:
i couldn't actually recreate the condition where it's possible to get both of these error messages to display at the same time, but cecile saw it. perhaps the path creation logic in horizon has gotten smarter. regardless, there's no reason to show both of these error messages, because it's a little weird. so now just show one or the other.


xgess:
@keybase/react-hackers

PR #19775, opened 2019-09-18
reset: hook up desktop modal

jakob223:
Issue: Y2K-160

cc @keybase/y2ksquad 

To test, 
```keybase account reset-start YOUR_USERNAME```
and then open the app. Click cancel and see it work.

### CODE REVIEW

File: shared/login/reset/modal.tsx [hunk 1]
```diff
@@ -1,15 +1,27 @@
...
+import * as AutoresetGen from '../../actions/autoreset-gen'
 
-export type Props = {
-  mapURL: string
-  timeLeft: string
-  onCancelReset: () => void
-}
+export type Props = {}
+
+const ResetModal = (_: Props) => {
```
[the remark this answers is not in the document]
jakob223: how do you set that on a functional component?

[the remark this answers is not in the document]
chrisnojima: `ResetModal.navigationOptions = {`

shared/util/timestamp.tsx
```diff
@@ -103,6 +103,31 @@ export const formatDuration = (duration: number): string => {
...
+    amt = d.hours()
+    label = 'hour'
+  } else if (d.minutes()) {
+    amt = d.minutes()
+    label = 'minute'
+  } else {
+    amt = d.seconds()
+    label = 'second'
+  }
+  if (amt > 1) {
```
buoyad: I think better as `amt !== 1`. or you could use `pluralize` from `util/string`

shared/actions/autoreset.tsx
```diff
@@ -1,3 +1,35 @@
-function* autoresetSaga() {}
+import * as Saga from '../util/saga'
+import * as Container from '../util/container'
+
+import * as NotificationsGen from './notifications-gen'
+import * as RouteTreeGen from './route-tree-gen'
+import * as AutoresetGen from './autoreset-gen'
+import * as RPCGen from '../constants/types/rpc-gen'
+
```
buoyad: :fire: blank lines

shared/actions/autoreset.tsx
```diff
@@ -1,3 +1,35 @@
...
+import * as Constants from '../constants/autoreset'
+
+const receivedBadgeState = async (
+  state: Container.TypedState,
+  action: NotificationsGen.ReceivedBadgeStatePayload
+) => {
+  const newResetState = action.payload.badgeState.resetState
+  if (state.autoreset !== newResetState) {
```
buoyad: This will fail every time, `newResetState` is a new object. I'd say compare the fields individually.

shared/actions/autoreset.tsx
```diff
@@ -1,3 +1,35 @@
...
+    return [
+      ...(newResetState.active ? [RouteTreeGen.createNavigateAppend({path: ['resetModal']})] : []),
```
buoyad: Hmm. Now it seems like this is firing too early to happen, like before the initial routing has happened. On login / restarting the app the nav doesn't happen, but switching tabs works.

shared/actions/autoreset.tsx
jakob223: I like them to organize the imports, but I guess we don't use them anywhere else so I can delete them

jakob223:
bump @buoyad

jakob223:
CI fail was on a thing that master was previously failing on. Merged master and merging.

PR #19791, opened 2019-09-19
Autoreset storybook: confirm reset

buoyad:
image

This also adds a `fullscreen` prop to modal that changes it on desktop to the version seen here.

cc @keybase/y2ksquad also cc @mlsteele I heard you're looking for a sticky modal footer tray like this one.

### DIFF
--- a/shared/login/reset/confirm.tsx
+++ b/shared/login/reset/confirm.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react'
 import * as Kb from '../../common-adapters'
 import * as Styles from '../../styles'
-import {SignupScreen} from '../../signup/common'
 
 const todo = () => console.log('todo')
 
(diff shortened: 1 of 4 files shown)


buoyad:
Added checkbox for stellar wallets to show up only if you've accepted the disclaimer.
image

shared/login/reset/confirm.tsx
```diff
@@ -0,0 +1,118 @@
...
+import * as Styles from '../../styles'
+
+const todo = () => console.log('todo')
+
+type Props = {
+  hasWallet: boolean
+}
+
+const ConfirmReset = (props: Props) => {
+  const [check1, setCheck1] = React.useState(false)
```
chrisnojima: i think this is too disconnected from the content. 
This should either be a map like `{wallet: false, ...}` or at least `checkWallet` etc

PR #19793, opened 2019-09-19
Remove userregion type from the go code

heronhaye:


heronhaye:
chat is flake and autoreset is fixed in another pr

PR #19795, opened 2019-09-19
remove nulls in variant types

chrisnojima:
The avdl variant code gen would make types like
`{type: 'a', somethingA: AThing | null} | {type: 'b', somethingB: BThing | null}`
 when really it should be
`{type: 'a', somethingA: AThing} | {type: 'b', somethingB: BThing}`

This pr fixes that and also removes some now unnecessary checks


chrisnojima:
cc: @malgorithms

PR #19797, opened 2019-09-19
Safe navigation in chat inbox

buoyad:
Fixes the issue where you quick-tap two rows on mobile and get two convos pushed. cc @keybase/y2ksquad

### DIFF
--- a/shared/router-v2/util.tsx
+++ b/shared/router-v2/util.tsx
@@ -1,5 +1,4 @@
 import {NavState} from '../constants/types/route-tree'
-import {memoize} from '../util/memoize'
 // We could have subnavigators, so traverse the routes so we can get the active
 // screen's index so we know when to enable the back button. Note this doesn't
 // support a subnavigator with a root you can hit back from.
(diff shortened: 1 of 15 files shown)


buoyad:
I'm moving where the `navKey` comes from out of `MakeRow` -> `Inbox`

buoyad:
k @chrisnojima `makeRow` is a function again, inbox is grabbing the key for all the rows.

shared/router-v2/util.tsx
```diff
@@ -13,13 +14,15 @@ export const getActiveIndex = (navState: NavState): number => {
   return navState.index
 }
 // Get active key inside any subnavigator so navigation closures are
-export const getActiveKey = (navState: NavState): string => {
-  if (!navState.routes) {
-    return navState.key
+export const getActiveKey = memoize(
```
chrisnojima: does this memoizing work at all. this thing is called recursively so you're blowing it away right

buoyad: Good catch, fixed.

shared/router-v2/util.tsx
```diff
@@ -1,4 +1,5 @@
 import {NavState} from '../constants/types/route-tree'
+import {memoize} from '../util/memoize'
```
chrisnojima: 🔥

PR #19800, opened 2019-09-19
first pass put-get

xgess:
adds commands for GETting and PUTting into the team storage
relies on https://github.com/keybase/keybase/pull/4408

```
$ keybase kvstore api -m '{"method": "put", "params": {"options": {"team": "nessy32,nessy32", "namespace": "personal-apps", "entryKey": "geocities2", "entryValue": "my geocities secretz"}}}'

{"result":{"teamName":"nessy32,nessy32","namespace":"personal-apps","entryKey":"geocities2","revision":1}}
```
```
$ keybase kvstore api -m '{"method": "get", "params": {"options": {"team": "nessy32,nessy32", "namespace": "personal-apps", "entryKey": "geocities2"}}}'

{"result":{"teamName":"nessy32,nessy32","namespace":"personal-apps","entryKey":"geocities2","entryValue":"my geocities secretz","revision":1}}
```

basic idea here is that there's new commands to the CLI for kvstore get and put. `go/service/kvstore.go` is where the main logic is. it calls into a revision cache (which hangs off G) to verify that the server can't lie, and a boxer to delegate encryption/signing and decryption/verifying. tests are pretty behavioral, i think, so it might make sense to start there. 

i was initially hoping this could be reviewed commit-by-commit. i'm not 100% sure anymore that it can. maybe. maybe not.


xgess:
TODO: running commands against a team that you aren't in crashes the service. so i'll fix that and remove this comment.

go/kvstore/boxer.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/go-crypto/ed25519"
+	"golang.org/x/crypto/nacl/secretbox"
+)
+
+type KVStoreBoxer interface {
+	Box(mctx libkb.MetaContext, entryID keybase1.KVEntryID, revision int, cleartextValue string) (ciphertext string,
+		teamKeyGen keybase1.PerTeamKeyGeneration, ciphertextVersion int, err error)
+	Unbox(mctx libkb.MetaContext, entryID keybase1.KVEntryID, revision int, ciphertext string, formatVersion int,
+		senderUID keybase1.UID, senderEldestSeqno keybase1.Seqno, senderDeviceID keybase1.DeviceID) (cleartext string, err error)
```
mlsteele: I'd like to get max involved with reviewing the crypto. It could be that [signencrypt](https://github.com/keybase/client/blob/master/go/chat/signencrypt/codec.go#L229) is a tool for this job

go/kvstore/boxer.go
```diff
@@ -0,0 +1,277 @@
...
+		Version: 2,
+	}
+	_, err = sigInfo.Verify()
+	return err
+}
+
+func (b *KVStoreRealBoxer) buildSignatureMsg(elements signatureElements) (ret []byte, err error) {
+	ret = append(ret, []byte(elements.EntryID.TeamID)...)
+	ret = append(ret, []byte(elements.EntryID.Namespace)...)
+	ret = append(ret, []byte(elements.EntryID.EntryKey)...)
```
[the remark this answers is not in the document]
xgess: lame attack, but yes. i agree. you are correct. will change. :)

go/client/cmd_kvstore.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/keybase/client/go/libkb"
+)
+
+func NewCmdKVStore(cl *libcmdline.CommandLine, g *libkb.GlobalContext) cli.Command {
+	subcommands := []cli.Command{
+		newCmdKVStoreAPI(cl, g),
+	}
+	return cli.Command{
+		Name: "kvstore",
+		// Usage:        "Manage a simple cleartext key to encrypted value store", //uncomment to unhide from command list
```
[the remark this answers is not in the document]
xgess: this looks like it's only for boolean flags on commands, not for whole commands

go/client/kvstore_api_handler.go
```diff
@@ -0,0 +1,147 @@
...
+	default:
+		return ErrInvalidMethod{name: c.Method, version: 1}
+	}
+}
+
+type getEntryOptions struct {
+	Team      string `json:"team"`
+	Namespace string `json:"namespace"`
+	EntryKey  string `json:"entryKey"`
+}
```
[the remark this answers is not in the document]
xgess: nah. i can push it down there.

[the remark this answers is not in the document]
xgess: wait a tick. it is there. this comment is on a different object, `getEntryOptions` which is basically just `keybase1.GetKVEntryArg` but without the sessionID. I could just reuse that object, but it's got a session floating around that we basically don't care about.

go/service/kvstore.go
```diff
@@ -0,0 +1,239 @@
...
+		TeamID:    teamID,
+		Namespace: arg.Namespace,
+		EntryKey:  arg.EntryKey,
+	}
+	prevRevision, err := h.fetchRevisionFromCacheOrServer(mctx, entryID)
+	if err != nil {
+		mctx.Debug("error fetching the revision for %+v from the local cache: %v", entryID, err)
+		return res, err
+	}
+	revision := prevRevision + 1
```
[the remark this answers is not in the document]
xgess: correct. i had initially planned on adding revision and conflict management to this PR, but i realized it wasn't strictly necessary for a very first pass, and it would be fairly easy to review in a single future PR. but yes. this is the idea. 
device `a` will pass in the revision with the second write and get a very helpful error message. and we can decide what to do in the case you've described; my instinct is overwrite if no revision is passed, and error with something very helpful (i.e. this entry changed since you've last seen it. you should fetch it first and take a look) if you used an old one. that way two devices can stay in sync either by both using revisions or both not using them.

go/kvstore/boxer.go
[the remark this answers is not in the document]
xgess: done

go/kvstore/cache.go
```diff
@@ -0,0 +1,114 @@
...
+	Revision   int
+	EntryHash  string
+	TeamKeyGen keybase1.PerTeamKeyGeneration
+}
+
+type kvCacheData map[keybase1.TeamID]map[string] /*namespace*/ map[string] /*entry*/ kvCacheEntry
+
+type KVRevisionCache struct {
+	sync.Mutex
+	data kvCacheData
```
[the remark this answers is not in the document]
xgess: done. PICNIC-533

go/service/kvstore.go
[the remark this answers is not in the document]
xgess: ticket so this doesn't get missed: PICNIC-534

go/libkb/globals.go
```diff
@@ -99,6 +99,7 @@ type GlobalContext struct {
 	paramProofStore        MerkleStore        // a cache and fetcher for param proofs
 	externalURLStore       MerkleStore        // a cache and fetcher for external urls
 	PayloadCache           *PayloadCache      // cache of ChainLink payload json wrappers
+	kvRevisionCache        KVRevisionCacher   // cache of revisions for verifying key-value store results
```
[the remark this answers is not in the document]
xgess: 👍 done

go/kvstore/boxer.go
```diff
@@ -31,74 +31,26 @@ func NewKVStoreBoxer(g *libkb.GlobalContext) *KVStoreRealBoxer {
 	}
 }
 
-type signatureElements struct {
-	EntryID           keybase1.KVEntryID
-	ClearBytes        []byte
-	Revision          int
-	Nonce             [24]byte
```
xgess: i removed the nonce from the associated data object. i think that's ok to do, but tell me if you think otherwise.

go/chat/signencrypt/codec.go
```diff
@@ -565,6 +566,59 @@ func OpenWhole(sealed []byte, encKey SecretboxKey, verifyKey VerifyKey, signatur
...
 }
 
+// include verification of associated data. see
+// https://en.wikipedia.org/wiki/Authenticated_encryption#Authenticated_encryption_with_associated_data_(AEAD)
+
+type AEADMessage struct {
+	V int               `codec:"v" json:"v"`
+	A [sha512.Size]byte `codec:"a" json:"a"`
+	M []byte            `codec:"m" json:"m"`
+}
```
AMarcedone: Add a comment to A saying that it is the hash of some associated data? In general names could be more verbose (since you have codec directives for packing).

go/chat/signencrypt/codec.go
```diff
@@ -565,6 +566,59 @@ func OpenWhole(sealed []byte, encKey SecretboxKey, verifyKey VerifyKey, signatur
...
+	if err != nil {
+		return ret, err
+	}
+	adHash := sha512.Sum512(adEncoded)
+	aeadMsg := AEADMessage{V: 1, A: adHash, M: msg}
+	clearBytes, err := msgpack.Encode(aeadMsg)
+	if err != nil {
+		return ret, err
+	}
+	return SealWhole(clearBytes, encKey, signKey, signaturePrefix, nonce), nil
```
AMarcedone: One idea might be have the output be a struct that itself contains the associated data. So that you have a single object having both the associated data and the encrypted one.
This is probably good enough though, as having the data part of the ciphertext could interfere with streaming decryption.

go/chat/signencrypt/codec.go
xgess: was thinking that was obvious from `[sha512.Size]` but you're right. just a second...

go/chat/signencrypt/codec_test.go
```diff
@@ -23,6 +23,24 @@ var plaintextInputs = []string{
...
+type arbitraryMsg struct {
+	Message string `codec:"m" json:"m"`
+}
+type arbitraryNested struct {
+	Fun    arbitraryMsg `codec:"f" json:"f"`
+	Boring arbitraryMsg `codec:"b" json:"b"`
+}
+
+var associatedDataInputs = []interface{}{
+	"",
```
AMarcedone: I am curious what happens if you try to use nil as associated data. Probably the same as "", but it would be interesting to know and maybe disallow one of the two.

go/kvstore/boxer.go
AMarcedone: it's ok. I would like to see some testing around the nonce though. For example:
- generate two nonces and assert they are different.
- encrypt something and assert the nonce field was actually set to something non zero
- encrypt something with a good nonce and try to decrypt with a different nonce.

go/chat/signencrypt/codec_test.go
xgess: it works just fine. as long as msgpack encodes it consistently, it really shouldn't matter what you pass in.

go/kvstore/boxer.go
xgess: i can add a couple nonce-related tests.

go/chat/signencrypt/codec.go
xgess: yeah - seems like a lateral move. i'd rather not do this.

go/chat/signencrypt/codec_test.go
xgess: added a test for it so the behavior is explicit.

go/kvstore/boxer.go
xgess: done!

go/kvstore/boxer.go
xgess: done.

AMarcedone:
OK, LGTM for the crypto design :)

go/chat/signencrypt/codec.go
```diff
@@ -565,6 +566,63 @@ func OpenWhole(sealed []byte, encKey SecretboxKey, verifyKey VerifyKey, signatur
...
+	Version       int               `codec:"v" json:"v"`
+	AssocDataHash [sha512.Size]byte `codec:"a" json:"a"`
+	Message       []byte            `codec:"m" json:"m"`
+}
+
+// SealWithAssociatedData is a wrapper around SealWhole which adds an associatedData object
+// (see AEAD ciphers) which must be message-packable into bytes. This exact object is required
+// to call OpenWithAssociatedData on the ciphertext.
+func SealWithAssociatedData(msg []byte, associatedData interface{}, encKey SecretboxKey, signKey SignKey, signaturePrefix kbcrypto.SignaturePrefix, nonce Nonce) (ret []byte, err error) {
+	adEncoded, err := msgpack.Encode(associatedData)
```
mlsteele: Chat uses `codec.MsgpackHandle{WriteExt: true}` I wonder if `WriteExt` is relevant

xgess: just dove into this. it does seem that most of the most similar binary encoding/decoding uses this pattern. good news though. msgpack.Encode and msgpack.Decode are just shorthand for exactly this. :)

PR #19802, opened 2019-09-19
PICNIC-359 - Fix Avatar Upload Panning Regression

thebearjew:
### Overview

- Moved the panning mouse handlers to the `Kb.Box` that renders the modal content rather than `Kb.MaybepPopup`.
- Made the `React.MouseEvent` persist so that we can use it when setting the offset position using `setState`. This was the main issue as React was printing console errors warning about the event being reclaimed.

### DIFF
--- a/shared/profile/edit-avatar/index.desktop.tsx
+++ b/shared/profile/edit-avatar/index.desktop.tsx
@@ -215,7 +215,6 @@ class EditAvatar extends React.Component<_Props, State> {
     if (!this.state.hasPreview || !this._image) return
 
     // Grab the values now. The event object will be nullified by the time setState is called.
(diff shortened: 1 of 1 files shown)


shared/profile/edit-avatar/index.desktop.tsx
```diff
@@ -298,8 +304,6 @@ class EditAvatar extends React.Component<_Props, State> {
             cursor: this.state.dragging ? '-webkit-grabbing' : 'default',
           },
         ])}
-        onMouseUp={this._onMouseUp}
-        onMouseDown={this._onMouseDown}
         onMouseMove={this._onMouseMove}
```
thebearjew: Added this back so that when panning within the white modal box, the panning continues into the overlay area.

PR #19806, opened 2019-09-19
fix incorrect blur call

chrisnojima:
if you are in chat, hit command+k , type a name, hit enter, you'll see a console error. This is due to an untyped ref


shared/chat/conversation-filter-input.tsx
```diff
@@ -20,9 +20,9 @@ export type Props = {
 }
 
 class ConversationFilterInput extends React.PureComponent<Props> {
-  _input: any
```
chrisnojima: this lets us do `this._input.blur()` which doesn't exist

PR #19811, opened 2019-09-19
Refactor some interfaces to play along with the tree sql storage engine

AMarcedone:
Also added a small optimization to the encoder to avoid generating a new msgpack encoding object every time we need to encode something.


go/logger/context_with_logging.go
```diff
@@ -0,0 +1,58 @@
+package logger
+
+import (
+	"testing"
+
+	"golang.org/x/net/context"
+)
+
+type ContextAndLogger struct {
```
maxtaco: is `logger.Context` ok?

go/libkb/context.go
```diff
@@ -690,3 +692,11 @@ func (m MetaContext) Keyring() (ret *SKBKeyringFile, err error) {
 	}
 	return m.ActiveDevice().Keyring(m)
 }
+
+var _ logger.CtxAndLogger = MetaContext{}
+
+func (m MetaContext) UpdateContext(c context.Context) logger.CtxAndLogger {
```
maxtaco: `MetaContext` already has something like this --- `WithContext`.  It's important that these things are always passed by value since `context.Context` always need to be passed by value.

go/logger/logger.go
```diff
@@ -12,14 +12,24 @@ type ExternalHandler interface {
...
-type Logger interface {
-	// Debug logs a message at debug level, with formatting args.
+type BaseLogger interface {
 	Debug(format string, args ...interface{})
+	Info(format string, args ...interface{})
+	Warning(format string, args ...interface{})
+	Error(format string, args ...interface{})
+}
+
+type CtxAndLogger interface {
```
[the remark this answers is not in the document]
AMarcedone: Since the logging objects we use on the client (logger.Logger) and the server (log.Logger) are different I thought using an interface with two different implementation would be a good way to go. I wrote the interface in a way that MetaContext already implements it on the client and we only need to write the server one. the stuff in context_with_logging.go is only used for testing.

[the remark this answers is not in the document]
maxtaco: But are interfaces passed by value? For instance, `context.Context` isn't an interface for this reason, I think....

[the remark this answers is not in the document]
maxtaco: But imagine this was the setup:

```go
type Context struct {
   logger logger.Logger
   ctx context.Context
}
```

then one liner:

```go
func (m MetaContext) ToLoggerContext() logger.Context {
  return logger.Context{logger : m.G().Log, ctx : m.ctx }
}```

File: go/logger/context_with_logging.go [hunk 4]
```diff
@@ -0,0 +1,58 @@
...
+	ctx context.Context
+	Logger
+}
+
+var _ CtxAndLogger = ContextAndLogger{}
```
maxtaco: The go convention is that the interface ends in `-er` and the instantiation doesn't.  It doesn't always work out. I'm still not a fan of `CtxAndLogger` versus `ContextAndLogger`. That's very unclear.

go/libkb/context.go
maxtaco: Rename to something like.... `UpdateContextToLoggerContext` ?

go/logger/context_with_logging.go
maxtaco: How about `logger.MiniContext` ?

maxtaco: Not a big fan of `ContextAndLogger` since once we add a 3rd field, we have to change everything.

go/merkletree2/encoders.go
```diff
@@ -25,33 +26,68 @@ const (
...
+	enc *codec.Encoder
+	dec *codec.Decoder
+}
 
 var _ Encoder = BlindedSHA512_256v1Encoder{}
 
-func (e BlindedSHA512_256v1Encoder) EncodeAndHashGeneric(o interface{}) (Hash, error) {
-	enc, err := msgpack.EncodeCanonical(o)
+func NewBlindedSHA512_256v1Encoder() BlindedSHA512_256v1Encoder {
+	var mh codec.MsgpackHandle
```
maxtaco: can't use the same codec Handle?

go/merkletree2/encoders.go
```diff
@@ -82,21 +118,47 @@ type SHA512_256Encoder struct{}
...
 	if err != nil {
 		return nil, err
 	}
-	return hasher.Sum(nil), nil
+	return e.HashKeyEncodedValuePairWithKeySpecificSecret(KeyEncodedValuePair{Key: kvp.Key, Value: encVal}, kss)
 }
 
-func (e SHA512_256Encoder) HashKeyValuePairWithKeySpecificSecret(kvp KeyValuePair, _ KeySpecificSecret) (Hash, error) {
-	return e.EncodeAndHashGeneric(kvp)
+func (e SHA512_256Encoder) HashKeyEncodedValuePairWithKeySpecificSecret(kevp KeyEncodedValuePair, kss KeySpecificSecret) (Hash, error) {
```
maxtaco: `kss` isn't used, is that a bug?  Should it be using hmac?

maxtaco: If this is a bug, let's make a new test to catch it.  Seems like it's not actually randomizing the tree?

go/merkletree2/encoders_test.go
```diff
@@ -38,22 +54,96 @@ func TestBlindedSHA512_256v1Encoder(t *testing.T) {
...
+	hashb1[0] = 0x01
+	hash1 := Hash(hashb1)
+
+	hashb2 := make([]byte, hashLength)
+	hashb2[0] = 0x02
+	hash2 := Hash(hashb2)
+
+	enc, err := encoder.Encode(Node{INodes: []Hash{hash1}})
+	require.NoError(t, err)
+	require.Equal(t, []byte{0x1, 0x91, 0xc4, 0x20, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, enc)
```
maxtaco: In general, hardcoded values aren't great since it's unclear how to update the tests if the code changes.

go/merkletree2/encoders.go
AMarcedone: It's not a bug. This encoder does not perform any blinding by design (and is only used in tests) so it ignores kss. The blind tree will use the BlindedSHA512_256v1Encoder 
(which has a test ensuring that all values affect the output, in particular see encoders_test:71)

I agree that the silently ignoring blinding is misleading though. Maybe I should throw/panic if a non nil kss is being given as input to this encoder?

go/merkletree2/encoders_test.go
AMarcedone: So do you suggest I remove the test? I think it is useful. I made another test above which does not hardcode values at all to catch logic errors, but this seems the only way to prevent a change in the msgpack encoding format (or in the library) from breaking proofs.
If an actual change is needed, one can always run this test, see where it fails and manually update the expected values to match the ones returned by the failing test (which is how I generated it).

go/merkletree2/encoders.go
maxtaco: sure

go/merkletree2/encoders_test.go
maxtaco: I think we should get this PR in, but it's a comment for the future. I've found myself in a position before when having to test against hardcoded vectors and not understanding what to update.

go/merkletree2/position.go
```diff
@@ -127,6 +141,33 @@ func (t *Config) getDeepestPositionForKey(k Key) (*Position, error) {
...
+	return min.Bytes()[1:]
+}
+
+func (t *Config) GetKeyIntervalUnderPosition(p *Position) (minKey, maxKey Key) {
+	var min, max big.Int
+
+	min.Set((*big.Int)(p))
+	n := uint(t.KeysByteLength*8 + 1 - min.BitLen())
+	min.Lsh(&min, n)
+	minKey = min.Bytes()[1:]
```
heronhaye: this is same as getMinKey?

go/logger/context_with_logging.go
maxtaco: I still want a better name for this ^

AMarcedone: I'll go for logger.Context. It is weird to me to use MiniContext for something that is bigger than a context.

go/libkb/context.go
AMarcedone: I think we can just use the WithContext method in this case. I do not need to return the interface. I'll delete this for now and we can revisit once we write the client code.

go/merkletree2/encoders.go
AMarcedone: I am not sure if that object is safe for concurrent use and if each encoder needs its own copy of the handle, so I decided to err on the side of caution. If you are convinced this is not the case, I can let both the encoder and decoder share the same object, or even declare it as a package level (constant) variable

go/merkletree2/position.go
AMarcedone: It is, but I am reusing the min object to compute the max, which is why I do not call the getMin function directly.

AMarcedone:
@maxtaco are you ok with the names I picked?

go/libkb/context.go
```diff
@@ -690,3 +691,9 @@ func (m MetaContext) Keyring() (ret *SKBKeyringFile, err error) {
 	}
 	return m.ActiveDevice().Keyring(m)
 }
+
+var _ logger.CtxAndLogger = MetaContext{}
```
maxtaco: var _ logger.Contexter = MetaContext{}

go/libkb/context.go
maxtaco: `UpdateContextToLoggerContexter`

go/logger/context_with_logging.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"golang.org/x/net/context"
+)
+
+type Context struct {
+	ctx context.Context
+	Logger
+}
+
+var _ CtxAndLogger = Context{}
```
maxtaco: `Contexter`

PR #19814, opened 2019-09-20
teams: patch holes in hidden chain made via FTL

maxtaco:
- IF we FTL load with holes, we'll wind up with stubbed link in the hidden chain
- Slow loads want a full chain, so the have to be sensitive to these holes, so they can be repopulated when a slow load happens
- Repro test case, and simple fix.


maxtaco:
CC @AMarcedone

PR #19816, opened 2019-09-20
libkb: don't clear secret store on oneshot logout

maxtaco:
- the general idea is that when we make an ActiveDevice, we'll tag it with the Keyring mode that we used when making the device.  Then we'll plumb that through, all the way to Logout. When Logout fires, only clear the secret if the secret was used to load the key into memory.
- also do some cleanup and refactor around Logout, make it dangle off of MetaContext, not GlobalContext


go/engine/selfprovision.go
```diff
@@ -138,16 +138,10 @@ func (e *SelfProvisionEngine) loadUserAndActiveDeviceKeys(m libkb.MetaContext) (
 	}
 	e.perUserKeyring = pukRing
 
-	activeDevice := e.G().ActiveDevice
-	encKey, err := activeDevice.EncryptionKey()
+	keys, err := e.G().ActiveDevice.DeviceKeys()
```
maxtaco: cc @joshblum I made this atomic, it's safer this way...

go/engine/signup.go
```diff
@@ -88,7 +88,7 @@ func (s *SignupEngine) Run(m libkb.MetaContext) (err error) {
 		return fmt.Errorf("Cannot signup because of currently logged in user: %s", clRes.Reason)
 	}
 
-	if err = m.G().Logout(m.Ctx()); err != nil {
+	if err = m.Logout(); err != nil {
```
patrickxb: don't we need something like `LogoutNoSwitch` here?

patrickxb: I made a ticket: https://keybase.atlassian.net/browse/TRIAGE-820

PR #19820, opened 2019-09-20
Do additional search for 'keybase' if query is phone/email

zapu:
On `keybase` tab, do additional `email` or `phone` search query if we discover query is a valid email address or a valid phone number. If we get a result and it's not duplicate (by assertion) of another result,  insert it at the beginning of result list.

cc @keybase/react-hackers @keybase/y2ksquad

### DIFF
--- a/shared/actions/team-building.tsx
+++ b/shared/actions/team-building.tsx
@@ -6,6 +6,8 @@ import * as RouteTreeGen from './route-tree-gen'
 import * as Saga from '../util/saga'
 import * as RPCTypes from '../constants/types/rpc-gen'
 import {TypedState} from '../constants/reducer'
+import {validateNumber} from '../util/phone-numbers'
+import {validateEmailAddress} from '../util/email-address'
 
 const closeTeamBuilding = () => RouteTreeGen.createClearModals()
 export type NSAction = {payload: {namespace: TeamBuildingTypes.AllowedNamespace}}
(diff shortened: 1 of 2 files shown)


shared/actions/team-building.tsx
```diff
@@ -40,6 +42,38 @@ const apiSearch = async (
   }
 }
 
+// If the query is a well-formatted phone number or email, do additional search
+// and if the result is not already in the list, insert at the beginning.
+async function specialContactSearch(users: TeamBuildingTypes.User[], query: string, region: string | null) {
```
jzila: I'm not sure why you're mixing promises and async here, and why you're creating so many closures. I think it'd be easier to read if you refactored to the following approach:

```typescript
const apiSearchOne = async (
  query: string,
  service: TeamBuildingTypes.ServiceIdWithContact
): Promise<TeamBuildingTypes.User | null> =>
  (await apiSearch(
    query,
    service,
    1 /* maxResults */,
    true /* serviceSummaries */,
    false /* includeContacts */
  ))[0]

const specialContactSearch = async (users: TeamBuildingTypes.User[], query: string, region: string | null) => {
  let result
  const phoneNumber = validateNumber(query, region)
  if (phoneNumber.valid) {
    result = apiSearchOne(phoneNumber.e164, 'phone')
  } else if (validateEmailAddress(query)) {
    result = apiSearchOne(query, 'email')
  }
  if (result && !users.find(x => x.id === result.id)) {
    // Overwrite `prettyName` to make the special result stand out.
    result.prettyName = query
    return [result, ...users]
  }
  return users
}
```

zapu: I agree, that's better. `apiSearchOne` was a closure because it passed specific set of arguments that only worked for code within `specialContactSearch` so it was more of a local function, or local specialization of a function. But it's fine as normal unexported function as well. Thanks.

zapu: it was also mixed up because I didn't know `(await x()[0])` is legal. TIL

jzila: It's `(await x())[0]`, but yeah

PR #19822, opened 2019-09-20
fix location for new RN

mmaxim:


shared/package.json
```diff
@@ -190,7 +191,6 @@
     "simple-markdown": "0.5.2",
     "tlds": "1.203.1",
     "typedarray-to-buffer": "3.1.5",
-    "url-parse": "1.4.7",
```
chrisnojima: guess this is a side effect of a bad merge. we should keep this and ditch 1.4.4

PR #19830, opened 2019-09-20
init: allow a "kbfs.mode" config option

strib:
A few users on slow desktops wanted the ability to run KBFS in "constrained" mode.  They currently have to override the systemd definition for the kbfs service if they want to do that on Linux, and I don't think there's any way to currently do it on other platforms. So just put it under the "kbfs.mode" path in the config file, to make it much easier to do.

Issue: HOTPOT-839


heronhaye:
Could we document this somewhere?

strib:
Yes, this is among several things I need to add to https://keybase.io/docs/kbfs/understanding_kbfs.

PR #19839, opened 2019-09-21
Fs: move mimeType/viewType, http logic, and url generation into Go

songgao:


songgao:
@chrisnojima I didn’t get to actually try to remove them. Will have a follow up PR for those.

go/kbfs/simplefs/simplefs.go
```diff
@@ -3022,3 +3010,55 @@ func (k *SimpleFS) SimpleFSConfigureDownload(
...
+
+	var encodedSegments []string
+	for _, segment := range strings.Split(path.Path, "/") {
+		encodedSegments = append(encodedSegments, url.PathEscape(segment))
+	}
+
+	u := url.URL{
+		Scheme:   "http",
+		Host:     address,
+		Path:     "/files" + strings.Join(encodedSegments, "/"),
```
strib: Is this assuming that `path.Path` starts with `/`?  Why not do `strings.Join` over the whole thing including "/files"?

go/kbfs/libhttpserver/content_type_utils.go
```diff
@@ -88,20 +94,91 @@ func (w *contentTypeOverridingResponseWriter) override() {
...
+	"image/png":  true,
+	"image/jpeg": true,
+	"image/gif":  true,
+	"image/webp": true,
+}
+
+func getGetGUIFileContextReturnType(
+	viewType keybase1.GUIViewType, contentType string) (
+	keybase1.GUIViewType, string, string) {
+	return viewType, contentType, strconv.Itoa(int(keybase1.GUIViewType_DEFAULT))
```
strib: Seems like the "invariance" will always be constant no matter what invariance is given.  Is that intentional and permanent?  Or is there some plan to change this function in the future?  I'm pretty confused as to what good this is doing, but I might just not be understanding what invariance is doing.

If there's a plan for future work here, it'd be great to mention that in a comment.

strib: Also, the `strconv` result can be saved into a constant (even if it's technically a `var`) to avoid re-doing it on every request.

go/kbfs/simplefs/simplefs.go
```diff
@@ -3022,3 +3010,55 @@ func (k *SimpleFS) SimpleFSConfigureDownload(
...
+		RawQuery: "token=" + token,
+	}
+
+	r, err := http.NewRequest("HEAD", u.String(), nil)
+	if err != nil {
+		return keybase1.GUIFileContext{}, err
+	}
+	resp, err := k.httpClient.Do(r)
```
strib: Going through the HTTP client here seems pretty heavyweight.  Is there no easy way to fake a header without doing that?

go/kbfs/libhttpserver/content_type_utils.go
songgao: Good catch! The invariance is actually supposed to be derived from viewType or contentType. It's to make sure viewType doesn't change from the time GUI learns about it to the time GUI request the resource through webview,.

go/kbfs/simplefs/simplefs.go
songgao: It's what we've been doing on the GUI side, and this shouldn't be worse that. I think we could avoid the http part by duplicating [this](https://github.com/golang/go/blob/8b6fa668ee1aaf6c00073115cc09957195ee30e1/src/net/http/fs.go#L196-L201), but if the logic there ever changes we'd get a wrong value here. Though with the invariance thing, it seems at least we won't end up serving wrong content-type (which could lead to security concerns). After typing this out it does seem like a good idea, so I'll be going with this route and update the PR.

go/kbfs/libhttpserver/content_type_utils.go
```diff
@@ -57,6 +57,73 @@ func getDisposition(defaultInlineValue bool, mimeType string) string {
...
+func getGUIFileContext(contentTypeRaw, contentDispositionRaw string) (
+	viewType keybase1.GUIViewType, invariance string) {
+	contentTypeProcessed := beforeSemicolon(contentTypeRaw)
+	disposition := beforeSemicolon(contentDispositionRaw)
+	if disposition == "attachment" {
+		return getGetGUIFileContextResp(keybase1.GUIViewType_DEFAULT)
+	}
+	switch {
+	case contentTypeProcessed == "text/plain":
+		return getGetGUIFileContextResp(keybase1.GUIViewType_TEXT)
```
strib: This might be more clear if you have a param outside the switch, and then make one call after the switch, like:

```go
viewType := keybase1.GUIViewType_DEFAULT
switch {
case contentTypeProcessed == "text/plain":
    viewType = keybase1.GUIViewType_TEXT
case supportedImgMimeTypes[contentTypeProcessed]:
	viewType = keybase1.GUIViewType_IMAGE
...
}
return viewType, getGetGUIFileContextResp(viewType)
```

This would avoid having `getGetGUIFileContextResp` needlessly return its input parameter.  You could probably even get rid of `getGetGUIFileContextResp()` entirely in that case.

go/kbfs/simplefs/simplefs.go
```diff
@@ -3011,16 +3014,59 @@ func (k *SimpleFS) SimpleFSConfigureDownload(
...
+		return "", err
+	}
+	f, err := fs.OpenFile(finalElem, os.O_RDONLY, 0644)
+	if err != nil {
+		return "", err
+	}
+	var buf [sniffLen]byte
+	n, _ := io.ReadFull(f, buf[:])
+	return http.DetectContentType(buf[:n]), nil
+
```
strib: Extra blank line.

go/kbfs/simplefs/simplefs.go
```diff
@@ -3032,30 +3078,17 @@ func (k *SimpleFS) SimpleFSGetGUIFileContext(ctx context.Context,
 	if err != nil {
 		return keybase1.GUIFileContext{}, err
 	}
-
-	var encodedSegments []string
-	for _, segment := range strings.Split(path.Path, "/") {
+	encodedSegments := []string{"/files"}
```
strib: Minor optimization: split `kbfsPath` above this, and calculate the exact length of the `encodedSegments` slice before you allocate it.

songgao:
@strib thanks! Please take another look; I added another commit to make `viewTypeInvariance` required as we don't use that to figure out the invariance anymore and it seems safer this way.

go/kbfs/libhttpserver/content_type_utils.go
```diff
@@ -72,32 +72,27 @@ var supportedImgMimeTypes = map[string]bool{
...
 	disposition := beforeSemicolon(contentDispositionRaw)
-	if disposition == "attachment" {
-		return getGetGUIFileContextResp(keybase1.GUIViewType_DEFAULT)
-	}
 	switch {
+	case disposition == "attachment":
+		viewType = keybase1.GUIViewType_TEXT
 	case contentTypeProcessed == "text/plain":
-		return getGetGUIFileContextResp(keybase1.GUIViewType_TEXT)
+		viewType = (keybase1.GUIViewType_TEXT)
```
strib: Why the parens on all of these?

go/kbfs/simplefs/simplefs.go
```diff
@@ -3078,8 +3077,10 @@ func (k *SimpleFS) SimpleFSGetGUIFileContext(ctx context.Context,
 	if err != nil {
 		return keybase1.GUIFileContext{}, err
 	}
-	encodedSegments := []string{"/files"}
-	for _, segment := range strings.Split(kbfsPath.Path, "/") {
+
+	segments := strings.Split(kbfsPath.Path, "/")
+	encodedSegments := append(make([]string, 0, len(segments)), "/files")
```
strib: The length should be `len(segments)+1` I think.

songgao:
I don't think test failures are relevant. Merging!

PR #19852, opened 2019-09-23
Use team building dialog for the people page

chrisnojima:
This moves using the profile/people page to use the team builder instead of the old search code
next step (different PR) is to replace wallet's use of the search


chrisnojima:
@keybase/react-hackers bump

shared/actions/typed-routes.tsx
```diff
@@ -3,12 +3,21 @@
 
 import * as RouteTreeGen from './route-tree-gen'
 
+// TODO i think this should go away. either we dispatch a team building action that's typed that produces these or
+// we actually type the routes / props
```
MarcoPolo: > we dispatch a team building action that's typed that produces these
I think dispatching actions is too much for this. It's just a simple mapping. Dealing with actions has overhead.

> we actually type the routes / props
I'd prefer this. But until these can be properly typed, I think this is the best solution.

shared/profile/search/container.tsx
```diff
@@ -2,6 +2,7 @@ import Search from '.'
 import * as ProfileGen from '../../actions/profile-gen'
 import * as RouteTreeGen from '../../actions/route-tree-gen'
 import {connect} from '../../util/container'
+// TODO deprecate NNOW
```
MarcoPolo: I'll do this in another PR

PR #19856, opened 2019-09-23
Bring electron 6 back and fix windows build

chrisnojima:
This fixes the osx package by updating the packager and fixes the windows build (hates jettify etc)
This also fixes not being able to launch the daemon if the path had a space in it


chrisnojima:
kicking test builds

PR #19859, opened 2019-09-23
PICNIC-163 Change Input Line Height to Prevent Emoji Clipping

thebearjew:
macOS rescales emoji characters as it desires based on if the screen is retina or not. So on non-retina the emojis are larger and on retina the emojis are smaller. This is a OS rendering decision, Slack, Spotify, and other electron applications also express the same behavior.

However our lineHeight on non-retina desktop was too small and would clip the enlarged emojis at the top by 2 pixels.


thebearjew:

PR #19861, opened 2019-09-23
libfs: add node wrapper for update history, and support revision ranges

strib:
This PR moves the processing for the `.kbfs_update_history` file to a node wrapper, so that it can be read via SimpleFS and chat links, as well as through fuse/dokan.

Also, it previously only supported listing the _entire_ history, which is painful and slow for TLFs with large histories.  Now, it supports revision ranges, or even single revisions, as in:

```
cat /keybase/team/myteam/.kbfs_update_history.101-107
keybase fs read /keybase/team/myteam/.kbfs_update_history.3019
```

Issue: HOTPOT-855


go/kbfs/libfs/node_wrappers.go
```diff
@@ -54,20 +89,76 @@ type specialFileNode struct {
...
-	if !perTlfWrappedNodeNames[name.Plaintext()] {
+	plain := name.Plaintext()
+	if !shouldBeTlfWrappedNode(plain) {
 		return sfn.Node.ShouldCreateMissedLookup(ctx, name)
 	}
 
-	switch name.Plaintext() {
-	case StatusFileName:
+	switch {
+	case plain == StatusFileName:
```
jzila: I was trying to think of a way to integrate the `perTlfWrappedNodeNames` and `perTlfWrappedNodePrefixes` here, perhaps by having them be a map to function handlers, but actually I don't think it would make the code any cleaner.

PR #19862, opened 2019-09-23
TOFU invites in regular teams

zapu:
- Enable "phone number" tab in "Add to a team" team-building modal.
- Add support for resolving server-trust assertions in `AddMemberTx`, so when phone number / email is already resolvable, it will add member directly instead of inviting (this already worked for proof-backed assertions, server-trust always fell back to making an invite).
- Use `AddMemberByAssertionOrEmail` in `service_helper.go` `AddMemberByID` so the above can actually work. This should be safe, it's already used in `AddMembers` (for batch adding more than one member). This enables adding members from CLI or GUI without special-casing TOFU. Next step could be that `AddMemberByID` just calls `AddMembers` with 1-element list.
- There is an old `InviteEmailMember` which will fail in cases where e-mail is already resolved. [TODO]
- Add parameterized SBS tests in `systests/sbs_test.go` that do two different scenarios for phone, email, and rooter assertions.

Edit: not quite done yet :sweat: A test broke that uncovered an issue with this code.
TODO:
- [x] Fix subsequent invites for same assertion should be rejected.

### DIFF
--- a/go/teams/transactions.go
+++ b/go/teams/transactions.go
@@ -510,6 +510,7 @@ func (tx *AddMemberTx) AddMemberByAssertionOrEmail(ctx context.Context, assertio
 
 	if !doInvite {
 		username = libkb.NewNormalizedUsername(upak.Username)
+		uv = upak.ToUserVersion()
 		invite, err = tx.addMemberByUPKV2(ctx, upak, role, botSettings)
 		m.Debug("Adding keybase member: %s (isInvite=%v)", username, invite)
 		return username, uv, invite, err
--- a/go/teams/transactions.go
+++ b/go/teams/transactions.go
@@ -145,7 +145,7 @@ func (tx *AddMemberTx) addMemberAndCompleteInvite(uv keybase1.UserVersion,
 	// Preconditions: UV is a PUKful user, role is valid and not NONE, invite
 	// exists. Role is not RESTRICTEDBOT as botSettings are set to nil.
 	payload := tx.changeMembershipPayload(uv.Uid)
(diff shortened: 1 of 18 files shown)


zapu:
cc @keybase/y2ksquad

go/teams/transactions.go
```diff
@@ -481,7 +481,23 @@ func (tx *AddMemberTx) AddMemberByAssertionOrEmail(ctx context.Context, assertio
 	var upak keybase1.UserPlusKeysV2
 
 	if isServerTrustInvite {
-		doInvite = true
+		user, resolveRes, err := g.Resolver.ResolveUser(m, assertion)
```
maxtaco: this is getting hairy, can you factor this out into a helper function?

maxtaco:
The thing I'm most afraid of in this PR is that the server gets a wildcard invite to mash. I don't see how that could have been affected by this, but please double-check too, thanks!

shared/constants/team-building.tsx
```diff
@@ -18,18 +18,6 @@ export const allServices: Array<Types.ServiceIdWithContact> = [
   'email',
   ...searchServices.slice(1),
 ]
-const searchServicesWithEmail: Array<Types.ServiceIdWithContact> = [
-  ...searchServices.slice(0, 1),
-  'email',
-  ...searchServices.slice(1),
-]
-
-export function servicesForNamespace(namespace: Types.AllowedNamespace): Array<Types.ServiceIdWithContact> {
```
buoyad: 🎉

PR #19869, opened 2019-09-23
kbfsedits: avoid rename infinite loop

strib:
Consider this sequence of edit events, in order by increasing revision number:

1. mkdir "a"
2. create file "a/foo"
3. mkdir "b"
4. rename "a" to "c"
5. rename "b" to "a"
6. rename "c" to "a/d"

`TlfHistory` processes these events in reverse, and tries to adjust the filenames of old events to look like the new file names, based on the directory renames that have happened since the file was edited. So in the above sequence, by the time it gets to revision 2 concerning file "a/foo", it tries to figure out what the latest name of directory "a" is now.  So it looks it up in the rename table, and finds out it's "a/d".  But then it iterates on the prefix of that path ("a"), to see if it's been renamed, finds "a/d" again, and it gets stuck in a loop.

The answer is that once we have followed one rename (which occured at some revision, in this case 5), we should ignore other renames that happen at the same time or further in the future, since they can't possibly affect the prefix we're currently considering.

So this commit tracks the revision of each rename event, and follows that rule.  It also adds a test that reproduces the infinite loop without this fix.

Issue: HOTPOT-856

### DIFF
--- a/go/kbfs/kbfsedits/tlf_history.go
+++ b/go/kbfs/kbfsedits/tlf_history.go
@@ -342,7 +342,7 @@ func (r *recomputer) processNotification(
 		// See if any of the parent directories were renamed, checking
 		// backwards until we get to the TLF name.
 		prefix := filename
-		rev := notification.Revision
+		latestRenameRev := notification.Revision
 		suffix := ""
 		for strings.Count(prefix, "/") > 4 {
 			var finalElem string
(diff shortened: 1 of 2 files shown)


go/kbfs/kbfsedits/tlf_history.go
```diff
@@ -341,15 +342,20 @@ func (r *recomputer) processNotification(
 		// See if any of the parent directories were renamed, checking
 		// backwards until we get to the TLF name.
 		prefix := filename
+		rev := notification.Revision
```
songgao: Maybe rename this to `latestRenamedRev` to make it clearer?

PR #19872, opened 2019-09-23
Added configuration parameters for blind merkle tree.

AMarcedone:
This builds on AMarcedone/y2k-598, so please review the differences with that branch only (i.e. the blindtree package).


heronhaye:
@AMarcedone can you change the base branch to 598?

AMarcedone:
@heronhaye done

go/blindtree/values.go
```diff
@@ -0,0 +1,78 @@
...
+		panic("Unrecognized Value Type")
+	}
+}
+
+type SigID []byte
+type Teamv1HiddenTail struct {
+	_struct struct{} `codec:",toarray"` //nolint
+	SigID   SigID
+	LinkID  sig3.LinkID
+	Seqno   keybase1.Seqno
```
heronhaye: What was the verdict about including SigID here? Do we need it?

AMarcedone: Removing both SigID and Seqno as we discussed. cc @maxtaco

PR #19890, opened 2019-09-24
Fix displaying the explanatory prompt under service tabs with an empty search

cjb:

If we're not looking at the Keybase service tab, then recommendations aren't being loaded.

Before:

Screen Shot 2019-09-24 at 12 22 30 AM

After:

Screen Shot 2019-09-24 at 12 22 38 AM

### DIFF
--- a/shared/team-building/index.tsx
+++ b/shared/team-building/index.tsx
@@ -320,7 +320,8 @@ class TeamBuilding extends React.PureComponent<Props, {}> {
   )
 
   _listBody = () => {
-    const showRecPending = !this.props.searchString && !this.props.recommendations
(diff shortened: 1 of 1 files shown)


shared/team-building/index.tsx
```diff
@@ -504,7 +505,7 @@ class TeamBuilding extends React.PureComponent<Props, {}> {
       />
     )
 
-    // Handle when team-buiding is making a new chat v.s. adding members to a team.
+    // Handle when team-building is making a new chat v.s. adding members to a team.
```
thebearjew: This was me sorry

PR #19897, opened 2019-09-24
dont want systemd if started with custom home

heronhaye:


maxtaco:
what do you think about specified via environment variable?

heronhaye:
Doesn't seem to work for me.

```
% HOME=/home/md/hello keybase status
Username:      modalduality
```

maxtaco:
I don't understand what you mean by "it doesn't work" --- does the repro not work, or the fix doesn't work?

maxtaco:
Right, my point is that this fails:

```
HOME=/tmp/foo XDG_RUNTIME_DIR= keybase status
```

for the same reason that this fails:

```
keybase --home=/tmp/foo status
```

And therefore, we should disable systemd when we see `HOME` for the same reason we disable it when we see `--home`

maxtaco:
@heronhaye I pushed another attempt ^?

zapu:
Isn't `$HOME` always set? so it's difficult to tell a non-standard home. Can probably ask the system if `$HOME` is real home dir for current user (the one that's also in `/etc/passwd`)

maxtaco:
Oh, @zapu, good point...

heronhaye:
(not sure what we're doing here, will leave this PR open for now)

heronhaye:
closing for now

PR #19900, opened 2019-09-24
No logsends from google ai

MarcoPolo:
@keybase/react-hackers 

Adds a new constant `isTestDevice` which is true if it's in a pre launch report test.


MarcoPolo:
merging this so I can start some mobile builds, https://github.com/keybase/client/pull/19902 fixes the failing CI

PR #19906, opened 2019-09-25
Do not send SBS system messages for KB invites

zapu:
Keybase-type invites are supposed to be transparent for the users. When adding someone without PUK, they are still listed in "members" list in the GUI (instead of "invites"), and chat with them should look like a chat with normal user.


zapu:
image

there is still this bug, i'll try to fix it here as well

jakob223:
> image
> 
> there is still this bug, i'll try to fix it here as well

Should be as simple as adding the word 'proved' in line 39 of `shared/chat/conversation/messages/system-sbs-resolve/index.tsx` and 'verified' to the beginnings of 35 and 37

PR #19911, opened 2019-09-25
Waiting states for recover password via paper key

buoyad:
Manages waiting states in the saga for submitting a paper key and new password. cc @keybase/y2ksquad


shared/actions/recover-password.tsx
```diff
@@ -123,18 +128,24 @@ const getPaperKeyOrPw = (
         yield Saga.put(RecoverPasswordGen.createRestartRecovery())
       }
     } else {
+      const state = yield* Saga.selectState()
+      if (anyWaiting(state, Constants.getPaperKeyWaitingKey)) {
+        yield Saga.put(WaitingGen.createClearWaiting({key: Constants.getPaperKeyWaitingKey}))
```
chrisnojima: its safe to just do this w/o checking

PR #19919, opened 2019-09-25
fix additional race of menu and app

chrisnojima:
its possible for the app to start up faster than the menu sometimes


shared/desktop/app/node.desktop.tsx
```diff
@@ -337,7 +339,13 @@ const start = () => {
 
   // Load menubar and get its browser window id so we can tell the main window
   menuBar(id => {
-    menubarWindowID = id
+    // its possible the app started up way before we get this id in rare cases
+    if (appStartedUp && id) {
```
jzila: Why would `id` not be set?

chrisnojima: Should be but who knows. Doesn't hurt

PR #19920, opened 2019-09-25
re init hidden package

joshblum:
fixes a failed seed check on restricted bots since the role would be `TeamRole_NONE` when the hiddenPackage was initialized

`▶ ERROR hidden team loader error: seed check at generation 3 wasn't found`

### DIFF
--- a/go/teams/loader.go
+++ b/go/teams/loader.go
@@ -732,7 +732,10 @@ func (l *TeamLoader) load2InnerLockedRetry(ctx context.Context, arg load2ArgT) (
 	if err != nil {
 		return nil, err
 	}
-	// update the hidden package once ret != nil
+	// Update the hidden package with team metadata once we process all of the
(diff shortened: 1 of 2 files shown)


go/teams/loader.go
```diff
@@ -724,6 +724,17 @@ func (l *TeamLoader) load2InnerLockedRetry(ctx context.Context, arg load2ArgT) (
...
 
+	if ret == nil {
+		return nil, fmt.Errorf("team loader fault: got nil from load2")
+	}
+
+	encKID, gen, role, err := l.hiddenPackageGetter(mctx, arg.teamID, ret, arg.me)()
+	if err != nil {
+		return nil, err
+	}
+	// update the hidden package once ret != nil
```
maxtaco: can you say why we do this? seems like we really only need to do this for the hidden bot case, wanting to skip seedchecks...

PR #19930, opened 2019-09-25
Surya/y2 k 673/nopw arch

heronhaye:
- [x] JS changes to match new RPCs
- [x] Sanity check in CLI and GUI


go/service/account.go
```diff
@@ -106,7 +106,7 @@ func (h *AccountHandler) ResetAccount(ctx context.Context, arg keybase1.ResetAcc
 
 	m.Debug("reset account succeeded, logging out.")
 
-	return m.Logout()
+	return m.LogoutKillSecrets()
```
heronhaye: We are resetting an account so we have a passphrase - kill secrets, don't force

go/service/login.go
```diff
@@ -27,10 +27,10 @@ func (h *LoginHandler) GetConfiguredAccounts(context context.Context, sessionID
 	return h.G().GetConfiguredAccounts(context)
 }
 
-func (h *LoginHandler) Logout(ctx context.Context, sessionID int) (err error) {
+func (h *LoginHandler) Logout(ctx context.Context, arg keybase1.LogoutArg) (err error) {
 	defer h.G().CTraceTimed(ctx, "Logout [service RPC]", func() error { return err })()
 	mctx := libkb.NewMetaContext(ctx, h.G()).WithLogTag("LOGOUT")
-	eng := engine.NewLogout()
+	eng := engine.NewLogout(libkb.LogoutOptions{Force: arg.Force})
```
heronhaye: Logout RPC call - always kill secrets, force if desired in rpc arg

go/client/cmd_deprovision.go
```diff
@@ -61,6 +69,7 @@ func (c *CmdDeprovision) Run() (err error) {
 		SessionID: 0,
 		Username:  username,
 		DoRevoke:  c.loggedIn,
+		Force:     c.force,
```
heronhaye: Force deprovision over unset passphrase if desired

go/engine/account_delete.go
```diff
@@ -70,5 +70,5 @@ func (e *AccountDelete) Run(m libkb.MetaContext) error {
 		return err
 	}
 	m.Debug("account deleted, logging out")
-	return m.Logout()
+	return m.LogoutWithOptions(libkb.LogoutOptions{KeepSecrets: false, Force: true})
```
heronhaye: deleting account, so nuke secrets forcibly

heronhaye: (even without passphrase)

go/engine/deprovision.go
```diff
@@ -82,7 +84,7 @@ func (e *DeprovisionEngine) attemptLoggedInRevoke(m libkb.MetaContext) error {
 	}
 
 	m.UIs().LogUI.Info("Logging out...")
-	if err = m.Logout(); err != nil {
+	if err = m.LogoutWithOptions(libkb.LogoutOptions{KeepSecrets: false, Force: e.force}); err != nil {
```
heronhaye: deprovision, don't keep secrets, allow user to select force

go/engine/paperprovision.go
```diff
@@ -54,9 +54,7 @@ func (e *PaperProvisionEngine) Run(m libkb.MetaContext) (err error) {
 	defer m.Trace("PaperProvisionEngine#Run", func() error { return err })()
 
 	// clear out any existing session
-	// XXX THIS IS DANGEROUS SINCE IT COULD CATASTROPHICALLY DELETE A NOPW KEYCHAIN ENTRY XXX
-	// XXX FIXME XXX
-	err = m.Logout()
+	err = m.LogoutKeepSecrets()
```
heronhaye: not sure?

go/engine/passphrase_recover.go
```diff
@@ -218,7 +218,8 @@ func (e *PassphraseRecover) loginWithPaperKey(mctx libkb.MetaContext) (err error
 
 	if err := e.changePassword(mctx); err != nil {
 		// Log out before returning
-		if err2 := RunEngine2(mctx, NewLogout()); err2 != nil {
+		// TODO FIX ME
+		if err2 := RunEngine2(mctx, NewLogout(libkb.LogoutOptions{})); err2 != nil {
```
heronhaye: not sure...why are we logging out on errors here? seems bad

go/engine/signup.go
```diff
@@ -94,13 +94,7 @@ func (s *SignupEngine) PaperKey() *libkb.PaperKeyPhrase {
...
 	defer m.Trace("SignupEngine#Run", func() error { return err })()
 
-	// Make sure we're starting with a clear login state. But check
-	// if it's fine to logout current user.
-	if clRes := libkb.CanLogout(m); !clRes.CanLogout {
-		return fmt.Errorf("Cannot signup because of currently logged in user: %s", clRes.Reason)
-	}
-
-	if err = m.Logout(); err != nil {
+	if err = m.LogoutKeepSecrets(); err != nil {
```
heronhaye: Behavior change: keep secrets when signing up to new user

go/libkb/context.go
```diff
@@ -571,7 +571,7 @@ func (m MetaContext) LogoutAndDeprovisionIfRevoked() (err error) {
 
 	if doLogout {
```
heronhaye: if revoked: kill secrets forcibly

go/engine/passphrase_recover.go
zapu: I think this is only in the code path where you are initially logged out, then you log in using paper key, and this lets you change your passphrase. If change fails, it wants to log you out so you can get to the initial state.

go/libkb/logout.go
```diff
@@ -97,16 +123,16 @@ func (m MetaContext) logoutSecretStore(username NormalizedUsername, killSecrets
 
 // LogoutSelfCheck checks with the API server to see if this uid+device pair should
 // logout.
-func (m MetaContext) LogoutSelfCheck() error {
-	g := m.G()
+func (mctx MetaContext) LogoutSelfCheck() error {
+	g := mctx.G()
```
zapu: will this pass lint? i think there is some rule where all receivers for given type in a file need to be named the same

go/engine/passphrase_recover.go
heronhaye: So what options do we want here? Kill secrets (since you have a passphrase) and force?

heronhaye:
cc @patrickxb @maxtaco (this PR makes it so it is basically impossible to clear your secret accidentally by moving the CanLogout/hasrandompw check into mctx.Logout). A subsequent PR will change `LoadHasRandomPw` to update on gregor notifications and store in config.json.

go/service/login.go
```diff
@@ -27,15 +27,15 @@ func (h *LoginHandler) GetConfiguredAccounts(context context.Context, sessionID
...
 	return engine.RunEngine2(mctx, eng)
 }
 
 func (h *LoginHandler) Deprovision(ctx context.Context, arg keybase1.DeprovisionArg) error {
-	eng := engine.NewDeprovisionEngine(h.G(), arg.Username, arg.DoRevoke)
+	eng := engine.NewDeprovisionEngine(h.G(), arg.Username, arg.DoRevoke, libkb.LogoutOptions{KeepSecrets: false, Force: true})
```
heronhaye: Deprovisioning - kill secrets with force. Since we have another device.

go/libkb/logout.go
```diff
@@ -128,11 +154,61 @@ func (m MetaContext) LogoutSelfCheck() error {
...
-		m.Debug("LogoutSelfCheck: logging out...")
-		return m.Logout()
+		mctx.Debug("LogoutSelfCheck: logging out...")
+		return mctx.LogoutKillSecrets()
 	}
 
 	return nil
 }
+
+func CanLogout(mctx MetaContext) (res keybase1.CanLogoutRes) {
```
heronhaye: This function is unchanged

heronhaye:
ping @zapu

go/engine/passphrase_recover.go
zapu: I think this is good actually, because `NewLoginWithPaperKey` logs you in and stores secrets in keychain (maybe it shouldn't? i couldn't figure out exactly), so we want to clear after ourselves here if anything fails

zapu: So yeah killSecrets and force

go/engine/login.go
```diff
@@ -140,7 +140,7 @@ func (e *Login) Run(m libkb.MetaContext) (err error) {
 	// clear out any existing session:
 	m.Debug("clearing any existing login session with Logout before loading user for login")
 	// If the doUserSwitch flag is specified, we don't want to kill the existing session
-	err = m.LogoutCurrentUserWithSecretKill(!e.doUserSwitch)
+	err = m.LogoutWithOptions(libkb.LogoutOptions{KeepSecrets: e.doUserSwitch})
```
zapu: there is already `LogoutKeepSecrets` in `checkLoggedInAndNotRevoked` that's called. What flow requires logging out again here?

go/engine/paperprovision.go
zapu: I don't think you should be able to land here if you are logged in. Maybe it's better to bail out if we are logged in (so we are in bad state from this engine's point of view) instead of logging out to clear the state.

go/engine/login.go
heronhaye: It only does that if LoggedInWrongUser error right? not always?

go/engine/paperprovision.go
heronhaye: Doesn't seem bad to me if you're logged in when you do this. We can just say provisioning a device is always done with login -s.

go/engine/login.go
zapu: you're correct, there are two cases in which `false, nil` is returned from `checkLoggedInAndNotRevoked` in which we would get this far, and only one of them (`LoggedInWrongUserError` logs out, `NoActiveDeviceError` doesn't).

go/engine/paperprovision.go
zapu: My questions is more like what exactly happens when you are user switching? doesn't it first logs the current user out (with keepSecrets)? So is the only expected state to be in this engine is logged out?

heronhaye:
CI failure is flake

PR #19933, opened 2019-09-25
kbpki_client: add id->username cache

strib:
`folderBranchOps.GetNodeMetadata` gets the normalized username given an ID on every call, which results in a service RPC.  However, this mapping can be easily cached since it basically never changes. So this adds a small LRU cache for those mappings.

Issue: HOTPOT-877

### DIFF
--- a/go/kbfs/libkbfs/interfaces.go
+++ b/go/kbfs/libkbfs/interfaces.go
@@ -823,6 +823,10 @@ type KBPKI interface {
 
 	// NotifyPathUpdated sends a path updated notification.
 	NotifyPathUpdated(ctx context.Context, path string) error
+
+	// InvalidateTeamCacheForID instructs KBPKI to discard any cached
+	// information about the given team ID.
+	InvalidateTeamCacheForID(tid keybase1.TeamID)
 }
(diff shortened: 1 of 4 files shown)


songgao:
What about in SBS case? Does normalized username still not change?

strib:
> What about in SBS case? Does normalized username still not change?

This is only looking up normalized usernames from UIDs.  In this SBS case, there is no UID, so this function (and cache) won't be used.

strib:
@songgao when I answered your question above, I was thinking about users (since that is the performance problem I'm solving here).  But it is true that _team_ names can change, both because of SBS resolutions and subteam renames.  I don't _think_ this particular cache would get hit much in those cases, but I just added a new commit (354e812) to invalidate the cache on team renames.  Thanks for the reminder.  Please take another look!

songgao:
NP! Is team name change by SBS resolutions is because of implicit team name?

strib:
> NP! Is team name change by SBS resolutions is because of implicit team name?

Both implicit teams containing SBS assertions, and subteam renames, can cause team name changes.

PR #19936, opened 2019-09-25
[Y2K-703] Remove pin option for ephemeral messages

jzila:


jzila:
Never mind, this actually works perfectly right now.

PR #19937, opened 2019-09-26
immer fs 1

songgao:


shared/actions/fs/index.tsx
```diff
@@ -151,8 +134,8 @@ const getSyncConfigFromRPC = (
     case RPCTypes.FolderSyncMode.partial:
       return Constants.makeTlfSyncPartial({
         enabledPaths: config.paths
-          ? I.List(config.paths.map(str => Types.getPathFromRelative(tlfName, tlfType, str)))
-          : I.List(),
+          ? config.paths.map(str => Types.getPathFromRelative(tlfName, tlfType, str))
+          : [],
```
chrisnojima: may not matter but just a note. `I.List() === I.List()` and `[] !== []` so either ensure 
1. this doesn't matter (likely)
1. change to accept undefined
1. or make `const emptyPaths = []` and just re-use that

shared/constants/fs.tsx
```diff
@@ -194,11 +206,11 @@ export const makeUploads = I.Record<Types._Uploads>({
   writingToJournal: I.Set(),
 })
 
-export const makeTlfs = I.Record<Types._Tlfs>({
-  loaded: false,
-  private: I.Map(),
-  public: I.Map(),
-  team: I.Map(),
+export const makeTlfs = (partial: Partial<Types.Tlfs>): Types.Tlfs => ({
```
chrisnojima: just a note about these helpers. before they  were strictly necessary as we needed the record factories, but now they're plain objects so their value is diminished. if we need all these fields filled then thats fine but also if its just used in a couple of places those things can just make the objects themselves

shared/fs/browser/index.stories.tsx
```diff
@@ -29,7 +29,13 @@ const _provider = {
 const storeCommon = Sb.createStoreWithCommon()
 const storeShowingSfmi = {
   ...storeCommon,
-  fs: storeCommon.fs.update('sfmi', sfmi => sfmi.set('showingBanner', true)),
+  fs: {
```
chrisnojima: note you can use immer `produce` for this kind of thing

shared/reducers/fs.tsx
```diff
@@ -96,463 +121,437 @@ const updatePathItem = (
...
-      const visibility = Types.getVisibilityFromElems(elems)
-      if (!visibility) {
-        return state
-      }
-      // @ts-ignore
-      return state.mergeIn(['tlfs', visibility, elems[2]], {
-        isIgnored: action.type === FsGen.favoriteIgnore,
-      })
+    draftState.tlfs[visibility] = new Map([
+      ...draftState.tlfs[visibility],
```
chrisnojima: likely quicker / simpler to just do
```
const visMap = new Map(draftState.tlfs[visibility])
visMap.set(elems[2], ...) // etc
```

shared/constants/fs.tsx
songgao: Yeah a lot of these have to be populated, so I figured I'd just keep them. This particular one is only used in the reducer for initial state, so I'll get rid of it.

shared/actions/fs/index.tsx
songgao: Ah, good point. I don't think there's correctness issue here. It'd matter in reducer for performance reasons but we do `isEqual` comparisons there.

I don't think a const emptyPaths would work in a lot of scenarios since we end up doing a `.map` later on and unfortunately `emptyPaths.map(a=>a) !== emptyPaths`

PR #19942, opened 2019-09-26
PICNIC-516 - Row Highlighting Regression On Dekstop

thebearjew:
Arrow keys highlight rows, now actually showing up after being lost in previous PR.


thebearjew:
cc @keybase/picnicsquad

PR #19945, opened 2019-09-26
update a stellar link

xgess:
change the link to something that actually has the answer. seems like stellar removed this content from their site and left it to each integrator to explain. this link might be a little technical for a curious, unsophisticated user of keybase, but the alternatives seem to be reddit answers or blockchain.com.

we could add a little FAQ doc somewhere like https://keybase.io/docs/stellar/faq with this (and perhaps other) answers. what do you think? is this good enough?


xgess:
@keybase/picnicsquad

PR #19955, opened 2019-09-26
Small electron scroll fixes

chrisnojima:


chrisnojima:
i tested ios and it looks fine

PR #19963, opened 2019-09-26
add copy button to withReveal CopyText

hyperobject:
brings the CopyText component back into spec by adding the copy button to components with the `withReveal` prop. 

@keybase/react-hackers @keybase/hotpotatosquad

### DIFF
--- a/shared/common-adapters/copy-text.tsx
+++ b/shared/common-adapters/copy-text.tsx
@@ -17,12 +17,15 @@ type Props = {
   hideOnCopy?: boolean
   onReveal?: () => void
   withReveal?: boolean
-  text: string
+  text?: string
+  placeholderText?: string
+  loadText?: () => void
 }
 
 const CopyText = (props: Props) => {
   const [revealed, setRevealed] = React.useState(!props.withReveal)
   const [showingToast, setShowingToast] = React.useState(false)
+  const [requestedCopy, setRequestedCopy] = React.useState(false)
 
   const setShowingToastFalseLater = useTimeout(() => setShowingToast(false), 1500)
   React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


shared/common-adapters/copy-text.tsx
```diff
@@ -35,6 +35,9 @@ const CopyText = (props: Props) => {
   const dispatch = Container.useDispatch()
   const copyToClipboard = (text: string) => dispatch(ConfigGen.createCopyToClipboard({text}))
   const copy = () => {
+    if (props.withReveal && !revealed) {
+      reveal()
+    }
```
songgao: Is this intentional? It seems copying shouldn't reveal the content in UI.

hyperobject: Yeah, it's intentional - I was specifically thinking of the copyText usage for stellar secret keys, which doesn't actually decrypt the secret key until `onReveal()` is called. (although that particular use case is still a little wonky here because `copyToClipboard` doesn't wait for whatever `onReveal` is doing)

songgao: I see, interesting. It seems the current interface is a bit confusing. Here's some ideas: perhaps we could add these props to the `CopyText` component:

```
loadText?: ()=>void
textPlaceholder?: string
```

If `text` is empty, we show `textPlaceholder` when revealed. Otherwise, show text when revealed. If `withReveal` is not set, we'd just call `loadText` in a `React.Effect` with empty dependent (i.e. like `componentDidMount`). For copy, we'd need to add a state (e.g. `requestedCopy`) to track if user has requested copy, and in the `onClick` event for the copy button we simply flip `requestedCopy`, and call `loadText()`. Then we need an effect that does the actual copy thing when `text` becomes non-empty, _and_ flip `requestedCopy` back.

In wallet's setting here https://github.com/keybase/client/blob/55f3b947934aff4b70c2c472466ca90ab07a4d23/shared/wallets/wallet/settings/index.tsx#L147 , we could just pass in `secretKey` as text, and that string as `textPlaceholder`.

Sorry it's a lot larger than it seems. What do you think? I'm open to other ideas!

hyperobject: Sounds like a good plan to me! I'll implement that.

songgao:
Looks great! I think we just need one more thing: if `withReveal` is set to `false`, `text` is not provided, we'd need to call `loadText` too in an effect. An alternative is to only allow `withReveal` when `text` is provided.

hyperobject:
@songgao I think this should already be covered by https://github.com/keybase/client/pull/19963/files#diff-40766f02cd35e07e59435b5d3b9654aeR42-R46; if no text is provided, the copy action will kick off the requestedCopy effect which calls loadText.

songgao:
That only happens when copy button is pressed right? If withReveal is set to false, and user does nothing, it'd never load the text and show the placeholder text unless text is provided in the first place.

hyperobject:
ahhhh good point, I'll add that in!

PR #19972, opened 2019-09-26
fix double encoding in filepreview URLs

songgao:


songgao:
Yeah I tested it and tried a few different files. I think encoding happens when you call `url.String()`, but yeah it's a bit confusing.

PR #19974, opened 2019-09-27
protocol: add support for setting floats in the config file

strib:
A future KBFS PR needs to read float values out of the config file, and this seemed easy enough to do. Hopefully it's not controversial, but let me know if you feel like it is.

Issue: HOTPOT-891


protocol/avdl/keybase1/config.avdl
```diff
@@ -195,6 +195,7 @@ protocol config {
     boolean isNull;
     union { null, boolean } b;
     union { null, int } i;
+    union { null, double } d;
```
strib: Wasn't sure whether it made the most sense to call this field `f` or `d`.  It's a bit weird using the `D` field in the Go code, but it felt right here.

go/kbfs/libkbfs/init.go
```diff
@@ -62,6 +62,15 @@ const CtxInitID = "KBFSINIT"
 // AdditionalProtocolCreator creates an additional protocol.
 type AdditionalProtocolCreator func(Context, Config) (rpc.Protocol, error)
 
+const (
+	configModePrefixStr            = "kbfs."
+	configModeStr                  = configModePrefixStr + "mode"
+	configBlockCachePrefixStr      = configModePrefixStr + "block_cache."
+	configBlockCacheMemMaxBytesStr = configBlockCachePrefixStr + "mem_max_bytes"
+	configBlockCacheDiskMaxFracStr = configBlockCachePrefixStr + "disk_max_fraction"
+	configBlockCacheSyncMaxFracStr = configBlockCachePrefixStr + "sync_max_fraction"
```
xgess: this seems harder to read than (and without significant benefits over) fully specifying these strings:

```golang
configModeStr = "kbfs.mode"
configBlockCacheMemMaxBytesStr = "kbfs.mode.mem_max_bytes"
configBlockCacheDiskMaxFracStr = "kbfs.block_cache.disk_max_fraction"
configBlockCacheSyncMaxFracStr = "kbfs.block_cache.sync_max_fraction"
```

protocol/avdl/keybase1/config.avdl
xgess: id say slight preference for whichever word / actual type (double vs float) is going to be used more frequently. looks like it's float from this PR. but i wouldn't go to the mat over this. 🤷‍♂

patrickxb:
This seems fine to me, @xgess's comments are reasonable (i.e. do we really need floats instead of percents?), but I don't have any problems adding floats to the config file data types if needed.

go/kbfs/libkbfs/init.go
strib: Oh oops, I didn't mean to have this commit in this PR, was supposed to be a separate one for KBFS reviewers.  I'll move it out, but also take your advice.  I was trying to minimize copy/paste of course, but happy to have it more readable if you think that's better.

strib:
Thanks both for the reviews.  I agree that percentages could work most of the time, but I felt like there were some case that floats would make more sense in the use cases I needed it for.  First is that we already expose it as a float from the CLI, so in case anyone used it already, I wanted an easy transition.  Second, if a user has 1 TB of free space, they still might reasonably want to restrict KBFS to using say 5 GB, and we couldn't get that kind of resolution with percents.

protocol/avdl/keybase1/config.avdl
strib: Changed to `f`.

PR #19975, opened 2019-09-27
init: read block cache config options from the config file

strib:
Depends on #19974.

* `kbfs.block_cache.mem_max_bytes`: the max size of the in-memory block cache.
* `kbfs.block_cache.disk_max_fraction`: the fraction of the disk's free space that the transient disk block cache may use.
* `kbfs.block_cache.sync_max_fraction`: the fraction of the disk's free space that the disk sync cache may use.

Issue: HOTPOT-891

### DIFF
--- a/go/kbfs/libkbfs/init.go
+++ b/go/kbfs/libkbfs/init.go
@@ -62,6 +62,15 @@ const CtxInitID = "KBFSINIT"
 // AdditionalProtocolCreator creates an additional protocol.
 type AdditionalProtocolCreator func(Context, Config) (rpc.Protocol, error)
 
+const (
+	configModePrefixStr            = "kbfs."
+	configModeStr                  = configModePrefixStr + "mode"
+	configBlockCachePrefixStr      = configModePrefixStr + "block_cache."
+	configBlockCacheMemMaxBytesStr = configBlockCachePrefixStr + "mem_max_bytes"
+	configBlockCacheDiskMaxFracStr = configBlockCachePrefixStr + "disk_max_fraction"
+	configBlockCacheSyncMaxFracStr = configBlockCachePrefixStr + "sync_max_fraction"
+)
+
 // InitParams contains the initialization parameters for Init(). It is
 // usually filled in by the flags parser passed into AddFlags().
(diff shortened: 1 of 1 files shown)


go/kbfs/libkbfs/init.go
```diff
@@ -659,7 +663,64 @@ func doInit(
...
+	ctx context.Context, kbCtx Context, params InitParams,
+	log logger.Logger) uint64 {
+	cap := params.CleanBlockCacheCapacity
+
+	// Use the capacity from the config file if none is provided on
+	// the command line.
+	if cap == 0 {
+		config := kbCtx.GetEnv().GetConfig()
+		capInt, ok := config.GetIntAtPath(configBlockCacheMemMaxBytesStr)
+		if ok {
```
jakob223: there's no fallback if it's not in the config file. Intentional?

jakob223: I imagine it might be nice to default to the values we had before.

strib: @jakob223 fall back to what?  `cap` is still set to whatever the parameter value was.  (Note that this is setting `capInt`, not `cap`, and it's `cap` that's returned below.

go/kbfs/libkbfs/init.go
```diff
@@ -659,7 +663,64 @@ func doInit(
...
+func getCacheFrac(
+	ctx context.Context, kbCtx Context, param, defaultVal float64,
+	configKey string, log logger.Logger) float64 {
+	frac := param
+
+	// Use the fraction from the config file if none is provided on
+	// the command line.
+	if frac == 0 {
+		config := kbCtx.GetEnv().GetConfig()
+		configFrac, ok := config.GetFloatAtPath(configKey)
```
jakob223: same question as above

jakob223:
Looks good modulo a question. Also, sorry for probably like 5 emails

go/kbfs/libkbfs/init.go
jakob223: Oh, I see. 👍

PR #19979, opened 2019-09-27
Fix Avatar Image Uploading File Type Restrictions

Ruegg:


jakob223:
Tried to find the part of the server that does image processing to make sure this was the right set of images, but I couldn't find what processes the images as the "kraken alternative".

Ruegg:
> Tried to find the part of the server that does image processing to make sure this was the right set of images, but I couldn't find what processes the images as the "kraken alternative".

I've used the same set of restrictions on file types that the web client for keybase.io uses :)

PR #19980, opened 2019-09-27
custom scrollbars. dark aware scrollbars. obtrusive aware scrollbars

chrisnojima:
This PR enables

1. Detection of osx auto hide scrollbar behavior
1. Custom css for dark mode scrollbars

cc: @adamjspooner

### DIFF
--- a/shared/desktop/renderer/main2.desktop.tsx
+++ b/shared/desktop/renderer/main2.desktop.tsx
@@ -99,11 +99,15 @@ const FontLoader = () => (
 let store
 
 const DarkCSSInjector = () => {
-  const className = useSelector(state => isDarkMode(state.config)) ? 'darkMode' : ''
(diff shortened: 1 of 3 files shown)


hyperobject:
Looks mostly good to me, a few potential issues:
* In dark mode with autohiding scrollbar, the scrollbar thumb doesn't seem like it's the right color: Screen Shot 2019-09-27 at 10 50 51 AM
* after starting app in dark mode + autohiding scrollbar, changing the system to not autohiding scrollbar causes the app's scrollbars to be in light mode

chrisnojima:
@hyperobject we don't (and can't believe) control the auto hiding scroll bars colors at all. in this PR we don't attempt to

heronhaye:
Looks great. Autohide is only on osx?

chrisnojima:
i believe so, but the code is agnostic

PR #19981, opened 2019-09-27
Tree optimization and type for surveyor response

AMarcedone:
Also, this allows to build a tree without any keys.


AMarcedone:
Added a couple of bugfixes.

PR #19982, opened 2019-09-27
PICNIC-163 - Force chat input height to match content-area

thebearjew:
This is a fix that solves a scrollbar issue on Linux.
Turns out that on macOS, the height of the `textarea` will be resized from `21` to `22` px to accommodate `line-height: 22px`.
This is not the case on Linux, and as a result a scrollbar is shown.

2019-09-27-090828_1924x146_scrot


heronhaye:
I did some random stuff and couldn't repro on this branch. Think it works. Thanks!

PR #19983, opened 2019-09-27
Add margin to conversation list area for pins and banners

hyperobject:
This PR fixes an issue where pinned messages and thread load status banners would partially hide the "digging ancient messages" component by conditionally adding margin to the top of conversation when a pinned message or thread load banner exists.

@keybase/hotpotatosquad @keybase/react-hackers

### DIFF
--- a/shared/chat/conversation/list-area/normal/index.native.tsx
+++ b/shared/chat/conversation/list-area/normal/index.native.tsx
@@ -188,13 +188,7 @@ class ConversationList extends React.PureComponent<Props> {
   render() {
     return (
       <ErrorBoundary>
-        <Box
-          style={Styles.collapseStyles([
-            styles.container,
-            this.props.hasStatus && {marginTop: 28},
-            this.props.hasPinnedMessage && {marginTop: 46},
-          ])}
-        >
+        <Box style={styles.container}>
           <NativeVirtualizedList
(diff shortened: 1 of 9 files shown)


shared/chat/conversation/list-area/normal/index.native.tsx
```diff
@@ -188,7 +188,13 @@ class ConversationList extends React.PureComponent<Props> {
   render() {
     return (
       <ErrorBoundary>
-        <Box style={styles.container}>
+        <Box
+          style={Styles.collapseStyles([
+            styles.container,
+            this.props.hasStatus && {marginTop: 28},
+            this.props.hasPinnedMessage && {marginTop: 46},
```
hyperobject: these are kind of magic numbery, but they're based on the rendered height of pinned messages and banners on mobile (which don't have nice constants to use, afaict)

shared/chat/conversation/pinned-message/index.tsx
```diff
@@ -137,7 +137,7 @@ const styles = Styles.styleSheetCreate(
         borderColor: Styles.globalColors.black_10,
         borderStyle: 'solid',
         left: 0,
-        position: 'absolute',
+        // position: 'absolute',
```
chrisnojima: 🔥

PR #19989, opened 2019-09-27
Disable chat reacjis if your role isn't high enough to post

buoyad:
cc @keybase/y2ksquad

### DIFF
--- a/shared/chat/conversation/messages/react-button/container.tsx
+++ b/shared/chat/conversation/messages/react-button/container.tsx
@@ -23,7 +23,7 @@ export type OwnProps = {
 
 export type WrapperProps = {
   active: boolean
-  cannotWrite: boolean
+  canWrite: boolean
   count: number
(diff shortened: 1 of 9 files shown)


shared/chat/conversation/messages/react-button/container.tsx
```diff
@@ -23,6 +23,7 @@ export type OwnProps = {
 
 export type WrapperProps = {
   active: boolean
+  cannotWrite: boolean
```
jzila: It looks like you've made `cannotWrite` mandatory everywhere. As such, I think it makes more sense to name it `canWrite`, since the check will otherwise always result in a harder-to-read double negative.

buoyad: Fair enough. I pushed a rebase + that change.

buoyad:
@keybase/react-hackers bump

Issue #9410, opened 2017-11-07
Can't stop automatic starting on Ubuntu 16.04 Linux


ziyaddin:
There are no settings to choose whether I want automatic starting or not. Would be great to add that kind of functionality. Otherwise, there is no way to stop automatic starting of the program.

maxtaco:
cc @oconnor663 for an update on this

oconnor663:
@ziyaddin take a look at `~/.config/autostart/keybase_autostart.desktop`:

```
# Uncomment the following line to disable Keybase autostart:
#Hidden=true
```

Apologies for not having a setting for this in the GUI yet.

ziyaddin:
@oconnor663 Thank you! No problem, it is not a big deal to disable it by using the mentioned method. The only issue is usability.

dgw:
Just wanted to drop a note here, since I found this issue and checked `~/.config/autostart/keybase_autostart.desktop` to find terminal commands instead:

```
# To toggle autostart on, run
# `keybase ctl autostart --enable`
# or to toggle off,
# `keybase ctl autostart --disable`.
# Note that this will overwrite any changes you have made.
```

heronhaye:
In Keybase 5.0, you can change this setting in the GUI on all platforms. Please make a new issue if you are still having trouble. Thanks!

Issue #19224, opened 2019-08-31
Disable autostart on boot


TheBubblePopped:
Where for gods sake is the button to disable autostart on bootup?

heronhaye:
If you're on Linux, try `keybase ctl autostart --disable`

tobypinder:
There are workarounds (as @heronhaye suggests) but every other Electron-like app I use has this toggle in the options. It becomes an impediment to rolling this out within an organisation if we have to tell people to follow complex post-install tasks from the terminal.

TheBubblePopped:
I uninstalled it before getting this answer and i will most likely never come back just for the reason @tobypinder pointed out. 
My though was that i alrdy spend too much time figuring out. 
But thank you heronhaye for answering anyway.

TheBubblePopped:
Keybase ctl autostart --disable didn't work on a friends pc. His keybase got uninstalled as well now. 
!Warning Permission denied opening guo config file!

heronhaye:
@TheBubblePopped It'd be helpful to do a `keybase log send` and also `cat ~/.config/autostart/keybase_autostart.desktop` to help us debug the problem.

heronhaye:
In Keybase 5.0, you can change this setting on all platforms. Please make a new issue if you are still having trouble. Thanks!

heronhaye:
As of Keybase 5.0, you can change the autostart settings in the GUI in the Settings tab on all platforms. Please make another ticket if you're still having trouble. Thanks!

Issue #13995, opened 2018-09-29
Allow to turn off autostart WITHOUT login.


tebruno99:
I am not currently ready to login with my account on this machine. Please allow me when I click on "SETTINGS" to actually view the Settings screen instead of trying to force me to login. This way I can turn of Autostart until I am actually ready.

my log id: 741dca3622dc2a26b429191c

heronhaye:
In Keybase 5.0, you can change this setting on all platforms. Please make a new issue if you are still having trouble. Thanks!

heronhaye:
As of Keybase 5.0, you can change the autostart settings in the GUI in the Settings tab on all platforms. Please make another ticket if you're still having trouble. Thanks!

tebruno99:
The issue is that you cannot do so before you login. Is that possible now on all platforms?

heronhaye:
Sorry, I misread. That is a good idea. On Linux this is possible with `keybase ctl autostart` in the CLI.

PR #19991, opened 2019-09-27
WIP: Linux start on login GUI item

taruti:
Still need to manage setting in guiConfig properly (or do rpc on startup)

### DIFF
--- a/shared/actions/platform-specific/index.desktop.tsx
+++ b/shared/actions/platform-specific/index.desktop.tsx
@@ -9,7 +9,7 @@ import path from 'path'
 import {NotifyPopup} from '../../native/notifications'
 import {execFile} from 'child_process'
 import {getEngine} from '../../engine'
-import {isWindows, socketPath, defaultUseNativeFrame} from '../../constants/platform.desktop'
+import {isLinux, isWindows, socketPath, defaultUseNativeFrame} from '../../constants/platform.desktop'
 import {kbfsNotification} from '../../util/kbfs-notifications'
 import {quit} from '../../desktop/app/ctl.desktop'
 import {writeLogLinesToFile} from '../../util/forward-logs'
@@ -404,11 +404,24 @@ const setOpenAtLogin = async (state: Container.TypedState) => {
     },
   })
 
-  if (!__DEV__ && SafeElectron.getApp().getLoginItemSettings().openAtLogin !== openAtLogin) {
-    logger.info(`Login item settings changed! now ${openAtLogin}`)
-    SafeElectron.getApp().setLoginItemSettings({openAtLogin})
+    if(isLinux) {
+      const enabled = (await RPCTypes.ctlGetNixOnLoginStartupRpcPromise()) === RPCTypes.OnLoginStartupStatus.enabled
+      if (enabled !== openAtLogin)
+        await setNixOnLoginStartup(openAtLogin)
+    } else {
+    if (SafeElectron.getApp().getLoginItemSettings().openAtLogin !== openAtLogin) {
+      logger.info(`Login item settings changed! now ${openAtLogin}`)
+      SafeElectron.getApp().setLoginItemSettings({openAtLogin})
+    }
   }
 }
+
+const setNixOnLoginStartup = async (enabled: boolean) => {
+  RPCTypes.ctlSetNixOnLoginStartupRpcPromise(enabled).catch(err => {
+    console.info(`Error in sending ctlSetNixOnLoginStartup: ${err.message}`)
+  })
+}
+
 
 export const requestLocationPermission = () => Promise.resolve()
 export const clearWatchPosition = () => {}
--- a/shared/settings/advanced/index.tsx
+++ b/shared/settings/advanced/index.tsx
@@ -101,7 +101,7 @@ const Advanced = (props: Props) => {
           </Kb.Box>
         )}
         {isLinux ? <UseNativeFrame {...props} /> : null}
-        {!Styles.isMobile && !isLinux && (
+        {!Styles.isMobile && (
           <Kb.Box style={styles.checkboxContainer}>
             <Kb.Checkbox
               label="Open Keybase on startup"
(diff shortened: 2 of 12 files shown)


jzila:
@taruti is this still WIP?

@heronhaye PTAL

heronhaye:
Seems fine. Looks like it still needs the gui checkbox work.

heronhaye:
@taruti LMK when you want me to review

taruti:
Ready for you in ~60min.

shared/settings/advanced/index.tsx
```diff
@@ -101,7 +101,7 @@ const Advanced = (props: Props) => {
           </Kb.Box>
         )}
         {isLinux ? <UseNativeFrame {...props} /> : null}
-        {!Styles.isMobile && !isLinux && (
+        {!Styles.isMobile && (
```
heronhaye: Don't understand how props.openAtLogin is set on GUI start

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -404,12 +404,24 @@ const setOpenAtLogin = async (state: Container.TypedState) => {
...
+    if (SafeElectron.getApp().getLoginItemSettings().openAtLogin !== openAtLogin) {
+      logger.info(`Login item settings changed! now ${openAtLogin}`)
+      SafeElectron.getApp().setLoginItemSettings({openAtLogin})
+    }
   }
 }
 
+const setNixOnLoginStartup = async (enabled: boolean) => {
+  RPCTypes.ctlSetNixOnLoginStartupRpcPromise({enabled}).catch(err => {
+    console.info(`Error in sending ctlSetNixOnLoginStartup: ${err.message}`)
```
buoyad: `logger.warn`

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -404,12 +404,24 @@ const setOpenAtLogin = async (state: Container.TypedState) => {
     },
   })
 
-  if (!__DEV__ && SafeElectron.getApp().getLoginItemSettings().openAtLogin !== openAtLogin) {
-    logger.info(`Login item settings changed! now ${openAtLogin}`)
-    SafeElectron.getApp().setLoginItemSettings({openAtLogin})
+  if (isLinux) {
```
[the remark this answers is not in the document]
buoyad: yes, we shouldn't set these if `!__DEV__`

shared/settings/advanced/index.tsx
buoyad: @heronhaye  In `actions/platform-specific/index.desktop` it is loaded and put in the store, then this component can get it.

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -9,7 +9,7 @@ import path from 'path'
 import {NotifyPopup} from '../../native/notifications'
```
buoyad: I think we need to change `initializeOpenAtLogin` in this file as well

shared/actions/platform-specific/index.desktop.tsx
```diff
@@ -404,12 +404,24 @@ const setOpenAtLogin = async (state: Container.TypedState) => {
     },
   })
 
-  if (!__DEV__ && SafeElectron.getApp().getLoginItemSettings().openAtLogin !== openAtLogin) {
```
heronhaye: Sorry I missed the configGuiSetValueRpcPromise part above this. This doesn't work because the user could change their setting in the CLI which doesn't set this with `keybase ctl autostart`. On Linux can we derive the initial state from the file itself instead of this config value? And we don't need to set it either

taruti: Is it ok to do RPC to service when the initialisation function runs? Is the service running and accepting RPC at that point? Confused because comments in e.g. `loadWindowState`.

heronhaye: Can you do it when the settings tab/advanced tab is open? cc @buoyad

heronhaye: For example see loadLockdownMode

buoyad: You're right @taruti it's probably not safe. @heronhaye's suggestion of loading when settings/advanced is opened sounds good.

taruti: I have been thinking over this quite a lot.

Currently we support the cli non-advertised ctl only on Linux. Is there a reason why we want to support it on Linux in a different fashion from Windows and Mac? I can do the code which reduces the race between GUI and the cli command, but not sure if it is really worth the added Linux only complexity.

heronhaye: I can't comment on how much more complex it is (cc @buoyad) but it seems reasonable that the implementation should be different if the inner workings are (xdg vs whatever the electron startup manager does).

buoyad: We can call the RPC on linux when settings/advanced is opened and use that to populate `state.config.openAtLogin`

taruti: Ok, updated the PR to do that. Do you want first to look at the latest changes to the PR, or have it rebased (conflicts)?

shared/actions/platform-specific/index.desktop.tsx
taruti: It should be fine now, the correct value gets loaded when opening Settings.

shared/actions/platform-specific/index.desktop.tsx
buoyad: Now it calls the RPC on startup, is the connection ready when starting the app up normally?

buoyad: Oh I misread that, nevermind. LGTM

jakob223:
🎉

taruti:
Rebased to master to solve conflicts.

heronhaye:
Tried this out. Seems to work. Thanks @taruti !

heronhaye:
Partially fixes https://github.com/keybase/client/issues/9410, https://github.com/keybase/client/issues/13995, https://github.com/keybase/client/issues/19224, https://github.com/keybase/keybase-issues/issues/3034.

PR #19992, opened 2019-09-27
Always return an error if Go didn't handle the message

MarcoPolo:
@mmaxim / @jzila 

I think it was possible if the unbox failed that we would drop the notification. This was because we didn't return an error even though the go side didn't handle it.

I changed it so all flows return an error if it failed to display the notification. This way Java knows if Go called the notification and it should mark the notification as seen.

(aside, the seenNotification deduping logic should probably be moved to the go side)


MarcoPolo:
Thanks, I'll wait for part 2 to get merged before merging this. https://github.com/keybase/client/pull/20004

PR #20001, opened 2019-09-27
display available bot commands when listing bot member settings; get …

joshblum:
…public bot commands


joshblum:
i think it's cleaner this way, UpdateCommands has to be called and needs a convID, seems easier with the current flow

joshblum:
(ci failure unrelated)

